PR #55, opened 2013-05-28
Initial implementation of turn restriction support

kschrab:
Here's my attempt to let graphhopper support turn costs/restrictions (turn cost = time a turn would cost, turn restriction = infinite turn cost, turn is prohibited). 
At the moment only simple turn restrictions will be imported. You can enable turn restriction support by `GraphHopper.enableTurnCosts()` and you need to re-import your osm files. If you want to use turn restrictions together with contraction hierarchies you must decide on preparation time if and which kind of turn cost calculation you want to use. Once the graph has been prepared for contraction hierarchies you cannot switch to another turn cost calculation. This behavior should be familiar since you need to decide on preparation time as well which EdgePropertyEncoder and which WeightCalculation you want to use.

Turn costs are stored in a separate storage. For each turn cost entry the edge id from which the turn goes, the edge id to which the turn goes and the costs of the specific turn is stored. The costs are encoded by TurnCostEncoder, which allows to store separate costs for car and bike in one integer value. When searching for shortest routes only restrictions will be considered, when searching for fastest routes restrictions AND turn costs will be considered. For pedestrians no turn costs or restrictions will be considered. 

Feel to try it out. I will add some more details to my implementation when I have a bit more time, maybe tomorrow.


jsroyal:
image

##### Vehicles #####


# Possible options: car,foot,bike,bike2,mtb,racingbike,motorcycle (comma separated)
# bike2 takes elevation data into account (like up-hill is slower than down-hill) and requires enabling graph.elevation.provider below

graph.flag_encoders=truck|turn_costs=true


# Enable turn restrictions for car or motorcycle. 
# Currently you need to additionally set prepare.ch.weightings=no before using this (see below and #270)
# graph.flag_encoders=car|turn_costs=true




##### Elevation #####


# To populate your graph with elevation data use SRTM, default is noop (no elevation)
# graph.elevation.provider=srtm


# default location for cache is /tmp/srtm
# graph.elevation.cache_dir=./srtmprovider/


# If you have a slow disk or plenty of RAM change the default MMAP to:
# graph.elevation.dataaccess=RAM_STORE



#### Speed, hybrid and flexible mode ####


# By default the speed mode with the 'fastest' weighting is used. Internally a graph preparation via
# contraction hierarchies (CH) is done to speed routing up. This requires more RAM/disc space for holding the
# graph but less for every request. You can also setup multiple weightings, by providing a coma separated list.
#prepare.ch.weightings=fastest


# Disable the speed mode. Should be use only with routing.max_visited_nodes or when the hybrid mode is enabled instead
prepare.ch.weightings=no


# To make CH preparation faster for multiple flagEncoders you can increase the default threads if you have enough RAM.
# Change this setting only if you know what you are doing and if the default worked for you.
# prepare.ch.threads=1


# The hybrid mode can be enabled with
# prepare.lm.weightings=fastest

# To tune the performance vs. memory usage for the hybrid mode use
# prepare.lm.landmarks=16

# Make landmark preparation parallel if you have enough RAM. Change this only if you know what you are doing and if the default worked for you.
# prepare.lm.threads=1


# avoid being stuck in a (oneway) subnetwork, see https://discuss.graphhopper.com/t/93
prepare.min_network_size=200
prepare.min_one_way_network_size=200


##### Routing #####


# You can define the maximum visited nodes when routing. This may result in not found connections if there is no
# connection between two points wihtin the given visited nodes. The default is Integer.MAX_VALUE. Useful for flexibility mode
# routing.max_visited_nodes = 1000000


# If enabled, allows a user to run flexibility requests even if speed mode is enabled. Every request then has to include a hint routing.ch.disable=true.
# Attention, non-CH route calculations take way more time and resources, compared to CH routing.
# A possible attacker might exploit this to slow down your service. Only enable it if you need it and with routing.maxVisitedNodes
# routing.ch.disabling_allowed=true


# If enabled, allows a user to run flexible mode requests even if the hybrid mode is enabled. Every such request then has to include a hint routing.lm.disable=true.
# routing.lm.disabling_allowed=true

# Control how many active landmarks are picked per default, this can improve query performance
# routing.lm.active_landmarks=4


# You can limit the max distance between two consecutive waypoints of flexible routing requests to be less or equal
# the given distance in meter. Default is set to 1000km.
routing.non_ch.max_waypoint_distance = 1000000


# You can permanently block certain areas from routing
# e.g. you can block certain areas with 
# block_area=lat1,lon1,lat2,lon2


##### Web #####

# if you want to support jsonp response type you need to add it explicitely here. By default it is disabled for stronger security.
# web.jsonp_allowed=true



##### Storage #####


# configure the memory access, use RAM_STORE for well equipped servers (default and recommended)
graph.dataaccess=RAM_STORE


# will write way names in the preferred language (language code as defined in ISO 639-1 or ISO 639-2):
# datareader.preferred_language=en


# Sort the graph after import to make requests roughly ~10% faster. Note that this requires significantly more RAM on import.
# graph.do_sort=true
### 


Sorry If I am not getting correct.
But the out is still not valid on  it.
Can you please help?

PR #99, opened 2013-08-22
Initial implementation of GTFS support

initdch:
It is a simple implementation to support public transport and GTFS import. I tested it with Dijkstra and real world data and it works very well.

The `GTFSReader` uses GraphStorage to generate a time-expanded graph from a gtfs file. The time is encoded in the graph structure. Therefore, for each station multiple nodes are created which represents the station at a given time. The distance between nodes are the time difference in seconds. 

For GTFS import the onebusaway-gtfs library is used. But trips which uses frequencies.txt are not yet supported.

 `GHPublicTransit` is modeled after `GraphHopper` and  can be used as high level interface. 
 There is also `PublicTransitMiniUI` which can be used to test and debug GTFS files.

### DIFF
--- a/core/files/travis-build.sh
+++ b/core/files/travis-build.sh
@@ -1,7 +1,7 @@
 HOME=$(dirname $0)
 cd $HOME/../..
 
-modules="core web tools"
+modules="core web tools simple-transit"
 for module in $modules; do
   cd $module
   echo "====== INSTALL $module ====="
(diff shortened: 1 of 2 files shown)


dtamm:
Any intentions on reloading this pull request? I am very interested in working with this kind of code (maybe even contribute to it).

karussell:
We rely here entirely on the community and this PR is very old. But we highly appreciate all contributions.

dtamm:
Then i will try myself with implementing the old files (which were coded for graphhopper-0.1.1) for the latest version of graphhopper and test it with some GTFS Data i got.

karussell:
Thanks! If you need help, just ask on the mailing list

dtamm:
Thanks, will do! Already tried myself on implementing the files this morning for graphhopper-0.4.1 and got some errors (due to missing classes)... will list them all and ask for help on the mailing list (after some tries of myself).

PR #209, opened 2014-05-15
Calculate and display the overall incline and decline

ratrun:
in order to smooth the input elevation data, which currently sometimes is not smooth at all.


ratrun:
After a long break I included the suggestion. Now the pull request contains only ascend and descend calculation as suggested. My tests showed that the elevation low pass filtering on reading SRTM is still not sufficient for calculating an overall ascend and descend. See https://graphhopper.com/maps/?point=48.079914%2C15.968628&point=47.995436%2C16.387482&vehicle=bike&elevation=true&layer=Omniscale 
Therefore I included a second filter. Initially I also changed the original elevation values into the  smoothed values. But I reverted that, because this would mess up all the existing SRTM tests and probably make things worse, as the smoothing algorithm also slightly changes the values of the initial and the last value.
In case that this request gets included, please note somewhere in the release notes that this calculation is a limited guess only and cannot calculate exact values due to principle reasons. Otherwise we probably cannot avoid endless discussions of the results, see http://regex.info/blog/2015-05-09/2568 . 
I didn't make detailed tests of the results based on a reference route, where the ascend and descend values are exactly known, because i don't know where I could find such an example on at least 100km distance case, with multiple mountain segments are included. Instead, I just compared the results with some segments I'm familar with and here the results look plausible. Please note that I'm living in Austria, where we have a lot of hills and mountains and I therefore don't care about 50 meters more or less.
In case that somebody finds a better filter and can prove that the results improve with it, then we always can exchange the filter into the improved one.

karussell:
Thanks - would still love to see two issues. Please would you mind to separate the filtering in another PR? Then we can easier integrate this without fear and polish the filtering/smoothing later.

For this PR I would like to avoid mixing this bike specific info in the general translation info. Please create two new translation keys e.g. "ascent %$1" and "descent %$1". Furthermore the GHResponse now contains a [hints field](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/GHResponse.java#L42) where I currently would like to prefer the values as they are not usable for all vehicles.

Thanks for the link too, interesting read.

ratrun:
I do not see a reason why separating the filtering would bring benefits. I know that without the filtering the calculated results are worse compared with the filtering. What do you fear? The filtering is only used for the calculation of the ascend and descent, so it cannot destroy/influence any existing functionality. Therefore I would like to keep it included.

As you again suggested usage on the translation keys I will create them. I had hoped that you forgot about it in the meantime ;)

I think that I understood the hints field on the java side. But I do not understand your plan for the java script side. Here I only found the usage as an error indication in this line: [main.js#L982] (https://github.com/graphhopper/graphhopper/blob/master/web/src/main/webapp/js/main.js#L982)

Could you please provide some more hints for the hints usage on the java script client side? I problably overlooked something.

karussell:
I would like to have this feature and feel the smoothing is not ready. But with 'not ready' I more mean that it could be applied in more scenarios like while import so that not only bike benefits.

> What do you fear? The filtering is only used for the calculation of the ascend and descent, so it cannot destroy/influence any existing functionality. Therefore I would like to keep it included.

I don't like to have it in an inconsistent manner. Smoothing should be done earlier if it is of an advantage.

Re the translation key: do not include the english translation, just provide EN (+DE) and the other languages fill fall back to english. I can do this, plus the JS side (when I finally find a bit time ;()

ratrun:
I didn't know about the automatic Engish language fall-back. I removed all the translation strings (except EN+DE) as proposed.

With a heavy heart I also included the requested elimiation of the smoothing. Now the patch is quite small.

Please note following commented test lines in GraphHopperIT.java:
This was the testresult with smoothing:
```
        //assertEquals(8, rsp.getAscendMeters(), 1e-1);
        //assertEquals(59, rsp.getDescendMeters(), 1e-1);
```
This is the testresult without the smoothing:
```
        assertEquals(84, rsp.getAscendMeters(), 1e-1);
        assertEquals(135, rsp.getDescendMeters(), 1e-1);
 ```
for 

https://graphhopper.com/maps/?point=43.730729%2C7.421288&point=43.727697%2C7.419199&vehicle=foot&elevation=true&layer=Omniscale

In case that you really want to have it this way, then I'll make one more commit and delete the two commented test results.

I don't think that this change is bike/foot related. The results are also of interest for other vehicles, even for car. 
I would ask you to make the JS side change as proposed.

**Comment by karussell** [comment]
Thanks, will look into it next week and add the JS where necessary!

> With a heavy heart I also included the requested elimiation of the smoothing. 

But please not without creating a new separate issue with the smoothing only :)

I want to highlight that I don't criticize you or anything, it is even the opposite that GraphHopper lives from many such improvements with this 'love for the detail'. Now for this case I just wanted to make sure that two independent issues get its own dedicated discussion and improvement. Also having smaller pull requests makes it easier to review, integrate, test and deploy.

File: core/src/main/java/com/graphhopper/util/PathMerger.java [hunk 1]
```diff
@@ -106,10 +130,29 @@ public void doWork( GHResponse rsp, List<Path> paths, Translation tr )
         {
             String debug = rsp.getDebugInfo() + ", simplify (" + origPoints + "->" + fullPoints.getSize() + ")";
             rsp.setDebugInfo(debug);
+            if (fullPoints.is3D)
+            {
+                int size = fullPoints.getSize();
+                double eleinArr[] = new double[size];
```
karussell: I would avoid creating a new array here, instead pass `fullPoints` directly to calcAscendDescend

As the smoothing will happen on import (and done in a separate issue). BTW: see also #507 where this then should be placed

karussell: I'll do ...

karussell:
@ratrun As I just saw the commit where you noted the GPL code for smoothing: would you please just provide a single commit for the ascend+ascend stuff without any smoothing? This is really important to keep GraphHopper free of licensing issues.

ratrun:
I used the GPL code just for experiments and replaced it later, no part of it is active any longer. I'll try to combine the changes into a single commit.

karussell:
Okay, thanks!

Just for the future: commiting GPL code into the repository should not be done. Also when looking for inspiration (aka implementing) then GPL code must not be used.

ratrun:
I'm sorry for the troubles. Now I think that I have tackled every comment. Please take a look. As suggested I didn't try  to implement the JS hint stuff.

karussell:
Sorry for pinging here again. Can you provide just one commit where the commits where you eliminated code are removed?

ratrun:
I'm still sometimes struggling with git. Is it OK now?

karussell:
Thanks and no problem :)
Now I can take it and rebase this one commit ... [this here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) has helped me to understand a bit more.

To avoid the merge commits you should do `git pull --rebase` instead of `git pull` ... BUT do this only if you know what you are doing and never push force into a none-feature branch :)

karussell:
Please have a look into [my changes](https://github.com/graphhopper/graphhopper/commits/issue_209_2). I added the translation stuff to the google doc and made an automatic update of the translation files, so there are these many changes. But actually we don't need this due to the simple arrows I use in the web UI. Let me know if you think we should remove the text or keep it.

ratrun:
Thank you. Looks good. As said, I totally agree to get rid of the translation stuff. I always wanted to have it that way and only included the translation keys because you insisted on them. But this is much more easy to understand in every language.

Now there is only one major problem: the missing smoothing and my algorithm, which cannot be used for smoothing during reading, because it messes up every start and end elevation. Do you already have an idea how to solve this? I think that a good 2D smoothing algorithm should have all evlevation data for every point available. When you do the smoothing during reading then I beleive that one probably needs a 3D smoothing algorithm. Anywhow, 3D smoothing is beyond my mathematical skills.

karussell:
Ok, I'll get rid of the translations ... re the smoothing: let us discuss this in the other issue.

karussell:
Hmmh, the translations are useful for navigational purposes .. I'll keep it but'll remove the 'web' prefix

karussell:
Is now also deployed.

PR #245, opened 2014-08-18
more none allowed ways for bycicle and foot

jotpe:
l know you wont get routed by bike over highways which are tagged with cycleway=none.


karussell:
Sorry, I do not remember the state of this PR. Can we close this and you wanted to open different ones regarding the bus and military access?

karussell:
Closing as I think that the original access restriction is not appropriated and should be done via 'soft constraints' (priorities). Please reopen if required or create a new.

Issue #111, opened 2013-10-09
More efficient graph storage via separating shortcuts from edges


karussell:
Separating shortcuts from edges could have some advantages:

 * This would reduce storage size. Because some fields are not necessary for shortcuts and some not for edges: no edgeIds in normal edges, no dynamic array for shortcuts necessary, no wayGeometry pointer for shortcuts, ...
 * related to #339
 * it would be possible to support multiple vehicles for CH -> multiple prepare calls

To do this we would need to think about some points:

 * probably we should store shortcuts different than edges as they do not need to be appended later on. E.g. just one block of edges per node
 * Having multiple profiles in one graph for the world wide case is extreme, so e.g. in the near future the edgeId limit of 2 billion is reached a lot quicker than with one profile per graph. Not if we have multiple, separate shortcut lists (every shortcut list forms a separate LevelGraph with the normal edges)

karussell:
The refactoring for this will be huge and can be summarized as 'use composition instead of inheritance':
 * there will be a class `GraphHopperStorage implements GraphStorage` with methods: `public Graph getBaseGraph()` and `T getGraph( Class<T> clazz )` returning e.g. the LevelGraph, later on we can also provide different Weighting to get specific LevelGraph
 * the DataAccess objects will be moved from GraphHopperStorage into the BaseGraph class (`BaseGraph implements Graph`)
 * the LevelGraphImpl (`LevelGraphImpl implements LevelGraph`) can also handle its DataAccess objects
 * LevelGraphStorage will be removed
 * Some methods of the base graph (like ensureNodeIndex) needs to be propagated to the LevelGraphs, this will be done with a listener mechanic
 * this opens up the possibility to register other graph extensions like we currently have with the TurnGraphExtension but more dynamically and not e.g. just one. Also the GraphExtensions would then be decoupled from the Graph and reside in the GraphHopperStorage (and should be probably called StorageExtension, not sure)

So this will not only make new things possible (multiple CHs) but also improve the architecture slightly as storage methods (create, loadExisting, flush, ...) are then separated from the traversal ones.

Currently I'm working on it to make it compile and one question I stumbled over is: the separation leads to the following uglier code:
```java
GraphHopperStorage ghStorage = new GraphHopperStorage(..);
Graph g = ghStorage.getBaseGraph();
g.edge(...)
```

instead of just:

```java
GraphHopperStorage g = new GraphHopperStorage(..);
g.edge(...)
```

[...]

PR #278, opened 2014-10-31
ExtendedStorageManager for an arbitrary amount of ExtendedStorage

b3nn0:
Hi,
This introduces an ExtendedStorageManager, which, by itself, is also an ExtendedStorage and can handle an arbitrary amount of sub-storages.
In addition it introduces an ExtendedStorageAccess (similar to NodeAccess) interface (and implementation).
To save space in common situation, during graph creation, one can specify either
a) a normal ExtendedStorage, in this case, it will mainly matche the old behaviour of storing ints directly in the graph or
b) an ExtendedStorageManager. In this case, the in-graph-extended-storage will point to the ExtendedStorageManager, which will then contain the ints.

It is possible to only use a subset of the ExtendedStorages that were provided during Graph creation. The ExtendedStorageManager will then figure out what to do, with the help of some properties that were written to the "properties" file.

The GHExtendedStorageAccess is able to correctly determine if it is working on a normal ExtendedStorage object or if it is working on an ExtendedStorageManager and will act accordingly.

A couple of tests were added to make sure that the ExtendedStorageManager and Access class work as intended.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -788,7 +788,7 @@ public class GraphHopper implements GraphHopperAPI
 
         if (encoder.supportsTurnCosts())
         {
-            result = new TurnWeighting(result, encoder, (TurnCostStorage) graph.getExtendedStorage());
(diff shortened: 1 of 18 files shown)


karussell:
According to email discussion: is there a newer version of this and any news about the CLA signing?

karussell:
Due to major refactorings happening since them and easier structure after #111 I'm closing this for now. Feel free to update or resurrect.

PR #298, opened 2015-01-10
Partially fixes 'layer not updated in URL' #266

### LINKED ISSUE #266: web interface: permalink URL not updated when change in layer
when watching the map via graphhopper.com/maps/ I can change the map layer in the upper right corner from Lyrk to osm.org, osm.de and others.

When doing so, the permalink in the browser's URL row does NOT change, it remains at layer=lyrk !!

Same is with the little symbol for permalink under the place search fields on the left screen.

PS: It seems that Lyrk layer is a bit outdated? any information from that tile provider??

drnextgis:
Updating permalink when base layer was changed


karussell:
Thanks a lot again! While this fixes the perma-link issue the URL still stays the same, do you have an idea for this?

PR #306, opened 2015-01-19
Maven version updated to 3.2.5

Haves1001:
I just changed the the version of maven to 3.2.5 as 3.2.3 is no longer available und the listed address. If 3.2.3 is required amother host should be selected.


karussell:
Thanks!

PR #312, opened 2015-01-29
fastest weight calculation should return seconds

zhiqiangZHAO:
it's just a small issue.
The speed is in KPH. the distance is meter.
weight should be in seconds.

Ah... almost all the tests failed.


karussell:
Thanks, will look into it! (although it is not important for the correctness and not really a bug but is important to fix)

zhiqiangZHAO:
Hi peter

agree with you in some extent.
But it confuse people.
And in the future, if the weight is combination of time and other term, like co2 emission, it will bring problem. Thanks.

karussell:
Well thats for sure if the documentation also says it is in seconds ;)

Issue #94, opened 2013-07-25
Wrong and missing turn instruction


karussell:
in this situation:

http://graphhopper.com/maps/?point=52.006528%2C4.372382&point=52.006908%2C4.369898

Graphhopper is not showing a 'sharp left turn', possibly because the from and to ways are the same (Poortlandplein, since that's the name of the square)

Here are other examples of missing instructions: 
- http://graphhopper.com/maps/?point=49.76661%2C43.648478&point=49.765854%2C43.658381&vehicle=FOOT
- http://graphhopper.com/maps/?point=48.614598%2C8.033752&point=48.61664%2C8.021629&vehicle=bike&elevation=true

Another improvement would be (create another issue?) to use a 10 or 20 meter tolerance to calculate the instructions. E.g. then the 'slight left turn' will be a 'left turn' [here](https://graphhopper.com/maps/?point=49.237356%2C8.091838&point=49.238273%2C8.092965&vehicle=bike&elevation=true&layer=Lyrk)

A further improvement is to add a name from the relation if no street name exists. E.g. [this route](https://graphhopper.com/maps/?point=51.231451%2C-0.503225&point=51.232419%2C-0.487347&vehicle=foot&locale=en-US&elevation=true&layer=OpenStreetMap.de) should show 'North Downs Way'

An other simpler idea (instead of junction analysis) would be to use changing OSM way ID to detect when to print instructions

We should attach names of the main road if there is a parallel cycle or foot path without a name.

Question:  I noticed that any time there is a fork in the road, the direction given is simply "Continue onto [highway name]". Would it be possible to specify whether to take the right or left side of the fork? (Also this is especially problematic at exits, which have no street name; the direction given is simply "Continue" and one must know that that means "Take the exit".)

jansoe:
To summarize my recent experience and this discussion on this topic, I think there are several fixes which could improve turn instructions:
+ if both incoming and outgoing edges have no streetname, instructions are completly ommitted (e.g. https://graphhopper.com/maps/?point=48.614598%2C8.033752&point=48.61664%2C8.021629&vehicle=bike&elevation=true)
+ if you first have to cross a link between the two opposing lanes before turning left, then entering the link dominates the left turn (e.g. https://graphhopper.com/maps/?point=52.508392%2C13.326287&point=52.509456%2C13.325692&layer=Lyrk)
[...]

PR #316, opened 2015-02-03
Added Roundabout Instruction

jansoe:
This introduces extra instructions for roundabouts (i.e. enter roundabout and take exit 3). This is achieved by subclassing Instruction as RoundaboutInstruction. 
Furthermore every instruction now contains its angle (radian).

Things still to improve:
* Better description by splitting insturctions for entering and leaving roundabout (in translation document)
* Improve angle calculation by considering slip roads (relates to #94)
* Expose full roundabout information to json

Code change has no measurable influence on calculation time of instructions (0.3ms of 3ms for Duesseldorfarea )

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -83,6 +83,13 @@
             <artifactId>osmosis-osm-binary</artifactId>
             <version>0.43.1</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+            <version>20140107</version>
+        </dependency>
+        
     </dependencies>
         
     <build>
--- a/core/src/main/java/com/graphhopper/routing/Path.java
+++ b/core/src/main/java/com/graphhopper/routing/Path.java
@@ -442,6 +442,7 @@ public class Path
                             int sign = Instruction.USE_ROUNDABOUT;
                             prevInstruction = new RoundaboutInstruction(sign, name, annotation,
                                     new PointList(10, nodeAccess.is3D()), 0);
+                            ways.add(prevInstruction);
 
                             if (prevName != null)
                             {
@@ -458,6 +459,7 @@ public class Path
                             } else // first instructions is roundabout instruction
                             {
                                 prevOrientation = ac.calcOrientation(prevLat, prevLon, latitude, longitude);
+                                prevName = name;
                             }
                         }
 
@@ -488,9 +490,6 @@ public class Path
                             .setContinuedStreet(!Helper.isEmpty(name) && (prevName.equals(name)))
                             .setDirOfRotation(delta)
                             .setFinished();
-
-
-                        ways.add(prevInstruction);
 
                         prevName = name;
                         prevAnnotation = annotation;
@@ -567,7 +566,7 @@ public class Path
                         orientation = ac.alignOrientation(prevOrientation, orientation);
                         double delta = (orientation - prevOrientation);
                         prevInstruction.setRadian(delta);
-                        ways.add(prevInstruction);
+
                     }
                     ways.add(new FinishInstruction(nodeAccess, adjNode));
                 }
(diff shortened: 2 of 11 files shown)


karussell:
Thanks! Also for the bunch of tests!

I'll try it tomorrow. Some notes:

 * we need to expose the roundabout information to JSON similar to instructions. As we add the angle to the roundabout instructions we should probably also do for all other.
 * Additionally we have to think how to expose the `clockwise` variable, either we use a separate property OR we use additional constants like `Instruction.USE_ROUNDABOUT_CW`,  `Instruction.USE_ROUNDABOUT_CCW`, `Instruction.USE_ROUNDABOUT_INVALID`? I tend to prefer the first, but would like to discuss.
 * the other points you mentioned are okay if done later
 * to avoid casting to RoundaboutInstruction you can assign it to a temporary variable: `RoundaboutInstruction tmpInstruction = new RoundaboutInstruction` and as a last step do `prevInstruction = tmpInstruction` -> read: casting is often ugly and sometimes even a sign of bad design (but yes, sometimes not possible in a different way).
 * Place the open bracket on a new line (maybe we can create an IntelliJ profile similar to what I have in NetBeans). This is none-standard Java coding style but the way we do it currently in GH
 * As the radian is a field in the Instruction class you can expose it directly there: `Instruction.getRadian`

Also as you have asked to be nit-picky ;)
 * To get the encoder you can grab it more directly, no need for string based stuff: `private final FlagEncoder encoder = new CarFlagEncoder(); private final EncodingManager manager = new EncodingManager(encoder);`
 * make the tests a little more separate - e.g. one method for clockwise roundabout and one for counterclockwise
 * instead `public boolean getContinuedStreet()` the getter for boolean is in Java: `isContinuedStreet()`
 * I would use only positives as variable instead of `unfinished` I would use `finished` as `!unfinished` is too complex for me ;)
 * The inline comment should come *before* somethings happens e.g. instead of 
`int someVar; // some comment` do it in two lines:
```
// some comment
int someVar;
```

Only sometimes I use one line and behind e.g. as you did sometimes to describe the if-else clauses.

devemux86:
:thumbsup:

jansoe:
thanx for the comments.
eventually I also missed some test failure, i will fix it tomorrow, and i adress your comments:
* i agree we should include the angle for all instructions in the json
* by this we could easily code for the clockwise variable: positve radian -> counterclockwise, negative -> clockwise, and if undetermined getRadian throws an error or special value >2pi 
* Unfortunately I can't assign the Roundabout Instruction to a temporary variable because, the instructions "lives" for consecutives iterator calls.
*  getRadian is already there :) and just overwritten in the RoundaboutInstruction class  

So far, more tommorow

karussell:
> eventually I also missed some test failure

Yeah, in RoutingAlgorithmIT there are some tests failing. Maybe some distance addition or points are missing?

> and if undetermined getRadian throws an error or special value >2pi 

undetermined can only be for tagging problems or strange roundabouts we do not know yet, right? I would prefer `Double.NaN`

> Unfortunately I can't assign the Roundabout Instruction to a temporary variable

Yeah, I just meant within the one if to avoid casting in the loop (I can also change afterwards and you see what I meant or maybe even leave it as it is)

> getRadian() is already there :)

ups, perfect :) !

jansoe:
> Yeah, in RoutingAlgorithmIT there are some tests failing. Maybe some distance addition or points are missing?

Are there still some test faiures? For me everything worked fine after fixing a bug when starting in Roundabout.

karussell:
Ah, missed the commit you did yesterday. Looks fine now. Do you want to add json stuff here (or should I?) and merge later or separate issue?

jansoe:
I would adress your comments right now, including json and then push here.
If you prefer to implement the json interface by yourself go ahead.

karussell:
Nice and yes, if you would do the json stuff it would be even better :)

PR #318, opened 2015-02-04
Incorporate filter search into the finding of the closest node search.

crowmagnumb:
The problem is as I discussed about the filtering. It's not a question of does the expanded search find a node or not, it's does the expanded search find a node that passes my filter. Inside of findClosest the findNetworkEntries returns a set of node ids, then we loop through these nodes using the edge explorer and the supplied filter. The basic search does not find a node that passes my filter so findClosest returns a "null" QueryResult. If I try and expand the search further I still don't find my node because no matter what value of maxRegionSearch, findNetworkEntries returns the first set of nodes that it finds but I already know those don't pass my filter, so once again they don't pass my filter and I get a "null" result. 

I've added the boolean to allow the filtering of the edges to be built into the search mechanism. I only added it to maintain backwards compatibility. I honestly don't see why this is not just done automatically.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
+++ b/core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
@@ -17,25 +17,6 @@
  */
 package com.graphhopper.storage.index;
 
-import com.graphhopper.coll.GHBitSet;
-import com.graphhopper.coll.GHTBitSet;
-import com.graphhopper.geohash.SpatialKeyAlgo;
-import com.graphhopper.routing.util.AllEdgesIterator;
-import com.graphhopper.routing.util.EdgeFilter;
-import com.graphhopper.storage.DataAccess;
-import com.graphhopper.storage.Directory;
-import com.graphhopper.storage.Graph;
-import com.graphhopper.storage.NodeAccess;
-import com.graphhopper.util.*;
-import com.graphhopper.util.shapes.BBox;
-import com.graphhopper.util.shapes.GHPoint;
-import gnu.trove.iterator.TIntIterator;
-import gnu.trove.list.array.TIntArrayList;
-import gnu.trove.procedure.TIntProcedure;
-import gnu.trove.set.hash.TIntHashSet;
-import java.util.*;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * This implementation implements an n-tree to get the closest node or edge from GPS coordinates.
@@ -51,7 +32,7 @@ public class LocationIndexTree implements LocationIndex
     private final Logger logger = LoggerFactory.getLogger(getClass());
     private final int MAGIC_INT;
     protected DistanceCalc distCalc = Helper.DIST_PLANE;
-    private DistanceCalc preciseDistCalc = Helper.DIST_EARTH;
+    private final DistanceCalc preciseDistCalc = Helper.DIST_EARTH;
     protected final Graph graph;
     private final NodeAccess nodeAccess;
     final DataAccess dataAccess;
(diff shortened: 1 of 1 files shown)


karussell:
Hmmh, I don't understand your usecase. Can you give me an example of how you use this? E.g. You want to find the closest bus station or something?

Also we need tests for this feature otherwise your use case will be lost or broken when a new refactoring comes in.

Furthermore I would prefer to avoid exploring within findNetworkEntries as this method is just about collecting nodes. So we need some kind of a loop over 'collect & explore & filter the edges' where the filter can also decide if the search should be stopped or not.

karussell:
Hmmh, you are holding more vehicles and location index returns earlier with only 'car edges'?

KenSoutherland:
Well, that is a pretty good example you just gave, and in a sense I am
doing something similar. I have a bunch of utility poles that I have to
visit. For each one I find the edge it is on and thus make a list of all
the edges I need to visit. Then starting with the first I route the user to
that edge, after that I route them to an adjacent edge if there is one in
the list. If not, I then use the findClosest() to find the closest edge
that is in my list of edges that I have to visit and then route them there,
and so on until I have visited all my edges.

KenSoutherland:
I'm not sure what you meant by that.

karussell:
So what you want is the best tour for N points? And what you currently do is a heuristic solution to that problem, it is not optimial, right?

KenSoutherland:
It is absolutely not optimal because N can be as large as say 1000. i.e. very large. It is not possible in our lifetimes to find the optimal solution. What I find is a damn fine route that is pretty darn good. Optimal? No.

KenSoutherland:
And incorporating the explore in the find results in a much more efficient algorithm. If this is not the desired behavior for all cases (again, I really can't see why not, at least from the stand-point of the findClosest() routine) then I propose a separate routine for this.

karussell:
That heuristic should be good, yes.

Still the LocationIndex is the wrong tool for what you want IMO. What you could do is to use Dijkstra with many destinations:

```java
QueryGraph qGraph = new QueryGraph(graph);
qGraph.lookup(yourLocations);
Dijkstra dijkstra = new Dijkstra(qGraph, encoder, weighting, TraversalMode.NODE_BASED) {
   @Override protected boolean finished() {
        // your condition e.g. a set of (virtual) nodes
    }            
};
// TODO currently not public
dijkstra.runAlgo();
```

Or just calculate ALL distances between ALL locations and feed an optimization tool (for <10 you can do brute force search). Calculating all distances or times is easy with our [Matrix API](https://github.com/graphhopper/directions-api/blob/master/README.md#matrix-api) but with plain GraphHopper and two simple for-loops that is often also acceptable fast.

karussell:
The problem with the LocationIndex and your usecase is that there is no order guaranteed, where Dijkstra guarantees that it visits the closer edges first.

karussell:
Still there might be indeed a problem if e.g. your graph contains car and bike streets, now you search for the closest bike street but the findNetworkEntries method returns only nodes exploring car streets (e.g. close to a motorway)

KenSoutherland:
It is not the routing problem I need solved. I have my routing problem and it's solved. What I wrote above was a great oversimplification of the problem, giving you just enough info necessary for this discussion. To say that using Dijkstra to solve my problem is missing the mark.

But you keep getting glimpses of my problem. Your example of bike vs. car streets is exactly the same issue as I am having. If you use my version, that problem goes away.  :)

karussell:
I've fixed this issue - see my new pull request: #488

Slightly more refactoring was necessary. It also should be faster on average now.

@crowmagnumb if my new change is not working as expected in your case, please let me know!

crowmagnumb:
Thanks! Sorry I haven't had time to check as I've sinced changed jobs and have not been working on this for a while. If I get back to that, and some day I hope to, I will definitely check it out.  Cheers.

PR #320, opened 2015-02-08
Support fords

florent-morel:
PR for ford support.
During graph creation, I check config.properties for FlagEncoder allowing crossing fords.

To enable ford crossing for a given FlagEncoder, amend config.properties with:
graph.flagEncoders.supportFord=foot

As for graph.flagEncoders, possible options are: car,foot,bike,bike2,mtb,racingbike,motorcycle (comma separated). A non-existing value will throw an Exception.

Note
I'm building a Map<String, FlagEncoder> mapEncoder to iterate over instantiated FlagEncoder to set BlockFords to false. Not sure the map is the most efficient way though, any suggestion?

Could you check this PR please?
Florent.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -536,9 +536,8 @@ public class GraphHopper implements GraphHopperAPI
         // osm import
         osmReaderWayPointMaxDistance = args.getDouble("osmreader.wayPointMaxDistance", osmReaderWayPointMaxDistance);
         String flagEncoders = args.get("graph.flagEncoders", "CAR");
-        String flagEncodersFord = args.get("graph.flagEncoders.supportFord", "");
         
-        setEncodingManager(new EncodingManager(flagEncoders, bytesForFlags, flagEncodersFord));
+        setEncodingManager(new EncodingManager(flagEncoders, bytesForFlags));
         workerThreads = args.getInt("osmreader.workerThreads", workerThreads);
         enableInstructions = args.getBool("osmreader.instructions", enableInstructions);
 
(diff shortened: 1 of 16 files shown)


karussell:
Thanks! I'll integrate the tests and your additional check.

But regarding the configuration: I do not see a reason to make it that complex. If you need a special setting why not do:

`hopper.setEncodingManager(new SomeEncoder().setBlockFords(true|false))` 

?

I would avoid to add a new method to FlagEncoder for every possible configuration, especially as this `ford` settings could potentially be different per country. One alternative could be to use a similar configuration schema per encoder like we already do for turnCosts:

`parseBoolean(propertiesStr, "turnCosts", false)`

florent-morel:
Hi, 
I changed the way the configuration is read and used a "turnCosts" approach.

We can now use:
graph.flagEncoders=car,foot|allowFords=true

I used boolean in AbstractFlagEncoder, as I'm not sure why there is int for turnCosts.
Could you please explain?

karussell:
Thanks!

Some minor things: Instead of creating an additional parameter in the AbstractFlagEncoder constructor just set the variable within the 'string'-constructors:

```java
this(bla, blup);
this.setBlockFords(parseBoolean(propertiesStr, "blockFords", false))
```

and either use blockFords as variable AND as configuration option or allowFords for both. Also remove `setBlockFords` and `isBlockFords` methods from the FlagEncoder interface. These should stay just public or even better protected methods in AbstractFlagEncoder.

florent-morel:
Hi, 

I'm not sure why the build failed on openjdk7 (1044.1). The other 2 and my local build went fine.

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-jar-plugin:2.4:jar (default-jar) on project graphhopper: Execution default-jar of goal org.apache.maven.plugins:maven-jar-plugin:2.4:jar failed: Plugin org.apache.maven.plugins:maven-jar-plugin:2.4 or one of its dependencies could not be resolved: Could not transfer artifact commons-lang:commons-lang:jar:2.1 from/to central (https://repo.maven.apache.org/maven2): GET request of: commons-lang/commons-lang/2.1/commons-lang-2.1.jar from central failed: Connection reset -> [Help 1]


Could you launch the build again manually to check if it was a temporary issue?

Regards,
Florent.

karussell:
Thanks! This is important as I've seen this before and needed an example for https://github.com/travis-ci/travis-ci/issues/3160 :)

PR #324, opened 2015-02-13
refactored BBox

jansoe:


karussell:
Important little change - thanks :) !

Some minor comments:
 * would you rename BBox.initInverse to BBox.createInverse ?
 * and in calcRouteBBox can you somehow use BBox.initInverse(_fallback.hasElevation())
 * reformat to current formatting rules (space in method args and '{' on newline)

karussell:
This refactoring uncovered a bug that the bounds for the graph is incorrectly stored. The elevation should be stored in the graph - I'll fix!

karussell:
Ok, fixed

karussell:
Turned out there was another bug in Helper methods. Nice and bad ;)

karussell:
And yet another thing which was unclearly defined in GH - thanks @jansoe for uncovering all this :) !

PR #333, opened 2015-02-17
Access GH from PostGIS using pl/python

rajanski:


karussell:
> pl/python function to enable graphhopper routing from postgis.
> Make sure to have created the extension plpythonu beforehand.

Thanks! Would you mind to explain in simple words what it does and what not? and why routing within postgis? Also how to (easily) set everything up e.g. a step by step installation on linux?

devemux86:
Indeed, considering that in Postgres we have also [pgRouting](http://pgrouting.org/), it would be interesting to see the results.

rajanski:
Hi there,

there is not much to set up os-wise, besides installing postgresql-plpython deb package. Then in postgres, simply say "create extension plypythonu". Afterwards, the function will be usable, with two coordinate pairs plus vehicle as inputs.
Best

R

karussell:
Thanks! But I would also like to know about the purpose, why not just use GH as it? Is it to make integration with SQL queries feeling more native?

rajanski:
The purpose is to enable postgres to very easily get route geometries between two points. I personally find this much easier using this function compared to setting up a pgrouting-enabled  database (and most time consumingly, importing the osm data into it )

karussell:
Thanks!

Merged, and moved to core/files/postgis https://github.com/graphhopper/graphhopper/commit/3f9d08032dfd89b3ca4ddf1104ed29071ff23d13

PR #338, opened 2015-02-25
Fix for #116: Make normal algorithms work on graph prepared for CH

### LINKED ISSUE #116: Make normal algorithms work on graph prepared for CH
It could be useful to prepare the graph for CAR but still have FOOT and BIKE vehicles or just CAR with other weighting available. I.e. it would be nice to have the possibility to run CH and normal algorithms (e.g. via setting edge filters)

The work in the branch shows that if we use the filter then CH-queries and preparation are 20% slower compared to disconnecting shortcuts and edges. Unsurprisingly is the number of normal edges going in the wrong direction (downwards) also 20%, i.e. iterating/loading them is waste of time.

This issue will allow us:

 * to improve turn instructions if CH is enabled and for this it is necessary to iterate over original edges only (very important, no space efficient workaround to my knowledge)
 * to support turn costs for CH it could be necessary to run normal algorithm while path extraction (important, but there exists a workaround via storing 'loops' at via point elsewhere)
 * running normal algorithms on a prepared graph (minor importance)

To make this issue complete we do in a [separate branch](https://github.com/graphhopper/graphhopper/tree/issue_116):
 - [x] removed LocationIndexTreeSC
 - [x] graph.getOriginalGraph (make sure no confusion with QueryGraph exists) or another technique to avoid a special edge filter e.g. to use LocationIndexTree with a graph containing shortcuts
 - [x] QueryGraph needs a getOriginalGraph method too (e.g. in Path but also in normal algorithms)
 - [x] make enabling multiple vehicles possible even if CH is enabled. But define which single vehicle should be used for CH -> done via new defaultVehicle which is also used if no vehicle was specified
 - [x] see if we can fix #288
 - [x] an efficient StopEdgeFilter which knows that original edges come before shortcuts and allow an efficient way to iterate over original edges. Does not help much as this is not often the case that there are 0 normal edges
 - [x] later think about reverse approach: always use base graph except level graph uses special 'graph for algorithm' in algo and path
 - [ ] Remove the necessity to do the CH preparation before the index creation. The location index had to be created AFTER the preparation but sometimes we need to change edges based on an location index (traffic data) BEFORE preparation.
 - [ ] after branch is merged (separate issue?): rename LevelGraph stuff to ShortcutGraph as we no longer need to store the levels, just if an edge is up- or downwards -> for the LevelEdgeFilter and only for normal edges as shortcuts are always upwards!

karussell:
This pull requests fixes #116 and does the following

 - it allows executing normal algorithms on a prepared graph (using a shortcut filter, see BaseGraph and Graph.getBaseGraph), although doing so is slower than on unprepared as it needs to filter away the shortcuts. #339 will fix this
 - the LevelEdgeFilter which was necessary before is now called more often (also for normal edges going downward levels) resulting in a slower routing and preparation for CH. This is roughly 18% and will also be fixed with #339
 - multiple vehicles are allowed, but only one is prepared via CH (the graphHopper.setDefaultVehicle or algorithm.defaultVehicle or the lexicographically first vehicle in the graph.flagEncoders list)
 - as normal traversal is allowed on a prepared graph we can now calculate proper instructions, e.g. the roundabout instructions will work for CH too
 - the location index creation is done before preparation which removes the need to create an in-memory index before preparation to feed custom data


devemux86:
Peter graph preparation seems more limited now.

By studying the PR I see in ```config-example.properties```:
"Enable turn restrictions for car (Currently prepare.chWeighting=fastest is required, see below)"
What if we want to prepare non CH graphs with car TR and multiple vehicles?

Also the algorithm.defaultVehicle=car is needed only with CH, right?

karussell:
Graph preparation is not limited. But thanks, that is a mistake :) ! Turn restrictions is only possible for `prepare.chWeighting=no` 

> What if we want to prepare non CH graphs with car TR and multiple vehicles?

'preparation' is only done for CH.

devemux86:
Ok! I chose wrong word, by "prepare" I was meaning "build" graphs.

jansoe:
I guess this comment should be removed now?
https://github.com/graphhopper/graphhopper/blob/issue_116_3/core/src/main/java/com/graphhopper/GraphHopper.java#L1042

karussell:
Yes, thanks!

karussell:
@devemux86 would you check if you also get a dramatical slow down on your devices. I've an old device and seeing now a lot slower than before. Not sure if we should release with that now :(

karussell:
Or if I should fix #339 before .. hmmh

devemux86:
@karussell you mean with CH off ?
Sure just let me integrate the updated code.

karussell:
CH off should be unchanged. CH on will be very slow, still not sure why not only 20% ...

devemux86:
Ok I'll try CH with 1 vehicle.
With more than 1 vehicles the build fails with "Currently only one vehicle is supported if you enable CH".

karussell:
> With more than 1 vehicles the build fails with "Currently only one vehicle is supported if you enable CH"

this should build and import with more vehicles, but only one vehicle (default vehicle) should be picked for CH

karussell:
See e.g. here https://github.com/graphhopper/graphhopper/blob/master/core/src/test/java/com/graphhopper/routing/GraphHopperIT.java#L291

devemux86:
Yes I have studied the new code changes and the docs.
If I put ```algorithm.defaultVehicle=car``` in config properties then the build fails, e.g. for car,foot,bike2
If I omit it the build proceeds.

karussell:
oh, okay. will have a look

devemux86:
Ok your example (foot,car) and default car works.

For the record my settings that fail are:
```
graph.flagEncoders=car|turnCosts=true,foot,bike2
graph.bytesForFlags=8
graph.dataaccess=RAM_STORE
algorithm.defaultVehicle=car
```
and Elevation enabled for bike2

jansoe:
maybe this is the reason https://github.com/graphhopper/graphhopper/pull/338/files#diff-b0eb0b2e6b07deaa542301bbc8904ca9R97

devemux86:
>@devemux86 would you check if you also get a dramatical slow down on your devices. I've an old device and seeing now a lot slower than before. Not sure if we should release with that now :(

@karussell  I have made some tests with CH specifically on older devices and unfortunately you are right.
There are times that I cannot even distinguish when I have CH on or off :(

(Lately I prefer playing with CH off, as I find Bidirectional A* to play nice and of course I can use multiple vehicles, TR and elevation)

karussell:
@jansoe what do you mean?
@devemux86 okay, then we will delay and I'll have to fix the other issue, hopefully it is nothing else somewhere hidden :)

> as I find Bidirectional A* fast

how fast is it for, say a 300km route on a recent device? (I have only a 4 year old samsung ;))

BTW: elevation can be used with CH

karussell:
@devemux86 would you try to compare with a sorted graph? (or do you already have it enabled?) `graph.doSort=true` or `hopper.setSortGraph`

devemux86:
No I didn't used a sorted graph. I'll make more tests.

devemux86:
I tried that and with extremely large routes >500km and even on powerful mobiles (also on Atlas at pc) and I think (unless I'm doing something wrong!) that CH on / off states appear to me with similar performance (without certainly an observable difference).

jansoe:
sorry was in a rush, I just wanted to point to that exception I currently get, but don't yet fully understand.
htps://github.com/graphhopper/graphhopper/blob/issue_116_3/core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java#L97
Will take a closer look tommorow.

karussell:
Ah, okay. Yes that is the same problem Emux has, will take a look.

@devemux86 thanks for the feedback! What is your best way to profile android apps on the device - simply include StopWatch or are there more efficient ways?

devemux86:
Eclipse has device monitors via ADT and AS has also some new monitors.
But about calculation speed I find often more convenient to just use counters like the one you mentioned.

karussell:
Okay, the problem is a simple one: levelFilter is null in PrepareContractionHierarchies.createAlgo IF the graph was already prepared and e.g. loaded from disc. If it was prepared and the routes are executes all was fine.

karussell:
Documented and fixed in #349

karussell:
The other bug @devemux86 mentioned is now reported here (will be fixed soon) #350

karussell:
Please let me know if this is working as expected (and only 20% or less slower), then we'll postpone #339 and release

renauddl:
Le 2/03/2015 11:01, Peter a écrit :
>
> Please let me know if this is working as expected (and only 20% or 
> less slower), then we'll postpone #339 
> <https://github.com/graphhopper/graphhopper/issues/339> and release
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/graphhopper/graphhopper/pull/338#issuecomment-76685304>.
>

Hi Peter,

just curious to know:
Do you still remove higher to lower during the contraction when this 
feature is activated?


*Renaud De Landtsheer, Ir, Phd*
Senior R&D Expert
CETIC
Rue des Frères Wright, 29/3
B-6041 Charleroi
Phone: +32 71 490 754

karussell:
We disconnect higher to lower shortcuts and filter away higher to lower edges.

PR #341, opened 2015-02-26
Android studio move

karussell:
Moving from eclipse for android to Android studio (IntelliJ)


karussell:
Merged

devemux86:
:worried:

karussell:
What is the problem?

devemux86:
Just concern since not all are [willing](https://groups.google.com/forum/#!topic/mapsforge-dev/qAsSpCuWoBg) to move to AS / Gradle.
But there is still the option to use the pom files for the Android sample.

karussell:
Of course, we won't move to gradle for the whole project as I currently do not see wide adoption and integration for gradle (I'm using e.g. NetBeans with good maven support). The downside is we have to sync those dependencies but that is okay for now.

Do you see other problems? Eclipse users can either use gradle, use the pom or even use the created jar files. I just want to get rid of something I don't like that much (I had to many fights with eclipse) with something which is now also the recommended system.

devemux86:
No, as you mentioned with Maven or jars someone can use the library in any IDE and with Gradle in AS (we follow the same couse in Mapsforge too).

devemux86:
By the way you have the ```AndroidDownloader``` class 2 times in Android module.
It's also outside the ```app``` folder, alone inside a ```src``` folder.

karussell:
Thanks, will remove

PR #342, opened 2015-02-26
apply method parameter

jansoe:
changed method to apply method parameter. Other option would be to remove parameter from method, but currently the method is called with the parameter throughout GraphHopper.


karussell:
Thanks

PR #345, opened 2015-02-28
Tuning for bicycles classifications for the sac_scale tag:

ratrun:
Allow highway=cycleway and sac_scale=hiking for every kind of bike.
Disallow sac_scale=mountain_hiking for every bike except for MTB.
Allow highway=* and sac_scale=hiking for every kind of bike, except for the racebike.


karussell:
Thanks! Would you mind to add a test for MTB, as it is not obvious to me how you still allow mountain_hiking :)

karussell:
Ah, okay. There is already a test which is still passing. Nice & thanks!

PR #355, opened 2015-03-05
Skip ways with sidewalk=no.

Rungee:
for example http://www.openstreetmap.org/way/155062104 .

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
@@ -206,6 +206,9 @@ public class FootFlagEncoder extends AbstractFlagEncoder
                 return 0;
         }
 
+        if (way.hasTag("sidewalk", "no"))
+       		return 0;
+       		
         if (way.hasTag("sidewalk", sidewalks))
             return acceptBit;
 


karussell:
That is important, thanks! Still I would avoid complete blocking in these cases and instead just decrease the priority. Would you mind to have a look into the collect method and provide this as a PR?

Rungee:
I assume that this tag was proposed for a simple reason. Walking along such a street can be very dangerous. But, if you insist on avoiding complete blocking, then I could try to move it into the collect method.
By the way, there is also a similar value "none" which is used much more often. Take a look at statistics
http://taginfo.openstreetmap.org/keys/sidewalk#values

karussell:
Probably you are right regarding the meaning but even primary roads are allowed and sometimes necessary to pass if there would be otherwise no way or a big detour. If foot should be blocked there would be foot=no? I'm not a mapper so maybe we need some more community feedback regarding this.

Multiple values can be handled.

karussell:
@ratrun do you have an opinion about this change?

karussell:
@ratrun and @Rungee - can we close in favour of #476?

ratrun:
I would close this here. Complete blocking is too restrictive, especially for devoloping countries.

karussell:
But we have an open issue for the priority handling of sidewalks? (lost a bit the overview)

karussell:
Closing due to #476 (and #463)

PR #356, opened 2015-03-05
Maven, Eclipse improvements / fixes for building

devemux86:
This PR is about some changes for being able to build GraphHopper repository in Eclipse.

- In .gitignore I added some Eclipse specific files

- In Maven for Android a fix for src path was needed or the build fails

- Android can now use compiler compliance level above 1.5 (which has issues with existing code)
I have put 1.6 though we could probably advance it to 1.7

- In order to build the Android sample a needed project.properties file is needed by ADT

If you want I could split the PR in order to better handle / discuss it.


android/app/pom.xml
```diff
@@ -114,7 +114,7 @@
         1. Use src/test/java for JUnit tests that don't call any Android APIs (directly or transitively)
         2. Use src/main/java for JUnit tests that call Android APIs  
         -->
-        <sourceDirectory>src</sourceDirectory>        
+        <sourceDirectory>src/main/java</sourceDirectory>
```
karussell: ah, maybe we can completely remove that now

android/app/project.properties
```diff
@@ -0,0 +1,14 @@
...
+#
+# To customize properties used by the Ant build system edit
+# "ant.properties", and override values to adapt the script to your
+# project structure.
+#
+# To enable ProGuard to shrink and obfuscate your code, uncomment this (available properties: sdk.dir, user.home):
+#proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
+
+# Project target.
+target=android-21
```
karussell: Shouldn't this be the minSdkVersion which is 8?

android/app/pom.xml
devemux86: I could try without it to see if the build succeeds.

android/app/project.properties
devemux86: That's the project build target and it's recommended to use the latest Android version.
There are ways to check the build inside the code for hanlding older version's code.

The versions you mean are minSdkVersion and targetSdkVersion, which are already defined in AndroidManifest.xml

karussell: Why the latest when we want to make sure that it runs on the oldest?

devemux86: That's what Lint tool does in IDEs, it parses the minSdkVersion and makes sure the code can run from that version and beyond.
Besides that if you want to use newer Android features that will appear only in new versions, you could not if you use old build target.

There is the official recommendation here:
http://developer.android.com/tools/projects/projects-eclipse.html
"Select a Compile With API version. This setting specifies what version of the SDK to compile your project against. We strongly recommend using the most recent version of the API."

android/app/pom.xml
karussell: Strange that maven compiled successfully with and without it for me

karussell:
BTW: Thanks!

Do you know why the apk is bigger like 2.4MB wasn't it more like 1MB. Is that due to the new mapsforge or did we forget to exclude a library?

android/app/pom.xml
devemux86: Maybe Eclipse is more strict with the Maven declarations?
I just tested it, it works fine without it at me too.

devemux86: Do you want to create a new complere PR ?

karussell: no, I just removed it in another commit

devemux86:
Actually it's the new Mapsforge.
But I checked the jars in Maven and I see an issue - the map component has more resources that we put in it.
Nice catch, we should handle it!

karussell:
I rebased and merged your PR - thanks! https://github.com/graphhopper/graphhopper/commit/6d5b98cb62ebab825fdcd298391965175e0f38ad

> the map component has more resources than we put in it.

You mean on the mapsforge side?

devemux86:
Yes, if you check mapsforge-map-0.5.0.jar in [Mapsforge server](http://ci.mapsforge.org/job/release-0.5.0/) you'll see that it should be only 259KB, not 1.3MB.

devemux86:
Thanks Peter!

Issue #360, opened 2015-03-10
Build Android directly on Travis


karussell:
It looks like it is possible to build the android apk on [travis](http://docs.travis-ci.com/user/languages/android/) but only for 'pure' Android projects.

Can we use the following workaround? https://groups.google.com/forum/#!searchin/travis-ci/android$20/travis-ci/3pzCgvxWZAo/LrSckrasFnkJ

karussell:
Asked [here](https://github.com/travis-ci/travis-ci/issues/3366) if there is a more official way of doing this

karussell:
Closing in favour of #803

PR #363, opened 2015-03-12
Android build target (Eclipse): upgrade to Android 5.1 (API 22)

devemux86:


karussell:
Okay, maybe we can have #360 before as it looks they do not support 22?

devemux86:
The change is for an Eclipse file (usable locally by the users).

The CI servers use their own repository and as you said they need to support the latest versions.
(e.g. I upgraded today Mapsforge too and we need to update our CI server too)

karussell:
I see, thanks! Do you know how to define that only a subproject is of type 'android' :) ?

devemux86:
Do you mean in CI ?

karussell:
Yes, for travis specific ... I do not see an option that we could use

devemux86:
We use Jenkins CI and Ludwig manages the server.
So I'm afraid I cannot help a lot here.. :(

karussell:
No problem I think I found a solution: https://groups.google.com/forum/#!searchin/travis-ci/android$20/travis-ci/3pzCgvxWZAo/LrSckrasFnkJ

PR #365, opened 2015-03-12
[WIP] Api refactor and bean abstraction for GHResponse

dewos:
----------

Moved to https://github.com/dewos/graphhopper-api 

----------

Hello @karussell, with this pull-request I would like to share a bean abstraction for GHResponse. Sure, you can infer almost everything from GHResponse directly, but deal with various Helpers, Json, Translations, and (initially) obscure business logic is not always ideal.

Right now it provides an almost (read the comments please) 1:1 map of the web REST response, and if merged will yield a standard and more user-friendly, serializable, optional API for free. Please take it into consideration.

Cheers.


karussell:
Thanks! I'm still not sure about the exact use case, also these convenient wrapper would introduce more RAM usage and CPU overhead and is not really optimizable.

Also who would use this and maintain :)? And isn't wrapResponse a code duplication?

And what do you mean here?

> Note: The calcRouteBBox fallback is actually ignored ('cause the wrapping is GraphHopperAPI unaware).

dewos:
PreNote: calcRouteBBox fallback in GraphHopperServlet is hopper.getGraph().getBounds() based.

Hello there. User-cases are specific, but I think it's still worth: 

1) It Is serializable (sometimes this really matters, for example for apps with an internal REST point which includes custom info AND the GH-routing in the same response).

2) It highlights a "standard" response structure (REST and Server), still keeping the low-level response avaiable.

3) It is more API-developer-friendly. PointsList and InstructionLists are NOT lists at all (misleading advertising here), and deal with them is really a mess right now (PointList is not even Iterable...). For get the Instruction.turnDirection you need a new TranslationMap()).doImport() in each class. Helpers everywhere... And so on... :)

For performance, sincerely I think that the extra RAM\CPU usage is negligible, moreover the wrapping would be **totally** optional. For mantaining it, it is really only a new little method with few lines inside, as all the beans are only getter\setter logic-free.

dewos:
PreNote: calcRouteBBox fallback in GraphHopperServlet is hopper.getGraph().getBounds() based.

Hello there. User-cases are specific, but I think it's still worth it: 

1) It Is serializable (sometimes it really matters, for example for apps with an internal REST point which includes custom info AND the GH-routing in the same response).

2) It highlights a "standard" response structure (REST and Server), still keeping the low-level response avaiable.

3) It is more API-developer-friendly. PointsList and InstructionList are NOT lists at all (misleading advertising here), and deal with them is really a mess right now (PointList is not even Iterable... I understand why, but still....). Obscure info retrival: example: Instruction.getInterval() missing from Java API. For Instruction.turnDirection you need a new TranslationMap()).doImport(). Oh, and helpers everywhere... and so on... :)

Simply, I like this more:
```java
        restResponse.getPaths().getInstructions().get(0).getInterval();
```
For performance, sincerely I think that the extra RAM\CPU usage is negligible, moreover the wrapping would be **totally** optional. For mantaining it, it is really only a new little method with few lines inside, as all the beans are only getter\setter logic-free.

karussell:
know what you mean and optionally is fine for me. But is the wrapper call (code duplication) really necessary?

> deal with them is really a mess right now

well, it is the 'old style' not sure if this is a mess :)

> new TranslationMap()).doImport(). 

Sure, also I don't think we need this anymore.

> PointList is not even Iterable

From the developer perspective Iterable seems to be a nice interface, but it is an ugly construct. Surely we have to improve in this area. Do you think an iterator based approach would make more sense?

```java
iter = pointList.iterator()
while(iter.hasNext()) {
  // graphhopperish style:
  iter.getLat; iter.getLon
  // or more Java:
  GHPoint point = iter.next();
}
```

> Oh, and helpers everywhere... and so on... :)

please be specific with this, I would like to improve as much as possible

dewos:
P.s. I updated my last post with some other examples.

> know what you mean and optionally is fine for me. But is the wrapper call (code duplication) really necessary?

For allow serialization with the same structure as REST response, yes, I think. The duplicated-code is the bare minimum in any case.

> well, it is the 'old style' not sure if this is a mess :)
Well.... actually.... Iterator and for(x : x) IS the "old style". We now have java8 streams. Hello 2015? :))

> From the developer perspective Iterable seems to be a nice interface, but it is an ugly construct. 

Developer Time is very important too. For the construct... All for taste.... What is clean for you is ugly for others. Anyway, Iterable is standard the-facto by now. 

> Do you think an iterator based approach would make more sense?

PointList and InstructionList shoud work with similar features, otherwise is a "breach of faith-contract" with the end-user. And all the most common infos (like interval) shoud be "1-getter-away" accessible. My two cents.

devemux86:
@karussell to provide GraphHopper for Android usage too, we should be careful with requested min Java version, as the max compiler compliance for Android is Java 7.

karussell:
> We now have java8 streams. Hello 2015?

I really like everything new. The problem really is that if I would do it everywhere in GraphHopper core you would not want and use it :). E.g. streams have a significant slowdown

> And all the most common infos (like interval) shoud be "1-getter-away" accessible. 

Actually the best would be just delegators like 
```java
instruction.getPoints() {
   return points.sublist(fromIndex, toIndex)
}

(please forgive the none java code ;))

> My two cents.

Important two cents. I really take this seriously, now I try to find how to realize this in the best way.

> example: Instruction.getInterval() missing from Java API

This is by desing. The 'interval' information you get just by instruction.getPoints and the interval array in json is very important to do it that way (here again a compomise). Otherwise we would have a point list for every instruction and a point list for the whole path, which would nearly duplicate the size of the json OR we would have to avoid the point for the whole path making it also inconvenient if one needs it (you would have to concatenate all the point lists from the instructions)

> from Emux: @karussell to provide GraphHopper for Android usage too, we should be careful with 
> requested min Java version, as the max compiler compliance for Android is Java 7.

I think @dewos comment was just a generic one AND here is the web module where I could think about Java 8. I do not see a reason why GraphHopper core should go Java 8 (yet).

**Comment by dewos** [comment]
Of course my Jdk8-Stream comment was only an innocent provocation. 

Actually, for the far far future, a Graphhopper "1.0 API" (means: re-build from scratch) would be the best option here, as it seems to me that the API evolved in too different directions (a pretty normal and physiological case for new project). 

For the near future, the beans would be a nice tempfix - addition (for serialization too). 
For Instruction.getInterval() in JAVA api can't you calculate and set it on addition time in InstructionList.add()? Or a InstructionList.getInterval(Instruction ins) at least.

**Comment by dewos** [comment]
P.s.
By the way, to be safe: the REST "interval" are only the pointer to the points (example: [2, 4]), not the complete point list.

**Comment by karussell** [comment]
> as it seems to me that the API evolved in too different directions

Can you elaborate? There are now three different layers of API:

 * the low level API (I do not see how we can change this without being slower)
 * the higher level embedded API (improvable, e.g. the separation of import and routing could be better) and
 * the external 'REST' API which we are now discussing

> the REST "interval" are only the pointer to the points (example: [2, 4]), not the complete point list

The problem is also that I would like to keep the embedded API as similar as possible to the external 'REST' API, and so something like this is unconvenient

> means: re-build from scratch

What do you find ugly? How would a 100% nice solution look like in your opinion? I don't think a re-build from scratch will solve anything. And why would be beans only a 'tempfix'?

> For Instruction.getInterval() in JAVA api can't you calculate and set it on addition time before 
> InstructionList.add()? Or an InstructionList.getInterval(Instruction ins) at least.

What do you mean here? There is no Java API getInterval at the moment just getPoints

**Comment by dewos** [comment]
Actually I'm discussing about making the intermedie "2.5" JAVA API similar to the rest response (which is less powerfull, but much more user-friendly and, I guess, good for the very most of GH user-base).

> What do you find ugly? How would a 100% nice solution look like in your opinion? 

Oh well, this is really a complex and timing-hungry question :)

First removing dead API (getLan + getLon etc...). 

Then for example using the beans (or a different version of them) on the "Embedded API" level too (so no problem for duplicate code etc), so GHPoint and GHInstruction will contains only the basic raw information, moving most of the logic in other classes (better separation of concerns, and impl switching possibile). Something like:

```java
List<Instruction> instructions = response.getInstructions();

ExportManager exportManager = new ExportManager(Translation tr);
List<Map<String, Object>> result = exportManager.createJson(instructions);
```

With this you can make "yours" personal version of ExportManager (for example).

Or (if you really like the "list" container class) an InstrucitonList\PointList implements List (with ad inner ArrayList as now) at least.

But they're just some ideas, making a good, clean and EASY Api is hard.

dewos:
> What do you mean here? There is no Java API getInterval at the moment just getPoints

Exactly, but the response shoud be equals (when possible).

> ....

Actually I'm discussing about making an "intermedie" JAVA API similar to the rest response (which is less powerfull, but much more user-friendly and, I guess, good for the very most of GH user-base).

> What do you find ugly? How would a 100% nice solution look like in your opinion? 

Oh well, this is really a complex and timing-hungry question :)

Let's try this (for a brand new API): 

- Easy things should be easy, and hard things should be possible.

- Clean beans "Response" "Point" and "Instruction" with only getter and setter without dead API methods (getLan + getLon, getEle etc...) or strange things like getExtraInfoJSON.

- Make the beans serializable, so bare minimal structure for providing "GH features" and nothing else. A good idea: use it  for the "Rest API" AND for the "Embedded API" level too. Yes, same class (with maybe minimal "transient" properties on rest response for optimization), you can directly serialize the response to Json so no problem for duplication of code. Hint: https://github.com/FasterXML/jackson-databind/

- Drop "paths" (for rest) and use plain List[Error], List[Instuction], List[Point] in the response. No need to complicate things. Out there are so many libs to deal with java "List".

- Moving ALL (or the very most possibile) of the "real" logic in other specialized optional classes like ExportManager, PathManager etc.. = Better separation of concerns, and implementation switching possibile.  "JSON\GPX etc..." are output-implementation details for an Instruction, not properties. 

Something like:

```java
// Very good for **most** users
List<Instruction> instructions = response.getInstructions();
List<Point> points = response.getPoints();
String polyline = response.getPolyline();// no static helper, do you see?
etc...

// Optional for power users
ExportManager exportManager = new ExportManager();
List<Map<String, Object>> result = exportManager.createJson(instructions);// just an example 'cause createJson is useless with this architecture as the beans are serializable

PathManager pathManager = new PathManager();
List<Point> splittedPoints = pathManager.split(points, 2, 4);// PointList.copy(int, int)
```
GHHopperAPI.route() will populate the beans (translation, generate polyline and so on), but the user will still be able to "fully play" with them (yes, sometimes it is needed and very welcome).

But they're just some ideas. Making a good, clean, and EASY Api is HARD (believe me, it's my daily work :)

In any case, for something less radical, this pull-request will instantly "cover" most of the problems for free (just a kindly reminder).

karussell:
> String polyline = response.getPolyline();// no static helper, do you see?

In general I understand what you mean and like the constructive critique you give although in this specific case I disagree with you :) because a string cannot be directly used and would have to be converted to a list of points using a static helper. Also please keep in mind that many of the things weren't decided in a rush, probably in too many cases I decided towards performance where I should have decided for usability. And eg. having many thousand points in a list makes a difference to the activity of the GC if you have several hundred requests per second (not uncommon!).

So, regarding List[XY]: that is a good thing, we should make it possible. Still List of Point won't happen as there are performance critical applications where this memory waste (and involved GC) would be ugly. But I'm sure that we find a nice API for this too.

> In any case, for something less radical, this pull-request will instantly 
> "cover" most of the problems for free (just a kindly reminder).

I think a step before is to make everything a bit user friendly like no `new TranslationMap()).doImport()`, no static helpers, the PointList being least iterable or iterator BUT without the serialization stuff. Then in a next step solving the (automated) serialization and de-serialization in one go and avoiding manual JSON writes and reads. 

What do you think of this strategy?

dewos:
> In general I understand what you mean and like the constructive critique you give although in this specific case I disagree with you :) because a string cannot be directly used and would have to be converted to a list of points using a static helper.

Wait, the bean is only a "plain" container, ALL the populating logic should be inside hopper.route(). 

```java
        GHRouteResponse res = gh.route(new GHRequest(49.6724, 11.3494, 49.6550, 11.4180));

        // .....in the meantime, in another dimension (INSIDE gh.route())
       ......
        res.setPolyline(WebHelper.encodePolyline(this.points, true):
        .....
        return res;
        // end other dimension

       String polyline = res .getPolyline();// String
```

> Still List of Point won't happen as there are performance critical applications where this memory waste (and involved GC) would be ugly. But I'm sure that we find a nice API for this too.

Ok, here I must play the special card (by Michael A. Jackson):

- The First Rule of Program Optimization: **Don't do it**.
- The Second Rule of Program Optimization (for experts only!): **Don't do it.... yet**.

More seriously, I understand your concerns for the routing, but for the response only I don't know..... By the way if it's really that important there must be a way somewhere. Maybe a LazyList with a cursor?

> I think a step before is to make everything a bit user friendly like no new TranslationMap()).doImport(), no static helpers, the PointList being least iterable or iterator BUT without the serialization stuff.

I suggest this: if you have to do something, do it for good.

> What do you think of this strategy?

I can make a new module (like graphhopper-api) in my fork and put some ideas in it (granting you the write permission). What you think?

karussell:
> I can make a new module (like graphhopper-api) in my fork and 
> put some ideas in it (granting you the write permission). What do you think?

Yeah, for the reading part we could evolve more independent of GraphHopper core. What do you think of our [external java client repo](https://github.com/graphhopper/directions-api-java-client/)?

karussell:
Also I'm not sure if its good to rely on GraphHopper core for the reading Java client.

dewos:
Ok, I'll think about it. Give me two o three days...

dewos:
Here we go @karussell , I build up the rough architecture. It is nice, I think. Some ideas include the high-level of the core, I warn you. The beans part is almost stable, all the remaning is a WIP. Please read all the comments and RUN the example. 

P.s.
Write permission granted

Have fun, see you.

devemux86:
By the way, why Jackson and not e.g. Gson ?

dewos:
@devemux86 

1) it is fast (gson too)
2) it is the suggested\semi-official serializer for Spring MVC, and http://vitalflux.com/java-top-10-java-based-web-development-frameworks-2014-2015 (so annotating it with @JsonSomething will make alot o people happy)

dewos:
Roadmap added to the first post.

karussell:
Thanks! Will look into it but probably not before Wednesday

api/src/main/java/com/graphhopper/client/AbstractHopperClient.java
```diff
@@ -0,0 +1,122 @@
...
+import com.graphhopper.util.shapes.BBox;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+// Please do not throw an error if hasError (with check everywhere)
+// Throws errors on errors is an anti-pattern
```
karussell: The intend is to enforce the hasError check before calling any of the methods, so no anti pattern just making sure people really look into the errors. And what to return for the distance or points if there is an error?

api/src/main/java/com/graphhopper/client/AbstractHopperClient.java
```diff
@@ -0,0 +1,122 @@
...
+         * - com.graphhopper.routing.Path
+         */
+        Downloader downloader = new Downloader("GraphHopper Test") {
+            @Override
+            public InputStream fetch( String url ) throws IOException {
+                return getClass().getClassLoader().getResourceAsStream("com.graphhopper/test.json");
+            }
+        };
+
+        GraphHopperWeb instance = new GraphHopperWeb().setPointsEncoded(false);
```
karussell: why encoded false?

api/src/main/java/com/graphhopper/client/AbstractHopperClient.java
```diff
@@ -0,0 +1,122 @@
...
+                routeInstructionList.add(instruction);
+            }
+        }
+
+        routeResponse.setInstructions(routeInstructionList);
+
+        return routeResponse;
+    }
+
+    private List<double[]> getCoordinates(PointList pointList) {
```
karussell: Why not use pointList.getLatitude(index) instead of toGeoJson? Also please avoid being ironic, not really helpful to understand what you mean.

api/src/main/java/com/graphhopper/internal/LazyPointList.java
```diff
@@ -0,0 +1,152 @@
+package com.graphhopper.internal;
+
+
+import com.graphhopper.bean.RoutePoint;
+
+import java.util.*;
+
+public class LazyPointList implements List<RoutePoint> {
```
karussell: Can we avoid Lazy naming here? I thought about lazy loading, but you meant lazy iteration pattern?

api/src/main/java/com/graphhopper/client/AbstractHopperClient.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+// Please do not throw an error if hasError (with check everywhere)
+// Throws errors on errors is an anti-pattern
+public abstract class AbstractHopperClient implements HopperClient {
```
karussell: shouldn't this class be somewhere in the unit tests as you are using the static json?

api/src/main/java/com/graphhopper/client/AbstractHopperClient.java
dewos: If you want to enforce the hasError check then make clear examples\documentation with it. No need to waste CPU clock at runtime in my opinion.

api/src/main/java/com/graphhopper/client/AbstractHopperClient.java
dewos: AbstractHopperClient right now is just a mock, read the roadmap.  TODO:  Working AbstractHopperClient.route() populating beans without instancing XInstruction InstructionLists or PointLists at all (aka PathMerger and Path logic, but with beans)

api/src/main/java/com/graphhopper/client/AbstractHopperClient.java
dewos: Just a demo: The idea is to drop toGeoJson at all and populate the beans directly.

api/src/main/java/com/graphhopper/internal/LazyPointList.java
dewos: Yes, http://it.wikipedia.org/wiki/Lazy_initialization. Is a performance optimization, hidden to the end-user (internal). "Lazy" is the global and universal name for it, in all languages ever.

api/src/main/java/com/graphhopper/client/AbstractHopperClient.java
dewos: Again: AbstractHopperClient right now is just a mock, read the roadmap.

dewos:
For the sake of clarity: beans are almost stable. All the remaning is a "proof of concept"\demo\mock. 

The idea is to drop Instruction, InstructionList and PointList (and all the toGeoJson() etc...), and, inside AbstractHopperClient.route() populating the beans without instancing anything else (aka PathMerger and Path logic, but with beans). 

No, I'm not ironic here. Is for real.

dewos:
Cleaned up, Peter is this better?

karussell:
As this is a huge undertaking what you plan this would be too much refactoring. Also I only see the need for the (de)serialization for the remote API, not sure if you plan this for desktop too like here: https://github.com/dewos/graphhopper/commit/c0aacc46d30a631be7743231c6be8e6bb21c0621#diff-327e4a118bc1174bd97ad3dcd440ef22R35

dewos:
Hi Peter, 
For the mapper: All the codes inside Example.java is an "user" example usage, not the lib (but of course an ipotetical WebHopperClient could use the mapper for serialize the response).

Yes the changes are many, but not SO many, and the API can become very very nice and clean. Do you like it? Have some other ideas?

P.s.
The diff your linked is very old, please look at the new commits.

dewos:
Hi Peter, 
For the mapper: All the codes inside Example.java are a "user" example usage, not the lib (but of course an ipotetical WebHopperClient could use the mapper for serialize the response).

Yes the changes are many, but not SO many, and the API can become very very nice and clean. Do you like it? Have some other ideas?

P.s.
The diff your linked is very old, please look at the new commits.

dewos:
Moved to https://github.com/dewos/graphhopper-api (still very "work in progress" and fluid).

karussell:
This is a nice initiative but still I fear just too big and too many things mixed in. If you just want to have better (de)serialization we should concentrate on this and not on refactoring. Still we can refactor afterwards into a cleaner API. Additionally currently it is not planned to change the JSON output.

Furthermore: I do not understand LazyPointList, if the purpose is using less memory, then I do not see how it does so.

dewos:
Yes Peter, I fear the same. A middle-way solution could be to make a parallel project thinking about the best API possible, wrapping for now the old API inside the new one to avoid the duplication of code\break things (so, Graphhopper inside HopperEngine, and GHResponse into the HopperResponse and so on...).

When it is ready and both of us are happy with it you can think about deprecating the old API (in core), and, after one or two version (like graphhopper 0.7+) switch from the wrapping logic to the real one (for performance of course). What do you think?

dewos:
P.s.
List<RoutePoint> is a "sweet" version of List<coordinates[]>, so you can populate only List<double[]> coordinates from the List<Path>, no need to populate both of them at route() time. More clear?

karussell:
I've experimented also with jackson and (de)serialization in this repo https://github.com/karussell/graphhopper-traffic-data-integration/ for a completely different purpose, but I found out that we could introduce some custom (de)serializer e.g. could be also the polyline stuff (see [here](https://github.com/karussell/graphhopper-traffic-data-integration/blob/master/src/main/java/com/graphhopper/traffic/demo/CustomGuiceModule.java#L53)). So not always we would need to change our domain classes but just the parsing/writing unit

devemux86:
Just a thought..
If possible, it would be favorable to avoid many external libraries dependencies (particularly for multi-platform components).

karussell:
Sure, the core won't get this stuff, when something like this will happen we'll have it in a separate module (and core will be splitted further into import and core-core ;))

dewos:
Hello there, has been a long week :)
Ok lets try to deal with a problem at time-

Serialized version of GHResponse
---------------------------------------------
@karussell what's wrong with the first pull request? ( like #abc7204 ). I do not believe we can do any better  without a real refactor and\or without duplicating the x.toGeoJson() logic.

About the external libraries dependencies, you should consider to annotate the beans with @JsonProperty for compatibility with Spring and many other frameworks. Don't worry it is a very light dep (just a module with ONLY few annotation and nothing else). No need to take in anything else.

<dependency>
  <groupId>com.fasterxml.jackson.core</groupId>
  <artifactId>jackson-annotations</artifactId>
  <version>${jackson-annotations-version}</version>
</dependency>

PR #368, opened 2015-03-13
Make tan calculation in AngleCalc more precise

zhiqiangZHAO:
Previous one is not correct for high lat area.
This one is correct.

Actually it's a Azimuth.
There is a very complex version in geotools, which we don't really need it.


karussell:
Thanks! With high lat you mean abs(lat) > 60 or something? (just to test it)
And has it performance implications if route is calculated with instructions?

zhiqiangZHAO:
@karussell yes, like Greenland. The angle it calculated are so wrong.
Actually, the original code is good for plane position (x, y).
But not correct for a (lat, lon).
the distance of 1 degree in lat doesn't always equal to distance of 1 degree in lon.

karussell:
Thanks again, fixed now!

PR #377, opened 2015-03-21
GPX: introduce indentation in xml elements

devemux86:
Introduce indentation in GPX xml elements and move closing tags in new line.

That way we may read more easily the produced gpx and process it further.

### DIFF
--- a/core/src/main/java/com/graphhopper/util/InstructionList.java
+++ b/core/src/main/java/com/graphhopper/util/InstructionList.java
@@ -186,22 +186,22 @@ public class InstructionList implements Iterable<Instruction>
 
         formatter.setTimeZone(tz);
         String header = "<?xml version='1.0' encoding='UTF-8' standalone='no' ?>"
-                + "<gpx xmlns='http://www.topografix.com/GPX/1/1' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'"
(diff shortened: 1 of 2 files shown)


devemux86:
I fixed also the unit test.

karussell:
Ok. Still opening an xml in a browser always formats you the xml, no?

devemux86:
Well for just viewing it, but we may need to process it with an editor.
It would be convenient to be in a user readable format.

For the record OSRM gpx files are just one line, even harder to comprehend.

karussell:
Sure, I understand. Are there any implications for other tools when including these new indentations (besides more kbytes)?

karussell:
Hmmh, it looks like now several more chars per trackpoint are added, which indeed means several KB more data. I don't like this as it is just for the purpose of debugging we should introduce a debug=true parameter or avoid this completely as other tools (like the browser or xmllint) are there to do pretty print of xml

devemux86:
It's your call but then we shouldn't have the current middle situation right now, where half the gpx file has some kind of indentation and the other is mixed up.

I suggest either to have it properly or have it compact (like OSRM).

devemux86:
Peter I'm closing this PR as I need to reorganize my repository for other PR.

If there is still interest in the future, I can resend it.

karussell:
Ok + I'm thinking now how we do this

BTW: If you do the PR suggestions on a branch you won't have this problem.

devemux86:
>BTW: If you do the PR suggestions on a branch you won't have this problem.

Thanks, that's exactly what I neglected to do and want to reorganize.

PR #379, opened 2015-03-23
GPX: fix timestamp representation

devemux86:
Peter I think that in GPX the timestamps should be in UTC with ISO 8601 representation and end with the 'Z' suffix.
http://www.topografix.com/gpx_manual.asp#time

And to conform with spec, probably it's not necessary to provide the timezone as parameter to the createGPX method.

### DIFF
--- a/core/src/main/java/com/graphhopper/util/InstructionList.java
+++ b/core/src/main/java/com/graphhopper/util/InstructionList.java
@@ -18,7 +18,6 @@
 package com.graphhopper.util;
 
 import java.text.SimpleDateFormat;
-
 import java.util.*;
 
 /**
@@ -168,7 +167,7 @@ public class InstructionList implements Iterable<Instruction>
      */
     public String createGPX()
     {
-        return createGPX("GraphHopper", 0);
+        return createGPX("GraphHopper", new Date().getTime());
     }
 
     public String createGPX( String trackName, long startTimeMillis )
--- a/core/src/main/java/com/graphhopper/util/InstructionList.java
+++ b/core/src/main/java/com/graphhopper/util/InstructionList.java
@@ -179,7 +179,7 @@ public class InstructionList implements Iterable<Instruction>
 
     public String createGPX( String trackName, long startTimeMillis, String timeZoneId, boolean includeElevation )
     {
-        SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
+        SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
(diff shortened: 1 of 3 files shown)


karussell:
Thanks, this would be really a nice cleanup :) ! Did you try some of the common GPX programs if they are okay with this? (basecamp, some online tools, josm, osm.org?)

devemux86:
Most readers (online, apps) are versatile enough to parse more than one timestamp patterns.
And I have found gpx files that don't follow the standard.

But going the other way, wherever I checked for official produced gpx files, I saw them following the spec.
(please inform us if someone has additional info)

karussell:
When we use 'Z' shouldn't we force +0? I.e.:

```java
SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
formatter.setTimeZone(TimeZone.getTimeZone("UTC"));
System.out.println(formatter.format(new Date()));
```
Otherwise I'll get incorrect time when using default time zone here in Germany.

devemux86:
What do you mean, don't you get with the above code the UTC time ?

karussell:
It will correctly use the local or specified time but ignores the offset as we always write 'Z'. I think you can easily try on your own. It should give you different times (except you are in london :)) but it will give you identical times

devemux86:
Right not it gave me ```2015-03-23T21:15:00Z```

devemux86:
That's the correct UTC time, what do you get?
(or I'm missing something here)

karussell:
Strange, that you get it correctly. I get `2015-03-23T22:18:42Z` 
Also it makes sense to force UTC and avoid the default or a specific time zone when we always use +0

devemux86:
The UTC timezone is needed or else as you said we'll get our current time.

I don't remember Java to have changes in that part (I'm in Linux with 7).

karussell:
Okay, but then this old code is wrong with your change:

```java
TimeZone tz = TimeZone.getDefault();
if (!Helper.isEmpty(timeZoneId))
   tz = TimeZone.getTimeZone(timeZoneId);
```
and we should always use UTC and remove the timeZone parameter, right?

devemux86:
Yes you are corrent, to be with the spec we shouldn't provide the user any timezone option, but use internally only UTC.

devemux86:
Peter I pushed a new commit removing the timezone parameter and making the UTC default.

Also I think that in case of not provided ```startTimeMillis```, the current date should be used and not the 0.

karussell:
Thanks & yes, you are right regarding the startTimeMillis

devemux86:
I pushed another commit regarding ```startTimeMillis```.

karussell:
Thanks!

Issue #362, opened 2015-03-11
Error in Dijkstra-One-To-Many cache


thingimon:
As mentioned in issue #239, there was an error in the cache of the Dijkstra-One-To-Many. However, the new version wrongly returns NOT_FOUND when to == currNode (i.e. searching the same path twice in a row). The inequality
```weights[to] < weights[currNode]```
should maybe not be strict?

karussell:
Yes, indeed!

karussell:
Thanks for reporting!

PR #380, opened 2015-03-23
DijkstraOneToMany - isWeightLimitReached

dos65:
Hi, i found this algo may return NOT_FOUND in case, where desired node is presented in heap and we reached earlier other node with same weight as desired and its weight equals to limit. Also this fix affects other 2 preparation tests.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/DijkstraOneToMany.java
+++ b/core/src/main/java/com/graphhopper/routing/DijkstraOneToMany.java
@@ -205,7 +205,7 @@ public class DijkstraOneToMany extends AbstractRoutingAlgorithm
 
     protected boolean isWeightLimitReached()
     {
-        return weights[currNode] >= weightLimit;
+        return weights[currNode] > weightLimit;
     }
 
     public void close()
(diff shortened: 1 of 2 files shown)


karussell:
This should be already fixed due to #362 (?)

karussell:
Ah, wait you removed the equal sign ...

dos65:
Yep, please look at new test.

dos65:
hmm. Maybe if current weight reached limit and equals its, we need check weights array and return "to" if it is contained there?

karussell:
What do you mean? The current solution seems good to me. Additionally we probably need to change the method name and do this (`>`) for all other algorithms.

dos65:
It get slightly more unnecessary iterations, but it is not critical. Ok I'll fix methods and check other algo. Can I rewrite edge counter in broken CH tests?

karussell:
> It get slightly more unnecessary iterations

Yes, but they are necessary :)

I currently investigate and will fix and let you review (as there is some inconsistency for CH and the other bidirectional algos when it comes to the finish and isWeightLimitReached methods)

karussell:
Okay, would you mind to review the changes? Then I'll merge

dos65:
Reviewed :) 
Thanks!

karussell:
Thank you :) !

PR #385, opened 2015-03-26
Made maximale possible speed value more explicit

jansoe:


karussell:
Thanks! Would you mind to introduce this variable for the other encoders as well :) ?

karussell:
Nice, thanks!

PR #386, opened 2015-03-30
Docs core weighting example

tomhoag:
Fixed a couple parameter name mismatches in the example code for the weighting docs.


tomhoag:
Moving to 0.4

PR #387, opened 2015-03-30
Changed var names to match those passed into methods

tomhoag:
Fixed a couple parameter name mismatches in the example code for the weighting docs.


karussell:
ups, thanks a lot!

Issue #270, opened 2014-10-22
Turn Costs for CH


karussell:
Turn costs work only without CH

Will be implemented via #1247

PR #394, opened 2015-04-13
DijkstraOneToMany traversal support for #270

dos65:
It is initial implementation of one-to-many with edge-based support.
But i am getting stuck with shortcuts creation. It seems like we need to contract and set levels for traversalIds, right?


karussell:
Thanks, will look into it!

> But i am getting stuck with shortcuts creation. It seems like we need to contract and set levels 
> for traversalIds, right?

You mean for CH? Yes, either nodeIds or edgeIds ... ah yes traversalId (2*edgeId) is more correct

karussell:
Your work inspired #912 ... so I think we can close here

PR #397, opened 2015-04-21
Added nearest API to snap point to road

Svantulden:
Hi, 

Added a nearest API call to Graphhopper that returns the nearest point on the prepared graph for your input point. We use this to snap a user's pinpoint to the road. It also gives the distance from the input point to the snapped point. 

You can use it by calling /nearest?point=lat,lon (jsonp is also supported and uses the same code as /route, just add &type=jsonp&callback=x)

All feedback is welcome!

### DIFF
--- a/web/src/main/java/com/graphhopper/http/GHServletModule.java
+++ b/web/src/main/java/com/graphhopper/http/GHServletModule.java
@@ -64,5 +64,8 @@ public class GHServletModule extends ServletModule
 
         serve("/route*").with(GraphHopperServlet.class);
         bind(GraphHopperServlet.class).in(Singleton.class);
+        
+        serve("/nearest*").with(NearestServlet.class);
+        bind(NearestServlet.class).in(Singleton.class);
     }
 }
--- a/web/src/main/java/com/graphhopper/http/NearestServlet.java
+++ b/web/src/main/java/com/graphhopper/http/NearestServlet.java
@@ -61,6 +61,7 @@ public class NearestServlet extends GHBaseServlet
     void writeNearest( HttpServletRequest req, HttpServletResponse res ) throws Exception
     {
         String pointStr = getParam(req, "point", null);
+        boolean enableElevation = getBooleanParam(req, "elevation", false);
         
         JSONObject result = new JSONObject();
         if (pointStr != null && !pointStr.equalsIgnoreCase("")) {
@@ -82,6 +83,10 @@ public class NearestServlet extends GHBaseServlet
                 JSONArray coord = new JSONArray();
                 coord.put(snappedPoint.lon);
                 coord.put(snappedPoint.lat);
+                
+                if (hopper.hasElevation() && enableElevation) {
+                    coord.put(snappedPoint.ele);
+                }
 
                 result.put("coordinates", coord);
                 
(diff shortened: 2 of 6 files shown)


karussell:
Really nice, thanks!

Would you mind to add one simple test similar to one of in GraphHopperServletIT?

Svantulden:
I've added a simple test for the nearest call. Is this what you meant?

karussell:
Yes & nice, that was fast!

> double lat = point.getDouble(0);
> double lon = point.getDouble(1);

Hmmh, that is a bit ugly as we always use lon,lat if a point is just returned as array. To make it according to the GeoJSON specification. Would it be possible to change this?

And one further question regarding: 

> assertTrue("returned point is not 2D: " + point, point.length() == 2);

is fetching elevation data with this method also possible? I.e. will it add a third dimension if enabled? Would be an interesting use case and probably easy to implement (?)

karussell:
> Hmmh, that is a bit ugly as we always use lon,lat if a point is just returned as array. 
> To make it according to the GeoJSON specification. Would it be possible to change this?

Or we can make it explicit like 'lat:..., lon: ..' ?

Svantulden:
> Hmmh, that is a bit ugly as we always use lon,lat if a point is just returned as array. To make it according to the GeoJSON specification. Would it be possible to change this?
Good idea, I have changed it now to use the GeoJSON specifications for a single Point

> is fetching elevation data with this method also possible? I.e. will it add a third dimension if enabled? Would be an interesting use case and probably easy to implement (?)
Hmm I don't quite know, as my data set doesn't contain elevation data yet. Theoretically it would be possible? I'll see what I can do. Do you know if it is GeoJSON valid if we add a third value to the coordinates array for elevation? Or would this be another property in and of itself?

karussell:
Thanks, again!

> Do you know if it is GeoJSON valid if we add a third value to the coordinates array for elevation? 

I think so. 

> Or would this be another property in and of itself? 

For the route we use just a third component, yes. 

BTW: Enabling elevation is very simple, I've [blogged about it yesterday](https://karussell.wordpress.com/2015/04/21/elevation-data-and-openstreetmap/). I think it is important for bike&foot routing.

devemux86:
>Do you know if it is GeoJSON valid if we add a third value to the coordinates array for elevation? 

http://geojson.org/geojson-spec.html#positions
"The order of elements must follow x, y, z order (easting, northing, altitude for coordinates in a projected coordinate reference system, or longitude, latitude, altitude for coordinates in a geographic coordinate reference system)."

Svantulden:
Thanks guys! I've added elevation to the output when available. I've also added a test for nearest with elevation. I mirrored the tests for graphhopper web with elevation, if that is okay. 

Is `hopper.hasElevation()` reliable enough to know if elevation is available in the dataset? You can also use `&elevation=true` or` false` in the url like you can with `/route`.

karussell:
Thanks!

> Is hopper.hasElevation() reliable enough to know if elevation is available in the dataset?

Yes, should be sufficient.

> When we tried it we ran into some memory problems on our machine and 
> because of our time schedule we decided to add it in later. 

Ah, okay. But they should be only add very few RAM like 10-20% (?)

karussell:
I've just rebased into one commit and merged it here: https://github.com/graphhopper/graphhopper/commit/d69d1114fecd688e3dc4057fc75e6b716052391b

Svantulden:
Awesome! Thanks for merging :+1:

Issue #143, opened 2014-01-15
Simple JavaScript UI to display and create graphs for unit tests


karussell:
This bites me very often and I create graphs on paper or via inkscape but viewing, creating or editing them is too costly. Therefor it would be nice to have a simple JavaScript map where I can create a sample graph and the app would export very simplified Java code. The trick is here to export into simple Java code, which is then readable from the JavaScript app back to display it too. Another option would be to export into some JSON which is then easily copy&pastable into a unit test (e.g. via Helper.createGraphFromJson("{some: json}")). But I prefer Java code if it is more compact.

Main purposes
- visualize graphs in unit tests
- visualize calculated routes
- create graphs for unit tests to avoid full or partial OSM import

Requirements
- lightweight JS client
- should run locally too, internet only required for fetching tiles. Or copy pasting to http://jsfiddle.net or http://jsbin.com?
- no DB or server necessary
- you can create polylines and connect them to graphs -> some leaflet plugins: http://makinacorpus.github.io/Leaflet.Snap/ and https://github.com/Leaflet/Leaflet.draw/ or https://www.mapbox.com/mapbox.js/example/v1.0.0/leaflet-draw/
- one can further edit the graph even after export etc
- export points as lat,lng list and Java code, Later: geoJson or gpx or to something like [this](http://arthur-e.github.io/Wicket/)
- print existing graph from unit test somehow
- import previously exported data or even 'Java' code
- embed this data directly in an URL to avoid copy or paste? or just use json and copy&paste or another good idea how to link from unit tests to this UI (could be hosted e.g. on graphhopper.com/maps/creator)
- Later: upload button for gpx file via leaflet plugin -> https://github.com/makinacorpus/Leaflet.FileLayer

ammagamma:
I made a little in-browser app to draw graphs and display very simple corresponding java code here:
https://github.com/ammagamma/graphhopper-creator .
[...]

PR #401, opened 2015-05-02
Flag encoder versions

andreaswolf:
This adds version numbers to all encoder classes. The numbers are stored with the graph and checked on reloading.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
@@ -52,6 +52,8 @@ public abstract class AbstractFlagEncoder implements FlagEncoder, TurnCostEncode
     protected long acceptBit;
     protected long ferryBit;
 
+    protected PMap properties;
+
     // This value determines the maximal possible speed of any road regardless the maxspeed value
     // lower values allow more compact representation of the routing graph
     protected int maxPossibleSpeed;
@@ -78,6 +80,15 @@ public abstract class AbstractFlagEncoder implements FlagEncoder, TurnCostEncode
     private boolean blockFords = true;
     protected final int speedBits;
     protected final double speedFactor;
+
+    public AbstractFlagEncoder(PMap properties) {
+        throw new RuntimeException("This method must be overridden in derived classes");
+    }
+
+    public AbstractFlagEncoder(String propertiesStr)
+    {
+        this(new PMap(propertiesStr));
+    }
 
     /**
      * @param speedBits specify the number of bits used for speed
@@ -714,6 +725,7 @@ public abstract class AbstractFlagEncoder implements FlagEncoder, TurnCostEncode
         throw new UnsupportedOperationException("Unknown key " + key + " for double value.");
     }
 
+    @Deprecated
     protected static double parseDouble( String str, String key, double defaultD )
     {
         String val = getStr(str, key);
@@ -722,6 +734,7 @@ public abstract class AbstractFlagEncoder implements FlagEncoder, TurnCostEncode
         return Double.parseDouble(val);
     }
 
+    @Deprecated
     protected static long parseLong( String str, String key, long defaultL )
     {
         String val = getStr(str, key);
@@ -730,6 +743,7 @@ public abstract class AbstractFlagEncoder implements FlagEncoder, TurnCostEncode
         return Long.parseLong(val);
     }
 
+    @Deprecated
     protected static boolean parseBoolean( String str, String key, boolean defaultB )
     {
         String val = getStr(str, key);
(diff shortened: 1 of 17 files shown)


karussell:
Thanks!

Looks good! Two minor things: would you use PMap instead of the new properties class? And would you rebase the changes into one commit without the [BRACKET] notation? (it is just uncommon for graphhopper, and maybe we introduce something for the different modules [web/mobile/...] but step by step :))

andreaswolf:
I adjusted to your comments. I did not squash the patch to make the changes between versions 1 and 2 clear. If you think the feature is worth merging, I’ll squash the commits to one.

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -714,6 +725,7 @@ public double getDouble( long flags, int key )
         throw new UnsupportedOperationException("Unknown key " + key + " for double value.");
     }
 
+    @Deprecated
```
karussell: that is indeed important :)
Would you also add this into the https://github.com/graphhopper/graphhopper/blob/master/core/files/changelog.txt and comment on the alternative?

core/src/main/java/com/graphhopper/storage/StorableProperties.java
```diff
@@ -168,6 +168,9 @@ public boolean checkVersions( boolean silent )
         {
             return false;
         }
+
+        // The check for the encoder version is not done here as this class does not know about the registered encoders
+
```
karussell: maybe point to the place where it is?

core/src/main/java/com/graphhopper/util/PMap.java
```diff
@@ -27,7 +27,7 @@
  */
 public class PMap
 {
-    private final Map<String, String> map;
+    protected final Map<String, String> map;
```
karussell: why do we need this protected?

core/src/test/java/com/graphhopper/routing/util/EncodingManagerTest.java
```diff
@@ -66,6 +71,13 @@ public void testEncoderAcceptNoException()
     }
 
     @Test
+    public void testEncoderWithWrongVersionIsRejected()
+    {
+        thrown.expect(IllegalArgumentException.class);
```
karussell: nice. I should have done this that way in lots of other places too :)

karussell:
Looks good, had just very minor comments.

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
andreaswolf: Will do, sure.

core/src/main/java/com/graphhopper/util/PMap.java
andreaswolf: Not anymore, needed this during migration. Will roll back.

core/src/test/java/com/graphhopper/routing/util/EncodingManagerTest.java
andreaswolf: I can go through some existing tests and adjust them, would be a good way for me to learn more about JUnit…

karussell: yeah, but not a high priority. maybe fix them when you stumble over them step by step :)

BTW: One more important issue for the testing stuff is a simple graph creation UI: https://github.com/graphhopper/graphhopper/issues/143

dos65: Also there is another way to define expected exception in test
`@Test(expected = IllegalArgumentException.class)`

karussell: Thanks, even better :) !

andreaswolf: AFAIK there’s no way to test for the exception message in the annotation, right?

andreaswolf:
First pushed my new three commits, after that a squashed version that contains all changes in one commit (and a minor unrelated change in a different commit).

core/src/test/java/com/graphhopper/routing/util/EncodingManagerTest.java
karussell: Probably but I don't know ... I was just used to catch+assert

karussell:
Nice, thanks!

core/src/test/java/com/graphhopper/routing/util/EncodingManagerTest.java
dos65: Yes, there is no way to test exception message using annotation, but i think it's doesn't matter in tests like this :).

PR #411, opened 2015-05-15
Add missing support for tag bicycle:forward=no detected during test a…

ratrun:
As a result of the OSM note discussion http://www.openstreetmap.org/note/319765 and the resolution with bicycle:forward=no I detected that graphhopper does currently not support this tag.

This change does two things:

1) Add missing support for tag bicycle:forward=no detected during test after OSM issue http://www.openstreetmap.org/note/319765 got fixed. 
2) Utilize the basic oneway logic in bike2 and handle the bike2 specific stuff afterwards such that we don't have to maintain the oneway stuff handling for bikes twice.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/Bike2WeightFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/Bike2WeightFlagEncoder.java
@@ -92,25 +92,14 @@ public class Bike2WeightFlagEncoder extends BikeFlagEncoder
     public long handleSpeed( OSMWay way, double speed, long encoded )
     {
         // handle oneways
-        if ((way.hasTag("oneway", oneways) || way.hasTag("junction", "roundabout"))
-                && !way.hasTag("oneway:bicycle", "no")
-                && !way.hasTag("cycleway", oppositeLanes))
+        encoded = super.handleSpeed(way, speed, encoded);
+        if (isBackward(encoded))
         {
-
-            if (way.hasTag("oneway", "-1"))
-            {
-                encoded |= backwardBit;
-                encoded = setReverseSpeed(encoded, speed);
-            } else
-            {
-                encoded |= forwardBit;
-                encoded = setSpeed(encoded, speed);
-            }
-        } else
-        {
-            encoded |= directionBitMask;
-            encoded = setSpeed(encoded, speed);
             encoded = setReverseSpeed(encoded, speed);
+        }
+        if (isForward(encoded))
+        {   
+            encoded = setSpeed(encoded, speed);
         }
         return encoded;
     }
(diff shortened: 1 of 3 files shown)


karussell:
Thanks! Good catch regarding the code duplication! But I'm not sure regarding the `bicycle:forward=no`. Is this really the right tag to use, there are only 73 such values world wide :) !

BTW: also feel free for a lane>1 bike penalty as discussed on the mailing list :)

ratrun:
Thanks for detecting the infrequent usage of `bicycle:forward=no`. Even tough its infrequent usage I think we should add the support for it, because the generic `vehicle:forward` is also supported. But now I additionally implemented support for `oneway:bicycle=yes`, which is much more common.

I did not extend anything about lanes. On the mailing list I just wanted to inform the user that we do not evaluate any lanes for bicycle routing. As a first guess it might sound as a good idea to avoid ways with multiple lanes.
But it is not that simple:
*   On steep ways it is much better to have a road with two or even more lanes such that the faster traffic has more space for passing by.
*   There would need to be some handling at crossings, which have turn lanes. I'm currenlty not very familiar with the OSM tagging sheme for turn lanes.
*   Many lanes within residential areas indicate danger for traffic jams. Outside of residential areas the maxspeed for cars is much more important.
*   Especially with the racing bike it is a good idea to choose roads with many lanes, except for rushing hours. On a Sunday morning it is really great to leave big cities on roads with many lanes.

Therefore I have improved the avoidance handling depending on the car maxspeed tag instead.
See: https://github.com/graphhopper/graphhopper/compare/master...ratrun:master 
For some unknown git reasons the two additional commits don't show up here under this pull request. Do I need to create a new pull request?

karussell:
Okay, thanks for the explanations. Makes sense.

> For some unknown git reasons the two additional commits don't show up here under 
> this pull request. Do I need to create a new pull request?

This pull request is against your branch `ratrun:bicycleforwardno` so you can add the changes there. BUT we should really separate every issue as a separate PR to avoid holding back changes that are okay because of changes that need work.

ratrun:
Thanks for pointing out the commit to the wrong branch. Could you please check once again? I added the patch with the maxspeed because of your comment about the lanes. Now I removed this change from this pull request. This pull request add now just the support for oneway:bicycle and bicycle:forwad=no.

karussell:
Thanks! 

One minor point: would you squash into one commit :) ? E.g. see [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html), if not, really no problem I can do this

ratrun:
Sorry, git only sometimes does what I intend, I always get errors which I can't interpret.

karussell:
no problem, I know what you mean ;) (I'll do it tomorrow)

devemux86:
@karussell thanks for the git squash hint!

karussell:
Merged here: https://github.com/graphhopper/graphhopper/commit/1df47ec6dae1f07579b168921bc4e42d6ebea040

Issue #412, opened 2015-05-16
Improve bicycle routing on footways


ratrun:
As a result of a discussion on the mailing list the following items should be checked and improved:

*  Maybe we should increase the preference also in case of presence of the tag `bicycle=yes`. Currently this is done only for `bicycle=designated`
* The speed values with a present surface tag needs to be decreased for footways.
* `highway=path` and `highway=footway` should be treated the same way.
* Eventually we also could increase the assumed speed for footway from rather conservative 6km/h to 8km/h, but I would not like to increase this speed further because I fear that this would have noteable negative effects in big cities. 
* We should influnce the assumed speed values based on the width tag: E.g. a 3 meters wide footway we could assume speed 16km/h, and fall back to the default value for a width<1.5 meters.

karussell:
> Maybe we should increase the preference also in case of 
> presence of the tag bicycle=yes. Currently this is done only for bicycle=designated

You mean for footpath? Then maybe, yes.

> The speed values with a present surface tag needs to be decreased for footways.

What do you mean here?

> highway=path and highway=footway should be treated the same way.

okay

> Eventually we also could increase the assumed speed for footway from rather conservative 6km/h to 8km/h, but I would not like to increase this speed further 
> because I fear that this would have noteable negative effects in big cities. 

We should avoid that. 

> E.g. a 3 meters wide footway we could assume speed 16km/h, and fall back to the default value for a width<1.5 meters. 

Hmmh, but often if the path is big there is also a high frequency of pedestrians. If it is a footway, it is a footway :). Only if bicycle=yes or designated is tagged we should think about improving preference for bike or even increasing speed.

ratrun:
Thank you. So I'll prepare a patch covering the following:

Change preference for `bicycle=yes` as we already do for `bicycle=designated`

> The speed values with a present surface tag needs to be decreased for footways.
Currently a `highway=footway` is assumed to have speed 6km/h. In case that a tag `surface=paved` is set for a footway, the speed boosts to 18km/h, which is sub-optimal. The reason is because the code for the suface tag assumes `highway track` at least.

Additionally I'll try to handle `highway=path` and `highway=footway` pretty much the same for the common bike profile with an exception for mountainbike.

I won't add any change for evaluating the width tag.

karussell:
> is set for a footway, the speed boosts to 18km/h, which is sub-optimal. 

Oh, that is not good. Important to fix & ensure via test.

[...]

PR #414, opened 2015-05-17
Improve avoidance of ways with high maxspeed values for cars

ratrun:
This change was triggered by a discussion on the mailing list , see https://lists.openstreetmap.org/pipermail/graphhopper/2015-May/001964.html .

Improve avoidance of ways with high maxspeed values for cars and control the avoidance with one car speed on AVOID_IF_POSSIBLE level and another speed limit for the REACH_DEST avoidance.

The AVOID_IF_POSSIBLE max car speed is now defined as 71 km/h for all bike types except for the racingbike.
The REACH_DEST max car speed is now defined as 81 km/h for all bike types except for the racingbike.
The AVOID_IF_POSSIBLE max car speed is defined as 81 km/h for the racingbike.
The REACH_DEST max car speed is defined as 101 km/h for the racingbike.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -57,6 +57,9 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
     private EncodedValue wayTypeEncoder;
     private EncodedValue preferWayEncoder;
 
+    // Car speed limit which switches the preference from UNCHANGED to AVOID_IF_POSSIBLE
+    private int avoidSpeedLimit;
+
     protected BikeCommonFlagEncoder( int speedBits, double speedFactor, int maxTurnCosts )
     {
         super(speedBits, speedFactor, maxTurnCosts);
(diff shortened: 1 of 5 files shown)


karussell:
Hmmh, I would like to avoid the complexity of having two speed limits for this. Where did you encounter a real world problem with the current status and is the change improving the situation?

ratrun:
This change was triggered by this an answer mail on the mailing list for this example

https://www.openstreetmap.org/directions?engine=graphhopper_bicycle&route=64.13323%2C-21.86841%3B64.13484%2C-21.87089#map=17/64.13394/-21.86962

in this mail: https://lists.openstreetmap.org/pipermail/graphhopper/2015-May/001964.html .

The user titled a `higway secondary maxspeed=50` as a "multi-lane fast traffic" road. During writing the answer I came to the conclusion that the current code with the check for >80km/h is not good because there are many roads with a speed limit of exactly 80.

The extension with the two limits is just logical in order to better serve the racing-bike use case. Most racing-bike people are able to cope with faster traffic. But there is also a limit for them. So I introduced the second limit.

karussell:
At least for this example I think it will be fixed once your change regarding bicycle=yes is deployed. Also traffic light recognition would help and I do not like this speed check that much. E.g. what if a high speed is allowed but the street is not crowded or if only 50 km/h and very full street. As we know these issues of the speed check I would like to avoid relying even more on them and making the code complexer.

ratrun:
You are right. The changes in #412 will probably solve the mentioned example. But I think that this change would improve the situation for other cases, where smaller roads run nearby faster roads. Would you accept the change if I reduce the complexity and use just a single maxspeed value?

karussell:
Yes, please just use one speed like before and maybe set different values for the different bikes (race vs. normal)?

ratrun:
Thanks for the improvement comment. Now it is indeed cleaner. Could you please check once more?

karussell:
Looks good, thanks!

PR #415, opened 2015-05-18
Path: Allow different visitors for instruction calculation

andreaswolf:
This change allows to use the existing methods with a custom visitor to calculate instructions.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/Path.java
+++ b/core/src/main/java/com/graphhopper/routing/Path.java
@@ -358,20 +358,12 @@ public class Path
     /**
      * @return the list of instructions for this path.
      */
-    public InstructionList calcInstructions( final Translation tr )
+    public InstructionList calcInstructions(final Translation tr)
     {
         final InstructionList ways = new InstructionList(edgeIds.size() / 4, tr);
-        if (edgeIds.isEmpty())
-        {
-            if (isFound())
-            {
-                ways.add(new FinishInstruction(nodeAccess, endNode));
-            }
-            return ways;
-        }
-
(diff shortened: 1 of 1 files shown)


karussell:
What usage scenario do you have in mind? And also there seems to be a test failing (or is this an uncovered bug?)

andreaswolf:
I need to replace the instruction calculation with my own code and I want to stick als close as possible to upstream Graphhopper (i.e. not mess around with a fork), so I need some extension points.

As for the tests, I actually did not run them (my bad…). I’ll have a look.

karussell:
What do you need to modify for the instruction calculation?

andreaswolf:
Most likely, I will need to completely override it, as I need all road segments in detail (not only tower-tower instructions) and some more information on the segments and the nodes. I did not look into that too deep, but that’s the general concept.

I also don’t know by now if I can reuse the Path class at all or if I need more data in there – but I’m optimistic that the existing data is enough.

karussell:
What I do not understand is if you need to do your own instruction calculation, why not just avoid instruction calculation at all (via hint: instructions=false) and then fetch all edges (calcEdges->gives you all you need) and then do you own instruction calculation?

andreaswolf:
Ah, ok. I think I got the whole thing wrong – just realized that I didn’t even get the refactoring 100% right :-/

Sorry for the hassle, I’ll close this merge request.

karussell:
It is really no problem, often new interaction leads to a better API!

But at the same time we have to make sure that we really need things we add :)

andreaswolf:
I’ll give my best to improve GraphHopper :-)

PR #417, opened 2015-05-20
Update EncodedValue.java

zhiqiangZHAO:
make sure java treated flags as an unsigned long, in case we use the highest bit.
Actually, there are more in the code.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/EncodedValue.java
+++ b/core/src/main/java/com/graphhopper/routing/util/EncodedValue.java
@@ -79,7 +79,7 @@ public class EncodedValue
         checkValue(value);
         // scale value
         value /= factor;
-        value <<= shift;
+        value <<<= shift;
 
         // clear value bits
         flags &= ~mask;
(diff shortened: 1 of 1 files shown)


karussell:
Thanks!

Do you have some hints what you mean with 'there are more in the code'? And would you mind creating a test case?

zhiqiangZHAO:
@karussell 
I use git grep to find the following class may have the same issue:

AbstractFlagEncoder.java
EncodedValue.java
EncodedDoubleValue.java

LevelGraphStorage.java:
double weight = (edge.getFlags() >> 2) / WEIGHT_FACTOR;

RAMIntDataAccess.java
return (short) (segments[bufferIndex][index] >> 16);

zhiqiangZHAO:
The test should be like:
speedEncoder = new EncodedDoubleValue("Speed", 56, 8, 1, 30, 255);
Long flags = -72057594037927936L;
assertEquals(255, speedEncoder.getDoubleValue(flags), 0.01);

karussell:
I'll fix. In RAMIntDataAccess it does not matter as it is casted to short anyway. The same reasoning applies why a signed shift is okay for the BitUtil* classes

PR #420, opened 2015-05-22
Round trip calculation

karussell:
I've been working on a simple round trip calculation in the last days and the results are not perfect yet but okayish. The method is using an alternative route calculation under the hood so as a side effect we have now basic alternative route calculation support, see the alternate_route branch.

The UI is not yet adapted only a rough workaround can make the round trip and/or the alternatives working there:
 * checkout and build graphhopper and start graphhopper web
 * Provide two locations, where only the first one is actually necessary and the additional `api.round_trip.max=1` parameter e.g. [this link](http://localhost:8989/?point=50.892639%2C13.444519&point=51.153509%2C13.233032&api.round_trip.max=1)
 * The alternative route calculation needs a similar parameter `api.alternatives.max=1`, multiple alternatives are returned using multiple `Path`s leading to a strange additional green line in the UI

Still do be done for round trip:
 - [x] update to latest master
 - [ ] filter round trips by the enclosing area
 - [x] fix UI

In another issue we then improve alternative routes:
 - [ ] calculate 'share' to improve quality
 - [x] breaking change for GraphHopper.route necessary where we return `List<AlternativeRoute>` and every AlternativeRoute is holding multiple `Path`s
 - [ ] add a similar penalty method that we are already using for round trip calculation
 - [ ] fix UI

Later we also need:
 - [ ] improve performance
 - [ ] do this for edge based traversal
 - [ ] do this for CH


karussell:
Closing due to #640

PR #421, opened 2015-05-23
Fix issue 412 bicycle routing issues for footway

ratrun:
Fix issues as discussed under issue #412 for handling of footways:

Changing the preference of highway=footway or path together with bicycle=yes from AVOID_IF_POSSILBE to UNCHANGED.
Avoid boosting speed for highway=footway from a fast surface.
Decrease the assumed speed for highway=path from 18 to 12.
Adding of a few more tests.
Change testing of explicit 4 into PUSHING_SECTION_SPEED in some tests.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -155,7 +155,7 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
         setHighwaySpeed("steps", PUSHING_SECTION_SPEED / 2);
 
         setHighwaySpeed("cycleway", 18);
-        setHighwaySpeed("path", 18);
+        setHighwaySpeed("path", 12);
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
```diff
@@ -349,13 +349,35 @@ public long handleWayTags( OSMWay way, long allowed, long relationFlags )
     int getSpeed( OSMWay way )
     {
         int speed = PUSHING_SECTION_SPEED;
+        Integer sInt = null;
+        Integer hwInt = null;
```
karussell: can you rename to surfaceSpeed and highwaySpeed?

karussell:
Thanks! I'll do just a minor renaming (no need for you)

karussell:
Minor tweak regarding the naming and using pushing sections not only footway, all tests still passing. Please review here the rebased commit: https://github.com/graphhopper/graphhopper/commit/e68e504890d481f1fbf6cae8db88120181ba129b

PR #422, opened 2015-05-25
Do not decrease preference for highway=track with maxSpeed >= avoidSpeedLimit

ratrun:
The preference should not decrease for the following case: `highway=track` and `maxspeed=100` . This was detected as feedback for the change at #414 on the https://lists.openstreetmap.org/pipermail/talk-de/2015-May/111195.html for highway=track.
According to taginfo there are indeed some ways tagged in this combination. 
-> Now we do not decrease the preference for this case and cases where the highway type is in the prefered highway type list.


karussell:
Interesting, thanks!

Issue #330, opened 2015-02-16
Strange Routing (Cycle)


ximex:
I tested around with the new routing implementation at osm (https://github.com/openstreetmap/openstreetmap-website/commit/250bb47cc2723c7af6fac0a6e9e900c1a50f9292)

I found this strange routing:
This first routing is much longer as the second one. I place the destination only some meters west in the second routing. The osm data looks correct.
![graphhopper](https://cloud.githubusercontent.com/assets/1078640/6217929/5af88d96-b619-11e4-972b-db720bc6902c.png)
![graphhopper2](https://cloud.githubusercontent.com/assets/1078640/6217930/5af999c0-b619-11e4-86ab-28be95445fed.png)

karussell:
GraphHopper bike routing is not about the shortest path, it is about the fastest but also including priorities like avoiding ugly surface, prefering a medium increase etc. Have a look at the time. They are similar:
https://graphhopper.com/maps/?point=47.8947%2C16.1056&point=47.900074%2C16.106408&vehicle=bike&elevation=true&layer=TF%20Cycle

So what do you think is the problem with the longer path, except that is longer in the distance :) ?

ximex:
first it is much longer. yes the time is nearly the same. from my experience in realworld this couldn't be right.
it's correct that the longer path has a smaller incline and a smother surface. But for me the distance should have more weight.
And if you look: the length of the route on highway=track is approx. as long as by the other route.

karussell:
> But for me the distance should have more weight.

Well, that is true for you but probably not for others and that is the reason bike routing is not easy :)

Maybe you could tweak the routing a bit and avoid such a big penalty directly in the code for that area? You'll need to tweak the BikeCommonFlagEncoder. Installation of GraphHopper is trivially once you have java installed.

We will make this more flexible from outside in a later version, but this personalization won't be accessible for OSM.org I fear ...

ximex:
i checked the data again in detail:
Short route: residential and track (grade1, grade2, "no grade")
Long route: tertiary, residential and track ( only "no grade")

Track Setting: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java#L156
Grade Settings: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java#L116

[...]

PR #423, opened 2015-05-25
Add support for class:bicycle

ratrun:
This change adds support for the tag [`class:bicycle'](http://wiki.openstreetmap.org/wiki/Class:bicycle)

The implementation uses the subtypes `touring`, `mtb` and `roadcycling`. 
See also #330.


karussell:
Nice & Thanks, also interesting :) !

PR #427, opened 2015-05-29
Android Sample: perform cleanup on exit, improvements at manifest

devemux86:
As users have posted in our [forum](https://groups.google.com/forum/#!topic/mapsforge-dev/Br1UQiMDwkk), at GH Android Sample there isn't any Mapsforge cleanup on app exit.
This PR handles this.

Also since Mapsforge doesn't yet support hardware acceleration (we're working on that) I set it to false in manifest, along with updating the sdk versions.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -1,29 +1,29 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="com.graphhopper.android"
-          android:versionCode="1"
-          android:versionName="0.1" >
+    package="com.graphhopper.android"
+    android:versionCode="1"
+    android:versionName="0.1" >
 
     <!-- mapsforge cache and saving maps -->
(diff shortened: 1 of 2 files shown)


karussell:
Thanks! Will it still run on my oldish 2.3 Android device ;) ?

Regarding the onDestroy: wouldn't it be simpler for mapsforge users to have a helper method which does this automatically or at least in one method call?

devemux86:
>Thanks! Will it still run on my oldish 2.3 Android device ;) ?

Yes we have min Android 2.3, and I still perform tests on that platform too. :)

>Regarding the onDestroy: wouldn't it be simpler for mapsforge users to have a helper method which does this automatically or at least in one method call?

You are right, I have thought too that the cleanup needs simplification (I'll open an issue).

devemux86:
BTW you can proceed with this PR if you like.
When we simplify Mapsforge cleanup I'll post here too ([issue](https://github.com/mapsforge/mapsforge/issues/620) opened).

karussell:
Thanks!

PR #431, opened 2015-06-01
Avoidanceattributestorage

engaric:
Attempt pull request of external stored avoidance attributes


engaric:
This was supposed to be to engaric branch.

karussell:
Ok. Still if you think that there are valueable pieces at some point feel free to post them as individual PRs!

Issue #424, opened 2015-05-27
get/setRadians misleading


karussell:
Some minor notes, which I stumbled over when [parsing the turn angle from json response](https://github.com/graphhopper/directions-api-java-client/commit/569d91a004fbf107f8d8579acb501f736f15dc1e#diff-8e8e01ea3bb10fbd1be8f469909ae512R198) in the java client:

 * getRadians should be renamed to getTurnAngle otherwise this is misleading naming. E.g. setRadian(getRadian()) should leave radian unchanged.
 * why are there two methods? .setRadian(deltaInOut).setDirOfRotation(deltaOut)
 * instead of 'double tmpRadian = Math.PI - clockwise * radian; tmpRadian *= clockwise;' shouldn't we do 'double tmpRadian = Math.PI * clockwise - radian;'
 * avoid encoder.K_ROUNDABOUT and use FlagEncoder.K_ROUNDABOUT

assigned @jansoe but of course I also can do this :)

jansoe:
I will do it.

[] rename getRadians to getTurnAngle is important!
[] refactor double tmpRadian = Math.PI * clockwise - radian
[] avoid encoder.K_ROUNDABOUT and use FlagEncoder.K_ROUNDABOUT
[] better documentation 

> why are there two methods? .setRadian(deltaInOut).setDirOfRotation(deltaOut)

set Radian just determines the radian (clockwise) between entrance (in) and exit (out).
setDirOfRotation determines the driving direction. Thus they need different input.

karussell:
@jansoe do you agree with my changes?

Issue #289, opened 2014-11-27
Via instruction is wrong


karussell:
The first instruction after a via point has to be recalculated in PathMerger. Otherwise we could get e.g. 'continue' when we need 'turn around' like [here](https://graphhopper.com/maps/?point=49.870245%2C11.892602&point=49.870628%2C11.892828&point=49.870245%2C11.892602&layer=Lyrk). Or fix this via merging all edges into one Path object and creating instructions out of that.

And not sure if the following is a bug: the via point should not insert a duplicate point in the pointlist, instead we need an explicit skip of FinishInstructions when adding to the point list. BUT keep in mind that we will have to tweak the code in Instruction.createJson

jansoe:
also include u-turn instructions for start points with direction (once #434 gets merged)

boldtrn:
I will close here as this is a duplicate at #1064 and the issue will be fixed via #1073.

Issue #225, opened 2014-06-24
Specify direction for start and end


crowmagnumb:
I have a case in which the starting direction on the street, as well as the starting point, is effectively known.  If I use the route tool this starting direction is not taken into account when I try to go from this point to another point. The directions will just start off saying "Continue On ..." when in fact they would preferably say "Turn Around".  However, I don't always *want* them to have to turn around as driving around the block to get going the other way is usually preferable (especially if it's a big truck). Is there a way to pass in the preferable starting direction.

jansoe:
I created a branch (#434) that addresses this feature. Unfortunately it is limited to the non-CH case (flexibility mode). Anyhow there are some conventions to be made, where I would appreciate your feedback. 
I implemented it such that one can specify the  ```heading``` parameter for start/stopover/end points as north based clockwise angle measured in radian (0, 2pi).  Are there preferences or reasons for different angle conventions (e.g. degree (0, 360) or (-180, 180), east-based, ...)?
Or even other suggestions?

devemux86:
I would suggest to follow the common bearing guidelines, e.g. like Android describes in ```Location``` class:
http://developer.android.com/reference/android/location/Location.html#getBearing()

devemux86:
BTW how do you think will the arrival ```heading``` parameter will work with the via points and end point ?

karussell:
is the bearing north based?

devemux86:
http://en.wikipedia.org/wiki/Bearing_(navigation)#Bearing_measurement
Yes, though the reference direction of 0° can be the magnetic of true north.

You can see also her:
http://en.wikipedia.org/wiki/Course_(navigation)

jansoe:
thanx, so I'll probably change to to degree.
 
> BTW how do you think the arrival heading parameter will work with the via points and end point ?

Sorry I do not really understand the question, but I'll try to answer:  if general ```heading=0``` means that the head of the car is pointing north, as well at start or via/end points.

karussell:
@devemux86 maybe if it helps: there are two options for via points (or end point). The first one tries to avoid a u-turn at via points and the second one enforces a heading e.g. to make sure that a truck arrives at a via point from a specific direction (e.g. otherwise it would be harder for the truck to enter a gate etc)

jansoe:
An implementation of this feature is now available by #434.

Nonetheless, it currently works only in the 'flexibility mode', i.e. without CH.

[...]

PR #434, opened 2015-06-03
Directions

jansoe:
still to do
* [x] handle case of query graph request at tower nodes
* [x] tests

### DIFF
--- a/core/src/main/java/com/graphhopper/GHRequest.java
+++ b/core/src/main/java/com/graphhopper/GHRequest.java
@@ -22,23 +22,29 @@ import com.graphhopper.util.Helper;
 import com.graphhopper.util.shapes.GHPoint;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 
 /**
  * GraphHopper request wrapper to simplify requesting GraphHopper.
  * <p/>
+ *
  * @author Peter Karich
  * @author ratrun
  */
 public class GHRequest
 {
     private String algo = "";
-    private List<GHPoint> points;
+    private final List<GHPoint> points;
     private final WeightingMap hints = new WeightingMap();
     private String vehicle = "";
     private boolean possibleToAdd = false;
     private Locale locale = Locale.US;
+
+    // List of favored start (1st element) and arrival heading (all other).
+    // Headings are north based azimuth (clockwise) in (0, 360) or NaN for equal preference
+    private final List<Double> favoredHeadings;
 
     public GHRequest()
     {
@@ -48,47 +54,150 @@ public class GHRequest
     public GHRequest( int size )
     {
         points = new ArrayList<GHPoint>(size);
+        favoredHeadings = new ArrayList<Double>(size);
         possibleToAdd = true;
     }
 
     /**
-     * Calculate the path from specified startPlace (fromLat, fromLon) to endPlace (toLat, toLon).
+     * Set routing request from specified startPlace (fromLat, fromLon) to endPlace (toLat, toLon)
+     * with a preferred start and end heading.
+     * Headings are north based azimuth (clockwise) in (0, 360) or NaN for equal preference.
+     */
+    public GHRequest( double fromLat, double fromLon, double toLat, double toLon,
+                      double startHeading, double endHeading )
+    {
+        this(new GHPoint(fromLat, fromLon), new GHPoint(toLat, toLon), startHeading, endHeading);
+    }
+
+    /**
+     * Set routing request from specified startPlace (fromLat, fromLon) to endPlace (toLat, toLon)
      */
     public GHRequest( double fromLat, double fromLon, double toLat, double toLon )
     {
         this(new GHPoint(fromLat, fromLon), new GHPoint(toLat, toLon));
     }
 
+
     /**
-     * Calculate the path from specified startPlace to endPlace.
+     * Set routing request from specified startPlace to endPlace with a preferred start and end heading.
+     * Headings are north based azimuth (clockwise) in (0, 360) or NaN for equal preference
      */
-    public GHRequest( GHPoint startPlace, GHPoint endPlace )
+    public GHRequest( GHPoint startPlace, GHPoint endPlace, double startHeading, double endHeading )
     {
         if (startPlace == null)
             throw new IllegalStateException("'from' cannot be null");
 
         if (endPlace == null)
             throw new IllegalStateException("'to' cannot be null");
+
         points = new ArrayList<GHPoint>(2);
         points.add(startPlace);
         points.add(endPlace);
+
+        favoredHeadings = new ArrayList<Double>(2);
+        validateAzimuthValue(startHeading);
+        favoredHeadings.add(startHeading);
+        validateAzimuthValue(endHeading);
+        favoredHeadings.add(endHeading);
     }
 
+    public GHRequest( GHPoint startPlace, GHPoint endPlace )
+    {
+        this(startPlace, endPlace, Double.NaN, Double.NaN);
+    }
+
+
+    /**
+     * Set routing request
+     *
+     * @param points          List of stopover points in order: start, 1st stop, 2nd stop, ..., end
+     * @param favoredHeadings List of favored headings for starting (start point) and arrival (via and end points)
+     *                        Headings are north based azimuth (clockwise) in (0, 360) or NaN for equal preference
+     */
+    public GHRequest( List<GHPoint> points, List<Double> favoredHeadings )
+    {
+        if (points.size() != favoredHeadings.size())
+            throw new IllegalArgumentException("Size of headings (" + favoredHeadings.size() +
+                    ") must match size of points (" + points.size() + ")");
+
+        for (Double heading : favoredHeadings)
+        {
+            validateAzimuthValue(heading);
+        }
+        this.points = points;
+        this.favoredHeadings = favoredHeadings;
+    }
+
+    /**
+     * Set routing request
+     *
+     * @param points List of stopover points in order: start, 1st stop, 2nd stop, ..., end
+     */
     public GHRequest( List<GHPoint> points )
     {
-        this.points = points;
+        this(points, Collections.nCopies(points.size(), Double.NaN));
     }
 
-    public GHRequest addPoint( GHPoint point )
+    /**
+     * Add stopover point to routing request.
+     *
+     * @param point          geographical position (see GHPoint)
+     * @param favoredHeading north based azimuth (clockwise) in (0, 360) or NaN for equal preference
+     */
+    public GHRequest addPoint( GHPoint point, Double favoredHeading )
     {
         if (point == null)
             throw new IllegalArgumentException("point cannot be null");
+
+
         if (!possibleToAdd)
             throw new IllegalStateException("Please call empty constructor if you intent to use "
                     + "more than two places via addPlace method.");
 
         points.add(point);
+        validateAzimuthValue(favoredHeading);
+        favoredHeadings.add(favoredHeading);
         return this;
+    }
+
+    /**
+     * Add stopover point to routing request.
+     *
+     * @param point geographical position (see GHPoint)
+     */
+    public GHRequest addPoint( GHPoint point )
+    {
+        addPoint(point, Double.NaN);
+        return this;
+    }
+
+    /**
+     * @return north based azimuth (clockwise) in (0, 360) or NaN for equal preference
+     */
+    public double getFavoredHeading( int i )
+    {
+        return favoredHeadings.get(i);
+    }
+
+    /**
+     * @return if there exist a preferred heading for start/via/end point i
+     */
+    public boolean hasFavoredHeading( int i )
+    {
+        if (i >= favoredHeadings.size())
+            throw new IndexOutOfBoundsException("Index: " + i + " too large for list of size " + favoredHeadings.size());
+
+        return !Double.isNaN(favoredHeadings.get(i));
+    }
+
+    // validate Azimuth entry
+    private void validateAzimuthValue( Double heading )
+    {
+        // heading must be in (0, 360) oder Nan
+        if (!Double.isNaN(heading) && ((Double.compare(heading, 360) > 0) || (Double.compare(heading, 0) < 0)))
+        {
+            throw new IllegalArgumentException("Heading " + heading + " must be in range (0,360) or NaN");
+        }
     }
 
     public List<GHPoint> getPoints()
@@ -164,9 +273,12 @@ public class GHRequest
         for (GHPoint point : points)
         {
             if (res.isEmpty())
+            {
                 res = point.toString();
-            else
+            } else
+            {
                 res += "; " + point.toString();
+            }
         }
         return res + "(" + algo + ")";
     }
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -26,9 +26,13 @@ import com.graphhopper.routing.*;
 import com.graphhopper.routing.ch.PrepareContractionHierarchies;
 import com.graphhopper.routing.util.*;
 import com.graphhopper.storage.*;
-import com.graphhopper.storage.index.*;
+import com.graphhopper.storage.index.LocationIndex;
+import com.graphhopper.storage.index.LocationIndexTree;
+import com.graphhopper.storage.index.QueryResult;
 import com.graphhopper.util.*;
 import com.graphhopper.util.shapes.GHPoint;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
@@ -36,12 +40,10 @@ import java.text.SimpleDateFormat;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicLong;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 /**
  * Easy to use access point to configure import and (offline) routing.
  * <p/>
+ *
  * @author Peter Karich
  * @see GraphHopperAPI
  */
@@ -87,7 +89,8 @@ public class GraphHopper implements GraphHopperAPI
     private double osmReaderWayPointMaxDistance = 1;
     private int workerThreads = -1;
     private boolean calcPoints = true;
-    // utils    
+    // utils
+    private static final AngleCalc ac = new AngleCalc();
     private final TranslationMap trMap = new TranslationMap().doImport();
     private ElevationProvider eleProvider = ElevationProvider.NOOP;
     private final AtomicLong visitedSum = new AtomicLong(0);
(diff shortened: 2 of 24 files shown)


karussell:
Thanks!

Will comment inline...

core/src/main/java/com/graphhopper/GHRequest.java
```diff
@@ -72,14 +75,17 @@ public GHRequest( GHPoint startPlace, GHPoint endPlace )
...
     }
 
     public GHRequest( List<GHPoint> points )
     {
         this.points = points;
+        initDirectionsList(points.size());
     }
 
-    public GHRequest addPoint( GHPoint point )
+    public GHRequest addPoint( GHPoint point , Double preferredDirection)
```
karussell: the comment regarding the unit of preferredDirection should go here

core/src/main/java/com/graphhopper/GHRequest.java
```diff
@@ -40,6 +38,10 @@
     private boolean possibleToAdd = false;
     private Locale locale = Locale.US;
 
+    // List of preferred start (1st element) and arrival directions (all other). 
+    // Directions are north based azimuth (clockwise) in (0,2pi) or NaN for equal preference
+    private List<Double> preferredDirections;
```
karussell: We should make points and preferredDirections make final, and therefor setDir method should not change the ref and just fill it

core/src/main/java/com/graphhopper/GHRequest.java
```diff
@@ -88,9 +94,87 @@ public GHRequest addPoint( GHPoint point )
...
+      this.preferredDirections = directions;
+      return this;
+    }
+
+    /**
+     * Sets preferred direction for a specific point of the request
+     * @param direction north based azimuth (clockwise) in (0,2pi) or NaN for equal preference
+     * @param position point for which direction is set. Starts with 0 for the start point  
+     */
+    public GHRequest setPreferredDirection(Double direction, int position)
```
karussell: can we avoid this method? I do not see a real need.

devemux86:
:+1:
Is this related to #225 ?

jansoe:
Yes (but will not work with CH')

core/src/main/java/com/graphhopper/routing/QueryGraph.java
```diff
@@ -356,6 +352,98 @@ private void createEdges( GHPoint3D prevSnapped, int prevWayIndex, GHPoint3D cur
...
+                VirtualEdgeIteratorState reverseEdge = virtualEdges.get(virtNodeIDintern * 4 + getPosOfReverseEdge(edgePos));
+                reverseEdge.setDispreferedEdge(true, true);
+            }
+        }
+    }
+
+    /**
+     * set edgeId disprefered (in direction adjNodeId)
+     */
+    public void setDispreferedEdge(int edgeId, int adjNodeId)
```
karussell: we need to find a better name (also prefer**r**ed) or maybe setInferiorEdge? And the sout statements should be removed. And do we really need to make this low level method public?

core/src/main/java/com/graphhopper/routing/VirtualEdgeIterator.java
```diff
@@ -135,6 +132,12 @@ public EdgeIteratorState setName( String name )
     }
 
     @Override
+    public boolean getBoolean(int key, boolean _default, PMap kwargs)
```
karussell: why the additional PMap here? fetching from a map while traversing millions of nodes should not be favoured. Instead map.get once into a local variable of the Weighting and pass this to this method e.g. via a simple Object param?

core/src/main/java/com/graphhopper/routing/VirtualEdgeIteratorState.java
```diff
@@ -143,7 +144,40 @@ public EdgeIteratorState setName( String name )
         this.name = name;
         return this;
     }
+    
+    @Override
+    public boolean getBoolean(int key, boolean _default, PMap kwargs )
+    {
+        if (key == EdgeIteratorState.DISPREFERED_STARTSTOPEDGE)
```
karussell: public constant variables should get better names and be shorter and/or better readable -> INFERIOR_START_STOP_EDGE or just INFERIOR_EDGE_KEY?

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -42,6 +42,7 @@
 public class EncodingManager
 {
     public static final String CAR = "car";
+    public static final String CARSTOPOVER = "carstopover";
```
karussell: same here: CAR_STOPOVER = "car_stopover" or do we need it now at all as the FastestWeighting is directly influenced?

core/src/main/java/com/graphhopper/routing/QueryGraph.java
jansoe: this method is requiered to disprefer the incoming edge at via points.
one could achieve this also by calculating the incoming azimuth, and then use the above method. But in my feeling this would be a quite indirect.

core/src/main/java/com/graphhopper/routing/util/ExcludeIdEdgeFilter.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class ExcludeIdEdgeFilter implements EdgeFilter
+{
+    private final TIntHashSet rejectedEdgeIds;
+    private final EdgeFilter baseFilter;
+
+    /**
+     * Creates an edges filter which extends an existing edge filter to exclude specific edge ids
+     */
+    public ExcludeIdEdgeFilter( EdgeFilter baseFilter, List<Integer> edgeIds )
```
karussell: it is okay to use TIntHashSet for the public constructor, not sure if copying into another set is necessary (decouples both sets of course)

core/src/main/java/com/graphhopper/routing/util/ExcludeIdEdgeFilter.java
```diff
@@ -0,0 +1,57 @@
...
+    {
+        this.baseFilter = baseFilter;
+        this.rejectedEdgeIds = new TIntHashSet(edgeIds.size());
+        rejectedEdgeIds.addAll(edgeIds);
+    }
+
+    @Override
+    public final boolean accept( EdgeIteratorState iter )
+    {
+        return baseFilter.accept(iter)? !rejectedEdgeIds.contains(iter.getEdge()) : false;
```
karussell: -> baseFilter.accept(iter) && rejectedEdgeIds.contains(iter.getEdge())

core/src/main/java/com/graphhopper/routing/util/ExcludeIdEdgeFilter.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public String toString()
+    {
+        return baseFilter.toString() + "excluding: " + rejectedEdgeIds.toString();
```
karussell: ", excluding:

core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
```diff
@@ -26,6 +26,8 @@
  */
 public interface EdgeIteratorState
 {
+    final static int DISPREFERED_STARTSTOPEDGE = -1;
```
karussell: Probably it is okay here, but still not sure. Or QueryGraph or Weighting :) ?

web/src/main/java/com/graphhopper/http/GHBaseServlet.java
```diff
@@ -120,6 +123,21 @@ protected String getParam( HttpServletRequest req, String string, String _defaul
         return new String[0];
     }
 
+    protected List<Double> getDoubleParamList( HttpServletRequest req, String string )
```
karussell: string should be renamed to key

core/src/main/java/com/graphhopper/GHRequest.java
jansoe: it is not necessary.
I just thought it is more convenient for using the low level API.  For example specifying only preferred start direction (without caring how many other points there are)

core/src/main/java/com/graphhopper/GHRequest.java
jansoe: we can do this, but it would brake at lot of points the common Usage of GHRequest.addPoint()
e.g. https://github.com/graphhopper/graphhopper/blob/master/core/src/test/java/com/graphhopper/GraphHopperTest.java#L519

core/src/main/java/com/graphhopper/GHRequest.java
karussell: Then I would just call addPoint(point, dir) the first time

core/src/main/java/com/graphhopper/GHRequest.java
karussell: no, I do not mean this. But I can refactor once you are done and you can review and see how this feels.

jansoe: ok, i get it now. will do it and you might refactor afterwards again :)

core/src/main/java/com/graphhopper/routing/VirtualEdgeIterator.java
jansoe: because, I used the PMap to pass the ```reverse``` setting of the routing algorithm further on. Thus it is contains a property not fixed for a weighting. So maybe right now, just use a boolean reverse instead of a PMap and refactor once we need more parameter?

core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
jansoe: I would not put it into QueryGraph, but Weighting seems also a good location for me

core/src/main/java/com/graphhopper/routing/VirtualEdgeIterator.java
karussell: Yes, I would just use the key and a boolean for now.

core/src/main/java/com/graphhopper/GHRequest.java
jansoe: quite true

core/src/main/java/com/graphhopper/routing/VirtualEdgeIterator.java
jansoe: to be clear, I would do ```getBoolean(int key, boolean _default, boolean reverse)```

karussell: Now I'm confused. Why do we need to specify a _default? I would implicitly assume that every key defines its own defaults indirectly.

jansoe: I thought the default construct is helpful, not to implement the key logic in every EdgeIteratorState but only where it is necessary (in our case in the VirtualEdgeIteratorState ). 
That is, if an Edge does not know how to handle a key, it just returns the default.

I think this behaviour is not really necessary, but  saved me from a lot of code duplications (```if key == DISPREFERED_EDGE return false```)

jansoe:
regarding the wording: i liked the dichotomy of ```preferredDirection``` vs. ```dispreferredEdge```, so  shall we do now ```superiorDirection```and ```inferiorEdge``` ? Or what do you think of ```favoredDirection``` and ```unfavoredEdge```?

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
jansoe: uups, yes. old code ....

karussell:
this is good: 'favoredDirection and unfavoredEdge'

karussell:
> setDispreferedEdge(int edgeId, int adjNodeId)
> this method is requiered to disprefer the incoming edge at via points.
> one could achieve this also by calculating the incoming azimuth, and then use the above method. 
> But in my feeling this would be a quite indirect. 

Hmmh, not sure what you mean here. I would prefer to have just two simple methods in QueryGraph: enforceDirection and dropDirectionEnforcement without any ifs before but probably with more required parameters. Even if we pass the endEdge if it is not necessary. BTW: why -2 in `paths.get(placeIndex - 2).getFinalEdge()`?

Also for me `QueryGraph.setDispreferedEdge` is misleading as it suggests one can set any edge but then throws exception for none-virtual edges.

>  Include instructions for UTurn at start

Why only at start? E.g. if costs are low u-turn can happen also elsewhere. But again: we should move this out of this issue and into: #289

jansoe:
> not sure what you mean here. I would prefer to have just two simple methods in QueryGraph: enforceDirection and dropDirectionEnforcement without any ifs before but probably with more required parameters

I mean, the two methods are applied in two different scenarios. In the one scenario I know only the direction, I want to prefer, in the other scenario I have no clue which direction I want to prefer but already I know the edge to disprefer. One could use the information in the second scenario, extract from the edge an direction, and then use this direction to feed it into the first scenario. But I dislike this, as it is unnecessary computational overhead. 
Regarding your suggestion to use one method for both, its sounds like putting two things together which do not belong together (I exaggerate here a bit, but in the end it has to look similar like this)
```java
favorDirectionOrUnfavorVirtualEdge(double favoredDirection, QueryResult queryPoint, boolean incoming, EdgeIteratorState unfavoredEdge)
{ if (unfavoredEdge == EdgeIteratorState.NO_EDGE)
{
favor direction
}
 else
{
unvafor edge
} 
```

To be more constructive, (1) in the GraphHopper Class I can remove a lot of ifs by doing all the validity checking within the QueryGraph and (2) a rename of  ```setDispreferedEdge``` in  ```setUnfavoredVirtualEdge``` should make the usage much more clear

This are my two cents, but if I still could not convince you, I will implement it like this, since you have the bigger picture in mind :)

jansoe:
> BTW: why -2 in paths.get(placeIndex - 2).getFinalEdge()

```placeIndex``` is initialised by 1, that is for ```placeIndex==1``` the first path is calculated and then stored in the list ```pathes``` at the first position (i.e. at 0 with a zero based index). This path is not affected by the requirement for straight via routing.
Then for the next path, i.e. ```placeIndex==2```, we have to get the final edge of the previous path (which was the first one), which is at the 0th position in the ```pathes``` list.

**Comment by jansoe** [comment]
>>    Include instructions for UTurn at start

> Why only at start? 

Here my reasoning was, that u-turns at via points are old behaviour. But with this code also u-turns could occur at start, which did not existed before. Therefore I regarded it as part of this issue, whereas the other u-turns seemed not directly related to this code.

But now I removed also the start u-turns to issue  #289

**Comment by jansoe** [comment]
open questions:
* [ ]  ```_default``` in ```EdgeIteratorState.getBoolean()```? (I'm fine with not)
* [ ] location of ```K_UNFAVORED_EDGE``` (I vote for FastestWeighting)
* [ ] 2 methods ```queryGraph.favorDirection``` and ```queryGraph.unfavorEdge``` or a single one (I prefer the former)

**Comment by jansoe** [comment]
* [ ] fix bug: direction parameters do not do not influence result for route? endpoint

**Comment by jansoe** [comment]
renamed ```direction``` parameter to ```heading```

File: web/src/main/java/com/graphhopper/http/GraphHopperServlet.java [hunk 13]
```diff
@@ -78,19 +79,49 @@ public void doGet( HttpServletRequest httpReq, HttpServletResponse httpRes ) thr
...
-            ghRsp = new GHResponse().addError(new IllegalArgumentException("Vehicle not supported: " + vehicleStr));
+            ghRsp.addError(new IllegalArgumentException("Vehicle not supported: " + vehicleStr));
         } else if (enableElevation && !hopper.hasElevation())
         {
-            ghRsp = new GHResponse().addError(new IllegalArgumentException("Elevation not supported!"));
-        } else
+            ghRsp.addError(new IllegalArgumentException("Elevation not supported!"));
+        } else if (favoredHeadings.size()>1 && favoredHeadings.size() != infoPoints.size())
+        {
+            ghRsp.addError(new IllegalArgumentException("#heading must be <=1 or equal #points"));
```
karussell: Here I slightly prefer `number of` over `#xy` :)

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -974,20 +1085,29 @@ public GHResponse route( GHRequest request )
...
             if (path.getTime() < 0)
+            {
                 throw new RuntimeException("Time was negative. Please report as bug and include:" + request);
+            }
 
             paths.add(path);
             debug += ", " + algo.getName() + "-routing:" + sw.stop().getSeconds() + "s, " + path.getDebugInfo();
 
+            // reset all direction enforcements in queryGraph to avoid influencing next path
+            queryGraph.clearUnfavoredStatus();
```
karussell: ah, we can't move this method into enforceHeading because we potentially call enforceHeading for start and end point, right?

karussell:
Looks now good to me!

core/src/main/java/com/graphhopper/GraphHopper.java
jansoe: correct :)

jansoe:
On my behalf, I'm done

karussell:
Would you mind to squash once again :) and move MyTest into the normal (j)unit testing?

jansoe:
> and move MyTest into the normal (j)unit testing

uups, this should not have been  comited , will remove it

devemux86:
I'm wondering if it's worth it to have also elsewhere the angle in human friendly (0, 360) e.g. like [here](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/util/RoundaboutInstruction.java#L70-L85)

karussell:
Hmmh, yes this should also be changed but probably in another issue: #424 

@jansoe the problem is now that turn_angle is in radian and already included in the json :(

karussell:
I would prefer consistent angles, and we can't revert turn_angle using degree without breaking several clients OR introducing two versions of the API for the switchover (see new #437). And as we already have an 'inconsistent state' (GPX export is using degree for gh:azimuth) I vote to keep using degree here and fix turn_angle via #437 later. 

Please vote for or against this :) !

devemux86:
I generally vote for using SI units in engines, though degrees (0-360) is the common way for angles.
Users can always convert the returned values later to their liking.

Of course you are right about existing implementation, an established API is hard to change.

karussell:
thanks. and yes, I would not have a problem with breaking the Java API (maybe we should do this even earlier) because this is relative easy to detect and fix for consuming developers, but the JSON API can't be changed as easy (this is what we now feel ourself with running the GH Directions API ;))

jansoe:
I vote for using degrees in this PR, and sometime, with other major changes,  also swap radians in roundabouts.

karussell:
Thanks a lot @jansoe ! Merged!

karussell:
BTW: should we throw an UnsupportedException if someone wants to use it with CH?

jansoe:
I'm undecided. On the one hand it will also work with CH in many cases, on the other hand, you sometimes get this artefacts ....

devemux86:
Great work @jansoe for a valuable feature!

karussell:
Yeah, me too. Still the problem is that CH is the default and people will report this as bug although this is 'known'. Maybe we throw an exception but allow via parameter in the GraphHopper class to make it still working?

PR #441, opened 2015-06-10
Update NameIndex.java

zhiqiangZHAO:
When write it, the address is in byte, not integer.
        names.setHeader(0, BitUtil.LITTLE.getIntLow(bytePointer));
        names.setHeader(4, BitUtil.LITTLE.getIntHigh(bytePointer));

### DIFF
--- a/core/src/main/java/com/graphhopper/search/NameIndex.java
+++ b/core/src/main/java/com/graphhopper/search/NameIndex.java
@@ -56,7 +56,7 @@ public class NameIndex implements Storable<NameIndex>
     {
         if (names.loadExisting())
         {
-            bytePointer = BitUtil.LITTLE.combineIntsToLong(names.getHeader(0), names.getHeader(1));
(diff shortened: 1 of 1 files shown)


karussell:
Ouch, thanks! Would you also have failing unit test for this (which is fixed after this change?)

zhiqiangZHAO:
I don't create unit test, since this error is obvious.
I think GH reade the bytePointer again to insert new names.
But actually when the GH loadExisting, it doesn't read new osm data, so, nobody really encouter problem with this.

karussell:
> I don't create unit test, since this error is obvious.

If there is a bug and no test fails we need a unit test to avoid the same bug appearing in the future.

>  I think GH reade the bytePointer again to insert new names.

I see, I'll try to recreate a test showing the problem (or if you would do it is appreciated :))

zhiqiangZHAO:
The test could be, write a nameIndex to disk, then loadExisting, then verify the bytePointer.
:)

PR #442, opened 2015-06-10
Update NameIndex.java

zhiqiangZHAO:
bytePointer is a long. not int.


karussell:
Ah, my previous commit is faulty: https://github.com/graphhopper/graphhopper/commit/1facdcadb195a6a39c635ba6a328e62b5e2b21fa

And we also need to change or better make obsolete the following in consuming classes: `if (nameIndexRef < 0) throw new IllegalStateException("Too many names are stored, currently limited to int pointer");`

karussell:
Fixed now

Issue #331, opened 2015-02-17
inconsistent behaviour of InstructionList.find


jansoe:
In general InstructionList.find(lat, lon, dist) finds the closest point on the route and then yields the upcoming instruction.
In the case the closest point on the route is the start or endpoint of an edge instead the previous instruction is returned.
See: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/util/InstructionList.java#L319

karussell:
You mean that in the case !validEdgeDistance the index++ could be missing? That would be indeed rare but ugly

jansoe:
yes, thanx for clarification :)

karussell:
Do you have a test case for this?

jansoe:
not yet, but I will write one

PR #444, opened 2015-06-10
fix for #331

jansoe:


karussell:
Thanks!

jansoe:
Question is, if it would ever have been occurred in a real world scenario :)

karussell:
still good to have it away :)

Issue #339, opened 2015-02-25
Separate list of traversal references for CH and normal edges


karussell:
Currently CH uses the same entry points (N_EDGE_REF) into the linked list of traversal references (E_LINKA, E_LINKB). This needs to be separated and therefor duplicated (e.g. N_CH_REF, E_CH_LINK_A, E_CH_LINK_A). But as the levels are then no longer necessary to store and also the shortcuts can be separated from edges we don't need more RAM (instead we'll even save a bit with #111)

This requires substantial refactoring of the `GraphStorage` retrieving a `base graph` and `routing graph` see [comment on 116](https://github.com/graphhopper/graphhopper/issues/116#issuecomment-75544401). This separation and the separation of edges and shortcuts (#111) will make multiple CH-preparations possible.

* fix bug in prepareGraph.getAllEdges -> do not return disconnected edge (ensure returning correct order) AND if disconnected in both directions they should be marked as removed
* sorting graph uses incorrect reference for baseGraph, probably better: create new scGraphWrapper
* Also investigate why there is exactly one node with no shortcuts and no upwards (first or last contracted node?), shouldn't there be more of them?
* fix documentation and examples

karussell:
This is now all internally possible and 'just' needs to be done API wise via adding multiple ch graphs in GraphHopperStorage besides the existing single chGraph. Once we have this we need #86 and the bug mentioned in #318 which will get a lot more urgent.

karussell:
Done via #447, doing multiple CH-prepared profiles in #464

PR #447, opened 2015-06-17
Refactoring of GraphHopperStorage to fix #111

### LINKED ISSUE #111: More efficient graph storage via separating shortcuts from edges

karussell:
This pull request is a major rewrite of some core parts of GraphHopper (storage and access of CHGraph, old name LevelGraph). This fixes #111. E.g. GraphHopperStorage now contains graphs like base graph and a ch graph which will allow for the physical separation of edges and shortcuts. 

Currently there are the following steps before we can merge this:
- [x] merge master into this
- [x] rename LevelGraph* to CHGraph*
- [x] separate node_ch_refs for CHGraphImpl to avoid base graph going through shortcuts
- [x] test with none CH algo on prepared graph should pass now
- [x] make base graph see only its edges (no shortcuts). edge count for level graph and base graph should change (e.g. getAllEdges.count)
- [x] all tests should pass now
- [x] fix all TODOs in the source code, and can we add edges count to locationindextree.calcChecksum? And can we avoid requiring the GraphHopperStorage and use a Graph instead in DijkstraBidirectionCHTest.createFactory
- [x] fix some bugs uncovered while refactoring, but do this in different issues
- [x] separate shortcut DataAccess in CHGraphImpl

Later we can also have multiple ch graphs but for that we should create a new issue besides #339

Some bugs uncovered while refactoring:
 - [x] setNodesHeader returns wrong used header size
 - [x] internalEdgeDisconnect -> use N_EDGE_REF for nodes.setInt
 - [x]  bug if next is called after setAdditionalField? if (extStorage.isRequireEdgeField()) iter.setAdditionalField(extStorage.getDefaultEdgeFieldValue());
 - [x] bug in test testFindShortcuts_Roundabout regarding skipped edges of shortcut 6,4
 - [x] bug in alledgeiterator detach => reverse can be incorrectly set if detach(false) and reverse=true, can only happen if detach is called on an already detached iterator which would involve a cast
 - [x] bug in alledgeiterator found: while iteration edgepointer can get bigger than valid edges and if it would be compacted then even array index out of bounds would get thrown
 - [x] if too many names are stored => wrong check for long value instead int! (check AllEdgesIterator and EdgeIterable)
 - [x] here it should be getNodes instead of getCapacity! removedNodes = new GHBitSetImpl((int) (nodes.getCapacity() / 4));
 - [x] shouldn't we throw an exception in CHGraphImpl instead of returning? public final void setLevel( int nodeIndex, int level ) { if (nodeIndex >= baseGraph.getNodes()) return;

(where marked they are fixed in the pull request but not master yet)


devemux86:
Thanks Peter for providing also a summary of API changes in changelog, it helps studying the code changes.

karussell:
If you find something you do not understand, please let me know!

Also this is now different and needs to be changed:
> call GraphHopperStorage.getBaseGraph(). To do routing with potentially shortcuts you call GraphHopperStorage.getGraph(encoder)

to "GraphHopperStorage implements now the Graph interface and delegates all necessary methods to the underlying base graph. To do routing you call getGraph(LevelGraph.class or Graph.class) where the parameter Graph.class currently returns the base graph and the behaviour is identical to GraphHopperStorage itself" (this is the convenient change I mentioned in the other issue)

karussell:
The additional refactoring for #339 will be done in a separate step. Now finalizing before like improving readability and performance checks. 

Also to all: as all tests pass please review.

While refactoring I found some bugs
 - [x] (fixed in branch+master) bug in test testFindShortcuts_Roundabout regarding skipped edges of shortcut 6,4
 - [ ] bug in alledgeiterator detach => reverse can be incorrectly set if detach(false) and reverse=true
 - [x] bug in alledgeiterator found: while iteration edgepointer can get bigger than valid edges and if it would be compacted then even array index out of bounds would get thrown
 - [ ] here it should be getNodes instead of getCapacity! removedNodes = new GHBitSetImpl((int) (nodes.getCapacity() / 4));

(where marked they are fixed in the pull request, only the first is also fixed in master too)

An open question is: move the BaseGraph implementation back into GraphHopperStorage to make everything simpler?

devemux86:
>An open question is: move the BaseGraph implementation back into GraphHopperStorage to make everything simpler?

And what about ```chGraph```?

karussell:
> And what about CHGraph?

CHGraph will stay separate. But moving BaseGraph back would have the advantage of avoiding the method delegation in GraphHopperStorage to the BaseGraph. Although I like the separation of storage and traversal stuff... but CHGraph currently also mixes storage and traversal :)

devemux86:
Ok you're probably right here.

karussell:
You mean moving it back into GraphHopperStorage? Just asking again as I'm still unsure ... probably I'll let if for now as it is until we find a reason pro or contra?

devemux86:
>You mean moving it back into GraphHopperStorage?

Yes. I'm a bit sceptical about ```GraphHopperStorage``` hosting [both](https://github.com/graphhopper/graphhopper/blob/issue_111/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java#L42-L44) types of graphs.
With the change it will be Base itself and still contain the CH?

karussell:
> I'm a bit sceptical about GraphHopperStorage hosting both types of graphs.

What do you mean? Or what do you think could improve the situation? Keep in mind that we'll need in the near future multiple chGraphs managed by the GraphHopperStorage. But via ghstorage.getGraph(class) it should be very simple to extend for more purposes or even other algorithms.

> With the change it will be Base itself and still contain the CH?

Yes, this would be a bit uglier design-wise but simpler.

devemux86:
I'm fine with current PR implementation.

>With the change it will be Base itself and still contain the CH?
>Yes, this would be a bit uglier design-wise but simpler.

Here I had second design thoughts but as you said it'll be simpler, so it's ok.

devemux86:
Well I suppose (depending on ram) it could work.
Sorry not having something usable in mind right now.

karussell:
Some numbers for the area 'Germany': 15% less storage and therefor RAM - as expected :) !
Routing same speed. But strangly CH routing is 20% slower, also preparation is 18% slower. To be investigated

devemux86:
:+1: 
What algorithm, bidirectional Dijkstra ?

karussell:
Yes, to better compare CH and none-CH I've picked bidir dijkstra in both cases -> See the measurement class (or graphhopper.sh measurement area.pbf)

karussell:
Although I only slightly improved the speed situation I'll merge this request soon as there are many advantages.

disadvantages, higly likely due to the additional EdgeAccess layer:
 * 12% slower CH preparation
 * 12% slower CH routing

advantages:
 * 10% less storage/RAM used
 * 5% faster routing, but highly likely via an unrelated change https://github.com/graphhopper/graphhopper/commit/3b197460a3a9aabb1dd5e8f47e47ef58bf49e668
 * separated edges from shortcuts storage, as well as CH entries (`nodesCH`) from normal traversal entries (`nodes`), allowing us to add multiple CH profiles in the near future
 * making other speed-up techniques easier pluggable in the future
 * better architecture via composition (EdgeAccess, CHGraphImpl and BaseGraph) instead of extension (`LevelGraphStorage extends GraphHopperStorage`)

devemux86:
I follow your code changes so :+1: for that.

karussell:
@devemux86 thanks for the feedback!

BTW: It would be now also possible to revert the speed we lost with #338 but we would need to introduce two integers for additional `ch refs` in the normal `edges` DataAccess storage, and the levels per node could be removed. Leading to a slight RAM increase of ~0.5GB = (2-1)*0.5GB for the world wide case. Then we could entirely avoid the LevelEdgeFilter for normal edges too, giving us a rough boost of 15-20%

BTW2: The levels should not be removed (although no longer needed right now). Instead we should just avoid to load them into RAM. Because when edge weights change and we need to recreate the CH profile the levels could then be reused and the CH preparation could be a lot faster then - at least from what I know from papers.

devemux86:
Peter I started playing with latest GH code.
I see with this PR old graphs do not work, right?

karussell:
Yes, because there is now a separate shortcuts and edges storage. Any problems I didn't think about with this :) ?

devemux86:
I try all the cases (e.g. loading a CH graph with a non CH config) and I see the same error message, let me conclude my tests and come back.

devemux86:
Yes I think something is not right (or I'm doing wrong!).

I have created a CH graph and I'm able to load it with ```graphHopper.setCHEnable``` either ```true``` or ```false```.

karussell:
That is like it should be :)

You can now run without performance lost normal routing algorithms even if CH is enabled, which means you can also load the graph without the ch datastructure and everything should work - at least in theory ;)

devemux86:
Oh I missed that step.
I support all possible GH cases in graph loading (via exception catch) - so it was strange to me.
It seems now that I can loosen the checks, thanks for the tip.

devemux86:
Since now we cannot be notified if a graph has CH or not, we need to always use ```graphHopper.setCHEnable``` at reader in order to be sure that we actually use the CH in CH graphs.
Since this ```set``` works with non CH graphs also, that would be not a problem.
On the contrary the false ```set``` means that we may loose CH in those graphs.
Is this thought correct?

karussell:
Ah, okay. I see what you mean. 

> Because the setCHEnable(true) works also with non CH graphs

That is a problem which we should indeed fix! Either we automatically detect CH and configure everything accordingly like we do partially for the flagEncoders, or we should throw an exception. Feel free to create an issue!

> On the contrary, the setCHEnable(false) means that we may loose CH in those graphs.

That is not such a big problem IMO as the results should be still correct, but yes this should be fixed probably at the same time.

Issue #451, opened 2015-06-25
SRTM elevation data provider not working


karussell:
Current URL http://dds.cr.usgs.gov/srtm/version2_1/SRTM3/ is gone. Similar to #346 our default URL is not longer working, switching to another for now. (until we host this ourself)

karussell:
No, it looks like the server still hosts the data but it seems not to return an error code even if the resource does not exist. instead it returns an html file.

karussell:
The real problem is the silent switch to the error stream in the download class

karussell:
Introduced in #426

Issue #281, opened 2014-11-06
FlagEncoder versioning


karussell:
If the access of a weight or something in the flags changed we have only the choice to increase the edge version which is ugly as people not using the one flagEncoder will also be forced to reindex it. So introduce a getVersion method and store this into the graph somehow.

Reading an older, different graph version is not the goal of this issue and is also a lot more complex but probably won't happen before 1.0

andreaswolf:
Do I understand this right that effectively a version should be stored for each flag (or rather, one version for the "bag of flags" of one encoder)? So a hypothetic structure could be:

```
[{
flag1: foo, flag2: bar, version: 1
},
{
flag3: baz, version: 3
}]
```

I think one byte should suffice for the version number, right?

karussell:
We could use a more fine grain solution but just one version per used encoder is sufficient for now. If the encoder version changes we throw an exception.

andreaswolf:
Ah, thought about it again… the used encoder versions can be stored once in the graph, right? My first draft was for each edge, but that does not make sense for a start.

karussell:
The purpose is entirely protecting the user from garbage data and so yes, once in the graph or in the properties would be sufficient. Maybe related to #313 and #285

karussell:
Fixed via #401

devemux86:
:+1:

devemux86:
Why not using the compatible encoders and skip the others [here](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/EncodingManager.java#L167-L170)?

karussell:
But when you load the graph sucessfully where only one encoders' version mismatches - how should the routing behave with the failing encoder?

Failing early is a lot better IMO

karussell:
The issue just makes sure that e.g. if the CarFlagEncoder is not used at all and there was a version increment happening we don't force the user with an edge version increment to refeed the data.

devemux86:
Thanks for the clarification Peter, I understand what you mean here.

devemux86:
@karussell I see that now if we try to load with the latest GH a graph created with GH just prior this PR, there is failure from [GraphHopperStorage](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java#L1407-L1411) with messages like:
```
WARNING: Encoding does not match:
[...]

PR #453, opened 2015-06-30
Android Sample: fix compile error, #451

devemux86:
Fix compile error in Android Sample due to #451 changes.


karussell:
Uhm, sorry for the improper fix https://github.com/graphhopper/graphhopper/commit/8384dcfe8e268aba1b6a306299c8df0b72e6057d the problem is that my Android environment is not existent ;)

karussell:
We should really fix #360 but never got the time for it ...

devemux86:
No problem, happy to help.

BTW I think the graphs in the server need updating, right now they're incompatible with the master due to #281.

karussell:
thanks, good hint as this does not happen automatically

karussell:
0.5 graphs should be also up-to-date

Issue #463, opened 2015-07-17
Avoid walking on major highways with sidewalk=no


neiljp:
There are various pedestrian ways around this roundabout - and walking across the junction is not advisable! I have set sidewalk=no where appropriate:
  http://overpass-turbo.eu/s/atC

This is an example unexpected route:
https://graphhopper.com/maps/?point=57.1554%2C-2.102809&point=57.153113%2C-2.099794&vehicle=foot&locale=en-US&elevation=true&layer=OpenStreetMap

While walking on service/residential (maybe unclassified?) may be reasonable (even preferable) with no sidewalk/pavement, the higher levels of road really shouldn't route pedestrians in the absence of a pavement, right? Just avoiding primary[_link?] upwards with sidewalk=no would be a good first change?

In future having notes regarding sidewalks on different sides might be useful (crossing roads where explicit crossings aren't included?), which I think isn't present at the moment (?), but just trying to avoid walking on a major road (itself) would be a good first stage.

Apologies in advance if this exists and is awaiting a database update; I don't know of many areas to check for this.

ratrun:
See code at https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java#L405 .
Your suggestions are already implemented. Note that there is no need for a "sidewalk=no" attribute on major in order to gain bigger avoidence level on major roads.

neiljp:
I don't claim to fully understand the implementation - I'm not sure how those two 'if' statements combine through weightToPrioMap - but the route proposed in my link above could either be construed as very poor or dangerous. In that particular case, Mounthooly (yellow road) one-way roads onto the roundabout have sidewalk on one side and a crossing between them, which would be the logical and preferred choice.

If the current code prefers sidewalk=yes/both/left/right and avoids highway=primary+, in my view either the map is tagged wrongly in some way (that isn't clear to me) or the routing is really non-ideal (and might need adjusting).

ratrun:
The suggested route is 374 m. The only alternative I see via Jute Street is 1.12km:

https://graphhopper.com/maps/?point=57.1554%2C-2.102809&point=57.15675%2C-2.101897&point=57.153113%2C-2.099794&vehicle=foot&locale=en-US&elevation=true&layer=OpenStreetMap

[...]

Issue #448, opened 2015-06-17
All UI parameters should be forwarded to the API, no ugly 'api.' hack


karussell:
This difference confused already several users when they have custom parameters that in the UI it is `api.xy` and for the API the URL parameter is just `xy`

Also it is inconsistent with `point`, `weighting` etc

Also update comment in GraphHopper.createWeighting

Issue #399, opened 2015-04-28
Priority bug for WORST value


karussell:
I got an email with:

*I think there may be a bug in the priority weighting (we’ve copied the motorcycle flag encoder) as if the GetDouble method uses the UNCHANGED value if the priority weighting returns zero which is also the weighting value for WORST.  This would mean that WORST roadtypes are treated the same as UNCHANGED.*

karussell:
The problem is that WORST is 0 which is also used for UNCHANGED in the special clause if ret == 0 then return UNCHANGED. Instead we should think about using negative values for bad priorities and still keep the order of the priorities. Then the following test should run through:

```java
@Test public void testUnchangedVersusWorstPriority_issue399()
{
    OSMWay way = new OSMWay(1);
    way.setTag("highway", "tertiary");
    long encoded = encoder.acceptWay(way);
    long flags = encoder.handleWayTags(way, encoded, 0);

    assertEquals(UNCHANGED.getValue(), encoder.getLong(flags, PriorityWeighting.KEY));
    assertEquals((double) UNCHANGED.getValue() / BEST.getValue(), encoder.getDouble(flags, PriorityWeighting.KEY), .1);

    MotorcycleFlagEncoder motorEncoder = new MotorcycleFlagEncoder(5, 5, 0)
    {
        @Override
        protected int calcPriority( OSMWay way, long relationFlags )
        {
            return PriorityCode.WORST.getValue();
        }
    };
    new EncodingManager(motorEncoder);

    encoded = motorEncoder.acceptWay(way);
    flags = motorEncoder.handleWayTags(way, encoded, 0);
    assertEquals(WORST.getValue(), motorEncoder.getLong(flags, PriorityWeighting.KEY));
    assertEquals((double) WORST.getValue() / BEST.getValue(), motorEncoder.getDouble(flags, PriorityWeighting.KEY), .1);
}
```

ratrun:
Why not just remove the value WORST from the PriorityCode enum ? 
This brings the benefit that priority changing upwards and downwards is symmetrical. It is unlogical that there should be 4 values for decreasing the priority down and only 3 values for boosting up.

karussell:
Yes, just removing WORST is also listed as possibility here, not sure. For 3 bits we have 8 possibilities and we should use them

karussell:
Will remove WORST for now as a workaround, still letting this open and postponing.

karussell:
[...]

PR #476, opened 2015-07-27
Foot handling improvement for avoidance handling triggered by issue 463.

ratrun:
See issue #463.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
@@ -397,17 +397,23 @@ public class FootFlagEncoder extends AbstractFlagEncoder
             weightToPrioMap.put(44d, AVOID_IF_POSSIBLE.getValue());
         }
 
-        if (way.hasTag("sidewalk", sidewalks))
-        {
-            weightToPrioMap.put(45d, PREFER.getValue());
-        }
-
         if (avoidHighwayTags.contains(highway) || maxSpeed > 50)
         {
-            weightToPrioMap.put(50d, REACH_DEST.getValue());
-
-            if (way.hasTag("tunnel", intendedValues))
-                weightToPrioMap.put(50d, AVOID_AT_ALL_COSTS.getValue());
+            if (way.hasTag("sidewalk", sidewalks))
+            {
+               weightToPrioMap.put(45d, REACH_DEST.getValue());
+            }
+            else
+            {
+                weightToPrioMap.put(45d, AVOID_AT_ALL_COSTS.getValue());
+            }
+        }
+        else
+        {
+            if (way.hasTag("sidewalk", sidewalks))
+            {
+                weightToPrioMap.put(45d, PREFER.getValue());
+            }
         }
     }
 
--- a/core/src/main/java/com/graphhopper/routing/util/PriorityWeighting.java
+++ b/core/src/main/java/com/graphhopper/routing/util/PriorityWeighting.java
@@ -29,7 +29,7 @@ import com.graphhopper.util.PMap;
 public class PriorityWeighting extends FastestWeighting
 {
     /**
-     * For now used only in BikeCommonFlagEncoder and MotorcycleFlagEncoder
+     * For now used only in BikeCommonFlagEncoder, FootEncoder and MotorcycleFlagEncoder
      */
     public static final int KEY = 101;
 
--- a/core/src/test/java/com/graphhopper/GraphHopperIT.java
+++ b/core/src/test/java/com/graphhopper/GraphHopperIT.java
@@ -200,8 +200,8 @@ public class GraphHopperIT
         req.getHints().put("heading_penalty", "300");
         GHResponse rsp = hopper.route(req);
 
-        assertEquals(873., rsp.getDistance(), 10.);
-        assertEquals(33, rsp.getPoints().getSize());
+        assertEquals(921., rsp.getDistance(), 10.);
+        assertEquals(38, rsp.getPoints().getSize());
     }
 
     @Test
--- a/core/src/test/java/com/graphhopper/routing/RoutingAlgorithmIT.java
+++ b/core/src/test/java/com/graphhopper/routing/RoutingAlgorithmIT.java
@@ -308,7 +308,7 @@ public class RoutingAlgorithmIT
         // prefer hiking route 'Teufelsloch Unterwaiz' and 'Rotmain-Wanderweg'        
         list.add(new OneRun(49.974972, 11.515657, 49.991022, 11.512299, 2365, 66));
         // prefer hiking route 'Markgrafenweg Bayreuth Kulmbach'
-        list.add(new OneRun(49.986111, 11.550407, 50.023182, 11.555386, 5165, 133));
+        list.add(new OneRun(49.986111, 11.550407, 50.023182, 11.555386, 6429, 122));
         runAlgo(testCollector, "files/north-bayreuth.osm.gz", "target/north-bayreuth-gh",
                 list, "FOOT", true, "FOOT", "fastest", true);
         assertEquals(testCollector.toString(), 0, testCollector.errors.size());
--- a/core/src/test/java/com/graphhopper/routing/util/FootFlagEncoderTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/FootFlagEncoderTest.java
@@ -202,6 +202,9 @@ public class FootFlagEncoderTest
         way.setTag("highway", "cycleway");
         assertEquals(PriorityCode.UNCHANGED.getValue(), footEncoder.handlePriority(way, 0));
 
+        way.setTag("highway", "primary");
+        assertEquals(PriorityCode.AVOID_AT_ALL_COSTS.getValue(), footEncoder.handlePriority(way, 0));
+        
         way.setTag("highway", "track");
         way.setTag("bicycle", "official");
         assertEquals(PriorityCode.AVOID_IF_POSSIBLE.getValue(), footEncoder.handlePriority(way, 0));
@@ -209,6 +212,16 @@ public class FootFlagEncoderTest
         way.setTag("highway", "track");
         way.setTag("bicycle", "designated");
         assertEquals(PriorityCode.AVOID_IF_POSSIBLE.getValue(), footEncoder.handlePriority(way, 0));
+        
+        way.setTag("highway", "cycleway");
+        way.setTag("bicycle", "designated");
+        way.setTag("foot", "designated"); 
+        assertEquals(PriorityCode.PREFER.getValue(), footEncoder.handlePriority(way, 0));
+        
+        way.clearTags();
+        way.setTag("highway", "primary");
+        way.setTag("sidewalk", "yes");
+        assertEquals(PriorityCode.REACH_DEST.getValue(), footEncoder.handlePriority(way, 0));        
     }
 
     @Test


karussell:
Hmmh, several tests are failing now. What causes this huge change in route quality?

ratrun:
I'm investigating the test failures. One failure happens with `heading_penalty`. Where can I find understandable documentation what `heading_penalty` and `heading` is about? I found the documentation on https://github.com/graphhopper/graphhopper/blob/master/docs/web/api-doc.md, but this does not tell me much. Are there any examples or is there somewhere a discussion when this got introduced?

karussell:
See #434
There are also heading-unrelated failures in RoutingAlgorithmIT

ratrun:
Thank you, now I understand the heading. In the test `testMonacoEnforcedDirection` the better solution is longer now. It is correct because sth direction instructs to go south at the start. The previous solution went north at the start. I have tested this with this URL

http://localhost:8989/?point=43.741069%2C7.426854&point=43.744445%2C7.429483&heading_penalty=300&heading=190&vehicle=foot&layer=Omniscale

Unfortunately the actual Monaco OSM data has very different highway tags compared to the checked in Monaco OSM data snapshot.

Now I need to look at `testNorthBayreuthFootFastestAnd3D()`.

ratrun:
The test `testNorthBayreuthFootFastestAnd3D` also needs to be adopted.

This is the current solution:

https://graphhopper.com/maps/?point=49.986111%2C11.550407&point=50.023182%2C11.555386&vehicle=foot&layer=Omniscale

With the modifcation we get routed on this longer solution, which is OK because there it contains a longer segment on the relation rwn "Markgrafenweg Bayreuth-Kulmbach", see

https://graphhopper.com/maps/?point=49.986111%2C11.550407&point=50.002554%2C11.538666&point=50.023182%2C11.555386&vehicle=foot&elevation=true&layer=WanderReitKarte

ratrun:
Can you please check it now?

karussell:
Thanks @ratrun for investigating this.

The Markgrafenweg thing is indeed more correct now, but the heading is incorrect. As 190 means south direction for the start, which is now north. (if another heading parameter would be provided the end or via could get a direction force too)

Hmmh, also there seems to be some UI issues, due to #448

I'll investigate why the heading does not work and fix the UI issue first.

karussell:
Can confirm that this all intended as it is. (The UI issue is fixed, still some improvements for heading necessary also with heading=190 for the second point and forcing a north heading is strange and will investigate this)

karussell:
Thanks @ratrun again!

neiljp:
Hi,
Should this fix resolve the original issue I reported?
(I've not seen it on the 'live' graphhopper instance on the website yet)

Also, I wanted to clarify that I was primarily objecting to major roads combined with no sidewalk; major roads with a ('good') sidewalk work fine, and if shorter are potentially better/shorter. In the example 'ideal' route given by (I think) @ratrun, the sidewalk on the (out)side of the roundabout would be preferable rather than going via the steps, but avoiding the section with no sidewalk.

karussell:
> Should this fix resolve the original issue I reported?

Yes, I think so. Please try it

> (I've not seen it on the 'live' graphhopper instance on the website yet)

Although we deploy even master branches we do so not automatically

neiljp:
I don't have a local install of GH to test, only the public instance.

karussell:
It is very easy to install one, only [3 commands](https://github.com/graphhopper/graphhopper/blob/master/docs/core/quickstart-from-source.md) if you have git and java

neiljp:
I've just tested this locally using the Scotland .pbf, and once I figured out I needed to change the router to use 'foot', it does route... but that provided is by no means ideal.
The test route I used was 'Canal Road, Aberdeen' to '57north'.

Existing online router provided non-ideal (walking, 1.36km, unsafe):
https://graphhopper.com/maps/?point=canal%20road%2C%20aberdeen&point=57north&vehicle=foot&locale=en-US&elevation=true&layer=Lyrk

Adding in an 'avoid no-sidewalk' reroute, this becomes (1.44km, safe):
https://graphhopper.com/maps/?point=canal%20road%2C%20aberdeen&point=57.154446%2C-2.099687&point=57north&vehicle=foot&locale=en-US&elevation=true&layer=Lyrk

Firstly, my local build just now, provides a route which I can best reproduce as follows (1.56km):
https://graphhopper.com/maps/?point=canal%20road%2C%20aberdeen&point=57.154446%2C-2.099687&point=57.153445%2C-2.098861&point=57north&vehicle=foot&locale=en-US&elevation=true&layer=Lyrk

Perhaps I have not configured it correctly locally (I don't think there is much to choose?) but that updated route for a (very similar to the original) test path seems a regression to me, based on the distance?

If I 'force' the route to go via Gallowgate (as per the original), it tries to turn off to go west as soon as possible, as if it is trying to go through the shopping centres and so on, eg. down Spring Garden -> Loch Street... which is about 1.79km.

As a separate but perhaps related issue, the router also still prefers the diagonal route via the steps, north of the big roundabout, rather than simply following the sidewalk around the edge of the roundabout.

karussell:
Based on distance and time this is indeed suboptimal. But as the [west north street](http://www.openstreetmap.org/way/137150843) is tagged with `sidewalk=left` it might just miss a bit of a configuration so that tertiary streets without a sidewalk are favoured over primary with sidewalk (?)

ratrun:
@karussell Yes. Currently `primary` and `tertiary` are both put into the avoidHighwayTags category and therefore the one with sidewalk is now choosen. By the way, for no obvious reason `secondary` is put into allowedHighwayTags, which does not make much sense. This should be moved to the avoidHighwayTags category.
I believe that we need two kind of avoidence types. One more strong for primary and secondary with the same ranking as currently. The other class could get `AVOID_IF_POSSIBLE`. I believe that I had this in the original code for bikes, but you didn't like it. Your argument was that this would be too complex. Would you accept it for the foot encoding?

karussell:
Hmmh, I would put tertiary and secondary into the allowedHighwayTags as those are less traffic (at least outer city). Or at least we should not put both into avoidHighwayTags IMO

> Your argument was that this would be too complex. Would you accept it for the foot encoding?

Hmmh, I still would like a more generic approach if we know outer/inner city. This smells too heuristical. Hmmh, we should think about a more stable approach somehow.

neiljp:
When it comes down to it, for foot at least, I wouldn't mind walking whatever was the shortest (fastest - maybe taking hills into account) route, as long as there is a (safe) route for me to use. Some roads do not permit (regular) foot passengers (eg. UK motorways, I believe?) and many between towns/cities do not have sidewalks.
My feeling is that either a (dynamic) route preference is necessary (walk on primary/sidewalk=no at all? prefer minor over major roads even if longer?), or perhaps some configurable thresholds for how much (amount? %?) extra distance to allow to add to a route if going an indirect way to trade against a more pleasant walk - eg. this footpath takes an extra mile but the route doesn't require walking on a 'major' road with no sidewalk for some distance.

karussell:
Well, 'safe' route sounds easy. But exactly your example is what is makes complex: the lack of data says that the tertiary has no side walks, and inner city this means the primary with sidewalk is saver!

> (dynamic) route preference is necessary

that will come for sure (and is already possible with GraphHopper) but is not the aim of this PR

So maybe we do not prefer the sidewalks that much and additionally remove the tertiary from the avoid list.

karussell:
I've now reintroduced the WORST prio code for sidewalk=no and primary roads, related to #399. This improves the situation. Please review!

Still the northern steps of the roundabout are taken which is okay as it is only a small detour and preferable in general.

karussell:
After fixing a bug (https://github.com/graphhopper/graphhopper/issues/399#issuecomment-129953614) introduced in the previous commit 44b4dc2 the situation is like before.
See the fix:
https://github.com/graphhopper/graphhopper/commit/d7c831b7643b53c77f2a17fc60460664a365af44

But after reviewing this again I don't think that the [3rd route](https://graphhopper.com/maps/?point=57.157053%2C-2.105598&point=57.154446%2C-2.099687&point=57.153445%2C-2.098861&point=57.147323%2C-2.095642&vehicle=foot&locale=en-US&elevation=true) is bad or too long as it is safer and avoids the tertiary road which could be unsafe without a sidewalk. What do you think?

karussell:
So if we want to fix this issue we would need to put tertiary and secondary out of the avoid list, which is only a bit ugly inner city but ironically would fix this issue here.

ratrun:
As indicated above, I think that a 2 step avoidence classification is the best we could reach for now. As expected I think that the result now would be what neiljp intends. If it can be detected in the future if an edge is located inner residential, the code can be improved again. Please take a look at caf3bfc and give it a try.

neiljp:
Should updating with a git pull from the repo as obtained from the devel quick start get the latest fixes being discussed here?
(I tried, and it seems to go via west north street still)

karussell:
We can surely workaround to make this situation working. The question is whether:
 * it is really the better route without sidewalk=yes for the tertiary
 * if it also improves the result for other situations as well - i.e. if there are other use cases
 * if we can get a similar behaviour without many if's - maybe instead we should create a more fluent prioritization for roads? like different penalty even for primary (10), secondary (9), tertiary (8), etc

karussell:
@neiljp as indicated: it will still go through this west north street with the new change as from outside both streets are nearly identical (tertiary without sidewalk preference, primary with sidewalk=left)

Besides this you need to do `./graphhopper.sh clean` additionally to git pull to force recreating the jars.

karussell:
To make my intend clear: I would like to avoid tuning towards single examples. Instead we should use examples to tune towards 'obvious intended behaviour' and make this happen with the minimal code and some unit tests. E.g. in the original request sidewalk=no was included in the response which is not only ugly but could be dangerous and this is fixed now.

But tuning sidewalk=yes & primary versus no sidewalk preferences & lower streets is a bit subjective at least with my current knowledge.

Let me know what you think.

ratrun:
I think that the example only helped to find a better solution in general. I would not have re-activated the discussion otherwise. I completely agree that the preferences are quite subjective. It also depends on the mapping behaviour you are used to. E.g. as far as i know in Vienna every highway must have a sidewalk per law. Therefore the present sidewalks are not explecitly mapped there. I also sometimes wonder how many higher classified roads one can find in some cities, especially in Africa. So I think that it is impossible to find a solution which fits everywhere. The "heuristic" is somhow built into the the conecpt of OSM because of its open and global structure.

neiljp:
Thanks for the build tips @karussell. I can confirm it now routes via west north st without the strange detours it did in the intermediate attempt, including going via the sidewalk on the roundabout rather than the steps (though this exposes some possible incorrect tagging of mine). Strangely, adjusting the route to go on the south side of the roundabout still seems to favour the longer off-roundabout route, but then I'm not sure exactly how the other north route was simplified either, to avoid the steps.

@karussell, I take your point that optimising for a given example is bad, as it can lead to reduced performance in other cases. Hopefully this is a good test case :)

Re safer ("saver"), GH is now routing that way since West North St has a sidewalk tag, but Gallowgate does not? I wouldn't originally have expected that kind of route to be prioritised (having sidewalk vs being shorter). I understand that this uses known information to find a 'safe' route, ie. that Gallowgate might not have a sidewalk and so it might not be shorter or reasonable. However, from the perspective of a mapper, I think it is important that if these kinds of assumptions are made, that this (and perhaps other aspects) is fed back into the OSM community - to ensure that both routers (including GH) don't get a bad press in this kind of strange situation, and that mappers know what is necessary in order to produce good routes. Currently, my understanding there is some fuzzy concept of a 'default' for a road type, and only using the sidewalk tag to indicate deviations from that - and that default varies per road type, country, locality... as per @ratrun 's comment.

neiljp:
I just had a look at the map near the area I showed, on the premise that I could add sidewalk tags to the tertiary/secondary roads too, in order to obtain more accurate routing. However, there are a significant number of these, and I question whether this is the correct way to go about 'solving' this problem.
In some other area, If I wished to indicate (tag) that a roundabout or dual-carriageway had only sidewalks on one/no side, my feeling is that GH will now preferentially route via them compared to nearby roads or shorter routes, simply because a sidewalk tag is present - am I wrong? Needing to say 'there are sidewalks on these slightly-major roads too' everywhere seems excessive.

My original issue was wrt avoiding sidewalk=no; would it be useful to focus on just that?

karussell:
Hmmh, yes. Maybe we should not prefer routes with safewalk=yes versus without safewalk tag - for simplicity we should make the assumption that all streets have a sidewalk. And just put a penalty on roads with sidewalk=no. If this does not solve the problem I'll try to remove secondary&tertiary from avoidhighways or use the solution from @ratrun

karussell:
Handling no tagging and safewalk=yes with the same priority fixed the detour. It now goes over gallowgate

neiljp:
@karussell Agreed this (particular case, at least) looks good here. The issue of sidewalk defaults, preference and major/minor roads is for another issue, I think.

I am interested regarding applications of sidewalk=no in general, but I will address that in a separate issue.

karussell:
> I am interested regarding applications of sidewalk=no in general, but I will address that in a separate issue

What do you mean? Isn't this fixed now?

neiljp:
@karussell I was going to put it elsewhere, but since you asked I'll clarify it here, and it can be copied elsewhere if it's relevant. Clearly some roads simply have no sidewalks for sections, and that should suggest to a router to take an alternative foot route if possible. However I'm intrigued as to whether things like 'linking' road junctions might automatically be avoided through pre-calculation, or whether explicit sidewalk=no tagging will always be necessary; this has significant implications for tagging- vs routing-level treatment of these situations. I'm thinking so far of 'link roads' (eg. which cross a dual-carriageway type pair of roads) and the roundabout issue already seen. The former is like the extension of Canal Place NE of the roundabout in my test-case; it's a 'link' to enable routing, it's purely a road feature and naturally has no sidewalk - and these kinds of roads are quite common at junctions, I think? TLDR: In your opinion, should all such links be explicitly tagged with sidewalk=no for foot routing purposes (potentially a lot of tagging to do), or maybe there is is some kind of way to detect 'pure junctions' or parts of roundabouts, and determine where there is no sidewalk automatically, before calculating routes.

karussell:
There might be a way to analyze crossings and enhance these roads with sidewalk tags but for now GraphHopper would need the tags (still unsure if I should suggest tagging these)

ratrun:
@karussell: Congratulations to the final solution. I wouldn't have expected that you intended to included this into 0.5, because the other changes are much more impressive compared to that peanut.
It was really an interesting discussion with a good result.

ratrun:
@neiljp: Just on more comment on
"However, from the perspective of a mapper, I think it is important that if these kinds of assumptions are made, that this (and perhaps other aspects) is fed back into the OSM community - to ensure that both routers (including GH) don't get a bad press in this kind of strange situation, and that mappers know what is necessary in order to produce good routes."
I think that the rule "We don't map for the render" should be extended to "We also don't map for the router".

neiljp:
@ratrun: As per what @karussell said, I agree about renderer vs router and I try to stay aware of that, but equally it is a true record of those sections of highway, that they have no sidewalk. On that basis, I don't see it as a bad thing - feel free to argue otherwise, here or elsewhere :)
My comment was basically a 'is this possible? if so, we don't need to tag!' question :) There are some cases in OSM where tagging has become complex, or has been proposed to be so, where instead some preprocessing enables a renderer to understand the existing data.

ratrun:
@neiljp: Sorry. I feel that I can't follow your complex English. This medium is awful.

neiljp:
@ratrun, sorry, I will rephrase:
- I agree that rendering for the router is bad, but I don't believe that this case fits that, as these sections of road truly do not have sidewalks;
- The last part was if it was possible to pre-process the data to 'add' sidewalk tags (or make some highways undesirable), based on analysing junction structures. Some renderers perform similar pre-processing, thereby avoiding directly retagging lots of data.

karussell:
@neiljp if that additional tags help mappers and if it is automatable, then we should definitely automate this and similar processes. But currently this won't happen soon as there are many more crucial things to do (if no sponsering happens ;)).

ratrun:
@neiljp thank you. Putting the `sidewalk=no` tag to a highway which does not have a sidewalk is not "mapping for the renderer", because then you just map what anybody can verify in a survey.

Now I understand the second part. This is what was Peter has implemented.

PR #480, opened 2015-07-29
make shorter CH preparation public available

karussell:
If you want to save CH preparation time and can accept a lot slower queries (still faster than without CH) you can now specify e.g. `prepare.contracted-nodes=97` which then only does the preparation for 97% of the nodes. In my experimental setup this leads to a 20% faster preparation but only a CH boost of 4. Same for 99%: you get 8% faster preparation but only a speed up factor of 9

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -38,7 +38,6 @@ import java.io.File;
 import java.io.IOException;
 import java.text.SimpleDateFormat;
 import java.util.*;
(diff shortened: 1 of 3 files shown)


devemux86:
Peter I think you mean here that with the less nodes, there is X% faster preparation and CH **slow down** of X factor.

karussell:
Indeed there was something missing. I meant the original CH boost factor was 47 and therefor if e.g. only 99% of the preparation is done the boost is only 9 i.e. 5 times slower. But for bigger areas this could make sense, only tested it for ~200km²

PR #487, opened 2015-08-03
Remove subnetworks for multiple vehicles

### LINKED ISSUE #86: Barriers lead to disconnected areas
Currently we completely remove small subnetworks in order to avoid failures when routing (as two disconnected areas won't have a path in between). Those subnetworks exists due to OSM errors or other things (?)

But the barriers can introduce further subnetworks ([example](http://www.openstreetmap.org/edit?lat=51.0757&lon=13.7388&zoom=18)). The problem is that only a subset of the allowed vehicles could be affected by those barriers. E.g. `barrier=bollard` limits cars but not pedestrians. Which means only for cars the disconnected subnetwork exists.

 - [ ] In order to fix that one needs to identify all subnetworks (per vehicle) and restrict access (setting both direction flags to 0). Now, if all allowed vehicles have the same restriction for one edge one can safely remove this edge.
 - [ ] Make this process working for the one way network removal (tarjan algorithm) too
 - [x] But still there will be problems related to this as in some cases an area is only reachable to "go out" or "go in". E.g. [this query](http://graphhopper.com/maps/?point=55.862711%2C-4.432372&point=55.864237%2C-4.251806) or [here](http://graphhopper.com/maps/?point=krakow&point=berlin). This will be fixed via #226

karussell:
This issue finally fixes #86. Some performance and manual tests are currently done...

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
+++ b/core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
@@ -41,7 +41,7 @@ public class DefaultEdgeFilter implements EdgeFilter
         this.encoder = encoder;
         this.in = in;
         this.out = out;
-    }
+    }    
 
     @Override
     public final boolean accept( EdgeIteratorState iter )
--- a/core/src/main/java/com/graphhopper/routing/util/PrepareRoutingSubnetworks.java
+++ b/core/src/main/java/com/graphhopper/routing/util/PrepareRoutingSubnetworks.java
@@ -21,9 +21,9 @@ import com.graphhopper.coll.GHBitSet;
 import com.graphhopper.coll.GHBitSetImpl;
 import com.graphhopper.storage.GraphHopperStorage;
 import com.graphhopper.util.*;
+import gnu.trove.list.TIntList;
 
 import java.util.*;
-import java.util.Map.Entry;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.slf4j.Logger;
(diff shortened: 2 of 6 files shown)


core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
```diff
@@ -41,7 +41,7 @@ public DefaultEdgeFilter( FlagEncoder encoder, boolean in, boolean out )
         this.encoder = encoder;
         this.in = in;
         this.out = out;
-    }
+    }
```
zhiqiangZHAO: typo?

core/src/main/java/com/graphhopper/routing/util/PrepareRoutingSubnetworks.java
```diff
@@ -73,45 +68,52 @@ public PrepareRoutingSubnetworks setMinOneWayNetworkSize( int minOnewayNetworkSi
...
         {
-            unvisitedDeadEnds = removeDeadEndUnvisitedNetworks(singleEncoder);
+            // mark edges for one vehicle as inaccessible
+            PrepEdgeFilter filter = new PrepEdgeFilter(encoder);
+            if (minOneWayNetworkSize > 0)
+                unvisitedDeadEnds += removeDeadEndUnvisitedNetworks(filter);
+
+            List<TIntArrayList> components = findSubnetworks(filter);
+            keepLargeNetworks(filter, components);
+            subnetworks = Math.max(components.size(), subnetworks);
```
zhiqiangZHAO: The idea here is: if this is a subnetwork for any vehicle, it will be removed ?

core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
karussell: Will remove the white space

core/src/main/java/com/graphhopper/routing/util/PrepareRoutingSubnetworks.java
karussell: The idea is to make the subnetwork calculation separate for every vehicle via the access restrictions instead of node removal.

Still the node removal occurs later on (in markNodesRemovedIfUnreachable) if no vehicle has access to the adjacent edges.

zhiqiangZHAO: I see. thank you Peter.

PR #488, opened 2015-08-03
Fixing potential location index lookup issue when filter is provided

karussell:
This fixes #318 

The iteration loop was moved out of findNetworkEntries to recall this method if no matching edges are found (i.e. edges that matches the edgeFilter)


karussell:
Merged

PR #502, opened 2015-08-23
Elevation control didn't show up

ratrun:
The elevation control in the web GUI does not show up. 8009563b0f7638459006fe6c306dcb579ad5203c was the reason. In `GHRequest.prototype.init = function (params)` parameter  `this.vehicle` was still used, although the assignment was removed. So this should be changed into 'params.vehicle'.

The same is true below in 'GHRequest.prototype.initVehicle = function (vehicle) {` where 'this.vehicle' needs to be replaced with 'vehicle'.


karussell:
Thanks!

PR #507, opened 2015-08-25
Make way-geometry store customizable in subclasses

jtheuer:


karussell:
Maybe the better way of doing it would be to define a more generic PointList modification interface? Which is currently DouglasPeucker and could be used for smoothing and more.

jtheuer:
Sure, if you prefer a more generic solution.

karussell:
With extending DouglasPeucker it should be already doable what you want, although not soo nice as with a more generic PointListModificator interface

jtheuer:
Looking deeper at the problem it would also be nice to have the full geometry, not only the pillarnodes.

Of course you can't change the tower coordinates but the for simplification and other enhancements it is always better to have the first and last coordinate, too.

karussell:
Yes, indeed.

jtheuer:
Another possibility that is less invasive would be to change the modifier of
```
EdgeIteratorState addEdge( int fromIndex, int toIndex, PointList pointList, long flags, long wayOsmId )
```
to ```protected``` this would follow the pattern you use for ```getElevation``` and would also allow me to perform the necessary modifications of the PointList. What do you think?

karussell:
Hmmh, what about the following?

If we move
```java
if (doSimplify) simplifyAlgo.simplify(pillarNodes);
```

into a method - e.g. `protected PointList filterNodes(PointList)` - and call this method with the full nodes (`pointList`) before the pillarNodes creation loop, then the simplify algorithm would improve too and your requirements are full filled.

jtheuer:
Sounds great!

karussell:
I've implemented a sketch for this, but several integration tests are failing and I'm unsure why the algorithm (visited nodes mismatch in GraphHopperIT.executeCHFootRoute) is effected, but won't have time to investigate in the next days.

Issue #501, opened 2015-08-22
Ferry issue for foot routing


karussell:
Reported [here](https://lists.openstreetmap.org/pipermail/graphhopper/2015-August/002323.html) and works for biking, so it should work for foot too.

That it does not work for car is okayish as ferries are not preferred due to costs and uncertain duration.

ratrun:
The reason for this finding is a wrong implementation of the duration tag. The format hh:mm:ss is currently interpreted as dd:hh:mm . See http://wiki.openstreetmap.org/wiki/Key:duration .

karussell:
Thanks for this finding! Maybe this should be also documented in the OSM wiki?

ratrun:
The wiki already documuments it since 16:18, 29 January 2013. I seems that a `dd:hh:mm` format was never documented. To be precise: the wiki documents a 'hh::mm::ss' format (with double colons). The double colons must be a typo, because the 25 real-world examples I found with overpass query   `way["duration"~"[0-9][0-9]:[0-9][0-9]:[0-9][0-9]"]({{bbox}}) ` in France but no match at all for `way["duration"~"[0-9][0-9]::[0-9][0-9]::[0-9][0-9]"]({{bbox}}) ` world-wide.

karussell:
Do you have a link for future reference?

ratrun:
Probably you meant this link: https://graphhopper.com/maps/?point=47.19415%2C-1.676384&point=47.19747%2C-1.671832&locale=fr-FR&layer=Lyrk

karussell:
Sorry, I meant the OSM docs link but now I can see the 'hh:mm:ss' :)
I thought before you meant 'PT02:25:06' which is not yet supported as it is very inconsistently used in OSM data.

PR #514, opened 2015-09-12
Fix parsing of duration tag for hh:mm:ss format.

ratrun:
According to http://wiki.openstreetmap.org/wiki/Key:duration there is no dd:hh:mm format, instead this format needs to be interpreted as hh:mm:ss

The modification also fixes issue #501

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
@@ -462,13 +462,14 @@ public abstract class AbstractFlagEncoder implements FlagEncoder, TurnCostEncode
     }
 
     /**
-     * This method parses a string ala "00:00" (hours and minutes) or "0:00:00" (days, hours and
-     * minutes).
+     * This method parses a string ala 'hh:mm', format for hours and minutes 'mm', 'hh:mm' or 'hh:mm:ss'
+     * FIXME: Add support for ISO_8601 
      * <p>
      * @return duration value in minutes
      */
     protected static int parseDuration( String str )
     {
+        int minutes = 0;
         if (str == null)
             return 0;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -462,13 +462,14 @@ protected double parseSpeed( String str )
     }
 
     /**
-     * This method parses a string ala "00:00" (hours and minutes) or "0:00:00" (days, hours and
-     * minutes).
+     * This method parses a string ala 'hh:mm', format for hours and minutes 'mm', 'hh:mm' or 'hh:mm:ss'
+     * FIXME: Add support for ISO_8601
```
karussell: The problem is not that it can't be added. The problem are several wrongly mapped duration. I think that it was like mixing minute stuff with month, although it is stated in the wiki: *duration=PT71M - A duration of 71 minutes. This is not the same as P71M which denotes 71 months.*

ratrun: In the meantime I have added support for ISO_8601 and intended to add it as separte pull request. I don't think that it is a good idea to just say that it is often filled out in a wrong way and even document in a comment a example of way with a wrong usage. First I would fix this way.
Shouldn't it be possible to add a simple plausability check based on the distance, when it comes to the decision if to consider the value of the duration tag? In case that the calulated speed is less then some value e.g. 0.1 km/h we could ignore it and use the calculated duration based on an the assumed ferry speed instead.

karussell: 100% agree - nice idea! Thanks & looking forward to the PR!

karussell: Maybe we can log these mismatches and create a blog post with these entries so that we or mappers could fix them

ratrun: The idea with the logging and the blogpost is very good. Maybe somebody might be willing to turn the result into a "Wochenaufgabe".

PR #516, opened 2015-09-14
Adding of comments for artificial way tag "estimated_distance"

ratrun:
During adding of support for duration in ISO 8601 format I discovered that the calculation of the artificial estimated_distance way tag is unnecessary most of the time. Now we compute it only in case that a way contains a "route" tag. This change should decrease the graph preparation time a little bit.
Additionally some comments about the estimated_distance handling are added or fixed.

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/OSMElement.java
+++ b/core/src/main/java/com/graphhopper/reader/OSMElement.java
@@ -127,7 +127,7 @@ public abstract class OSMElement
     }
 
     /**
-     * Chaeck that the object has a given tag with a given value.
+     * Check that the object has a given tag with a given value.
      */
     public boolean hasTag( String key, Object value )
     {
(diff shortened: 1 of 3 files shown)


karussell:
Thanks!

Still I would like to avoid breaking the contract that such stuff is available in the encoder. Also the computational overhead should be neglectable.

I cannot remind it anymore but I needed the distance estimation not only for ferry stuff in another project.

karussell:
The other changes look good. Why was there a test change necessary?

core/src/test/java/com/graphhopper/reader/OSMReaderTest.java
```diff
@@ -648,7 +648,7 @@ public long handleNodeTags( OSMNode node )
         OSMWay way = new OSMWay(1L);
         way.getNodes().add(1);
         way.getNodes().add(2);
-        way.setTag("highway", "motorway");
+        way.setTag("route", "ferry");
```
karussell: I mean: why is this change necessary? Maybe we should add a new test specific for route=ferry? (and avoid removing the test for highway=motorway?)

ratrun: The test is testing the calculated estimated_center. This artificially created tag is not used anywhere in the productive code, just in tests.

ratrun:
Now the PR contains only comment and file format changes, but no code change.

karussell:
Thanks!

Issue #322, opened 2015-02-10
Incorrect snapping for ferry routes. LocationIndex should know only start and end


DanySK:
Hi all,
I rely on Graphhopper for simulating opportunistic networks of mobile devices in real-world setups. The project is called Alchemist and it is supported by University of Bologna. The high number of randomly displaced devices allowed me to spot some curious behaviour, that I was able to reproduce in the web version of Graphhopper: apparently, navigable lines are interpreted as legit foot path.

To reproduce, please visit this link:
https://graphhopper.com/maps/?point=51.502091%2C-0.123076&point=51.500435%2C-0.12225&vehicle=foot&elevation=true&layer=Lyrk

I guess that such behaviour is anomalous. I make widespread use of LocationIndex.findClosest() in order to be able to map possibly wrong devices position to plausible coordinates, but they often end up suggesting... swimming. This is one other example: https://graphhopper.com/maps/?point=51.500341%2C-0.122609&point=51.500408%2C-0.119519&vehicle=foot&elevation=true&layer=Lyrk

karussell:
That is a ferry, not really swimming and not anomalous. In order to map sequential data points to edge ids you might be interested in our new [map matching component](https://github.com/graphhopper/map-matching)

But I see what you mean and maybe the route should not start directly on a ferry. Not sure.

karussell:
See the corresponding [OSM layer](https://graphhopper.com/maps/?point=51.502091%2C-0.123076&point=51.500435%2C-0.12225&vehicle=foot&elevation=true&layer=OpenStreetMap) for a more detailed look of the ferry routes.

DanySK:
Than you for your answer, I understand.

Unfortunately the (nicely powerful) map matching is not a viable work around in my case: I often need, for instance, to displace hypothetical network nodes on a real-world map, and in order to do that I use the method mentioned above. As you surely realise, this is disruptive if the simulated area comprises navigable water streams. I am currently working this around using bike paths, but is clearly sub optimal.

Anyway, apart from my very problems, which I believe should not be considered since graphhopper aims at a general navigator solution, it appears to me that:
* I agree with you that it makes little sense starting a route from a ferry. One problem here is that if you try to start to navigate from a bridge, even a little drift in GPS precision may lead to the suggestion of taking a ferry. I noted also that if ferries and bridges intersect, on intersection points some times a ferry is suggested as closer point.
* I also agree that ferries should not be mixed. More in general, they should not be considered as plain roads.
[...]

PR #519, opened 2015-09-16
Adding support for parsing duration in ISO_8601 format

ratrun:
This change is the result from a discussion of PR #514: Now we are supporting the ISO 8601 format of the duration tag. In case that errors of the tag values are detected, the respective OSMWay ID and the duration tag value gets logged.


karussell:
Thanks!

Why are there failing tests?

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -26,6 +26,7 @@
 import com.graphhopper.util.*;
 
 import java.util.*;
+import javax.xml.datatype.*;
```
karussell: Hmmh, we cannot rely on this xml stuff in the encoder, this won't work for Android and iOS. Isn't it possible with something else or by hand where we support just the majority of possibilities?

ratrun:
No. I don't intend to re-invent the wheel. I found an alternative on Java 8: https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html, but I can not find an alternative for Java 7.

karussell:
Ok, but we really need to support Android. The rules for the core are very strict and even 1.6

Also a simple solution via SimpleDateFormat could be possible without much effort but I would understand if that is too much effort

devemux86:
>Ok, but we really need to support Android.

:+1:

ratrun:
I'm wondering if anybody is really creating graphs on Android. Does this make sense?

devemux86:
>I'm wondering if anybody is really creating graphs on Android. Does this make sense?

No, for creating graphs.
But should be able to built against Java 7 (or 6 as Peter mentioned) and don't have linked classes causing issues in mobile environments (e.g. Android publishing with ProGuard will fail on that).

At least until we have a more modular GH. :)

ratrun:
It looks as we could use joda, see http://www.joda.org/joda-time/. Its license is Apache license 2.0. I could try to Switch to using it.
Would this be accepted?
If yes, what is the best way to include it? As jar release file? In source?

devemux86:
May I ask why we need an external library (or advanced class).
Current Java cannot suffice us to implement the functionality?

ratrun:
Sure one could implement the ISO 8601 stuff, but as indicated I won't. This obstacle is too high for the benefit. It is really stupid that there is not even an #ifdef in Java.

karussell:
I have another idea: why not move the iso calculation out and do this in OSMReader e.g. convert the result into minutes which is then read by the flagencoder.

This way the flag encoders stay independent of xml dependencies.

> I'm wondering if anybody is really creating graphs on Android. Does this make sense?

Android does not use the import stuff etc BUT it is using the FlagEncoders so it needs to know all the classes when loading it I think.

ratrun:
Thank you. This should be easy. I'll do it in the late afternoon.

ratrun:
I moved the parsing now into the reader. Now there are no travis errors any more, but I'm still not sure if I broke Android. I'm refering to the new class within the abstract flag Encoder.
Could you please check?
"P2M" converted into minutes obviously returns a slightly different result under openjdk. Therfore I relaxed this test in order to pass tarvis test.

karussell:
Ah, I meant a bit different not sure if hiding like this helps
@devemux86 what do you think? How is this handled on Android?

devemux86:
Good news, more thorough search reveals that there is ```javax.xml.datatype``` package from Android API 8 (below our min).
https://developer.android.com/reference/javax/xml/datatype/package-summary.html

karussell:
Thanks & nice. And it looks like j2objc has support for this as well: http://j2objc.org/reference/javax/xml/datatype/package-summary.html

Still I would like to avoid this in the encoders and move it to OSMReader which can depend on xml or pbf stuff etc :)

ratrun:
@karussell : Did I understand you right that I should check every way for the presence of the "duration" tag and convert into minutes (or maybe better seconds) there? I suppose that this is rather costly in terms of graph generation time, as 99.98% there won't be such a tag. Is the goal to keep the XML parsing dependecy out of the encoders really worth that?

karussell:
Yes, but why would this be expensive? I doubt this, as this is just a hashmap 'get' :)

ratrun:
Thanks @devemux86 for the Android investigation and thanks @karussell for the rebase/stash git lection. I hope that its now simlar to the way you intended it. I'm not sure about the log level for the paring findings. I set it to 'warn', but eventuelly it should be only 'info'.

karussell:
Thanks, looks good!

Would you mind if I or you reduce the commit message? I prefer to have the commit messages short and to the point (but not too short :)) and refer to the issue or discussion.

> I'm not sure about the log level for the paring findings.

We had already a warning if the parseDuration failed, so this should be good.

ratrun:
Short enough commit message now? I prefer longer commit messages, as from expierence I learned that it usually helps detecting introduced problems later on. But this change is so well reviewed and test covered, so that there won't be any problems :wink:

karussell:
Hmmh, I always use just the issue title, as I think the explanation why and how should go into more detail like only possible in the discussion like here. Will include the longer comment here:
**The changes are the result of issue #519. Now we are supporting the ISO 8601 format of the duration tag. In case that errors of the tag values are detected, the respective OSMWay ID, the duration tag value and its interpretation in minutes gets logged.**

I've now merged the changes - thanks! Would you mind to include a succeeding real OSM test? Like you have done with the failing test-osm2.xml?

ratrun:
This is a real world example: https://graphhopper.com/maps/?point=46.023668%2C9.235931&point=46.011032%2C9.284134&point=45.983007%2C9.259964&locale=en&vehicle=car&weighting=fastest&elevation=false&layer=Omniscale 
From the via to the end-point it currently takes a 49km de-tour. When the change goes on-line also the second 15 minutes ferry route should be taken like this: https://graphhopper.com/maps/?point=46.023668%2C9.235931&point=46.013178%2C9.284134&point=45.983007%2C9.259964&locale=en&vehicle=car&weighting=fastest&elevation=false&layer=Omniscale

karussell:
Ok. [Here](https://gist.github.com/karussell/da39dcbea2b8b9cc9e06) is the result of a world wide run

ratrun:
Thank you. Now I have fixed the findings in the OSM database. At some occurencies similar issues as described under #322 occurred where multiple edges of the same ferry contained the same long duration tag. I have fixed theses in the OSM database by removing the duration from the segments and moving it into the duration tag of the relation.

It turned out that the unrealistic duration detection for very short ferries was too aggressive (and buggy too because of rounding before the check). I have fixed that in the code, see the latest changes in  f4235be

What I have learned so far: 
1) The OSM wiki about the duration tag needs to clarifiy especially for short ferries if the average waiting is to be included in the duration or not. Obviously some mappers include it and some do not. I believe that an average waiting time should be included. If you agree, then I will update the wiki.
2) The logging occurs for every flagencoder seperately. Therefore some IDs get logged twice or even more often. I don't think that we need to fix that short term.
3) It turned out that there is a more general problem for very short ferries in distance, which we probably should address in a new separate issue. Please take a look to the "FIXME" comments. A real world testcase for this is here: https://graphhopper.com/maps/?point=52.019398%2C9.423244&point=52.019015%2C9.426355&locale=de-AT&vehicle=car&weighting=fastest&elevation=false&layer=Omniscale
With the latest change the route is now found for bike, foot and car. Anyhow, the duration in the instruction is wrong for all. I would call the new issue something like this: "Duration shown in the instructions is too short for short ferries, This occurs for ferries, which are tagged with a duration tag, where the value includes an average waiting time".

karussell:
Thanks for the findings! So if multiple ferries share the same ways then they have to be mapped via relations - maybe one should add this to the wiki too.

> turned out that the unrealistic duration detection for very short ferries was too aggressive (and buggy too because of rounding before the check). I have fixed that in the code, see the latest changes in f4235be

Thanks! Would you mind to add a small unit test showing the bug?

> I believe that an average waiting time should be included

Hmmh, I think that this should not be part of the ferry. The duration is really only the duration IMO. One could add another tag for average waiting time. But mixing both feels 'unexpected' to me.

ratrun:
Now I included unit tests (for car) and also slightly improved the code for very slow short ferries. I also removed the FIXMEs, as the duration shown as part of the instructions needs to be handled as part of a separate issue.

Originally the same idea with the two tags for the duration came into my mind. So you would specify the "duration" from when the ship/boat leaves the start location to when it arrives at the end? 
This also was my first interpretation as well, but I must say that I read the code first and only after that looked on to the wiki page. 
In case that it is a ship, it must be faster than at least 1km/h. But this might not be the case for a ferry on ropes on a river. I remeber a real world ferry on ropes example for which I have a GPX trace and therefor was able to look it up. This ferry is incredible slow, the slowest I ever saw. It made 88 meters in about 5 minutes, which is 1.056km/h. For these types of ferries we now use the shortest possible speed.
What do you say to the 10 minutes duration for a 77 meters distance from the example? Is this a tagging error or is the average waiting time included? I believe that here the waiting time is included. 

I plan to update the wiki saying that the waiting time is not included and if the ferry is tagged as a relation, that a duration tag should be set as part of the relation.

karussell:
> So you would specify the "duration" from when the ship/boat leaves the start location to when it arrives at the end?

Yes, but [I've asked](http://forum.openstreetmap.org/viewtopic.php?id=50716) this on the OSM forum

> What do you say to the 10 minutes duration for a 77 meters distance from the example? Is this a 
> tagging error or is the average waiting time included? I believe that here the waiting time is included. 

I had not yet the time to look deep into the issues and I highly appreciate that you did :) ! But yes, this looks like there waiting time included. I would add an OSM note so that someone there can have a look.

> I plan to update the wiki saying that the waiting time is not included and if the ferry is tagged 
> as a relation, that a duration tag should be set as part of the relation.

Let us wait a bit for the osm forum entry ... otherwise go ahead :+1:

karussell:
> Is something else left open?

No, looks good, thanks. Will hopefully test a bit more on Monday and merge then.

karussell:
I've updated [the gist](https://gist.github.com/karussell/da39dcbea2b8b9cc9e06) with the new ww results - now using bike&foot&mtb&racingbike (coming very soon :))

Regarding the change: this is IMO not more descriptive, it is more the minimal resolution or something (+underscore is against Java style guide). So I would leave this for another issue for now.

ratrun:
Ok. I'm not that familiar with Java. Can you do cherry picking or do I need to revert 3fccc38?

karussell:
Yes, cherry pick should work

karussell:
I've merged your changes, see [this commit](https://github.com/graphhopper/graphhopper/commit/f3b37c363d59eab77b24437dc5add8bc6909dc71) - thanks!

Issue #259, opened 2014-09-10
Multilingual Street names in Instructions depending on the 'Locale' of a query


stephanosch:
When asking for routing in Greek language, is it possible to use the Greek name of the roads within the OSM? In general, to allow multiple names depending on the 'locale'

For example in OSM we have:
```xml
<way id='71' visible='true' version='1'>
     …
    <tag k='name' v='Arodafnis' />
    <tag k='name:el' v='Αροδαφνης' />
  </way>
```

devemux86:
Peter how do you think we should handle this?

In [Mapsforge](https://github.com/mapsforge/mapsforge/blob/master/docs/Getting-Started-Map-Writer.md#basic-options) we have the parameter ```preferred-language``` which defines the language to use primarily.

Should we add a similar parameter here too?

devemux86:
I mean for start, during graph creation to let user select a language to use.

At a later stage we could think about multilingual graphs.

karussell:
> I mean for start, during graph creation to let user select a language to use.

Yes, this should be very easy to do

> At a later stage we could think about multilingual graphs.

Should be also doable when we introduce a new `getName(locale)` method but the storage will be a bit more complex. Maybe we let the user decide which languages to parse up front and then we create for every language a reference array as well as the nameindex, this would avoid that there are too many languages with only a few used values (e.g. when creating a new reference array for any new encountered language). We will see and should think about it.

karussell:
The preferred language is now changable - thanks @devemux86 - will keep it still open until multiple street names can be stored&retrieved.

devemux86:
@karussell 
In a next more advanced step, we could use a language regex pattern (see [here](https://github.com/mapsforge/mapsforge/blob/dev/mapsforge-map-writer/src/main/java/org/mapsforge/map/writer/util/OSMUtils.java#L46)), for matching the osm name.
That could work if we iterate though ```OSMElement``` tags (currently ```protected```).

karussell:
Not sure, what you mean. When we support multiple languages for import and for retrieval we will probably never be able to support an arbitrary number of languages.

devemux86:
Right now we try to retrieve the preferred language exactly as proposed by the user.
We could extend the findings by ignoring case, parsing osm name tags with ```-``` or ```_```, etc.
e.g. en, en-US, en_US, EN_US could be parsed in the same name.

karussell:
[...]

PR #522, opened 2015-09-19
Multilingual graphs: preferred language on import, #259

devemux86:
First step for multilingual graphs #259.

Specify the ```preferred-language``` for way names during graph import.
I followed our methodology in Mapsforge [map-writer](https://github.com/mapsforge/mapsforge/blob/master/docs/Getting-Started-Map-Writer.md#basic-options).
Language code as defined in ISO 639-1 or ISO 639-2.

* If no preferred language is specified, only the default language with no tag will be imported.
* If a language is specified, it will be imported if its tag is found, otherwise fall back to default language.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -31,6 +31,7 @@ import com.graphhopper.storage.index.LocationIndexTree;
 import com.graphhopper.storage.index.QueryResult;
 import com.graphhopper.util.*;
 import com.graphhopper.util.shapes.GHPoint;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -65,6 +66,7 @@ public class GraphHopper implements GraphHopperAPI
     private final String fileLockName = "gh.lock";
     private boolean allowWrites = true;
     boolean enableInstructions = true;
+    private String preferredLanguage = "";
     private boolean fullyLoaded = false;
     // for routing
     private double defaultWeightLimit = Double.MAX_VALUE;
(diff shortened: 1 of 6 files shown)


karussell:
Thanks! Can you add a small test for this? Maybe add two different tags to one of the osm.xml files

devemux86:
Yes certainly I'll try that.

devemux86:
Peter please check if this is right implemented.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -607,6 +630,7 @@ public GraphHopper init( CmdArgs args )
 
         workerThreads = args.getInt("osmreader.workerThreads", workerThreads);
         enableInstructions = args.getBool("osmreader.instructions", enableInstructions);
+        preferredLanguage = args.get("osmreader.preferred-language", null);
```
karussell: Here it should be `preferredLanguage = args.get("osmreader.preferred-language", preferredLanguage);` to avoid overwriting possible initialized values

karussell:
Looks, good! (only the tiny hint)

devemux86:
I fixed the hints.
Though I think we still need the ```Helper.isEmpty(preferredLanguage)``` because in case of empty string, the default language should be used.
(Travis seems on hold)

karussell:
Yes, that is necessary but I would do "string.isEmpty" instead, making it more clear for others. And for that we have to make sure string won't be null

devemux86:
@karussell I really approve your persistence to detail! :+1:

karussell:
not really ;) 
as I'm not always that consistent ... as e.g. this Helper.isEmpty is used and none-null is not always enforced (but not always can be enforced easily)

karussell:
Thanks - merged: https://github.com/graphhopper/graphhopper/commit/d782f9da9c7283640aed3eee236a4ad927cd4c33

PR #524, opened 2015-09-23
Curvature weighting

boldtrn:
Currently there is an Exception thrown when calculating the distance between two edges.


boldtrn:
Ups, hier hab ich mich leider verklickt.

Issue #510, opened 2015-09-02
Put numbers on markers in web UI


karussell:
Similar to what we have in our [route optimization editor](https://graphhopper.com/blog/2015/07/21/graphhoppers-new-route-optimization-editor/)

![route editor](https://graphhopper.com/blog/wp-content/uploads/2015/07/route-editor-overview.png)

PR #525, opened 2015-09-24
Put numbers on markers in web UI

ratrun:
Implementation for issue #510. I didn't expect this to be that easy. So far the markers for the instructions still contain the icon without a number. There it is a more tricky because I have no idea where to get the index from, therefore I didn't attempt it. I believe that this eventually would require changes at the server side and could be added in a later step. 
The icons looks good also for two digit numbers. They probably do no look that nice for 3 digit numbers, but this probably will never happen - and if so, I don't care.


karussell:
:+1: ! Also thanks for the typo corrections :)

karussell:
[Deployed !](https://graphhopper.com/maps/?point=Leipzig%2C%20Deutschland&point=Hof%2C%20Deutschland&point=M%C3%BCnchen%2C%20Deutschland&point=Stuttgart%2C%20Deutschland&locale=de&vehicle=car&weighting=fastest&elevation=false&layer=Omniscale)

ratrun:
Whow, this was really fast!

Issue #462, opened 2015-07-17
Make age of data available in the UI


karussell:
Currently the age is included in the info json but not visible to the user. As several profiles can have different dates, we need to extend the json in this direction and e.g. print the oldest date?

zhiqiangZHAO:
what you mean "age"? the date of osm snapshot?

karussell:
The date of the import (could be indirectly the date of the OSM) and/or the date of the CH preparation. See [here](https://github.com/graphhopper/graphhopper/blob/master/web/src/main/java/com/graphhopper/http/InfoServlet.java#L73)

PR #530, opened 2015-09-27
Make age of data available in the UI

ratrun:
Implementation of issue #462.

### DIFF
--- a/web/src/main/webapp/js/main.js
+++ b/web/src/main/webapp/js/main.js
@@ -31,6 +31,7 @@ var routeSegmentPopup = null;
 var elevationControl = null;
 var activeLayer = '';
 var i18nIsInitialized;
+var metaVersionInfo;
 
 var iconFrom = L.icon({
     iconUrl: './img/marker-icon-green.png',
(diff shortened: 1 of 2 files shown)


web/src/main/webapp/js/main.js
```diff
@@ -1070,15 +1080,19 @@ function routeLatLng(request, doQuery) {
 
             var hiddenDiv = $("<div id='routeDetails'/>");
             hiddenDiv.hide();
+            var versionInfoHiddenDiv = $("<div id='versionDetails'/>");
+            versionInfoHiddenDiv.hide();
```
karussell: Add the version details to the hiddenDiv, then no extra hide() and toggle() call is necessary

karussell:
Thanks! I will have to make a few changes before this also works on GraphHopper Maps though ... but should not take too long.

ratrun:
I have finished an extension providing support for the OSM data date via the `osmosis_replication_timestam` field according to   https://wiki.openstreetmap.org/wiki/PBF_Format#Definition_of_the_OSMHeader_fileblock.

These extensions are here: https://github.com/ratrun/graphhopper/tree/OSMfileHeaderDate. It is probably better to create a separate PR, but if you prefer I can add it to here such that you need to merge only once.

karussell:
Merged this [here](https://github.com/graphhopper/graphhopper/commit/f668788fc750d06f49556cb63caad57ac78ca72b). Please create a new PR for the other nice work!
(The work looks really good, thanks! I would directly return the osmDataDate and reuse the `formatDateTime` method for formatting)

karussell:
This is now live. Currently we use 2 JVMs but there could be more. So I introduced a semi-official json format for the info page ala:
`"car":{"import_date": ...}, "motorcycle":{...}`

and I simply pick the car profile which could be different to others:

`https://graphhopper.com/api/1/info?debug=true&key=[YOUR_KEY]`

Issue #498, opened 2015-08-18
Apache HTTP Client Removal


devemux86:
Android 6.0 SDK (API 23) was released and removes some deprecated classes / methods.

In our case it's about the removal of ```Apache HTTP Client``` used in ```AndroidDownloader```.
http://developer.android.com/preview/behavior-changes.html#behavior-apache-http-client

The recommended way is use of ```HttpURLConnection``` as seen in ```Downloader```.

karussell:
The problem is how can we remove it when it causes several issues without it? Should we use something like OkHttp instead for Android? To my knowledge it is based on HttpURLConnection but corrects some limitations regarding certs etc.

devemux86:
The best way would be handling the situation with the recommended Android way (there must be a solution somehow - though I have not managed investing time in it).

https://developer.android.com/about/versions/marshmallow/android-6.0-changes.html#behavior-apache-http-client
We can continue with Apache HTTP APIs by declaring the folowing:
```gradle
android {
    useLibrary 'org.apache.http.legacy'
}
```

karussell:
As it is only a demo I would prefer a way which won't make problems in a year. So I would prefer to not go the legacy route and use either the Android way or the external okhttp, where I made good experiences with in our [Java/Android clients](https://github.com/graphhopper/directions-api-java-client)

devemux86:
Usually avoiding to load a library/app with too many external dependencies, so I'd first review the official recommendation.
But it's your call at the end. :smile:

karussell:
Well, I do not care that much, as this is a demo and people can choose a different download mechanism. But yes, if there would be a stable and reusable download component without deps this would be best :)

devemux86:
>Also if people use http to download there should not be any problems

So why the ```https``` in Sample if it can solve the situation?

karussell:
https is enforced for everything on GraphHopper.com

I could make an exception for the maps, but it could make problems with the enabled [hsts](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security). Not sure if this applied only to browsers (I don't think so), so if the downloaded fetches once a https site it will never fetch a http site again.

devemux86:
Well not making promises, I can review the ```https``` issue, if another solution doesn't appear in the meantime.

karussell:
[...]

PR #533, opened 2015-10-04
Android Gradle advance, improvements

devemux86:


karussell:
Thanks!

devemux86:
I kept Android compile version to 22 (Lollipop) as we need first to see #498.

devemux86:
BTW Peter why you have commented the ```jcenter```?
I know Google in recent AS builds puts only that and no ```maven``` at all.

karussell:
> BTW Peter why you have commented the jcenter?

I do not know but all repositories should be identical, and I prefer to have consistency across compilation, but if it makes problems then include it again.

Issue #425, opened 2015-05-27
Improve motorcycle profile


karussell:
Include the [bendiness of a road](http://www.technomancy.org/openstreetmap/bendy-roads/) and prefer higher values in the weighting, make it possible to change the influence on the fly (so do not pre-calculate priority)

boldtrn:
Hi Peter @karussell ,

we are currently working on that feature. Right now, we did a quite easy implementation. Created a weighting that knows the curvy roads and returns 0, when weighting a curvy road (we basically have a list of curvy roads).

The biggest issue is, that even when using the Motorcycle flagEncoder, routes that could be done on a highway are routed over that highway. 

Since you understand all the internals, how would you tackle the issue?

I think a good solution would be to add a curvature parameter (the higher the more curvy the road is) to each edge in a preprocessing step (when loading the .pbf). There is a nice python library that does something similar and exports .kml files (https://github.com/adamfranco/curvature).

Afterwards, a weighting like: ```distance / speed*curvature```could be feasible. However, I am still not sure about the highway thing. When riding the motorcycle one wants to avoid highways, as long as there is at least a more or less acceptable alternative.

Best,
Robin

karussell:
Yes, you'll probably need both: mark as highway and increase weighting for that, and mark the 'bendiness' of each edge. Both done for the import so that it is accessible somehow in the weighting (either in the flags or in a separate DataAccess object or similar).

boldtrn:
Hi Peter @karussell ,

thanks for your comment. 

Since, I lack some understanding for the internals of graphhopper could you please explain me these points:
- How to check for the road type in a weighting implementation?
- How do the flags work and how would I create a new flag?
  - I've read your post about the internals. There you describe that each edge consists of parameters. But somehow I am not able to find something that explains how to work with these properties and flags. I've seen the bike2 encoder. You basically add an ```EncodedDoubleValue```. Is this how to add new Flags?
- Where to apply the curvature calculation?
  - I've seen the method `processWay` in the `OSMReader`, is that the correct location?
  - Alternatively is it the `preProcess` in the `OSMReader`

Cheers,
Robin

karussell:
> How to check for the road type in a weighting implementation?

[...]

Issue #493, opened 2015-08-10
Make creating CH profile more configurable


karussell:
If one wants to create CH profile for foot and bike but disable for car due to turn costs support this is not possible. We should implement a list of weightings where the CH profile should be configured for like chWeightings instead of chWeighting.

Furthermore it should be possible to use different weightings for different vehicles e.g. curvature for motorcycle and priority for bike.

Partially implemented through #623

easbar:
This issue seems quite important considering the changes for #1776, #1708  and also #1820. Currently CH profiles are configured like this:

```
graph.flag_encoders=car,motorcycle|turn_costs=true,bike
prepare.ch.weightings=fastest|u_turn_costs=30,shortest
prepare.ch.edge_based=edge_and_node
```

which means:

- there are three flag encoders/vehicles one of which needs turn cost support so there will be a `TurnCostStorage`
- there are two weightings for CH preparation so for each of the vehicles there will be at least two CH preparations (one for each weighting)
- for every encoder with `turn_costs=true` there will be in fact two preparations for each weighting (one node- and one edge-based). if we set `prepare.ch.edge_based=off` we only get the node-based one and if we set `prepare.ch.edge_based=edge_or_node` only the edge-based one

so in total we get the following preparations:

- car|fastest|node
- car|shortest|node
- motorcycle|fastest|node
- motorcycle|fastest|edge|u_turn_costs=30
- motorcycle|shortest|node
- motorcycle|shortest|edge
- bike|fastest|node
- bike|shortest|edge

The drawbacks of the current method are:

1) its not intuitive and rather hard to understand
2) we cannot select the weighting for each vehicle individually, for example skip bike|shortest, but add car|short_fastest
3) we cannot customize the weightings, for example set the time/distance factor on `ShortFastestWeighting`. this is also problematic, because if we did set this factor it would have to be included in the profile name, otherwise we would not be able to distinguish weightings with different choices of this factor
4) we cannot use block area (I guess for the same reason)
5) the custom weightings in #1776 cannot be mapped to this current simplistic configuration scheme


Bullets 1) and 2) are easy to fix: We could just use a list similar to the one I have written above. The only downside would be that it breaks backward compatibility (we could add a converter though).

[...]

PR #534, opened 2015-10-04
Added CurvatureWeighting to improve Motorcycle routing.

boldtrn:
We created a weighting that creates fun routes for riding with a motorcycle (at least what is fun to me, when riding the motorcycle). Unfortunately we have no clue on how to create UnitTests for the routing, therefore no tests were added.

Thins that would be also nice to add are:
- Consider elevation. Roads that include elevation are usually fun to ride with the motorcycle.
- Allowing the user to create a list of nice roads that are preferred.

Also I modified the MotorcycleFlagEncoder in a way that improved the routing in my opinion. This PR belongs to #425


core/src/main/java/com/graphhopper/routing/util/CurvatureWeighting.java
```diff
@@ -0,0 +1,138 @@
...
+    {
+        super(flagEncoder, pMap);
+        this.flagEncoder = flagEncoder;
+        this.nodeAccess = ghStorage.getNodeAccess();
+    }
+
+    @Override
+    public double getMinWeight( double distance )
+    {
+        return 0.001 * distance;
```
karussell: This is probably not correct. It needs to be a good approximation to make A* working fast and correct. If it is too small A* will be too slow, if it is bigger then a single value in real world then route results can be incorrect.

core/src/main/java/com/graphhopper/routing/util/CurvatureWeighting.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * A really small bendiness or a bendiness greater than 1 indicates an error in the calculation. Just ignore them.
+     * We use bendiness > 1.2 since the beelineDistance is only approximated,
+     * therefore it can happen on straight roads, that the beeline is longer than the road.
+     */
+    protected double correctErrors( double bendiness )
+    {
+        if (bendiness < 0.01 || bendiness > 1.2)
+        {
+            logger.info("Corrected a bendiness of: " + bendiness);
```
karussell: no logs in Weighting please :)

Either throw an exception, add more unit tests or make it working all the time. There are really rare cases where none of them is possible.

karussell:
Would you mind to look into the failing integration tests? If the new results are indeed an improvement?

Also I would not want to have the expensive calculations for the distance in the Weighting class, can you check if you do them before and store them somewhere how much faster this is? I would be fine with it if it is like only 10% worse

karussell:
If you want to fine tune the routing I recommend to specify some external variable like `bendiness_booster` which you can e.g. pass through the constructor and take from the URL parameter, which influences the weighting. GraphHopper passes it from the URL in the PMap and you can easily grab like [so](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/FastestWeighting.java#L48)

BTW: The calcWeight method should be **very** fast due to two reason
 * faster execution speed for none-CH and CH
 * faster CH preparation

So precalculate as much as possible and e.g. leave only some external variables open which configures the final routing.

> Consider elevation. Roads that include elevation are usually fun to ride with the motorcycle.

Let us create a new issue for this as it could be a bit more effort. But you could check if elevation is available let it influence the priority. Is up and downhill different? If we can avoid this it would save us lots of resources (RAM) and make it easier as both weights would be identical. Otherwise a similar approach like the one in bike2 would be necessary: like [here](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/Bike2WeightFlagEncoder.java#L68) but other changes are then required like described [the docs](https://github.com/graphhopper/graphhopper/blob/master/docs/core/create-new-flagencoder.md#different-forward-and-backward-weights)

boldtrn:
Hi peter,

I just moved the bendiness calculation to the MotorcycleFlagEncoder, probably still some work to do, but it works (at least on my machine ;)) and it should be better than the previous version. However, I am not sure if its a good idea to add it to the MotorcycleEncoder or if creating a new Encoder would be better?

I will look into the height thing in the future. In general driving upwards is preferred by most motorcyclists, but I think we should not care about that currently. Any height changes are preferred to driving horizontally. So this could be a second parameter of the motorcycle encoder. Thanks for the references to the docs.

The `booster` is a good idea. That should be considered in the future.

karussell:
Thanks, would you also have a look into the failing tests? (use `mvn clean test verify` or run the RoutingAlgorithmIT, IT stands for integration test)

I would for now make the bendiness only available to motorcycle, maybe you/we find a way later to extract and reuse it elsewhere.

boldtrn:
Hi Peter,

yes I know about the tests. The errors occur because I changed the speed and priority of some streets. Therefore the expected values differ from the actual values. Are you fine when I modify the  tests or should I rather reset the speed and priorities of the streets?

karussell:
After verifying the new routes e.g. via starting the UI with the areas locally I'm fine with resetting the expected values.

> So if I understand you correctly, I shall also add a similar Test for the CurvatureWeighting, correct?

Yes, best would be an integration test (but we already have it when using the new weighting?) and one or two unit tests without import requirements just set some bendiness values in the edge and assert the weighting and test(s) for the bendiness calculation itself

boldtrn:
Hi Peter,

I just fixed the Tests and basically copied the fastest test to use it for the curvature weighting, which produces the same results. 

> one or two unit tests without import requirements just set some bendiness values in the edge and assert the weighting and test(s) for the bendiness calculation itself

Alright, I can do something like this as well.


Best,
Robin

boldtrn:
Hi Peter,

I just added the required test. But I had to copy the `initExampleGraph` method from the `Bike2WeightFlagEncoderTest`. The method should be probably added to some `TestHelper` but I am not sure about the appropriate position. This is probably something you should decide ;)

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -940,9 +940,16 @@ public Weighting createWeighting( WeightingMap weightingMap, FlagEncoder encoder
                 return new PriorityWeighting(encoder, weightingMap);
             else
                 return new FastestWeighting(encoder, weightingMap);
+        } else if ("curvature".equalsIgnoreCase(weighting) || weighting.isEmpty())
```
karussell: weighting.isEmpty should be removed here (and is already handled one if before) ... I make this in the squashed commit ...

karussell:
Thanks again, this looks good. 

> But I had to copy the initExampleGraph method from the Bike2WeightFlagEncoderTest.

This is okay for now I think.

But I cannot create a squashed commit, I'm not that experienced with git. Can you somehow rebase the changes so that we have only one commit for all the changes? Or e.g. extract a patch and use that for one commit?

Also the germany.c_300.kml should be probably removed from the pull request as I cannot find the usage of it (?)

boldtrn:
> Also the germany.c_300.kml should be probably removed from the pull request as I cannot find the usage of it (?)

Removed

> But I cannot create a squashed commit, I'm not that experienced with git. Can you somehow rebase the changes so that we have only one commit for all the changes? Or e.g. extract a patch and use that for one commit?

It should be possible by just click on the Auto-Merge button in github?

karussell:
> It should be possible by just click on the Auto-Merge button in github?

No idea, I only use command line git. I could probably do it via extracting a patch out of this PR but then you won't show up as the author which is far from what I want :)

... but maybe I found a [different simpler solution](http://stackoverflow.com/questions/5308816/how-to-use-git-merge-squash)

karussell:
Okay via the following author fake it was possible and very easy:
```bash
git merge --squash issue_534
git add .
git commit --author "your name <your@ema.il>" -m "improve ..."
```

So thanks again - I'll keep it open until successful deployment in a few days

devemux86:
@karussell nice Git hint!

boldtrn:
@karusell I am a terminal user as well. But all prs that are automatically mergeable, can be merged by github automatically. You just need to click on the green auto merge button in the bottom of the pr. So no terminal magic is required at this point. After merging you can also delete the branch. Next time you pull from master the branch is merged to the master.

karussell:
The problem with this button is the clumsy history. For one PR I prefer just one commit :) ... although I have to admit that single commits from my side aren't as clean as I want them to ;)

karussell:
This should be live now. But **fastest** weighting only, due to current small limits of multiple vehicles and CH. See #493.

As contributor I've added @boldtrn someone else?

boldtrn:
Hi Peter,

sorry for the late reply. Sounds great :) @seeebiii and @ChristianSeitzer  were part of the team as well. If you don't mind it would be nice if you could add them as well :)

Hopefully I will find some time in several weeks to improve the curvature weighting as discussed above.

Best,
Robin

karussell:
Thanks to them as well - added them!

PR #535, opened 2015-10-06
Display OSM data date for import in web UI

ratrun:
Now we are able to read the age of OSM data from the osmosis_replication_timestamp field according to https://wiki.openstreetmap.org/wiki/PBF_Format#Definition_of_the_OSMHeader_fileblock, propagate
and display this information in the web UI.

Please note that I'm now directly returning the osmDataDate and reuseing the `formatDateTime` method for formatting as suggested in #530. Thank you for this input. This change triggered that I had to make `formatDateTime` public such that it can be used in the test.


ratrun:
Usage of `formatDateTime` was probably not a good idea. As feared, travis - which apperently is running in a different time zone, is failing now:

Failed tests: 
  OSMReaderTest.testDataDateWithinPBF:855 expected:<2014-01-02T0[1]:10:14+0000> but was:<2014-01-02T0[0]:10:14+0000>

karussell:
But this is a good indication that this needs improvement as we should use UTC :)

karussell:
Ups, thanks!

karussell:
Hmmh, indeed then formatDateTime seems to be wrong and needs this fix. I'll see what I can do.

ratrun:
Thanks for the code snip input. Now I found the two problems and fixed them. If needed, I can merge the commits into one.

karussell:
I've simplified a bit of the date parsing/formatting part and unified the approach in all the (3) places. See [this commit](https://github.com/graphhopper/graphhopper/commit/7111537a54e65f20076dae15a4d24e733f149991) and [this branch](https://github.com/graphhopper/graphhopper/commits/issue_535)

The travis tests [do pass](https://travis-ci.org/graphhopper/graphhopper/builds/84788783). Let me know if you are fine and I'll squash the commits and merge then!

Is it e.g. intended that there is no date formatting [here](https://github.com/graphhopper/graphhopper/commit/7111537a54e65f20076dae15a4d24e733f149991#diff-c8a086dc671dd6aee9861a2ba6245afcR168)?

ratrun:
I missunderstood your comment. The code is cleaner now indeed. But the mixed datatime format in the display does not look that great. I wanted to avoid it. But if it is fine for you, then its good. Additionally I noticed a small typo -  the "Jar Date" should be better "Jar date":

Import date: 2015-10-11T17:27:37Z
GH Version: 0.6
Jar Date: 2015-10-11T19:26:34+0200

karussell:
Thanks!

> But the mixed datatime format in the display does not look that great.

Well this is the same but you are right that we should definitely change this in the pom.xml too ... I'll also fix the typo

karussell:
Merged.

Issue #437, opened 2015-06-08
Improvements to JSON format of web module


karussell:
* different APIs of the Directions API should use same time unit: seconds (currently routing API has ms)
 * with alternative paths?
 * different layout for via-points? 'legs'
 * turn_angle is in radian

michaz:
> use JSON request instead URL parameter

Are there any examples where this is considered a good idea? I thought the whole point about REST was that a resource is identified by its URL alone, and a shortest path is _the_ perfect example of a resource.

navitia.io seems to agree.

karussell:
> Are there any examples where this is considered a good idea?

ElasticSearch, our Route Optimization API, mapzen, ...

> I thought the whole point about REST was that a resource is identified by its URL alone, and a shortest path is the perfect example of a resource.

Why? There are lots of possible parameters. E.g. it is time, vehicle size etc dependent. I do not see it as a resource. I see it more as a 'calculation job'...

michaz:
> Why?

Because it doesn't change the world, because it returns the same result until it is explicitly changed by something, because it can be cached, ... 

Google search is GET

michaz:
> remove insecure JSONP?

+1

Not only because security, but also because we are not a web framework.

I think in a few years we will say "Oh, you want our API a bit different? Go use a web tool to click together a proxy/gateway which changes it like you want it."

karussell:
> because it returns the same result until it (the road network) is explicitly changed by something, because it can be cached

I would argue against this (as real time road changes will invalidate this and it is unclear which routes to invalidate), but I think this is indeed discussible :)

Let me give you one "beauty" argument as why we shouldn't care much about principled web stuff here:

we'll be able to configure a CH vehicle profiles with a JSON file. Also the default vehicle profiles for non-CH should be configurable by JSON and now, we are forced due to beauty reasons ;), to use the **same JSON** to request our API to dynamically change the vehicle profile per request.

> And, of course, as soon as you start thinking about subscriptions to routes

Yes, but then you could also argue about a resource that you subscribe to ... again: I wouldn't care much. BTW: [navitia api](https://github.com/CanalTP/navitia/wiki/OpenTripPlanner-and-Navitia-comparison#history-1) is reeaally old ;)

michaz:
[...]

PR #536, opened 2015-10-06
Issue 310: Generated GPX file contains both <rte> and <trk> tags

### LINKED ISSUE #310: Generated GPX file contains both <rte> and <trk> tags
Hi, 

as discussed separately with Peter, I am encountering issues with the GPX graphhopper generates.
Indeed, as they contain both the rte and trk tags, the softwares I use are computing both thus generating a longer route than created.

As an example, I created the following mountain bike route: 
https://graphhopper.com/maps/?point=43.87287%2C7.006069&point=43.866868%2C7.005318&point=43.851581%2C6.99415&point=43.844123%2C7.026186&point=43.839665%2C7.087083&point=43.843999%2C6.993184&point=43.849879%2C6.967607&point=43.872792%2C7.006075&vehicle=foot&elevation=true&layer=TF%20Cycle

Here is a recap of the behavior I am getting with different tools:
 - josm display both the route and the track when opening the GPX file.
 - Oruxmaps on Android considers both rte and trk, thus a 35km track becomes a 60km route+track overall. The elevation profile is pretty clear: you see more than 20kms at elevation 0 and then the real track starts.
 - this website also has the same symptoms : http://www.visugpx.com/index.php?i=N4SHU8rKaV (the URL is valid for 24hours). Here is a screenshot (which is more or less what Oruxmaps does as well):

http://www.imagehangar.com/view.php?pic=small/29703700_1422363331.jpg

I'm not sure what should exactly be the content of the gpx and I understand both route and track have their use but it would be nice to have the choice when downloading the gpx.
For example, http://www.openrunner.com/ has 2 different buttons to generate the gpx: one for route and one for track.

Any thoughts about this?

Regards,
Florent.

ratrun:
Launch a popup window in JS where the user can select which GPX data parts shall be exported.
Additionally add the possibility for exporting of GPX waypoints.
Please note that it was necessary to update the jquery-ui version, as otherwise the popup diolog did not show up. I somewhere found the suggestion to update the version on stackoverflow and the updated version works.

Fixes #310


karussell:
:+1: nice!

karussell:
In the console it says to me: `$(...).sortable is not a function` and the web UI is scrambled a bit

Is the a js module missing?

ratrun:
Yes, I'm sorry, one file was missing. Now its added and the change in update-translations.sh reverted.

karussell:
Thanks! This is now merged, let me know if this is okay for you!

Note that I've changed the css IDs to avoid collision in the future and also the URL parameters have now a `gpx.` prefix. Furthermore additionally to `FIXME` in the source you should add it to the issue text. See the default change for 'route' via #437

karussell:
deployed, need to document the new parameters

PR #545, opened 2015-10-12
Make SRTMProvider resolution independent, fixes #544 and #543

### LINKED ISSUE #543: SRTM provider woeks only for 1201 resolution
SRTM can be in both 1201 and 3601 resolution. the provider should not be resolution dependent.
See discussion here:
https://discuss.graphhopper.com/t/elevation-data-from-local-files/175/17

### LINKED ISSUE #544: Unable to compile on windows with maven in program files
When placing maven in c:\program files\ and running ```graphhopper.sh build``` the build fails - probably something with the grep...
As a work around I place MAVEN_HOME in the environment variables.

HarelM:
Fixes #543 *make SRTMProvider resolution independent* and #544 *fix for graphhopper.sh under windows*

See [discussion](https://discuss.graphhopper.com/t/elevation-data-from-local-files/175)

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/dem/SRTMProvider.java
+++ b/core/src/main/java/com/graphhopper/reader/dem/SRTMProvider.java
@@ -64,7 +64,8 @@ public class SRTMProvider implements ElevationProvider
 
     private static final BitUtil BIT_UTIL = BitUtil.BIG;
     private final Logger logger = LoggerFactory.getLogger(getClass());
-    private final int WIDTH = 1201;
+    private final int DEFAULT_WIDTH = 1201;
+    private final int WIDTH_BYTE_INDEX = 0;
     private Directory dir;
     private DAType daType = DAType.MMAP;
     private Downloader downloader = new Downloader("GraphHopper SRTMReader").setTimeout(10000);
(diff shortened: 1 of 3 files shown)


karussell:
Thanks!

I also understand now why you had to adjust the loading logic. Nice.

What I would need is a small test case like done for SRTMTest just ensuring loading & get is working also for resolution!=1201. Can you produce srtm data with a different resolution or are there some files under a public domain? If not I'll have to ask them first, if they (commercial) allow usage for the GraphHopper project.

HarelM:
Hi @karussell,
I added one file with a resolution that is different than 1201. see:
core/files/N31E034.hgt.zip
And added the relevant test inside another test:
```assertEquals(95, instance.getEle(31, 34), 1e-1);```
Hope that's acceptable, if not, you can move the above line to a separate test.

karussell:
Thanks! Do you know under which license this stands?

HarelM:
I'm guessing the same as Aster and SRTM, you can ask Ohad Manor on the discussing, it's the same one. although this is used only for testing and in not shipped with the product so I'm not sure you need any license, but I'm not a big expert when it comes to stuff like that.

karussell:
> I'm not sure you need any license, but I'm not a big expert when it comes to stuff like that.

Yeah, I think even if only used for tests we need to be sure as we 'spread' it.

I'll see if I can get first a license agreement for Aster and we could use the raw data from there ... or I create a simple faked ele data somehow.

karussell:
Uhm, after an hour of search and broken links I gave up. Can you point me to the aster data AND a site describing their license / attribution? What I found was just [this](https://lpdaac.usgs.gov/citing_our_data) (nothing precise) and [this](https://lpdaac.usgs.gov/how_can_i_get_free_aster_l1b_data_over_united_states_and_its_territories) 'download description' and the links to these ['apps'](http://asterweb.jpl.nasa.gov/data.asp). But when I try to get a license info of a concrete data set I only get [404 sites](https://lpdaac.usgs.gov/404/products%20aster%20products%20table%20astgtm). I've read that there are public domain 30m resolution ASTER data for the US. That is something I find interesting. Also I have contact someone there regarding the license .. we'll see.

HarelM:
I don't have anything concrete. try ask Ohad in the discussion...

karussell:
ASTER is not possible to be included in GraphHopper due to [licensing](https://lpdaac.usgs.gov/dataset_discovery/aster/aster_policies) and the need for purchase. So we'll have to create fake data first.

HarelM:
I know there is SRTM data over the US that has the same resolution, I'm sure you will be able to find a file for the tests...

HarelM:
I think this might be what you need:
http://dds.cr.usgs.gov/srtm/version2_1/SRTM1/Region_01/
Also, Since this data is in the same format as above in almost the same address and since this check-in allows using this data I would recommend starting to look for the relevant file here and if it was not found use the SRTM3. this way you'll have better resolution over the US :-)

karussell:
@HarelM thanks! Can you elaborate a bit on this? Do you mean the referenced data has the same resolution for all files so we can use this for the tests?

And does the referenced data already contain SRTM3 or do you mean we would have to implement a fall back additionally?

HarelM:
Not sure I understand your questions. This data is 3601x3601 resolution so it can be used for the tests.
Regardless of the tests, these files can be used to calculate elevation (although they do not cover a lot of area) so instead of downloading SRTM3 data and parse it try and find the relevant file in the repository above (for better resolution) and if there is no file there go to the SRTM3 repository.

karussell:
> This data is 3601x3601 resolution so it can be used for the tests.

Perfect - thanks! I'll do this tomorrow and we can finally integrate your PR!

> so instead of downloading SRTM3 data ...

this part I do not understand :)
Is SRTM3 data not covering the full area and you mean one could use SRTM1 as fallback as it has the same resolution?

HarelM:
The names are confusing, I had the same misunderstanding (I think).
SRTM3 is 3 arc resolution files covering the entire planet (1201x1201).
SRTM1 is 1 arc resolution files covering some parts in the US (3601x3601).
My suggestion was to update the SRTMProvider to try and look for the SRTM1 files for better resultions, in case the files does not exists fall back to STRM3.

karussell:
> The names are confusing, I had the same misunderstanding (I think).

ah, okay. The problem is that there where also different versions and the most recent version had also a better 30m resolution and was named SRTM3 if I correctly remember. 

> My suggestion was to update the SRTMProvider to try and look for the SRTM1 files for better resultions, in case the files does not exists fall back to STRM3.

Hmmh, one could implement such a logic (in another issue) but I think if the user does this it should not be hard with a big wget magic. E.g. he downloads the 1201 data and overwrites the available higher resolution data if he likes.

HarelM:
That's true, That's what I did in order to discover this issue :-).

karussell:
Hmmh, what I do not understand is why the export into an image does only partially work. Half of the image seems be empty. Somewhere maybe a scaling issue occurs or a precision mistake in the export code (?)
See here: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/reader/dem/HeightTile.java#L135

karussell:
temp

HarelM:
Are you sure the higher precision file is truly full of data?
it is also easy to see that the makeARGB logic and getHeight logic are different.
I currently do not have time to dig into the code, but the problem I faced were due to getCapacity which was completely irrelevant to what was in the file, so this would be my first guess...

karussell:
The capacity was indeed wrong (now it is `long len = width * width;`) but still not fixing the issue. And no, I'm not sure if it is full of data but I always expect the error on GraphHopper-side first :)

But I've now used your original zip file for the image export and this looks good, except a small purple area (1020*83) in the upper left corner. So I'm okay now with using the new 3601² source files with slightly corrupted file for the tests. Still we should keep this in mind, maybe there is still a bug in our code.

karussell:
So thanks again. Merged: 
https://github.com/graphhopper/graphhopper/commit/2e60c80beb19f46c14d9d3b5cccc084b873e4944

Please reopen or create a new issue if you found problems with it!

PR #566, opened 2015-11-23
Update config-example.properties

boldtrn:
I didn't get the essence of that property and needed to look it up in the code. It was not obvious to me that we don't store the way names if we set this setting to false. I hope other users don't struggle with that anymore.

### DIFF
--- a/config-example.properties
+++ b/config-example.properties
@@ -29,7 +29,7 @@ graph.flagEncoders=car
 # MMAP_STORE_SYNC could be used otherwise but will be a lot slower
 graph.dataaccess=RAM_STORE
 
-# if you want to reduce storage size and you don't need instructions for the resulting path use:
+# if you don't need turn instruction, you can reduce storage size by not storing way names:
 # osmreader.instructions=false
 
 # will write way names in the preferred language (language code as defined in ISO 639-1 or ISO 639-2):


boldtrn:
Just saw that the CI for one environment fails because of a maven exception. This commit should not break anything :/

karussell:
Thanks, docs should be clear :) !

> if you want to reduce storage (because way names are not stored) size and you don't need instructions for the resulting path use

As we are at it: Would you mind to reformulate this without the brackets? E.g. ?
`if you don't need way names e.g. for the turn instructions of a path you can reduce storage size significantly by:`

boldtrn:
Hi Peter,

I removed the brackets and wrote it like:
`if you don't need turn instruction, you can reduce storage size by not storing way names:`

I am not sure if this is actually better than your proposed formulation. I think it is better this way since the parameter is called `osmreader.instructions`. So it seems natural to disable instructions using this parameter (as a bonus you safe storage).

karussell:
Ok, sounds good too :) Now would you mind to squash the commits into one :) ?

http://de.gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

boldtrn:
Well I seem to be unable to squash the commits. When I squash the commits the repository does not accept my changes anymore. I will close this PR and create a new one with a single commit.

karussell:
When you squash you need to push force the changes. **But** keep in mind that you should never elsewhere force your push, so do this only for feature branches or here like pull requests where you are the only one with access to the branch, as this will otherwise mess up lots of stuff for other potential contributors :)

karussell:
> Just saw that the CI for one environment fails because of a maven exception. This commit should not break anything :/

Ugly. It looks like travis has currently some network issues :(

Issue #230, opened 2014-07-07
Car: Different max speed for inner and outer city


karussell:
Currenty max speed relies on the maxspeed tag in the city or village and often streets do not have such a tag. We need to grab the boundaries of the city and take this into account. 

This issue is related to #75 and #175

Possible improvements afterwards:
- [Route through Prague](https://graphhopper.com/maps/?point=50.1255%2C14.4588&point=50.0332%2C14.4003&locale=de&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale) should be 20 minutes not 16
- [Laval, France](https://graphhopper.com/maps/?point=48.038496%2C-0.825005&point=48.096885%2C-0.70158&locale=de&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale) should be 20 not 15 minutes and using one of the northern motorways not the southern (as first suggestion)
- [Mettmann Germany](https://graphhopper.com/maps/?point=51.255899%2C6.939411&point=51.250527%2C7.014427&locale=de&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale) should avoid the inner city and slightly slower (not 8min but more like 10min)
- [Bayreuth, Germany](https://graphhopper.com/maps/?point=49.956963%2C11.582508&point=49.936085%2C11.549892&locale=de&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale) should use the nordring (as first suggestion) and avoid the inner city, also not 5min but more like 10
- [Paris](https://graphhopper.com/maps/?point=48.924304%2C2.603073&point=48.826983%2C2.243614&locale=de&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale) oh, not through Paris! And not 33min but more like 40-45min (at least)
- [London](https://graphhopper.com/maps/?point=51.556369%2C-0.155525&point=51.458499%2C-0.09613&locale=de&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale) Time is 20 but should be more like 40min (but here going through the city is similar slow ;))
- [London](https://graphhopper.com/maps/?point=Cricklewood%2C%20NW2%202DQ%2C%20London%2C%20United%20Kingdom&point=51.539983%2C-0.178871&locale=de&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale) should be more like 10-15 minutes not 6
[...]

Issue #75, opened 2013-06-10
Interpret traffic lights


NopMap:
While I was driving through town today, rejecting the routing suggestions of my professional car navigation, I had this idea: The navigation was suggesting a road that looked faster according to distance and street class. But I know that there are a series of traffic lights on that road and it is actually faster to drive around it on a road without lights.

Common navigation systems are usually pretty bad about this, they are too optimistic when you drive downtown. I heard from a navigation guy a while ago that the commercial navigation data does not contain traffic lights.

So basically we should evaluate the (500000) traffic lights in the OSM data and if an edge has a traffic ligth, add a small speed penalty to that edge for car and bike traffic.

Also, I found that there is a tag for pedestrian crossings 
http://wiki.openstreetmap.org/wiki/DE%3ATag%3Ahighway%3Dtraffic_signals

Evaluating this could make a route much safer, but it's rather complicated, I don't have a simple idea on how to do it.
- [ ] clean up import and make node tags influence possible
- [ ] tune speed decrease but keep in mind that traffic lights could be off at night so use a low influence as start.

jansoe:
I implemented a [brach](https://github.com/jansoe/graphhopper/tree/TrafficLights2) to take traffic lights into account. Although this leads to quite realistic travel times, the routing now becomes sometimes a bit arkward, trying to circumvent traffic lights by going around in zick-zack. It seems, that one can only take traffic lights into account, if general turncosts are taken into account.

andreaswolf:
@jansoe I need further information (where on the route are traffic lights, roundabouts, crosswalks, …) for the application in my diploma thesis. As far as I see, your information is added during the prepare step, right? Do you know of a way to preserve that information for later processing? Is there a possibility to extract all this info from a final route?

If you need help with the turncost implementation (which would also be interesting for me), drop me a line.

PR #571, opened 2015-11-23
Added Elevation to Motorcycle Routing

boldtrn:
Basically I added elevation data to the motorcycle routing. I worked with srtm elevation data. The values significantly changed when using the cigar data. But somehow the cigar data seemed not to work that good? Even though you mention it would be the better one.

The elevation widget for the cigar data didn't work in the UI.

Fixed a serious bug when calculating the bendiness. Results should be better now.

BTW: Why did you changed the min bendiness in the CurvatureWeighting to 1? Actually the Bendiness can be 0.
BTW2: Successfully squashed the commits ;) 

Best,
Robin

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -576,9 +576,9 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
 
         if (way.hasTag("scenic", "yes"))
         {
-            if (weightToPrioMap.lastEntry().getValue() < BEST.getValue() )
-            // Increase the prio by one step
-            weightToPrioMap.put(110d,  weightToPrioMap.lastEntry().getValue() + 1);
+            if (weightToPrioMap.lastEntry().getValue() < BEST.getValue())
+                // Increase the prio by one step
+                weightToPrioMap.put(110d, weightToPrioMap.lastEntry().getValue() + 1);
         }
     }
 
@@ -604,18 +604,17 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
             wayType = WayType.ROAD;
         else
             wayType = WayType.OTHER_SMALL_WAY;
-        
+
         boolean isPushingSection = isPushingSection(way);
         if (isPushingSection && !partOfCycleRelation || "steps".equals(highway))
             wayType = WayType.PUSHING_SECTION;
-        
+
         if (way.hasTag("bicycle", intendedValues))
         {
             if (isPushingSection && !way.hasTag("bicycle", "designated"))
                 wayType = WayType.OTHER_SMALL_WAY;
-            else
-                if ( (wayType == WayType.OTHER_SMALL_WAY ) || (wayType == WayType.PUSHING_SECTION) )
-                    wayType = WayType.CYCLEWAY;
+            else if (wayType == WayType.OTHER_SMALL_WAY || wayType == WayType.PUSHING_SECTION)
+                wayType = WayType.CYCLEWAY;
         } else if ("cycleway".equals(highway))
             wayType = WayType.CYCLEWAY;
 


karussell:
> But somehow the cigar data seemed not to work that good? Even though you mention it would be the better one.

Yes, they are better. What do you mean with 'not working' in this context?

> BTW: Why did you changed the min bendiness in the CurvatureWeighting to 1? Actually the Bendiness can be 0

Sorry if I 'fixed' something without asking you for assistance. This values needs to be exact (it should not overestimate) and I tried to understand all the max and min values of the formula you used and obviously made some mistakes. I'll have a look again.

karussell:
Would you mind to fix the formatting that e.g. the opening bracket has to be on a new line? E.g. use NetBeans or the IntelliJ settings in the core/files folder or manually adjust Eclipse. Otherwise there are many unimportant changes here.

boldtrn:
Allright formating should be fixed.
Also minBendiness is at .1. Shouldn't become lower than this.

You'r welcome ;)

> Yes, they are better. What do you mean with 'not working' in this context?

It does not show anything, at least on my machine, not sure if this is something local. But haven't tried in depth.

boldtrn:
Ok, I just changed some things. I adjusted the min length that a edge needs to have to be even calculated and the minimum number of nodes (3).

Also I did a tag based rating. Streets that have sidewalks or lighting are rated worse.

I changed the speeds of the streets. Especially the speed of secondary and tertiary. You can drive there way faster on a motorcycle.

core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
```diff
@@ -350,19 +365,63 @@ public void applyWayTags( OSMWay way, EdgeIteratorState edge )
     {
         double speed = this.getSpeed(edge.getFlags());
         double roadDistance = edge.getDistance();
-        double beelineDistance = getBeelineDistance(way);
+        double beelineDistance = getBeelineDistance(edge);
+        PointList pointList = edge.fetchWayGeometry(3);
```
karussell: I would reduce the number of calling the relative expensive fetchWayGeometry. E.g. pass the pointList to the getBeelineDistance method?

core/src/test/java/com/graphhopper/routing/RoutingAlgorithmIT.java
```diff
@@ -104,13 +105,14 @@ public void testMonacoMotorcycle()
         list.add(new OneRun(43.730729, 7.42135, 43.727697, 7.419199, 2703, 119));
         list.add(new OneRun(43.727687, 7.418737, 43.74958, 7.436566, 3749, 170));
         list.add(new OneRun(43.728677, 7.41016, 43.739213, 7.4277, 3175, 169));
-        list.add(new OneRun(43.733802, 7.413433, 43.739662, 7.424355, 2423, 141));
-        list.add(new OneRun(43.730949, 7.412338, 43.739643, 7.424542, 2253, 120));
+        list.add(new OneRun(43.733802, 7.413433, 43.739662, 7.424355, 5636, 298));
+        list.add(new OneRun(43.730949, 7.412338, 43.739643, 7.424542, 5467, 277));
```
karussell: Did you look if the new routes in monaco are more reasonable or why is there this big difference?

karussell:
Thanks for the improvements! Would you revert the formatting changes in OSMReader and the tests? Also graphhopper.sh needs a revert so that we'll fix this in the separate issue.

core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
```diff
@@ -93,30 +100,37 @@ public MotorcycleFlagEncoder( int speedBits, double speedFactor, int maxTurnCost
...
         // spielstraße
         defaultSpeedMap.put("living_street", 5);
         defaultSpeedMap.put("service", 20);
         // unknown road
-        defaultSpeedMap.put("road", 20);
+        defaultSpeedMap.put("road", 40);
         // forestry stuff
         defaultSpeedMap.put("track", 15);
 
+        enjoyableTags.add("scenic");
```
karussell: Interesting! Did not knew about this. Is this usable for bike too @ratrun ?

http://taginfo.openstreetmap.org/keys/scenic

core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
```diff
@@ -93,30 +100,37 @@ public MotorcycleFlagEncoder( int speedBits, double speedFactor, int maxTurnCost
...
+
+        // List is typcially set in urban areas
+        frustratingTags.add("lit");
```
karussell: Oh, why is this frustrating :) ?

core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
boldtrn: Yes sure, I thought about that as well, but thought the method call would not be expensive.

core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
boldtrn: Unfortunately it is not used very often, especially on roads. It's oftimes used on trails and paths http://overpass-turbo.eu/s/cVF. But there are some roads and according to the wiki it should be used for beautiful roads e.g. http://wiki.openstreetmap.org/wiki/Talk:Key:highway#Scenic.2Fhistorical_roads.

core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
boldtrn: Exactly that was a quick and probably dirty fix, but the results work quite well, at least for my area. But you are right if #230 is fixed we should use these results. I hope you are fine if we leave this here until #230 is ready?

But in general lit roads are a sign of high traffic volume and we don't want traffic.

core/src/test/java/com/graphhopper/routing/RoutingAlgorithmIT.java
boldtrn: To be honest, no I haven't checked them. Something quite crazy is happening here. With this file `core/files/monaco.osm.gz` the results are crazy (and actually not reasonable) and the distances are as set above. 
If I use the osm.pbf file from http://download.geofabrik.de/europe/monaco.html, the distances are fine and the routes are absolutely ok. So I really don't know what to do here?

boldtrn:
Your comments except for the tests are pushed. I can squash them when its mergeable.

core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
ratrun: Thank you for the hint about the scenic tag. I didn't know it, although I did search for something like this a long time ago, The wiki page was created in 2012 and I searched before that time. I'll prepare a PR for the bicycles. I'll intent to increase the preferene by one level. Although this changes looks easy it might take some time.

core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
karussell: Sure, this is fine. BTW: also #75 could be used to determine slow down.

core/src/test/java/com/graphhopper/routing/RoutingAlgorithmIT.java
karussell: Keep in mind that "shortest" is used and a relative old version of monaco ... but I'll have a look if you don't find the reason. Maybe due to the elevation?

core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
karussell: nice and no time pressure please :) !

core/src/test/java/com/graphhopper/routing/RoutingAlgorithmIT.java
boldtrn: Yes I compared it with shortest. I think the issue is related to speed limites. Probably in the old version of Monaco the speed limits are not set properly. Since I changed the default speed lmits, the routing could be wrong. That would be my guess, since the elevation thing is only relevant when using the curvature routing.

PR #579, opened 2015-11-26
Unique paths for pr

boldtrn:
This PR contains the possibility to request unique paths. 

This is possible by adding `unique_paths=true` to the request. Currently this only works with `Dijkstra` and the `DijkstraBi`.


karussell:
What are the advantages over a [custom weighting](https://github.com/graphhopper/graphhopper/blob/master/docs/core/weighting.md)?

boldtrn:
The idea was to make it independant from the weighting. So that fastest, shortest, curvature, what ever works with unique paths. But I think adding it to the weighting seems natural.

How about extending the weighting interface so that additional weighting strategies can be defined (basically a strategy pattern). This would allow us to define additional strategies (or boosters/modifiers) for a given weighting. 

A possible use case are the unique paths. Other usecases are possible as well (Even though I cannot think of one).

karussell:
> How about extending the weighting interface so that additional weighting strategies can 
> be defined (basically a strategy pattern). This would allow us to define additional 
> strategies (or boosters/modifiers) for a given weighting.

Yes, this is a good idea - I would avoid fiddling with the algorithms as it they are complex enough and we hopefully get more of them and tune them further.

We also need this more configurable `Weighting` stuff when we [improve the flexibility](https://discuss.graphhopper.com/t/more-flexibility-in-gh/71) (hopefully soonish), where we can post a profile via JSON etc. When I initially thought about it I think we provide a map (json struction) and grab certain properties for fast calculation in `calcWeight` but as this is not implemented I cannot say if it works reasonable (easy to use and fast). The problem with calcWeight is that it is potentially invoked millions of times per route (hundres&thousands of times only for CH) calculation and it needs to be very fast.

boldtrn:
I just added the Interface for the weighting strategy. What do you think? Are you happy with that architecture?

karussell:
> @karussell I just added the Interface for the weighting strategy. What do you think? Are you happy with that architecture?

Hmmh :) ... then the rest of the changes like in GraphHopper.java won't be necessary, right?

We really have to be careful here regarding performance and need a mechanism which adds no overhead. E.g. using the 'new' ;) for-loop pattern is slower than via indices. What do you think about the superWeighting mechanism instead (used for TurnWeighting)? One would have to measure the fastest and potentially refactor the TurnWeighting. 

Still I think this has to be decoupled from this issue in the first place, maybe finally we can implement a simple AvoidEdgeIdWeightingStrategy or something, here?

core/src/main/java/com/graphhopper/routing/util/AbstractWeighting.java
```diff
@@ -32,6 +38,40 @@ public AbstractWeighting( FlagEncoder encoder )
             throw new IllegalStateException("Not a valid name for a Weighting: " + getName());
     }
 
+
+    @Override
+    public double calcWeight( EdgeIteratorState edgeState, boolean reverse, int prevOrNextEdgeId )
+    {
+        double weight = calcPureWeight(edgeState, reverse,prevOrNextEdgeId);
+
+        if(adjustments != null){
```
karussell: btw: lazy initialization here is not good as although calcPath is single threaded in core GH, this can change and others can have multi-threaded algos. A weighting should be seen as read-only.

karussell:
> One would have to measure the fastest 

BTW: there is a simple procedure to measure if there is a difference: `./graphhopper.sh measurement your.pbf [count]`. Where the count parameter can be e.g. 2 and then the current and the last commit are used to run the measurement. Simple mechanism to compare performance for multiple commits.

core/src/main/java/com/graphhopper/routing/util/AbstractWeighting.java
boldtrn: Yes I thought about it and thought that a null check is faster then querying `size`.

boldtrn:
> Hmmh :) ... then the rest of the changes like in GraphHopper.java won't be necessary, right?

The changes in GraphHopper.java are still necessary. The issue is, calcWeight does not know if the edge was used or not. So you have to update, once a `Path` was created. (https://github.com/graphhopper/graphhopper/pull/579/files#diff-59d362881b641ff1b933a3785503c20eR1125)

When avoiding already explored edges, no path will be found (I tried that first).

> What do you think about the superWeighting mechanism instead (used for TurnWeighting)

I thought about chaining weighting objects (this is basically the same as in TurnWeighting?). I chose this approach because I think it is more flexible and less magic is involved. But performance wise it is probably faster, especially if no modification is set.

boldtrn:
@karussell I just thought about this issue. Probably something like the TurnWeighting makes more sense in this case, because we do not change anything if we only use a regular Weighting. I will change this PR accordingly.

boldtrn:
@karussell My changes are pushed. I hope you like it. 

Unfortunately I pulled from the graphhopper master and github shows all changes made including the ones from master.

karussell:
> Unfortunately I pulled from the graphhopper master and github shows all changes made including the ones from master.

Hmmh, that is strange and hard to review, also merging this will lead to a confusing history - can you clone the latest master again and copy just the 3 files you changed?

As you have the abstract weighting class: I would either refactor the TurnWeighting to use the abstract class also there or remove the abstract class completely as all 'abstract' classes with just one implementation are a bit overhead :) 
I do not like 'premature optimization' for code too, although I have to admit that GH is not 100% free of this ;)

boldtrn:
Ok, I will close this PR.

PR #580, opened 2015-11-27
Increase bike priority level by one step in case of scenic=yes tag.

ratrun:
Triggered by #571.


karussell:
Simple&nice - thanks!

PR #582, opened 2015-11-28
Random Round Tours

boldtrn:
This PR contains the possibility to create random round tours. When passing `tour_length=APPROX_DISTANCE_IN_KM` to the request it returns a roundtour of approximately the requested distance. The actual distances can vary a lot, since I only calculate the beeline distance of the waypoints.

The current results are ok, but not good. Especially because the unique_paths PR is not included here (this will probably improve the results.

Currently this feature is not available in the GUI. We just use the first waypoint and create a route.

Something that could be added as well is to calculate a number of tours and select the best one, to remove bad results created by this approach.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -1118,7 +1118,8 @@ public class GraphHopper implements GraphHopperAPI
         // Every alternative path makes one AltResponse BUT if via points exists then reuse the altResponse object
         PathWrapper altResponse = new PathWrapper();
         ghRsp.add(altResponse);
-        boolean isRoundTrip = AlgorithmOptions.ROUND_TRIP_ALT.equalsIgnoreCase(algoOpts.getAlgorithm());
+        boolean isRandomRoundTrip = AlgorithmOptions.ROUND_TRIP.equalsIgnoreCase(algoOpts.getAlgorithm());
+        boolean isRoundTrip = AlgorithmOptions.ROUND_TRIP_ALT.equalsIgnoreCase(algoOpts.getAlgorithm()) || isRandomRoundTrip;
         boolean isAlternativeRoute = AlgorithmOptions.ALT_ROUTE.equalsIgnoreCase(algoOpts.getAlgorithm());
 
         if ((isAlternativeRoute || isRoundTrip) && points.size() > 2)
@@ -1151,6 +1152,9 @@ public class GraphHopper implements GraphHopperAPI
 
             sw = new StopWatch().start();
             RoutingAlgorithm algo = tmpAlgoFactory.createAlgo(queryGraph, algoOpts);
+            if(algo instanceof RoundTripAlgorithm){
+                ((RoundTripAlgorithm) algo).prepare(locationIndex, new DefaultEdgeFilter(encoder), points.get(0));
+            }
             algo.setWeightLimit(weightLimit);
             String debug = ", algoInit:" + sw.stop().getSeconds() + "s";
 
@@ -1193,7 +1197,7 @@ public class GraphHopper implements GraphHopperAPI
             }
         } else if (isRoundTrip)
         {
-            if (points.size() != altPaths.size())
+            if (points.size() != altPaths.size() && !isRandomRoundTrip)
                 throw new RuntimeException("There should be the same number of points as paths. points:" + points.size() + ", paths:" + altPaths.size());
 
             pathMerger.doWork(altResponse, altPaths, tr);
--- a/core/src/main/java/com/graphhopper/routing/AlgorithmOptions.java
+++ b/core/src/main/java/com/graphhopper/routing/AlgorithmOptions.java
@@ -62,6 +62,10 @@ public class AlgorithmOptions
     /**
      * round trip algorithm based on alternative route algorithm
      */
+    public static final String ROUND_TRIP = "roundTrip";
+    /**
+     * round trip algorithm based on alternative route algorithm
+     */
     public static final String ROUND_TRIP_ALT = "roundTripAlt";
     private String algorithm = DIJKSTRA_BI;
     private Weighting weighting;
--- a/core/src/main/java/com/graphhopper/routing/RoundTripAlgorithm.java
+++ b/core/src/main/java/com/graphhopper/routing/RoundTripAlgorithm.java
@@ -1,0 +1,124 @@
+/*
+ *  Licensed to GraphHopper and Peter Karich under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for 
+ *  additional information regarding copyright ownership.
+ * 
+ *  GraphHopper licenses this file to you under the Apache License, 
+ *  Version 2.0 (the "License"); you may not use this file except in 
+ *  compliance with the License. You may obtain a copy of the License at
+ * 
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.routing;
+
+import com.graphhopper.routing.util.EdgeFilter;
+import com.graphhopper.routing.util.UniquePathWeighting;
+import com.graphhopper.routing.util.Weighting;
+import com.graphhopper.routing.util.tour.TourWayPointGenerator;
+import com.graphhopper.storage.Graph;
+import com.graphhopper.storage.index.LocationIndex;
+import com.graphhopper.util.shapes.GHPoint;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class implements the round trip calculation via randomly generated points
+ * <p>
+ * @author Robin Boldt
+ */
+public class RoundTripAlgorithm implements RoutingAlgorithm
+{
+    private final Weighting weighting;
+    private LocationIndex locationIndex = null;
+    private EdgeFilter edgeFilter = null;
+    private GHPoint from = null;
+    private final double distanceInKm;
+    private Graph g;
+    private AlgorithmOptions opts;
+    private int numberOfVisitedNodes = 0;
+    private double weightLimit;
+
+    public RoundTripAlgorithm( Weighting weighting, double distanceInKm, Graph g, AlgorithmOptions opts )
+    {
+        this.distanceInKm = distanceInKm;
+        this.g = g;
+        this.opts = opts;
+        if (!(weighting instanceof UniquePathWeighting))
+            weighting = new UniquePathWeighting(weighting);
+        this.weighting = weighting;
+    }
+
+    public void prepare( LocationIndex locationIndex, EdgeFilter edgeFilter, GHPoint from )
+    {
+        this.locationIndex = locationIndex;
+        this.edgeFilter = edgeFilter;
+        this.from = from;
+    }
+
+    private RoutingAlgorithm getRoutingAlgorithm()
+    {
+        return new DijkstraBidirectionRef(g, opts.getFlagEncoder(), this.weighting, opts.getTraversalMode());
+    }
+
+    public List<Path> calcRoundTrips()
+    {
+        List<Integer> points = TourWayPointGenerator.generateTour(from, locationIndex, edgeFilter, this.distanceInKm);
+        List<Path> pathList = new ArrayList<Path>(points.size() - 1);
+        RoutingAlgorithm routingAlgorithm;
+
+        for (int i = 1; i < points.size(); i++)
+        {
+            routingAlgorithm = this.getRoutingAlgorithm();
+            routingAlgorithm.setWeightLimit(weightLimit);
+            Path path = routingAlgorithm.calcPath(points.get(i - 1), points.get(i));
+            pathList.add(path);
+            this.numberOfVisitedNodes += routingAlgorithm.getVisitedNodes();
+            if (weighting instanceof UniquePathWeighting)
+                ((UniquePathWeighting) weighting).addPath(path);
+        }
+
+        return pathList;
+    }
+
+    @Override
+    public Path calcPath( int notNeeded, int notNeededAsWell )
+    {
+        throw new UnsupportedOperationException("Not supported.");
+    }
+
+    @Override
+    public List<Path> calcPaths( int notNeeded, int notNeededAsWell )
+    {
+        if (locationIndex == null)
+        {
+            throw new IllegalStateException("You have to call prepare before calculating Paths");
+        }
+        return calcRoundTrips();
+    }
+
+    @Override
+    public void setWeightLimit( double weightLimit )
+    {
+        this.weightLimit = weightLimit;
+    }
+
+    @Override
+    public String getName()
+    {
+        return AlgorithmOptions.ROUND_TRIP;
+    }
+
+    @Override
+    public int getVisitedNodes()
+    {
+        return this.numberOfVisitedNodes;
+    }
+
+}
--- a/core/src/main/java/com/graphhopper/routing/RoutingAlgorithmFactorySimple.java
+++ b/core/src/main/java/com/graphhopper/routing/RoutingAlgorithmFactorySimple.java
@@ -63,6 +63,10 @@ public class RoutingAlgorithmFactorySimple implements RoutingAlgorithmFactory
             altRouteAlgo.setMinPlateauFactor(opts.getHints().getDouble("alternative_route.min_plateau_factor", 0.2));
             altRouteAlgo.setMaxExplorationFactor(opts.getHints().getDouble("alternative_route.max_exploration_factor", 1));
             return altRouteAlgo;
+        } else if (AlgorithmOptions.ROUND_TRIP.equalsIgnoreCase(algoStr))
+        {
+            double distanceInKm = opts.getHints().getDouble("round_trip.distance", 1);
+            return new RoundTripAlgorithm(opts.getWeighting(), distanceInKm, g, opts);
         } else if (AlgorithmOptions.ROUND_TRIP_ALT.equalsIgnoreCase(algoStr))
         {
             RoundTripAltAlgorithm altRouteAlgo = new RoundTripAltAlgorithm(g, opts.getFlagEncoder(), opts.getWeighting(), opts.getTraversalMode());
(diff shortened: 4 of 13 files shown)


ratrun:
I tried this branch out because it sounds interesting. But it does not work for me. Whenever I add `&tour_length=100`, no route is found any longer.

boldtrn:
@ratrun haven't tested it for Austria. This might happen sometimes, but shouldn't happen regularly. What area did you use?

ratrun:
The first log line contains the coordinates and the exact URL. It looks as you can choose any. I tried at least 8 times with differnt start and end point and never got a result.

boldtrn:
@ratrun thanks for pointing out that there was a mistake. I fixed it and got rid of the `((LocationIndexTree) locationIndex).setMaxRegionSearch(100);`

Attached is a screenshot of the generated route (thats actually a good one - took me about 10 repetitions - but you should get a result every time). Currently edges are taken twice, but this should be fixed by #579.

screen shot 2015-11-29 at 14 13 17

ratrun:
Thank you. I can confirm that now I always get a result.
I would have a small suggestion for improvement (except for unique paths avoidance): It would be great if one could add another `iteration_attempts` parameter, which instructs how many routes get calculated and only the result with the lowest overall weighting result should be returned. This `iteration_attempts` parameter must be limited to a hard-coded value, e.g. 20 such that the computation does not take too long. Most probably @karussell will not want deploy this possibility to the graphhopper main server because of the load. But this would be an interesting feature for private instances.

karussell:
Thanks for the work. I would like to integrate #420 in the next days first and then adapt this solution to integrate as second alternative. The unique path calculation should be done in the weighting, where we have to improve the API to allow combination of weightings like suggested by @boldtrn. 

Furthermore we will definitely deploy this to the API at some point, even if it is load intense, also probably to the GraphHopper Maps demo. That will happen once we have the [flexibilty work improved and production ready](https://discuss.graphhopper.com/t/more-flexibility-in-gh/71).

karussell:
The parts to integrate this into GH are now mostly done: https://github.com/graphhopper/graphhopper/tree/alternate_route2 Would you mind to integrate this?

Currently you need to trigger alternative calculation via `&algorithm=alternativeroute&alternative_route.max_num=4` and something similar will be necessary for round trips. Later we can enable this via the UI somehow ...

karussell:
@boldtrn The alternative routing and 'round trip alt' algorithm are now integrated in the master. You can now easily integrate this algorithm (or any improved version you might have) via adding it to RoutingAlgorithmFactorySimple ... and 'roundtrip' is reserved for you. Feel free to throw away my approach which is currently slower and would need a quality boost to come near your results.

ratrun:
@boldtrn Do you plan to integrate your solution as suggested by Peter? I was thinking about if I should give it a try. But I'm sure it would be much easier for you and if you already started that would make sense at all.

boldtrn:
@ratrun Yes, this is planned. I might work a bit on this issue today. I hope it's quite straight forward. I wanted to integrate this in the next time.

ratrun:
Thanks. Great.

boldtrn:
@ratrun I updated my branch. Let me know what you think.

karussell:
Thanks @boldtrn that you took the time to reintegrate it!

> round_trip.distance

I would use meters as this is the same unit we use for the response 

We also need at least one unit and one integration test e.g. in GraphHopperIT for this round trip algo.

> GHPoint from, locationIndex

Can we avoid somehow the prepare? The 'lat,lon' stuff is meant for the higher level API and lower level should always use node or edge IDs. Why do you have only 'lat,lon' available when generating the random points? Or could you create the random points at the higher level layer instead? Especially for the unit tests and other integrations it is harder to handle 'lat,lon's instead of IDs.

Can we reduce the complexity a bit? E.g. remove TourStrategyFactory, merge the TourStrategy classes into one and moving TourWayPointGenerator methods into the RoundtripAlgo class? I would avoid introducing an abstraction layer if there is just one implementation or do you have already alternates for them (then it would be okay).

And is the method in CoordinateProjection maybe already covered from a DistanceCalcXY method? If not, I would put the method into an existing Helper class to avoid classes just for one method (?)

ratrun:
@boldtrn Thank you. This really works very well and the feature is great. I intend to use it for planning of cycle tours in spring. My tests from my location look promising. It chooses very similar cycle routes which I would take.
I have not looked into your implementation yet. 
Do you know a possibility to extend the algorithm and also somehow use a second point or alternatively a heading parameter as input in addition? It would be really great if one could influence the direction one intends to go. One explame is that I want to go to direction east. Currently the only possibility is to give it multiple tries and hope that the direction changes to the one you intend.

I'll try to implement a web interface for it based on my master, such that the usage becomes more simple.

@karussell : 
>>round_trip.distance

>I would use meters as this is the same unit we use for the response

I believe that kilometers are what users would expect, so this would be more user friendly. But as the end users will not manually mangle URI parameters, I don't care.

ratrun:
I merged your branch into my master, implemented the WEB UI and pushed the changes.

boldtrn:
@karussell 

> I would use meters as this is the same unit we use for the response

Yes sure, we can change this. I started with km since that's usually the unit the user is interested in.

> We also need at least one unit and one integration test e.g. in GraphHopperIT for this round trip algo.

Yes, you are right. I will add some tests.

> Can we avoid somehow the prepare? The 'lat,lon' stuff is meant for the higher level API and lower level should always use node or edge IDs. Why do you have only 'lat,lon' available when generating the random points? Or could you create the random points at the higher level layer instead? Especially for the unit tests and other integrations it is harder to handle 'lat,lon's instead of IDs.

For the round trip generation I use a technique that is called point projection. I use the starting point and project this point into a certain direction over a certain distance. Then I use the locationIndex to find the closest node. We could do it in the GraphHopper.java class but than we have to pass the generated points to the roundtrip algorithm via RoutingOptions. If we want to generate multiple roundtrips (for alternative paths), we need to generate all these points beforehand. What do you think, should where and how should we generate these points?

> Can we reduce the complexity a bit? E.g. remove TourStrategyFactory, merge the TourStrategy classes into one and moving TourWayPointGenerator methods into the RoundtripAlgo class? I would avoid introducing an abstraction layer if there is just one implementation or do you have already alternates for them (then it would be okay).

The current strategy is not the best, but with this layer adding new strategies is really easy and straight forward. I think we will see a lot of interesting strategies in the future. @ratrun just requested adding a second point. This could be easily done by just adding another strategy. In the StrategyFactory on could define what strategy should be picked according to the current situation (e.g. using the two-point strategy only if distance > 50km).

> And is the method in CoordinateProjection maybe already covered from a DistanceCalcXY method? If not, I would put the method into an existing Helper class to avoid classes just for one method (?)

I haven't found something like that. Sure, where would you move it? 

> Do you know a possibility to extend the algorithm and also somehow use a second point or alternatively a heading parameter as input in addition? 

@ratrun Yes both should be quite straight forward. Just create a new TourStrategy. Probably you have to extend everything a bit to set an initialBearing. But should be quite easy. Let me know if you need more details. I will add another helper method to the TourStrategy, which should make it easier to generate several points.

> I merged your branch into my master, implemented the WEB UI and pushed the changes.

Great news :) Thanks.

boldtrn:
I just committed the helper functions and some tests (not enough).

karussell:
> I haven't found something like that. Sure, where would you move it? 

I would put it into DistanceCalcEarth as the earth radius is also there (or could be also related to AngleCalc). Furthermore the `projectCoordinateRandomBearingAndModifiedDistance` is something which should be testable and I would put it as none-static method into the round tour class so unit tests can avoid randomness via creating a fixed seed or a fixed result and method overridign.

Hmmh, thinking about this a bit more: as we have heading stuff, maybe also 'bearing' should renamed to 'heading' and probably there is some code for this too. Will look into it.

> The current strategy is not the best, but with this layer adding new strategies is really easy and straight forward. I think we will see a lot of interesting strategies in the future.

Okay, I'm fine with it if there are already two strategies :) but it is overhead if we just plan for them as then it is nearly always the case that refactorings etc are necessary.

> For the round trip generation I use a technique that is called point projection. I use the starting point and project this point into a certain direction over a certain distance. Then I use the locationIndex to find the closest node. We could do it in the GraphHopper.java class but than we have to pass the generated points to the roundtrip algorithm via RoutingOptions. 

Okay, that sounds doable. Will have to think about it, because an algo setup which requires the location index is too heavyweight and should be 'somehow decoupled' ;). In the worst case instead of passing the locationindex I would provide a lookup hook.

karussell:
Kind of related a post on HN regarding abstractions vs. unhandleable mess https://news.ycombinator.com/item?id=11093733

karussell:
When thinking about this more ... 

we should move the point creation out of the algorithm and e.g. provide the created points as input for the algorithm or let clients explicitly define points or just one point and a heading etc. Also the randomness should move out of the algorithm if not even to the client itself e.g. providing a random heading direction. What do you think? Should I make a branch to show what I mean (when I find time ;))?

boldtrn:
@karussell Ok, I am excited. Moving the generated points as input to the algorithm is a good idea (even though the algorithm won't do very much right now if there are no points to generate). For a simple single point strategy generating a random heading by the user seems appropriate. I am not sure about more complex strategies, but I am excited about your branch :).

ratrun:
@boldtrn 
>Thank you for the advice. The extension with an initial bearing (=heading) was pretty easy and it works well. As entering of a distance value and bearing value in the GUI as numerical values is not so super intuitive, I'll try to try to implement a two points approach on the WEB UI and calculate the distance parameter and the bearing on the UI. 

Here is the basic workflow:
The user will need to enter a first point. Then or before he/she will be able to click on a simple button for round tours. Once pressed, the icon for the second point will change to a center symbol. When the user has entered an initial position of the second point, the JS code will compute the radius distance between the start point and the second point and the direction (=bearing). I intend to use Leaflet distanceTo and http://stackoverflow.com/questions/3932502/calcute-angle-between-two-latitude-longitude-points. The target tour distance gets simply computed with 2*PI*DistanceBetweenTheTwoPoints. At movement of the any of the two points the same thing will happen again. On a click on the center point of the tour, the center point of the tour gets randomly moved to a point on a circle with the start point as center of the circle, and the tour gets recalculated. In case if you have better ideas please let me know.

boldtrn:
@ratrun 

Thanks for explaining. That is actually a nice idea to work with the two points. I don't know if you have made any progress here?

I just recently added the roundtrip feature to a project of mine. There is use https://github.com/aterrien/jQuery-Knob to allow a user to set the bearing. A distance and starting point has to be entered additionally.

ratrun:
@boldtrn Thanks for the feedback on the GUI idea. I haven't implemented anything as this PR unfortunately is still not accepted and a numerical entry field of the bearing was good enough for me.

karussell:
Please have a look: https://github.com/graphhopper/graphhopper/compare/issue_582

 * merged master
 * removed the old round trip algo
 * made classes better pluggable and testable -> avoid statics a bit
 * avoid randomness via seeds
 * avoid a bit casting
 * use meters instead km
 * improved integration test, added unit test
 * renamed a few things

Please question naming and everything. This is for now a rough sketch of what I meant. Still missing is the part where one calculates the initial heading from the second point.

Also I would somehow use the `TourPointGenerator` which has more control over the whole process instead of the `TourStrategy` but for now both is still used.

boldtrn:
@karussell 

Thanks. Looks nice. Unfortunately I cannot comment in the code, so I'll just add my comments here.

`// TODO use the second point to calculate the initial direction`
Actually I am not sure if we should really do it that way. I experimented with a knob and had quite good feedback. See the Roundtour generation at [Kurviger.de](https://kurviger.de/).

`Object algoControl = null;`
Would be great if we could create a defined class here. Currently it is only used for the `TourPointGenerator` right?

`long seed = request.getHints().getLong("round_trip.seed", 0L);`
Great idea to pass it in the request!

`Rates already used Paths worse.`
Javadoc needs an update :)

Something I personally would love to improve is the `RoundTripPointGenerator`. There is to much randomness when generating the points in my opinion. it would be great if there would be a more intelligent solution. When the street network density is low (e.g. in the Alps), there are situations were this approach performs not very well. That might be a different ticket, though.

I am still a fan of the strategy, since this makes the whole approach customizable.

karussell:
> Thanks. Looks nice. Unfortunately I cannot comment in the code, so I'll just add my comments here.

Uhm, ugly. And I cannot push and add to your commits so I've created the branch ... probably I should have opened yet another PR to make it the 'github way'

> Actually I am not sure if we should really do it that way. I experimented with a knob and had quite good feedback. 

From the API usage I prefer the "point&heading" parameters but the current system prefers "2 points" as with 1 point we get a bit problems with the QueryResult. Still we can workaround this with duplicating the first QueryResult as the second... I'll do that and try to avoid the currently required second point and also we should try to read the (first) `heading`

> See the Roundtour generation at Kurviger.de.

:+1: I really like that :)

>  Would be great if we could create a defined class here. Currently it is only used for the TourPointGenerator right?

Yes. That was necessary to avoid the special cast. 

A better approach to put it into the hints but this is only usable for string values :(. And that is by intention as I wanted all 'real objects' being explicit in the AlgorithmOptions or being later created from the strings. Hmmh. Now the Object serve the generic purpose and could be used from any algo. Not sure how to solve this better, a specific class would not make sense too.

> Something I personally would love to improve is the RoundTripPointGenerator

You want to improve the implementation details? That should indeed go into another issue :)
For now I would be happy if we can get this merged as a first step :)

karussell:
Regarding the 'algoControl': what if we calculate the points and assign the new points and do the regular routing, where we just need to intercept the weighting change somehow. Maybe with a custom RoutingAlgorithmFactory? I mean this part [here](https://github.com/graphhopper/graphhopper/compare/issue_582#diff-6a0717ac71575944279a2e916b7b6b85R72) (except the AvoidPathWeighting) is duplicate to the loop in GraphHopper, and if we can somehow merge this into the normal process we do not need a new roundtrip class, maybe just a factory and the point creation strategy. A further benefit would be that the used algorithm would stay customizable and not fixed to DijkstraBidirectionRef

boldtrn:
> Uhm, ugly. And I cannot push and add to your commits so I've created the branch ... probably I should have opened yet another PR to make it the 'github way'

Yes, probably you should have created a PR to my PR :D.

> Still we can workaround this with duplicating the first QueryResult as the second

Yes, could be a nice workaround.

> A better approach to put it into the hints but this is only usable for string values :(. And that is by intention as I wanted all 'real objects' being explicit in the AlgorithmOptions or being later created from the strings. 

Could we maybe create an empty Interface called `AlgorithmControl` (or something like that) to specify what type of object to put in. I am actually not even sure if it's allowed to create empty interfaces, but it would improve the readability for the user.

> I mean this part here (except the AvoidPathWeighting) is duplicate to the loop in GraphHopper, and if we can somehow merge this into the normal process we do not need a new roundtrip class, maybe just a factory and the point creation strategy.

Yes you are right. I think I it copied from the GraphHopper class ;). I did not do this for two reasons.
a) We have to add the AvoidPathWeighting to the GraphHopper class, but it would be only used for the roundtrips
b) With random roundtrips you regularly end up with bad routes. Therefore, it is better to calculate several routes and select the best one. We might solve that using the alt paths approach and offer the user for example 3 different routes for one round tour request. But still we need to change the generated points with each iteration, which again changes the core routing functionality of the GraphHopper class.

I would love to implement it in the GraphHopper class, but I thought you would not like it.

karussell:
> I am actually not even sure if it's allowed to create empty interfaces, but it would improve the readability for the user.

Good idea. And yes, that is allowed and called 'marker interface'

> I would love to implement it in the GraphHopper class, but I thought you would not like it.

I would not convolute the GraphHopper with round trip specific stuff, that is right. But if we could find a way to abstract this away and bundle it in a round trip algo factory or something that this would be nice. But maybe also part of another PR :)

karussell:
I've applied the suggestions to the branch. Please have a look.

boldtrn:
> Good idea. And yes, that is allowed and called 'marker interface'

Thanks I didn't know that.

> I've applied the suggestions to the branch. Please have a look.

Looks nice, but tests are failing?

> I would not convolute the GraphHopper with round trip specific stuff, that is right. But if we could find a way to abstract this away and bundle it in a round trip algo factory or something that this would be nice. But maybe also part of another PR :)

Mhm, if we want to do that, I think we should do it now.
I think there is no way to avoid changes to the GraphHopper class. At least we have to add the AvoidPathWeighting to the path calculation, and update it with the used paths. Another issue is how to calculate multiple paths and how to find the best one.

karussell:
> Looks nice, but tests are failing?

Fixed

> Mhm, if we want to do that, I think we should do it now.

Okay, sure. I'll have a look.

karussell:
The route method has kind of 3 parts: 1. point lookup, 2. route loop and 3. path merger.

Now if I understood this correctly for the round trip we need: 1. create points 2. route loop with adjusted weighting and 3. path merger with potentially retry at 1?

boldtrn:
> Now if I understood this correctly for the round trip we need:

Exactly.

karussell:
I had a bit success in moving 2 and 3 out into another class, but then for a new 'round trip implementation' we would need to copy and paste most of it. Hmmh, tricky

neiljp:
@boldtrn I had a look at your kurvinger.de, since I was interested in the roundtrip work. I'm not fluent in german, but this is for motorbikes? Is the intention to generate round-trips which are of approximately the length (double-headed arrow) in km? I ask since I've used it in some cases and got ones which are eg. 22km instead of "3", though some generated routes were about 3... should it do that? :)
(the use-case I looked at was based on starting at a rail station, since there have been other tools I looked at in the past where people travelled by rail and then did a loop from there - not sure that applies to a motorbike, though ;))

boldtrn:
@karussell Yes it's really tricky.

@neiljp Yes this is true. The algorithm works best for distances greater than 50km. The usual input is at around 200km. Motorcyclists often ride round tours as day trips. So the intention was to easily generate round trips.

neiljp:
@boldtrn Ah, shame it doesn't work well for shorter, as it'd be great to have time/distance-guided routes for walking/cycling too :)

boldtrn:
@neiljp I optimized the algorithms at [Kurviger](https://kurviger.de/) for routes with more than 50km. Maybe you are able to create a nice algorithm for short distances. Feel free to contribute it.

karussell:
I've created #709 which integrates this PR in a clean fashion (at least it seems so ;)). If someone can have a look at it I would merge or improve it.

@neiljp the current roundtrip algorithm from @boldtrn seems to work for every vehicle rather good.

neiljp:
@karussell That's great news - I look forward to seeing it in action :)

karussell:
> @karussell That's great news - I look forward to seeing it in action :)

You can easily check the branch #709 out - we would love to get feedback as early as possible :)

Also we do not yet have a UI for this, maybe we need something similar like @boldtrn did on kurviger, also @ratrun had something if I remember correctly.

PR #584, opened 2015-11-30
Intelli j profile

boldtrn:
Here is my current formatting profile for IntelliJ. Probably not everything matches exactly, but better than the default intelliJ settings.


karussell:
Thanks this will be a nice help for other intelliJer :)

Regarding 'style': can you try to create a separate branch for every feature (and PR) so that the other changes (like for the unique path) do not show up here :) ? And we can merge and discuss them separately.

boldtrn:
Sorry, was my bad. I created a separate branch, but forgot to checkout the master first.

PR #586, opened 2015-11-30
Unique Paths

boldtrn:
Ok here it is, finally. Needed to reset my remote repository. Just saw that my motorcycle curvature elevation PR is gone, due to that. I will add this PR again, as well.


boldtrn:
I guess we can close this PR. The unique paths are added to the round tours by now. So this shouldn't be needed anymore.

PR #587, opened 2015-11-30
Added Elevation Data to Motorcycle Flag Encoder

boldtrn:
Readded #571. 

This time I haven't changed the IntegrationTests, therefore they fail. I still think that this issue occurs because the speed limit for some minor roads in Monaco is not set in the monaco test file. With the current monaco file everything works as excpected.


boldtrn:
Had a more detailed look into that issue. Seems that the fastest weighting prefers small and actually slow streets because most probably no speed limit is set on these streets. I changed the default speed to favour typical rural roads. Right now there are two possibilities in my opinion:

* Change the test data and accept a rather bad route
* Create a motorcycle curvature encoder

boldtrn:
Ok I reverted the speed changes. This is actually not part of this PR and the elevation changes should not be blocked by speed changes. Therefore, I reverted the changes.

We should consider, creating an own encoder (or profile) for the curvature stuff to define different speed limits.

@karussell Could you have a look at the still failing tests. This is due to the `OSMTagParserTest`. It parses `P2M` as `5356800` but expects it to be smaller than `5274000`.

karussell:
@boldtrn thanks! I've opened #588

boldtrn:
I tried to merge the current master because the tests were failing and it is fixed in the current master. I guess the history is not really user-friendly due to that. I will close this PR and add the most important changes to this branch later.

PR #596, opened 2015-12-07
Refactoring necessary for alternative routes

karussell:
This refactoring prepares for a proper integration of alternative routes introduced in #420. This is a breaking change even for the higher level API (GraphHopper class) as we need multiple routes in GHResponse which is currently not possible, but luckily this change does not change the JSON layout! And as the GraphHopperWebTest shows the new code still parses old JSON and vice versa should be possible too (not tested until now)

This change does:

 * GHResponse now wraps multiple AltResponse 
 * renames GraphHopper.getPaths to calcPaths as 'get' sounds too cheap
 * a new method `RoutingAlgorithm.calcPaths` is added returning `List<Path>`
 * minor modifications to the RouteSerializer to return multiple paths if available
 * GHResponse.getErrors returns now own errors as well as errors from contained AltResponse entries and also returns an error if no alternative is added which is nice. BUT we needed to introduce a new method `hasRawErrors` to directly check the underlying error list and avoid adding other errors for some circumstances. Not sure if this is the best approach but led to backward compatibility for `ghResponse.hasErrors` as well as to fewer code.

**Please review and test for your integration for any glitches!**


devemux86:
:+1: 
Interesting concept.

boldtrn:
:+1: Nice, I have to test this in the next days.

karussell:
Okay, and I'll try to make #420 working with this. It is not perfect but will probably polish it while including 420

karussell:
This refactoring is successfully used with the alternative route calculation: https://github.com/graphhopper/graphhopper/tree/alternate_route2 Currently you need to trigger alternative calculation via `&algorithm=alternativeroute&alternative_route.max_num=4`

boldtrn:
@karussell Would you mind if we refactor the calcPaths method in terms of length? Break it down into smaller methods. The method is now approximately 200 lines long. It becomes really hard to focus on the essentials of that function.

karussell:
Closing due to #640

Issue #597, opened 2015-12-07
Bike routing penalty for flyover/overpass bridge


nemobis:
I find GraphHopper almost magic for bike routing and I'm recommending it to everyone since a few months, thanks a lot.

I've been fighting a bit with the directions at https://graphhopper.com/maps/?point=45.511279%2C9.254265&point=45.517805%2C9.266946&vehicle=bike&weighting=fastest&elevation=true&layer=TF+Cycle , IMHO blatantly wrong because they ignore a provincial cycling route (via Idro on naviglio Martesana) and suggest instead to go in the middle of a motorway junction (via Padova at Cascina Gobba, with a cavalcavia). The desired routing is: https://graphhopper.com/maps/?point=45.511279%2C9.254265&point=45.517805%2C9.266946&vehicle=bike&weighting=fastest&elevation=true&layer=TF+Cycle

![gobba-cavalcavia](https://cloud.githubusercontent.com/assets/901528/11639499/89a88154-9d2d-11e5-9895-3d2d6743d7f1.png)

![gobba-bici](https://cloud.githubusercontent.com/assets/901528/11639503/8db008b2-9d2d-11e5-829e-a8cee69606e3.png)

Multiple possible errors or deficiencies are discussed at https://lists.openstreetmap.org/pipermail/talk-it/2015-December/thread.html#50340 ; I hope the matter is not with the alleged 2 meter saving in elevation or with the alleged 90 meters in distance saved.

One thing I think should be considered is that the bridge, especially being over a motorway, is moderately steep; currently bridges seem to have no weight in that sense: https://github.com/graphhopper/graphhopper/blob/fed03547acea0b0edbbb90dc723b944250463ef9/core/src/main/java/com/graphhopper/routing/util/Bike2WeightFlagEncoder.java#L161
I think you might want to estimate a steepness, or a slowdown, depending factors such as length of the bridge and kind of underlying obstacle.

As far as I can see, the data is correct: https://wiki.openstreetmap.org/wiki/Key:ele , https://wiki.openstreetmap.org/wiki/Key:bridge , https://wiki.openstreetmap.org/wiki/Key:height don't mention any further detail on bridge steepness being recommended or even allowed ; I asked at https://wiki.openstreetmap.org/wiki/Talk:Key:height#Bridge_steepness just in case.

ratrun:
[...]

Issue #598, opened 2015-12-07
Bike routing should prefer roads with 30 km/h or 15 km/h speed limit


nemobis:
Same issue as https://github.com/graphhopper/graphhopper/issues/597 , please see for my test case.

30 km/h zones and speed limits are increasingly used to create ways that, while still allowed for motor vehicles and mixed usage, are recommended for bicycles:
* http://www.ecf.com/advocary/mobility/traffic-calming30-kh/
* http://www.copenhagenize.com/2010/01/30-kmh-zones-work.html
A 15 km/h limit is more extreme and can be taken as a strong indication of a road being practically forbidden to cars, almost equivalent to a cycleway. (An even lower limit on an extended way would practically be like a pedestrian area, I guess; I'm not sure if that's in use.)

To the contrary, roads which lead to motorway junctions may theoretically have a 50 km/h speed limit, but, when the limit becomes (or was) 90 km/h at a few meters distance, drivers are unlikely to respect the lower limit.

Concretely, I think that in addition to setAvoidSpeedLimit() https://github.com/graphhopper/graphhopper/blob/fed03547acea0b0edbbb90dc723b944250463ef9/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java#L201 there should be:
* a setPreferSpeedLimit() set at 15 for bike and MTB, perhaps 30 for racing bikes, so that via Idro in the example is chosen over normal roads (as long as the surface or type is reasonable);
* a setting to also avoid or discourage getting 50-100 m far or closer from a way whose speed limit is above the avoided one; or alternatively, to discourage going over ways which in 50-100 more meters turn into highway=motorway_junction, highway=primary,  highway=motorway, highway=motorway_link or similar.

ratrun:
First we need to fix the discovered maxspeed car specific code problems in #597, then we can have a look if further perference of low maxspeed limits is needed and usefull.

nemobis:
Sure, thank you.

karussell:
@ratrun is this here fixed due to your fix in #600 or will it need further changes?

ratrun:
#600 is only one of three changes.

A second change is necessary to boost the influence of the internally assigned preference. 

A third and most complex change is needed for this:

[...]

PR #600, opened 2015-12-08
Compensate loss of road type speed by low maxspeed values via increase of preference

ratrun:
Bicycle related change: Compensate the loss of road type speed by increasing the priority preference in case that a low maxspeed tag limits the speed below the guessed road type speed. 
This change was triggered by issue #597 and #597. See discussion in #600.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
@@ -791,16 +791,18 @@ public abstract class AbstractFlagEncoder implements FlagEncoder, TurnCostEncode
     }
 
     /**
-     * @param force should be false if speed should be changed only if it is bigger than maxspeed.
+     * @param way: needed to retrieve OSM tags
+     * @param speed: speed guessed e.g. from the road type or other tags
+     * @return The assumed speed. 
      */
-    protected double applyMaxSpeed( OSMWay way, double speed, boolean force )
+    protected double applyMaxSpeed( OSMWay way, double speed )
     {
         double maxSpeed = getMaxSpeed(way);
-        // apply only if smaller maxSpeed
+        // We obay speed limits
         if (maxSpeed >= 0)
         {
-            if (force || maxSpeed < speed)
-                return maxSpeed * 0.9;
+            // We assume that the average speed is 90% of the allowed maximum
+            return maxSpeed * 0.9;
         }
         return speed;
     }
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -319,6 +319,30 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
         return oldRelationFlags;
     }
 
+     /**
+     * Apply maxspeed: In contrast to the implementation of the AbstractFlagEncoder, 
+     * we assume that we can reach the maxspeed for bicycles in case that the road 
+     * type speed is higher and not just only 90%.
+     * <p>
+     * @param way: needed to retrieve OSM tags
+     * @param speed: speed guessed e.g. from the road type or other tags
+     * @return The assumed avererage speed.
+     */
+    @Override
+    protected double applyMaxSpeed( OSMWay way, double speed)
+    {
+        double maxSpeed = getMaxSpeed(way);
+        if (maxSpeed >= 0)
+        {
+            // We strictly obay speed limits, see #600
+            if ( maxSpeed < speed )
+            {
+                    return maxSpeed;
+            }
+        }
+        return speed;
+    }
+    
     @Override
     public long handleWayTags( OSMWay way, long allowed, long relationFlags )
     {
@@ -326,13 +350,11 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
             return 0;
 
         long encoded = 0;
+        double wayTypeSpeed = getSpeed(way);
         if (!isFerry(allowed))
         {
-            double speed = getSpeed(way);
-
-            // bike maxspeed handling is different from car as we don't increase speed
-            speed = applyMaxSpeed(way, speed, false);
-            encoded = handleSpeed(way, speed, encoded);
+            wayTypeSpeed = applyMaxSpeed(way, wayTypeSpeed);
+            encoded = handleSpeed(way, wayTypeSpeed, encoded);
             encoded = handleBikeRelated(way, encoded, relationFlags > UNCHANGED.getValue());
 
             boolean isRoundabout = way.hasTag("junction", "roundabout");
@@ -353,7 +375,7 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
         if (relationFlags != 0)
             priorityFromRelation = (int) relationCodeEncoder.getValue(relationFlags);
 
-        encoded = priorityWayEncoder.setValue(encoded, handlePriority(way, priorityFromRelation));
+        encoded = priorityWayEncoder.setValue(encoded, handlePriority(way, wayTypeSpeed, priorityFromRelation));
         return encoded;
     }
 
@@ -469,7 +491,7 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
      * <p>
      * @return new priority based on priorityFromRelation and on the tags in OSMWay.
      */
-    protected int handlePriority( OSMWay way, int priorityFromRelation )
+    protected int handlePriority( OSMWay way, double wayTypeSpeed, int priorityFromRelation )
     {
         TreeMap<Double, Integer> weightToPrioMap = new TreeMap<Double, Integer>();
         if (priorityFromRelation == 0)
@@ -477,7 +499,7 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
         else
             weightToPrioMap.put(110d, priorityFromRelation);
 
-        collect(way, weightToPrioMap);
+        collect(way, wayTypeSpeed, weightToPrioMap);
 
         // pick priority with biggest order value
         return weightToPrioMap.lastEntry().getValue();
@@ -520,7 +542,7 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
      * @param weightToPrioMap associate a weight with every priority. This sorted map allows
      * subclasses to 'insert' more important priorities as well as overwrite determined priorities.
      */
-    void collect( OSMWay way, TreeMap<Double, Integer> weightToPrioMap )
+    void collect( OSMWay way, double wayTypeSpeed, TreeMap<Double, Integer> weightToPrioMap )
     {
         String service = way.getTag("service");
         String highway = way.getTag("highway");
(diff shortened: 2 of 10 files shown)


karussell:
This PR changes the factor 0.9 to 1.0 for bike? But is this really the (only) cause of this behaviour as the official bike route should be preferred and could weight out the speed disadvantage.

karussell:
I'm not sure about this comment:

> @return The assumed speed. If the maxspeed tag is available then normally it decreases the speed guessed from the road type (if maxspeed < speed).
> * We allow us to go maxSpeedIgnoranceFactor faster compared to the speed limit, see issue #597

The maxSpeedIgnoranceFactor is always 1 and it will never increase the speed as indicated with the second comment or what do you mean here?

ratrun:
The problem with the 0.9 factor is that the derived road type speed does not get decreased, therefore I removed it for the bikes.
The `maxSpeedIgnoranceFactor` is for discussion. What value do you suggest? In case that we set it to 1.0, then I'll completey remove it. But maybe we should allow at least 10% tolerance up with value 1.1, or even 30% ? The reason is that I haven't discovered any radar for usualy bike speed values around 20 km/h.

ratrun:
I forgot to ask about the reasoning for the 0.9 maxspeed factor for bikes. I suppose that it is assumed that one usually cannot reach the maxspeed value because of acceleration phases... This strange factor should get explained as well. I also was asking myself why the road type speed is used in comparison without any factor.

karussell:
> The problem with the 0.9 factor is that the derived road type speed
> does not get decreased, therefore I removed it for the bikes.

But if e.g. maxspeed=15 then it will cap higher speeds to this value?

> But maybe we should allow at least 10% tolerance up with value 1.1, or even 30% ?

Ah, okay - now I understand the intention of this value better, thanks. 20 or 30% more than allowed do not sound unrealistic up to 30km/h IMO. But what was again the reason we want and should violate the speed sign :) ?

> I also was asking myself why the road type speed is used in comparison without any factor.

 * maxspeed * 0.9 => average speed
 * guessed speed values => already average speed

ratrun:
> But what was again the reason we want and should violate the speed sign :) ?

Let's assume two nearly parallel alternative ways of `highway=residential`. One of it is designated as a cycle route and in order to make it safe for cyclists, the allowed speed is maximal 15km/h. This dedicated cycle route segment is 5% longer compared to the residential alternative.
Our road type speed on the none limited way is the 18km/h default for highway=residential. Issue #597 showed that in this case the none dedicated way is chosen, even if the cycle route is of network "rcn" and therefore classified as VERY_NICE. The main reason is that our fastest weighting limits the speed to the 15km/h limit and consequently the priority boosting starts from a bigger weight. The other reason is the priority boosting is not strong enough. But this sub-problem gets addressed in a separate PR.

Please re-check the modifications in 6995423 . Here I included your suggestions from above. From my perspective the PR is ready for merge now.

karussell:
> the allowed speed is maximal 15km/h

You mean the average speed is then used as a limit or is there a maxspeed=15 tag on the road somewhere? Hmmh, if there is really a sign on the road I do not really like the idea of ignoring it, even for cyclists. As e.g. then this road could get unsafer for pedestrians if we suggest racing through it.

Probably boosting the priority should be improved like you also suggested - but you mean that this was not sufficient?

ratrun:
>Hmmh, if there is really a sign on the road I do not really like the idea of ignoring it, even for cyclists.

In my experiments I came to the conclusion that priority boosting is not sufficient. The argument is also in contradiction to you first feedback which was  

>20 or 30% more than allowed do not sound unrealistic up to 30km/h IMO

Due to this statement I assiged the maxSpeedIgnoranceFactor to 1.3. But I'm also fine with 1.2 or 1.15. I only wanted to discuss it here. 

Regarding:

>As e.g. then this road could get unsafer for pedestrians if we suggest racing through it.

This argument not valid IMHO. We are not really suggesting to race through it. We only relax the assumed limit due to the weighting dilemma identified above. Rideres are responible for respecting the limits themselve.

I have now pulled this change and the announce prio boosting change to https://github.com/ratrun/graphhopper/tree/master . It allows to play with two sliders, where one can dynamically choose the preference of a fast / nice route an a steep/flat route. This feature is really nice for testing and finding of alternatives.

karussell:
> The argument is also in contradiction to you first feedback

Yes. I was thinking more about it and about our goals and I changed my perception of this issue, of course, even now, it is not set in stone. But we should not violate signs, there has to be other ways (I hope). Is the maxspeed sign really there?

ratrun:
I would be interested about your fears that this is against our goals. I can't follow your thoughts in this aspect. If you think about the estimation of the travel time, then I'm convinced that one can neglect the influence of this change on it.

And if it is your argument that we should not violate signs, then I believe that I gave reasons above that is not very useful in this case without leaving room of tolerance.

>Is the maxspeed sign really there?

I can't tell if the maxspeed sign is really correctly mapped in issue 597. The answer to this question is also not really relevant. Or ? The observed region is one of the best I have discovered so far in OSM when it comes to bicycle relevant tagging, so I'm very convinced that it is correctly mapped. 

>there has to be other ways (I hope) 

Well, I'm sure there are a many. One would be to generally reduce the assumed way type speed values. But this seems to be a wrong approach to me. All others, which come to my mind, are even worse.

I have re-tested #597 once again with different values for the `maxSpeedIgnoranceFactor` and priority boosting factor values. Due to a recent slight improvement in the boosting implementation my statement above that priority boosting only is not sufficient for the expected result in #597 is not true any longer.  In my current implementation at https://github.com/ratrun/graphhopper/tree/master a `maxSpeedIgnoranceFactor=1.0` and `niceLevel=2` also leads to the expected result. 

I believe that as we currently assume 18km/h as average cycling speed under good conditions and due to the fact that there are regions where the general speed limit gets reduced to 15km/h in order to better attract cyclists, a maxSpeedIgnoranceFactor of at least 1.1 is the best compromise. This suggested solution also covers cases, where the speed limit reaches 10km/h or lower, where an average cyclist really needs to dramatically reduce its speed.

karussell:
> I would be interested about your fears that this is against our goals. I can't follow your thoughts in this aspect.

Okay :)

We have 'speed' and we have 'priority' and we should keep this separate. My second fear is that we can loose trust of our users. GraphHopper can and will be used more and more also as a navigation tool (locus add-on, pocket maps, cruiser, ...) then people can and will recognize that we estimate higher than allowed speeds for the average speed, even if it is just 10%. And if it is just 10% I think we should try to reach the goal via some different approach like with the 'priority'.

Maybe the real underlying problem is the that we store "too rough" speed values? And we should use a speedFactor of 1 (or even 0.5?) and the necessary speedBits (5 or 6)?

> I have re-tested #597 once again with different values for the maxSpeedIgnoranceFactor

Thanks again for your time - I really appreciate your time here and this discussion! Again: I'm not sure how the end solution will or should look like but we should try hard to make routing of GraphHopper simple, logical, legal and of high quality ... somehow ;)

> This suggested solution also covers cases, where the speed limit reaches 10km/h or lower

Can't we keep the speed and raise the priority? ... but probably you tried that and we would need higher values for the priority?

ratrun:
> then people can and will precognize that we estimate higher than allowed speeds for the average speed, even if it is just 10%.

Eventually yes, although I wouldn’t expect any blame from one of the addressed users. But you are  right that there could come offences from others.

> Maybe the real underlying problem is the that we store "too rough" speed values? 

Yes indeed, in the observed case it plays a little role that internally a speed value of 15km/h gets rounded down to 14km/h. I remember it as a kind of design decision not to use odd values in for the road type speed assumptions. If we increased the amount of speed bits, we could avoid that small error. But I don't think that this would help a lot.

> I think we should try to reach the goal via some different approach like with the 'priority'. 

Yes and no. Let me try to explain: In the observed case a priority boosting solution only would work because the cycle route along the river is tagged as route with network=rcn. This we classify as `VERY_NICE`. I  prefer a solution which does not rely that internally we assign a priority value > `UNCHANGED` based on the road classification. 

> Can't we keep the speed and raise the priority? 

Yes we could do that. We could compensate the loss of speed in case that the tagged 'maxspeed' limits our assumed average road speed by automatically assigning e.g. one step higher priority value for compensation. Such code would get implemented around https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java#L535  

I didn't try this solution because you once gave feedback in another discussion about avoidSpeedLimit. If I remember correctly you didn't want the mapping from 'maxspeed' to priority not to become too complicated. Therefore I expected this argument again. It is also not that easy to decide how to implement that approach best. The questions one needs to answer is what influence the prioritisation we have calculated so far should have. The easiest way would be just to increase the priority by one step value to the better. In the reported case we therefore would assign BEST. If you prefer it this way, I'll update the PR.

karussell:
> I remember it as a kind of design decision not to use odd values in for the road type speed assumptions

It is just a result of the limited bits.

> I prefer a solution which does not rely that internally we assign a priority value > UNCHANGED based on the road classification. 

What do you mean here?

>  you didn't want the mapping from 'maxspeed' to priority not to become too complicated. 

I want to keep things simple. I do generally tend to try making things working via simple implementation and if that is not possible of course am okay with more complex solutions.

Do you think that the issue with two values for avoidSpeedLimit is related and that we even could fix this issue here too with that? Having two issues indicating the same, more complex solution would be enough and then this should be done this way.

> It is also not that easy to decide how to implement that approach best. 

IMO there is no 'best' solution. But we should aim for the 'best possible right now' :) !

> In the reported case we therefore would assign BEST. If you prefer it this way, I'll update the PR.

If the avoidSpeedLimit issue is unrelated or would not fix this issue here, then I would appreciate this here.

ratrun:
> Do you think that the issue with two values for avoidSpeedLimit is related and that we even could fix this issue here too with that? Having two issues indicating the same, more complex solution would be enough and then this should be done this way.

No, this discussion covered unrelated cases, where I wanted to classify the tollerance for high maxspeeds in two categrories and not just one.

> IMO there is no 'best' solution. But we should aim for the 'best possible right now' :) ! 

In this case I don't see anything which might change based on which we could find a better solution in the future.

> If the avoidSpeedLimit issue is unrelated or would not fix this issue here, then I would appreciate this here.

I'll update the PR soon.

ratrun:
The proposed change is implemented now. But as expected #597 is not solved. Further boosting of the priority level is needed additionally. I changed my mind and will add this change to this same PR instead of creating a separate new one.

karussell:
Thanks!

So this PR only increases the priority if the maxSpeed is lower then the guessed speed?
Then we should change the PR title ;)

BTW: removing the 'boolean force' parameter seems like a good idea

> I changed my mind and will add this change to this same PR instead of creating a separate new one.

which PR you mean here?

ratrun:
> BTW: removing the 'boolean force' parameter seems like a good idea 

I already forgot about that.

> which PR you mean here?

Now I remember the reason why I wanted to create a separate PR for increasing of the boosting via the priority influence. The problem is that this breaks a lot of testcases. I have now commited it in such a way that the intention becomes clear: 0f1eb4dda013660dae5ef84d7cac739a41738095 withoput breaking any tests.
Do you want to discuss this in a separate PR? In this case, I will rename this PR and also address #598.

karussell:
Yes, let us do this in another PR.

> I will rename this PR and also address #598 

If code changes are necessary, please create a new PR.

I can now merge your changes except the last commit? Or do you want to squash everything into one commit which should be merged?

ratrun:
Now I combined everything except for the last change into a single commit.

karussell:
Thanks :) !!

Issue #450, opened 2015-06-21
Create separate source module for import and core


karussell:
The whole OSM reading stuff should be in a separate source module 'import-osm' which needs a bit refactoring so that we can finally do

```java
hopper = new GraphHopperOSM().setOSMFile(...)
hopper.importOrLoad()
```

and on Android etc we should be able to avoid setting OSM related stuff and just do:
```java
hopper = new GraphHopper();
hopper.load(ghFolder)
```

We should use the change to separate OSM better from the core. Furthermore we could easier integrate other readers like #277 and #616.

devemux86:
You mean like split into modules for graph creation, perform routing etc (like the current ```tools```) ?

karussell:
Yes. Just moving the import code out of `core`, so that Android and iOS does not need to exclude those xml reading stuff. Or if Android at some point implements its, it can just add it as dependency

devemux86:
:+1: 
In mobile platforms the import (aka creation) of graphs is a bit difficult due to devices specifications.

devemux86:
@karussell nice work!

PR #602, opened 2015-12-13
Alternate route2

boldtrn:
This is a first idea of how to refactor the calcPaths method. The tests are still failing because I changed the behavior using Exceptions at some point. But I guess this can be fixed with little effort.


karussell:
What is the reasoning behind this refactoring?

The exception handling should not change as I would like to be able to indicate multiple problems with the request e.g. if one or more location cannot be found.

karussell:
Ups, I see that you mention the reasoning in another comment :)

> @karussell Would you mind if we refactor the calcPaths method in terms of length? Break it down into smaller methods. The method is now approximately 200 lines long. It becomes really hard to focus on the essentials of that function.

boldtrn:
I will revert the exception handling. I ran out of time but wanted to show the approach and discuss, if we want to proceed with this. Instead of using a method. The inner class approach seems to be a good idea to me.

karussell:
I'll have a look into it. In general I do not like large methods but I also do not like premature optimization - i.e. if there is no problem then there is no problem ;)

boldtrn:
Maybe I was over-engineering, but in my opinion the calcPaths method is too long. It is easy to get lost and it's really hard to get the essentials of that functions. It should be split up into several smaller methods in my opinion.

karussell:
I see you argument and although 150 lines is kind of 'big' it is also acceptable at the moment.

Mainly the method is a 3 step process and does the job of feeding the low level API: 
 * do location lookups (QueryResult), create QueryGraph and Weighting
 * calculate single paths
 * merge paths if necessary

Instead of artifically trying to separate methods I would propose a different refactoring where we try to split:

 * setup (Config)
 * import, CH preparation (Writing) and
 * loading, querying (Reading) 

into own classes/methods and if still necessary tie them together via the GraphHopper class. And we should do this after #450 is completed.

boldtrn:
Ok let's close this PR.

> 
* do location lookups (QueryResult), create QueryGraph and Weighting
* calculate single paths
* merge paths if necessary

In my opinion we should separate the calcPath method into these three methods in the future.

Splitting the GraphHopper.java class into 3 separate Classes that are wrapped by the GraphHopper.java class is actually another refactoring, that is really important. A class should have one purpose.

karussell:
> In my opinion we should separate the calcPath method into these three methods in the future.

Okay. Then this refactoring can be done also on its own (be it before the split or after)

> A class should have one purpose.

Yes, definitely. And one has to define what 'one purpose' is, so a bit subjective. Until now it was 'make some routes' ;)

PR #612, opened 2015-12-17
Added GHInput Import

boldtrn:
The Add-Button was not working, since GHInput was not found.


karussell:
Oh, Thanks! We need some kind of tests here :(

karussell:
Some experience with selenium :) ?

boldtrn:
Yes a little, but functional testing of JS is really hard. I'd always wanted to create functional tests using behat and mink http://mink.behat.org/en/latest/. But never tried it.

PR #615, opened 2015-12-20
Change default precision from 1 m to 10 m for distance label

fbonzon:
Unless you are looking at directions for a very short distance, say total distance about < 1 km (which is not common), moving the mouse on the elevation profile by just 1 pixel will never make only the 3rd decimal digit change. 2nd decimal digit changes too. Thus 3rd digit is not significant and can be removed.

I'm hesitating about 2nd digit. It could be removed too in most cases. With directions about > 10 km long, the 2 digit is not significant when mousing over the elevation profile.

#### Other minor changes

- ForceAxisBounds = false is default anyway
- Keeping "yAxisMin: 0" alone is useless unless "forceAxisBounds = true"
- Add missing exports
- Remove trailing space

### DIFF
--- a/web/src/main/webapp/css/style.css
+++ b/web/src/main/webapp/css/style.css
@@ -1,6 +1,6 @@
 body {
     color: #000;
-    font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;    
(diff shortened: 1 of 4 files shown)


karussell:
Thanks! Also for the other minor fixes!

Issue #277, opened 2014-10-29
DataReader for Ordnance Survey data


karussell:
@engaric has mentioned that he is working on a new reader, nice work! See [repo](https://github.com/engaric/graphhopper) and [mailing list entry](https://lists.openstreetmap.org/pipermail/graphhopper/2014-October/001294.html) and [this entry](https://lists.openstreetmap.org/pipermail/graphhopper/2014-November/001374.html)

Some notes from him:
- [ ] Fix exception 'org.opengis.referencing.NoSuchAuthorityCodeException: No code "EPSG:27700" from authority "EPSG" found for object of type "EngineeringCRS".'
- [ ] Find common behaviour for OSM and OSurvey, do it via interfaces and/or abstract classes.
- [ ] Move the import process out of the core to reduce dependencies. Of course also the OSM part should move either into tools or every importer gets his separate module.
- [ ] Avoid the heavy logging like in OSITNElement or OsItnInputFile and use unit tests instead, 
  which makes the import procedure also more future proof. See OSMReaderTest where we do similar stuff for OSM and just use artificial OSM XML
- [ ] For better portability (Android, JS, Objective-C) I abandoned java reflections in core and we are at java 6. So the reader 'guessing' should be done more statically e.g. similar to EncodingManager.parseEncoderString.
- [ ] Improve code formatting and class naming (OSITN vs. OsItn -> maybe OSurvey? As OS is already occupied from open source and operating system?)

(My) current instructions

``` bash
#0.
git clone https://github.com/engaric/graphhopper
#1. changed the config.properties to use OSurvey: 
reader.implementation=OSITN
#2. get graphhopper.sh from official gh master and overwrite current (somehow not in-sync?)
#3. 
wget https://www.ordnancesurvey.co.uk/docs/sample-data/os-mastermap-itn-layer-sample-data.zip#sample-data-download
#4.
unzip os-mastermap-itn-layer-sample-data.zip && mv Initial/58096-SX9192-2c1.gz sample.osm.gz
#5.
./graphhopper.sh import sample.osm.gz
```

karussell:
@engaric would still love to see this merged. Should be simpler now that OSM is slightly better separated from core. Also reflection is no longer used in core. Will close for now and hope for a PR :) + let us/me know your questions!

PR #616, opened 2015-12-22
Shapefile import via geotools

vvikas:
Shape file reader : 
- Abstract shapefile reader
- Implementation for OSM shape files
- Basic test case on small sample shapefile
- graphhopper.sh modifications

Run it via :  ./graphhopper.sh web - asia_india.shp


core/pom.xml
```diff
@@ -22,6 +22,7 @@
         <netbeans.hint.license>apache20</netbeans.hint.license>        
         <maven.build.timestamp.format>yyyy-MM-dd'T'HH:mm:ssZ</maven.build.timestamp.format>        
         <builddate>${maven.build.timestamp}</builddate>        
+        <geotools.version>10.8</geotools.version>
```
mprins: You should use a currently supported version of Geotools, eg. 14.1

core/pom.xml
```diff
@@ -92,8 +93,36 @@
...
+        </dependency>
+
     </dependencies>
         
+
+    <repositories>
+        <repository>
+            <id>maven2-repository.dev.java.net</id>
+            <name>Java.net repository</name>
+            <url>http://download.java.net/maven/2</url>
```
mprins: this repository does not actually exist

core/pom.xml
```diff
@@ -92,8 +93,36 @@
...
+            <name>Open Source Geospatial Foundation Repository</name>
+            <url>http://download.osgeo.org/webdav/geotools/</url>
+        </repository>
+        <repository> <!--Add the snapshot repository here-->
+            <snapshots>
+                <enabled>true</enabled>
+            </snapshots>
+            <id>opengeo</id>
+            <name>OpenGeo Maven Repository</name>
+            <url>http://repo.opengeo.org</url>
```
mprins: this repository does not actually exist

core/pom.xml
vvikas: I had used, mvn clean test verify
that seemed to use java 1.6 hence had to use previous version. 
Here tests are run against 1.7 and 1.8 , so will update to newer version.

karussell:
Hi Vikas,

thanks for this PR! Shapefiles are very valuable, especially to import or handle proprietary data. Is there a special format expected or certain attributes? E.g. how is the speed value per vehicle guessed? Or will it just import the topology and handling other attributes will be done elsewhere?

In order to integrate this we need to do some work like #450. This is already planned and creates a separate `osm-import` module. Then we need an additional refactoring separating OSM stuff from FlagEncoders. See also the discussion and the available refactoring in #277, maybe there are even some valuable refactorings which can be used here?

Recently I had the pleasure with playing with shp files on my own (maybe I'll blog about this) and geotools is just too huge to be included in the normal distribution or even in core. But after these refactorings we should be able to put the geotools stuff into a separate module like `shp-import` which then provides an implementation of `DataReader`, identical to what then OSM will provide.

The unit tests looks good. Additionally we need a stronger validation here e.g. ensuring that certain networks exists and the location lookup works, plus a simple Dijkstra run.

> Run it via : ./graphhopper.sh web - india-latest.osm

Do you mean `india-latest.shp` here instead? When doing the import: how large is the shp file and how much RAM is used for the import? Also after the necessary refactorings the shp files should work directly like osm or pbf files: 
`./graphhopper.sh web india-latest.shp`

vvikas:
Yes ! ./graphhopper.sh web india-latest.shp

vvikas:
I have abstracted out the logic to handle shape file handling and building graph. 
And processJunctions and proccessRoads methods need to be implemented by the
ShapeFile reader depending on the data.
One implementation I have provided for, is for files at : http://download.geofabrik.de
Different vendors have different schemes, who can create their shape file reader by implementing above methods.

Vehicle speeds are not explicitly calculated. For now just using 'maxspeed' in shape file, if not present an Average Speed is used. This will need improvements, which will again depend on the "schema of shape file".

devemux86:
Interesting implementation and useful too for proprietary data.

> we should be able to put the geotools stuff into a separate module like `shp-import`

:+1: 
Like Peter mentioned, it's better to not not mix the various `import` modules.
Probably a core module and extensions like the `osm-import` or `shp-import` with their own dependencies is probably the best way.

core/pom.xml
vvikas: First repository is not there, rest looks fine
Had picked form : http://docs.geotools.org/latest/userguide/tutorial/quickstart/maven.html

vvikas:
@karussell there is gc overhead in one build, should increase the maven memory in this case ?

karussell:
> And processJunctions and proccessRoads methods need to be implemented by the
ShapeFile reader depending on the data. One implementation I have provided for, is for files at : http://download.geofabrik.de

Ah, okay!

How do you determine if something is a junction vs. a bridge?

> Vehicle speeds are not explicitly calculated. For now just using 'maxspeed' in shape file, if not present an Average Speed is used. This will need improvements, which will again depend on the "schema of shape file".

Perfect!

> @karussell there is gc overhead in one build, should increase the maven memory ?

Hmmh, what do you mean here? I was asking about memory usage just because of comparison to OSM importer.

Regarding the refactoring: I'll try to implement #450 in the next weeks and then it shouldn't take much to change this PR to be merged. What do you think?

vvikas:
>Hmmh, what do you mean here? I was asking about memory usage just because of comparison to >OSM importer.

Please check attached log for asia_india.shp , for zip of ~300 MBs with RAM of 1G took about 5-6 minutes.
[import.log.zip](https://github.com/graphhopper/graphhopper/files/70620/import.log.zip)

The memory usage I was taking about was in the 'Tarvis CI build' on of its test instance is failing due to OOM.

I will take at look at #450 and other issues and see how much is to be done (refactoring,redesign, etc) and commit as per availability.

karussell:
Just to let you know: this is not forgotten :) As this is important work I'm working on #450 in a new branch and will provide this soonish as a new pull request. Then this work needs to be refactored (into a new maven module) to integrate it better in the new structure.

karussell:
@vvikas would you have a bit time to see if an integration of this PR is possible with the new separation done in #740?

vvikas:
Great , changes looks good. Will try to integrate with it. Should be done by end of July.

karussell:
Thanks a lot for the feedback! Will then try to integrate this in the next weeks.

karussell:
BTW: Merged the changes so this one here could be better separated too

core/pom.xml
```diff
@@ -22,6 +22,7 @@
         <netbeans.hint.license>apache20</netbeans.hint.license>        
         <maven.build.timestamp.format>yyyy-MM-dd'T'HH:mm:ssZ</maven.build.timestamp.format>        
         <builddate>${maven.build.timestamp}</builddate>        
+        <geotools.version>13.2</geotools.version>
```
mprins: You will want to update this to the latest maintainance release (14.x). The current stable (15.x requires java 8)

karussell: As this won't be in the core we can use java 8 like we do with the web module too.

karussell:
Adaptions to this PR done in #874

PR #617, opened 2015-12-23
Small fixes to make alternate routes work with the web module 

elibar:
1) Fix "algorithm" parameter name
2) Add "alternative_route.max_num" as a parameter on the http request


karussell:
Thanks!

PR #618, opened 2015-12-23
Refactor prepare contraction hierarchies

thehereward:
Just a tiny amount of tidying.
I'm hoping to have a look through more of the code when I get time.
🎁

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
+++ b/core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
@@ -819,100 +819,10 @@ public class PrepareContractionHierarchies extends AbstractAlgoPreparation imple
         AbstractBidirAlgo algo;
         if (AlgorithmOptions.ASTAR_BI.equals(opts.getAlgorithm()))
         {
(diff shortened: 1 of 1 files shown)


karussell:
Thanks! 

Would you mind to squash the two commits into one :) ?

Also why do you need the creation separated? Would it make sense for you to change `private` to `protected`? Then one could modify the creation process via subclassing.

Furthermore I would call it createAStar... instead of getAStar...

thehereward:
Thanks for such a quick reply!

> Would you mind to squash the two commits into one :) ?

Sure, will do :-)


> Also why do you need the creation separated?

It makes the code a bit easier to read: the `createAlgo` method is now much shorter and is just responsible for choosing which algorithm to create. The new methods are then responsible for actually creating the algorithm.


> Would it make sense for you to change private to protected? Then one could modify the creation process via subclassing.

This sounds more like a new feature than refactoring :-) I don't know enough about the code base to know if it is something we would want to make possible (and if it is there might be a better way of achieving it). If it is something we want to do then it should probably be addressed as a separate issue?


> Furthermore I would call it createAStar... instead of getAStar...

Good point, I will change this.

karussell:
> then it should probably be addressed as a separate issue

You are right!

thehereward:
Methods renamed and commits squashed :-)

karussell:
Thanks!

PR #619, opened 2015-12-23
Non-breaking changes: correct spellings and remove redundant semi-colons

thehereward:
Should all be non-breaking, very safe changes.
🎁


karussell:
Thanks again!

PR #620, opened 2015-12-24
Remove duplicated and unreachable code.

thehereward:
~~As a seasonal bonus this should increase test coverage (`calcNormalizedDist` was previously not covered by tests).~~ 
🎁

Edit: turns out it was covered by tests that I had just not noticed yet.


karussell:
Thanks again :) !

I would like to reserve 'getXY' for "nearly zero effort" operations ... would you mind to rename `getShrinkFactor` into `calcShrinkFactor` :) ?

thehereward:
Sure, no problem.

Issue #439, opened 2015-06-09
Add speed, time and more informations to instructions


nside:
It'd be useful to get the speed of route segments returned in a GHResponse. Not clear to me how to expose it though (add it to points or an array of speeds or something else?)

karussell:
Yes, would be nice and it should be already relative easy to customize GraphHopper to do so. But for GraphHopper itself we need a more generic functionality that could also return road type, surface, CO2 usage and more. So this issue is also related to #311

sebalis:
I came across something that would suit a more generic functionality the other day. I noticed that a park that I could use for walks between the 32nd Chaos Communication Congress and my hotel was closed at night. After tweeting about it and including a link to a Graphhopper routing (only for illustration), I was eventually pointed to this issue. See https://twitter.com/SebaLis/status/681440331754270720 – although I had never expected Graphhopper to do anything different from what it had done. (As I wrote in the thread, I only meant to lightly complain about the park being something of a "gated community", which also happened to be the congress motto.) It was only after SK53 and karussell responded that I started thinking about ways to handle this in Graphhopper.

So, imagine there is a routing that uses segments with any kind of restriction – an area or gate that is closed at certain times, a road with tolls or non-default speed limits, an underpass with a maximum height, a bridge with maximum weight etc. I think it would be good to first return the original (optimal, unrestricted) solution together with a hint that it has certain restrictions. This could be displayed in the style of a warning icon with a minimal amount of text pointing to the fact that there are restrictions, perhaps mentioning the type(s) but nothing more. The user could then click on the warning to reveal a set of form elements relevant to the restriction(s). So this would be an input for the time when the user wants to use the routing, checkboxes to avoid toll roads or speed restrictions, inputs for height or weight of the user’s vehicle. After completing and submitting this extra form, the route could be recalculated with those extra conditions taken into account.

Thanks for your work, I hope my idea will be useful in some way. :-)

karussell:
I've thought about this a bit more and instead of attaching this to the instructions we should create a new `details` sections. This way we can show all the details not just a summary of the streets between turns. Also interesting to fetch arbitrary OSM data from a route and potentially interesting for the [map matching component](https://github.com/graphhopper/map-matching).

The only problem is the potential big result set for long routes, but we could optimize this a bit via using the interval instead of repeating the geometry here again. Here is what I thought:

```json
"details": {
 "surface": { "unpaved": [[0, 3], [7,11], ...] },
[...]

PR #621, opened 2015-12-26
Conditional Access Restrictions

### LINKED ISSUE #374: Recognize conditional dates
Constructions should be avoided and assuming a daily re-import we should recognize the [conditional](http://wiki.openstreetmap.org/wiki/Conditional_restrictions) tag including the date format defined in [opening_hours](http://wiki.openstreetmap.org/wiki/Key:opening_hours)

boldtrn:
This PR adds the feature of checking conditional Access Restrictions. Currently only restrictions that last longer than one day are supported. The WayAcceptor currently uses the current day as comparison value.

Currently the following date formats are supported:
* 2014 Aug 10-2014 Aug 14
* Aug 10-Aug 14
* Jul-Aug
* Aug
* Sa-Su
* Su
* All of the above formats support the ranges in revers, e.g. Nov-Jan (matches Nov, Dec, Jan)

Things that are not supported yet:
* Multiple conditions like:
 * (2014 Aug 10-2014 Aug 14; Sa-Su)
 * no @ (2014 Aug 10-2014 Aug 14); no @ (Su)
* Things like: easter, PH, SH, summer, winter, ...

(Allowance of already restricted ways, like `yes @ (Su)` is now supported)

Partially fixes #374


karussell:
Thanks!

The unsupported things are no problem. One thing which I do not understand is 'Allowance of already restricted ways': Do you mean that `access=no` and `access:conditional=yes @ (date time)` is not yet possible?

boldtrn:
> Do you mean that access=no and access:conditional=yes @ (date time) is not yet possible?

Yes. This is not yet implemented. During the acceptWay method we have to add an allowance check. Currently, this seems like a conceptual issue to me. We are using a accept fast or fail fast strategy in the acceptWay method.

Once we returned 0, we cannot return the acceptBit. I didn't want to change that without discussing it.

karussell:
In acceptWay we have a section where one such condition could be included IMO:

```java
        String firstValue = way.getFirstPriorityTag(restrictions);
        if (!firstValue.isEmpty())
        {
            if (restrictedValues.contains(firstValue) /*&& HERE*/)                return 0;
            if (intendedValues.contains(firstValue))                return acceptBit;
        }
```

The foot&bike need simpler handling:
```java
if (way.hasTag(restrictions, restrictedValues) /*&& HERE*/) return 0;
```

Later on we can put this all into one method. For now unit tests should be sufficient e.g. for car&bike.

boldtrn:
@karussell I added the allowance in all encoders (motorcycle, foot, bicycle, and car) and wrote tests accordingly.

karussell:
> Do you think this is fixed with the name change? Or do we have to change something on the conceptual level?

Hmmh, not sure as we still would get `if !rejector.accept(way) then return 0` (or maybe I'm wrong here with mind-based renaming ;))

What about one class (like `ConditionalTags`) with two more direct methods like `reject` and `accept`?

boldtrn:
I did some refactoring. I guess it should readable now :).

karussell:
Nice - thanks! Just a tiny thing to discuss (sorry for the nit picking): are there shorter and still nice method names? http://stackoverflow.com/questions/2230871/when-is-a-java-method-name-too-long

I can't find better ones, just would move the `is` to the beginning: `isRestrictedWayConditional(ly)Permissive` and `isPermissiveWayConditional(ly)Restricted`

or could we remove the `conditional(ly)` as the class and therefor the variable has this in the name?

boldtrn:
You are right. The `is` should be in front.

I try to name methods in a way that no additional documentation is necessary (whenever possible ;)). And I think the `conditional(ly)` is important in this case. Without it would be: `isPermissiveWayRestricted`, this sounds a bit weird, I guess. But this is probably something that is discussable, so I can remove it if you want to?

karussell:
> I try to name methods in a way that no additional documentation is necessary (whenever possible ;))

Nice, that is exactly like it should be :) !

> this sounds a bit weird, I guess

Yes, you are right. Let us keep this but rename to `permissive` instead `permissed` :)

boldtrn:
> Let us keep this but rename to permissive instead permissed :)

Ok I changed permissed to permitted. I think this should be the better naming for this?

karussell:
Okay, this should be better :)

Also `isRestrictedWayConditionallyPermissed` need this change and in `isPermittedWayConditionallyRestriced` a 't' is missing in restricted.

To make it mergable we also need to avoid throwing exception here:
`if (conditionalTag.contains(";"))` and just print a warning instead (?)

In the class ParsedCalendar the methods should use the Java style like `isYearless` instead of `yearless`. I know the latter has advantages and we had this style in the early days, but the Java standard makes it easier for others when using autocomplete etc

In the new files the license header is missing.

> // TODO: Ferries have conditionals, like opening hours or are closed during some time in the year

What would be needed to implement this?

boldtrn:
> What would be needed to implement this?

Mhm, most ferries I saw I have quite complex opening hours (or 24/7). We are not yet able to parse these opening hours. See this query http://overpass-turbo.eu/s/dvv. So currently no easy solution available. Some ferries are seasonally closed, but these opening hours are setup quite complex.

Firs we have to consider the opening_hours tag (Some things change here, so we'd need a OpeningHoursParser).
Then we have to support combined opening_hours like: `opening_hours=Mo-Su 8:00-20:00; Nov-Mar off`.
Then we have to parse the opening hours of (every?) way.

karussell:
Ok, then we postpone this for ferries, of course. Also related to #439 

> Then we have to parse the opening hours of (every?) way.

Those with highway tags and also probably of every nodes with barriers

karussell:
I made a few minor modifications and while trying this on real OSM data I stumbled over:

 * the log info is too much for the case `conditionalTag.contains(";")` I vote for not logging this at all and adding this as todo for #374 
 * May I add the format `dd.MM.` and remove the failing test? Or why did you add an explicit test for this that this should fail? The same for `DateRangeParser.parseDateRange("Sat");`

boldtrn:
> the log info is too much for the case conditionalTag.contains(";") I vote for not logging this at all and adding this as todo for #374

I think this is a good idea. Also I think this could be doable with some clever transformations. I just found the [OpeningHoursParser of OSMAnd](https://github.com/osmandapp/Osmand/blob/master/OsmAnd-java/src/net/osmand/util/OpeningHoursParser.java). The do not have a good strategy for the `;` as well (as far as I can tell from skimming the code - see the FIXME ;)).

> May I add the format dd.MM. and remove the failing test? Or why did you add an explicit test for this that this should fail? The same for DateRangeParser.parseDateRange("Sat");

I added the failing tests to show the limitations of the parsing and I wrote them for me while developing to check if everything behaves as expected. I tested against the specification, which is: `Sat` is not the correct form for `Saturday`, according to [OSM](http://wiki.openstreetmap.org/wiki/Key:opening_hours).  I think there was a reason why I did not add the `dd.MM` format. Maybe only because it is not correct according to the documentation, but I am not sure.

karussell:
> Maybe we should investigate their code in more detail?

Please not. It is GPL :(

> Maybe only because it is not correct according to the documentation, but I am not sure. Please go ahead ;).

Okay, then will keep rejecting `Sat` but accepting `dd.MM(.)`

karussell:
If 'Aug 10-Feb' is used then currently all days in February do not lay inside this interval. Is this according to the wiki or do we need to fix this (potentially in a separate issue)

karussell:
I've merged this now

karussell:
Created a new issue for this https://github.com/graphhopper/graphhopper/issues/645

PR #623, opened 2015-12-30
Allowing multiple weightings/vehicle using CH

boldtrn:
This PR allows users to setup multiple weightings using CH. You have to define the weightings in the properties file like: 
`prepare.chWeighting=fastest, shortest`

### DIFF
--- a/config-example.properties
+++ b/config-example.properties
@@ -42,6 +42,9 @@ graph.dataaccess=RAM_STORE
 # graph but less for every request.
 # prepare.chWeighting=fastest
 #
+# You can also setup multiple weightings, by providing a coma separated list like:
+# prepare.chWeighting=fastest, shortest
+#
 # Disable the speed-up mode (contraction hierarchies, CH) via enabling the flexibility mode:
 # prepare.chWeighting=no
 #
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -31,7 +31,6 @@ import com.graphhopper.storage.index.LocationIndexTree;
 import com.graphhopper.storage.index.QueryResult;
 import com.graphhopper.util.*;
 import com.graphhopper.util.shapes.GHPoint;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -83,7 +82,7 @@ public class GraphHopper implements GraphHopperAPI
     // for CH prepare    
     private boolean doPrepare = true;
     private boolean chEnabled = true;
-    private String chWeightingStr = "fastest";
+    private List<String> chWeightingList = new ArrayList<String>(Arrays.asList("fastest"));
     private int chPrepareThreads = -1;
     private ExecutorService chPreparePool;
     private int preparePeriodicUpdates = -1;
@@ -294,21 +293,60 @@ public class GraphHopper implements GraphHopperAPI
     }
 
     /**
+     * Wrapper method for {@link GraphHopper#setCHWeightings(List)}
+     *
+     * @deprecated
+     * This method is used as a deprecated wrapper to not break the JavaApi.
+     * This will be removed in the future.
+     * Please use {@link GraphHopper#setCHWeightings(List)} or {@link GraphHopper#setCHWeightings(String...)}
+     */
+    @Deprecated
+    public GraphHopper setCHWeighting( String weightingName )
+    {
+        return this.setCHWeightings(Arrays.asList(new String[]{weightingName}));
+    }
+
+    /**
+     * Wrapper method for {@link GraphHopper#setCHWeightings(List)}
+     */
+    public GraphHopper setCHWeightings( String... weightingNames )
+    {
+        return this.setCHWeightings(Arrays.asList(weightingNames));
+    }
+
+    /**
      * Enables the use of contraction hierarchies to reduce query times. Enabled by default.
      * <p>
      *
-     * @param weighting can be "fastest", "shortest" or your own weight-calculation type.
+     * @param weightingList A list containing multiple weightings like:
+     *                     "fastest", "shortest" or your own weight-calculation type.
      */
-    public GraphHopper setCHWeighting( String weighting )
+    public GraphHopper setCHWeightings( List<String> weightingList )
     {
         ensureNotLoaded();
-        chWeightingStr = weighting.toLowerCase();
+
+        if (weightingList.isEmpty())
+            throw new IllegalArgumentException("It is not allowed to pass an emtpy weightingList");
+
+        this.chWeightingList.clear();
+        for (String weight : weightingList)
+        {
+            weight = weight.toLowerCase();
+            weight = weight.trim();
+            this.chWeightingList.add(weight);
+        }
         return this;
     }
 
-    public String getCHWeighting()
+    /**
+     * Returns all CHWeighting names
+     */
+    public List<String> getCHWeightings()
     {
-        return chWeightingStr;
+        if(this.chWeightingList == null || this.chWeightingList.isEmpty()){
+            throw new IllegalStateException("No CHWeightings are set");
+        }
+        return this.chWeightingList;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


karussell:
Thanks! This is indeed a very minor change - nice :) !

We could even avoid breaking the Java API when offering the old method setCHWeighting(String).

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -297,18 +297,29 @@ private GraphHopper setUnsafeMemory()
...
+     * @param weightingList A list containing multiple weightings like:
+     *                     "fastest", "shortest" or your own weight-calculation type.
      */
-    public GraphHopper setCHWeighting( String weighting )
+    public GraphHopper setCHWeighting( List<String> weightingList )
     {
         ensureNotLoaded();
-        chWeightingStr = weighting.toLowerCase();
+
+        if(weightingList.isEmpty())
```
karussell: we should throw an exception here saying at least one weighting is required.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -297,18 +297,29 @@ private GraphHopper setUnsafeMemory()
...
+            return this;
+
+        this.chWeightingList.clear();
+        for (String weight: weightingList)
+        {
+            weight = weight.toLowerCase();
+            weight = weight.trim();
```
karussell: should we assume already trimmed strings?

karussell:
Ah, okay the method `public List<String> getCHWeighting()` breaks the Java API and should be added to core/files/changelog.txt

config-example.properties
```diff
@@ -42,6 +42,9 @@ graph.dataaccess=RAM_STORE
 # graph but less for every request.
 # prepare.chWeighting=fastest
 #
+# You can also setup multiple weightings, by providing a coma separated list like:
+# prepare.chWeighting=fastest, shortest
```
devemux86: Probably comma separated without spaces ?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -959,7 +974,8 @@ public Weighting getWeightingForCH( WeightingMap weightingMap, FlagEncoder encod
         // get requested weighting name
         String weightingStr = weightingMap.getWeighting().toLowerCase();
         if (weightingStr.isEmpty())
-            weightingStr = chWeightingStr;
+            // TODO Ugly workaround or valid?
+            weightingStr = getCHWeighting().get(0);
```
karussell: This is valid as we always assume at least 'fastest' in the weighting list. Maybe throw an exception here if it is empty with a more clear exception than the ArrayOutOfBounds thing

config-example.properties
karussell: for the encoders we accept also spaces I think, trimming is easy and should be done before calling setCHWeighting I think. But for now I would be even okay with trimming inside the setter, not sure.

boldtrn: I thought about that as well. What is the downside of trimming? We do not allow weighting names like `" name"` or `"name "`. The benefit is that our config files are not sensible to whitespace.

I would not expect a config file to be sensible to whitespaces.

> trimming inside the setter

Do you think adding logic to a setter is bad?

boldtrn:
> We could even avoid breaking the Java API when offering the old method setCHWeighting(String).

How about creating a setCHWeighting method that uses varargs. This would not break the setter.

> public List<String> getCHWeighting()

How about providing a `String getCHWeighting()` method and add a `@deprecated `. The method would return the first element of list. This method could be also used in the `getWeightingForCH` method. We could remove that method in the future.

Therefore, this would be a pure feature that does not break the api and could be merged soon (since I'd like to see that feature in the master branch ;)).

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: You are right. I will update the PR

config-example.properties
karussell: > Do you think adding logic to a setter is bad?

Yeah. That is the only thing I don't like about it. But when reading the config file we should trim and then set with the array which is then used 'as it'. Maybe even do not do toLowerCase but throw exception if upper case?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -294,21 +293,57 @@ private GraphHopper setUnsafeMemory()
...
+        for (String weight : weightingList)
+        {
+            weight = weight.toLowerCase();
+            weight = weight.trim();
+            this.chWeightingList.add(weight);
+        }
         return this;
     }
 
+    public List<String> getAllCHWeightings()
```
karussell: here I vote for just 'getCHWeightings'

karussell:
Perfect ideas!

Just for the toLowerCase and trimming I would like to know your oppinion e.g. if just doing it or throwing exceptions etc

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: Me too, but overloading of methods is only valid if you use different parameters. Same method name with different return type is not valid.

boldtrn:
> Just for the toLowerCase and trimming I would like to know your oppinion e.g. if just doing it or throwing exceptions etc

Actually, I don't think we are doing any harm with the toLowerCase (was there before actually) and the trimming.

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: Ah, ups, you're right. You wanted to keep the old method, hmmh. I think we should break the API here as the new method should also have a nice name and the old method is probably not that often used.

config-example.properties
boldtrn: I actually don't think that adding logic to getters/setter is a bad practice. As long as we are trying to improve the quality of the "seted" value.

The pro of doing it in the setter is that we have a central/unique place to execute the correction of weighting values.

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: Ok, so I will change the name from `getAllCHWeightings` to `getCHWeightings` and change the name of `getCHWeighting` to `getFallbackCHWeighting` to be used in the `getWeightingForCH` method.

After writing the above I just realized, that `getWeightingForCH` and `getCHWeighting` is probably irritating.

Maybe we refactor the `getAllCHWeightings ` to `getCHWeightingNames`?

karussell: > After writing the above I just realized, that getWeightingForCH and getCHWeighting is probably irritating.

Hmmh, that is right. I would get rid of it and just keep `getCHWeightings` and `getWeightingForCH`. Currently new GH versions broke the Java API and if it is properly documented in the changelog.txt there was no problem or even questions in the forum.

BTW: Maybe we should rename `getWeightingForCH` to `findCHWeighting` to indicate the small overhead? But would be a different issue

boldtrn:
Ok everything should be fixed now. I added the renaming to the changelog.txt. What do you think, can we merge this branch?

karussell:
Thanks! One minor thing now is that the setter and getter of the `List<String>` are slightly 'misnamed'. Would you rename `public GraphHopper setCHWeighting( List<String> weightingList )` to setCHWeighting**s** and then squash your changes to one commit?

boldtrn:
Ok sure, I can do that. But I would leave the fallback method using varargs without the s? To not break the API at this point.

karussell:
> But I would leave the fallback method using varargs without the s? To not break the API at this point.

Hmmh, good question. Or keep setCHWeighting with one string and do setCHWeightings with the varargs?

boldtrn:
> Hmmh, good question. Or keep setCHWeighting with one string and do setCHWeightings with the varargs? As I would mark the old setCHWeighting method as deprecated and remove it at some point (?)

Ok this sounds reasonable. Change both List+Varargs to setCHWeightings and add a setCHWeighting method that is set deprecated?

karussell:
Ok!

boldtrn:
PR is squashed. Thanks.

karussell:
> PR is squashed. Thanks.

Oh, I have to thank: thanks :) !

One thing which I will do in another change:
`String tmpCHWeighting = args.get("prepare.chWeighting", "fastest");`

This should be renamed to `prepare.chWeightings` with a fallback to `prepare.chWeighting` which will be also removed later.

karussell:
Merged with a few changes in the GraphHopper class. https://github.com/graphhopper/graphhopper/commit/641d222d12221aac87bab5a23aefc40fae5edb71

karussell:
Thanks!

PR #630, opened 2015-12-31
Increase of priority influence for bike encoders

ratrun:
This change doubles the influence of the priority influence for the bicycles encoders bike and mtb. The priority influence for the racebike is increased by factor 1.5. See #597, #598 and #600.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -785,4 +785,10 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
     {
         specificBicycleClass = "class:bicycle:" + subkey;
     }
+    
+    @Override
+    public double defaultPriorityBoostLevel()
+    {
+        return 2.0;
+    }
 }
--- a/core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
@@ -317,6 +317,12 @@ public class CarFlagEncoder extends AbstractFlagEncoder
         else
             return "destination: " + str;
     }
+    
+    @Override
+    public double defaultPriorityBoostLevel()
+    {
+        return 1.0;
+    }
 
     @Override
     public String toString()
--- a/core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
@@ -132,4 +132,9 @@ public interface FlagEncoder extends TurnCostEncoder
      * @return true if already registered in an EncodingManager
      */
     boolean isRegistered();
+
+    /**
+     * @return the default boost level for the priority
+     */
+    public double defaultPriorityBoostLevel();
 }
--- a/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
@@ -385,6 +385,12 @@ public class FootFlagEncoder extends AbstractFlagEncoder
 
         return PriorityWeighting.class.isAssignableFrom(feature);
     }
+    
+    @Override
+    public double defaultPriorityBoostLevel()
+    {
+        return 1.0;
+    }
 
     @Override
     public String toString()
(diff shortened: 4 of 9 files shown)


karussell:
Thanks for this change. It is a simple change and still makes design problems we currently have in GraphHopper obvious:

 * With this change we would force all encoders to implement a PriorityWeighting specific method which is not good. As a temporary solution we could instead use `encoder.getBool(KEY, 0)`. (Still such a variable belongs to the weighting not into the encoder/storage, which we'll change later)
 * Also the bike routing is currently advertised as `fastest` although it uses the `priority` weighting under the hood. And boosting the priority even more could improve several things but at the same time would be a lot less `fastest`. We could introduce a `tourism-bike` or `trekking-bike` instead, but it would be probably better to reflect the difference via applying a different weighting formula on the same vehicle. Similar problem we have for `foot` vs. `hike`

Would you be okay with postponing this change?

ratrun:
> With this change we would force all encoders to implement a PriorityWeighting specific method which is not good. As a temporary solution we could instead use encoder.getBool(KEY, 0). (Still such a variable belongs to the weighting not into the encoder/storage, which we'll change later)

One arguments against this is that the priority is pretty much of the same level of importance compared to the speed stuff which is handled in the AbstractFlagEncoder and as it is only a default value, I see that this setting belongs to the flag encoders.

>Also the bike routing is currently advertised as fastest although it uses the priority weighting under the hood. And boosting the priority even more could improve several things but at the same time would be a lot less fastest. We could introduce a tourism-bike or trekking-bike instead, but it would be probably better to reflect the difference via applying a different weighting formula on the same vehicle. Similar problem we have for foot vs. hike

Isn't this a bit off topic? I completley agree, but I this problem arose already with the introduction of the PrioityWeighting over about two years ago. You are right that the change will make this more obvious. But only very few people look and question themselves about a missleading weighting setting in one of the URI parameters. But people do look at the route results.

I considered this change as starting point based on which a low hanging fruit for first simple but important flexibility becomes possible. Therefore I asked you to consider it for 0.6. Thanks for checking. From what I understood about the announced flexibilisation approach so far I thought that it still would be used, as it allows a lightway high level flexibility most people would be happy with. But as this does not seem to be the case I'm fine with postponing. I'll wait with further PRs and routing report fixing until the annouced flexibility improvements are settled and released.

karussell:
> Isn't this a bit off topic?

As you say: this change will make our old decision more obvious and I would avoid that. Instead we should takle these problems in 0.7 where we can better separate this and introduce different behaviour but keep the old stuff.

> I'll wait with further PRs and routing report fixing until the annouced flexibility improvements are settled and released.

Further changes are still welcome. Even in this case, but as I said it is a rather hard decision to keep things for a few weeks as they are and then e.g. make it possible that we deploy `fastest bike` and `priority bike`

karussell:
I think mainly issue #493 is missing (as CH is our default). Then multiple vehicles can use different weightings and we can offer an improved `priority bike` (like proposed in this PR) without overwriting the `fastest bike` option. This should not be much effort, I'll think if this could happen in 0.6.

karussell:
Hmmh, there would be more changes required than only specifying to which vehicle a chWeighting applies (like UI, createWeighting).

What we could instead do is that the parameter you introduced is changable per URL and uses the current value as default. This would not change current installations (nor affect the unit tests) but you would be able to change this in your local installation.

ratrun:
I'm using the URL parameter in my master, but I didn't expect that this simple solution would be acceptable for the general solution as it does not cover the CH case. The name for the parameter I'm using is niceLevel.
Question: Will #493 allow to specify something like this in the config.properties? 
```
vehicle:bicycyle:niceLeve=1
vehicle:foot:niceLeve=1
```
Can these settings be queried via `pMap.getDouble("vehicle:bicycyle:niceLeve", 1.0);` ? 

If so, what naming conventions do you suggest?

karussell:
> I didn't expect that this simple solution would be acceptable for the general solution 
> as it does not cover the CH case.

Currently it doesn't in the GraphHopper class directly, one would need to overwrite some behaviour but when creating a weighting - be it for CH routing or none-CH routing - you can specify configuration. The only CH requirement is that you need these config changes directly after import and on-query changes are currently just ignored.

> Question: Will #493 allow to specify something like this in the config.properties?

No. But this is exactly the problem: we need a way to define a 'weighting alias' and make this accessible with a JSON, which is the goal of the flexiblity improvement.

> Can these settings be queried via pMap.getDouble("vehicle:bicycyle:niceLeve", 1.0); ? 

No, currently you'll need to provide config settings via URL. The naming convention is like `namespace.sub_namespace.key`

ratrun:
Thank you.

Conclusion of the PR: The purpose of this PR will become valid in a different from when the mentioned flexibility extensions become available. Therefore I'm closing this PR.  Fixes for #597, #598 and #600 can only be provided when the possiblity dor flexibility extensions becomes available.

Issue #634, opened 2016-01-07
Odd choices in bicycle routing, perhaps due to 50 km/h limit on some of the involved segments


sebalis:
Hi, as requested by @karussel in the thread after https://twitter.com/SebaLis/status/684729267419234304 I have this to report:

On a local road, Graphhopper makes some strange choices in bicycle routing. This is a large inner-city road, dual carriageway and part of a German “B” road but still a city road that is not grade-separated, intersects with all roads and has a pavement that is partly a cycle lane next to it. In OSM it’s modelled as two ways for the two directions tagged with highway:secondary and oneway:yes, and a maxspeed:50 tag. The pavement and cycle lane is modelled as one and the same sequence of ways having both the foot:designated and bicycle:designated tags.

https://graphhopper.com/maps/?point=52.022785%2C8.528824&point=52.019959%2C8.526238&locale=en&vehicle=bike&weighting=fastest&elevation=true&layer=Omniscale shows that a bicycle routing on the road from one crossing to the third-next one branches off the road at an intermediate crossing using lanes that are actually intended only for turning into a residential road and back from that road into the main one. These are segments marked highway:residential and oneway:yes but with no speed limit (of course the regular 50 km/h speed within German cities still applies, I suppose the limit is there on the main road because in some rare cases a 70 km/h applies to main city roads).

When the start and end positions are moved from the road to the pavement/cycle lane as shown in https://graphhopper.com/maps/?point=52.022844%2C8.528749&point=52.019986%2C8.526099&locale=en&vehicle=bike&weighting=fastest&elevation=true&layer=Omniscale the routing still tries to use these residential segments and ends up with an even more nonsensical path.

@karussell speculated on Twitter that the missing speed limit was what made these segments so popular for Graphhopper’s cycle routing. But that does not make much sense as a) bicycles should not avoid a road with a 50 km/h limit and b) using these two short segments ends up with a longer and non-straight overall route, particularly in the second case, and that extra length and the required turns should outweigh the advantage of hypothetically being faster than 50 km/h on these short segments. Also, the pavement/cycle lane that is used in the second case does not have a speed limit either of course and is also a more direct route than the solution that is returned.

ratrun:
The main problem is that the choosen cycleway segment is tagged as `highway=cycleway` and the other way as path. For path we assume a lower speed. An additional small thing I noticed is a small tagging problem in relation 2418376: This cycle route does not include any network tag.

karussell:
[...]

PR #635, opened 2016-01-07
Change speed for highway=path into the cycleway speed for all …

ratrun:
…bicycle types

Make sure that highway=path + way.hasTag("bicycle", intendedValues) results in the same highway speed as for a cycleway. See #634.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -424,14 +424,29 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
         }
 
         // Until now we assumed that the way is no pushing section
-        // Now we check, but only in case that our speed is bigger compared to the PUSHING_SECTION_SPEED
-        if ((speed > PUSHING_SECTION_SPEED)
-                && (!way.hasTag("bicycle", intendedValues) && way.hasTag("highway", pushingSections)))
-        {
-            if (way.hasTag("highway", "steps"))
-                speed = PUSHING_SECTION_SPEED / 2;
-            else
-                speed = PUSHING_SECTION_SPEED;
+        // Now we check that, but only in case that our speed is bigger compared to the PUSHING_SECTION_SPEED
+        if (speed > PUSHING_SECTION_SPEED)
+        { 
+            if (way.hasTag("highway", pushingSections))
+            {
+                if (!way.hasTag("bicycle", intendedValues))
+                {
+                 if (way.hasTag("highway", "steps"))
+                    speed = PUSHING_SECTION_SPEED / 2;
+                 else
+                    speed = PUSHING_SECTION_SPEED;
+                }
+                else
+                {
+                   if (way.hasTag("bicycle", "designated") || way.hasTag("bicycle", "official"))
+                      speed = getHighwaySpeed("cycleway");
+                   else
+                   {
+                       // (way.hasTag("bicycle", "yes") || way.hasTag("bicycle", "permissive"))
+                       speed = PUSHING_SECTION_SPEED * 2;
+                   }
+                }
+            }
         }
 
         return speed;
@@ -547,8 +562,11 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
     {
         String service = way.getTag("service");
         String highway = way.getTag("highway");
-        if (way.hasTag("bicycle", "designated"))
-            weightToPrioMap.put(100d, PREFER.getValue());
+        if (way.hasTag("bicycle", "designated") || way.hasTag("bicycle", "official"))
+            if ("path".equals(highway))
+                weightToPrioMap.put(100d, VERY_NICE.getValue());
+            else
+                weightToPrioMap.put(100d, PREFER.getValue());
         if ("cycleway".equals(highway))
             weightToPrioMap.put(100d, VERY_NICE.getValue());
 
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -163,8 +163,9 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
         setHighwaySpeed("living_street", 6);
         setHighwaySpeed("steps", PUSHING_SECTION_SPEED / 2);
 
-        setHighwaySpeed("cycleway", 18);
-        setHighwaySpeed("path", 12);
+        final int CYCLEWAY_SPEED = 18;  // Make sure cycleway and path use same speed value, see #634
+        setHighwaySpeed("cycleway", CYCLEWAY_SPEED);
+        setHighwaySpeed("path", CYCLEWAY_SPEED);
         setHighwaySpeed("footway", 6);
         setHighwaySpeed("pedestrian", 6);
         setHighwaySpeed("track", 12);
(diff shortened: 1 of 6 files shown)


karussell:
Thanks!

As this relates to the other issue ... shouldn't we use the high speed of 18km/h only if the path is also tagged as bicycle=designated? Just path doesn't look this fast e.g. see [the wiki images](http://wiki.openstreetmap.org/wiki/Tag:highway%3Dpath)

ratrun:
Now I included your suggestion "shouldn't we use the high speed of 18km/h only if the path is also tagged as bicycle=designated?"

I believe that handling of smoothness is another more complex issue, which sbetter hould be handled in a separated commit.

ratrun:
@karussell Can you please check this PR? It seems that we forgot about it.

karussell:
Not forgotten just too many things.

core/src/test/java/com/graphhopper/routing/util/BikeFlagEncoderTest.java
```diff
@@ -68,23 +68,40 @@ public void testGetSpeed()
...
 
-        assertEquals(6, encoder.getSpeed(way));
+        assertEquals(PUSHING_SECTION_SPEED * 2, encoder.getSpeed(way));
         assertPriority(UNCHANGED.getValue(), way);
 
         way.clearTags();
+        way.setTag("highway", "cycleway");
+        assertEquals(18, encoder.getSpeed(way));
+        assertPriority(VERY_NICE.getValue(), way);
+        int cyclewayspeed = encoder.getSpeed(way);
```
karussell: there is no assert here - is this intended?

karussell: and a check is missing where highway=path without another tag resulting in the path speed=12 ?

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
```diff
@@ -163,8 +163,9 @@ protected BikeCommonFlagEncoder( int speedBits, double speedFactor, int maxTurnC
         setHighwaySpeed("living_street", 6);
         setHighwaySpeed("steps", PUSHING_SECTION_SPEED / 2);
 
-        setHighwaySpeed("cycleway", 18);
-        setHighwaySpeed("path", 12);
+        final int CYCLEWAY_SPEED = 18;  // Make sure cycleway and path use same speed value, see #634
+        setHighwaySpeed("cycleway", CYCLEWAY_SPEED);
+        setHighwaySpeed("path", CYCLEWAY_SPEED);
```
karussell: why are they equal? Shouldn't they be different and we should use the cycleway speed instead when highway=path and designated?

karussell:
@ratrun would mind looking into this here :) ?

ratrun:
The cycelway speed for path is taken only if in addition one of the intendedValues is set.

karussell:
> The cycelway speed for path is taken only if in addition one of the intendedValues is set.

What do you mean here and would you mind to have a look into the questions (see also in comments above):

> why are they equal? Shouldn't they be different and we should use the cycleway speed instead when highway=path and designated?
> there is no assert here - is this intended?
> and a check is missing where highway=path without another tag resulting in the path speed=12 ?

 I really would like to get this merged :)

core/src/test/java/com/graphhopper/routing/util/BikeFlagEncoderTest.java
ratrun: Please check the whole code and not only the diff. I suppose you are missing an assert after the assignment to `cyclewayspeed`, but I'm just fetching it and use it in further tests below.

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
ratrun: Sorry. I overlooked these question here. This is a good one. But the answer is not that easy.
The explicit highway speed for path is actually not used at all, because of these lines on line 441 in BikeCommonFlagEncoder.java
```
                   if (way.hasTag("bicycle", "designated") || way.hasTag("bicycle", "official"))
                      speed = getHighwaySpeed("cycleway");
```
But because ot the check in line 428 
`if (speed > PUSHING_SECTION_SPEED)`
it is important that it is assigned with a highway speed > PUSHING_SECTION_SPEED. 
Maybe there might be a simpler and more elegant solution, but I didn't one. Please try to play around with it, then you'll see what I mean.

ratrun:
>>>What do you mean here? And would you mind to have a look into the questions (see also in comments above):

Here I meant line 441:

```
                   if (way.hasTag("bicycle", "designated") || way.hasTag("bicycle", "official"))
                      speed = getHighwaySpeed("cycleway");
```

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
karussell: Just to understand this better: if highway=path without any other tags then it is accessible for bike and we need some asserts and a lower speed or am I overlooking something?

karussell: Ah, got it. Still I'll add a unit test for this case...

core/src/test/java/com/graphhopper/routing/util/BikeFlagEncoderTest.java
```diff
@@ -68,23 +68,40 @@ public void testGetSpeed()
...
+        assertEquals(PUSHING_SECTION_SPEED * 2, encoder.getSpeed(way));
         assertPriority(UNCHANGED.getValue(), way);
 
+        way.setTag("surface", "unpaved");
+        assertEquals(PUSHING_SECTION_SPEED * 2, encoder.getSpeed(way));
+
+        way.setTag("surface", "paved");
+        assertEquals(PUSHING_SECTION_SPEED * 2, encoder.getSpeed(way));
+
         // Pushing section Ok !!
```
karussell: btw: do you know what this comment means :) ?

karussell:
Made some minor edits, added the small test case and squashed, see the change which I intend to merge soon, here: https://github.com/graphhopper/graphhopper/compare/issue_635

karussell:
The only minor question left is: why do we have 'get off the bike', 'cycleway' and 'way'? What purpose does 'way' have - shouldn't it be empty instead?

ratrun:
>The only minor question left is: why do we have 'get off the bike', 'cycleway' and 'way'? What >purpose does 'way' have - shouldn't it be empty instead? 

The original intention was to make the instructions more clear from a bike riders perspective and give hints what kind of way to seek and expect. I think that I would not try this again if I had to re-implement that idea. So if you don't like the idea please delete it.

karussell:
> I think that I would not try this again if I had to re-implement that idea. So if you don't like the idea please delete it.

I like the information about getting of the bike and cycleway, although the API should be a bit different in the future but for now we keep this. What I mean is that the 'way' is not informative (compared to get off the bike and cycleway) and I would leave it empty instead?

core/src/test/java/com/graphhopper/routing/util/BikeFlagEncoderTest.java
ratrun: Sorry, I forgot. I believe `// Pushing section Ok !!`meant that the next test
```
        way.setTag("surface", "paved");
```
 is fine in case that the result is that it is a pushing section. Please delete the comment

ratrun:
>I like the information about 'getting off the bike' and 'cycleway', although the API should be a bit different in the future ('more separate') but for now we keep this. What I mean is that 'way' is not informative (compared to get off the bike and cycleway) and I would leave it empty instead?

Yes this is fine. Maybe the problem here was that I'm was thinking about the German term "Weg" or colloquialism "Wegerl", which means something small. The English term way is probably not 1:1 equivivalent.

karussell:
Maybe we should translate explicitely 'narrow way' (or road/path)?

karussell:
Thanks again for this contribution and sorry for the delay!

PR #641, opened 2016-01-15
Center vehicle icons vertically in buttons + minor JS code refactor

fbonzon:
This PR will move down vehicle icons 1 or 2 px down, so they are centered inside their button.

Use Mapbox retina tiles when possible. It’s commented out here. For people who register their own API token and uncomment.

Simplify subdomains of tiles URL. In Leaflet, value “abc” is already default for {s}.


karussell:
Thanks! Would you mind to squash into 1 commit?

fbonzon:
Done.

PR #648, opened 2016-01-26
Wrap coordinates around dateline

fbonzon:
When browsing another copy of the world, left or right the central one, after right-clicking on the map to add start, end, or intermediate points, error like this one is displayed:
```
Cannot find point 0: 52.50452,373.351135
Cannot find point 1: 52.516639,373.428383
```
This PR will make all clicks with longitude -180 to +180 degrees.

Uses the [wrap()](http://leafletjs.com/reference.html#latlng-wrap) function from Leaflet.

It's not a complete solution to this issue, as the route is always drawn on the main copy of the world. However it is still an improvement:
* Error is gone
* After routing results success, map is automatically panned to the route anyway (in central copy of world), and brings you back to the main world

### DIFF
--- a/web/src/main/webapp/js/main-template.js
+++ b/web/src/main/webapp/js/main-template.js
@@ -325,14 +325,14 @@ function setToEnd(e) {
 }
 
 function setStartCoord(e) {
-    ghRequest.route.set(e.latlng, 0);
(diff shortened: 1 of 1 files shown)


karussell:
Thanks!

PR #640, opened 2016-01-15
Alternative route for flexibility

karussell:
This pull request finally contains the alternative path refactoring, as well the the alternative routing itself (edge&node based traversal), the necessary UI changes and the old round trip algorithm.

The quality of finding alternatives has increased compared to the old proposal due to the introduction of the 'share' property. Still performance is not optimal.

To enable it just pass `algorithm=alternativeroute` or the Java equivalent. The result looks like:
alt_routes

To calculate more than one alternative use `alternative_route.max_paths=3` (2 alternatives)

This PR makes the PRs #420 & #596 obsolete

TODOs: major speed improvements possible (paper 'Improved Alternative Route Planning'), use bidirectional A*? make alternative search working for CH, post processing of the alternatives to filter away alternatives sharing too much with other alternatives

round trip calculation needs integration tests and a working end path removal or just the addition/replacement via #582

PR #658, opened 2016-02-04
Fix GraphHopperWeb to pass alternative_route hints to the server

elibar:
Passing all hints that starts with alternative_route as quay parameters on the route request

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
@@ -103,15 +103,25 @@ public class FootFlagEncoder extends AbstractFlagEncoder
         safeHighwayTags.add("track");
         safeHighwayTags.add("residential");
         safeHighwayTags.add("service");
+        
+        safeHighwayTags.add("trunk");
+        safeHighwayTags.add("trunk_link");
+        safeHighwayTags.add("primary");
+        safeHighwayTags.add("primary_link");
+        safeHighwayTags.add("secondary");
+        safeHighwayTags.add("secondary_link");
+        safeHighwayTags.add("tertiary");
+        safeHighwayTags.add("tertiary_link");
+        
 
-        avoidHighwayTags.add("trunk");
-        avoidHighwayTags.add("trunk_link");
-        avoidHighwayTags.add("primary");
-        avoidHighwayTags.add("primary_link");
-        avoidHighwayTags.add("secondary");
-        avoidHighwayTags.add("secondary_link");
-        avoidHighwayTags.add("tertiary");
-        avoidHighwayTags.add("tertiary_link");
+//        avoidHighwayTags.add("trunk");
+//        avoidHighwayTags.add("trunk_link");
+//        avoidHighwayTags.add("primary");
+//        avoidHighwayTags.add("primary_link");
+//        avoidHighwayTags.add("secondary");
+//        avoidHighwayTags.add("secondary_link");
+//        avoidHighwayTags.add("tertiary");
+//        avoidHighwayTags.add("tertiary_link");
 
         // for now no explicit avoiding #257
         //avoidHighwayTags.add("cycleway"); 
--- a/web/src/main/java/com/graphhopper/http/GraphHopperWeb.java
+++ b/web/src/main/java/com/graphhopper/http/GraphHopperWeb.java
@@ -23,9 +23,10 @@ import com.graphhopper.GHResponse;
 import com.graphhopper.GraphHopperAPI;
 import com.graphhopper.util.*;
 import com.graphhopper.util.shapes.GHPoint;
-import java.util.ArrayList;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
 
 import org.json.JSONArray;
 import org.json.JSONObject;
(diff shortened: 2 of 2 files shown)


karussell:
Are the FootFlagEncoders intended to be included here?

We should not filter the hints (i.e. remove the `if(key.startsWith("alternative_route.")){`) and calling URLEncode on the value and key is very important here.

elibar:
Hi 
The FootFlagEncoders changes are not intended to be included in this pull request. Though  I think we should change the foot weights. I stopped using the avoidHighwayTags weights as I got bad feedback from pedestrians about the generated routes.

If removing the `if(key.startsWith("alternative_route.")` we should remove also the code that deals with the key, so we will not have the key query parameter twice in the request

karussell:
> Maybe this is a little bit off topic here

@ratrun Thanks for this JS catch. Would you mind to open up a new issue for this? Also the `.` is intended to work in JS like it supposed to be: would you mind to try: `{...., "alternative_route": { "max_paths" : 3 } }` instead?

> If removing the if(key.startsWith("alternative_route.") we should remove also the code 
> that deals with the key, so we will not have the key query parameter twice in the request 

@elibar what do you mean here? I would like to avoid changing code here again, if there are new parameters (e.g. besides alternative_route) necessary to add.

elibar:
There is a special handle for the "key" parameter
` String tmpKey = request.getHints().get("key", key);` 
and 
` if (!tmpKey.isEmpty())`
`                url += "&key=" + tmpKey;`

If we have code that copy all hints to query parameters the "key" parameter will be added twice to the URL, once by the above code and once by the code that handle all hints.

karussell:
> There is a special handle for the "key" parameter

Ah, yes, indeed, also: 'instructions' and 'elevation'

A bit tricky is `calcPoints` which would be the only param which we should just ignored in the hints loop.

ratrun:
>Would you mind to open up a new issue for this? Also the . is intended to work in JS like it supposed to be: would you mind to try: {...., "alternative_route": { "max_paths" : 3 } } instead?

Thank you. Shortly after having sent this comment, I had the same idea. But this required a couple of ofther changes as well. See d4c9c7400210ee6a033ef52f36e1583c33af1fe4 . The good thing is that URI parameter renaming and a new issue is not necessary.

karussell:
> But this required a couple of ofther changes as well.

What do you mean here? It looks like you did not base the change on the most recent master (?)

ratrun:
@karussell: Somehow your response does not appear here. I received the following via the mail notification only.

>>    But this required a couple of other changes as well.

>What do you mean here? It looks like you did not base the change on the most recent master (?)

Yes, the change is based on my master, not the official one. The reason was the following statement in #640 about the alternative_route WEB UI support:

>I'm a bit unsure about this. We can add a checkbox for this but we cannot display this option if a CH-based query is running yet and also the performance is not yet there where I would like it. We will announce it though in the release notes and pointing to this issue. Then others (and me) can investigate how to implement this in CH and make it faster for one of the releases afterwards.

Despite that I wanted to point to a working implementation. It might help others, because it took me a couple of attempts to get it working.

karussell:
> @karussell: Somehow your response does not appear here. I received the following via the mail notification only.

Sorry, I removed it as I saw you updated the change and it pointed before to a the big commit :) and now it is clear.

> Despite that I wanted to point to a working implementation. It might help others, because it took me a couple of attempts to get it working.

Probably not a bad idea. Still I like to have the UI clean, maybe show these (and later) advanced options only when clicking on a small gear wheel icon or something?

ratrun:
>Still I like to have the UI clean, maybe show these (and later) advanced options only when clicking on a small gear wheel icon or something?

The UI remains unchanged for the end-user in case that a CH server instance is used in the backend. For CH backend instances the sliders, which do not make sense otherwise, are hidden.

karussell:
> The UI remains unchanged for the end-user in case that a CH server instance is used 
> in the backend. For CH backend instances the sliders, which do not make sense otherwise, are hidden.

Sure, but even for flexibility mode I would avoid it. That won't be the only option to come ...

ratrun:
>Sure, but even for flexibility mode I would avoid it. That won't be the only option to come ...

Yes. But I think that the web UI source code should always support all the features the core server source code provides. Otherwise it gets difficult for contributors, as we would always need to look up documentation or code for the available parameters and add them to the URL by hand. This is not convenient at all.

Do I understood it right that you intend to be able to announce from a server side configuration setting which GUI elements to show on the web UI? In this case, this probably should be added as part of the flexibility enhancements right from the start. Another more complicated solution is to dynamically create the web UI using npm based via build settings. But I would like avoid that as this sounds too complicated.

karussell:
> This is not convenient at all.

I understand what you mean but keep in mind that there are already many features not accessible from the UI. This is partly from the fact that is does not make much sense and partly from the lack of time to do this properly like it was the case for alternative routes. And just adding more controls won't make the UI easier for beginners.

> Do I understood it right that you intend to be able to announce from a server side configuration

I didn't mean that here.

> But I would like avoid that as this sounds too complicated.

This is already done with e.g. the elevation support and is not really complex (or maybe I misunderstand what you meant?)

karussell:
I've fixed this - @elibar let me know if this works for you this way.

karussell:
@ratrun for now I make several features more visible via the new ['overview'](https://github.com/graphhopper/graphhopper/blob/master/README.md#overview) in the readme - let me know if something is missing or needs explanation. Please create a separate issue for the UI change.

Issue #292, opened 2014-12-04
SRTM Service Connection Refused


Andy1605:
Hello,

If i pull the current master branch of graphhopper i am not compatible with the maps from the maps repository anymore (different graph versions 7 and 9)! So i decided to build maps on my own.
Small maps are no problem (like Schwaben(GER)) but with bigger maps like Germany, i get the following exception while building with graphhopper.sh:

Exception in thread "main" java.lang.RuntimeException: Couldn't process file /home/maps/OSMs/germany.osm, error: java.net.ConnectException: Connection refused
    at com.graphhopper.reader.OSMReader.writeOsm2Graph(OSMReader.java:321)
    at com.graphhopper.reader.OSMReader.readGraph(OSMReader.java:152)
    at com.graphhopper.GraphHopper.importData(GraphHopper.java:623)
    at com.graphhopper.GraphHopper.process(GraphHopper.java:592)
    at com.graphhopper.GraphHopper.importOrLoad(GraphHopper.java:565)
    at com.graphhopper.tools.Import.main(Import.java:16)
Caused by: java.lang.RuntimeException: java.net.ConnectException: Connection refused
    at com.graphhopper.reader.dem.SRTMProvider.getEle(SRTMProvider.java:302)
    at com.graphhopper.reader.OSMReader.getElevation(OSMReader.java:562)
    at com.graphhopper.reader.OSMReader.addNode(OSMReader.java:547)
    at com.graphhopper.reader.OSMReader.processNode(OSMReader.java:522)
    at com.graphhopper.reader.OSMReader.writeOsm2Graph(OSMReader.java:291)
    ... 5 more
Caused by: java.net.ConnectException: Connection refused
    at java.net.PlainSocketImpl.socketConnect(Native Method)
    at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:339)
    at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:200)
    at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:182)
    at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
    at java.net.Socket.connect(Socket.java:579)
    at sun.net.NetworkClient.doConnect(NetworkClient.java:175)
    at sun.net.www.http.HttpClient.openServer(HttpClient.java:432)
    at sun.net.www.http.HttpClient.openServer(HttpClient.java:527)
[...]

PR #668, opened 2016-02-21
Updated SRTM Provider URL

boldtrn:
I created a mirror for the SRTM data at http://srtm.motoroute.me/. I set that URL as the new default. See also #292


karussell:
Nice & thanks a lot!

PR #674, opened 2016-03-03
Update documentation for simple routing example now using PathWrapper

ammagamma:
doc/core/routing.md seems to be outdated as for example it says that the InstructionList can be obtained directly from the GHResponse. However, first one needs to call getBest() or getAll() to obtain a PathWrapper.


karussell:
Thanks!

PR #676, opened 2016-03-06
More appropriate MIME type for GPX exports

fbonzon:
There is no [official assignment](http://www.iana.org/assignments/media-types/media-types.xhtml) for the GPX format., however [Wikipedia](https://en.wikipedia.org/wiki/GPS_Exchange_Format) and several others use `application/gpx+xml`.


karussell:
What problems do you have with `application/xml`? I would avoid introducing this change as a default because it could break existing clients. Still we can add it to potential later improvements here #437 .

fbonzon:
I don't have a real problem with this MIME type. Just trying to improve it. Good to see it's already in the improvements list.

OK to delay this change to a next API major version upgrade, when changes are OK to break existing clients.

karussell:
> I don't have a real problem with this MIME type

okay, that is good to hear! 

> Good to see it's already in the improvements list.

Yeah, added it :)

And: still thanks for this!

karussell:
Fixed

PR #678, opened 2016-03-08
Move constant CSS from js to css

fbonzon:
These 2 CSS declarations `float: right;` and `right: 0;` are actually constant and can be moved to the css file.

Tested on my dev instance.

Code cosmetic:
- 0 times whatever is 0, so we may omit "px"
- Remove trailing white space (sorry, done automatically by my editor)

### DIFF
--- a/web/src/main/webapp/css/style.css
+++ b/web/src/main/webapp/css/style.css
@@ -8,8 +8,10 @@ body {
     min-width: 700px;
 }
 #map {
-    /* set size via JS */
+    /* Size and position ("relative" or "absolute") set via JS. */
     cursor: default;
+    float: right;
+    right: 0;
 }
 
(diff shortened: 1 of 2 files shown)


karussell:
Was there a problem before with the constant CSS in JS?

fbonzon:
No problem.

It seems more efficient to have these 2 directives once in CSS, rather than to keep overwriting via JS them without any actual change, as the map adjusts its size due to browser changing window size. Up to you, maybe the "more readable" code argument will win.

karussell:
Can we keep it as it is? I do not see an advantage and as the properties belong to each other the JS way is slightly more expressive.

fbonzon:
That's fine, if it's your preference.

PR #681, opened 2016-03-09
Feature maximum visited nodes

### LINKED ISSUE #473: Make maximum visited nodes configurable
To ensure that one can run none-CH in production we need an option to avoid full graph exploration

boldtrn:
This PR contains the ability to limit the number of visitedNodes that a non-ch RoutingAlgorithm visits.

This can be specified in the config and via an URL parameter.

Fixes #473

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -72,6 +72,7 @@ public class GraphHopper implements GraphHopperAPI
     private boolean simplifyResponse = true;
     private TraversalMode traversalMode = TraversalMode.NODE_BASED;
     private final Map<Weighting, RoutingAlgorithmFactory> algoFactories = new LinkedHashMap<Weighting, RoutingAlgorithmFactory>();
+    private int maxVisitedNodes = Integer.MAX_VALUE;
     // for index
     private LocationIndex locationIndex;
     private int preciseIndexResolution = 300;
@@ -686,6 +687,8 @@ public class GraphHopper implements GraphHopperAPI
 
         // routing
         defaultWeightLimit = args.getDouble("routing.defaultWeightLimit", defaultWeightLimit);
+        maxVisitedNodes = args.getInt("routing.maxVisitedNodes", Integer.MAX_VALUE);
+
         return this;
     }
 
@@ -1093,10 +1096,16 @@ public class GraphHopper implements GraphHopperAPI
         QueryResult fromQResult = qResults.get(0);
 
         double weightLimit = request.getHints().getDouble("defaultWeightLimit", defaultWeightLimit);
+        int maxVisistedNodesForRequest = request.getHints().getInt("routing.maxVisitedNodes", maxVisitedNodes);
+        if(maxVisistedNodesForRequest > maxVisitedNodes){
+            ghRsp.addError(new IllegalStateException("The routing.maxVisitedNodes parameter has to be below or equal to:"+maxVisitedNodes));
+            return Collections.emptyList();
+        }
+
         String algoStr = request.getAlgorithm().isEmpty() ? AlgorithmOptions.DIJKSTRA_BI : request.getAlgorithm();
         AlgorithmOptions algoOpts = AlgorithmOptions.start().
                 algorithm(algoStr).traversalMode(tMode).flagEncoder(encoder).weighting(weighting).
-                hints(request.getHints()).
+                maxVisitedNodes(maxVisistedNodesForRequest).hints(request.getHints()).
                 build();
 
         boolean viaTurnPenalty = request.getHints().getBool("pass_through", false);
--- a/core/src/main/java/com/graphhopper/routing/RoundTripAltAlgorithm.java
+++ b/core/src/main/java/com/graphhopper/routing/RoundTripAltAlgorithm.java
@@ -48,6 +48,7 @@ public class RoundTripAltAlgorithm implements RoutingAlgorithm
     private final TraversalMode traversalMode;
     private double weightLimit = Double.MAX_VALUE;
     private int visitedNodes;
+    private int maxVisitedNodes = Integer.MAX_VALUE;
     private double maxWeightFactor = 2;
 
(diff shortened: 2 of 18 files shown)


boldtrn:
@karussell I think we have an issue with our current travis configuration. I forgot to run the whole test-suite before pushing and it failed for me locally. Since the js tests succeed, travis seems to think that all tests succeed. Btw, I am currently fixing the issues ;).

boldtrn:
Ok. PR should be fine. What do you think?

core/src/main/java/com/graphhopper/routing/RoundTripAltAlgorithm.java
```diff
@@ -209,6 +212,17 @@ public void setWeightLimit( double weightLimit )
...
 
     @Override
+    public void setMaxVisitedNodes( int numberOfNodes )
+    {
+        this.maxVisitedNodes = numberOfNodes;
+    }
+
+    protected boolean isMaxVisitedNodesExceeded()
+    {
+        return maxVisitedNodes > getVisitedNodes();
```
karussell: here is a minor mistake

karussell: but not important to add a test as we have your nice PR soon and throw this away :)

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1093,10 +1096,15 @@ public GHResponse route( GHRequest request )
         QueryResult fromQResult = qResults.get(0);
 
         double weightLimit = request.getHints().getDouble("defaultWeightLimit", defaultWeightLimit);
+        int maxVisistedNodesForRequest = request.getHints().getInt("routing.max_visited_nodes", maxVisitedNodes);
+        // TODO Do we want to throw an exception at this point?
```
karussell: yes, I would prefer throwing an exception or even better: adding an error to the response

karussell:
Thanks - looks really good!

> I think we have an issue with our current travis configuration. I forgot to run the 
> whole test-suite before pushing and it failed for me locally. Since the js tests succeed, 
> travis seems to think that all tests succeed. 

Do you mean that although tests are failing locally, travis is fine with it because only the JS tests are passing?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1093,10 +1096,15 @@ public GHResponse route( GHRequest request )
         QueryResult fromQResult = qResults.get(0);
 
         double weightLimit = request.getHints().getDouble("defaultWeightLimit", defaultWeightLimit);
+        int maxVisistedNodesForRequest = request.getHints().getInt("routing.max_visited_nodes", maxVisitedNodes);
```
devemux86: Here the parameter shouldn't be `"max_visited_nodes"` ?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -686,6 +687,8 @@ public GraphHopper init( CmdArgs args )
 
         // routing
         defaultWeightLimit = args.getDouble("routing.defaultWeightLimit", defaultWeightLimit);
+        maxVisitedNodes = args.getInt("routing.maxVisitedNodes", Integer.MAX_VALUE);
```
devemux86: The default value can probably be written as `maxVisitedNodes` to follow current pattern.

boldtrn:
> Do you mean that although tests are failing locally, travis is fine with it because only the JS tests are passing?

Yes, have a look at this test. It fails but travis does not fail https://travis-ci.org/graphhopper/graphhopper/jobs/114835803#L1152

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: problem is that we are very inconsistent with this and it has a reason: from URL or JS one expects `some_thing` from within Java one expects `someThing`. What we could do is internally convert all `someThing` into `some_thing` or migrate all remaining `someThing` to my preferred naming `some_thing` "somehow"

karussell:
> Yes, have a look at this test.

Outch - ugly! Thanks for finding this!

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: Yes sure, we can change this.

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: Yes we can change the name in both cases (config and url) so we do not have the `routing.` in front. I thought it would be better understandable since the number of parameters is growing.

core/src/main/java/com/graphhopper/routing/RoundTripAltAlgorithm.java
boldtrn: Ah yes, we can actually delete the isMaxVisitedNodesExceeded method, since it is not used here. We delegate the maxVisitedNodes to the Dijkstra and the Dijkstra cares about the maxVisistedNodes.

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: Ok sure, I change that.

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: The `routing` in front is good as indeed the count is growing. Also please for now use `someThing` notation - I've created #682 to make everything more consistent or to be able to use both.

boldtrn: Ok, so I shall use camel case in the url?

boldtrn:
Ok, I updated my branch. I renamed the URL parameter using camelCase.

karussell:
Thanks a lot! One minor in the exception -> routing.max_visited_nodes. Would you also squash the commits into one?

boldtrn:
Squashed it

karussell:
Thanks - merged!

PR #690, opened 2016-03-25
Gradle: update Android build tools to 23.0.3

devemux86:
Update Android build tools to latest version.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'com.android.application'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.2"
+    buildToolsVersion "23.0.3"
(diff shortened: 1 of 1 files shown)


karussell:
Thanks! Does the maven config also need an update?

devemux86:
No because in GH we build the Android sample only with Gradle.
Maven would require external [Maven Android SDK Deployer](https://github.com/simpligility/maven-android-sdk-deployer) procedures.

devemux86:
If you want we can create an issue and I'll see what can be done.

Though it isn't something automatic, users would have to update manually their local repo first.

karussell:
okay

BTW: I deploy the apk via maven and the sdk deployer via the pom.xml and it works (if one installs stuff before this ;))

PR #691, opened 2016-03-28
Updated SRTM URL to kurviger.de

boldtrn:
Since I renamed my motorcycle route planer to Kurviger I had to change the base URL to [Kurviger.de](https://kurviger.de/). The old url will be still valid for the foreseeable future.


karussell:
Somehow I'm getting in a branch `CertificateException: No subject alternative DNS name matching srtm.kurviger.de found` for the https version (and also when trying to connect via the browser this fails). Should we just use http?

boldtrn:
Sorry, should be fixed.

PR #693, opened 2016-03-29
Set ContentType to application/gpx+xml

boldtrn:
The ContentType for .gpx files should be `application/gpx+xml`. Also see here for more details: https://en.wikipedia.org/wiki/GPS_Exchange_Format

I had a small amount of users that received the file as `filename.gpx.xml`. It seems like this fixes the issue.

### DIFF
--- a/web/src/main/java/com/graphhopper/http/GraphHopperServlet.java
+++ b/web/src/main/java/com/graphhopper/http/GraphHopperServlet.java
@@ -199,7 +199,7 @@ public class GraphHopperServlet extends GHBaseServlet
         boolean withTrack = getBooleanParam(req, "gpx.track", true);
         boolean withWayPoints = getBooleanParam(req, "gpx.waypoints", false);
         res.setCharacterEncoding("UTF-8");
-        res.setContentType("application/xml");
+        res.setContentType("application/gpx+xml");
         String trackName = getParam(req, "trackname", "GraphHopper Track");
         res.setHeader("Content-Disposition", "attachment;filename=" + "GraphHopper.gpx");
         long time = getLongParam(req, "millis", System.currentTimeMillis());


karussell:
Hmmh, difficult to decide as there are lots of applications outside that we would need to re-test again. See duplicate #676

boldtrn:
Sorry, haven't seen the other ticket. I changed it for [kurviger.de](https://kurviger.de/) and wanted to commit the improvement back to GraphHopper. One major decision towards that content type was also this [link](https://trac.openstreetmap.org/ticket/4768). I think it should not break any other applications. But we still have to test it :(.

karussell:
Maybe @ratrun you have some experience which applications we really should test. At least 2 online tools, 2 navis and this Garmin application which I cannot remember anymore. Hmmh, but wait: the file type should not influence the bytes (?) so if both are equal we should be able to change this even without much testing?

boldtrn:
@karussell 

This only changes the http header information of the content-type. This should only influence services that download gpx files from GraphHopper. The file itself should stay the same.

Some browser/pc/setting combination changes the file ending of the downloaded file to .xml. 

BTW: I guess you mean BaseCamp.

karussell:
> BTW: I guess you mean BaseCamp

ah, yes.

So, we just need to make sure that the API clients still work properly.

karussell:
Ok, I've asked one and the client still works with this change. 

Can you add this fallback if the client is explicitly asking for it?

```java
if(req.getContentType().equals("application/xml"))
 res.setContentType("application/xml")
else
 res.setContentType("application/gpx+xml")
```

karussell:
Fixed

boldtrn:
Thanks. :+1:

Issue #695, opened 2016-04-04
bicycle=dismount is not evaluated


polarbearing:
While graphhopper knows the idea of pushing a bike on ```path, footway, pedestrian and steps```, the explicit access tag ```bicycle=dismount``` (used about [40000 times](http://taginfo.openstreetmap.org/tags/bicycle=dismount)) is not evaluated. If found on any highway, it should create such [pushing section](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/BikeFlagEncoder.java#L54). 

E.g. I have a ```highway=secondary``` here which is for a long time ```access=no, foot=yes, bicycle=dismount```.

karussell:
Thanks that is indeed missing!

Regarding your example if access=no and foot=yes is already set GraphHopper should already tell you about getting off the bike. Or do you have a example link?

polarbearing:
See [here](https://graphhopper.com/maps/?point=52.457788%2C13.625622&point=52.457003%2C13.62545&locale=en&vehicle=bike&weighting=fastest&elevation=true&layer=Omniscale) and switch between bike, car and foot.

karussell:
Ah, okay [this way](http://www.openstreetmap.org/way/43167434). Maybe the current implementation would use 'get off the bike' if there is no access=no and just a foot:yes. We need to check this and make it working.

polarbearing:
Yes, I should have given you the way number. 

The evaluation order in OSM is typically that first access=no forbids everything, then individual permissions are opened case by case. Thus you do not need a long list of *=no tags. 
Thanks.

polarbearing:
Just for the records, I just removed a bunch of unnecessary turn restrictions from a newbie a the northern end of the road segment. But it seems the issue is still within the road, if I get the start/stop point [narrower only on the affected segment](https://graphhopper.com/maps/?point=52.457435%2C13.625547&point=52.457055%2C13.625482&locale=en&vehicle=foot&weighting=fastest&elevation=true&layer=Omniscale), I get the same results.

ratrun:
I think that this issue can be closed, it should be solved for a while.

PR #696, opened 2016-04-04
Add support for bicycle:dismount.

ratrun:
Add support for the vehicle tag for bicycles and foot.
Theses changes were triggered by issue #695.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -44,7 +44,7 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
     protected static final int PUSHING_SECTION_SPEED = 4;
     private long unpavedBit = 0;
     // Pushing section heighways are parts where you need to get off your bike and push it (German: Schiebestrecke)
-    protected final HashSet<String> pushingSections = new HashSet<String>();
+    protected final HashSet<String> pushingSectionsHighways = new HashSet<String>();
     protected final HashSet<String> oppositeLanes = new HashSet<String>();
     protected final Set<String> preferHighwayTags = new HashSet<String>();
     protected final Set<String> avoidHighwayTags = new HashSet<String>();
@@ -69,7 +69,7 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
     {
         super(speedBits, speedFactor, maxTurnCosts);
         // strict set, usually vehicle and agricultural/forestry are ignored by cyclists
-        restrictions.addAll(Arrays.asList("bicycle", "access"));
+        restrictions.addAll(Arrays.asList("bicycle", "vehicle", "access"));
         restrictedValues.add("private");
         restrictedValues.add("no");
         restrictedValues.add("restricted");
@@ -261,7 +261,7 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder
             return 0;
 
         // use the way if it is tagged for bikes
-        if (way.hasTag("bicycle", intendedValues))
+        if (way.hasTag("bicycle", intendedValues) || way.hasTag("bicycle", "dismount"))
             return acceptBit;
 
         // accept only if explicitly tagged for bike usage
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
```diff
@@ -261,7 +261,7 @@ public long acceptWay( OSMWay way )
             return 0;
 
         // use the way if it is tagged for bikes
-        if (way.hasTag("bicycle", intendedValues))
+        if (way.hasTag("bicycle", intendedValues) || way.hasTag("bicycle", "dismount"))
```
karussell: Should we just add 'dismount' to the intendedValues here?

core/src/main/java/com/graphhopper/routing/util/BikeFlagEncoder.java
```diff
@@ -86,6 +86,7 @@ boolean isPushingSection( OSMWay way )
         String trackType = way.getTag("tracktype");
         return way.hasTag("highway", pushingSections)
                 || way.hasTag("railway", "platform")
+                || way.hasTag("bicycle", "dismount")
```
karussell: here we should use `isPushingSection` instead (?)

karussell:
Thanks! I think I'll have time this week to look through all the nice work of you :) Sorry for all the delay ...

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
ratrun: I had this first. It is not a good idea for two reasons:
1) The code in `getSpeed` in BikeCommonFlagEncoder.java would require at least one more 'if'
2) It is not logical. A way with `bicycle=dismount` is not really intended for bicycle usage such as the other values `designated` or `official`.

core/src/main/java/com/graphhopper/routing/util/BikeFlagEncoder.java
ratrun: Thank you. This is a good optimisation. I have updated the PR with this included.

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
karussell: makes sense!

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
```diff
@@ -428,12 +428,22 @@ int getSpeed( OSMWay way )
...
+            {
+                if (way.hasTag("highway", "steps"))
+                    speed = PUSHING_SECTION_SPEED / 2;
+                else
+                    speed = PUSHING_SECTION_SPEED;
+            }
             else
-                speed = PUSHING_SECTION_SPEED;
+            {
+                if (way.hasTag("bicycle", "dismount"))
```
karussell: instead of the separate 'if' maybe just 
```java
if (way.hasTag("highway", pushingSections) || way.hasTag("bicycle", "dismount"))
```
this way we would handle the steps too?

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
```diff
@@ -679,7 +689,7 @@ public double getDouble( long flags, int key )
 
     boolean isPushingSection( OSMWay way )
     {
-        return way.hasTag("highway", pushingSections) || way.hasTag("railway", "platform");
+        return way.hasTag("highway", pushingSections) || way.hasTag("railway", "platform") || way.hasTag("bicycle", "dismount");
```
karussell: this is now a bit confusing isPushingSection vs. `pushingSection` variable: maybe the variable should be renamed to `pushingSectionHighways`?

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
ratrun: Thank you. This was a very good capture. It was even possible to reduce it further with elimination if one more if as we didn't have a test for fast steps. I made some more test for it.

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
ratrun: This was an excellent capture. I was able to reduce the code complexity even further  - and added two more tests for the steps.

karussell:
Sorry again for the delay - now merged!

Issue #644, opened 2016-01-20
Wrong graph folder unnecessary creation


devemux86:
When we call `graphHopper.load` with a wrong path that doesn't exist, the process informs us with an exception about that, but also creates unnecessarily the path.

That's inconvenient as we have to remove the unneeded created empty folder afterwards.
Can we skip the `GHDirectory.mkdirs` in case the graph loading has failed?

karussell:
Sure, that is ugly.

ammagamma:
https://github.com/graphhopper/graphhopper/pull/700

karussell:
Should be fixed with #700

karussell:
Found a way to fix this and keep backward compatibility: https://github.com/graphhopper/graphhopper/commit/0de10bfd2ba7bd33c7e9980a73ad00b66b7a6898

More complex than #700 :(. See #724 for discussion why this is still necessary.

devemux86:
Nice! :+1: 
Also the needed removal of the `mkdirs` in `GHDirectory` constructor.

PR #700, opened 2016-04-13
Issue #644

ammagamma:
Prevents the creation of an empty folder when loading a Graph from a non-existent path.

The directory is created in GraphHopper.load() at line 837:

`        GHDirectory dir = new GHDirectory(ghLocation, dataAccessType);`

The directory is created before it is checked if the graph loading was successful. Therefore we can not simply omit the mkdirs call if the graph loading failed unless there would be a possibility to create the storage without creating the directory).
Also GraphHopper.load() is not only used to load an existing graph, but also in GraphHopper.importOrLoad() to initialize the GHStorage etc. (the actual loading is done only later in Graphhopper.process() in this case). 
The solution in this pull request is the least invasive I could come up with.
I check for the existence of the path in load() and then call the old load() method (under the new name
"initializeStorage") if the path exists. Otherwise false is returned. I am not entirely sure if initializeStorage is a good name and there seem to be quite a few things going on in this method.

By the way no exception is thrown if the path does not exist upon calling load(), but only false is returned. The user gets the confusing exception: "Call load or importOrLoad" when calling GraphHopper.route(), which can be seen in GraphHopperTest.testNoNPE_ifLoadNotSuccessful(). This is confusing because the user called load() before but still gets this exception.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -697,7 +697,7 @@ public class GraphHopper implements GraphHopperAPI
      */
     public GraphHopper importOrLoad()
     {
-        if (!load(ghLocation))
+        if (!initializeStorage(ghLocation))
         {
             printInfo();
             process(ghLocation);
@@ -793,6 +793,14 @@ public class GraphHopper implements GraphHopperAPI
      */
     @Override
     public boolean load( String graphHopperFolder )
+    {
+        if (!(new File(graphHopperFolder).exists()))
+            return false;
+
+        return initializeStorage(graphHopperFolder);
+    }
+
+    private boolean initializeStorage(String graphHopperFolder)
     {
         if (Helper.isEmpty(graphHopperFolder))
             throw new IllegalStateException("graphHopperLocation is not specified. call init before");
--- a/core/src/test/java/com/graphhopper/GraphHopperTest.java
+++ b/core/src/test/java/com/graphhopper/GraphHopperTest.java
@@ -455,6 +455,15 @@ public class GraphHopperTest
     }
 
     @Test
+    public void testDoNotCreateEmptyFolderIfLoadingFromNonExistingPath()
+    {
+        instance = new GraphHopper().
+                setEncodingManager(new EncodingManager("CAR"));
+        assertFalse(instance.load(ghLoc));
+        assertFalse(new File(ghLoc).exists());
+    }
+
+    @Test
     public void testFailsForMissingParameters() throws IOException
     {
         // missing load of graph


karussell:
Thanks! 

Would you mind [sending](https://graphhopper.com/#contact) me your email, so that we can send you the [CLA](https://graphhopper.com/agreements/individual-cla.html) & sign it? See our [contributing docs](https://github.com/graphhopper/graphhopper/blob/master/.github/CONTRIBUTING.md#license-agreement) why :)

ammagamma:
Ok, sent it :)

karussell:
Thanks - merged!

> By the way no exception is thrown if the path does not exist upon calling load(), but only false is returned. 

Hmmh, indeed. As we now have a bit better separated the load and importOrLoad method, could we throw an exception in load but none in importOrLoad if no path exists?

ammagamma:
Yes, I think we can throw an exception, I just was not sure because it is so close to 
the API.

karussell:
Loading from an empty dir is a caller mistake (at least if no tests say sth. different ;)) and an exception can be thrown. But importOrLoad should not throw an exception if load fails due to an empty dir.

If you have&take the time I would appreciate a fix :)

ammagamma:
We would have to change GraphHopperTest.testNoNPE_ifLoadNotSuccessful() (what does NPE mean btw ?), but as I said above this test is misleading anyway because it asserts that the exception says load was not called even though it was called. Of course we would also have to change GraphHopperTest.testDoNotCreateEmptyFolderIfLoadingFromNonExistingPath(), which was just added in this PR, because now an exception would be expected. Ok, I will create a pull request for this.

karussell:
> what does NPE mean btw 

NullPointerException :)

> this test is misleading anyway because it asserts that the exception says load was not called even though it was called.

Hmmh, not sure I understand what you mean. The test should verify that if the route method is called without a successfully loaded graph, that then the exception with 'Call load or import..' is thrown, the message could be a bit more clear and should say 'Do a successful call to load or import' or something :)

ammagamma:
Ok, changing the error message to "do a successful load .." would be clearer also in my opinion. If we throw an exception in load when passing a non-existing path this test won't work anymore because this exception would occur when calling load() already instead of later when calling route(). We would have to have some kind of unsuccessful load, which is not due to a non-existing path. For example we could use an existing empty directory / simply create the ghLoc before calling load.
In this case we still assert that an unsuccessful load is detected in route() and yields an informative error message (for example instead of yielding a NPE), which as I understand it now is the intention of the test.

karussell:
> which as I understand it now is the intention of the test.

Yes, I think the graph storage was null without a check in `route()` and resulted in an not very informative NPE.

ammagamma:
[#702](https://github.com/graphhopper/graphhopper/pull/702)

Issue #665, opened 2016-02-17
Curvature Weighting produces Exception for Path#calcMillis


boldtrn:
As discussed [here](https://discuss.graphhopper.com/t/calculating-time-should-not-require-to-read-speed-from-edge-in-wrong-direction/509/9), there are cases that result in an Exception using the CurvatureWeighting.

Interesting is that in these cases the speed for that edge is 0, but we never set an edge to speed 0 using the getSpeed() method in the CarEncoder.

The weighting returns 0. And we get the following Exception:
```
java.lang.IllegalStateException: Calculating time should not require to read speed from edge in wrong direction. Reverse:false, fwd:false, bwd:true
	at com.graphhopper.routing.Path.calcMillis(Path.java:266)
	at com.graphhopper.routing.Path.processEdge(Path.java:238)
	at com.graphhopper.routing.PathBidirRef.extract(PathBidirRef.java:95)
	at com.graphhopper.routing.DijkstraBidirectionRef.extractPath(DijkstraBidirectionRef.java:132)
	at com.graphhopper.routing.AbstractBidirAlgo.calcPath(AbstractBidirAlgo.java:64)
	at com.graphhopper.routing.AbstractRoutingAlgorithm.calcPaths(AbstractRoutingAlgorithm.java:121)
	at com.graphhopper.GraphHopper.calcPaths(GraphHopper.java:1327)
```

To reproduce this Exception, load the current germany.osm.pbf. And visit this URL: http://localhost:8989/?point=48.873748%2C8.412781&point=49.099864%2C7.993197&locale=de-DE&vehicle=motorcycle&weighting=curvature&elevation=false&layer=Omniscale

ammagamma:
I think I made a little progress with this and reproduced the error in a unit test. I figured out that there is a single edge that causes the error and when I looked up the coordinates I found out that its the ferry crossing the Rhine river near Karlsruhe :) 

A routing request from one riverside to the other only works  (again using the map for all of Germany from Geofabrik) when going from West to East but not the other way around. 

This works:
http://localhost:8989/?point=48.978881%2C8.25488&point=48.977273%2C8.2569&locale=de-DE&vehicle=motorcycle&weighting=curvature&elevation=false&layer=Omniscale

This does not:
http://localhost:8989/?point=48.977273%2C8.2569&point=48.978881%2C8.25488&locale=de-DE&vehicle=motorcycle&weighting=curvature&elevation=false&layer=Omniscale

[...]

Issue #367, opened 2015-03-13
Speed zero should leed to access=no


zhiqiangZHAO:
looks like something wrong here:
https://graphhopper.com/maps/?point=38.942477%2C-77.418699&point=38.95018%2C-77.404784&locale=en-US&layer=Lyrk

zhiqiangZHAO:
looks like something wrong with an edge: 
http://www.openstreetmap.org/edit?editor=id&way=141900856#map=18/38.95066/-77.40282

zhiqiangZHAO:
I am test car with turn restriction.

zhiqiangZHAO:
The problem is related speed=0.

zhiqiangZHAO:
https://github.com/graphhopper/graphhopper/issues/242

The original edge has a tag: max_speed=1, I remove it.

There shouldn't be an edge with 1 MPH for car in reality.

In Graphhopper, the speed shouldn't be 0.
If a edge with speed 0, should change it to unnavigable.

karussell:
Thanks for reporting and investigating. I think we should set this as unnavigable in the import if speed is 0, otherwise we would have to check the speed on every read which is too expensive

zhiqiangZHAO:
I prefer to ignore maxspeed=1, it's ridiculous.

Yes. Definitely there shouldn't be an edge with speed=0.
map matching found it as the origin, but it's not searchable due to speed=0.

zhiqiangZHAO:
what I have saw the real maxspeed is 5 MPH.
Usually it's in parking lot.
How could be 1 MPH, it's slower than walk.

karussell:
> I prefer to ignore maxspeed=1, it's ridiculous

Hmmh, not sure. At least we'll set accessibility to no in case of speed is 0

zhiqiangZHAO:
That's also good.
At least the road with very low speed are very rare.

zhiqiangZHAO:
btw, OSRM don't have this problem. 
http://www.openstreetmap.org/directions?engine=osrm_car&route=38.95103%2C-77.40297%3B38.95082%2C-77.40479#map=18/38.95102/-77.40388

http://www.openstreetmap.org/directions?engine=osrm_car&route=38.95111%2C-77.40283%3B38.95082%2C-77.40479#map=18/38.95102/-77.40381

I guess this due to the compression of speed in GH.

karussell:
They do not take into account the road probably due to speed == 0

[...]

PR #703, opened 2016-04-17
adds tests reproducing the error described in issue665

ammagamma:
Adds three tests: 
1) creates the error described in [#665](https://github.com/graphhopper/graphhopper/issues/665) using OSM data and the GraphHopper API

2) shows that the Graph resulting from the OSM data is invalid in this scenario

3) reproduces the invalid Graph and shows how the error is produced using the bidirectional Dijkstra algorithm between the two nodes of the Graph directly

### DIFF
--- a/core/src/test/java/com/graphhopper/reader/OSMReaderTest.java
+++ b/core/src/test/java/com/graphhopper/reader/OSMReaderTest.java
@@ -36,8 +36,10 @@ import org.junit.Before;
 import org.junit.Test;
 
 import com.graphhopper.GraphHopper;
+import com.graphhopper.GHRequest;
 import com.graphhopper.reader.dem.ElevationProvider;
 import com.graphhopper.reader.dem.SRTMProvider;
+import com.graphhopper.routing.DijkstraBidirectionRef;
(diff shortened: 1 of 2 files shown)


ammagamma:
I simply put all three tests inside OSMReaderTest for now.

karussell:
Thanks again!

The bug is that we do not set the reverse speed in the motorbike encoder and so the `speed==0` leads to `access==false` due to the fix for #367 when calling `reverseFlags`. This method is called for 'opposite' base nodes ie. when the flags are stored relatively to the adjacent node and needs swap all speeds and access properties. I'm not yet sure why this happens only that infrequent.

In the Bike2WeightFlagEncoder we handle reverse speed via calling a special method `handleSpeed` in `handleWayTags` which makes sure both speeds are always properly filled.

So the real reason is that adding 'reverse speed' is overly complicated and too many things can go wrong when adding this to an encoder.

But probably also the magic happening in the fix for #367 is not good.

karussell:
Thanks a bunch again to @ammagamma for making it very easy for me to reproducing this. Closing due to #711

Issue #131, opened 2013-12-11
Distance matrix ( Origin / Destination )


matsjg:
It would be nice to use Graphhopper to calulate a distance matrix (or Origin / Destination) matrix from a list of nodes.
This is useful for route optimization. 

Either by repetitive use of one-to-all shortest path algorithms like Dijkstra's or maybe use an all-to-all shortest path algorithms such as the Floyd-Warshall algorithm.

IsNull:
I need the distance matrix feature for a mean of 100x100 (all-to-all) queries. _The matrix will then be further processed by genetic algorithms for a complex delivery planning and routing problem._

Now before I'm going to implement this in my Fork I wanted to investigate into the current situation of this matter here. 

* Wouldn't it make sense that GraphHopper provides a (non-optimised) default implementation of the Matrix API? (For the sake of a common API...)
* Since you already have implemented an optimised version of this and holding it back for a reason, what would happen if I send you a pull-request with matrix functionality? 😄 

This leads to an interesting situation for an (semi) open source project - how to deal with contributions which partially cover closed source features which where intended to be paid for?

That being said, I can somehow understand that you want to keep portions back of your code. On the other hand, this will always lead to such problems, especially if your community grows. Thank you for providing GraphHopper Open Source anyway 👍

karussell:
> Wouldn't it make sense that GraphHopper provides a (non-optimised) default 
> implementation of the Matrix API? (For the sake of a common API...)

That is a good idea, yes.

> Since you already have implemented an optimised version of this and holding it back for a reason, what would happen if I send you a pull-request with matrix functionality? :smile:

That is a very good question and for the last >2 years there was no need to answer this ;)

We will of course not reject it and integrate it if our standards like CLA and unit&integration tests are in place. Commercially it could temporarily make sense to reject this, but our aim is to push our open source projects with the help of our business/company and not the other way around (easier said than done!). Furthermore the 'danger' of a fork would then be higher and we really want to avoid this mess for the sake of our community & quality etc.

> That being said, I can somehow understand that you want to keep portions back of your code. On the other hand, this will always lead to such problems, especially if your community grows.

We do not see this as a real problem. In fact we will have to think about closed vs. open features from time to time but our business model is not founded on closed source although it makes it a bit easier at the start with closed source. BTW: see [here](https://graphhopper.com/#os-comparison) for a current comparison.

> Thank you for providing GraphHopper Open Source anyway :+1:

Thanks for this kind words!

[...]

Issue #682, opened 2016-03-10
CamelCase vs. under_score notation


karussell:
We should allow both or migrate to just under_score notation. E.g. prepare.ch_weightings vs. prepare.ch_weightings or prepare.minNetworkSize vs. prepare.min_network_size

Also check CH properties for better naming in general:
 * `routing.ch.heading.force`  instead of `force_heading_ch`
 * `prepare.ch.weightings` instead of `prepare.chWeightings`

Others: 

 * a bit ugly: `alternative_route.max_paths` vs. `algorithm=alternativeRoute` 
 * make all routing options with or without a `routing.` prefix. See `alternative_route` vs. `routing.flexMode.xy`
 * what about the names in storableproperties
 * names in measurement properties result
 * translation keys

devemux86:
In long term it'd be probably better to have only one notation.
But what about API current compatibility?

karussell:
At the moment I tend to convert all get and put calls with camelCase to under_score and rewrite all put requests and the config to use under_score. This way we always use one notation internally, but keep compatibility.

karussell:
I'll soon implement this where all of the old settings should work as before (except maybe rarely used like force_heading_ch) or other options introduced in 0.7-SNAPSHOT

The plan is to

 * switch all properties to under score
 * use parameters without `routing.`

for all the areas mentioned above.

What do you think?

devemux86:
I think it's a good plan. :+1:

karussell:
Please have a look into #719

PR #705, opened 2016-04-21
Distance matrix support (WIP)

IsNull:
This pull request implements a basic API for distance matrix queries as discussed in #131. As such, I want to review the proposed API before refining it and adding tests. 

Please have a look and tell me if the proposed API goes into the right direction. I'd like to avoid putting too much work into something you don't consider for merge. You find the primary code under the package /matrix in the core project.

**Edit**
- Added some unit tests
- Adapted to refactorings


IsNull:
The `DijkstraOneToMany`algorithm returns zero for some distances when the cache is used. I am not sure if I have stumbled over a Bug or if I just used it wrong. Current usage is simply as follows:

```java
        DistanceMatrix matrix = new DistanceMatrix();

        DijkstraOneToMany algorithm = (DijkstraOneToMany)routingFactory.createAlgo(graph, underlyingAlgo);


        for (int origin : origins) {

            algorithm.clear(); // Clear cache on different start node
            DistanceMatrix.DistanceRow row = matrix.addRow(origin);

            for (int destination : destinations) {
                Path path = algorithm.calcPath(origin, destination);
                row.addDestination(destination, path.getDistance(), path.getTime());
            }
        }
```

I.e. the algorithm is reused with cache for all destinations of a single origin. For each new origin, clear() is called since the documentation states that its necessary to clear the cache for different start nodes.

The code works if `DijkstraOneToMany` is created as new instance for each usage, but this obviously is not the idea.

karussell:
If you feel if that is a bug I need a simple unit test showing it. Also keep in mind that DijkstraOneToMany is not thread safe, so for our use case you would need the same approach as the other algorithms (destroy queue etc on every request) or you need manage it in ThreadLocal (more efficient but more RAM usage and more complex). Furthermore DijkstraOneToMany won't work for CH (at least not now)

IsNull:
I can not follow you completely where you see a multithreading issue. For each user request a new `MatrixAlgorithm` instance is being created, the only thing shared between Threads would be the `MatrixAlgorithmFactory` which is thread safe since it does not use any class fields and just returns new instances.

> Furthermore DijkstraOneToMany won't work for CH (at least not now), so I would recommend a simple but new dijkstra class similar to the others and thread safe.

Ok that is a problem, because  using the DijkstraOneToMany would make it impossible to use the CH optimisations. 

Before I am able to implement a custom Dijkstra in need however to fully understand the threading issues. Could you please explain in Detail how the algorithm is not thread-safe, if for each usage a new instance is created?

Thanks for your time :)

karussell:
> Could you please explain in Detail how the algorithm is not thread-safe, if for each usage 
> a new instance is created?

Then there is no problem, but if you look into the data structures then you'll see that recreating per request is very expensive in contrast to the very cheap recreation for the normal algorithms.

IsNull:
Understood. As for the matrix API itself, this seems to be less of an issue. Usually, the matrix api will be used with few requests, but each request is very expensive. So a single initialisation per request is not too big of an issue.

Anyway, the DijkstraOneToMany implementation is now working, at least the very basic unit tests confirm this. 

See #707, #708 for the DijkstraOneToMany bug.

karussell:
> So a single initialisation per request is not too big of an issue IMHO.

That would be the wrong approach. Please keep in mind that creating such a data structure is really very expensive and does not make sense to be recreated for every request, instead just see the other algorithms how this is handled. Also the Matrix API can be used for one-to-10 or just 10-to-10 requests not only for heavy weight stuff.

IsNull:
Yea I actually prefer clean solutions as well. I just have not fully grasped all the design decisions in the code. For example,  there are many `checkAlreadyRun();` in your algorithms, so by design these algorithms can only be used once. So I assumed that an algorithm usually only runs once.

I guess we are not talking about the same initialisation? 

Now the  `DijkstraOneToMany` is a different story, as it can repeatedly be used and does actually prepare memory for all nodes 

```java
        parents = new int[graph.getNodes()];
        Arrays.fill(parents, EMPTY_PARENT);

        edgeIds = new int[graph.getNodes()];
        Arrays.fill(edgeIds, EdgeIterator.NO_EDGE);

        weights = new double[graph.getNodes()];

        Arrays.fill(weights, Double.MAX_VALUE);
```
Are these the  `complex data structures` you are referring to? Just ensuring we are on the same page.

The more I think about it, it would probably make sense to move the cache part out in a shared, Thread-safe component. The cache obviously has to support different start nodes, otherwise its likely that the cache does not help much. This cache can probably be generalised and does not necessary have to be for a specific algorithm. Its probably similar to the CH, just not calculated up front but something like a cache for already calculated path partials / shortcuts etc. Maybe I just reinvent something or this approach is otherwise unfeasible... Any thoughts on this? 

Before inventing a new cache/data-structure & algorithm, I will do also a bit research on existing algorithms & data-structures which are better suited for repeated queries.

karussell:
> Are these the complex data structures you are referring to? Just ensuring we are on the same page.

Yes, although 'complex' is more 'expensive'

> The cache obviously has to support different start nodes

Ah, yes. [Virtual nodes](https://github.com/graphhopper/graphhopper/blob/master/docs/core/low-level-api.md#what-are-virtual-edges-and-nodes) are another problem which you would get with reusing.

> Before inventing a new cache/data-structure & algorithm

I wouldn't go this more complex way as the first trial. Just put enough tests that later on one can refactor simple to more complex solutions with the same results.

IsNull:
>I wouldn't go this more complex way as the first trial. Just put enough tests that later on one can refactor simple to more complex solutions with the same results.

Yep that was the idea in the first place. I almost got carried away :) So I'd say for a first version, I will leave the DijkstraOneToMany as it is now, but it should only be used for large matrices. Im not going to hack in any threading since this would lead to memory leaks - for every thread which accesses the matrix api, a new DijkstraOneToMany would be in memory. So this needs quite some thought and I think we need to design this from ground up.

I will add some more tests and do some clean up. If you are going to merge the refactoring I will adopt the code as necessary.

IsNull:
I am reiterating about the matrix API and performance thoughts:

* The CH's should be used since they give a massive performance boost
* The CH already acts as a `global cache` for routing algorithms
* We do not really need to expand the full path - only the time/distance is interesting for further optimisations

So if in shortcuts of your CH implementation the the distance/time is stored, its getting very fast since the full path does not need to get expanded.

karussell:
The Matrix should work for both cases (with and without CH). But I would just start with the simple interface via the double for loop and then improve from there.

IsNull:
@karussell Yes, lets keep this pull request simple in terms of the provided matrix algorithms.

There are two open points which we probably should address before a merge can be considered:

* The creation/handling of the `AlgorithmOptions` is not optimal and also has some code redundancies with `GrahpHopper.java` Could you have a look at this - to solve this we probably have to refactor some portions out of the `GrahpHopper.java` class.

* The `GHRequest`/`GHResponse` are also similar to the `GHMatrixRequest`/`GHMatrixResonse`. Would it make sense to refactore this to a common  `GHRequest`/`GHResponse` base, which then probably also gets used in the new code from the point above?

core/src/main/java/com/graphhopper/matrix/DistanceMatrixService.java
```diff
@@ -0,0 +1,186 @@
...
+        List<QueryResult> originQNodes = lookupNodes(request.getOrigins(), algoOpts.getFlagEncoder());
+        List<QueryResult> destinationQNodes = lookupNodes(request.getDestinations(), algoOpts.getFlagEncoder());
+
+        List<QueryResult> merged = new ArrayList<>(originQNodes);
+        merged.addAll(destinationQNodes);
+
+        queryGraph.lookup(merged);
+
+        List<Integer> originNodes =  mapToNodes(originQNodes);
+        List<Integer> destinationNodes =  mapToNodes(destinationQNodes);
```
karussell: Why do we need this conversion step?

core/src/main/java/com/graphhopper/matrix/DistanceMatrixService.java
```diff
@@ -0,0 +1,186 @@
...
+
+        MatrixAlgorithm algorithm = matrixAlgorithmFactory.build(queryGraph, algoOpts);
+        DistanceMatrix matrix = algorithm.calcMatrix(originNodes, destinationNodes); // TODO
```
karussell: What is the TODO here?

core/src/main/java/com/graphhopper/matrix/DistanceMatrixService.java
karussell: Ah, okay. You are separating the low level API from the higher one and just require the algorithm to know 'nodes', no need for QueryResults.

core/src/main/java/com/graphhopper/matrix/DistanceMatrixService.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * Builds the AlgorithmOptions
+     * // TODO Refactor: The following is almost copy & paste from GraphHopper.calcPoints()
+     *
+     * @param request The matrix request
+     * @param hopper Instance of hopper
+     * @return
+     */
+    private AlgorithmOptions.Builder buildOptions(GHMatrixRequest request, GraphHopper hopper){
```
karussell: Yeah, it is a good idea to separate this in GraphHopper as well.

core/src/main/java/com/graphhopper/matrix/algorithm/DijkstraOneToManyMatrixAlgorithm.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * This implementation uses the DijkstraOneToMany algorithm, which can reuse data structures as long
+ * as the same start-node is in use.
+ *
+ * This means we reuse the algorithm/cache per row.
+ * I.e. for each origin, there is a new DijkstraOneToMany / cache created.
+ *
+ * @author Pascal Büttiker
+ */
+public class DijkstraOneToManyMatrixAlgorithm extends AbstractMatrixAlgorithm {
```
karussell: As said before: I would not want to have this as an option as this explodes for large areas. Instead a simple one-to-many approach via the normal Dijkstra or bidir. Dijkstra class could be implemented.

core/src/main/java/com/graphhopper/matrix/algorithm/MatrixAlgorithm.java
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * Results in quadratic complexity on-top of the underlying route algorithm.
+     * Not very efficient.
+     */
+    String OneToOneLoop = "one-to-one-Loop";
+
+    /**
+     * This implementation uses for each node a one-to-many Dijkstra
+     */
+    String OneToManyDijkstra = "one-to-many-Dijkstra";
```
karussell: here we should move this to AlgorithmOptions and use names without dashes and camel case. See also #682

core/src/main/java/com/graphhopper/matrix/algorithm/OneToOneLoopMatrixAlgorithm.java
```diff
@@ -0,0 +1,63 @@
...
+ * assuming a quadratic matrix (origins == destinations)
+ *
+ * For Dijkstra this is O( (|E|+|V|log(|V|))^2 )
+ *
+ * @author Pascal Büttiker
+ */
+public class OneToOneLoopMatrixAlgorithm extends AbstractMatrixAlgorithm {
+
+    private final RoutingAlgorithmFactory routingFactory = new RoutingAlgorithmFactorySimple();
+    private final AlgorithmOptions underlyingAlgo;
```
karussell: naming 'underlyingAlgo' is misleading. Furthermore we can disable calculating instructions and paths in the hints

karussell:
> Would it make sense to refactore this to a common GHRequest/GHResponse base, which then probably also gets used in the new code from the point above?

Subclassing is convenient but has the downside of forcing us to use the same stuff. I would for now go a separate and code duplicative way. 

Keep in mind that we later want an improved version of this where we potentially have hundreds of points and need a possibility to implement the GHMatrixResponse very memory efficient e.g. in raw arrays like `times[][]; distances[][]` instead of "row objects"

core/src/main/java/com/graphhopper/matrix/DistanceMatrixService.java
IsNull: Its no longer required, I will remove it.

core/src/main/java/com/graphhopper/matrix/algorithm/DijkstraOneToManyMatrixAlgorithm.java
IsNull: Yeah its still there since i did some experimenting. I'll remove it.

Probably you should add a javadoc warning to`DijkstraOneToMany`, or even a hard limit + throw an exception if its that problematic.

core/src/main/java/com/graphhopper/matrix/algorithm/MatrixAlgorithm.java
IsNull: I didnt use the AlgorithmOptions since the MatrixAlgorithm & Factory was completely separated, but probably we can have them in there any way. Maybe we just prefix matrix algorithms with `Matrix`?
I.e.:  `MatrixOneToOneLoop` ?


As a side note: I wonder if it is a good idea to have the algorithm configurable for the matrix. We could probably choose the best algorithm automatically based on Graph Size and query input?

core/src/main/java/com/graphhopper/matrix/algorithm/OneToOneLoopMatrixAlgorithm.java
IsNull: > naming 'underlyingAlgo' is misleading.

Yes, it used to be the algorithm instance. I'll rename it.

>Furthermore we can disable calculating instructions and paths in the hints

Yes we should disable this generally for matrix requests. Where can I find the flag keys for this?

core/src/main/java/com/graphhopper/matrix/DistanceMatrixService.java
IsNull: Well I need the QueryResults due to the necessary `queryGraph.lookup(merged);` to properly initialise the queryGraph.

But since this code is almost 1:1 identical to the GraphHopper default route Graph initialisation, I really think this should be refactored in a shared Graph-Preparation class.

IsNull: Well I need the QueryResults due to the necessary `queryGraph.lookup(merged);` to properly initialise the queryGraph.

But since this code is almost 1:1 identical to the GraphHopper default route Graph initialisation, I really think this should be refactored in a shared Graph-Preparation class.

core/src/main/java/com/graphhopper/matrix/algorithm/MatrixAlgorithm.java
karussell: > As a side note: I wonder if it is a good idea to have the algorithm configurable for the matrix. We could probably choose the best algorithm automatically based on Graph Size and query input?

Yes, maybe that would be best for now. Later one could have a heuristic value to lower quality and improve speed or something

core/src/main/java/com/graphhopper/matrix/algorithm/OneToOneLoopMatrixAlgorithm.java
karussell: See the new Parameters class -> Routing.CALC_POINTS and Routing.INSTRUCTIONS

IsNull:
I have merged your refactoring and adapted my code so it works with it. 

>Keep in mind that we later want an improved version of this where we potentially have hundreds of points and need a possibility to implement the GHMatrixResponse very memory efficient e.g. in raw arrays like times[][]; distances[][] instead of "row objects"

Sounds reasonable, I will refactor the distance matrix response to use an array based data structure.

IsNull:
I've refactored the matrix data structures to 2d arrays for maximum performance and minimal memory consumption. I've also collapsed the internal used DistanceMatrix class and the external visible GHMatrixResponse, which now uses the DistanceMatrix as property.

It is optimised so it can be directly turned into JSON by Jackson or a similar mapper. (Well thats not important for you since you apparently do this stuff manually in your web service anyway.)

IsNull:
The `GHMatrixResponse` currently only holds the calculated matrix. We can later add more information such as completion state, runtime duration etc. if required.

Anyway, please have a look and tell me what needs more work in order to consider merging this PR.

core/src/main/java/com/graphhopper/matrix/DistanceMatrix.java
```diff
@@ -0,0 +1,124 @@
...
+        if(weights != null) weights[originIndex][destIndex] = weight;
+    }
+
+    /**
+     * Gets the distance matrix cell for a single (origin --> destination) entry
+     * @param originIndex The origin index
+     * @param destIndex The destination index
+     * @return THe cell of the origin-destination
+     */
+    public DistanceMatrixCell getCell(int originIndex, int destIndex){
```
karussell: here I would avoid creating an object, even if it is on-demand and even if the JVM is fast with these things :)

I would instead create three methods like `getDistance(int, int)`, see similar work I've done very recently to improve the memory usage of the directions api [matrix client](https://github.com/graphhopper/directions-api-java-client/blob/master/client/src/main/java/com/graphhopper/api/MatrixResponse.java#L79)

core/src/main/java/com/graphhopper/matrix/algorithm/MatrixAlgorithmFactory.java
```diff
@@ -0,0 +1,20 @@
...
+ * @author Pascal Büttiker
+ */
+public interface MatrixAlgorithmFactory {
+
+    /**
+     * Creates a MatrixAlgorithm instance using the given options
+     * @param g The Graph to run the algorithm on
+     * @param opts The options for the algorithm
+     */
+    MatrixAlgorithm build(Graph g, AlgorithmOptions opts );
```
karussell: I would use the same naming `createAlgo` for now? (`RoutingAlgorithm createAlgo( Graph g, AlgorithmOptions opts );`)

core/src/main/java/com/graphhopper/matrix/algorithm/OneToOneLoopMatrixAlgorithm.java
```diff
@@ -0,0 +1,64 @@
...
+                origins.length, destinations.length,
+                true,  // Include distances
+                true,  // Include times
+                true); // Include weights
+
+        for (int i=0;i<origins.length;i++) {
+
+            int origin = origins[i];
+            for (int j=0;j<destinations.length;j++) {
+                RoutingAlgorithm algorithm = routingFactory.createAlgo(graph, routingAlgoOptions);
```
karussell: is `calc_paths=false` and `instructions=false` set somewhere?

core/src/test/java/com/graphhopper/matrix/AbstractMatrixAlgorithmTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Base class for matrix tests, takes MatrixAlgorithm implementations and runs tests against them.
+ */
+public abstract class AbstractMatrixAlgorithmTest {
+
+    private static final double DELTA = 0.000001;
```
karussell: `.1` is sufficient (and no constant necessary) as distance is in meters.

core/src/test/java/com/graphhopper/matrix/OneToOneLoopMatrixAlgorithmTest.java
```diff
@@ -0,0 +1,26 @@
...
+import com.graphhopper.routing.util.ShortestWeighting;
+import com.graphhopper.routing.util.TraversalMode;
+
+import static com.graphhopper.util.Parameters.Algorithms.MATRIX_ONE_TO_ONE;
+
+
+/**
+ * Tests the OneToOneLoopMatrixAlgorithm
+ */
+public class OneToOneLoopMatrixAlgorithmTest extends AbstractMatrixAlgorithmTest {
```
karussell: Is CH and non-CH possible and tested this way?

karussell:
Thanks for the work!

> It is optimised so it can be directly turned into JSON by Jackson or a similar mapper

This is indeed nice and I should have done similar with the normal routing objects too. For really big matrices (>1000²) it can be beneficial if you stream the results directly into the JSON, of course no requirement for the first shot but we should keep this in mind and make it possible at some point.

Two questions: is the web integration already here? And as it is using normal routing, is it working&tested for CH and non-CH?

core/src/main/java/com/graphhopper/matrix/DistanceMatrix.java
IsNull: Ok np, I'll refactor it.

core/src/main/java/com/graphhopper/matrix/algorithm/OneToOneLoopMatrixAlgorithm.java
IsNull: No I forgot that one. Added it now see here: https://github.com/graphhopper/graphhopper/pull/705/commits/4ab0c347d30ea898779724e9ccbaaefb884f6f70

IsNull:
To your last two questions:

I did not do any tests related CH / non-CH yet. I'm even not quite sure how such a test should look like? I assumed since I just reuse normal routing this will work out of the box since the normal routing is already tested properly?

As a side note: The project where I want to use this is currently in "pre production", meaning that I'm still evaluating stuff and not working primarily on it. This is why you see this infrequent progress :). Anyway, as soon that we get startet on it we will use the distance matrix api for real world scenarios and have some production grade tests on top of it. The performance of it will also be evaluated then.

To the web integration: Honestly, I am not sure if im going to do this. We will internally create a new web service based on spring boot which just uses the graphhopper core. (Writing a solid rest api takes very few lines with spring boot) On top of that, spring boot delivers a lot of integrations for free, which we need for our Microservice architecture: Health/Metrics, Netflix component integration such as Eureka and Feign and probably also some authentication. Some additional features such as configuring/uploading an osm file and a general UI for configuration will also be required.

We probably make this project open source, but since it will be mission critical it has to be under our full control.

karussell:
> I assumed since I just reuse normal routing this will work out of the box since the normal routing is already tested properly?

Yes, still important to assert what **should** work, probably not now but very certainly later on.

> Honestly, I am not sure if im going to do this

It is important to have and define some structure like the matrix java API defines the structure, also without a web component it wouldn't be useful for half of the GraphHopper users I fear. Furthermore the Matrix API json structure is even a lot easier than the routing api

IsNull:
> Yes, still important to assert what should work, probably not now but very certainly later on.

Agreed. I will certainly do some manual testing soon when integrating graphhopper. If you help me with improved unit tests which test both CH and non-CH I'm happy to work on this as well.


>It is important to have and define some structure like the matrix java API defines the structure, also without a web component it wouldn't be useful for half of the GraphHopper users I fear. Furthermore the Matrix API json structure is even a lot easier than the routing api

Well a matrix request could potentially be a quite long running operation, putting the system under stress. The way I'm going to implement this is by designing an async API which returns immediately after a matrix request with a job-id, which can be polled later on. Also, the number of concurrent running matrix requests should be limited (maybe 2-3). There needs to be a queuing system which gets async handled, and this goes hand in hand with the async rest api.

So implementing this just quick and dirty might pose some stability threads.

The thing is, implementing very basic support would be 4 lines of code if there would be a framework in place. Spending time manually parsing a request etc. just seems a waste of time to me - just compare the below code which probably is enough for a rest API in Spring (or alternatives like Jersey):

```java
@RestController
@RequestMapping("/api/matrix")
public class MatrixResource {
    @Autowired
    private MatrixService matrixService;
    
    @RequestMapping(
        method = RequestMethod.GET,
        produces = MediaType.APPLICATION_JSON_VALUE)
    public ResponseEntity<GHMatrixResponse> getAll(@RequestBody GHMatrixRequest request) {
        return matrixService.matrix(request);
    }
}
```

Anyway, matrix api support in the core is a feature on its own. If someone needs the matrix api in the web project, by all means, he can contribute to this project as well :)

I hope you can understand my point.

karussell:
> So implementing this just quick and dirty might pose some stability threads.

For sure, but eg. serializing json is a subproblem and I would want the simple version of this all in the web module if we have matrix calculation in the core

> Spending time manually parsing a request etc. just seems a waste of time to me

I agree. But those frameworks come to a cost: the learning curve and there are many to choose from, e.g. we use dropwizard internally. And so I would for now provide simple GET requests for the matrix ("reference implementation"). We can add simple tests and if one day we move to a simpler framework like dropwizard or spring boot those tests should still work.

> If someone needs the matrix api in the web project, by all means, he can contribute to this project as well :)

I have to disagree. I want to make most of the features available and how should we advertise that matrix is now included but no one except Java devs can use it? There are features which have no explicit UI like the round trip feature, although one can still make it work using some parameters. But there is no feature (or very few) which are not accessible from the web module. Trust me: having this in the web module from the beginning is very important.

karussell:
Regarding framework: I think convenience is really important especially if you have many REST endpoints, but we don't have that and for us performance is more important. For me this is one of the reasons to stay a bit more low level. See [here](https://www.techempower.com/benchmarks/#section=data-r12&hw=peak&test=query). (Probably more important for simple a-to-b routing than it is for matrix requests).

aj1310:
Is the matrix support still on the roadmap? I could contribute if help is needed.

kalexmills:
Ditto to alewaros' comment. Looking to gain this support for my own graphhopper instance.

TimPigden:
Hi, we're very interested in using a one-to-many. but it looks like this has faded away over the issue of web support. Is someone able to comment?

karussell:
Seems like the original creator lost interests, see https://discuss.graphhopper.com/t/one-to-many-requests/2183

Issue #707, opened 2016-04-25
DijkstraOneToMany cache broken after calcPath(from, to) where from == to


IsNull:
While working on the distance matrix, I stumbled over a little bug in
DijkstraOneToMany. If you invoke `DijkstraOneToMany.calcPath(from, to)` and from / to is the same node, all future calls to `calcPath(from, to)` will yield a zero path/distance.

This is obviously a rare case, since it does not really make sense to query for a path to itself, but for the sake of a clean framework I think such edge cases should be handled cleanly anyway.

See the pull request #708 which shows the issue.

karussell:
Fixed via #708

PR #708, opened 2016-04-25
Dijkstra one to many zero distance issue

IsNull:
Unit test which exposes the issue #707


karussell:
See the merged fix.

A cleaner fix would have been to move the `finished` method to:
```java
if (finished())
  return currNode;
 if (heap.isEmpty() || isMaxVisitedNodesExceeded() || isWeightLimitExceeded())
  return NOT_FOUND;
```

but this would add a (minor) overhead for nearly all other calculations, which for now seems to be unnecessary.

PR #709, opened 2016-04-26
Improvement for roundtrip calculation

karussell:
This PR improves the integration of pull request #582. You can try this out via appending `&algorithm=roundTrip&round_trip.seed=0&round_trip.distance=10000&routing.flexibleMode.force=true` to the URL. Change the seed to introduce some randomness.

The separation seems to work and the round trip looks (still) good :)

car-good

I have created an interface RoutingTemplate and there are 3 implementations: via routing, round trip and alternative. The implemetations are stateful and not thread-safe (which both does not really matter), I could also get rid of the AlgorithmControl.

Also the `algorithm=roundtrip|alternativeRoute` is now a bit fake as it should more be something like `routing_template=roundtrip|alternativeRoute` and then we would have the option to additionally control the used algorithm under the hood. For now I vote to keep it as it is and let the roundtrip implementation choose the underlying routing algo.

Quality-wise this looks already good, also the distance approximation is good. But there are two minor issues one with the the tail route and one with the motorway problem:

roundtour-increase-area

There are some `TODO NOW`s

Also I'm not sure about the naming `AlternativeRoute` (append Algorithm?) vs. `AlternativeRouting` (the template).


core/src/main/java/com/graphhopper/routing/util/AvoidPathWeighting.java
```diff
@@ -0,0 +1,81 @@
...
+    public AvoidPathWeighting( Weighting superWeighting )
+    {
+        super(superWeighting);
+    }
+
+    /**
+     * This method adds the specified path to this weighting which should be penalized in the
+     * calcWeight method.
+     */
+    public void addPath( Path path )
```
karussell: I would vote for providing a `add(Collection<EdgeIteratorState>` method here instead, then we could use this weighting for more cases, not just paths.

core/src/main/java/com/graphhopper/routing/util/tour/TourStrategy.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Returns the distance in KM that is used for the generated point #iteration
+     */
+    public abstract double getDistanceForIteration( int iteration );
+
+    /**
+     * Returns the bearing between 0 and 360 for the current iteration.
+     * <p>
+     * TODO NOW make identical orientation as the heading
```
karussell: Here I mean this heading definition https://github.com/graphhopper/graphhopper/blob/master/docs/core/routing.md#heading

karussell:
> There is only one minor difference for the round trip test:

This is due to an improvement of the situation as now virtual nodes are introduced for all points and before they were only junction-precise (except maybe the start)

core/src/main/java/com/graphhopper/routing/util/tour/TourWayPointGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+        tour.add(startIndex);
+        return tour;
+    }
+
+    private static QueryResult generateValidPoint( GHPoint from, double distanceInKm, double bearing, LocationIndex locationIndex, EdgeFilter edgeFilter, int triesAvailable )
+    {
+        int counter = 0;
+        while (true)
+        {
+            GHPoint generatedPoint = CoordinateProjection.projectCoordinate(from.getLat(), from.getLon(), distanceInKm, bearing);
```
karussell: If we retry here, then there is no change for from, distanceInKm nor bearing. Should we really retry this much or make retry procedure simpler, like always reducing `distanceInKm = distanceInKm / 2` or something?

core/src/main/java/com/graphhopper/routing/util/tour/TourStrategy.java
karussell: @boldtrn is this also north based? then it should be already identical I think (?)

boldtrn: Yes this is north based. The idea is to get heading per iteration. I create that for every iteration.

The idea is to project a point with a given heading/bearing for every iteration. This changes for every iteration.

karussell:
See my comment here https://github.com/graphhopper/graphhopper/pull/709#discussion_r61255859 and the solution in RoundTripRoutingTemplate where maxRetries can be specified in the config.properties. The retry is without randomness and reduces distance a bit, we could instead also slightly change the heading as well, not sure.

core/src/main/java/com/graphhopper/routing/template/RoundTripRoutingTemplate.java
```diff
@@ -0,0 +1,180 @@
...
+        this.ghResponse = ghRsp;
+        this.locationIndex = locationIndex;
+        this.maxRetries = maxRetries;
+    }
+
+    @Override
+    public List<QueryResult> lookup( List<GHPoint> points, FlagEncoder encoder )
+    {
+        double distanceInMeter = ghRequest.getHints().getDouble("round_trip.distance", 1000);
+        long seed = ghRequest.getHints().getLong("round_trip.seed", 0L);
```
boldtrn: How about not using the seed at all, if no seed was passed to the algorithm?

core/src/main/java/com/graphhopper/routing/util/tour/TourWayPointGenerator.java
boldtrn: Yes, one try should be enough. I haven't changed that when migrating the code. My old solution looked a bit different.

boldtrn:
> we could instead also slightly change the heading as well, not sure.

Yes, changing the heading is a good idea as well.

I had a look at your improvement. In general it looks good and flexible.

karussell:
> How about not using the seed at all, if no seed was passed to the algorithm? Or generating a random seed if no seed is available?

The problem about this server-side randomness would be that we loose reproducibility, i.e. all requests should return identical results (for the same GH version) and having an explicit seed in the parameters makes this possible. Without this you cannot really debug or reproduce issues anymore (easily).

> I had a look at your improvement. In general it looks good and flexible.

Okay & Thanks! 

Should we do the heading stuff in another PR? And also it would be nice to hear that your custom routing stuff from kurviger works with this PR or what needs to be changed.

core/src/main/java/com/graphhopper/routing/template/RoundTripRoutingTemplate.java
```diff
@@ -0,0 +1,180 @@
...
+        GHPoint start = ghRequest.getPoints().get(0);
```
boldtrn: As far as I can see, we don't check if there is at least one point in ponts. We should add this somewhere.

karussell:
Updated to latest master and added a point.size check. I will review once again and merge afterwards if no objections.

karussell:
Merged

karussell:
There is a new parameters `round_trip.points` creating support points of the round trip and also the `heading` parameter influences the direction of the round trip. See https://github.com/graphhopper/graphhopper/commit/130870905667ca602552fbb6f39b3ff1c3dc5bfc

PR #710, opened 2016-04-27
Android Sample: update Gradle

devemux86:
- Update Gradle plugin 2.1.0
- Include `gradle.properties` for [dexing-in-process](http://developer.android.com/tools/revisions/gradle-plugin.html#revisions)

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -1,0 +1,19 @@
+# Project-wide Gradle settings.
+
+# IDE (e.g. Android Studio) users:
+# Gradle settings configured through the IDE *will override*
+# any settings specified in this file.
+
+# For more details on how to configure your build environment visit
+# http://www.gradle.org/docs/current/userguide/build_environment.html
+
+# Specifies the JVM arguments used for the daemon process.
+# The setting is particularly useful for tweaking memory settings.
+# Default value: -Xmx10248m -XX:MaxPermSize=256m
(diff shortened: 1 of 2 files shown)


android/gradle.properties
```diff
@@ -0,0 +1,19 @@
...
+# any settings specified in this file.
+
+# For more details on how to configure your build environment visit
+# http://www.gradle.org/docs/current/userguide/build_environment.html
+
+# Specifies the JVM arguments used for the daemon process.
+# The setting is particularly useful for tweaking memory settings.
+# Default value: -Xmx10248m -XX:MaxPermSize=256m
+# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx2048m
```
karussell: oh, wow. Is that really necessary so high?

devemux86: Yes I'm afraid it's the official instruction for enabling dexing-in-process, see [here](http://developer.android.com/tools/revisions/gradle-plugin.html#revisions) (I updated the PR doc).

devemux86: It's just an option for faster builds.
If you feel uncomfortable with that we can remove it.

karussell: I think this would just fail in my personal case having open other apps as well :)

PR #711, opened 2016-04-29
Reproducing & fixing reverse speed error

karussell:
This fixes the bug in #665. The work to find this was done in #703.

This bug could be reproduced also for `bike2` which is very strange as I've never seen this failing in 'real world'.

The important part now is to set the flags for both speeds explicitly via
```java
flags = handleSpeed(way, ferrySpeed, flags);
```

or in motorcycle encoder via setSpeed&setReverseSpeed.

This is explicitly mentioned now in the create-new-flagencoder.md guide but the whole process should be made easier later on too.


karussell:
Ah, before merging: I'll have to move the OSMReader independent test to a bike2 graph-only test. Just for clarity...

PR #714, opened 2016-05-05
Gradle properties: use default settings

devemux86:
After testing (also in Mapsforge), let's stay on the safe side with the default settings.

Users can always use custom settings via their profiles.


karussell:
Thanks!

PR #722, opened 2016-05-15
Validate the start of a roundtrip

boldtrn:
Currently we do not validate the start of a round trip. This can lead to an empty response. Caused by an exception:

```
java.lang.IllegalStateException: Do not call QueryGraph.lookup with invalid QueryResult
```

This PR validates the start point.


karussell:
Thanks!

PR #730, opened 2016-05-31
initial work towards generic weighting&encoding ('more flexible' work)

karussell:
See [the forum](https://discuss.graphhopper.com/t/more-flexibility-in-gh/71) for more discussion.

There are lots of TODOs and currently no integration tests, so it is not time to merge but to play around for you. You can currently configure the average speed of the highways the (car-like) vehicle should use. E.g. overwrite residential, tertiary and secondary via [this request](http://localhost:8989/?point=52.462913%2C13.387871&point=52.529486%2C13.399372&vehicle=generic&weighting=generic&highways.secondary=35&highways.tertiary=30&highways.residential=30).


core/src/main/java/com/graphhopper/routing/util/GenericFlagEncoder.java
```diff
@@ -0,0 +1,478 @@
...
+
+        maxPossibleSpeed = 140;
+
+        //
+        // TODO do something about the access values (forestry, agricultural, destination, private, delivery, customers)
+        //
+        List<String> highwayList = Arrays.asList(
+                /* reserve index=0 for unset roads (not accessible) */
+                "_default",
+                "motorway", "motorway_link", "motorroad",
```
boldtrn: Why do you define motorrad here? I thought this would not be not an highway value - http://overpass-turbo.eu/s/idk

core/src/main/java/com/graphhopper/routing/util/GenericFlagEncoder.java
```diff
@@ -0,0 +1,478 @@
...
+                "trunk", "trunk_link",
+                "primary", "primary_link", "secondary", "secondary_link", "tertiary", "tertiary_link",
+                "unclassified", "residential", "living_street", "service", "road", "track",
+                "ferry", "shuttle");
```
boldtrn: I also could not find anything about shuttle in OSM, is this description more general to also support non-osm sources?

karussell: Sorry, this is a typo: should be shuttle_train http://wiki.openstreetmap.org/wiki/Proposed_features/shuttle_train

boldtrn: There is no need to be sorry :)

But I am wondering: this is used in the route tag and not the highway tag, but we never look in the route tag for it (or I missed it?).

karussell: > There is no need to be sorry :)

:) sure :+1: !

Yes, this is used in the route tag but I was trying to avoid a new tag key here as there are no combinations of e.g. motorway and ferry or with shuttle. This is just to avoid using yet another set of bits.

boldtrn: Yes I am with you in that point. But I mean we are not reading the route tag yet in this PR, right?

But what about the motorroad, see my comment above?

boldtrn: Thanks for your answers :), helped me a lot to understand what's happening.

devemux86:
We cannot combine the regular encoders (e.g. car) with the `DataFlagEncoder`'s features like store of highway tags?
I mean using a `CarFlagEncoder` and retrieving the road type of an edge, as `generic` appears as an independent vehicle for the graph.

karussell:
Not yet, but this is one of the remaining tasks.

PR #736, opened 2016-06-11
Android Sample: update to Mapsforge 0.6.1

devemux86:
We released [Mapsforge 0.6.1](https://groups.google.com/forum/#!topic/mapsforge-dev/F_uBzeJ4Thc).

Let's update GraphHopper sample with that.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -41,10 +41,10 @@ dependencies {
        exclude group: 'org.apache.xmlgraphics', module: 'xmlgraphics-commons'
     }
 
-    compile group: 'org.mapsforge', name: 'mapsforge-core', version: '0.6.0'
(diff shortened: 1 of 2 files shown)


devemux86:
We should change also the download link to `0.8` graphs.

Though I don't find them to work, with the error "Encoder foot was used in version 1, but current version is 2".

karussell:
Ups. Thx & will do :)

karussell:
I'm a bit lazy, now we do not need manual editing anymore ;)
https://github.com/graphhopper/graphhopper/commit/dc16f31e53c3aba75e8c011306e83326c7e65651

devemux86:
:+1: 
Any automation is handy for our life as developers!

devemux86:
I updated the PR resolving the conflicts.

devemux86:
Peter shouldn't we put also here an `Android` label for easy reference?

PR #740, opened 2016-06-12
Separate core module from osm reading

### LINKED ISSUE #450: Create separate source module for import and core

karussell:
Fixes #450 e.g. removing xml dependencies from model classes (node, way, relation). Moves several OSM related classes and tests into the module and makes it possible to avoid this overhead for Android. The breaking change now is that all code requiring OSM import has to change

 * maven dependency from `core` to `reader-osm`
 * change `GraphHopper` to `GraphHopperOSM`

But this makes integrating like #616 very easy e.g. in a submodule without effecting current dependencies which I really like.

The OSM parsing stuff is still included in the FlagEncoders like converting units to SI units or workarounds for certain OSM issues as well as parsing strings into objects. Such a change would be more intense and requiring us to convert the OSM data into some kind of intermediate object format. Another option would be to define a subset of OSM which is stricter and less complex but still valid OSM and use this as the intermediate format. But this is for another issue...


devemux86:
Analysing this very useful (and big) modules decoupling, will the elevation data import via `tif` files remain as part of  `graphhopper-core` module (along with `xmlgraphics-commons` dependency)?

karussell:
Currently yes as I'm not sure how the dependencies should look like. E.g. should it be `readerosm->elevation->core` or `elevation->readerosm->core` or `elevation->core && readerosm->core`?

devemux86:
Core should be independent (if possible) from all graph creation business.

But what you describe is a good point for thinking, i.e. how the elevation and various readers interact with each other..

karussell:
BTW: maybe we should use `GraphHopper` for osm stuff for backward compatibilty and the core graphhopper class will then be `GraphHopperCore` or something or keep it as it is `GraphHopperOSM extends GraphHopper`?

devemux86:
Hmm nice thought. For clear API the 2nd solution (with data suffix) appears more appropriate i.e. `GraphHopperOSM` extends `GraphHopper`.
Or ease we could have the irregularity of `GraphHopper` for `osm` and `GraphHopperShp` for `shp`.

karussell:
> Or we could have the irregularity of GraphHopper for osm and GraphHopperShp for shp?

Indeed, this would be too ugly. So let us break and we already report the user what to do: https://github.com/graphhopper/graphhopper/pull/740/files#diff-59d362881b641ff1b933a3785503c20eR774

karussell:
At least for the elevation module this could work as follows. E.g. in the current Import file (`tools` module) we can do:

```java
public class Import
{
    public static void main( String[] strs ) throws Exception
    {
        CmdArgs args = CmdArgs.read(strs);
        GraphHopper hopper = new GraphHopperOSM().init(args);
        // one line glue code of OSM and elevation module 
        hopper.setElevationProvider(ElevationHelper.createElevationProvider(args));
        hopper.importOrLoad();
        hopper.close();
    }
}
```

The same would be necessary in the web module. The downside is that we would need this glue code and also explicitely define the elevation module as separate dependency besides `reader-osm` in all downstream dependencies which need osm import with elevation. Not sure if this is okay.

Another possibility to avoid the glue code would be very simple reflection code asking for elevation and if not existing falls back to `ElevationProvider.NOOP`. I would try to avoid this as e.g. iOS will have problems with this but maybe it is not a problem at all when we provide the possibility to do this by hand.

karussell:
Merged now.

PR #749, opened 2016-06-22
Add more translations

fbonzon:
[Live demo](http://françois.ch/graphhopper/?locale=fr-CH&layer=OpenStreetMap)


web/src/main/webapp/js/map.js
```diff
@@ -164,12 +164,12 @@ function initMap(bounds, setStartCoord, setIntermediateCoord, setEndCoord, selec
         },
         contextmenu: true,
         contextmenuItems: [{
-                text: 'Route ',
```
karussell: here is a space after 'Route' ... maybe we should use a parametrized translation. And did you add those keys to they spreadsheet? If not, I can do it.

karussell:
Thanks - looks good!

web/src/main/webapp/js/map.js
fbonzon: Added the space after "Route".

Same question about parametrized translation can be raised below for keys "start_label", "intermediate_label", "end_label", where a number is added after the translated string + space.

I don't think parametrized translation is necessary here.

Yes, all these keys were already added on the spreadsheet a few months ago.

fbonzon:
Addressed your comment.

karussell:
Thanks!

Issue #272, opened 2014-10-24
'undefined' in location. GraphHopper Maps


karussell:
Very rarely there occures 'undefined' in the location input. It was happening after (right) click something on the markers or so.

sebalis:
There’s a very clear way of triggering this: change the transport mode by clicking on Foot or Bicycle etc. while the input fields for start and destination are still empty. This causes the fields to be filled with the string 'undefined'. Unfortunately there are some map objects that actually have this name. So when one of the two fields (say, start) is filled in with actual data, the web interface thinks that the other field has also been set, to a place in Zimbabwe. Example: Change to foot routing while the fields are empty, enter „Harare“ in the start field, choose the Harare (Zimbabwe) item that is offered – you are given a route before you entered your destination.

karussell:
This should be fixed with #1180 - will deploy it in a few minutes.

PR #750, opened 2016-06-22
Translate Zoom control tooltips and autocomplete results

fbonzon:
See the **[live demo](http://xn--franois-xxa.ch/graphhopper/?locale=fr-CH&layer=OpenStreetMap)**

The menu with autocomplete suggestions now displays geographic names in the same language as the other UI elements. As mentioned in the [documentation](https://graphhopper.com/api/1/docs/geocoding/#url-parameters), currently only French (fr), English (en), German (de) and Italian (it) are supported. Other languages will continue to display English names.

Translation strings for "zoom in" and "zoom out" have already been committed earlier today.


web/src/main/webapp/js/autocomplete.js
```diff
@@ -1,5 +1,6 @@
 var formatTools = require('./tools/format.js');
 var GHInput = require('./graphhopper/GHInput.js');
+var GHRoute = require('./graphhopper/GHRoute.js');
```
fbonzon: Fixing unrelated bug here. This is required for line 71 below.

web/src/main/webapp/js/autocomplete.js
```diff
@@ -81,7 +83,9 @@ AutoComplete.prototype.createGeocodeURL = function (ghRequest, prevIndex) {
     var path = this.createPath(this.host + "/geocode?limit=6&type=" + this.dataType + "&key=" + this.key);
     if (prevIndex >= 0 && prevIndex < ghRequest.route.size()) {
         var point = ghRequest.route.getIndex(prevIndex);
-        path += "&point=" + point.lat + "," + point.lng;
+        if (point.isResolved()) {
```
fbonzon: avoids `point=undefined,undefined` in the URL

karussell:
Thanks!
Finally this old bug gets resolved :) #272

web/src/main/webapp/js/translate.js
```diff
@@ -6,11 +6,11 @@ var mathTools = require('./tools/math.js');
 
 function tr2(key, args) {
     if (key === null) {
-        log("ERROR: key was null?");
+        console.log("ERROR: key was null?");
         return "";
     }
     if (defaultTranslationMap === null) {
-        log("ERROR: defaultTranslationMap was not initialized?");
```
karussell: Thanks, it looks like `log(` is no longer defined (?)

We should then fix the several other places too (maybe another issue)

web/src/main/webapp/js/main-template.js
```diff
@@ -158,6 +159,8 @@ $(document).ready(function (e) {
 
                 // execute query
                 initFromParams(urlParams, true);
+
+                checkInput();
```
karussell: Is this necessary for the `undefined` bug?

web/src/main/webapp/js/translate.js
fbonzon: Correct, log() is undefined. There are more instances, but are all commented out, so I did not fix those.

web/src/main/webapp/js/main-template.js
fbonzon: No, unrelated. This has to do with the order of things happening.

When the ready() function is executed, the autocomplete menu for the first 2 locations fields (empty From and To) is already finished, and still has "en" locale. One must wait until the API call to retrieve translations to know the locale.

fbonzon:
#272 is not solved here. It is a different problem unfortunately.

I am solving bad URLs sent to geocoding API, that had a bad location bias. Inspect network traffic in the browser developper console. It happens for example with the 2nd field ("To") when both location fields are initially empty.

web/src/main/webapp/js/main-template.js
fbonzon: The checkInput(); on line 207 below now happens before the new instance here. It becomes redundant, and could probably be removed.

PR #757, opened 2016-07-27
Hint change for Java API

stevendlander:
`config.properties` refers to a request hint that must be set to force the use of flexible mode.  The hint mentioned in the comment does not appear to be correct (does not work).  The correct hint should be `routing.ch.disable=true`


karussell:
Thanks!

Issue #500, opened 2015-08-22
Geometry limited by 2^31


karussell:
As the geometry ref is a positive integer only and we point into the DataAccess object with this int (without any conversion) there is a limit of 2^31 ints encoded from a list of "lat,lon,ele" which could be reachable for the world wide case when all ways are included.

See [this mailing list post](https://lists.openstreetmap.org/pipermail/graphhopper/2015-August/002324.html) by @jtheuer

karussell:
We increased the limit to (2^32-1) which makes this issue a lot less critical. Will still let it open.

TODO: 
 - [ ] recreate 0.6 maps for Android demo

highsource:
Hi,

I'm trying to get my head around this issue.

From what I see in `BaseGraph`, it uses `DataAccess` (a kind of abstract byte storage) to save bytes. `PointList` gets encoded into a list of `lat`, `lon` and possibly `ele`, 4 bytes each which are then saved in `wayGeometry` altogether as a byte array.

I'm not quite sure where the limit is. Is it the type of `geoRef`? I see this:

```
        long geoRef = Helper.toUnsignedLong(edges.getInt(edgePointer + E_GEO));
```

Is the problem that `geoRef` is stored as unsigned int? Why can't we store it as long instead? Probably because it will add 4 bytes per edge and that would be too much of memory? How many edges do we have? If it's only 178 094 237 (not sure about the number above), 4 extra bytes per edge will be around 700 MB, that does not seem too much.
We can add just one or two bytes if it's critical, that would increase the capacity to 2^40/2^48 which will be more than enough - without significantly increasing the size of the edge storage. Or am I missing something?

Or is there a limit of DataAccess size like 4GB file size or something?

> To reduce severity of this issue one can additionally reduce the storage usage by:

This actually does not increase the possible space of `geoRef` but reduces the rate of `geoRef` growth, right? The size of the byte array of the point list will be smaller so `geoRef` will grow slower.

I think we can win 1 bit if we don't save `len * dim` but just `len`. `dim` is 2 or 3 so this would give at least 1 bit.

Concerning proposed solutions:

> * reusing the area when calling setWayGeometry of existing edge, if new point list is smaller

I think this might be quite hard to implement correctly.

> * using delta encoding #234

Quite an approach.

> * use only 2 or 3 bytes for the elevation data instead of 4

I think this is a pretty good idea. Who needs elevation with 32-bit precision?

> * use negative values too (convert into 'unsigned int' -> long)

[...]

PR #769, opened 2016-08-17
When updating geometry, reuse area of existing edge #500

highsource:
Solves a part of #500 by reusing the geometry area of existing edge if the new point list is not larger.

I had to expose maxGeoRef to be able to test it, I hope it's ok.

I've also decided it would be good to erase existing contents if new point list is smaller than existing. Otherwise we'll have some old coordinates laying around.

Another improvement would be to store the maximum size of the area. Then it would be possible to reuse it even if point list grows but stays beyond that maximum size. Would require storing max size next to count which is probably not so good.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/BaseGraph.java
+++ b/core/src/main/java/com/graphhopper/storage/BaseGraph.java
@@ -28,8 +28,6 @@ import com.graphhopper.search.NameIndex;
 import com.graphhopper.util.*;
 import static com.graphhopper.util.Helper.nf;
 
-import java.util.Arrays;
-
 import com.graphhopper.util.shapes.BBox;
 
 /**
@@ -839,11 +837,6 @@ class BaseGraph implements Graph
 				final int count = wayGeometry.getInt(existingGeoRefPos);
             	if (len <= count)
             	{
-            		// If old array was smaller, erase contents
-            		if (len < count)
-            		{
-            			eraseWayGeometryAtGeoRef(count, existingGeoRef);
-            		}
             		setWayGeometryAtGeoRef(pillarNodes, edgePointer, reverse, existingGeoRef);
                     return;
             	}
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -822,41 +829,81 @@ private void setWayGeometry_( PointList pillarNodes, long edgePointer, boolean r
...
-                    tmpOffset += 4;
-                }
+            	long existingGeoRefPos = existingGeoRef * 4L;
+				final int count = wayGeometry.getInt(existingGeoRefPos);
+            	if (len <= count)
+            	{
+            		// If old array was smaller, erase contents
+            		if (len < count)
+            		{
+            			eraseWayGeometryAtGeoRef(count, existingGeoRef);
```
karussell: I know it is cleaner: but why would we need to clear this?

highsource: I'm ok to drop the erasure. Zeroes would make it easier to see empty parts. It would also make data blocks better compressible. But that's not really necessary.

highsource: Should I remove the erasure and send another PR?

karussell: You should be able to just do another commit in your branch and it should show up here. But I'll do the code formatting and merge this with your changes (+could remove the method) - let me know what to do :)

Regarding compressibility: that is indeed true. But I'm unsure how we can apply compression to the geometry and still keep everything else working & fast.

highsource:
Ok, will do. And yes, you'll need to reformat. I used default Eclipse code formatter, but will probably switch to IntelliJ later on.

Compressibility - I was not thinking about realtime compression, more about storage compression on disc. But let's drop it for now.

karussell:
> Compressibility - I was not thinking about realtime compression, more about storage compression on disc. But let's drop it for now.

We 'store' stuff in-memory (via byte or int arrays) in the identical format we store data on disc. So reading stuff into memory is very simple and also switching to memory mapped. E.g. for memory mapped we read edge properties from a DataAccess but the same happens for an in-memory DataAccess. So at the end there is (currently) no difference between storage compression and realtime compression ... I hope I have not misunderstood what you mean :)

highsource:
Yes, I understand that and you got me right. I was just thinking of use cases like data transfer where you download the prepared files of the graph as a zipped package and the unzip them. But really, let's drop it, it does not seem to be a direct use case at the moment.  
I'll send the fix shortly.

karussell:
> But really, let's drop it, it does not seem to be a direct use case at the moment.

Ah, I now understand what you mean: that is indeed true and could lead to smaller compressed files. We are currently doing compressed graphhopper files for Android (".ghz").

highsource:
I've removed erasure for now. It's easy to add it later on.

karussell:
Thanks a lot for this fix! I've just applied minor code formatting ;) and moved the test into the abstract tester, so that all graph implementations run through this. BTW: there is no need to close the graph as this is part of the shutdown method. Also I've removed the (hopefully?) unrelated checkGraph call and some other method calls. Furthermore have added you as contributor :)

See https://github.com/graphhopper/graphhopper/commit/dfd62e2c8fa73f95eceed05b918cb607fdaac813

Issue #713, opened 2016-05-04
Elevation should be guessed from start and end if going through a tunnel


karussell:
[This Arlberg tunnel](https://graphhopper.com/maps/?point=47.143613%2C10.312042&point=47.129849%2C10.083421&vehicle=car&elevation=true) is the best example, which is I think a lot smoother than we show ;)

The same should be done for bridges like [here](https://graphhopper.com/maps/?point=50.421261%2C12.060757&point=50.43299%2C12.079597&locale=en&vehicle=car&elevation=true&layer=TF%20Outdoors)

highsource:
Hi,

I'd like to give it a try (as a "getting started in Graphhopper" issue)
I understand the issue pretty well but I need a couple of pointers to get started.

Here's my understanding of the task.

Elevation providers give elevation on the surface - which is very different for nodes of the bridge or tunnel. So instead of taking the provided elevation we should somehow calculate elevation of tunnel/bridge nodes.

To do this, for each tunnel/bridge (for verbosity I'll only refer to bridge from now on) we'll need to find "entry" nodes and "inner" nodes of the tunnel. Then we'll need to calculate the elevation of "inner" nodes based on the "outer" nodes, possibly taking into account the surface elevation as maximum (for tunnels) or minimum (for bridges). The tunnel may "dive in" in places where surface elevation is too low.

The connected component of the tunnel will be normally quite simple (just connecting two entry/exit nodes), but there may be more complex structures with multiple entry/exit nodes. Still, these structures will be relatively small.

Now to do this we'll need to identify individual tunnels and find out inner and outer nodes.
A tunnel is basically a connected component where all edges have type "tunnel". Outer nodes are those which have "tunnel" as well as non-"tunnel" edges. Inner nodes only have "tunnel" edges.

The easiest would to traverse the graph in some postprocessing step. If we encounter a "tunnel" edge then we build the connected component from this edge (starting from the nodes of the edge), detect inner/outer nodes (by looking at edges), the calculate elevations of inner nodes and then mark edges as "tunnel-processed".

However this postprocessing will require (and modify) the already built graph. I'm not sure it this is OK, maybe this is needed to be done while reading?

If this is the case then it is harder. We'll need to maintain some kind of tunnel nodeset structure while reading and use them to recalculate elevations after everything's read.
Something like:
* When we read a tunnel edge, we check if its nodes already have associated tunnel nodesets. 
[...]

PR #772, opened 2016-08-21
Saving K_TUNNEL and K_BRIDGE edge flags #713

highsource:
This PR saves `tunnel:yes` and `bridge:yes` in two additional generic flags (just like `K_ROUNDABOUT`).

The drawback is that since these are generic flags, they are currently duplicated among all the flag encoders. As a result, users will need 64-bit edge flags in most cases (2+ encoders).


karussell:
Thanks, will investigate. Related to #713

PR #776, opened 2016-08-23
More details to cannot find point exception

boldtrn:
This is a first implementation idea to add more details to the Cannot find Point exception.

Let me know what you think.


core/src/main/java/com/graphhopper/util/exceptions/CannotFindPointException.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public int getPointIndex()
+    {
+        return this.pointIndex;
+    }
+
+    @Override
+    public Map<String, String> getIllegalArgumentDetails(){
+        Map<String, String> illegalArgumentDetails = new HashMap<>(1);
+        illegalArgumentDetails.put("not_found_point_index", String.valueOf(this.pointIndex));
```
boldtrn: Maybe we should add this String to the Parameters file?

karussell: For now this is okay

core/src/main/java/com/graphhopper/util/exceptions/GHIllegalArgumentException.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.Map;
+
+/**
+ *
+ * Extends IllegalArgumentException because we explicitly catch that in the GraphHopperServlet - TODO we might use something else?
+ *
+ * @author Robin Boldt
+ */
+public abstract class GHIllegalArgumentException extends IllegalArgumentException
```
karussell: Instead of an abstract class I would prefer to have an interface. E.g.

interface GHException {
 Map<String, String>getDetails()
}

Then we can extend from IllegalArgumentException and implement GHException for the CannotFindPointException but can create other exceptions with such details easily.

WDYT?

boldtrn:
> Instead of an abstract class I would prefer to have an interface

That's a great idea, thanks. I updated my branch.

core/src/main/java/com/graphhopper/util/exceptions/GHException.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ *
+ * Extends IllegalArgumentException because we explicitly catch that in the GraphHopperServlet - TODO we might use something else?
+ *
+ * @author Robin Boldt
+ */
+public interface GHException
+{
+    public abstract Map<String, String> getIllegalArgumentDetails();
```
karussell: Would you mind to rename to getDetails :) ?

core/src/main/java/com/graphhopper/util/exceptions/GHException.java
```diff
@@ -0,0 +1,31 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.util.exceptions;
+
+import java.util.Map;
+
+/**
+ *
+ * Extends IllegalArgumentException because we explicitly catch that in the GraphHopperServlet - TODO we might use something else?
```
karussell: remove the TODO?

boldtrn:
Sorry for the inconvenience. Fixed it

karussell:
Thanks and no need to sorry :) ... also I forgot one more thing: would you add a test grabbing the int value? E.g. [here](https://github.com/graphhopper/graphhopper/blob/master/web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java#L147) or whereever you think it is appropriated?

boldtrn:
> add a test grabbing the int value

Sure, forgot about that. I added the test. 

BTW: Have you tested the "Squash Merge" Button of the GitHub UI? It is extremely useful and additionally I don't need to do it on the command line ;)

karussell:
> Klickibunti

Of course I'm a fan of exactly this too ;) ... looks like even wikipedia knows about it https://de.wikipedia.org/wiki/Klickibunti

Regarding formatting -> we'll fix this soonish :) !

karussell:
Thanks - merged!

boldtrn:
> Thanks - merged via UI ;) !

Nice!

> Regarding formatting -> we'll fix this soonish :) !

Yes, I'd appreciate that :+1: 

> looks like even wikipedia knows about it

I found that as well :D

core/src/main/java/com/graphhopper/util/exceptions/CannotFindPointException.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public int getPointIndex()
+    {
+        return this.pointIndex;
+    }
+
+    @Override
+    public Map<String, String> getDetails()
+    {
+        Map<String, String> deatils = new HashMap<>(1);
```
devemux86: "deatils" a typo here?

karussell: Probably also a nice usecase to use Collections.singletonMap?

karussell: Will fix

boldtrn: > "deatils"

Sorry about that :( That's a typo, yes

PR #777, opened 2016-08-23
Add a PointOutOfBoundsException and load BBox from configuration

boldtrn:
When you use GH for a non-world-wide usage, the BBox can become pretty inaccurate. I believe this is due to ferries that are included in some OSM extracts (like from Geofabrik).

Therefore the supported area looks really big. If a user want to create a route in that area, a lot of points will be not found. Therefore this PR contains the ability to specify an excact BBox. If a user requests a point outside of that BBox he/she will receive a PointOutOfBoundsException.

Tests are failing right now, because the tests use Points that are actually outside of the BBox ;) I also updated the BBox Contains function.

Let me know what you think and if I should proceed with this PR. This is currently live on [Kurviger](https://kurviger.de/), but I would extend it for GH, if that is desired.


karussell:
Not sure I understand the use case. Why is the bbox more precise? Like it looks now it is just a duplicate from the storage?

boldtrn:
> Why is the bbox more precise?

Ahh sorry, I think I missed that in the explanation. You can define the BBox in the config file.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -692,6 +696,15 @@ public GraphHopper init( CmdArgs args )
         maxVisitedNodes = args.getInt(Routing.INIT_MAX_VISITED_NODES, Integer.MAX_VALUE);
         maxRoundTripRetries = args.getInt(RoundTrip.INIT_MAX_RETRIES, maxRoundTripRetries);
 
+        // Bounds BBox
+        double left = args.getDouble(Parameters.Config.BB_LEFT, -180);
+        double right = args.getDouble(Parameters.Config.BB_RIGHT, 180);
+        double bot = args.getDouble(Parameters.Config.BB_BOT, -90);
+        double top = args.getDouble(Parameters.Config.BB_TOP, 90);
+
+        bounds = new BBox(left, right, bot, top);
```
boldtrn: The bounds are initialized here

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -95,6 +97,8 @@
     private ElevationProvider eleProvider = ElevationProvider.NOOP;
     private FlagEncoderFactory flagEncoderFactory = FlagEncoderFactory.DEFAULT;
 
+    private BBox bounds = new BBox(-180, 180, -90, 90);
```
karussell: here you could use BBox.createInverse

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -717,9 +730,26 @@ public GraphHopper importOrLoad()
         {
             printInfo();
         }
+        updateBounds();
         return this;
     }
 
+    private void updateBounds(){
```
karussell: instead of this extra method you can use bbox.update

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1015,6 +1045,10 @@ public Weighting createTurnWeighting( Graph graph, FlagEncoder encoder, Weightin
         return weighting;
     }
 
+    public BBox getBounds(){
```
karussell: I would for now not expose this BBox to the outer space.

karussell:
Hmmh, I do not like the fact of having two BBoxes. Do you have ideas about that? Maybe overwrite the graph bbox? (not 100% sure if this has problematic outcomes ;))

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: This represents the BBox for the world-wide usecase. The inverse is not existing, right? There are test cases that do not call the init function, this is basically only for the test cases.

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: Ah nice, I havent' seen that function :)

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: I use that in the InfoServlet to return the correct BBox to the user. Right now you can get the storage, that has a getBounds method as well.

boldtrn:
> Maybe overwrite the graph bbox?
I thought about that, but figured, that it might not be a good idea. Especially since it could have some unexpected negative outcome.

I think these are two different BBox. The one is the BBox for user inputs, the other is for internal calculation. Fox example I am fine, if GraphHopper calculates a route that contains streets that are outside of the BBox, since the OSM Export was done with the complete-ways option. But I would not want a user to select a waypoint outside of the BBox.

karussell:
> Especially since it could have some unexpected negative outcome.

Which one :) ?

> But I would not want a user to select a waypoint outside of the BBox.

Hmmh, I still do not really understand why we would need this.

Couldn't you check this via the returned route if the snapped coordinates are too far away of the specified coordinates? What information does this add for the user that we cannot add right now?

boldtrn:
> Which one :) ?

I don't know :), I am just cautious when changing something. 

> What information does this add for the user that we cannot add right now?

That he placed a waypoint outside of the supported area. Right now, if you import an area like Kurviger, the BBox looks almost like world-wide coverage. But that is not be true, It just contains some ways that are thousands of km long (ferries, I think). The actual issue is the inaccurate BBox, that can be fixed with this PR. The OutOfBoundsExceptions is essentially only a goodie.

karussell:
> I don't know :), I am just cautious when changing something. 

Sure, still this should be covered in our tests and if not we have to add tests, not fear ;)

> The actual issue is the inaccurate BBox, 

I do not think that the bbox is inaccurate, as it includes no wrong information. Probably better would be to have a polygon with more precise data, but ferries could also destroy those. The bbox/polygon can be just a rough hint for the area. So a more reliable method to make your requirements *"a waypoint outside of the supported area"* working is to calculate the distance from specified vs. snapped coordinate and if >1km let the use know that there is no road nearby.

boldtrn:
> I do not think that the bbox is inaccurate

Not in that sense, but it creates the assumption that everything in the gray box is covered, which is wrong. If a user selects a waypoint in the gray box, but outside of the actual covered area, and we return that there is no road nearby, this is irritating, especially if there is actually a road at this point, but it's not part of the included area. That's why I propose this feature. This is meant only for the user, I don't want to change anything else

karussell:
> This is meant only for the user, I don't want to change anything else

Sure, but we should not introduce a feature which introduces duplicate information and where the required use case can be solved differently.

boldtrn:
> required use case can be solved differently

How would you return an exact representation of the covered area? As a polygon that is generated after the import (BTW I love that solution)?

karussell:
I mean, your problem seems to me not showing the boundary but instead "giving the user a proper hint what to do on 'points not found' error". Showing a more precise boundary is a different and more complex task and I would implement something like this only if really required, using a defined one from the configuration would be fine, but should probably also reflect in the imported data (outside should be stripped out) or at least avoid the duplication.

devemux86:
Maybe I'm missing something here, but since graph already provides the bbox, why we need in the implementation the additional `bounds` bbox to be updated by the graph's one?

karussell:
I think this is the case if no one is provided by the config

boldtrn:
> bbox to be updated by the graph's one?

> I think this is the case if no one is provided by the config

Exactly, that was the idea.

> reflect in the imported data (outside should be stripped out)

Maybe you are right and I approached the problem incorrect. Right now I generate an OSM extract using the `--complete-ways` options, which leads to these issues. Incomplete ways are not an issue for GH (I think they are cut at the BBox edge when extracting)?

The Behavior of GH is correct, and generates a BBox that represents the outer boundaries for all ways.

If that works, I think my extract process is the issue and not GH. Therefore, I propose to only include the PointsOutOfBounds Exception and the check if the point is inside of the bounds in this PR.

WDYT?

karussell:
> Right now I generate an OSM extract using the --complete-ways options, which leads to these issues

Any feedback on this would be welcome!

> The Behavior of GH is correct, and generates a BBox that represents the outer boundaries for all ways.

BTW: not 100% correct, e.g. pillar nodes could still be outside of the bbox. 

> I propose to only include the PointsOutOfBounds Exception and the check if the point is inside of the bounds in this PR.

It could be indeed valuable to have this exception instead of the generic "could not find point x"

boldtrn:
> It could be indeed valuable to have this exception instead of the generic "could not find point x"

Ok I updated my branch to reflect the discussed behavior. Tests are still failing, since we add a Point that is actually out of bounds in the tests:
`GHResponse grsp = instance.route(new GHRequest(11.1, 50, 11.2, 52.01).setVehicle("foot"));`

The bounds are at 52 in this case. If I change the point to 52 the resulting route is different. Not sure which solution you propose?

karussell:
Do you have a link to the specific test?

boldtrn:
> We can change the test I think. Do you have a link to the specific test?

Sure: https://github.com/graphhopper/graphhopper/blob/master/reader-osm/src/test/java/com/graphhopper/reader/osm/GraphHopperOSMTest.java#L576

karussell:
Try changing the query location (and inside the bbox) where the route remains identical. See test-osm3.xml for the network

core/src/main/java/com/graphhopper/util/exceptions/PointOutOfBoundsException.java
```diff
@@ -0,0 +1,25 @@
...
+    public PointOutOfBoundsException( String var1, int pointIndex )
+    {
+        super(var1, pointIndex);
+    }
+
+    @Override
+    public Map<String, String> getDetails()
+    {
+        Map<String, String> deatils = new HashMap<>(1);
+        deatils.put("out_of_bounds_point_index", String.valueOf(this.pointIndex));
```
karussell: Do you think it would make sense to just use "point_index" for all such exceptions? This way the client needs no special code and we too :) Furthermore it should be clear enough

reader-osm/src/test/java/com/graphhopper/reader/osm/GraphHopperOSMTest.java
```diff
@@ -573,7 +573,7 @@ public void testFootOnly()
         assertEquals(2, instance.getGraphHopperStorage().getAllEdges().getMaxId());
 
         // A to E only for foot
-        GHResponse grsp = instance.route(new GHRequest(11.1, 50, 11.2, 52.01).setVehicle("foot"));
+        GHResponse grsp = instance.route(new GHRequest(11.1, 50, 11.199999, 52).setVehicle("foot"));
```
boldtrn: Actually I don't really understand that fix. I looked inside the corresponding xml file, but could not find the points?

boldtrn:
Ok, I adapted the PR. I hope it's ok that way?

karussell:
Thanks, looks good. Will just check regarding the test ...

karussell:
Merged

boldtrn:
> Will just check regarding the test ...

Maybe you can explain the test a bit? I checked the .xml file but it basically contains only 5 nodes and some ways that connects these nodes. Where does the intermediate points come from? Are they created by GH?

With intermediate points I mean for example the returned points in the response of the footOnly test. The results contains points that are not specified in the xml.

karussell:
> The results contains points that are not specified in the xml. 

These are virtual nodes, and only a point calculated via "snapped to edge". Previously to this changes we snapped not to the edge but snapped to the point (from eastern) and got the exact coordinates

boldtrn:
> These are virtual nodes, and only a point calculated via "snapped to edge".

Ah ok, I thought so. My understanding was that we only create virtual nodes if these are nodes in OSM. 

Thanks for the explanation :)

karussell:
See [here](https://github.com/graphhopper/graphhopper/blob/master/docs/core/low-level-api.md) and [here](https://github.com/graphhopper/graphhopper/blob/master/docs/core/technical.md) for more details.

boldtrn:
> See here and here for more details.

Yes, I read both files some time ago, but still the behavior felt strange to me. 

When having a look at the above test. When I am querying the following point:  `11.2, 52` the last point of the route is: `11.199999910609154,51.99999992547405`, but the route contains 4 points, whereas when querying: `11.199999, 52` the last point is `11.199999341702974,51.99999957559544` and the route contains 3 points. I was just wondering why the last point is not  `11.2, 52`. Especially since we did not add any other points in the OSM file. Also why is a fourth point added here? That created some headache for me and I fixed the above test using trial and error.

karussell:
> I was just wondering why the last point is not 11.2, 52

Either the tower node is not stored this way in GraphHopper due to rounding errors, or the point is snapped to the edge and creates virtual node before the actual point (you can find out via debugging)

> Also why is a fourth point added here? 

Because the snapped location was on an edge somewhere after 11.2,52

boldtrn:
Thanks for the detailed answer. Helped me a lot!

boldtrn:
> Any feedback on this would be welcome!

Ok, so right now I can confirm that exports without complete-ways parameter work in GH and that the BBox is correct. If any issues occur, I will let you know :)

Issue #396, opened 2015-04-21
Miles instead of Kilometers?


DenisCarriere:
Is it possible to include the desired measurement unit in the URL parameter?

I am trying to have the output distances as miles instead of kilometers.

Issue #95 mentioned it, but it doesn't seem to work anymore:

> But you can also force the locale via an URL parameter e.g. to show miles vs. km:
> http://graphhopper.com/maps/?point=Kiel&point=M%C3%BCnchen&locale=en_US

karussell:
We decided to do this on the client side but not yet done

devemux86:
I agree with that approach.
I think the internal calculations should be done in the SI system and the clients could convert the values as they like.

chimbs86:
So I was trying my hand at this. I see that the com.graphhopper.routing.Path
has a getDistance() method which returns a distance. And here I can add a getDistanceInImperialUnits() as a wrapper around the getDistance() which will use a metric to impeerial conversion and then modify the web interface to display this. Is that a valid approach? Or would you suggest something else instead

chimbs86:
Alternatively, I could only modify the GHResponse to display the imperial distances and thus not change any underlying functionality. Now that I think about it, that seems like a better alternative. 
res.setDistance(distance).setTime(time); this method called in the graphhopper.java will do just that.

karussell:
Hi, thanks. Just converting this in the web UI would be sufficient, so no server or API change is necessary. For that one would need somehow an options button where the user can change these settings. And by default just using km is fine :)

ammagamma:
#778

karussell:
Fixed - thanks @ammagamma !

fbonzon:
The scale could be updated too, to show only the relevant unit (km or mi) depending on this new units switch, instead of showing both. Sent #782.

PR #778, opened 2016-08-24
Makes it possible to use imperial units (miles&feet) in GH web (Issue #396)

ammagamma:
This PR makes it possible to switch to imperial units in GH web.

- Adds a new url parameter useMiles(=true/false) and two buttons in the instructions
  window to choose the unit system.
- Stores the chosen unit system as a flag inside GHRequest (but not in
  GHRequest.api_params as it is not used by the GH api, similar to the do_zoom
  parameter).
- Adds some tests for the unit conversions.
- Updates leaflet.elevation version to (so far unreleased version) 0.0.4 which
  features the option to show the elevation data in feet instead of meters.
- Slightly increases the left margin of the elevation plugin window to make the
  y-axis label visible. For some reason the label is placed further to the left
  than in earlier versions of leaflet.elevation.
- Copies the white-theme style class used earlier for the elevation window to
  the new leaflet.elevation css file.

Not sure if this is the cleanest solution to do it. Also I was not 100% happy to use the unreleased 0.0.4 version of leaflet.elevation, but previous versions only support meters, and changing it by hand seemed worse than using the updated version directly from the master branch.


web/src/main/webapp/js/graphhopper/GHRequest.js
```diff
@@ -70,7 +71,7 @@ GHRequest.prototype.init = function (params) {
         else if (val === "true")
             val = true;
 
-        if (key === "point" || key === "mathRandom" || key === "do_zoom" || key === "layer")
+        if (key === "point" || key === "mathRandom" || key === "do_zoom" || key === "layer" || key === "useMiles")
```
karussell: we should use underscores in all parameters ... not sure why and for what mathRandom was ;)

web/src/main/webapp/js/graphhopper/GHRequest.js
```diff
@@ -159,7 +163,7 @@ GHRequest.prototype.createGPXURL = function (withRoute, withTrack, withWayPoints
 };
 
 GHRequest.prototype.createHistoryURL = function () {
-    return this.createPath("?" + this.createPointParams(true));
+    return this.createPath("?" + this.createPointParams(true)) + "&useMiles=" + !!this.useMiles;
```
karussell: Shouldn't this be within this.createPath?

web/src/main/webapp/js/main-template.js
```diff
@@ -554,18 +554,40 @@ function routeLatLng(request, doQuery) {
             mapLayer.addDataToRoutingLayer(geojsonFeature);
             var oneTab = $("<div class='route_result_tab'>");
             routeResultsDiv.append(oneTab);
-            tabHeader.click(createClickHandler(geoJsons, pathIndex, tabHeader, oneTab, request.hasElevation()));
+            tabHeader.click({useMiles: request.useMiles}, createClickHandler(geoJsons, pathIndex, tabHeader, oneTab, request.hasElevation()));
```
karussell: why do we need a parameter here and why not pass to createClickHandler?

karussell:
Thanks! Will try and play around soonish with it, potentially introduce a 'gears' icon to hide some config buttons like this or introduce new round tour buttons etc

web/src/main/webapp/js/graphhopper/GHRequest.js
ammagamma: Ok, I was not sure if I should follow the convention of  "do_zoom" or "mathRandom". I will change the url parameter from "useMiles" to "use_miles", then. Shall I change "mathRandom" to "math_random" for consistency ? What about the java script variables ? In the GHRequest constructor function for example there is a variable `this.do_zoom` and another one `this.dataType` so both underscore and camel case are used next to each other.

ammagamma:
> potentially introduce a 'gears' icon to hide some config buttons like this or introduce new round tour buttons etc

Of course the UI / arrangement of the buttons etc. can be improved...

> I tried it before and got a black elevation window.

I had to copy/paste the white-theme style class to the leaflet.elevation css file, and I suppose this was also added to the original css file previously. Actually they recommend to create a separate .less file for custom themes: https://github.com/MrMufflon/Leaflet.Elevation#customizing-styles, but not sure if this is really necessary.

web/src/main/webapp/js/main-template.js
ammagamma: I now think we actually do not need it and can pass use_miles directly to createClickHandler as you say. I think I tried this before and the elevation inset did not get updated when the unit buttons were clicked, but I must have made something wrong then. I got it working now and of course this is the better solution.

web/src/main/webapp/js/graphhopper/GHRequest.js
ammagamma: createPath is also used by, e.g., createGPXURL etc. Because use_miles is nothing the Java API should worry about I thought it should not be put into createPath. Of course, having use_miles inside GHRequest is a bit of a workaround, because I assume GHRequest was meant to be the request object that communicates with graph hopper, use_miles (and do_zoom (?)) are more about the client side representation of the results.

karussell: Ah, ok.

> while use_miles (and do_zoom (?)) are more about the client side representation of the results.

Yes, we are storing state in the URL. Although a bit ugly, kind of common.

web/src/main/webapp/js/graphhopper/GHRequest.js
karussell: Well, inside JavaScript and Java we should aim for camelCase variable naming as it is standard there (not 100% sure for JS). But outside of code (configuration, json, URL) we should use under score to make it consistent accross languages and data formats. That was not clear from the beginning so there might be still inconsistency in JS, hopefully not anymore in Java core. See #682

karussell:
> I had to copy/paste the white-theme style class to the leaflet.elevation css file, and I suppose this was also added to the original css file previously.

Nice, thanks for explaining this :) !

karussell:
Thanks - merged!

karussell:
One minor improvement could be done when we avoid a request to the server for the switch, but as this should be rare I'm fine with it

karussell:
This is now live.

PR #780, opened 2016-08-26
ALT algorithm for a more dynamic speed up

karussell:
I've developed over the last months a new speed up algorithm (actually the initial prototype was done in a day ;))

Enable with:
- `prepare.lm.weightings=fastest` in the config.properties
- append `&ch.disable=true` if CH is enabled

The speedup ranges from roughly 5 to 20 times faster, with an average of about 12 for e.g. Bavaria. I.e. it is not as fast as CH which is for the same area 8-10 times faster. The per-request memory requirements are in a similar range (i.e. ~12 times smaller => ~12 times less visited nodes). The static memory requirements are a bit smaller compared to CH but for world wide and performance reasons this could be more depending on the compression and number of the landmarks.

Some other good properties are:
- The LM preparation can be parallelized and should be possible within a few minutes even for world wide coverage
- A recalculation of the preparation is not strictly necessary if e.g. traffic jams ("slow downs")
- Should work for edged_based and also with turn costs and restrictions (although not yet tested)
- Should work for all other non-CH properties like `pass_through` etc or exclude ferries or motorways
- Could be used for time-dependent edge properties (as we could enable/disable them per request)
- A "fastest car" preparation could be re-used for motorcycles (but not the otherway around, property of underestimation most hold)

There are only a few remaining TODOs:

- [x] for larger areas like Europe we need: calculate factor to scale weight better into one short (e.g. currently truncated at 65K making it ugly for larger areas)
- [x] for world wide we need: make subnetworks working via landmark mapping and just select e.g. the largest 32 subnetworks. This is also easy and I'm glad I found this memory efficient solution. Also every node will get a subnetwork id and we can easily decide before a request if the route will be found.
- [x] add integration tests
- [x] fallback to beeline approximation if weight is too big. This could boost performance for large areas, test this.
- [x] force picking the landmarks location being always accessible in both directions => not necessary if we find subnetworks via Tarjan algo and still keeps e.g. norway and UK connected
- [x] Print warning if too many (>10%) entries maxed out.
- [x] use ConnectionNotFoundException instead of runtime exception
- [x] test with edge based & turn costs
- [x] make alternatives working with landmarks
- [x] better heuristic for landmark picking, currently some landmarks are too close to be efficiently used. This will boost performance and hopefully reduce the boost factor variation
- [x] clean up spatial index integration but merge specific PR before web/src/main/java/com/graphhopper/http/DefaultModule.java

Hint: use graph.do_sort=true for 5-10% better routing and preparation performance

Later
- faster and parallel landmark finding
- avoid searching for every profile via reusing landmarks of previous weighting?
- use previously created landmarks from e.g. json file for world wide use case
- test and make "cross weighting usage" working (e.g. motorcycle uses precalculated car weights)
- although CH and LM are combineable LM actually slows down CH. Investigate this.
- use ALT for faster CH preparation


devemux86:
@karussell impressive stats and properties!

karussell:
This is now rebased to latest master. It finally works for EU-wide queries and should work world wide, so that we can include this in 0.9 after some TODOs are solved (updated them above)

The main things I've fiddling with were related with subnetwork issues, which are now covered in the tests. 

Now we detect if two nodes are in the same subnetwork and only then proceed with this landmark-approach otherwise use the beeline algorithm. And so we have to store the subnetwork ID per node, which made the SubnetworkStorage necessary. Furthermore we need a special landmark-mapping (2) so that we know which node IDs the landmarks have in which subnetwork, so that we can still use only **one** large array, where we store all the weights of the landmarks (1). So the data structures are as follows:

SubnetworkStorage.da: just a large byte array storing the subnetwork ID for every node

LandmarkStorage.landmarkWeightDA: (1) the first bytes of this DataAccess store the landmark weights:
```
         from weight lm1 | to weight lm1 | from lm2    | to lm2 | ...
--------------------------------------------------------------------------------------
node1 |  short value     | short value   | short value |
--------------------------------------------------------------------------------------
node2 | ...
```
and (2) then the landmark-subnetwork mapping follows:

```
               landmark1   | lm2          | ...
--------------------------------------------------------------------------------------
subnetwork1 |  int node id | int node id  | ...
--------------------------------------------------------------------------------------
subnetwork1 | ...
```

karussell:
It is interesting how the visited nodes area and count varies with only minor changes to the start&end location. There is probably potential to improve this.

alt-visited-nodes

boldtrn:
Excellent progress @karussell. 

> It is interesting how the visited nodes area (yellow) varies with only minor changes to the start&end location (route is blue). There is probably potential to improve this or some bug - maybe related to some overflow or too imprecise storage of the weights (using only short)?

Do you think this is a bug? My first impression (really just an impression, I have nothing to support this) is that you placed the points in the upper picture at a location that might be valid for taking two possible "good highways". When looking at the right point in the upper picture you can see how the other reasonable good highway was followed for some time (the yellow line going north/east). It is hard to explain :). But maybe I am wrong?

karussell:
I do not know if this is a bug, have not much investigated yet. Before I thought the starting and ending area looked strange as there were sometimes many yellow nodes, but now the more I think about it it is probably how the algo works ...

karussell:
Finally it 'works' world wide. The remaining problem is that the landmark count for EU+Asia+Africa is too small (when using 16) as those areas are connected via 'landline' and ferries, see [this](http://bl.ocks.org/d/b35e484d56387c02e02478d60527171c). Not sure about the perfect solution. We could disconnect those via introducing border restrictions, increase the landmarks (will try now), artificially introduce subnetworks for the landmarks or something.

5 subnetworks were created: EU+Asia+Africa; Australia; Northern America; Southern America; Malaysia. Strange, where is Japan? One big enough subnetwork near Malaysia (>500K nodes) still has problems to create the landmarks: "next start node not found in big enough network of size 806392, first element is 103853491, 5.515810225512316,95.84255935644863"

BTW: creating the landmark weights takes ~70 minutes for car. Including subnetwork calculation, landmark assignment and finally the weight calculation for every subnetwork.

karussell:
As we need more separated areas e.g. to split EU/Asia/Africa into its parts we could also need #926

karussell:
Finally <1s response times for most continental >4000km (e.g. EU or US) routing requests :) !

karussell:
This should be in a good shape now. Except the spatial index integration (we'll merge the related PR before this here) and the splitCountries method which should be optional and probably done independently of the landmarks creation. I'll do this and also try to improve performance a bit. Regarding the other stuff I've added @boldtrn to review this, hope this is okay :) ? You can also have a look if round tour creation could use the landmark speed up easily (similar like [I did with the alternative routes](https://github.com/graphhopper/graphhopper/pull/780/commits/2a30f3dde2f6f137fa6df9aff73284cfe57a669e)?)

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -882,11 +920,16 @@ public void postProcessing() {
...
             createCHPreparations();
+        if (!isCHPrepared())
+            prepareCH();
+
+        if (lmFactoryDecorator.isEnabled())
+            lmFactoryDecorator.createPreparations(ghStorage, traversalMode);
 
-        if (!isPrepared())
-            prepare();
+        prepareLM();
```
boldtrn: Why don't we need to check if LMs are prepared?

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
```diff
@@ -93,11 +94,11 @@ public void init(CmdArgs args) {
         if (enableThis)
             setDisablingAllowed(args.getBool(CH.INIT_DISABLING_ALLOWED, isDisablingAllowed()));
 
-        setPreparationPeriodicUpdates(args.getInt("prepare.updates.periodic", getPreparationPeriodicUpdates()));
-        setPreparationLazyUpdates(args.getInt("prepare.updates.lazy", getPreparationLazyUpdates()));
-        setPreparationNeighborUpdates(args.getInt("prepare.updates.neighbor", getPreparationNeighborUpdates()));
-        setPreparationContractedNodes(args.getInt("prepare.contracted_nodes", getPreparationContractedNodes()));
-        setPreparationLogMessages(args.getDouble("prepare.log_messages", getPreparationLogMessages()));
+        setPreparationPeriodicUpdates(args.getInt(CH.PREPARE + "updates.periodic", getPreparationPeriodicUpdates()));
```
boldtrn: Should we put them in the Parameters file?

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
```diff
@@ -210,7 +212,7 @@ public CHAlgoFactoryDecorator setWeightingsAsStrings(String... weightingNames) {
 
     public List<String> getWeightingsAsStrings() {
         if (this.weightingsAsStrings.isEmpty())
-            throw new IllegalStateException("Potential bug: chWeightingList is empty");
+            throw new IllegalStateException("Potential bug: weightingsAsStrings is empty");
```
boldtrn: Why is this a potential bug? Maybe we can add more information here?

core/src/main/java/com/graphhopper/routing/lm/LMAlgoFactoryDecorator.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.List;
+
+/**
+ * This class implements the A*, landmark and triangulation (ALT) decorator.
+ *
+ * @author Peter Karich
+ */
+public class LMAlgoFactoryDecorator implements RoutingAlgorithmFactoryDecorator {
+    private int landmarkCount = 8;
+    private int activeLandmarkCount = landmarkCount / 2;
```
boldtrn: So should landmarks be even or factors of 2?

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+    private int minimumNodes = 500_000;
+    private SubnetworkStorage subnetworkStorage;
+    private Map<String, Polygon> borderMap = new HashMap<>();
+
+    public LandmarkStorage(GraphHopperStorage graph, Directory dir, int landmarks, final Weighting weighting, TraversalMode traversalMode) {
+        this.graph = graph;
+        this.minimumNodes = Math.min(graph.getNodes() / 2, 500_000);
+        this.encoder = weighting.getFlagEncoder();
+        this.weighting = weighting;
+        this.lmSelectionWeighting = new ShortestWeighting(encoder) {
```
boldtrn: Would it make sense to optionally provide the weighting for the preparation. Makes probably sense for cases where shorter distance does not need to be lower weight?

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
karussell: No, too unimportant

karussell:
> This looks very good to me. Overall query time for Badenwürttemberg was reduced 4-5 times.

With how many landmarks? It should be more like 10-20 with 16 or 32.

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
karussell: It just should not happen ... but this wasn't introduced in this PR :)

core/src/main/java/com/graphhopper/routing/lm/LMAlgoFactoryDecorator.java
karussell: No, this does not matter. But I'll make the defaults a bit different as tests show that 8 active landmarks are nearly always the best configuration despite the total number of landmarks

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
karussell: What do you mean here? The tricky part is that we are using the normal weighting for the preparation but the landmark selection heuristic works best with shortest. See phase 1 vs. 2 in createLandmarks

karussell:
And how did you determine the performance? Did you use the "measurement" class? This class also produces stats for a few different active landmarks

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: This is indeed a bit inconsistent: the if is now inside the prepareLM method.

boldtrn:
> With how many landmarks? It should be more like 10-20 with 16 or 32.

Sorry I should have given more details. I just enabled the lm preparation in the config. Didn't do any tuning.

> And how did you determine the performance? Did you use the "measurement" class? This class also produces stats for a few different active landmarks

No I just measured the request time, so it also contains the time to lookup the coordinates and so. Not a really sophisticated test, but the request time was reduced by factor 4 or 5. Therefore, the actual routing was faster.

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: Yes I saw that later on :). Not sure if we should change this though. Maybe rename the perpare method to prepareIfNotPrepared for ch an lm. This would make the code a bit more readable IMHO. WDYT?

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
boldtrn: I know :). I just stumbled over it and was wondering why we are throwing such a generic exception here.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+    boolean isInitialized() {
+        return initialized;
+    }
+
+    /**
+     * This method calculates the landmarks and initial weightings to & from them.
+     */
+    public void createLandmarks() {
+        if (isInitialized())
+            throw new IllegalStateException("Initialize the landmark storage once!");
```
boldtrn: "Initialize the landmark storage **only** once!"? Not sure though.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+        }
+
+        // introduce a factor to store weight without loosing too much precision 
+        // AND making it compatible with weighting.calcWeight
+
+        // roughly approximate the maximum distance of the current area but limit to world wide case
+        // too small is dangerous regarding performance, e.g. for Germany at least 1500km is very important otherwise speed is at least twice as slow e.g. for just 1000km
+        BBox bounds = graph.getBounds();
+        double maxTmp = 30_000_000;
+        double distanceInMeter = !bounds.isValid() ? maxTmp : (Helper.DIST_EARTH.calcDist(bounds.maxLat, bounds.maxLon, bounds.minLat, bounds.minLon) > 50_000 ? maxTmp : 1_000_000);
```
boldtrn: Could we split this line in several lines putting the results in a variable with a naming that explains what happens? It is hard to get what is happening here.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+        // 'to' and 'from' fit into 32 bit => 16 bit for each of them => 65536
+        factor = weightMax / (1 << 16);
+
+        if (Double.isInfinite(factor) || Double.isNaN(factor))
+            throw new IllegalStateException("Illegal factor calculated from distance " + distanceInMeter + " and weight " + weightMax);
+
+        LOGGER.info("init landmarks for subnetworks with node count greater than " + minimumNodes + ", weightMax:" + weightMax
+                + ", factor:" + factor + " from max distance:" + distanceInMeter / 1000f + "km");
+
+        // special subnetwork 0
```
boldtrn: Is this the unclear_subnetwork? Is this documented why it is special?

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+        // UNCLEAR_SUBNETWORK == 0 should only be used if subnetwork is too small
+        Arrays.fill(subnetworks, (byte) UNSET_SUBNETWORK);
+
+        // we cannot reuse the components calculated in PrepareRoutingSubnetworks as the edgeIds changed in between (called graph.optimize)
+        // also calculating subnetworks from scratch makes bigger problems when working with many oneways
+
+        StopWatch sw = new StopWatch().start();
+        // TODO make better testable for smaller area
+        SpatialRuleLookupArray ruleLookup = new SpatialRuleLookupArray(createEU_Africa_RU_BBox(), 0.1, true);
+        int edges = splitCountries(borderMap, ruleLookup);
```
boldtrn: rename method to makeBorderCrossingEdgesInaccessible?

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+        LOGGER.info("Made " + edges + " edges inaccessible. Calculated country cut in " + sw.stop().getSeconds() + "s, " + Helper.getMemInfo());
+        sw = new StopWatch().start();
+        TarjansSCCAlgorithm tarjanAlgo = new TarjansSCCAlgorithm(graph, new DefaultEdgeFilter(encoder, false, true), true);
+        List<IntArrayList> graphComponents = tarjanAlgo.findComponents();
+        LOGGER.info("Calculated tarjan subnetworks in " + sw.stop().getSeconds() + "s, " + Helper.getMemInfo());
+
+        EdgeExplorer tmpExplorer = graph.createEdgeExplorer(new RequireBothDirectionsEdgeFilter(encoder));
+
+        int nodes = 0;
+        MAIN:
```
boldtrn: Why is there a label? BTW seems to be unused as well.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+        for (int nodeId = 0; nodeId < subnetworks.length; nodeId++) {
+            subnetworkStorage.setSubnetwork(nodeId, subnetworks[nodeId]);
+        }
+
+        LOGGER.info("Finished landmark creation. Subnetwork node count sum " + nodes + " vs. nodes " + graph.getNodes());
+        initialized = true;
+    }
+
+    // TODO calculate from GeoJSON, #926
+    protected BBox createEU_Africa_RU_BBox() {
```
boldtrn: Should be supported now.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+            // TODO cache rule for all nodes to avoid calling "2*nodes" of lookup
+            if (ruleAdj != ruleBase) {
+                inaccessible++;
+                allEdgesIterator.setFlags(encoder.setAccess(allEdgesIterator.getFlags(), false, false));
+            }
+        }
+        return inaccessible;
+    }
+
+    private String myprint(int... ints) {
```
boldtrn: This method seems to be unused, should we delete it?

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+        System.out.println(str);
+        return str;
+    }
+
+    /**
+     * This method creates landmarks for the specified subnetwork (integer list)
+     *
+     * @return landmark mapping
+     */
+    private boolean createLandmarks(final int startNode, final byte[] subnetworks, IntArrayList subnetworkIds) {
```
boldtrn: Rename this method as it has the same name as createLandmarks?

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
boldtrn: You are right, sorry I missed that.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
boldtrn: I just found more references to this. Maybe we should always reference to the UNCLEAR_NETWORK variable and add some documentation at the variable to explain its use?

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+        final int subnetworkId = landmarkIDs.size();
+
+        // 1a) pick landmarks via shortest weighting for a better geographical spreading
```
boldtrn: Where did we picked a landmark?

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+        // 'fastest' has big problems with ferries (slow&very long) and allowing arbitrary weighting is too dangerous
+        Weighting initWeighting = lmSelectionWeighting;
+        Explorer explorer = new Explorer(graph, this, initWeighting, traversalMode);
```
boldtrn: Thought at first this would be the EdgeExplorer. Maybe explicitly call this LandmarkExplorer?

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -0,0 +1,745 @@
...
+                lastNode = currFrom.adjNode;
+                return finishedFrom;
+            } else {
+                lastNode = currTo.adjNode;
+                return finishedTo;
+            }
+        }
+
+        public boolean setSubnetworks(final byte[] subnetworks, final int subnetworkId) {
+            if (subnetworkId > 127)
```
boldtrn: Put this in a variable?

karussell: This is not variable ;(

karussell:
Thanks for the review will do what you said and also try to be more descriptive in code and with more javadocs.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
boldtrn: Yes, put it in a constant, sorry ;).

boldtrn:
> Thanks for the review will do what you said and also try to be more descriptive in code and with more javadocs.

Sorry I was really picky with the naming this time. I tried to understand what was happening and just saw that in some cases the names didn't reflected the logic :).

karussell:
I've added the possibility to use hand picked landmarks instead of calculated once per subnetwork, which is not only faster but >2 times faster for country sized areas.

Will no change as requested now. And include the spatial index stuff to split the area in smaller subnetworks (only for world wide scenario) for better speed of landmarks.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -163,9 +165,9 @@ boolean isInitialized() {
     /**
      * This method calculates the landmarks and initial weightings to & from them.
      */
-    public void createLandmarks() {
+    public void createLandmarksForSubnetwork() {
```
karussell: oh, it looks like intelliJ has renamed this at the same time ;)

karussell:
BTW: I found that the spatial index json (map.geo.json) had lots of IDs "-99". Now that we throw exceptions in those duplicative cases I had to edit this field and stumbled over e.g. France where this was obviously incorrect (FRA) and also for Norway (NOR). For the others I've just put an increasing counter before the -99 ...

karussell:
Three last steps to make this production ready (if no other objections :))

 - [ ] integration test for reading the landmarks
 - [ ] unit tests for splitting the areas according to the SpatialRuleLookup
 - [ ] run again world wide and test if flexible mode is able to cross some of the "special area" border which will fail for the landmark routing via 'connection not found'.

boldtrn:
> BTW: I found that the spatial index json (map.geo.json) had lots of IDs "-99". Now that we throw exceptions in those duplicative cases I had to edit this field and stumbled over e.g. France where this was obviously incorrect (FRA) and also for Norway (NOR).

Oh, I missed that. There is actually an issue open for that. Sorry for the troubles. https://github.com/datasets/geo-countries/issues/5

karussell:
> I've added the possibility to use hand picked landmarks instead of calculated once per subnetwork, which is not only faster but >2 times faster for country sized areas.

This is no longer true - I've tuned autopicking landmarks with a simple trick using no 'fastest' or 'shortest' weighting but just use BFS and return 1 as weighting which leads to 5% faster landmarks than handpicked and roughly 22x faster response times for Germany (22x faster on average!).

karussell:
> Oh, I missed that. There is actually an issue open for that. Sorry for the troubles.

No trouble at all - it was fixed in <5min :)

karussell:
Nice, everything works, also world wide with even better response times after the recent weighting change. Will merge after a few cosmetics.

karussell:
Merge now.

I had to made a few minor changes to the spatialrule lookup @boldtrn please have a look if this is ok.

boldtrn:
Awesome! Happy to see this feature merged.

> I had to made a few minor changes to the spatialrule lookup @boldtrn please have a look if this is ok.

Looks good to me. The factory is a good idea. So it's easy to inject custom SpatialRules :+1:

Issue #773, opened 2016-08-22
replace json.org library with own wrapper


karussell:
Currently we use a json libary with a potential ugly license in web module (_The software shall be used for good, not evil._). [Others](http://blog.sonatype.com/2012/03/use-json-well-youd-better-not-be-evil/) don't like it too. Android has a Apache licensed version of it [here](https://android.googlesource.com/platform/libcore/+/310eafdfeb5db23a4d5e405bbe607cffd1f4b24a/json/src/main/java/org/json/JSONObject.java#26) claiming that: "_this class was written without inspecting the non-free org.json sourcecode_"

So we should look into alternatives to replace this.

Also for several tasks we need a json parser in the **core** in the future. Such a parser should be fast, small, actively maintained, usages according to [this](http://mvnrepository.com), simple to use for basic configuration reading stuff, available on maven central and the Apache License is preferred. Also Java 7 is required for now, reading into POJOs would be good but not required.
- https://github.com/fangyidong/json-simple, this is actually the json.org project?!? 25KB, 200 stars, 420 usages, Apache, seem inactive
- https://github.com/ralfstx/minimal-json 30KB, 340 stars, MIT, 36 usages
- https://github.com/square/moshi ~100KB, 2000 stars, Apache, 16 usages
- https://github.com/jdereg/json-io ~100KB, 170 stars, Apache, 14 usages
- https://github.com/google/gson ~230KB, 5700 stars, Apache, 2360 usages
- http://johnzon.apache.org/index.html 240KB?, does not apply, Apache, <10 usages
- https://github.com/FasterXML/jackson-jr ~300KB (equivalent to databind, jackson core is 280KB), 80 stars, Apache, usages 3K for normal jackson, jackson-jr <20
- https://github.com/owlike/genson ~400KB,120 stars, Apache, <20

Probably a lot simpler would be if we pick one for now, create a simple interface and then could switch afterwards.

karussell:
The size is not that important as on Android side one could use ProGuard, on server side it does not matter and for iOS we probably need a different implementation.

devemux86:
Well Gson has been more than good all this time. :+1: 

Also note that Android has a base [JSON API](https://developer.android.com/reference/org/json/package-summary.html) integrated.

karussell:
Looks like they just copied the JSON lib we are currently using despite the license thing.

karussell:
moshi looks even more streamlined (100KB) and is very similar to gson

[...]

PR #781, opened 2016-08-26
Change graph based on overlay data

karussell:
This PR makes it possible to feed external speed or access changes via GeoJSON. For now it is not yet integrated in e.g. GraphHopper core or web. But you can easily add it via:

```java
GraphHopper myHopper = new GraphHopperOSM() {

    public void postProcessing()
    {
        super.postProcessing();
        FeedOverlayData instance = new FeedOverlayData(getGraphHopperStorage(), 
                getEncodingManager(), getLocationIndex(), new GHsonBuilder().create());
        long updates = instance.applyChanges(fileOrFolderAsString);
    }
}
```
pom.xml:
```xml
        <dependency>
            <artifactId>graphhopper-reader-overlay-data</artifactId>
            <version>0.8-SNAPSHOT</version>           
        </dependency>
```

E.g. see FeedOverlayDataTest or an [example json file](https://raw.githubusercontent.com/graphhopper/graphhopper/reader-overlay-data/reader-overlay-data/src/test/resources/com/graphhopper/reader/overlaydata/overlaydata1.json)

Keep in mind that the order of any coordinate array is lon,lat and e.g. the bbox is min_lon,min_lat,max_lon,max_lat like specified in the [Geojson spec](http://geojson.org/geojson-spec.html). Also currently only three cases for a Feature are implemented:
 * `type=Point`, 
 * `type=MultiPoint` and 
 * no `type` and no `properties.coordinates` entry but with a bounding box entry `bbox`

It furthermore introduces a Json reader wrapper as discussed in #773 backed by Gson.

The overlay change functionality will stay in separate module `reader-overlay-data`. It goes even further and as we need map matching functionality in this reader module, we potentially have to move the currently separate [map matching component](https://github.com/graphhopper/map-matching) into this repository here. Keeping it necessary and still depend on it would not work well.

Keep in mind: the feeding mechanism is not thread safe, so you should **not** use this for real time updates without some locking. And it would work only for non-CH, but this method here is before the preparation and works with CH as applied only once.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ local.properties
 *.zip
 .DS_Store
 /core/nbproject/
+/reader-overlay-data/nbproject/
--- a/pom.xml
+++ b/pom.xml
@@ -10,9 +10,7 @@
     <packaging>pom</packaging> 
     <url>https://graphhopper.com</url> 
     <inceptionYear>2012</inceptionYear>
-    <description>
-        Super pom of GraphHopper core, a fast Java road routing engine.
-    </description>
+    <description>Super pom of GraphHopper, the fast and flexible routing engine</description>
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
@@ -79,6 +77,7 @@
         <module>tools-lgpl</module>
         <module>core</module>
         <module>reader-osm</module>
+        <module>reader-overlay-data</module>
         <module>tools</module>
         <module>web</module>
     </modules>
--- a/reader-overlay-data/pom.xml
+++ b/reader-overlay-data/pom.xml
@@ -4,10 +4,10 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>com.graphhopper</groupId>
-    <artifactId>graphhopper-reader-osm</artifactId>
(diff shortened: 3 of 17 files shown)


karussell:
Currently the default is to pick the first vehicle if non is specified in the file, maybe we should apply the change to all vehicles instead.

karussell:
Will merge soon

karussell:
Attached is a real world example ([berlin area](http://download.geofabrik.de/europe/germany/berlin-latest.osm.pbf)) to block access (car) and lower speed (foot)
[overlaydata1.json.zip](https://github.com/graphhopper/graphhopper/files/444996/overlaydata1.json.zip)

for foot => http://localhost:8989/?point=52.51006%2C13.375436&point=52.510655%2C13.373221&vehicle=foot

for car => http://localhost:8989/?point=52.51055%2C13.375779&point=52.510955%2C13.375688&vehicle=car

karussell:
Is now merged to master: https://github.com/graphhopper/graphhopper/commit/9a9f34ad67ec661568d35a1b4ebf43012d21fec6

karussell:
Using serialization classes e.g. 'ConfigurationPojo' in the core will avoid using this 'heavier' dependency into the core. We can deserialize the json in the web module into such a 'serialization object' and forward this.

PR #783, opened 2016-08-26
Update .gitignore

fbonzon:
You probably mean a specific file here, and not
`/some/other/web/src/main/webapp/js/config/options_prod.js`


karussell:
Thanks!

PR #784, opened 2016-08-26
Avoid broken images when GraphHopper is accessed with "index.html"

fbonzon:
Example URL with broken images: https://graphhopper.com/maps/index.html?point=46.649436%2C7.558594&point=46.963385%2C7.69043&locale=en-US&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale

With this PR, images in instructions will show up.


karussell:
Hmmh, why do you think this is important? Will have to test if other scenarios (local, production) will break

karussell:
Okay, this is nice to have if running the UI locally against any routing server.

fbonzon:
I think it is important, because the functionality of GraphHopper Maps should not depend on what URL you access it from.

I reached the URL with index.html from my FTP client. Right-clicked on index.html file icon and selected "Open URL in browser", which was the fastest was to reach the instance I had just uploaded on my server.

The local files case you mentioned is even more useful, yes.

I [tested](http://xn--franois-xxa.ch/graphhopper-dirname/index.html?point=Berlin%2C%20Germany&point=Munich%2C%20Germany&locale=en-US&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=OpenStreetMap) it. It will not break production.

karussell:
Yes, had also tested this. Thanks again :) !

PR #785, opened 2016-08-28
Gradle: advance to Android 7 Nougat

devemux86:
@karussell do we need to advance also Android version in pom.xml [here](https://github.com/graphhopper/graphhopper/blob/master/android/app/pom.xml#L121)?


karussell:
Let us keep the pom.xml for now as it seems just used from me ;)

PR #787, opened 2016-09-08
Added the possibility to add a name to the finish instruction

boldtrn:
Preparing to fix this [Issue](https://github.com/graphhopper/directions-api-java-client/issues/15)

Therefore, we need the possibility to add a name to the FinishInstruction. This method will be used [here](https://github.com/graphhopper/directions-api-java-client/blob/master/client/src/main/java/com/graphhopper/api/GraphHopperWeb.java#L304).

### DIFF
--- a/core/src/main/java/com/graphhopper/util/FinishInstruction.java
+++ b/core/src/main/java/com/graphhopper/util/FinishInstruction.java
@@ -57,7 +57,7 @@ public class FinishInstruction extends Instruction
     @Override
     public String getTurnDescription( Translation tr )
     {
-        if (rawName || !name.isEmpty())
+        if (rawName)
             return getName();
 
         return tr.tr("finish");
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/graphhopper/util/FinishInstruction.java
```diff
@@ -41,7 +57,7 @@ public FinishInstruction( PointAccess pointAccess, int node )
     @Override
     public String getTurnDescription( Translation tr )
     {
-        if (rawName)
+        if (rawName || !name.isEmpty())
```
devemux86: The additional check is not used in other instructions, is it needed?

boldtrn: Good point. No, I don't thinks so.

karussell:
Thanks!

PR #790, opened 2016-09-10
More turn instructions

don-philipe:
With some little changes in the `Path` class, now more turn instructions are available. Now only the `continue` instruction on a road whose name don't change is omitted. See also #94.

Had to adjust several tests. I hope this is sufficient for testing this feature.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/Path.java
+++ b/core/src/main/java/com/graphhopper/routing/Path.java
@@ -541,12 +541,12 @@ public class Path
                     prevName = name;
                     prevAnnotation = annotation;
 
-                } else if ((!name.equals(prevName)) || (!annotation.equals(prevAnnotation)))
+                } else
                 {
-                    prevOrientation = AC.calcOrientation(doublePrevLat, doublePrevLong, prevLat, prevLon);
+                    double tmpPrevOrientation = AC.calcOrientation(doublePrevLat, doublePrevLong, prevLat, prevLon);
                     double orientation = AC.calcOrientation(prevLat, prevLon, latitude, longitude);
-                    orientation = AC.alignOrientation(prevOrientation, orientation);
-                    double delta = orientation - prevOrientation;
+                    orientation = AC.alignOrientation(tmpPrevOrientation, orientation);
+                    double delta = orientation - tmpPrevOrientation;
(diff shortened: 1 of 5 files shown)


karussell:
Thanks!

> With some little changes in the Path class, now more turn instructions are available. 

'More instructions' are not necessarily better, what exactly does it change? And does it change the performances of calculating the instructions?

> Had to adjust several tests. I hope this is sufficient for testing this feature.

As this improves turn instructions I would like to have a test added which shows this.

devemux86:
>'More instructions' are not necessarily better, what exactly does it change?

Indeed, now regardless name change we have a new instruction?

Also "continue" instruction seems producing the same?
And that's the most frequent complaint for GraphHopper instructions, too much unnecessary continues. :smile:

karussell:
@don-philipe would you mind to explain the changes a bit? (plus see my questions)

don-philipe:
Sorry for the delay, was very busy the last weeks.

> Also "continue" instruction seems producing the same way?
And that's the most frequent complaint for GraphHopper instructions, too much unnecessary continues.

I implemented this feature mainly addressing routing for blind persons. They need the information about direction changes even if the OSM-Tags don't change. In #94 someone suggested that this feature should be optional (only for certain routing profiles?).

> 'More instructions' are not necessarily better, what exactly does it change?

The changes in the Path class affect the decision about the instruction generation. In the original code instruction where only generated when name or annotation of an edge changed (plus first instruction and roundabout instructions). With my changes the last "else if" branch is transformed into an "else" branch. Now an instruction is produced with every new edge. At the end of this branch another "if" decides whether to accept this instruction or not. This was neccessary because the direction calculation is done in this branch. Instructions are now accepted if the edge name or annotation changes (as in previous version) or if the direction is not "continue".
All in all we have now also instructions with direction changes not only when the OSM-Tags are changing, but as well when they don't change (as in first example in #94).

> And does it change the performances of calculating the instructions?

I haven't done measures regarding perfomance. But as explained above, there is more calculation do to now (at least one calculation for every edge).

karussell:
> In #94 someone suggested that this feature should be optional (only for certain routing profiles?)

Ah, okay. This would make sense. But as it is now it is not mergable. Maybe you or someone else has tome for this :) ?

PR #791, opened 2016-09-10
Implements a more precise way to add new intermediate points in web UI

ammagamma:
Previously, the order of intermediate points was simply equal to the order in which the points were added. This PR changes this behavior such that a newly added point is inserted between the closest two existing waypoints.
Edit: I just realized there is still a problem with alternative routes ... (fixed in second commit)


karussell:
Thanks! Will try it out in the next days

web/src/main/webapp/js/main-template.js
```diff
@@ -547,7 +584,11 @@ function routeLatLng(request, doQuery) {
             var geojsonFeature = {
                 "type": "Feature",
                 "geometry": path.points,
-                "properties": {"style": style}
+                "properties": {
+                    "style": style,
+                    name: "route",
+                    snappedWaypoints: path.snapped_waypoints
```
karussell: I know it is a bit nitpicking but as we 'emulate' GeoJSON we would need underscore - does this make sense :) ?

I.e.
snapped_waypoints: path.snapped_waypoints

karussell:
Thanks again - works perfectly :) !

Would you mind to fix the minor mentioned thing and (if possible) add a quick test for it [here](https://github.com/graphhopper/graphhopper/tree/master/web/src/test/webapp/spec)?

web/src/main/webapp/js/main-template.js
ammagamma: Ok, sure I fixed this.

ammagamma:
> Would you mind to fix the minor mentioned thing and (if possible) add a quick test for it here?

I would like to add a test, but this does not seem very easy at the moment. We could run a kind of integration test in the web browser (maybe using [phantomJS](http://phantomjs.org/)). Alternatively, we could extract the function that finds the correct index for the waypoint and test it, but in this case we would still have to mock some of the dependencies, like the mapLayer and some more Leaflet stuff (L.latLng etc.). As far as I can tell for both solutions we are missing some infrastructure right now. 

I also thought about if some of the functionality needed here should be available within graphhopper, so we could reduce the amount of logic in the UI layer. For example when calculating a route for given via-points, is it possible to keep track which parts of the resulting route lie between which via-points ? What I mean is this:

Request: A --> B --> C (route from A to C via B)
Response: A --> x_1 --> x_2 --> ... --> B --> x_45 -->x_46 --> ...-->x_99 --> C (exact coordinates of total route)

Is there a way to obtain the via-points that are the start and end of a given route segment, e.g. is it possible to call something like `response.intermediatePoints(x_45)` and retrieve (B, C) ? As far as I could tell the graphhopper response only returns the whole route (x_1, x_2, ...., x_99) and the via-points (A, B, C) separately, but no information which coordinates lie between which via-points.

karussell:
> I would like to add a test, but this does not seem very easy at the moment

Ok, no problem. I just thought somehow assuming 3 points and then add another and asserting the correct index is somehow possible.

> We could run a kind of integration test in the web browser (maybe using phantomJS).

I would prefer to keep one tool for now (Jasmine) but we should investigate alternatives.

> we would still have to mock some of the dependencies, like the mapLayer and some more Leaflet stuff 

That would be okay for now

> I also thought about if some of the functionality needed here should be available 
> within graphhopper, so we could reduce the amount of logic in the UI layer

Interesting thought. We already can calculate distances to 'edges' but this is not published as API.

> As far as I could tell the graphhopper response only returns the whole route (x_1, x_2, ...., x_99) 
> and the via-points (A, B, C) separately, but no information which coordinates lie between which via-points.

There are two possibilities with which one could achieve this

 * the sign of the instruction indicates if this is a via point and then one could merge the point indices of the `interval` array
 * or using `snapped_waypoints`

We could make it easier with a simple array where values correspond to the point indices for every via-point, something like `waypoints_intervals`

web/src/main/webapp/js/main-template.js
karussell: Thanks!

ammagamma:
>>   we would still have to mock some of the dependencies, like the mapLayer and some more Leaflet stuff

>  That would be okay for now

I will try this for now.

karussell:
Thanks a bunch!

web/src/test/webapp/spec/routeManipulationSpec.js
```diff
@@ -0,0 +1,74 @@
+// because leaflet-src.js requires window, document etc. we need to provide these variables...
```
ammagamma: This is pretty ugly. Maybe you have a better idea ? We could keep the tests separate from leaflet entirely by reimplementing/copy&pasting the functionality of `L.latLng`, but I do not think this is much nicer. We could also put this workaround into a separate file.

ammagamma:
I separated the method for finding the correct index and wrote two tests for it. Maybe you can have a look again now. Let me know if you do not like the naming etc. !

ammagamma:
> >    We could run a kind of integration test in the web browser (maybe using phantomJS).

> I would prefer to keep one tool for now (Jasmine) but we should investigate alternatives.

I was thinking about a tool that allows us to simulate clicks on the map and for example check whether the url changes in a correct way, not sure if this can be done only with Jasmine.

ammagamma:
> We could make it easier with a simple array where values correspond to the point indices for every via-point, something like waypoints_intervals

If this was available we would at least not have to find the next way point by iterating over the route.

web/src/test/webapp/spec/routeManipulationSpec.js
karussell: No idea, but looks okay to me. We are kind of mocking away these dependencies which could lead to trouble but for now this should be sufficient. In long term we should try to move more parts into a better structured code ;( (more seperate from UI)

karussell:
> I separated the method for finding the correct index and wrote two tests for it. Maybe you can have a look again now. Let me know if you do not like the naming etc. !

Thanks a bunch, this is what I meant and looks good :) !

> I was thinking about a tool that allows us to simulate clicks on the map and for example check whether the url changes in a correct way

ah, yes. Something like selenium? I see this more like the integration testing which is currently not existant and possibly also important. The unit testing part should be more or less decoupled from the UI and doable with jasmine (I know the architecture is currently bad to have this properly separated)

> If this was available we would at least not have to find the next way point by iterating over the route.

Ok

karussell:
Merged and already live :) ! Thanks again!

oblonski:
Cool. Thanks a lot. It looks great.

boldtrn:
@ammagamma Nice feature, works pretty good. When start and end is the same point, there is an issue, maybe you can checkt?

karussell:
@boldtrn do you have an example as I cannot reproduce this.

boldtrn:
Sure sorry, I should have provided an example from the start.

Try this route: https://graphhopper.com/maps/?point=49.439557%2C9.118652&point=52.596375%2C13.458252&point=52.562995%2C9.250488&point=49.439557%2C9.118652&locale=de-DE&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale

Now add a point western of the route, for example in Siegen, using the right click dialog. It will be the 1st waypoint, wereas it should be the 3rd.

For example: https://graphhopper.com/maps/?point=49.439557%2C9.118652&point=50.847573%2C8.009033&point=52.596375%2C13.458252&point=52.562995%2C9.250488&point=49.439557%2C9.118652&locale=de-DE&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale

ammagamma:
@boldtrn thanks for testing, this is indeed a problem, which occurs whenever a waypoint is crossed multiple times. I will look for a better solution.

ammagamma:
> ah, yes. Something like selenium? I see this more like the integration testing which is currently not existant and possibly also important. 

Yes thats what I meant

ammagamma:
https://github.com/graphhopper/graphhopper/pull/795

PR #796, opened 2016-09-16
Issue #713 part 1: store tunnel and bridge tags.

highsource:
This PR adds a few utility methods to check if the edge is a tunnel or a bridge edge.


karussell:
Thanks! Also for moving this to the new encoder!

For performance reasons it is very wise that you put the int values into a separate field :). Now I fear we'll produce a big cluttered class at the end. But we'll see if those method are necessary (or if we move this to the weighting instead) and fix later.

PR #798, opened 2016-09-19
Interpolates elevation for tunnels and bridges, fixes issue #713

highsource:
This is the fix for #713. It adds elevation interpolation for tunnels and bridges.

For this to work you'll need to turn on the `generic` flag encoder:
```
graph.flag_encoders=generic,car,foot
```

This will probably also require using 8 bytes for flags:

```
graph.bytes_for_flags=8
```

Elevation can be turned off via:

```
graph.elevation.interpolate.bridges=false
graph.elevation.interpolate.tunnels=false
```

(Both settings are `true` by default.)

See [this tweet](https://twitter.com/orless/status/777916527567839232) for example.

How it works.

For each tunnel (or bridge) edge, the algorithm first finds a connected component of tunnel (bridge) edges. Thus we can find "inner" and "outer" nodes of the tunnel. Outer nodes have tunnel and non-tunnel edges whereas inner nodes only have tunnel edges. Outer nodes are considered to be on the Earth surface, therefore we can trust the elevation provider for their elevation.

Once we've found inner and outer nodes, we'll interpolate elevations of inner nodes based on the elevations of the outer nodes. Interpolation method depends on the number of outer nodes. Cases for 0 or 1 outer nodes are trivial. For 2 outer nodes we use linear interpolation. For 3 outer nodes we use planar interpolation. For 4 and more outer nodes we use [the following method](http://math.stackexchange.com/a/1930758/140512).

This is how we get elevations of the inner nodes. Elevation of pillar points is then calculated using linear interpolation using distances from the pilar node to the tower nodes.

Few further hints.

* I had to add some default implementation to `DataFlagEncoder.flagsDefault`, as it was called during initialization.
* I had to add `motorroad` and `forrestry` to the `highwayList` in `DataFlagEncoder`, otherwise GraphHopper was failing with `Graph not prepared for highway=forestry` error during initialization.
* `DataFlagEncoder` seems to require 26 bits which I think is a bit too much. It does not feel completely finished yet, but is sufficient for this issue.

I'd be grateful for the review and merge.


karussell:
> Few further hints.

Thanks! The DataFlagEncoder is a bit unpolished at the moment.

> DataFlagEncoder seems to require 26 bits which I think is a bit too much

This is correct and will continue to grow probably to 64bit. E.g. we store the maximum speed for cars currently forward&backward leading already to 2*5bits - we could probably do a bit better if we do not store multiples of 5 but use e.g. 0,1,2,3,4, ... for common speeds like 5, 10, 20, 30, 50?

Also the intention is to fully move everything to this encoder (as often "raw data" is necessary) and potentially breaking the `long`-flags into two `int`s and adding a third int, then providing `edge.getData(int)`. This way we are no longer limited by 64bits. But for the most use cases 64 will be sufficient, at least this is the plan :)

> Elevation can be turned off via:

You mean interpolation here? I would not make this configurable. I would enable interpolation as soon as elevation is enabled and the DataFlagEncoder is there.

highsource:
I understood the intention of the `DataFlagEncoder` but I also think we'll need more that 64 bits in the long run. Optimizing speed bits would give 2-4 bits most (I don't think we can go below 8 different speeds, that's at least 3 bits per direction, but 4 bits is more realistic).

>> Elevation can be turned off via:
> You mean interpolation here? I would not make this configurable. I would enable interpolation as soon as elevation is enabled and the `DataFlagEncoder` is there.

Yes, interpolation. I've added these configuration options for a backward compatibility path. I'm OK to drop it. Should I do it or could you do it upon merge?

karussell:
> (I don't think we can go below 8 different speeds, that's at least 3 bits per direction, but 4 bits is more realistic).

What is even worse: we need two speed values for hgv and potentially others too. Highly likely a dynamic storage size per edge should be introduced but then we cannot easily use it like an array and would need a binary search for every index. Not sure yet how to solve this efficiently.

> I've added these configuration options for a backward compatibility path.

Currently no one uses DataFlagEncoder so we should be save :)

> I'm OK to drop it. Should I do it or could you do it upon merge?

I can do this

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -872,6 +930,41 @@ public void postProcessing() {
...
+		if (!ghStorage.getEncodingManager().supports("generic")) {
+			logger.warn("Elevation interpolation of bridges or tunnels is turned on, but generic flag encoder is not enabled. "
+					+ "Either enable generic flag encoder using graph.flag_encoders=generic,... "
+					+ "or turn off bridge and tunnel interpolation using graph.elevation.interpolate.bridges=false and graph.elevation.interpolate.tunnels=false.");
+			
+		} else {
+			final FlagEncoder genericFlagEncoder = ghStorage.getEncodingManager().getEncoder("generic");
+			if (!(genericFlagEncoder instanceof DataFlagEncoder)) {
+				throw new IllegalStateException("Generic flag encoder equired for elevation interpolation of bridges and tunnels is enabled but does not habe the expected type " + DataFlagEncoder.class.getName() + ".");
+			} else {
```
karussell: I know this is a bit of taste: but I prefer to avoid nested if-else and here one could avoid the else as the if stops in every case.

core/src/main/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolator.java
```diff
@@ -0,0 +1,114 @@
...
+import com.graphhopper.util.BreadthFirstSearch;
+import com.graphhopper.util.EdgeExplorer;
+import com.graphhopper.util.EdgeIterator;
+import com.graphhopper.util.EdgeIteratorState;
+import com.graphhopper.util.PointList;
+
+import gnu.trove.set.TIntSet;
+import gnu.trove.set.hash.TIntHashSet;
+
+public abstract class AbstractEdgeElevationInterpolator {
```
karussell: Would you mind to add one sentence description to every new class. I know the current state is not always this way but we should improve inline documentation a bit :)

core/src/main/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolator.java
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	private void interpolateElevationsOfPillarNodes() {
+		final EdgeIterator edge = storage.getAllEdges();
+		while (edge.next()) {
+			if (isStructureEdge(edge)) {
+				int firstNodeId = edge.getBaseNode();
+				int secondNodeId = edge.getAdjNode();
+
+				double lat0 = storage.getNodeAccess().getLat(firstNodeId);
```
karussell: I would assign nodeAccess to a local variable outside of the loop. Potentially the JVM is already doing this but it also reads a bit faster :)

core/src/main/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolator.java
```diff
@@ -0,0 +1,114 @@
...
+	private final NodeElevationInterpolator nodeElevationInterpolator;
+	private final ElevationInterpolator elevationInterpolator = new ElevationInterpolator();
+
+	public AbstractEdgeElevationInterpolator(GraphHopperStorage storage, DataFlagEncoder dataFlagEncoder) {
+		this.storage = storage;
+		this.dataFlagEncoder = dataFlagEncoder;
+		this.nodeElevationInterpolator = new NodeElevationInterpolator(storage);
+	}
+
+	protected abstract boolean isStructureEdge(EdgeIteratorState edge);
```
karussell: Can we rename this method? E.g. isInterpolableEdge or something similar?

core/src/main/java/com/graphhopper/reader/dem/ElevationInterpolator.java
```diff
@@ -0,0 +1,99 @@
+package com.graphhopper.reader.dem;
+
+import com.graphhopper.util.PointList;
+
+public class ElevationInterpolator {
+
+	public static final double EPSILON = 0.00001;
+	public static final double EPSILON2 = EPSILON * EPSILON;
+
+	public double calculateElevation(double lat, double lon, double lat0, double lon0, double ele0, double lat1,
```
karussell: It should be more clear what this method does. Especially as the parameters are somehow unexpected. Add a comment here?

karussell: Ah, now I understood. This is the case for 2 outer points, the other calculateElevation methods or for 3 outer points and 4+

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -396,8 +408,11 @@ public double getMaxspeed(EdgeIteratorState edge, int accessType, boolean revers
 
     @Override
     public long flagsDefault(boolean forward, boolean backward) {
-        // just pick car mode to set access values?
-        throw new RuntimeException("do not call flagsDefault");
+//         just pick car mode to set access values?
+//        throw new RuntimeException("do not call flagsDefault");
+    	// TODO This is called on each of the encoders so I had to replace the runtime exception with something,
+    	// but I'm not sure this is correct.
+    	return setAccess(0, forward, backward);
```
karussell: I'll check why this got called. But we could just remove this method and the super method will be called:
```java
long flags = speedEncoder.setDefaultValue(0);
return setAccess(flags, forward, backward);
```

core/src/test/java/com/graphhopper/reader/dem/BridgeElevationInterpolatorTest.java
```diff
@@ -0,0 +1,133 @@
...
+import com.graphhopper.storage.RAMDirectory;
+import com.graphhopper.util.EdgeIteratorState;
+import com.graphhopper.util.Helper;
+
+import gnu.trove.set.hash.TIntHashSet;
+
+public class BridgeElevationInterpolatorTest {
+
+	private static final double EPSILON= 0.0000000001d;
+	private static final ReaderWay BRIDGE_WAY;
```
karussell: I do not like mutable static variables that much - you do not write to it but it is allowed and could then introduce problems for multithreaded unit test execution. In test we also do not need to care about object creation so I'd vote to just remove the static and put the initialization in the `@Before` method.

core/src/test/java/com/graphhopper/reader/dem/BridgeElevationInterpolatorTest.java
```diff
@@ -0,0 +1,133 @@
...
+	private GraphHopperStorage graph;
+	private DataFlagEncoder dataFlagEncoder;
+	private BridgeElevationInterpolator bridgeElevationInterpolator;
+	
+	@Before
+	public void init(){
+		graph = new GraphHopperStorage(new RAMDirectory(), new EncodingManager("generic,foot", 8),
+				true, new GraphExtension.NoOpExtension()).create(100);
+
+		dataFlagEncoder = (DataFlagEncoder) graph.getEncodingManager().getEncoder("generic");
```
karussell: BTW: If you dislike casts you can also do:
```
dataFlagEncoder = new DataFlagEncoder();
FootFlagEncoder foot = new FootFlagEncoder();
graph = new GraphHopperStorage(new RAMDirectory(), new EncodingManager(Arrays.asList(dataFlagEncoder,foot), 8),
				true, new GraphExtension.NoOpExtension()).create(100);
```

core/src/test/java/com/graphhopper/reader/dem/TunnelElevationInterpolatorTest.java
```diff
@@ -0,0 +1,306 @@
...
+import com.graphhopper.routing.util.EncodingManager;
+import com.graphhopper.storage.GraphExtension;
+import com.graphhopper.storage.GraphHopperStorage;
+import com.graphhopper.storage.NodeAccess;
+import com.graphhopper.storage.RAMDirectory;
+import com.graphhopper.util.EdgeIteratorState;
+
+import gnu.trove.set.hash.TIntHashSet;
+
+public class TunnelElevationInterpolatorTest {
```
karussell: Do we really need a separate test for tunnels? What is different with them from the code concept? Or maybe create an abstract tester with two subclasses or one test with two instances?

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
```diff
@@ -56,17 +59,64 @@ public void testHighway() {
         edge = GHUtility.createMockedEdgeIteratorState(0, flags);
         assertEquals("_default", encoder.getHighwayAsString(edge));
     }
+    
+    @Test
+    public void testTunnel() {
```
karussell: Strange that this appears as change, as it is already there: https://github.com/graphhopper/graphhopper/blob/master/core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java#L61-L103

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
```diff
@@ -56,17 +59,64 @@ public void testHighway() {
...
+        edge = GHUtility.createMockedEdgeIteratorState(0, flags);
+        assertEquals("bridge", encoder.getTransportModeAsString(edge));
+        assertFalse(encoder.isTransportModeTunnel(edge));
+        assertTrue(encoder.isTransportModeBridge(edge));
+    }    
 
     @Test
     public void testHighwaySpeed() {
-        Map<String, Double> map = new HashMap<>();
+        Map<String, Double> map = new LinkedHashMap<>();
```
karussell: Ah, ups - thanks! Maybe we should require a LinkedHashMap for `encoder.getHighwaySpeedMap` etc?

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -373,6 +373,66 @@ public void testSRTMWithInstructions() throws Exception {
...
+        GHResponse rsp = tmpHopper.route(new GHRequest(43.74056471749763, 7.4299266210693755, 43.73790260334179, 7.427984089259056).
+                setAlgorithm(ASTAR).setVehicle(vehicle).setWeighting(weightCalcStr));
+        PathWrapper arsp = rsp.getBest();
+        assertEquals(351.38545671859913, arsp.getDistance(), .1); // Without interpolation: 356.79372559007476  
+        assertEquals(6, arsp.getPoints().getSize());
+        assertTrue(arsp.getPoints().is3D());
+        
+        assertEquals(Helper.createPointList3D(
+                43.7405648173518, 7.429926635993867, 17.0,
+                43.74016602501644, 7.42986703134776, 19.042999267578125, // Without interpolation: 23.0
```
karussell: Oh, the many digits after comma are a bit ugly as they suggests a false precision. Should we round it when interpolating e.g. to `cm` or even just `dm`?

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
highsource: I think `speedEncoder` was `null`, but I'll check.

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
highsource: I think it's a bit weird to pass have an implementation class like  `LinkedHashMap` in the signature. I'll think of options.

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
highsource: This is due to using `equals` of `PointList`. Normally doubles are compared with certain precision, this does not work when doing `equals` on the whole object.
I think I'll just check elevations with some epsilon.

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
highsource: Perhaps formatting? Spaces vs. tabs.

core/src/test/java/com/graphhopper/reader/dem/TunnelElevationInterpolatorTest.java
highsource: Tunnel test is actually the main test. Test for bridges is to smoke-test (if it works at all) as well to check interpolation of pillar nodes. So that's the same concept but these tests test different parts of interpolation.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
highsource: I think `speedEncoder` was `null`, but I'll check again. I did not want to change the `DataFlagEncoder` a lot, that did not feel completely finished yet and it was not in the focus.

core/src/main/java/com/graphhopper/GraphHopper.java
highsource: Ok, that's really a matter of taste. I prefer this type if/else blocks, make me think less (like I don't have to check if the if block stops or not). I'd like to leave it as is if it's not that important for you.

core/src/main/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolator.java
highsource: Will do.

core/src/main/java/com/graphhopper/reader/dem/ElevationInterpolator.java
highsource: I'll give these methods more elaborate names.

highsource:
Thank you for the review! I'll make the changes and commit shortly.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
highsource: Yes, `speedEncoder` is `null`. I'll leave `return setAccess(0, forward, backward);` for now, I've also filed #800 for this.

core/src/main/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolator.java
highsource: ps. We could also use [License Maven Plugin](http://www.mojohaus.org/license-maven-plugin/), it can automatically add or check license headers. I used it for a number of GPL projects and it worked very well.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
karussell: Ah, yes. One would need the carFwdFlagEncoder here ... but we also have backward, hmmh. I would like to get rid of this.

core/src/test/java/com/graphhopper/reader/dem/TunnelElevationInterpolatorTest.java
karussell: Ok, that is good.

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
karussell: I meant in general: when we store elevation with so many digits people could assume that we have also very precise data, so regardless of this ugliness in the test it might be wise to cut off precision?

highsource: Makes sense.

highsource: I think I'll round it to 0.5m. I think that's the precision we get from elevation provider so we shouldn't pretend our interpolation is much better.
Next, 0.5m would allow us to use just 16 bits for elevations (14 bits for 0...16384, 1 bit for sign, 1 bit for .0/0.5).

core/src/test/java/com/graphhopper/reader/dem/TunnelElevationInterpolatorTest.java
```diff
@@ -183,55 +187,56 @@ public void interpolatesElevationOfTunnelWithThreeOuterNodes() {
...
+        assertEquals(40, na.getElevation(4), PRECISION);
+        assertEquals(20, na.getElevation(5), PRECISION);
+        assertEquals(30, na.getElevation(6), PRECISION);
+        assertEquals(40, na.getElevation(7), PRECISION);
+    }
+
+    @Test
+    public void interpolatesElevationOfTunnelWithFourOuterNodes() {
+
+        // @formatter:off
```
karussell: You should be able to trick all formatters via `//` on all lines

core/src/test/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolatorTest.java
```diff
@@ -0,0 +1,83 @@
...
+        edgeElevationInterpolator = createEdgeElevationInterpolator();
+
+        interpolatableWay = createInterpolatableWay();
+        normalWay = new ReaderWay(0);
+        normalWay.setTag("highway", "primary");
+    }
+    
+    @After
+    public void tini() {
+        graph.close();
```
karussell: Here it is safer to use Helper.close(graph) as this checks for `null` and null could happen e.g. if init failed.

BTW: tini is a fun name :) I used `setUp` and `tearDown`

core/src/test/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolatorTest.java
```diff
@@ -0,0 +1,83 @@
...
+import com.graphhopper.storage.GraphHopperStorage;
+import com.graphhopper.storage.RAMDirectory;
+import com.graphhopper.util.EdgeIteratorState;
+
+import gnu.trove.set.hash.TIntHashSet;
+
+/**
+ * @author Alexey Valikov
+ */
+public abstract class AbstractEdgeElevationInterpolatorTest {
```
karussell: The abstract test class should be named XYTester otherwise it will be executed automatically from maven and fail

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
karussell: 0.5 is good here but might not be sufficient in general, not sure. E.g. I know from a customer using GH with more precise elevation data plus doing interpolation in between so they really need `cm` precision at least otherwise they would have to big steps in-between making some other routing algorithms a bit more ugly

core/src/test/java/com/graphhopper/reader/dem/TunnelElevationInterpolatorTest.java
highsource: No, not Eclipse. :( It strips leading spaces in comments.

core/src/test/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolatorTest.java
highsource: No problem to rename, but I don't think Maven/JUnit does not execute test in abstract classes (I've just rechecked it).

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
highsource: Ok, unerstood. 0.01m then.

core/src/test/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolatorTest.java
karussell: Ah, ok then you can keep it. Maybe this changed in some more recent maven version.

highsource:
Done.

.gitignore
```diff
@@ -2,7 +2,7 @@ target/
 *~
 TODO.txt
 *-gh/
-*.osm
+*.osm*
```
karussell: This is a bit dangerous as we have many *osm.gz files under core/files/ ... probably better to create a folder like osm-data (and put the temp osm files there?) which is then ignored?

core/src/main/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolator.java
```diff
@@ -0,0 +1,161 @@
...
+import com.graphhopper.util.PointList;
+
+import gnu.trove.set.TIntSet;
+import gnu.trove.set.hash.TIntHashSet;
+
+/**
+ * Abstract base class for tunnel/bridge edge elevation interpolators. This
+ * class estimates elevation of inner nodes of a tunnel/bridge based on
+ * elevations of entry nodes. See
+ * <a href="https://github.com/graphhopper/graphhopper/issues/713">#713</a> for
```
karussell: I do not link to github explicitly inside of the code and just say `#xy`. This way we can easily migrate to other issue trackers in the future.

core/src/main/java/com/graphhopper/reader/dem/ElevationInterpolator.java
```diff
@@ -0,0 +1,156 @@
...
+            double ele = 0;
+            for (int index = 0; index < size; index++) {
+                ele += eles[index] * vs[index] / v;
+            }
+            return roundToPrecision(ele);
+        }
+    }
+
+    public double roundToPrecision(double value) {
+        return Math.round(value / PRECISION) * PRECISION;
```
karussell: See also Helper.round (not sure if applicable though)

karussell:
Thanks again!

Merged via https://github.com/graphhopper/graphhopper/commit/90a3844cd3f9768cef97d4db491d064cc1e08bfc

PR #802, opened 2016-09-20
Added intergration test based on RNV, also GtfsHelper with useful time routines

highsource:
I've added a `GtfsHelper` with basic time conversion routines.

I've also added an intergation test based on GTFS feed from RNV (https://opendata.rnv-online.de/datensaetze/gtfs-general-transit-feed-specification). It is open data (the license is [	
dl-de-by-2.0](https://www.govdata.de/dl-de/by-2-0), we'll need to add attribution). I think it's good to test with real datasets from Germany.


highsource:
ps. I'll be contributing via pull requests, is this OK for you?

karussell:
> ps. I'll be contributing via pull requests, is this OK for you?

Sure, added you also to the commiters - feel free to create branches and create PR against them.

README.md
```diff
@@ -125,3 +125,7 @@ Here is a list of the more detailed features including a link to the documentati
  * scales from small indoor-sized to world-wide-sized graphs
  * Find nearest point on street e.g. to get elevation or 'snapp to road'
  * Enables the [map matching component](https://github.com/graphhopper/map-matching) (flexibility only)
+
+# Attribution
```
karussell: Maybe we should link to a separate Readme inside the gtfs reader module? As I fear there will be more than one of these attribution statements necessary and it is not really related to all GH users e.g. the once not using public transport. Or use a different test DB where this is not necessary?

karussell:
Looks good to me. @michaz will comment & merge

README.md
highsource: I think all of the GTFS feeds in Germany require attribution in one way or another so we can't avoid attribution statements in the long run.
I've added a separate `README.md` for `reader-gtfs`, moved the attribution statement there.

michaz:
Nice. I've been using the Berlin-Brandenburg dataset to play around with locally and was hesitant to check it in, but yes, it's good to have integration tests against real data. Thanks!

Issue #803, opened 2016-09-22
Android build fails on build server


karussell:
Due to latest android version update we probably need to update the Android SDK there, until then we do not have snapshots of GH :/

devemux86:
What server you mean?

karussell:
Our internal build/dev server ... or we need to somehow make creation of snapshots possible for travis. But this can be more complex due to Android SDK setup. Not sure how others handle this.

devemux86:
If it can help in Mapsforge and VTM I have organized lately their snapshots publishing, but via Gradle (see [here](https://github.com/mapsforge/vtm/blob/master/deploy.gradle)).

karussell:
Is it working on travis?

devemux86:
>Is it working on travis?

What do you mean?
The publishing is on Sonatype ([here](https://oss.sonatype.org/content/repositories/snapshots/org/mapsforge/)), Travis is still used for code testing.

karussell:
When we run tests on travis many people just do the deploy there for the snapshots, are you doing this on travis or on your own machine? On my own machine I had this working with the maven android stuff, just need to update the Android SDK I think

devemux86:
>When we run tests on travis many people just do the deploy there for the snapshots, are you doing this on travis or on your own machine?

You mean the GH `travis-build.sh` script?

>On my own machine I had this working with the maven android stuff, just need to update the Android SDK I think

A lot of things have [broken](https://github.com/simpligility/maven-android-sdk-deployer/issues) lately with Android + Maven, I hope you make it working.

karussell:
> You mean the GH travis-build.sh script?

Yes, if we would add 'deploy' there we would get snapshots deployed (easily) but without android

devemux86:
That's a good thought, but what about Sonatype credentials?

For Android see our [.travis.yml](https://github.com/mapsforge/vtm/blob/master/.travis.yml), very simple and works for both Android and Java modules in the project.

karussell:
> That's a good thought, but what about Sonatype credentials?

There was either some travis magic involved I think

> For Android see our .travis.yml, very simple and works for both Android and Java modules in the project

Ah, this might work for us too. The android part is interesting and wasn't existent when I looked into it the last time I think.

devemux86:
See [here](https://docs.travis-ci.com/user/languages/android) for all options regarding Android.
You can even enable emulators for running tests.

mprins:
[...]

Issue #701, opened 2016-04-13
Make GraphHopper compiling and running on JDK9


karussell:
Out of couriosity I tried JDK9 and got already lots of success :)
- [x] runs out of the box in JDK9 with jdk8 compiled jars
- [x] Maven compiles with [a small workaround](http://stackoverflow.com/a/36605759/194609) in the parent pom.xml. The workaround won't be necessary for a new maven release I think
- [x] most tests pass
- [x] except the memory mapping tests, which return 'RuntimeException: unable to unmap the mapped buffer'. [Here](http://mail.openjdk.java.net/pipermail/core-libs-dev/2015-September/035055.html) is some discussion about this.

mprins:
You could specify two profiles, one for jdk8 and one for jdk9 and as part of the profile set up includes/excludes for the surefire an failsafe tests. I would not recommend doing it from the commandline.

something like:

```xml
<profiles>
  <profile>
    <id>jdk8</id>
    <activation>
      <activeByDefault>false</activeByDefault>
      <jdk>1.8</jdk>
    </activation>
    <properties />
    <build>
      <plugins>
        <plugin>
          <artifactId>maven-failsafe-plugin</artifactId>
        </plugin>
        <plugin>
          <artifactId>maven-surefire-plugin</artifactId>
        </plugin>
      </plugins>
    </build>
  </profile>
  <profile>
    <id>jdk9</id>
    <activation>
      <activeByDefault>false</activeByDefault>
      <jdk>1.9</jdk>
    </activation>
    <properties />
    <build>
      <plugins>
        <plugin>
          <artifactId>maven-failsafe-plugin</artifactId>
          <configuration>
            <systemPropertyVariables />
          </configuration>
        </plugin>
        <plugin>
          <artifactId>maven-surefire-plugin</artifactId>
          <configuration>
           <!-- this uses a method described here: 
                  https://issues.apache.org/jira/browse/SUREFIRE-131 to exclude a singe 
                  test method, ofcourse any config can ce used here -->
            <test>!SomeClassTest#methodtoskipTest</test>
          </configuration>
        </plugin>
      </plugins>
    </build>
  </profile>
</profiles>
```
Will activate the relevant config based on the JDK running the build.

karussell:
Thanks this sounds good. Do you know of an annotation like `@IgnoreIfEnvVarIsXY`?

[...]

PR #806, opened 2016-09-23
Travis-CI build enhancements

mprins:
As discussed in  https://github.com/graphhopper/graphhopper/issues/803#issuecomment-249179426 this PR brings JDK 9 builds on Travis-CI and automatic deployment of snapshots to the central repository.

  - [x] Start testing on Java 9 with Maven 3.3.9
  - [x] upgrade maven plugins and require Maven >= 3.2
  - [x] set up for publishing snapshots to maven central

The latter needs to have the secure keys generated and added in the `.travis.yml` (there are some placeholders there now) so that after a succesful build of the master branch the snapshot artifacts are deployed.

This build no longer uses the `/core/files/travis-build.sh` not sure if that can be removed, it has some out-commented npm stuff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,12 +45,10 @@ before_install:
          sudo rm /etc/mavenrc;
      fi
 
-install:
-  # install without any testing to get all dependencies in place
-  # - mvn install -Dmaven.test.skip=true -B -V -fae -T2.2C
+install: true
 
 script:
-  - mvn clean test verify
+  - mvn clean test verify -B
   #- ./core/files/travis-build.sh
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,72 @@
 language: java
+sudo: false
+matrix:
+  fast_finish: true
+  include:
+      # Java 9 needs to be manually installed/upgraded
+      # see: https://github.com/travis-ci/travis-ci/issues/2968#issuecomment-149164058
+    - jdk: oraclejdk9
+      env: JVM=latest
+      sudo: required
+      dist: trusty
+      group: edge
+  allow_failures:
+    - jdk: oraclejdk9
+
+env:
+  global:
+    - BASEURL=https://www-eu.apache.org/dist/maven/maven-3/VERSION/binaries/apache-maven-VERSION-bin.zip
+    - FILE=apache-maven-VERSION-bin.zip
+    - DIR=apache-maven-VERSION
+    - VERSION=3.3.9
+    - secure: "CI_DEPLOY_USERNAME=central_username"
+    - secure: "CI_DEPLOY_PASSWORD=central_password"
+
 jdk:
   - oraclejdk8
-# do not install anything instead return true via unix command true
-install: true
-script: ./core/files/travis-build.sh
+
+before_install:
+   # update maven
+   - if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+         wget --no-check-certificate $(echo -n $BASEURL | sed -e 's#VERSION#'$VERSION'#g');
+         unzip -qq $(echo -n $FILE | sed -e 's#VERSION#'$VERSION'#');
+         export M2_HOME=$PWD/$(echo -n $DIR | sed -e 's#VERSION#'$VERSION'#');
+         export PATH=$M2_HOME/bin:$PATH;
+     fi
+   # update java 9
+   - if [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$JVM" == "latest" ]; then
+         sudo apt-get update -qq;
+         sudo /bin/echo -e oracle-java9-installer shared/accepted-oracle-license-v1-1 select true | sudo debconf-set-selections;
+         sudo apt-get install -y oracle-java9-installer;
+         sudo apt-get install -y oracle-java9-unlimited-jce-policy;
+         sudo update-java-alternatives -s java-9-oracle;
+     fi
+   - if [ "$TRAVIS_JDK_VERSION" == oraclejdk9 ]; then
+         sudo rm /etc/mavenrc;
+     fi
+
+install:
+  # install without any testing to get all dependencies in place
+  # - mvn install -Dmaven.test.skip=true -B -V -fae -T2.2C
+
+script:
+  - mvn clean test verify
+  #- ./core/files/travis-build.sh
+
+after_success:
+  # deploy snapshot artifacts to maven central
+  - if [ "$TRAVIS_JDK_VERSION" == "oraclejdk8" ] &&
+       [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
+         travis_wait mvn deploy --settings core/files/settings.xml -DskipTests=true -B;
+    else
+        echo "Not deploying snapshot artifact for $TRAVIS_BRANCH";
+    fi
+
 notifications:
   email:
     - github@graphhopper.com
 
-# enable container-based stack
-sudo: false
+cache:
+  directories:
+  - $HOME/.m2
+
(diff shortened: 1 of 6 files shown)


karussell:
Already really nice - thanks!

mprins:
@karussell this needs a little something on your side to add the secure keys to the travis file; deployment should the start working after merging into master.

Also I've set up to allow test failures on java 9, and it's showing some with a message of `unable to unmap the mapped buffer`. eg. https://travis-ci.org/graphhopper/graphhopper/jobs/162408743#L1011

karussell:
Thanks - will read next week about how to setup the secure keys.

Regarding the JDK9: in #701 I already stumbled over this ugliness, but I'm unsure how to fix this yet (see the link there). Maybe we can exclude or ignore (for now) a single test for a specific JDK version?

.travis.yml
```diff
@@ -1,12 +1,70 @@
...
+     fi
+   - if [ "$TRAVIS_JDK_VERSION" == oraclejdk9 ]; then
+         sudo rm /etc/mavenrc;
+     fi
+
 install: true
-script: ./core/files/travis-build.sh
+
+script:
+  - mvn clean test verify -B
```
karussell: I've looked up what the -B means (--batch-mode) ... why is this necessary?

core/pom.xml
```diff
@@ -74,41 +74,43 @@
...
-                </plugin>                
-                
-                <!-- create jar with test classes to be reused in other projects like external or our reader-osm -->                
+                </plugin>
+
+                <!-- create jar with test classes to be reused in other projects like external or our reader-osm -->
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-jar-plugin</artifactId>
+                    <version>3.0.2</version>
```
karussell: As you already changed this in the parent pom, this is not necessary here (?) and when not being explicit here, we can then easier upgrade versions later on in the parent pom

web/pom.xml
```diff
@@ -80,22 +80,23 @@
...
+
     </dependencies>
 
     <build>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>                
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.5.1</version>
```
karussell: Same here: we could remove the version (?)

.travis.yml
mprins: it would prevent the build waiting until it times out if any interactive response would be required

core/pom.xml
mprins: that would only work if the parent pom has a pluginManagement section where the version is specified otherwise the result of the version resolution of the plugin is undefined (maven will just pick one that's available), see: https://maven.apache.org/pom.html#Plugin_Management

web/pom.xml
mprins: as above, not really a good idea

core/pom.xml
karussell: Ups, thanks for pointing this out :) ! Maybe it would then be better to have a plugin section in the root pom or are there disadvantages?

karussell:
Uhm, strange. How it comes I have a comment in my emails regarding the 2 profiles but not here?

> You could specify two profiles, one for jdk8 and one for jdk9...

karussell:
Nice! Deployment seems to work (according to the time stamps of the snapshot) but still something timed out: https://travis-ci.org/graphhopper/graphhopper/jobs/163212260

>   travis_jigger $! $timeout $cmd

Before I can merge it: would you mind to sign the [CLA](https://github.com/graphhopper/graphhopper/blob/master/.github/CONTRIBUTING.md) (you do not give away your rights, just agree to the Apache License) and send this to us. Or just send me your email so that I can send you an easier to handle digital signing doc.

devemux86:
>travis_jigger $! $timeout $cmd

That probably comes from [travis_wait](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received), is it needed (default 10min)?

core/pom.xml
mprins: You can setup the plugins in the parent pom like:
```xml
<build>
    <pluginManagement>
            <plugins>
                <plugin>
                    <groupId>org.apache.maven.plugins</groupId>
                    <artifactId>maven-compiler-plugin</artifactId>
                    <version>3.5.1</version>
                    <configuration>
                        <source>${java.version}</source>
                        <target>${java.version}</target>
                    </configuration>
                </plugin>
...
            </plugins>
        </pluginManagement>
```
and then in the build section of the child projects and possibly the parent as well specify the plugins to be used for the build of that artifact. You can find a working example in https://github.com/B3Partners/flamingo-ibis/blob/master/pom.xml

the downside is that there is some duplication of plugins (they both the `pluginManagement` and the `build` section of a pom will have a `plugins` section)

mprins:
the `travis_wait` may not be necessary, if there is output within the 10 mins. if the deploy produces no log output for 10min or more travis will fail the deploy as it seems stalled (to travis)

mprins:
>Uhm, strange. How it comes I have a comment in my emails regarding the 2 profiles but not here?
>>You could specify two profiles, one for jdk8 and one for jdk9...

@karussell I commented on the original issue #701

karussell:
Merged - thanks a lot again @mprins - really nice now :) !

We'll improve this now in an iterative manner later on ...

karussell:
Currently the snapshots are not deployed, e.g. see https://travis-ci.org/graphhopper/graphhopper/jobs/168299877

It says: 
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.7:deploy (default-deploy) on project graphhopper-parent: Failed to deploy artifacts: Could not transfer artifact com.graphhopper:graphhopper-parent:pom:0.9-20161017.130625-2 from/to ossrh (https://oss.sonatype.org/content/repositories/snapshots): Failed to transfer file: https://oss.sonatype.org/content/repositories/snapshots/com/graphhopper/graphhopper-parent/0.9-SNAPSHOT/graphhopper-parent-0.9-20161017.130625-2.pom. Return code is: 401, ReasonPhrase: Unauthorized. -> [Help 1]
```

Any clue what changed?

Also it says:
```
This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid and setguid executables.

If you require sudo, add 'sudo: required' to your .travis.yml
```

And we have `sudo: false` in the root but in the matrix we have `sudo: required`. Is this correct?

mprins:
> Any clue about this?
The username and password are invalid; looking at the .travis.yml the are:
```
    - secure: "CI_DEPLOY_USERNAME=central_username"
    - secure: "CI_DEPLOY_PASSWORD=central_password"
```

those still need to be update to your (encrypted) credentials as I wrote up top

>> The latter needs to have the secure keys generated and added in the `.travis.yml` (there are some placeholders there now) so that after a successful build of the master branch the snapshot artifacts are deployed.

see: https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml (and https://docs.travis-ci.com/user/encryption-keys )

ps When using the Travis gem I would not use the --add option, it will ruin the `.travis.yml` layout


> And we have sudo: false in the root but in the matrix section we have sudo: required. Is this correct?

We are changing to a different build image as part of the build matrix; using 
```yaml
sudo: required
dist: trusty
group: edge
```
the sudo is required to update the java 9 jdk on that machine to a current version (there is no relation to the snapshot deployment failure)

karussell:
> those still need to be update to your (encrypted) credentials as I wrote up top

I think this is a merge issue (or something else) as I've definitely updated them - thanks for pointing me to this :) !

> ps When using the Travis gem I would not use the --add option, it will ruin the .travis.yml layout

what do you mean here?

> We are changing to a different build image as part of the build matrix; using 

Thanks for the explanation!

karussell:
Works now - thanks again @mprins :) ! E.g. see [here](https://oss.sonatype.org/content/groups/public/com/graphhopper/graphhopper-core/0.9-SNAPSHOT/)

karussell:
Just as a reference there is a web UI to create those keys: http://rkh.github.io/travis-encrypt/public/index.html

PR #807, opened 2016-09-26
Use Weighting.calcMillis instead of Path.calcMillis, fixes #393

### LINKED ISSUE #393: Improve time calculation for turn costs
If turn costs are enabled the time delay is added only when doing the routing. We have to add the time delay in Path.calcMillis too. Probably we need a separate method in TurnWeighting for this to avoid mixing weight and time.

karussell:
This allows the user to create a custom Weighting which also changes the time calculation and a time delay can be now stored in the edge flags

Fixes #393 

TODOs:

 - [ ] add a test for this new flexibility
 - [ ] test performance

Would highly appreciate feedback as this changes the internals a bit. Also if we should assume for now that turn costs are in seconds or if we need a separate method (getTurnWeight vs. getTurnMillis)

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/AStar.java
+++ b/core/src/main/java/com/graphhopper/routing/AStar.java
@@ -147,7 +147,8 @@ public class AStar extends AbstractRoutingAlgorithm {
 
     @Override
     protected Path extractPath() {
-        return new Path(graph, flagEncoder).setWeight(currEdge.weight).setSPTEntry(currEdge).extract();
+        return new Path(graph, weighting).
+                setWeight(currEdge.weight).setSPTEntry(currEdge).extract();
     }
 
     @Override
--- a/core/src/main/java/com/graphhopper/routing/AStarBidirection.java
+++ b/core/src/main/java/com/graphhopper/routing/AStarBidirection.java
@@ -153,7 +153,7 @@ public class AStarBidirection extends AbstractBidirAlgo {
 
     @Override
     protected Path createAndInitPath() {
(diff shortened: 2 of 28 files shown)


karussell:
Merged via: https://github.com/graphhopper/graphhopper/commit/0bf6b14b2982638db5e7002a797f4d1c657e32a4

karussell:
Hmmh, I'm not satisfied here. We would need to break API again if we have to calculate time, distance, transfers, weight in Dijkstra, currently we just calculate weight. I'm thinking about a new class `WeightingResult` with time, distance, weight, ... and just one method:

```java
class Weighting {
  void calculate(WeightingResult, EdgeIteratorState, reverse, prevEdge);
  // or for easier backward compatibility returning the first parameter:
 // WeightingResult calculate(WeightingResult, EdgeIteratorState, reverse, prevEdge);
}
```

Then we could easily fill the time or weight etc in this method. It would be even more customizable as we could subclass WeightingResult and add parameters. We could even replace `TurnWeighting` when we store the turnCosts in the WeightingResult (or a subclass).

easbar:
But isn't it like we are calculating only the weight while the routing algorithms are running and the distance/time is calculated while we are unpacking the path path anyway ?

karussell:
`WeightingResult` is definitely the wrong name ;) ... as I meant e.g. time as input parameter.

For time-dependent algorithms we need the time as additional input parameter to consider time dependent costs (weights).

Maybe we allow input parameters like time (and transfers, turns) and output parameters like travel cost, travel time and turn costs?

PR #812, opened 2016-10-02
Upgrade Leaflet to version 1.0.1

fbonzon:
Allow fractional zoom levels (new feature of Leaflet 1.0), and make the route geometry exactly fit the map viewport (with margin of 42 pixels).

removeFrom() → remove() (function name changed in Leaflet 1.0).

Demo: [françois.ch/graphhopper-leaflet/](http://françois.ch/graphhopper-leaflet/)


karussell:
Thanks!

There seems to be some issues:

 * cannot drag or click the markers (when changing the layers it appears for a tiny time, maybe some z-layer problem)
 * no popup shows up when clicking on the turn instructions
 * still some issues with scale remove `TypeError: scaleControl.remove is not a function`

fbonzon:
Should I rebase this PR to HEAD, so you have no merge conflicts ?

Did you bring your own dev instance? Missing remove() hints you are still using 0.7.7 Leaflet version. Clear cache? Did you run "npm update" to upgrade Leaflet ?

I don't see these issues on my demo: [françois.ch/graphhopper-leaflet/](http://françois.ch/graphhopper-leaflet/)

karussell:
Ah, ok. Will try what you suggested :)

karussell:
It works now. 

The only problem I now have is a bit blurred impression of the map - is this due to the fractional zoom? Should we disallow this for now?

E.g. compare the [current version](https://graphhopper.com/maps/?point=52.536834%2C13.302742&point=52.519057%2C13.486248&layer=Lyrk) with [a local version](http://localhost:8989/?point=52.536834%2C13.302742&point=52.519057%2C13.486248&layer=Lyrk). As Lyrk supports high resolution and it looks good in the current version I think this might be either a bug in leaflet or due to some misconfiguration.

karussell:
Hmmh, I cannot reproduce the issue on your deployment - see [here](http://françois.ch/graphhopper-leaflet/?point=52.536834%2C13.302742&point=52.519057%2C13.486248&layer=Lyrk). Strange.

karussell:
Oh, even more strange: locally the issue seems to be fixed too ...

karussell:
Merged - thanks!

karussell:
This is now also deployed!

fbonzon:
I don't have access to your localhost version, and can't see this blurred effect you experienced.

Yes, this is due to fractional zoom levels. Tiles look a bit different, because they can't be streched with any ratio now, but Leaflet is smart and uses tiles from next zoom level when needed. I tested on deployed version and don't see issues.

If you don't like the new fractional zoom or get complaints, just remove this line:

https://github.com/graphhopper/graphhopper/blob/dab77ef1145e60f383ceace1712d0aa618d68204/web/src/main/webapp/js/map.js#L53

karussell:
Yes, I do not see this anymore. Maybe a cache issue or something. We'll wait for user complaints then before disabling fractional zoom.

PR #814, opened 2016-10-02
Add *.html to extensions where we normalize line endings on checkout

fbonzon:
Affects only 1 file in the repository:

`/web/src/main/webapp/index.html`

Sort lines by extension.


karussell:
Thanks!

Issue #754, opened 2016-07-13
Routing through gates that don't have restrictions for car profile


HarelM:
Hi,

Please see our issue here related to routing through gates - in short - car routing doesn't pass through gates that doesn't have an access restriction when it should
IsraelHikingMap/Site#185
You can play around with the routing here - the following case is where a small settlement has an entrance gate that can be passed but still should be mapped:
http://israelhiking.osm.org.il/#/15/32.9176/35.5987?s=aUhQ0SBk9d

I think this also describes it:
https://graphhopper.com/maps/?point=32.913603%2C35.599136&point=32.911135%2C35.600359&locale=en-us&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale

We are using GH version 0.6 on windows with latest JVM available.

karussell:
Related #236

boldtrn:
We have this issue as well. In general I think this is mostly an OSM issue and the access=true tags should be added to barriers in general. Assuming access=true for gates might result in dangerous routes.

zstadler:
I suggest taking the same approach used for routing through fords - add an optional configuration setting. As a result, no changes to the general policy are made and local mapping policies can be easily applied.

I assume that means adding 

```javascript
        this.setBlockByDefault(properties.getBool("block_by_default", true));
```
to ```CarFlagEncoder(PMap properties)```, right!?

karussell:
Would be possible, yes. But I would rather avoid tweaking the relative 'static' XYFlagEncoders and instead improve the new DataFlagEncoder where you can use the GenericWeighting. But the DataFlagEncoder currently does not support barriers, maybe you would have to provide a PR for this :) ?

zstadler:
I can provide a PR of the change to CarFlagEncoder, in order to provide an interim solution to this issue.

boldtrn:
I think we can close this issue with the PR or should we create a solution for all flag encoders?

PR #815, opened 2016-10-02
Add an optional configuration setting for routing through barriers

zstadler:
An interim solution for issue  graphhopper/graphhopper#754


karussell:
Thanks. This is somehow mixed with a new car4wd profile? Sounds interesting but should be put in another PR.

zstadler:
It was not my intention to mix the two topics. 

I was not aware that if I continue development on the master branch of the forked repository, after submitting the pull request, will cause a change in the contents of the pull request.

What would be the best practice to recover form this unintended mix-up?

karussell:
We can close here and you can create one branch for every new feature that you intend to merge (Then you can easily create one PR per feature branch). And then in your master you merge all the features.

zstadler:
OK. Dropping this PR.
Opened PR #817 instead

PR #817, opened 2016-10-03
Add an optional configuration setting for routing through barriers

zstadler:
Interim solution for Issue #754

PR #819, opened 2016-10-03
Add car4wd routing profile

zstadler:
Allow routing through tracks regardless of their tracktype value.
Including support for block_fords and block_barriers config settings.


karussell:
Interesting. Would you mind to add some tests defining what should be different to a normal `car`?

zstadler:
Notes:
1. The prioritization by ```highway``` tag, originally taken from the ```motorcycle``` profile, was removed because I was unable to observe its impact.
2. The ```car4wd``` test suite is identical to the ```car``` test suite except for the test concerning ```tracktype=grade4```.

karussell:
> The car4wd test suite is identical to the car test suite except for the test concerning tracktype=grade4.

Could you extend the car test and just overload the method which is different?

karussell:
Thanks for the fix. I still cannot see the difference to car just from source code. Would you mind to explain? E.g. car currently also also grade4.

zstadler:
[This section of the ```acceptWay``` method](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java#L195-L199) is removed in Car4WDFlagEncoder.java:

```java
@@ -192,12 +76,6 @@
             return 0;
         }

-        if ("track".equals(highwayValue)) {
-            String tt = way.getTag("tracktype");
-            if (tt != null && !tt.equals("grade1") && !tt.equals("grade2") && !tt.equals("grade3"))
-                return 0;
-        }
-
         if (!defaultSpeedMap.containsKey(highwayValue))
             return 0;
```

As a result, the value of the ```tracktype``` tag is no longer a criteria for allowing the route through a ```highway=track``` way.

As seen above, the ```car``` profile does not use ```tracktype=grade4```, and this is also checked by the [corresponding test](https://github.com/graphhopper/graphhopper/blob/master/core/src/test/java/com/graphhopper/routing/util/CarFlagEncoderTest.java#L58-L59).

karussell:
Thanks. I'll then also remove the confusing assignment of grade4+5 speeds in car.

karussell:
In the commit message you stated:

 * Add an optional configuration setting for blocking barriers by default

This would be good but I cannot see this (obviously) in the code.

zstadler:
> In the commit message you stated:

> - Add an optional configuration setting for blocking barriers by default

> This would be good but I cannot see this (obviously) in the code.

This was previously merged in [Add an optional configuration setting for routing through barriers #817](817), and already available in this profile.

PR #826, opened 2016-10-08
Fix broken scalecontrol when switching between km to miles.

ratrun:
Switching between imperial and metric units got broken because of the leaflet update. This is fixed here.


karussell:
What do you mean with broken scale control? I cannot see something broken here.

ratrun:
Calculate a route such that distances are displayed in kilometers. When you now change the units, the left side buttom map scale GUI element still shows km, altough it should show miles. If you debug it in the browser you'll see something like ```scaleControl.remove``` is not a function.

karussell:
I cannot reproduce this, did you refresh the browser cache?

ratrun:
Strange. I initially detected the problem under nwjs, where it cannot be triggered from the browser cache as the page always gets completely loaded initally. Did you run `npm update` ? Eventually you did not and therefore you cannot reproduce it?

karussell:
Yes. Do you see this on https://graphhopper.com/maps too? The latest version is already deployed there. Hmmh, strange. @fbonzon or @ammagamma can you reproduce this problem?

fbonzon:
@ratrun You did not run `npm update`. The main.js file in your commit still reads "L.version = '0.7.7';"

ratrun:
Thanks @fbonzon for pointing me on my still not updated leaflet version. I did run npm update, but probably in a wrong folder.

PR #833, opened 2016-10-13
Fix background color of spinner icon

fbonzon:
So it matches the white background it is diplayed on, not grey.


karussell:
Thanks!

karussell:
Ah, that was something (small) I always wanted to be fixed :) !

PR #834, opened 2016-10-14
Fix minor CSS issues

fbonzon:
- Cursor when dragging location points should show arrows up and down, not only up, as we can reorder points both up or down
- Remove duplicate "color"
- No need to specify units for 0 something


karussell:
Thanks!

PR #842, opened 2016-10-18
add buildweb argument to graphhopper.sh which builds the graphhopper jar with the web module

edy:
I am using docker to run graphhopper. This pull request adds a new argument to the graphhopper.sh.
Now one can create (not only) docker images with a prebuilt graphhopper jar which also contains the web module.

### DIFF
--- a/graphhopper.sh
+++ b/graphhopper.sh
@@ -27,12 +27,13 @@ FILE=$2
 function printUsage {
  echo
  echo "./graphhopper.sh import|web <your-osm-file>"
- echo "./graphhopper.sh clean|build|help"
+ echo "./graphhopper.sh clean|build|buildweb|help"
  echo
  echo "  help        this message"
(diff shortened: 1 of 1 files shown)


karussell:
Thanks!

karussell:
Would you mind to send me an email? See our [contributing guide](https://github.com/graphhopper/graphhopper/blob/master/.github/CONTRIBUTING.md) regarding the CLA.

karussell:
BTW: regarding docker: maybe you can help us there :) ? I know there are several docker files around in the web but I would prefer one which is maintained in the repo.

edy:
well, i could write a Dockerfile for this project and send a you PR 😃

karussell:
This would be really nice! Currently we put those 'misc' files a bit hidden in `core/files` like the postgis folder. But as a separate step we can/should move the required files&folders into a new `/files`

Issue #549, opened 2015-10-19
Improve location lookup to avoid geocoding mismatch


karussell:
Geocoding is done separately and can have big mismatches where the geocode location snaps to e.g. a closer motorway and not to the intended nearby street of a specific house. Or where a building is directly under or near a bridge etc.

To prevent the users from these mismatches we need to include an address or street hint for the location lookup - we could name this `point_hint` or `address_hint` or similar. Then while lookup we calculate the Levenshtein distance for every point (if cached for one request it should be cheap) and we sort by this distance and filter out bad results. If only 'bad results' no filtering should happen. The question is how much a similarity should influence the snapped distance, i.e. what to do if the street Y looks a bit more similar than X but X is closer.

Furthermore two more improvements are possible
- the 'heading' parameter could be used as a `direction_hint` and improve the routing near a motorway (where one could exclude certain snapped locations based on this hint). 
- we can provide a subnetwork counter in the 'nearest' endpoint and have to include the address hint there too or return a list of matches with different streets etc.

Related to #729

An example can be many buildings on 'the corner' of a junction. E.g. 'Laufamholzstraße 154 Nürnberg' which currently geocodes to [49.46553,11.154669](https://graphhopper.com/maps/?point=49.46553,11.154669) and so it snaps to the closest 'Vorrastraße' but should snap to something like [49.465687,11.154631](https://graphhopper.com/maps/?point=49.465687,11.154631). Now if we specify Laufamholzstraße or even the full address in the routing we should be able to make this happen.

boldtrn:
I am currently looking into this issue. Something I was wondering about is how to design the `point_hint` (I think point hint might be better, since it belongs to the point?). Would we make the behavior of `point_hint` the same as point, meaning, we have a list of hints, exactly the same length as point, where the first hint is meant for the first point, and so on. Or would we only add a point_hint for certain points, e.G. those that were resolved from GeoCoding, whereas we might not want this feature for lat,lon points. Therefore, we might want to have `poin_hint_1` for `point_1`, but don't add `point_hint_0`, because we don't care about `point_0`, for whatever reason.

karussell:
I would keep it simple and assume that the number of 'point_hint' is
identical to the number of 'point' arguments. And if not throw an exception.

karussell:
Hmmh, interesting. Not sure how to handle this properly. We probably
need some heuristic before applying levenshtein e.g. in combination with
the longest common substring:

[...]

Issue #729, opened 2016-05-30
Improve edge filtering for LocationIndex


karussell:
The current method

``` java
QueryResult findClosest( double lat, double lon, /*->*/EdgeFilter edgeFilter )
```

is too simple for the following use cases:
1. If the Weighting returns infinity we need to exclude those edges too. E.g. just include weighting in the EdgeFilter or combine 'access determination' with weight calculation somehow.
2. Sorting edges according its weight should be possible, e.g. for map matching and via-point finding of the round trip calculation. But the problem here is: how do we combine the distance from the request point to the snapped point and the weight? And when can we stop increasing the radius? See one example: https://discuss.graphhopper.com/t/5000

Update: the BlockAreaWeighting makes problems. Before the creation of this we need the LocationIndex, but the LocationIndex needs the Weighting. Or to put it differently: a block-edge-Weighting should be possible for the LocationIndex, but then it operates on the base graph. Then, after it is done, the block-edge-Weighting needs to operate on the newly created QueryGraph to know about virtual edges and nodes. Potential fix: throw an error if query points lay inside the blocked area. Then BlockAreaWeighting does not need to know QueryGraph (only BaseGraph)

PR #846, opened 2016-10-22
Improve location lookup to avoid geocoding mismatch

boldtrn:
As mentioned in #549 I am currently looking into improving the location lookup. So what I did, was to extend the LocationIndex Interface for a new method, that matches not necessarily the closest but the best match, by comparing not only geographic distance but also the Levenstein distance.

Right now this is really simplistic and only a brief draft of how this could look like. As usual I like to discuss a draft first, before implementing too much detail :). However, the shown implementation matches the correct Street, for the example in #549, see the new IT.

So things that need to be done:
- [x] Implement the `point_hint` parameter - also decide about the design of this parameter.
- [x] Implement a more sophisticated 'Good Match', for example if the name is 5% of the perfect match. But then we would have to check if the distance from the best match from the current closestMatch is smaller than from the current. Not sure if this is worth the trouble?
- [x] Prepare strings before comparing them. For example: remove white spaces, maybe punctuation?
- [x] Right now I just copied the `findClosest` method of the LocationIndexTree, maybe we find a good way to unify the duplicated code.
- [x] Decide if we want to create a method in EdgeIteratorState, that only returns the street name, without the relation.
- [x] Decide if we need the implementation in the LocationIndex classes :)
- [x] Decide about the Levenstein Distance implementation, right now I use Apache Common Langs, but adding a dependency for "just" the Levenstein Distance might not be a good idea, so maybe we want to implement our own version. We can also copy&paste it, plus adding the tests and adding the dependency to the Notice file
- [x] Update API documentation
- [x] Write Unit Tests for new Edgefilters

Things for the future:
- [ ] Cache the calculated distances, for example in a Map. Right now the distances are calculated several time per name, since several edges might have the same name. Also we should compare the speed of caching vs. recalculation. In general Levenstein distance is not cheap to calculate, but for short strings this might be irrelevant?
- [ ] Create different versions of String similarity checking, e.G. Phonetic algorithms like some phonetic based stuff like soundex, metaphone etc? See [here](https://en.wikipedia.org/wiki/Phonetic_algorithm), [here](http://stackoverflow.com/q/2344320/194609) and [here](https://en.wikipedia.org/wiki/Approximate_string_matching)

Let me know what you think :)


karussell:
> so maybe we want to implement our own version.

We can also copy&paste it, plus adding the tests and adding the dependency to the Notice file

> Cache the calculated distances, for example in a Map. Right now the distances are calculated several time per name, since several edges might have the same name. 

Good idea. But for the first working prototype performance is not an issue if it is unchanged without the point hints

> Right now I just copied the findClosest method of the LocationIndexTree, maybe we find a good way to unify the duplicated code.

Yes, would be very good if there is only one method working with point hints and without. Should be possible with a changed EdgeFilter (see below)

> Prepare strings before comparing them. For example: remove white spaces, maybe punctuation?

The string preparation is probably the hardest part where we have to concentrate to increase the quality. Probably we could apply normal search index preparations like tokenizing the input and query and then count matching strings only (strings above a certain threshold). etc

> Decide if we want to create a method in EdgeIteratorState, that only returns the street name, without the relation

For now I would avoid this. We should be prepared for this as there are other unavoidable differences like 'Xy Straße' vs 'Xystraße' vs. 'Xy Strasse' etc. Maybe instead levenshtein a different method is better? Like some phonetic based stuff like soundex, metaphone etc? See [here](https://en.wikipedia.org/wiki/Phonetic_algorithm), [here](http://stackoverflow.com/q/2344320/194609) and [here](https://en.wikipedia.org/wiki/Approximate_string_matching)

Lucene could be used for ideas :)

> Decide if we need the implementation in the LocationIndex classes :)

I would not change the lookup Java API as we could implement a new edge filter which does the name comparison and caching. Then we also only need the improvement in the LocationIndexTree class only.

karussell:
To make things simpler we should probably assume that `points_hint` just contains the street name for now.

Later on we could try to find common 'street' wordings (street, way, road) for the different languages to detect the street inside the hint (and also in `edge.getName()`). Then we will need the locale and have different locales of the street names stored in the graph too. As this is complex we should avoid locale specific stuff, although this could improve quality a lot.

boldtrn:
> We can also copy&paste it, plus adding the tests and adding the dependency to the Notice file

Ok, sure, I can do that :).

> The string preparation is probably the hardest part where we have to concentrate to increase the quality. 

I thinks so, too. I think we can get to a okayish quality with simple things. If we see that we need more we can also take i18n and so on into account. Don't you think, that most of the time we will match against geocoding results? These should be exact.

> Like some phonetic based stuff like soundex, metaphone etc?

Let's make the String matching abstract, so we can interchange the algorithms and compare them easily :)?

> I would not change the lookup Java API as we could implement a new edge filter which does the name comparison and caching.

How would you do that? I think we cannot model more complex things in an EdgeFilter?

karussell:
> How would you do that? I think we cannot model more complex things in an EdgeFilter?

As a first prototype I would just reject all non-matching edges. To tune this we can then sort the results somehow (here we need to mix the distance with quality which might be a bit hacky) See #729

> Let's make the String matching abstract, so we can interchange the algorithms and compare them easily :)?

As you said: I would keep it simple and try several but then pick the best :)

> Don't you think, that most of the time we will match against geocoding results? These should be exact.

... there are one billion geocoders in the world ;) But yes, I think you are right and some combination of longest common substring (for preprocessing or in combination) and similarity comparison should do it I hope :)

boldtrn:
> As a first prototype I would just reject all non-matching edges.

Ok, this is implemented now. I also added the http parameter and did some refactoring. So right now this should be basically usable, but there are still some open points ;). I tested this in the browser and it works in the browser as well. Hopefully, I can fix the open points this week.

Use String Matching:
http://localhost:8989/?point=49.46553%2C11.154669&point=49.464749%2C11.150651&locale=de-DE&vehicle=car&weighting=fastest&elevation=false&use_miles=false&layer=Omniscale&point_hint=%22Laufamholzstra%C3%9Fe%22&point_hint=%22%22

Don't use it:
http://localhost:8989/?point=49.46553%2C11.154669&point=49.464749%2C11.150651&locale=de-DE&vehicle=car&weighting=fastest&elevation=false&use_miles=false&layer=Omniscale

Same result as:
https://graphhopper.com/maps/?point=49.46553%2C11.154669&point=49.464749%2C11.150651&locale=de-DE&vehicle=car&weighting=fastest&elevation=false&use_miles=false&layer=Omniscale

core/src/main/java/com/graphhopper/routing/template/ViaRoutingTemplate.java
```diff
@@ -65,6 +66,9 @@ public ViaRoutingTemplate(GHRequest ghRequest, GHResponse ghRsp, LocationIndex l
         queryResults = new ArrayList<>(points.size());
         for (int placeIndex = 0; placeIndex < points.size(); placeIndex++) {
             GHPoint point = points.get(placeIndex);
+            if(ghRequest.hasPointHints()){
+                edgeFilter = new LevensteinDistanceEdgeFilter(encoder, ghRequest.getPointHint(placeIndex));
```
karussell: I would use a wrapper of the old edgeFilter, otherwise we could accept stuff that edgeFilter would reject.

core/src/main/java/com/graphhopper/routing/util/LevensteinDistanceEdgeFilter.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.graphhopper.util.EdgeIteratorState;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.regex.Pattern;
+
+/**
+ * @author Robin Boldt
+ */
+public class LevensteinDistanceEdgeFilter extends NameSimilarityEdgeFilter{
```
karussell: Levenstein -> Levens**h**tein -> https://en.wikipedia.org/wiki/Levenshtein_distance

core/src/main/java/com/graphhopper/routing/util/LevensteinDistanceEdgeFilter.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        String name = iter.getName();
+        name = removeRelation(name);
+        name = prepareName(name);
+
+        int perfectDistance = Math.abs(soughtName.length()-name.length());
+        int levDistance = StringUtils.getLevenshteinDistance(soughtName, name);
+        // TODO: Maybe we should match 5% diff as good or something like that?
+        return levDistance <= perfectDistance;
```
karussell: I'm not sure I understand the criteria. Would you mind adding ~4 examples to show that it works correctly if it should accept and reject? (2 for both=>4 examples)

Also for now we should avoid yet another subclass and move this into one (merge this class and NameSimilarityEdgeFilter)

core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -0,0 +1,64 @@
...
+
+        name = nonWordCharacter.matcher(name).replaceAll("");
+        name = name.toLowerCase();
+
+        return name;
+    }
+
+    protected String removeRelation(String edgeName){
+        if(edgeName != null && edgeName.contains(", ")){
+            edgeName = edgeName.substring(0, edgeName.lastIndexOf(','));
```
karussell: Nice hack :) !

web/src/main/java/com/graphhopper/http/GraphHopperServlet.java
```diff
@@ -90,6 +90,13 @@ public void doGet(HttpServletRequest httpReq, HttpServletResponse httpRes) throw
                     throw new IllegalArgumentException("heading list in from format: " + e.getMessage());
                 }
 
+                List<String> pointHints = Collections.EMPTY_LIST;
+                pointHints = new ArrayList<String>(Arrays.asList(getParams(httpReq, "point_hint")));
+                if (pointHints.size() > 0 && pointHints.size() != requestPoints.size()) {
+                    // TODO: Not sure if this is nice?
+                    throw new IllegalArgumentException("If you pass point_hints, you need to pass a hint for every point, empty hints will be ignored");
```
karussell: This is good! We do similar things for heading I think. Minor: `point_hints` -> `point_hint`

karussell:
After addressing my reviewed points and removing the dependency to apache commons I would be also fine to merge this as this does not really effect other parts of the system - or what do you think?

core/src/main/java/com/graphhopper/routing/template/ViaRoutingTemplate.java
boldtrn: You are right. Internally I use the DefaultEdgeFilter and accept an Edge only if it's accepted by the DefaultEdgeFilter, but wrapping the existing EdgeFilter is better.

core/src/main/java/com/graphhopper/routing/util/LevensteinDistanceEdgeFilter.java
boldtrn: > I'm not sure I understand the criteria. Would you mind adding ~4 examples to show that it works correctly if it should accept and reject? (2 for both=>4 examples)

Sure, I want to add tests anyway :)

> Also for now we should avoid yet another subclass and move this into one (merge this class and NameSimilarityEdgeFilter)

I thought we wanted to compare Levenshtein to e.G. Soundex, therefore I wanted to make it extendable. Maybe we could also make this configureable, like routing algorithms. For example one user might prefer StringMatchingAlgo A the other prefer StringMatchingAlgo B. We could allow a user to pass an argument like `matching_algorithm=soundex`.

boldtrn:
> or what do you think?

Yes I think so to. This PR still needs some polishing, but I think we are getting close to mergeability :)

boldtrn:
I updated the PR.

Something I don't like right now is that if we cannot find a similar edge, we throw an exception. Since the parameter is a `hint` I would expect that it returns the closest match as fallback.

WDYT?

boldtrn:
I added two more tests, that show an issue of the current approach using the `perfectDistance`. When comparing a short to a long String, for example `Hauptstraße` to `a`. This is obviously a constructed case, but still, if the hint is long and the edgeName is short, there is a good chance of mismatches, especially, since we allow small typos.

karussell:
> Something I don't like right now is that if we cannot find a similar edge, we throw an exception. Since the parameter is a hint I would expect that it returns the closest match as fallback.

Indeed, sounds good.

boldtrn:
I updated my PR. Fallback to closest edge if no edge with similar name is found. I also changed it a bit. I calculate the number of allowed typos from the length of the point_hint. So with every 20 characters of the point_hint, 1 typo is allowed.

karussell:
> Fallback to closest edge if no edge with similar name is found.

Nice, this is simple :) ! How easy is it to store the best result even if the filter is given, so that we can avoid two requests?

boldtrn:
> How easy is it to store the best result even if the filter is given, so that we can avoid two requests?

Hm, I don't think that this is possible easily, since we don't do anything with an edge if the EdgeFilter is not accepting it, right? 

To support that we probably have to change the LocationIndex and EdgeFilter a bit. We would have to accept every edge that would be accepted by e.G. the DefaultEdgeFilter. In a second step we would check name similarity, this would be somewhat similar to my first commit in this PR, but without changing the API.

Since lookups don't take that long and it only impacts users that use the point_hint I would use the current solution. Changing the location index might have an impact on the lookup for non point_hint users (only an assumption).

karussell:
Indeed, you are right this would make unnecessary complex for now.

core/src/main/java/com/graphhopper/util/StringUtils.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * This class is a partial Copy of the org.apache.commons.lang3.StringUtils
+ * that can be found here: https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/StringUtils.java
+ *
+ * The library can be found here: https://commons.apache.org/proper/commons-lang/
+ *
+ * @author Robin Boldt
+ */
+public class StringUtils {
```
karussell: Would you mind putting the apache commons files into a separate package (`com.graphhopper.apache(.)commons`?) and use the original license header?

core/src/test/java/com/graphhopper/routing/util/NameSimilarityEdgeFilterTest.java
```diff
@@ -0,0 +1,227 @@
...
+        // Distance - PerfectDistance = 1
+        edgeFilter = new NameSimilarityEdgeFilter(new DefaultEdgeFilter(new CarFlagEncoder()), "az");
+        assertTrue(edgeFilter.accept(edge));
+
+        // Distance - PerfectDistance = 2
+        edgeFilter = new NameSimilarityEdgeFilter(new DefaultEdgeFilter(new CarFlagEncoder()), "xy");
+        assertFalse(edgeFilter.accept(edge));
+    }
+
+    static class TestEdgeIterator implements EdgeIteratorState {
```
karussell: Can you use GHUtility.createMockedEdgeIteratorState here?

Oh it looks like there is another helper ShortFastestWeightingTest.createEdge ... will try to fix this.

web/src/main/java/com/graphhopper/http/GraphHopperServlet.java
```diff
@@ -90,6 +90,12 @@ public void doGet(HttpServletRequest httpReq, HttpServletResponse httpRes) throw
                     throw new IllegalArgumentException("heading list in from format: " + e.getMessage());
                 }
 
+                List<String> pointHints = Collections.EMPTY_LIST;
+                pointHints = new ArrayList<String>(Arrays.asList(getParams(httpReq, "point_hint")));
+                if (pointHints.size() > 0 && pointHints.size() != requestPoints.size()) {
+                    throw new IllegalArgumentException("If you pass point_hints, you need to pass a hint for every point, empty hints will be ignored");
```
karussell: point_hints -> point_hint

core/src/test/java/com/graphhopper/routing/util/NameSimilarityEdgeFilterTest.java
boldtrn: Good point. I created something like GHUtility.createMockedEdgeIteratorState because I needed different mock methods here.

core/src/main/java/com/graphhopper/util/StringUtils.java
boldtrn: No problem

boldtrn:
Fixed review comments. Should be ready for merging?

karussell:
Thanks!

karussell:
Nice: tested this a bit and this works already great for all my scenarios. Will make sure I'll deploy this feature fast :)

Issue #831, opened 2016-10-13
Improve Weighting class


karussell:
Copied from PR #807 

I'm not satisfied here. We would need to break API again if we have to calculate time, distance, transfers, weight in a routing algo like Dijkstra. Currently we just calculate the weight.

Now I'm thinking about a new class `WeightingResult` with time, distance, weight, ... and just one method:

```java
class Weighting {
  void calculate(WeightingResult res, EdgeIteratorState edge, boolean reverse, int prevEdge);
  // or for easier backward compatibility returning the first parameter:
 // WeightingResult calculate(WeightingResult, EdgeIteratorState, reverse, prevEdge);
}
```

Then we could easily fill the time or weight etc in this method. It would be even more customizable as we could subclass WeightingResult and add parameters. We could even replace `TurnWeighting` when we store the turnCosts in the WeightingResult (or a subclass).

Another advantage is that we currently would calculate the weighting and then the 'millis' but this work of calculating the time is already (often) done in calculating the weighting and so it would be duplicative which we can avoid via this new method.

One problem remains: how we incorporate the time stuff required for the `pt` branch.

PR #848, opened 2016-10-25
Merge master into pt

karussell:
This brings the PT branch to master. All tests pass but please review before, especially the clashing classes like Dijkstra and Path

> Aaah - this is VERY confusing that all the **old** comments regarding some containing commits now show up in this pull request!!

Okay, squashed all changes into a single commit for now


core/src/main/java/com/graphhopper/routing/Path.java
```diff
@@ -214,7 +217,7 @@ public void next(EdgeIteratorState edgeBase, int index) {
                         // TODO This should already be in the SPT, we shouldn't need to calculate it again here.
                         time += ((TimeDependentWeighting) weighting).calcTravelTimeSeconds(edgeBase, time / 1000.0) * 1000.0;
                     } else {
-                        time += calcMillis(dist, edgeBase.getFlags(), false);
+                        time += weighting.calcMillis(edgeBase, false, EdgeIterator.NO_EDGE);
```
karussell: Here we need previousEdge too

karussell:
(GitHub going nuts now when I edit a comment it shows up a previous version ... anyway)

We need to work out a solution for #831 with public transport in mind.

karussell:
I've fixed the issues and also merged latest changes from pt branch back to here so it should be mergable without conflicts again. Would you mind to review @michaz ?

michaz:
Thanks! Quite possible that some of the changes in the core on the pt branch can be reverted and left in a new branch (time-dependent) soon.

karussell:
Probably we can also consider to merge pt into master as soon as it is at least 'alpha' status (e.g. slightly tuned web API&UI and a bit more unit test coverage) because 0.9 has still several months to stabilize & evolve. And maintaining two not so similar branches requires a bit work.

PR #849, opened 2016-10-26
Add Dockerfile

edy:
As promised, I created a simple Dockerfile for GraphHopper.

Build an image:
`docker build -t graphhopper:master .`

Run tests:
`docker run -it graphhopper:master ./maven/bin/mvn clean test verify`

Run graphhopper:
`docker run -it -p 8989:8989 -v /home/username/graphhopper-maps/:/data graphhopper:master ./graphhopper.sh web /data/europe-latest.osm.pbf`

You could add this Project to docker hub and let them build the images for you, which everyone can use.

Please let me know if there is something i should change.


stefpb:
+1

karussell:
Thanks a lot! As we have currently already many files in the project root, is it possible to move the 2 new docker files to e.g. core/files or would it make things ugly/complicated?

karussell:
And I assume switching to a stable branch like 0.8 is possible via

`docker build -t graphhopper:0.8 .`

Or what does graphhopper:master mean?

edy:
If I move the Dockerfile to `core/files`, then it would be unpossible to `COPY` the repository data into the docker image.

> The <src> path must be inside the context of the build; you cannot COPY ../something /something, because the first step of a docker build is to send the context directory (and subdirectories) to the docker daemon. -- https://docs.docker.com/engine/reference/builder/#/copy

graphhopper:master was just an example for a tag name of the docker image. One would build an image of the 0.8 branch by first checking it out and then use `docker build -t graphhopper:0.8 .`
https://docs.docker.com/engine/reference/commandline/build/#/tag-image--t

edy:
There could be a Dockerfile in `core/files` which downloads the released zip files, like i did in https://github.com/edy/graphhopper-docker

karussell:
Hmmh, difficult. I thought I could just fetch the docker file and work with it (install it etc) without being attached to any repository or data?

For normal production use I would prefer fetching the stable zip. But sometimes a tweak or bug fix is necessary. Maybe like in your 2nd suggestion but with `git checkout 0.8`?

> graphhopper:master was just an example for a tag name of the docker image

Sorry, have not yet any idea of the concepts for docker :) which makes it a bit hard to decide what to do

edy:
> Maybe like in your 2nd suggestion but with git checkout 0.8

I'd go with the 0th (Dockerfile inside the project root).

Pros:
- you can `git checkout` your branch/tag/commit and build an image for this version
- Docker Hub has automated builds which are triggered on pushes (commits or tags). Every pushed tag will create a new docker image tagged as eg. `graphhopper:0.8.2`

Cons:
- two more files inside the project root 😃 

My other suggestions require more work because the version number must be changed manually inside the Dockerfile on every new release.

karussell:
Ok, thanks for the explanation - convinced & merged :)

PR #850, opened 2016-10-28
Android: use VTM OpenGL vector map library

devemux86:
We released [VTM 0.6.0](https://groups.google.com/forum/#!topic/mapsforge-dev/5w9Sl7Kljzk) OpenGL vector map library today. :slightly_smiling_face: 

Here is an update for Android sample, i.e. GraphHopper + VTM working together:
- Use VTM OpenGL renderer with offline Mapsforge vector maps
- Show its overlay API use for the routing markers and calculated route line


karussell:
Really nice update - thanks a lot @devemux86 !

devemux86:
Probably you could update now the Android [screenshot](https://github.com/graphhopper/graphhopper/tree/master/android) with something more OpenGL in 2.5D!

karussell:
Do you know how the pom.xml snippet looks like with those native libs?

karussell:
> Probably you could update now the Android screenshot with something more OpenGL in 2.5D!

Yeah, tried this but I still rely on (old) Maven ;(

devemux86:
>Do you know how the pom.xml snippet looks like with those native libs?

Can you try declaring the natives with the `<classifier>` tag, e.g.:
```xml
<classifier>natives-armeabi</classifier>
```

karussell:
Thanks, that was it! Never heard about classifier, always thought the system path is the way to go, nice!

This is just wow :) ! Not only the rotation & tilt is nice but also the rendering speed!

devemux86:
And without an active file tile cache!
We can add it in the future for an even better experience. :slightly_smiling_face:

karussell:
Hmmh, tried a landscape screenshot but the text does not rotate. So I included for now a smaller screenshot but it does not look pretty: https://github.com/graphhopper/graphhopper/blob/master/docs/android/index.md

(I removed the readme in the Android subproject - good idea or should we at least point to the android docs fromt here?)

devemux86:
>tried a landscape screenshot but the text does not rotate

Probably because app is [locked](https://github.com/graphhopper/graphhopper/blob/master/android/app/src/main/AndroidManifest.xml#L25) for portrait?

Screenshot indeed should be a large one, at least HD.
Android subproject could use a docs link and the same screenshot, it's good for demonstration there too. :slightly_smiling_face:

karussell:
Ok, I'll try.

> Android subproject could use a docs link and the same screenshot, good for demonstration there too. :slightly_smiling_face: 

Problem of this dup info: I completely forgot about it ... I'll re-add a link to the docs + the screen.

devemux86:
I'd replace that [line](https://github.com/graphhopper/graphhopper/blob/master/android/app/src/main/AndroidManifest.xml#L25) with:
```xml
android:configChanges="keyboardHidden|orientation|screenSize"
```
to avoid application's restart with device orientation changes.

karussell:
Nice, that improves it even more - thanks again!

PR #852, opened 2016-10-28
Solve mmap for jdk9

karussell:
I'm trying to get the jdk9 build green via ignoring some tests regarding memory mapping

First shot ... maybe it already works and/or @mprins can help :) ?

Later on we should make it working or mark those tests with annotations to ignore them as discussed [here](https://github.com/graphhopper/graphhopper/issues/701#issuecomment-251655568)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,13 +41,14 @@ before_install:
          sudo update-java-alternatives -s java-9-oracle;
      fi
    - if [ "$TRAVIS_JDK_VERSION" == oraclejdk9 ]; then
+         export PROFILE="-P jdk9"
          sudo rm /etc/mavenrc;
      fi
 
 install: true
 
 script:
-  - mvn clean test verify -B
+  - mvn $PROFILE clean test verify -B
   #- ./core/files/travis-build.sh
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,15 +41,13 @@ before_install:
          sudo update-java-alternatives -s java-9-oracle;
      fi
    - if [ "$TRAVIS_JDK_VERSION" == oraclejdk9 ]; then
-         export PROFILE="-P jdk9";
          sudo rm /etc/mavenrc;
      fi
 
 install: true
 
 script:
-  - mvn $PROFILE clean test verify -B
-  #- ./core/files/travis-build.sh
+  - mvn clean test verify -B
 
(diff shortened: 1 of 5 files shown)


mprins:
I think there's no need to use the P switch on the Mavens command line, Mavens should activate the profile because it knows about the JVM.
Not sure about having both activebydefault and jdk selectors in the activation.

.travis.yml
```diff
@@ -47,8 +47,7 @@ before_install:
 install: true
 
 script:
-  - mvn clean test verify -B
-  #- ./core/files/travis-build.sh
+  - mvn -DfailIfNoTests=false clean test verify -B
```
karussell: The argument `-DfailIfNoTests=false` is necessary because the lgpl tools module has no tests

karussell:
Now this runs at least locally, several issues where there ...

 * the jdk9 has 9-ea as version number
 * there are `--add-exports` and `-XaddExports` options which is a bit confusing. Look like locally was a different than for travis, will try now `--add-exports`
 * found a possibility to clean Android buffers too as the name there is just slightly different maybe @devemux86 can check if this runs or how to tweak this so that we could close a map and open a different one (without the memory problem) on Android. See [this discussion](https://discuss.graphhopper.com/t/spatial-limitation-in-android-offline-on-device/298)
 * One remaining real issue for the JDK9 build: GraphHopperStorageWithTurnCostsTest>GraphHopperStorageTest.testEnsureSize:184 » OutOfMemory

karussell:
Strange, although we get `java version "9-ea"` it is still saying

`Unrecognized option: --add-exports:java.base/jdk.internal.ref=ALL-UNNAMED`

And the same happens for the other option `-XaddExports`

karussell:
Argh, it was the colon! Tunaki [from stackoverflow](http://stackoverflow.com/questions/40313090/java-9-options-add-exports-vs-xaddexports-not-recognized#comment67883333_40313090) solved this :) 

So now this strange error occurs:
```
testEnsureSize(com.graphhopper.storage.GraphHopperStorageWithTurnCostsTest)  Time elapsed: 0.135 sec  <<< ERROR!
java.lang.OutOfMemoryError: Java heap space - problem when allocating new memory. Old capacity: 1048576, new bytes:49283088, segmentSizeIntsPower:18, new segments:48, existing:1
```

PR #855, opened 2016-11-01
Add bahn.de/PTE-based test suite for RNV

highsource:
Here's a prototypical implementation of a test suite based on bahn.de data via PTE.

Trip queries are read from the CSV file (`rnv-trips-least-duration-01.csv`) and then executed in the GraphHopper. No assertions on results, just printing out. I've also reduced test cases to trips with just one leg because RNV does not have correct transfers.

As I ran the test it showed that GH PT sometimes gives good results, but in most cases it gives arrival time which is later than bahn.de. I hope these test cases will help in the analysis.


michaz:
The problem here could be similar to the reason transfers don't work on the RNV dataset.

In RNV, every pole is one stop. A stop (or stop area) with poles in either direction is two stops in the RNV dataset. We currently only, strictly use the closest stop from the start and the destination. 

That's probably why we often don't find a route.

In VBB, this is much easier, because all allowed transfers are explicit edges.

This test will only start giving nice results when we implement a transfer heuristic.

michaz:
(The VBB dataset is currently too much for Travis, otherwise I'd suggest we do a similar test suite for VBB right away...)

highsource:
No, I'm explicitly checking only those routes which only had 1 leg with PTE.
I understand that it might be a bit early for this test. But I think it's still good to see the difference to bahn.de - as early as possible.

michaz:
Exactly. :)

michaz:
Anecdote: Testing the VBB dataset against bahn.de is super unfair, because they have more pessimistic transfer times in their GTFS than in Hafas, apparently. :-(

highsource:
bahn.de re-processes data delivered by transport companies on import. So it's more for orientation.

I'll try to implement OpenTripPlanner-based test data production as well. That should give fair results.

michaz:
traveltimes

Wow. Will start analysing this tomorrow.

highsource:
Below y=x (to the right/bottom) means GraphHopper reports shorter travel times compared to bahn.de?

michaz:
> Below y=x (to the right/bottom) means GraphHopper reports shorter travel times compared to bahn.de?

Yes. Those seem to be trips where the web site says they should take really long. Are those maybe stops where only one bus goes every few hours, and GraphHopper just walks?
I haven't had time to look at this in any way yet, just wanted to show the picture. :)

highsource:
Should I do a VBB suite as well? Does it help?

michaz:
Yes, please! It helps a lot. RNV will then be our unrestricted-walking/multimodal/experimental-behavior scenario, and VBB will be our strictly-do-what's-in-gtfs scenario.

michaz:
bahn.de seems to produce unexpectedly long solutions sometimes:
screen shot 2016-11-18 at 12 41 19

@highsource I wonder, would it be difficult to re-generate the .csv file using the `VrnProvider` instead of bahn.de? If it's easy for you, please feel free to just commit.

highsource:
This should not be too complicated but will take some time.

BTW I get 1h23m for this query with VRN.

highsource:
Hm, this is a very good approach to identify weaknesses in bahn.de. Kind of cross-comparison for the same queries among providers. This might be pretty useful.

michaz:
(Wouldn't need to be exactly the same instances, of course. Can be newly randomized.)

michaz:
Next version:
traveltimes
In the previous picture, everybody was walking, because the schedule didn't fit the data -- I'm using the new schedule now, which covers all of November -- so we were basically just plotting bahn.de travel time against distance.

Now, something like a diagonal is emerging. :-)

michaz:
The bulge over the diagonal starts around an hour or two. Maybe that's where bahn.de is starting to use long-distance trains which we don't know about.

highsource:
I think long-distance is disabled in the query.

michaz:
> I think long-distance is disabled in the query.

Hmm, okay, nice try though. ,-)

michaz:
Ah, it's the S-Bahn.

PR #868, opened 2016-11-11
Add icons in context menu

fbonzon:
Width parameter can be omitted. By default width adapts to the longest menu item. Prevents line break with UI in some languages with longer strings, e.g. French.

Before:
google chrome-screenshot-001

After:
google chrome-screenshot-002

Demo: https://gh.dev.bikewm.com/


karussell:
Nice & Thanks!

PR #870, opened 2016-11-12
Improve access restriction handling for bicycle routing

ratrun:
Give highway=cycleway higher preference compared to access restrictions. The change was triggered by the discussion in issue#863 about allowing a combination of highway=cycleway with "vehicle=no". Also BikeFlagEncoder access restriction testcases are improved for similar cases.


core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
```diff
@@ -248,8 +248,19 @@ public long acceptWay(ReaderWay way) {
...
+                return acceptBit;
+            if (!isSacScaleAllowed(sacScale))
+                return 0;
+        }
+
         // use the way if it is tagged for bikes
-        if (way.hasTag("bicycle", intendedValues) || way.hasTag("bicycle", "dismount"))
+        if (way.hasTag("bicycle", intendedValues) || 
+            way.hasTag("bicycle", "dismount") || 
+           (way.hasTag("highway", "cycleway")))
```
karussell: Would you mind to remove the double brackets `(way....)`? Also for the above statement

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
```diff
@@ -248,8 +248,19 @@ public long acceptWay(ReaderWay way) {
         if (!highwaySpeeds.containsKey(highwayValue))
             return 0;
 
+        String sacScale = way.getTag("sac_scale");
```
karussell: Nice, I can see why you moved this before the improved 'has bicylce tag' check. Would you mind to add a test for sac_scale that would fail if put afterwards?

karussell:
Thanks!

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
ratrun: There is already a test for this, which failed. I would not have recognized this otherwise!

karussell: Ah, great :)

PR #872, opened 2016-11-14
Added Max Point Distance for Non-Ch Requests

### LINKED ISSUE #734: Make it possible to restrict routing requests above a certain limit
Currently we can protect graph exploration to a certain limit but if e.g. certain points are 'obviously' out of scope we could avoid the calculation at all and introduce a `routing.max_distance` or something which is simply checked on the server side.

boldtrn:
This PR intents to fix #734. It adds the option to define a maximum beeline between two consecutive points. The distance will be only checked for non-ch requests.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -40,8 +40,7 @@ import com.graphhopper.storage.index.QueryResult;
 import com.graphhopper.util.*;
 import com.graphhopper.util.Parameters.CH;
 import com.graphhopper.util.Parameters.Routing;
-import com.graphhopper.util.exceptions.DetailedIllegalArgumentException;
-import com.graphhopper.util.exceptions.MaxPointDistanceExceededException;
+import com.graphhopper.util.exceptions.PointDistanceExceededException;
 import com.graphhopper.util.exceptions.PointOutOfBoundsException;
 import com.graphhopper.util.shapes.BBox;
 import com.graphhopper.util.shapes.GHPoint;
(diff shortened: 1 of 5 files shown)


karussell:
Thanks!

I'm not sure about the naming. Do you think we need at some point parameters for CH, non-CH and different boost techs like [landmarks](https://github.com/graphhopper/graphhopper/pull/780)? If yes, then probably just max_waypoint_distance is enough for the non-ch case without the 'non-ch' addition? And for all ch and landmark related configs we just use e.g. `lm.max_waypoint_distance`?

And ugly that there is `way_point_max_distance` with a complete different meaning. But we could find a better name for this and make this accessible in the docs, plus deprecate the old naming.

boldtrn:
> If yes, then probably just max_waypoint_distance is enough for the non-ch case without the 'non-ch' addition?

Not sure, I chose the naming because that is actually the meaning of the parameter, whereas I would assume that max_waypoint_distance is enforced independent from CH, which is not true right now.

> If yes, then probably just max_waypoint_distance is enough for the non-ch case without the 'non-ch' addition?

We just got rid of the `.` in parameters, so I feel like that I would chose something different, but yes denoting or creating some sort of category would be great. Maybe with `__`, we should think about this; maybe in a different issue or in the forum?

> And ugly that there is way_point_max_distance with a complete different meaning

You are right, I have not considered that. This is ugly indeed. Any naming suggestions here? I somehow feel that the current name describes the meaning pretty well.

karussell:
> whereas I would assume that max_waypoint_distance is enforced independent from CH, which is not true right now.

Yes, you are right. Can you still slightly rename it to `non_ch_max_waypoint_distance`?

> We just got rid of the . in parameters

Not if they can easily fall into a category like `ch.disable` or are for config.properties only (`routing...`)

BTW: maybe we should introduce a category 'non_ch' and name it `non_ch.max_waypoint_distance`?

> I somehow feel that the current name describes the meaning pretty well.

Maybe `perpendicular_max_waypoint_filter_distance` or some 'filter/smoothing/cleaning' addition?

karussell:
One minor addition could be useful: if CH disabling is possible or CH is disabled and this parameter is still MAX we should probably print out a warning? Or already set a high limit per default?

boldtrn:
> BTW: maybe we should introduce a category 'non_ch' and name it non_ch.max_waypoint_distance?

Thanks, I did that :).

> Maybe perpendicular_max_waypoint_filter_distance or some 'smoothing/cleaning' replacing the filter?

Not sure about the naming, so I thought I won't touch it for now, however I can change it if you want?

> if CH disabling is possible or CH is disabled and this parameter is still MAX we should probably print out a warning? Or already set a high limit per default?

Yes sounds good, but I would only add it to the config file, not to the code itself. Therefore, we don't break existing code.

Let me know what you think or if I should improve anything with the current solution?

karussell:
> Not sure about the naming, so I thought I won't touch it for now

Yes. I've put a note in #437

> Therefore, we don't break existing code

Nice idea, I'll change the config-example.properties

karussell:
> Nice idea, I'll change the config-example.properties

Ups - you've already done this - thanks!

PR #874, opened 2016-11-16
Refactored OSM Shapefile importer

PGWelch:
Fixes / changes completed

1. Highway tag – we’re now using this (not used before, previously the code was using a hardcoded speed or max speed if non-zero).

1. Max speed tag now passed properly to the encoders. Added a filter for value “0” as in Geofabrik shapefiles 0 means no tag, but 0 will probably just mean 0 speed in the Graphhopper flag encoders…

1. One way tags now used properly. Previously the code was only checking for “0” and treating this as boolean. You also have to check for against the direction of digitisation (e.g. value “-1” if its one-way against the directory of digitisation). Also Geofabrik is using an odd (unfortunately non-standard) convention for their shapefile oneway field; I now read this convention but turn into the standard convention in the ReaderWay I pass to the flag encoders. This could break in the future if Geofabrik clean this up (the code will throw an exception with a useful message if it breaks).

1. Encoding manager now used properly. I now fill a ReaderWay with the tags and pass to the EncodingManager (the EncodingManager was not used previously), so we should get proper behaviour over multiple speed profile types (car, bike etc). Previously there was some hardcoding to “car”.

1. Fixed partially connected graph bug. With a shapefile from Geofabrik we don’t have node objects, only the geometry of the road (they don’t have a separate node export shapefile). Worse still, the road geometry generally isn’t split at junctions / intersections. Previously the code was assuming that any point at the start or the end of a road geometry line needed to be a node, and was splitting edges wherever one of their geometry points corresponded to the start or end of another edge. This fails badly though, because many times you get two edges crossing each other (e.g. at crossroads) where neither edges are broken at the crossroads – resulting in edges not being split at the crossroads, the graph being only half-connected and hence some pretty odd routes. Instead I’ve changed the code to create a node – and therefore split edges – based on any identical coordinate shared in two or more edges. This is (I believe) the best we can do with Geofabrik shapefile data but may fail in some odd / rare circumstances (e.g. as no z-level considered). This also means each coordinate (pillar or tower) gets stored in a hashmap during the build, which is not memory-friendly for very large files. However for a HERE importer I believe the data would be setup differently (so we’d be able to do this differently).

1. Edge length was being rounded down to int metres, I’ve kept it in double instead.

1. Added the estimated_distance and estimated_centre artificial tags

1. Ensured code handles shapefile multilinestring or linestring geometry (both are possible, previously multilinestring was assumed).

1. Unit tests - two tests for travel times, comparing between building using pbf (the existing reader project) and the shapefile project. One test is between some predefined points with known travel times, the other chooses many pairs of random points (200 currently). For the random pairs of points, we’re getting a mean agreement in travel times between the pbf and shp of around 99%, which is pretty good as the shapefile isn’t taking into account lots of other tags : ) Obviously in some cases we get less agreement presumably due to tags we’re not using…

1.  Unit tests - we now have a (very basic) geometry test

1.  Unit tests - we now have a test for a single one way example

1. Small amounts of code tidying (no code conventions applied though).

 Improves #616


karussell:
Thanks, will review!

reader-shp/src/main/java/com/graphhopper/reader/shp/OSMShapeFileReader.java
```diff
@@ -0,0 +1,329 @@
...
+							} else {
+								pillars.add(point);
+							}
+						}
+					}
+				}
+
+			}
+		} catch (Exception e) {
+			LOGGER.error("Exception while processing for roads: ", e);
```
karussell: Here we have to throw an exception, e.g. `throw new RuntimeException("some info", e)`

reader-shp/src/main/java/com/graphhopper/reader/shp/ShapeFileReader.java
```diff
@@ -0,0 +1,94 @@
...
+import com.graphhopper.storage.Graph;
+import com.graphhopper.storage.GraphHopperStorage;
+import com.graphhopper.storage.GraphStorage;
+import com.graphhopper.storage.NodeAccess;
+import com.vividsolutions.jts.geom.Coordinate;
+
+/**
+ * ShapeFileReader takes care of reading a shape file and writing it to a road network graph
+ *
+ * @author Vikas Veshishth
```
karussell: Feel free to add yourself here too :)

reader-shp/src/main/java/com/graphhopper/reader/shp/GraphhopperShp.java
```diff
@@ -0,0 +1,26 @@
...
+	@Override
+	protected DataReader createReader(GraphHopperStorage ghStorage) {
+		OSMShapeFileReader reader = new OSMShapeFileReader(ghStorage);
+		for (EdgeAddedListener l : edgeAddedListeners) {
+			reader.addListener(l);
+		}
+		return initDataReader(reader);
+	}
+
+	public void addListener(EdgeAddedListener l) {
```
karussell: Did I understood the purpose of the listeners: adding custom attributes to the graph? Why are multiple listeners necessary?

reader-shp/src/main/java/com/graphhopper/reader/shp/Utils.java
```diff
@@ -0,0 +1,17 @@
+package com.graphhopper.reader.shp;
+
+import com.graphhopper.util.PointList;
+import com.vividsolutions.jts.geom.Coordinate;
+import com.vividsolutions.jts.geom.GeometryFactory;
+
+public class Utils {
+	public static String toWKT(PointList list) {
```
karussell: Is it necessary to create a new factory for every call? Or is performance overhead neglectable and problematic stuff to ugly like non-thread safety?

reader-shp/src/test/java/com/graphhopper/reader/shp/ShapeFileReaderTest.java
```diff
@@ -0,0 +1,289 @@
...
+	/**
+	 * Build the graphs once only for the various tests
+	 */
+	@BeforeClass
+	public static void setupBeforeClass() {
+		try {
+			new File(tempOutputDirFromShp).mkdirs();
+			new File(tempOutputDirFromPbf).mkdirs();
+			
+			System.out.println("Building from shapefile");
```
karussell: No logging necessary in the tests IMO

karussell:
When trying ['sachsen'](http://download.geofabrik.de/europe/germany/sachsen-latest-free.shp.zip) I get:

```java
Caused by: java.lang.NullPointerException
	at com.graphhopper.reader.shp.ShapeFileReader.getFeatureIerator(ShapeFileReader.java:56)
	at com.graphhopper.reader.shp.OSMShapeFileReader.processJunctions(OSMShapeFileReader.java:82)
```

Will look into it tomorrow.

PGWelch:
Re: sachsen. I've just tried this file and its fine. Route built below using this code.

	GraphhopperSHP gh = new GraphhopperSHP();
	gh.setWayPointMaxDistance(0);
	gh.setStoreOnFlush(false)
			.setDataReaderFile(
					"C:\\Users\\Phil\\Dropbox\\Business\\Companies\\Graphhopper\\Shapefile importer\\Sachsen-dataset\\gis.osm_roads_free_1.shp")
			.setGraphHopperLocation(new File("c:\\temp").getAbsolutePath())
			.setEncodingManager(new EncodingManager(new CarFlagEncoder())).setCHEnabled(false).importOrLoad();
	
	GHPoint dresden = new GHPoint(51.0504 ,13.7373);
	GHPoint leipzig = new GHPoint(51.3397,12.3731);
			
	GHRequest request = new GHRequest(dresden,leipzig).setVehicle("car");
	GHResponse response = gh.route(request);
	if(!response.hasErrors()){
		System.out.println(Utils.toWKT(response.getBest().getPoints()));
		
	}

The exception suggests the shapefile can't be loaded? You've unzipped and referenced the roads file yes?

dresden-to-sachsen

reader-shp/src/main/java/com/graphhopper/reader/shp/ShapeFileReader.java
```diff
@@ -0,0 +1,94 @@
...
+	public ShapeFileReader(GraphHopperStorage ghStorage) {
+		this.graphStorage = ghStorage;
+		this.graph = ghStorage;
+		this.nodeAccess = graph.getNodeAccess();
+	}
+
+	@Override
+	public void readGraph() throws IOException {
+		// TODO why 1000?
+		graphStorage.create(1000);
```
PGWelch: OK, there's no parameterless overload for create so I'll leave this line as is.

karussell:
Works now surprisingly well - thanks a bunch!

> The exception suggests the shapefile can't be loaded? You've unzipped and referenced the roads file yes?

I specified the folder not the shp file ... was too late yesterday ;) ... we should still add a NPE check (?)

reader-shp/src/test/java/com/graphhopper/reader/shp/ShapeFileReaderTest.java
PGWelch: Prefer to keep the printlns if OK. There's some console output from the tests I like to check : )

reader-shp/src/main/java/com/graphhopper/reader/shp/ShapeFileReader.java
PGWelch: Changed in the next commit

reader-shp/src/main/java/com/graphhopper/reader/shp/OSMShapeFileReader.java
PGWelch: Changed in the next commit. I've changed the exceptions to be unchecked (as I hate checked exceptions!) and don't catch them anymore (other than wrapping them in a runtime exception to uncheck them if needed). The original exception is likely to be more informative than a rethrown one anyway.

reader-shp/src/main/java/com/graphhopper/reader/shp/OSMShapeFileReader.java
```diff
@@ -0,0 +1,329 @@
...
+import com.vividsolutions.jts.geom.MultiLineString;
+
+import gnu.trove.impl.Constants;
+import gnu.trove.map.hash.TObjectIntHashMap;
+
+/**
+ * OSMShapeFileReader for files present at : http://download.geofabrik.de/ It extracts the data as per the structure of
+ * shape files
+ *
+ * @author Vikas Veshishth, Philip Welch
```
PGWelch: Changed in the next commit.

reader-shp/src/test/java/com/graphhopper/reader/shp/ShapeFileReaderTest.java
karussell: The problem is that if every test does this we would be overwhelmed by log statements (we have >1K tests ;)). And every test should just print ok to command line if succeeded. For development purposes you can use this (and throw away afterwards) although I always recommend to formulate the things you check on command line with your eyes into an assert statement. Do you know what I mean :) ?

reader-shp/src/main/java/com/graphhopper/reader/shp/GraphhopperShp.java
PGWelch: The listeners don't modify anything, they're only used to help me output debugging information when checking over the building of the graph. They're not actually called from the commited code (just an external main I was using to run a debug from). They are useful though, so I'd prefer keep them in temporarily.

reader-shp/src/main/java/com/graphhopper/reader/shp/GraphhopperShp.java
```diff
@@ -0,0 +1,26 @@
+package com.graphhopper.reader.shp;
+
+import java.util.HashSet;
+
+import com.graphhopper.GraphHopper;
+import com.graphhopper.reader.DataReader;
+import com.graphhopper.reader.shp.OSMShapeFileReader.EdgeAddedListener;
+import com.graphhopper.storage.GraphHopperStorage;
+
+public class GraphhopperShp extends GraphHopper {
```
PGWelch: Changed in the next commit. Probably not gramatically correct though! OSM is an acronymn, SHP isn't. Licence file added.

reader-shp/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+                    </archive>
+                        	                    
+                    <!-- for standalone usage -->
+                    <descriptorRefs>
+                        <descriptorRef>jar-with-dependencies</descriptorRef>
+                    </descriptorRefs>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
```
PGWelch: Changed in the next commit.

reader-shp/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.graphhopper</groupId>
+    <artifactId>graphhopper-reader-shp</artifactId>
+    <version>0.9-SNAPSHOT</version>
+    <packaging>jar</packaging>
+    <name>GraphHopper Reader for Shapefile Data</name>
+	
+    <properties>      
+        <geotools.version>13.2</geotools.version>
```
PGWelch: Updated to 15.2 in next commit.

reader-shp/.gitignore
```diff
@@ -0,0 +1,2 @@
+#Don't ignore my Malta test data used in the automated test!!
+!malta-latest.osm.pbf
```
PGWelch: Changed in the next commit.

karussell:
It looks like geotools is not JDK9 ready, which is a bit ugly:
```
java.lang.IllegalArgumentException: org.geotools.data.DataStoreFactorySpi is not an ImageIO SPI class
	at com.graphhopper.reader.shp.ShapeFileReaderTest.initHopper(ShapeFileReaderTest.java:118)
	at com.graphhopper.reader.shp.ShapeFileReaderTest.setupBeforeClass(ShapeFileReaderTest.java:131)
```

Do you have an idea if we somehow can workaround this without changing geotools? This was the only thing I found in the web but maybe this need a change to geotools directly: https://github.com/levigo/jbig2-imageio/pull/18

karussell:
Ah, it is already known to them: https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-5289?filter=allopenissues

PGWelch:
hmm.. Probably best to wait until they've fixed it then.

Dr Philip WelchOpen Door Logistics LimitedSpecialists in open source solutions for transport logistics.


Email: info@opendoorlogistics.com
Linkedin:  http://uk.linkedin.com/in/welchphilip
Skype: opendoorlogistics
Tel:  +44 (0) 20 8144 4009
Website:  http://www.opendoorlogistics.com






---- On Thu, 17 Nov 2016 12:19:27 +0000 Peter&lt;notifications@github.com&gt; wrote ---- 

Ah, it is already known to them: https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-5289?filter=allopenissues

karussell:
For now I've just excluded the shp module from JDK9 build and will still merge.

karussell:
Merged it via this commit https://github.com/graphhopper/graphhopper/commit/1960b0f6239cb66fd6847fb865880a10e78eb025

The ugliness here is that I wanted it in a single commit (for simplicity) but now it looks like a change only from vvikas still when you look at 'git log' the author is "vvikas and Philip Welch" although it did not pick up the second email, obviously. I hope this is okay and have added you as a contributor in [our list](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTORS.md) and hopefully more changes are coming :) (?)

PGWelch:
Cool. Don't worry about who the change shows up from - it's not a problem.

PR #875, opened 2016-11-17
Use Access Tags, in particular destination

### LINKED ISSUE #733: Add support for restriction value 'destination' for car
From: https://discuss.graphhopper.com/t/restrictedvalues-destination/475

The access restriction value destination is not regarded. A high penalty has to be applied when changing from an unrestricted way to a way with destination restriction.

Test case 1: Legal path avoids the hamlet Heitersberg and will make a ~20km detour.
https://graphhopper.com/maps/?point=47.414905%2C8.343494&point=47.415366%2C8.346804&locale=en-US&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale

Test case 2: Fastest path is through Bellikon 
https://graphhopper.com/maps/?point=47.36249%2C8.367419&point=47.415231%2C8.346503&locale=en-US&vehicle=car&weighting=fastest&elevation=true&layer=Omniscale

boldtrn:
I added parsing of the access=destination tag to the CarFlagEncoder, which intents to fix #733.

I also added parsing of access values to the DataFlagEncoder. Right now this happens only for a car like vehicle. I created sort of 3 categories: ACCESSIBLE, EVENTUALLY_ACCESSIBLE, NOT_ACCESSIBLE. Whereas access values like customers, destination, and so on are EVENTUALLY_ACCESSIBLE.

Right now I add a penalty of factor 10 to edges that are EVENTUALLY_ACCESSIBLE. This could be improved if we would add "geospatial knowledge" to the GenericWeighting. For example if an edge is close to a waypoint, the EVENTUALLY_ACCESSIBLE penalty could be very low, whereas if the edge is far from a waypoint the penalty would be comparably high.


core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
```diff
@@ -37,6 +37,9 @@
```
karussell: Cannot comment on the faundau_destination.osm.xml file somehow ... so here it goes :)

Can we avoid another osm xml? E.g. I found destination tags in krems.osm.gz or north-bayreuth.osm.gz

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -37,6 +37,7 @@
  * @author Peter Karich
  */
 public class DataFlagEncoder extends AbstractFlagEncoder {
+
```
karussell: The changes here look nice and reasonable to me - would you mind to add a unit test in combination with the GenericWeighting?  (not just the integration test)

core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
boldtrn: Ah nice, haven't seen that before, thanks.

karussell:
Thanks!

Issue #562, opened 2015-11-16
Remove LGPL code from core


karussell:
A few places use LGPL, it is not an issue for most users and companies but for some due to restrictiv license obligations. We need to 
- move IntDoubleBinHeap out of the core
- replace trove4j with hppc or similar, see #34 
- can we replace OSM-binary? (this is no longer in the core)

devemux86:
How about separating more clearly ```core``` from ```tools```?
i.e. ```core``` for reading graphs and routing and ```tools``` (or else) for reading osm and creating graphs.
Right now graph creation is integrated in ```core```, so the osm read dependencies.

karussell:
:+1: 
that is the next step :) which will - at the end - move all OSM stuff into `tools` or even better into a new `osm-import` module? Otherwise e.g. web needs a dependency on `tools` which also contains unnecessary swing stuff. 

But e.g. the heap can't be moved up-stream as `core` depends on it. Maybe we'll replace the classes there at some point, but for me this is none critical. Just wanted them separated.

karussell:
The question is also a bit what we do with the GraphHopper class which belongs to the `core` but needs the osm import and we don't want reflection. 

So the result will be something like:
```java
GraphHopper hopper = new GraphHopper() {
     protected DataReader createReader( GraphHopperStorage ghStorage )    {
        return initOSMReader(new OSMReader(ghStorage));
    }
}
...
hopper.importOrLoad();
```

or more explicitely:

```java
GraphHopper hopper = new GraphHopper().setDataReader(new OSMReader(osmFile));
...
hopper.importOrLoad();
```
Both would keep in mind e.g. cases like #277 

Or is this too complex for the normal developer? And we should implement additionally a special GraphHopper subclass?

```java
GraphHopper hopper = new GraphHopperOSM().setOSMFile(file);
...
hopper.importOrLoad();
```

Or even avoid renaming and name the subclass `GraphHopper` and the super class e.g. `GraphHopperCore`?

devemux86:
(I agree that reflection should be avoided)

Your 3rd suggestion with separate classes seems more feasible to normal developers, which separates better the osm import.

karussell:
Have to think about this - just to be sure - you meant the `GraphHopperOSM` style? Let us move to the new #563

K-Leon:
I just had a look in IOS and Android Ports of Graphhopper:

Right now there are dependencies on LGPL Code for both - Android and iOS? 

[...]

Issue #34, opened 2013-04-19
Compare trove4j with hppc


karussell:
[hppc](http://labs.carrotsearch.com/hppc.html) looks [fast](http://mail-archives.apache.org/mod_mbox/mahout-dev/201103.mbox/%3CAANLkTi=gbd-GFxpQu_86haq16-gikTrwwZ8x4Tp250jH@mail.gmail.com%3E), is 2 times smaller and is apache licensed. so we should try out if speed is the same after replacing it. test:
- preparation of contraction
- routing speed for CH
- routing speed for normal algorithms. as more nodes are visited => collection lib needs to be more efficient

Related to #562

karussell:
There are lots of alternatives these days which might be worth to investigate, also maybe they are more active or even faster: see collections alternative https://github.com/Vedenin/useful-java-links/#collections

karussell:
But it looks like after 4 years there will be a 3.1 version of trove4j :) https://bitbucket.org/trove4j/trove/issues/64/31-release#comment-26676294

karussell:
Will reopen this due to the licensing stuff needs to be simplified at some point

PR #878, opened 2016-11-20
Replace trove4j with hppc

karussell:
This fixes an early issue #34 and improves #562 

The advantages:

 * core no longer relies on any LGPL dependency, which makes using the library a no-brainer then on Android & iOS (license-wise one could have argumented that e.g. dalvik merges trove with our core and 'infects' it and for iOS it could be even worse)
 * hppc is a smaller library (<2MB instead of >3MB)
 * seems to work very similar fast now these days

Disadvantages

 * took a bit time to make it working, see [here](https://groups.google.com/forum/#!topic/java-high-performance-primitive-collections/KvqyWiMTX8g) for the detailed issues
 * some unknown bugs or performance changes?

Not yet done:

 - [x] hppc has only 0 for default value but -1 was required in constructor -> GHTBitSet (TIntHashSet); OSMShapeReader -> TObjectIntHashMap; OSMReader -> edgeIdToOsmWayIdMap (GHIntLongHashMap);
 - [x] fix some license headers (GHIntHashSet.java)
 - [x] review by community
 - [x] deployed and tested to staging server world wide
 - [x] deploy to production (also ww)
 - [x] is hashmixing only important for mixing two hashes or also when resizing a hash? Or other way around if trove4j also this problem we can probably ignore it for now but must be aware of it e.g. if there are performance problems.
 - [x] fix android docs about trove (docs/android/index.md#Problems)
 - [ ] fix iOS with @clns its help

### DIFF
--- a/core/src/main/java/com/graphhopper/coll/GHIntArrayList.java
+++ b/core/src/main/java/com/graphhopper/coll/GHIntArrayList.java
@@ -49,7 +49,7 @@ public class GHIntArrayList extends IntArrayList {
     }
 
     public final GHIntArrayList fill(final int max, final int value) {
-        // TODO use System.arraycopy and direct buffer somehow
+        // TODO fill 100 then copy using System.arraycopy, then duplicate this via copying again etc
         for (int i = 0; i < max; i++) {
             add(value);
         }
--- a/core/src/main/java/com/graphhopper/coll/GHIntArrayList.java
+++ b/core/src/main/java/com/graphhopper/coll/GHIntArrayList.java
@@ -1,0 +1,77 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for 
+ *  additional information regarding copyright ownership.
+ * 
+ *  GraphHopper GmbH licenses this file to you under the Apache License, 
+ *  Version 2.0 (the "License"); you may not use this file except in 
+ *  compliance with the License. You may obtain a copy of the License at
+ * 
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.coll;
+
+import com.carrotsearch.hppc.IntArrayList;
+import java.util.Random;
+
+/**
+ *
+ * @author Peter Karich
+ */
+public class GHIntArrayList extends IntArrayList {
+    public GHIntArrayList() {
+        super(10);
+    }
+
+    public GHIntArrayList(int capacity) {
+        super(capacity);
+    }
+
+    public GHIntArrayList(GHIntArrayList list) {
+        super(list);
+    }
+
+    public final GHIntArrayList reverse() {
+        final int[] buffer = this.buffer;
+        for (int start = 0, end = size() - 1; start < end; start++, end--) {
+            // swap the values
+            int tmp = buffer[start];
+            buffer[start] = buffer[end];
+            buffer[end] = tmp;
+        }
+        return this;
+    }
+
+    public final GHIntArrayList fill(final int max, final int value) {
+        // TODO use System.arraycopy and direct buffer somehow
+        for (int i = 0; i < max; i++) {
+            add(value);
+        }
+        return this;
+    }
+
+    public final GHIntArrayList shuffle(Random random) {
+        int[] buffer = this.buffer;
+        int max = size();
+        int maxHalf = max / 2;
+        for (int x1 = 0; x1 < maxHalf; x1++) {
+            int x2 = random.nextInt(maxHalf) + maxHalf;
+            int tmp = buffer[x1];
+            buffer[x1] = buffer[x2];
+            buffer[x2] = tmp;
+        }
+        return this;
+    }
+
+    public static GHIntArrayList from(int... elements) {
+        final GHIntArrayList list = new GHIntArrayList(elements.length);
+        list.add(elements);
+        return list;
+    }
+}
--- a/core/src/main/java/com/graphhopper/coll/GHIntArrayList.java
+++ b/core/src/main/java/com/graphhopper/coll/GHIntArrayList.java
@@ -39,9 +39,10 @@ public class GHIntArrayList extends IntArrayList {
 
     public final GHIntArrayList reverse() {
         final int[] buffer = this.buffer;
+        int tmp;
         for (int start = 0, end = size() - 1; start < end; start++, end--) {
             // swap the values
-            int tmp = buffer[start];
+            tmp = buffer[start];
             buffer[start] = buffer[end];
             buffer[end] = tmp;
         }
--- a/core/src/main/java/com/graphhopper/coll/GHSortedCollection.java
+++ b/core/src/main/java/com/graphhopper/coll/GHSortedCollection.java
@@ -17,9 +17,8 @@
  */
 package com.graphhopper.coll;
 
-import gnu.trove.iterator.TIntIterator;
-import gnu.trove.set.hash.TIntHashSet;
-
+import com.carrotsearch.hppc.cursors.IntCursor;
+import java.util.Iterator;
 import java.util.Map.Entry;
 import java.util.TreeMap;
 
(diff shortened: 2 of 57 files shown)


karussell:
Can I interpret the two thumbs as review :) ? @devemux86 @boldtrn? Just wanted someone to have a quick scan over it that I did not introduce very obvious problems :) (will also have another look now)

core/src/test/java/com/graphhopper/routing/AbstractRoutingAlgorithmTester.java
```diff
@@ -592,7 +592,7 @@ public void testViaEdges_BiGraph() {
     public void testViaEdges_WithCoordinates() {
         GraphHopperStorage ghStorage = createTestStorage();
         Path p = calcPath(ghStorage, 0, 1, 2, 3);
-        assertEquals(Helper.createTList(9, 1, 2, 8), p.calcNodes());
```
karussell: virtual nodes get a different id because of the hashmap `edge2res` in QueryGraph creates them in a different order

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -219,14 +217,14 @@ private void prepareRestrictionRelation(ReaderRelation relation) {
...
+    private LongSet getOsmWayIdSet() {
         return osmWayIdSet;
     }
 
-    private TIntLongMap getEdgeIdToOsmWayIdMap() {
+    private IntLongMap getEdgeIdToOsmWayIdMap() {
         if (edgeIdToOsmWayIdMap == null)
-            edgeIdToOsmWayIdMap = new TIntLongHashMap(getOsmWayIdSet().size(), 0.5f, -1, -1);
-
+            edgeIdToOsmWayIdMap = new GHIntLongHashMap(getOsmWayIdSet().size(), 0.5f);
```
karussell: Here we do not return -1 anymore if empty but we can easily check if the key is contained via the direct method

reader-shp/src/main/java/com/graphhopper/reader/shp/OSMShapeFileReader.java
```diff
@@ -53,13 +50,12 @@
  * @author Philip Welch
  */
 public class OSMShapeFileReader extends ShapeFileReader {
-    private static final int COORD_STATE_UNKNOWN = -1;
+    private static final int COORD_STATE_UNKNOWN = 0;
```
karussell: empty value in the hashmap is now 0 ... maybe replace this with a contains check? But as we do not use 0 this should be fine too

devemux86:
Oh sorry @karussell I just expressed my opinion for your decision. :slightly_smiling_face:

karussell:
Got a bit feedback from the hppc author in the aboved linked thread.

Tried also [koloboke](https://github.com/leventov/Koloboke) as it was mentioned from him and it could be better in terms of speed but the perf numbers are identical at least for CH and non-CH routing. Also both communities are rather small but hppc is used more ([90 deps](http://mvnrepository.com/artifact/com.carrotsearch/hppc)) and from bigger projects like ElasticSearch and does everything we need (at least now).

@devemux86 no problem

boldtrn:
> Oh sorry @karussell I just expressed my opinion for your decision. :slightly_smiling_face:

Me, too :). But I will have a look at your changes and will comment if anything might be an issue.

core/src/main/java/com/graphhopper/coll/GHIntArrayList.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public GHIntArrayList(GHIntArrayList list) {
+        super(list);
+    }
+
+    public final GHIntArrayList reverse() {
+        final int[] buffer = this.buffer;
+        for (int start = 0, end = size() - 1; start < end; start++, end--) {
+            // swap the values
+            int tmp = buffer[start];
```
boldtrn: Maybe we should define the int outside of the for? Not sure if this gets optimized, but I introduced a "bug" with String once, so I became a bit cautious with these things.

core/src/main/java/com/graphhopper/coll/GHIntArrayList.java
```diff
@@ -0,0 +1,77 @@
...
+        for (int i = 0; i < max; i++) {
+            add(value);
+        }
+        return this;
+    }
+
+    public final GHIntArrayList shuffle(Random random) {
+        int[] buffer = this.buffer;
+        int max = size();
+        int maxHalf = max / 2;
```
boldtrn: Why only half? For example the implementation of Collections.shuffle does it for every element except the first.

core/src/main/java/com/graphhopper/coll/GHSortedCollection.java
```diff
@@ -105,14 +105,21 @@ public int pollKey() {
...
+        GHIntHashSet set = e.getValue();
         if (set.isEmpty()) {
             throw new IllegalStateException("internal set is already empty!?");
         }
-        int val = iter.next();
-        iter.remove();
+
+        Iterator<IntCursor> iter = set.iterator();
+        final int val = iter.next().value;
+        set.removeAll(new IntPredicate() {
```
boldtrn: Why not just use http://carrotsearch.github.io/hppc/releases/0.7.0/api/com/carrotsearch/hppc/IntHashSet.html#removeAll-int- ?

core/src/main/java/com/graphhopper/util/GHUtility.java
```diff
@@ -359,6 +361,18 @@ public static int getEdgeFromEdgeKey(int edgeKey) {
     }
 
     /**
+     * @return a copy of inputArray from [offset, offset+length)
+     */
+    public static LongIndexedContainer copy(LongArrayList inputArray, int offset, int length) {
```
boldtrn: Is this faster than construction a LongArrayList with the exptected lenght and this method http://carrotsearch.github.io/hppc/releases/0.7.1/api/com/carrotsearch/hppc/LongArrayList.html#add-long:A-int-int- ?

core/src/main/java/com/graphhopper/coll/GHSortedCollection.java
boldtrn: Sorry, I should have read the thread first.... 

> It's better to use removeAll(IntPredicate predicate).

core/src/main/java/com/graphhopper/coll/GHIntArrayList.java
karussell: To avoid a "re-shuffle" but this does not really matter to me

core/src/main/java/com/graphhopper/coll/GHSortedCollection.java
karussell: Hmmh, now that I'm re-reading this this looks strange to me as remove(int) seems to be very similar. Maybe he misunderstand something.

core/src/main/java/com/graphhopper/util/GHUtility.java
karussell: Ah, nice. It should be identical and so we can remove this.

karussell:
Thanks for the review @boldtrn - do you understand the hashmixing problem Dawid mentioned? [Here](https://github.com/carrotsearch/hppc/blob/master/hppc-examples/src/test/java/com/carrotsearch/hppc/examples/HppcExample_006_HashAndScatterMaps.java) is some code&explanation for this

boldtrn:
> do you understand the hashmixing problem Dawid mentioned? 

I am not sure. This sounds to me like a typical, hash collision scenario. Bascially with open addressing you are calculating a hash (index of an array) for a key. If you put an element into the hashtable you are putting it into the index of the array. If the index is already used, there are different strategies to resolve the issue, one is linear probing. So what you do with linear probing you just put the element at index+1 if the slot is available, if not put it at index+2, and so on. So the problem with this is that in the worst case you end up with linear put and get operations, whereas this should be constant for a hashmap.

So from what I understand is that if you are downsizing an existing map, it might happen that you create a lot of these collisions and therefore you end up with a hash map that might be as bad as a random array.

So if you are iterating over the map in key order for creating a new table, the chance for creating very bad tables is increased. Lets assume you have 3 key collisions for the first key in your map, so you start to put them at index: 0,1,2. In the bigger map that's not a problem as the next key might be at index 5, but with the smaller map this element might be put at index 1, but index 1 is occupied so you put it at index 3. This goes on and on, therefore the lookup of these elements becomes very bad. If you would look for the element that should be at index 1, you go there and compare the key, you find the key does not match so you need to check index 2, and so on.

So if you would do this with iterating randomly over the hashmap the chances for creating a bad hashmap might be decreased.

That's how I understand it, but I am not sure if he meant that :).

core/src/main/java/com/graphhopper/coll/GHIntArrayList.java
boldtrn: Ok, so if you don't need a guaranteed complete shuffle this should be fine.

karussell:
Yes, this part I understand. But I'm not sure if we would currently stumble over such a problem as we 1. create a different hashing strategy depending on the size of the hash (`HashOrderMixing.constant`) and 2. do not merge two hashes somewhere else (the two would even need to be of the same size)

boldtrn:
Ok, in this case I guess we would not stumble over the problem.

devemux86:
If I may report that using extensively GraphHopper with hppc lately and not seen so far any issues with the change. :slightly_smiling_face:

karussell:
Nice & Thanks for the feedback - we did the same observation :) !

PR #879, opened 2016-11-21
making findNetworkEntriesSingleRegion public

kodonnell:
To allow use of `findNetworkEntriesSingleRegion` when extending `LocationIndexTree` to create different search algorithms than `findNetworkEntries`. Use cases are e.g. map-matching and custom implementations e.g. [here](https://github.com/graphhopper/map-matching/pull/67).


karussell:
Thanks!

Issue #789, opened 2016-09-09
Block area and roads per request


karussell:
We need a built-in functionality for the [forbiddenEdgeId-Weighting](https://github.com/graphhopper/graphhopper/blob/master/docs/core/weighting.md) to block an area or a list of lookups.

Currently this is only possible as a post processing step before the routing via the [overlay mechanism](https://github.com/graphhopper/graphhopper/pull/781)

abxaz92:
I subscribe too

boldtrn:
How would something like this look like?

The way I understand this issue is, that I would add for example:
- a list of edgeIds that are forbidden or have an massively increased weight for one particular request
- a list of coordinates with distances, e.g. `forbidden_point=lat,lng,radius`, all edges around this point in a certain radius are found ad added as forbidden edgeIds. Radius in this scenario could also be the maxRegion parameter of the LocationIndexTree, so no exact meter requirement but more like a "big area" or "small area".

karussell:
We talked a bit offline about the new Data API I plan to build via POSTing JSON and probably here it would also make more sense to start with the JSON API for the `route` end point and include a more or less identical format but not persistent. What I mean is: the Data API will allow us to disable certain edges or reduce/increase speed in real time but persistent for all following route requests (all non-CH). The format should be the same which is already implemented in the [overlay reader #781. Now a POST request with an identical format against the `route` endpoint would only temporarily limit or change certain edge properties. So then there are three changes possible:
 
 * before an import, also applicable to CH, already implemented with the overlay reader
 * real time permanent update (for all later requests), will be implemented in #845
 * real time temporary update (just per request), should be implemented here, but for now probably the `forbidden_point` style you have in mind is sufficient. And create a new POST API for the `route` endpoint in a different issue. I've also a rough picture for this and its features.

But besides this more intense changes there are several other quirks with this issue: e.g. how to handle a location lookup if the edge is in the forbidden list (?) Also what should we do if the area is too big (i.e. too many edges would be involved)?

karussell:
Fixed via #890

jsroyal:
Is that this problem solved? Can I see the merge ID of PR?

boldtrn:
@jsroyal just look at the comment above and yes it is fixed.

jsroyal:
Thank you! @boldtrn, I went through the code and it is nice implemented.
and I am not able to block the path in web UI  as I can see the given see the ##899 vehicle=generic&ch.disable=true.
[...]

PR #890, opened 2016-11-28
Allow to block area per request

boldtrn:
This PR intents to improve #789. You can either block areas (given as BBox), points (closest edge from Lat,Lng), or by edge_id.

This is currently implemented only for the GenericWeighting, as my Approach is to inject the Weighting and change the weight of certain edges.

Here is an example. This is the unblocked route:
block_nothing

This is the route with the black area blocked:
blocked_area

This is the route with the black area and black point blocked:
blocked_area_point

All three cases can be seen in the new Integration test.

Things we might want to add:

- Returned the blocked edge ids, this would allow us to cache the blocked edges for further requests (not sure how much speed this would get us).
- Add the possibility to manage arrays or lists in the HintsMap. Right now a limitation of the current approach is that the parameters are single strings, for example: `routing.non_ch.blocked_points=lat,lon,lat,lon`. This approach works for now, but it would be nice to be able to add every point seperated by &. On the other hand urls might get very long, as the naming is currently rather long.
- Add a (helper)class that searches the Graph for edges that are in a BBox, or maybe circle, or ...

BTW: I added calling super() in GHBitSetImpl(), I figured we might forgot to add it at this position, but I am not sure, so if that was wrong, I can remove it?

BTW2: I did a rough speed comparison on my laptop. I planned a route accross Baden Württemberg, from Lindau to Karlsruhe, so approx. 260km. Without blocking anything the request took about 1s. With blocking a huge area (approx. half Baden Württemberg - BBox: `8.2645,48.1056,9.9124,48.8557` - the blocked edge set **contains 790876 edges**) the request took about 1.5s, which surprised myself, I thought the speed would be massively decreased. 

Unblocked:
unblocked_bw

Blocked:
blocked_bw

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -90,6 +90,7 @@ public class GraphHopper implements GraphHopperAPI {
     private boolean simplifyResponse = true;
     private TraversalMode traversalMode = TraversalMode.NODE_BASED;
     private int maxVisitedNodes = Integer.MAX_VALUE;
+    private String blockedRectangularAreas = "";
 
     private int nonChMaxWaypointDistance = Integer.MAX_VALUE;
     // for index
@@ -645,6 +646,7 @@ public class GraphHopper implements GraphHopperAPI {
         maxVisitedNodes = args.getInt(Routing.INIT_MAX_VISITED_NODES, Integer.MAX_VALUE);
         maxRoundTripRetries = args.getInt(RoundTrip.INIT_MAX_RETRIES, maxRoundTripRetries);
         nonChMaxWaypointDistance = args.getInt(Parameters.NON_CH.MAX_NON_CH_POINT_DISTANCE, Integer.MAX_VALUE);
+        blockedRectangularAreas = args.get(Routing.BLOCKED_RECTANGULAR_AREAS, "");
 
         return this;
     }
@@ -950,12 +952,12 @@ public class GraphHopper implements GraphHopperAPI {
         final GHIntHashSet blockedEdges = new GHIntHashSet();
         final List<Shape> blockedShapes = new ArrayList<>();
         // We still need EdgeIds for PointBlocking
-        final boolean blockByShape = hints.getBool(Parameters.NON_CH.BLOCK_BY_SHAPE, false);
+        final boolean blockByShape = hints.getBool(Routing.BLOCK_BY_SHAPE, true);
         GraphBrowser browser = new GraphBrowser(graph, locationIndex);
         EdgeFilter filter = new DefaultEdgeFilter(encoder);
 
         // Add Blocked Edges
-        String blockedEdgesStr = hints.get(Parameters.NON_CH.BLOCKED_EDGES, "");
+        String blockedEdgesStr = hints.get(Routing.BLOCKED_EDGES, "");
         if (!blockedEdgesStr.isEmpty()) {
             String[] blockedEdgesArr = blockedEdgesStr.split(",");
             for (int i = 0; i < blockedEdgesArr.length; i++) {
@@ -964,11 +966,11 @@ public class GraphHopper implements GraphHopperAPI {
         }
 
         // Add Blocked Points
-        String blockedPointsStr = hints.get(Parameters.NON_CH.BLOCKED_POINTS, "");
+        String blockedPointsStr = hints.get(Routing.BLOCKED_POINTS, "");
         if (!blockedPointsStr.isEmpty()) {
             String[] blockedPointsArr = blockedPointsStr.split(",");
             if (blockedPointsArr.length % 2 != 0) {
-                throw new IllegalArgumentException(Parameters.NON_CH.BLOCKED_POINTS + " need to be defined as lat,lon");
+                throw new IllegalArgumentException(Routing.BLOCKED_POINTS + " need to be defined as lat,lon");
             }
 
             double lat;
@@ -981,11 +983,17 @@ public class GraphHopper implements GraphHopperAPI {
         }
 
         // Add Blocked Rectangular Areas
-        String blockedAreasStr = hints.get(Parameters.NON_CH.BLOCKED_RECTANGULAR_AREAS, "");
+        String blockedAreasFromRequest = hints.get(Routing.BLOCKED_RECTANGULAR_AREAS, "");
+        String blockedAreasStr;
+        if (!this.blockedRectangularAreas.isEmpty() && !blockedAreasFromRequest.isEmpty()) {
+            blockedAreasStr = this.blockedRectangularAreas + "," + blockedAreasFromRequest;
+        } else {
+            blockedAreasStr = this.blockedRectangularAreas + blockedAreasFromRequest;
+        }
         if (!blockedAreasStr.isEmpty()) {
             String[] blockedAreasArr = blockedAreasStr.split(",");
             if (blockedAreasArr.length % 4 != 0) {
-                throw new IllegalArgumentException(Parameters.NON_CH.BLOCKED_RECTANGULAR_AREAS + " need to be defined as left,bottom,right,top");
+                throw new IllegalArgumentException(Routing.BLOCKED_RECTANGULAR_AREAS + " need to be defined as left,bottom,right,top");
             }
 
             double left;
@@ -999,9 +1007,9 @@ public class GraphHopper implements GraphHopperAPI {
                 top = Double.parseDouble(blockedAreasArr[4 * i + 3]);
 
                 final BBox bbox = new BBox(left, right, bottom, top);
-                if(blockByShape){
+                if (blockByShape) {
                     blockedShapes.add(bbox);
-                }else {
+                } else {
                     browser.findEdgesInShape(blockedEdges, bbox, filter);
                 }
             }
@@ -1009,33 +1017,33 @@ public class GraphHopper implements GraphHopperAPI {
         }
 
         // Add Blocked Circular Areas
-        String blockedCircularAreasStr = hints.get(Parameters.NON_CH.BLOCKED_CIRCULAR_AREAS, "");
+        String blockedCircularAreasStr = hints.get(Routing.BLOCKED_CIRCULAR_AREAS, "");
         if (!blockedCircularAreasStr.isEmpty()) {
             String[] blockedCircularAreasArr = blockedCircularAreasStr.split(",");
             if (blockedCircularAreasArr.length % 3 != 0) {
                 //TODO: Do we ant radius or diameter?
-                throw new IllegalArgumentException(Parameters.NON_CH.BLOCKED_CIRCULAR_AREAS + " need to be defined as lat,lng,radius");
+                throw new IllegalArgumentException(Routing.BLOCKED_CIRCULAR_AREAS + " need to be defined as lat,lng,radius");
             }
 
             double lat;
             double lng;
             int radius;
             for (int i = 0; i < blockedCircularAreasArr.length / 3; i++) {
-                lat= Double.parseDouble(blockedCircularAreasArr[3 * i]);
+                lat = Double.parseDouble(blockedCircularAreasArr[3 * i]);
                 lng = Double.parseDouble(blockedCircularAreasArr[3 * i + 1]);
                 radius = Integer.parseInt(blockedCircularAreasArr[3 * i + 2]);
                 Circle circle = new Circle(lat, lng, radius);
-                if(blockByShape){
+                if (blockByShape) {
                     blockedShapes.add(circle);
-                }else {
+                } else {
                     browser.findEdgesInShape(blockedEdges, circle, filter);
                 }
             }
 
         }
 
-        cMap.put(Parameters.NON_CH.BLOCKED_EDGES, blockedEdges);
-        cMap.put(Parameters.NON_CH.BLOCKED_SHAPES, blockedShapes);
+        cMap.put(Routing.BLOCKED_EDGES, blockedEdges);
+        cMap.put(Routing.BLOCKED_SHAPES, blockedShapes);
 
         return cMap;
     }
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -17,6 +17,7 @@
  */
 package com.graphhopper;
 
+import com.graphhopper.coll.GHIntHashSet;
 import com.graphhopper.reader.DataReader;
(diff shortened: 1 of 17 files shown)


core/src/main/java/com/graphhopper/util/Parameters.java
```diff
@@ -141,5 +141,11 @@
          * Describes the max allowed distance between two consecutive waypoints of a non ch request. Distance is in Meter
          */
         public static final String MAX_NON_CH_POINT_DISTANCE = ROUTING_INIT_PREFIX + NON_CH_PREFIX + "max_waypoint_distance";
+
+        public static final String BLOCKED_EDGES = ROUTING_INIT_PREFIX + NON_CH_PREFIX + "blocked_edges";
```
boldtrn: Ok, I see. Will remove it.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -938,6 +941,97 @@ public Weighting createWeighting(HintsMap hintsMap, FlagEncoder encoder) {
         throw new IllegalArgumentException("weighting " + weighting + " not supported");
     }
 
+    private GHIntHashSet getBlockedEdges(HintsMap hints, FlagEncoder encoder) {
```
boldtrn: I had exactly the same thought :), this is what I meant with:

> Add a (helper)class that searches the Graph for edges that are in a BBox, or maybe circle, or ...

Will set up something :)

boldtrn: Ok, I moved the common code to a different class, but the method is still quite long, not sure if we want to move everything out of the GraphHopper class?

boldtrn:
I updated my branch, with your proposed changes, also I added the function to Block Circular Areas.

karussell:
Wow, that was fast - thanks! I'll have a closer look now!

Another idea (maybe you already have it included) I had for larger areas could be to switch to ID-less blocking via something like `if(bbox.contains(edge.getAdjNode->lat,lon)) return infinity`

boldtrn:
> Another idea (maybe you already have it included) I had for larger areas could be to switch to ID-less blocking via something like if(bbox.contains(edge.getAdjNode->lat,lon)) return infinity

Excellent idea! I added this. There is one ugly thing however, I needed to pass a NodeAccess object to the GenericWeighting. Not sure if there is a better option to get the Coordinates of an Edge in the Weighting?

Anyway, for a single large shape the performance boost of this technique is significant. A user can select to block by shape by passing `non_ch.block_by_shape=true`, this will use the Shape->contains method in the edge instead of collecting EdgeIds beforehand. This technique is not used for single point blocking, so if one blocks some points+areas, the weighting will check for EdgeIds+Shape contains.

karussell:
Nice - thanks!

> Coordinates of an Edge in the Weighting

The only more costly option would be to do edge.fetchWayGeometries.

Instead of this:

> cMap.put("node_access", getGraphHopperStorage().getNodeAccess());

the QueryGraph.getNodeAccess is necessary here (plus a currently failing test ;))

> A user can select to block by shape by passing non_ch.block_by_shape=true

Ok. Later on we'll have to make this more production grade ie. restrictive (limit by area and number of geometries) and also should automate the switching to the 'shape blocking'.

boldtrn:
> QueryGraph.getNodeAccess

Oh ok, I wasn't aware of that, thanks. The problem with this is, that the QueryGraph has not been created at this point. So what I could do is to either create a new QueryGraph at the current position or change the calcPaths method to check if the weighting is a GenericWeighting and inject the blocking Attributes there.

> Ok. Later on we'll have to make this more production grade ie. restrictive (limit by area and number of geometries) and also should automate the switching to the 'shape blocking'.

Ok, I can add more safety features here.

karussell:
> The problem with this is, that the QueryGraph has not been created at this point.

Hmmh, ugly. Injecting means we would need to add a new Weighting.setGraph method. But moving it before the weighting creation would not work for CH:

```java
weighting = ((PrepareContractionHierarchies) tmpAlgoFactory).getWeighting();
routingGraph = ghStorage.getGraph(CHGraph.class, weighting);
```

boldtrn:
Injecting means we would need to add a new Weighting.setGraph method.

Yes, I would add a setNodeAccessMethod. Probably something like:

```
if(weighting instanceof GenericWeighting)
   weighting.setNodeAccess(queryGraph.getNodeAccess);
``` 

> But moving it before the weighting creation would not work for CH

Yes exactly. Therefore I would create a second RoutingGraph, which is also very ugly. I am not sure about the perfomance impact of creating a RoutingGraph?

karussell:
We cannot do that for CH as the weighting is expected to be stateless and not created per request (chGraph.getWeighting). So the correct way I think would be to pass the QueryGraph (as `Graph`) to createWeighting.

boldtrn:
> We cannot do that for CH as the weighting is expected to be stateless and not created per request (chGraph.getWeighting).

Yes, but blocking per Request won't work for CH anyway? Also I think the GenericWeighting is meant to be used for non-ch? Maybe I am miss thinking something right now?

karussell:
> Also I think the GenericWeighting is meant to be used for non-ch?

GenericWeighting can be used also for CH but then with fixed parameters only. That is a possible use case: try out the weighting with non-CH until you are happy, then freeze your weighting and use CH for best performance.

> Yes, but blocking per Request won't work for CH anyway?

It should work, not per request, but if the blocking parts are known before preparation.

boldtrn:
> QueryGraph (as Graph) to createWeighting

Thanks again :) But at the time of createWeighting, the QueryGraph does not exist yet, does it?

karussell:
> if(weighting instanceof GenericWeighting)
>    weighting.setNodeAccess(queryGraph.getNodeAccess);

Let us introduce a new method `Weighting.setGraph(Graph g)` - ie. we could set also a non-QueryGraph like necessary if we call createWeighting(...., Graph) for CH.

> Thanks again :) But at the time of createWeighting, the QueryGraph does not exist yet, does it?

I would create it for non-CH before the createWeighting call

boldtrn:
> we could set also a non-QueryGraph like necessary if we call createWeighting(...., Graph) for CH.

I just started with this change, then I figured, that we call the createWeighting for CH before the Graph exists at: https://github.com/graphhopper/graphhopper/blob/f73a6892d0bd26bb184acb7db124ff7133205c62/core/src/main/java/com/graphhopper/GraphHopper.java#L780

Solving this issue for non-ch is easy, but for CH I think we would need to inject the graph in the post-processing step. I think we could do this before this line: https://github.com/graphhopper/graphhopper/blob/0f7562a6ac9f24c49d8d58b6f9cb5c4e09438156/core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java#L98

Do you think this will work? I would call the set graph method in this case if the weighting is an instanceof the GenericWeighting.

boldtrn:
Ok, so I added the Graph to the Genericweighting. It works with CH, as well, see the TODO in the GenericWeighting, this needs to be made configureable, maybe even in the config?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -839,7 +839,7 @@ private void initCHAlgoFactoryDecorator() {
         if (!chFactoryDecorator.hasWeightings())
             for (FlagEncoder encoder : encodingManager.fetchEdgeEncoders()) {
                 for (String chWeightingStr : chFactoryDecorator.getWeightingsAsStrings()) {
-                    Weighting weighting = createWeighting(new HintsMap(chWeightingStr), encoder);
+                    Weighting weighting = createWeighting(new HintsMap(chWeightingStr), encoder, null);
```
karussell: here we have to pass the graphHopperStorage (ie. base graph)

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -914,17 +914,20 @@ private boolean isPrepared() {
...
         String weighting = hintsMap.getWeighting().toLowerCase();
 
         if (encoder.supports(GenericWeighting.class)) {
             DataFlagEncoder dataEncoder = (DataFlagEncoder) encoder;
             ConfigMap cMap = dataEncoder.readStringMap(hintsMap);
-            cMap = setupBlocking(cMap, hintsMap, dataEncoder);
-            return new GenericWeighting(dataEncoder, cMap);
+            cMap = setupBlocking(cMap, hintsMap, dataEncoder, graph);
+            GenericWeighting genericWeighting =  new GenericWeighting(dataEncoder, cMap);
+            genericWeighting.setGraph(graph);
```
karussell: hmmh, maybe remove it from the interface and just have the graph in the constructor?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1125,20 +1127,20 @@ else if (!(tmpAlgoFactory instanceof PrepareContractionHierarchies))
...
-                    routingGraph = ghStorage.getGraph(CHGraph.class, weighting);
-
+                    queryGraph = new QueryGraph(ghStorage.getGraph(CHGraph.class, weighting));
+                    queryGraph.lookup(qResults);
                 } else {
                     checkNonChMaxWaypointDistance(points);
-                    weighting = createWeighting(hints, encoder);
+                    queryGraph = new QueryGraph(ghStorage);
+                    queryGraph.lookup(qResults);
+                    weighting = createWeighting(hints, encoder, queryGraph);
```
karussell: Perfect!

core/src/main/java/com/graphhopper/routing/weighting/GenericWeighting.java
```diff
@@ -73,10 +75,14 @@ public GenericWeighting(DataFlagEncoder encoder, ConfigMap cMap) {
         accessType = gEncoder.getAccessType("motor_vehicle");
         blockedEdges = cMap.get(Parameters.NON_CH.BLOCKED_EDGES, new GHIntHashSet(0));
         blockedShapes = cMap.get(Parameters.NON_CH.BLOCKED_SHAPES, Collections.EMPTY_LIST);
-        nodeAccess = cMap.get("node_access", null);
-        if(!blockedShapes.isEmpty() && nodeAccess == null){
-            throw new IllegalStateException("You have to pass a valid NodeAccess if you want to use Shape Blocking");
+        // TODO Uncomment this, if you want to use Blocked Shapes with CH
```
karussell: What do you mean here?

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: Yes, I would have done that, but I think it's not possible with CH.

boldtrn:
I just updated this PR to allow configuring blocked areas for CH using the config.properties, right now I added it only for rectangular areas, but we can add circular areas if we want to follow the approach?

karussell:
See my new branch: https://github.com/graphhopper/graphhopper/compare/issue_890

I've merged the branch change-graph-api into these changes, renamed the GraphBrowser class to GraphEdgeIdFinder (?) and also removed a few API options like blocking by edges (as we should not now expose edges to the outside - if necessary, we should consider traversal IDs). Also searching by lat,lon is very similar to lat,lon,radius so I merged them into one option and use ';' for 'object'-separation.

Also I used a simpler rectangular definition (lat1,lon1,lat2,lon2) - does this makes sense to you? Definitely not GeoJSON but our `point` is also not geojson and this definition is consistent to point IMO :)

Also I've added a test for small areas and larger areas to trigger 'shape search' and 'edgeId search'.

karussell:
Several classes and methods needs more javadocs according to #891 and also we should add a simple UI (ctrl+drag?) to block one area (+force ch.disable=true)?

> "Fake Graph that returns a Node Acess ..."

Can we avoid this somehow? E.g. create a graph with some content instead of these many source code lines?

karussell:
With the change-graph-api branch merged in we have now three possibility to block an area

 * after the import (suitable for CH too)
 * via separate Data API for permanent changes (suitable for traffic changes that apply to all requests)
 * via Routing API parameter `block_area`(suitable for individual trials)

boldtrn:
Thanks @karussell. Really good improvements.

>  renamed the GraphBrowser class to GraphEdgeIdFinder (?)

I wasn't happy with the name GraphBrowser either. Not sure about the naming here. This class does several things right now:
- parsing blocked areas from hints
- searching Edge Ids

> Also searching by lat,lon is very similar to lat,lon,radius so I merged them into one option and use ';' for 'object'-separation.

Brilliant idea!

> Also I used a simpler rectangular definition (lat1,lon1,lat2,lon2) - does this makes sense to you?

Yes. I am personally used to the left,bottom,right,top scheme, but the minLat,minLon,maxLat,maxLon is also a standard way to define a BBox (and I think this one is the default for GraphHopper?).

Will work on some of the open points now :).

boldtrn:
I just saw that the reader-overlay module was removed from the pom. Is this intended?

boldtrn:
I added a simple UI to show the rectangle blocking feature. I set it up that one can use the right mouse button to set one of the two Points of the blocking BBox. For actual production use, we probably want a more sophisticated approach? I think for now this should be ok?

Maybe in a future version we might want add something like [Leaflet.draw](https://github.com/Leaflet/Leaflet.draw). Not sure if we want to go down that way?

karussell:
> I just saw that the reader-overlay module was removed from the pom. Is this intended?

Yes, moved most of the stuff into the core and created a new json module where some minor classes remain

> translations are missing :)

Would you mind to add the English & German ones to the spreadsheet? The rest will be adopted often very fast from the community :)

> I added a simple UI to show the rectangle blocking feature.

:+1: nice. Will have a try!

> An obvious limitation right now is that you can only create one rectangular blocking area, no circular, no multiple areas

That is sufficient for now IMO

> Init the blocked polygon from the parameter to the map

What do you mean here?

boldtrn:
> What do you mean here?

I just pushed that change :) Basically if you have a blocked rectangle in your request url, it will be shown on the map. Otherwise you see that the route is taking a detour, but you cannot see why.

> Would you mind to add the English & German ones to the spreadsheet?

Sure, no problem. Wanted to wait until you approve the changes, before changing the spreadsheet.

karussell:
Hmmh, this somehow does not work (yet) :) ... but maybe leaflet draw is indeed better suited for this as we can avoid that many clicks.

karussell:
Sorry, vehicle=generic wasn't set. This is a bit ugly, hmmh

karussell:
ok, let us move the UI into a separate issue. This one is already relative big :)

karussell:
Merged via https://github.com/graphhopper/graphhopper/commit/78d19a80327b35f2f3fe419b4e4d0f4d2d94b9b6

Thanks!

PR #893, opened 2016-12-02
Jaro Winkler Point Hint Improvement

boldtrn:
Switched from Levenshtein to Jaro Winkler for NameSimilarityEdgeFilter.


core/src/test/java/com/graphhopper/routing/util/NameSimilarityEdgeFilterTest.java
```diff
@@ -69,9 +69,88 @@ public void testAccept() {
...
+
+
+
+
+
+
     }
 
-    @Test
+    @Ignore
```
boldtrn: The problem with this PR are these examples. I used Google Maps to get Addresses and compared them to the OSM Address. The differences are very big sometimes.

core/src/test/java/com/graphhopper/routing/util/NameSimilarityEdgeFilterTest.java
```diff
@@ -69,9 +69,88 @@ public void testAccept() {
...
+    public void testThatShouldSucceed(){
+        EdgeFilter edgeFilter;
+        EdgeIteratorState edge;
+
+        // The Problem is that Rd vs Road is abreviated, if we have Road, it works
+        edgeFilter = new NameSimilarityEdgeFilter(new DefaultEdgeFilter(new CarFlagEncoder()), "257 Main Rd, Claremont, Cape Town, 7708, Afrique du Sud");
+        edge = createTestEdgeIterator("Main Road");
+        assertTrue(edgeFilter.accept(edge));
+
```
boldtrn: This example could be fixed by a dictionary that matches Rd to Road and S to South

core/src/test/java/com/graphhopper/routing/util/NameSimilarityEdgeFilterTest.java
```diff
@@ -69,9 +69,88 @@ public void testAccept() {
...
+        // Just too much difference Between Google Maps and OSM @ 32.121435,-110.857969
+        edgeFilter = new NameSimilarityEdgeFilter(new DefaultEdgeFilter(new CarFlagEncoder()), "7202 S Wilmot Rd, Tucson, AZ 85701");
+        edge = createTestEdgeIterator("South Wilmot Road");
+        assertTrue(edgeFilter.accept(edge));
+
+        // @ 37.307774,13.581259
+        edgeFilter = new NameSimilarityEdgeFilter(new DefaultEdgeFilter(new CarFlagEncoder()), "Via Manzoni, 50/52, 92100 Agrigento AG, Italy");
+        edge = createTestEdgeIterator("Via Alessandro Manzoni");
+        assertTrue(edgeFilter.accept(edge));
+
```
boldtrn: This shows the problem with the dictionary approach. Av in Spain means Avenida, whereas in the US it would be Avenue.

core/pom.xml
```diff
@@ -81,6 +81,13 @@
             <version>${json.org.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>info.debatty</groupId>
+            <artifactId>java-string-similarity</artifactId>
```
karussell: I see a small disadvantages for e.g. iOS to have this in a separate jar just for the sake of one class. Can we just copy this class (and remove the two interfaces) to a package like we did for apache commons (plus add this in the NOTICE for proper attribution)? Also then we could remove the Levenshtein class?

core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -19,85 +19,183 @@
...
+import info.debatty.java.stringsimilarity.JaroWinkler;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.regex.Pattern;
 
 /**
- * Abstract Class that defines the basis for NameSimilarity matching using an EdgeFilter.
+ * This Class defines the basis for NameSimilarity matching using an EdgeFilter.
```
karussell: Would you add some more explanation here? E.g. *to snap coordinates not necessarily to its closest edge but instead pick an edge with a similar name we need to define similary and this happens here*

minor typo: Class->class

core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -19,85 +19,183 @@
...
-    private static final double LEVENSHTEIN_ACCEPT_FACTOR = .05;
-
+    private static final Pattern NON_WORD_CHAR = Pattern.compile("[^\\p{L}]+");
     private final EdgeFilter edgeFilter;
-    private final String soughtName;
-    private static final Pattern nonWordCharacter = Pattern.compile("[^\\p{L}]+");
+    private final List<String> pointHint;
 
-    public NameSimilarityEdgeFilter(EdgeFilter edgeFilter, String soughtName) {
+    private JaroWinkler jw = new JaroWinkler();
```
karussell: Can be final

core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -19,85 +19,183 @@
...
     @Override
     public final boolean accept(EdgeIteratorState iter) {
-        if(!edgeFilter.accept(iter)){
+        if (!edgeFilter.accept(iter)) {
             return false;
         }
 
-        // Don't check if PointHint is empty anyway
-        if(soughtName.isEmpty()){
+        // Don't check if point hint is empty anyway
```
karussell: this comment is not necessary IMO

core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -19,85 +19,183 @@
...
-                return name.equals(soughtName);
+        List<String> edgeName = prepareName(name);
+
+        // TODO Splitting Strings to merge them again is not very elegent, maybe we rather want to keep the OriginalPointHint and the OriginalEdgeName?
+        if(isJaroWinklerSimilar(listToString(pointHint), listToString(edgeName), .8))
+            return true;
+
+        return false;
+
+        /*
```
karussell: please remove the commented code, should be in the previous commit

core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -19,85 +19,183 @@
...
+        // System.out.println(hint + " vs. edge:" + name + ", " + levDistance + " <= " + factor);
+        return levDistance <= factor;
     }
 
-    private boolean isLevenshteinSimilar(String name){
-        int perfectDistance = (int) (Math.abs(soughtName.length()-name.length()) + Math.ceil(soughtName.length()*LEVENSHTEIN_ACCEPT_FACTOR));
-        int levDistance = StringUtils.getLevenshteinDistance(soughtName, name);
-        return levDistance <= perfectDistance;
+
+    public static String longestSubstring(String str1, String str2) {
```
karussell: let us remove this method here too :)

core/src/test/java/com/graphhopper/routing/util/NameSimilarityEdgeFilterTest.java
```diff
@@ -69,9 +69,88 @@ public void testAccept() {
...
+        edgeFilter = new NameSimilarityEdgeFilter(new DefaultEdgeFilter(new CarFlagEncoder()), "Av. Juan Ramón Ramírez, 12, 02630 La Roda, Albacete, Spain");
+        edge = createTestEdgeIterator("Avenida Juan Ramón Ramírez");
+        assertTrue(edgeFilter.accept(edge));
+
+
+    }
+
+
+    /**
+     * We ignore Typos for now, most GeoCoders return pretty good results, we might allow some typos
```
karussell: Do you mean here 'we might allow some typos **in the future**'?

core/pom.xml
boldtrn: Yes sure, if we want to merge this into master. I haven't done this, yet, as I considered the branch to be a bit "experimental - playing around with JaroWinkler".

core/src/test/java/com/graphhopper/routing/util/NameSimilarityEdgeFilterTest.java
boldtrn: I'd love to allow Typos, or speaking more general, allow different string lengths and differences of the common substring, without allowing too many false positives. I think, we want to match different Geocoder sources, for now. Maybe the Osm data contains a Typo, that is not present in the Geocoder source or maybe we want to match input typed in by hand. More realistic examples are in the method above. That's why I have not removed that test for now, we can remove it, if you think we don't need it anyway.

core/pom.xml
karussell: Ah, ok. I thought JaroWinkler is for the cases from real world always an improvement and we should replace the current approach (?)

core/src/test/java/com/graphhopper/routing/util/NameSimilarityEdgeFilterTest.java
karussell: Hmmh, not sure. Maybe keep it and so we can easily enable it if we have a new approach or similar.

core/src/main/java/com/graphhopper/debatty/java/stringsimilarity/JaroWinkler.java
```diff
@@ -0,0 +1,144 @@
+package com.graphhopper.debatty.java.stringsimilarity;
+
+import java.util.Arrays;
+
+/**
+ * This class is copied from: https://github.com/tdebatty/java-string-similarity/blob/master/src/main/java/info/debatty/java/stringsimilarity/JaroWinkler.java
+ * and slightly modified. *
```
karussell: Please [add the license](https://github.com/tdebatty/java-string-similarity/blob/master/LICENSE.md) directly to this file, similar to the Apache license headering.

core/src/main/java/com/graphhopper/debatty/java/stringsimilarity/JaroWinkler.java
```diff
@@ -0,0 +1,144 @@
...
+     */
+    public final double getThreshold() {
+        return threshold;
+    }
+
+    /**
+     * Compute JW similarity.
+     * @param s1
+     * @param s2
+     * @return
```
karussell: Please remove undocumented parameters and return values or add text to it.

core/src/main/java/com/graphhopper/debatty/java/stringsimilarity/JaroWinkler.java
```diff
@@ -0,0 +1,144 @@
...
+        return jw;
+    }
+
+
+    /**
+     * Return 1 - similarity.
+     * @param s1
+     * @param s2
+     * @return
+     */
```
karussell: Same here.

core/src/main/java/com/graphhopper/debatty/java/stringsimilarity/JaroWinkler.java
boldtrn: A, completely overlooked that, thanks!

PR #894, opened 2016-12-02
Measurement UI

karussell:
With this simple measurement.html you can evaluate output from `./graphhopper.sh measurement your.pbf`

Related https://github.com/graphhopper/map-matching/pull/73


kodonnell:
Some comments:

- this only works when running server and browser on same host
- it's, um, not pretty = ) Also got a few problems with e.g. the graphs.
- as far as I can tell, you don't need a server (and if you did you could use e.g. python -m SimpleHTTPServer or similar too, which do the same thing). E.g. you could just have a single text file which contains a list of all measurements (which can be e.g. recreated with a simple script or at the end of `./graphhopper measurement ...`). Then you fetch that file, then fetch all the other files defined in it. It's not super pretty, but it removes a dependency on e.g. npm, and is of similar prettiness to using http-server as a directory browser. (To be fair, I have to admit to having used this hack before too.)
- if you're using node/npm, why not get it to manage your dependencies (jquery etc.) instead of including compiled code in git? Graphhopper is already a pretty slow repo to clone. You could then use nice npm stuff like run scripts, etc. Even better, you could write an actual app which takes care of the yucky stuff (e.g. finding measurement files, stripping out the relevant stuff, returning as JSON, etc.). Same applies for other web apps.
- the overall workflow seems a bit clunky ... the user has to manually run all the measurements as text files, and then run this separately. The timelines don't make sense (the time doesn't actually reflect the time the code changes were made, and hence they have to hope they ran things in the right order, and didn't have old ones lying around - e.g. from other branches or previous runs - otherwise they'll get misleading results). And nor can they associate the performance with actual codes changes.
- for now, I don't really see what value this offers the user over and above e.g. what's found [here](https://github.com/graphhopper/map-matching/pull/73) - in fact I'd say it offers less, as the aforementioned actually compare commits etc. If it's simply graphs, then it's something pretty easy to do as part of the command line tool. And that saves you all the hassle of a web app and the cruft to support it.

Couple of asides:
- if you could have a single instance which runs measurements for every commit (travis?), then you could serve an entire history, and that'd be useful.
- that made me think - should we create a separate repo for measurements? The problem with the current approach is that the measurement functionality is tied to the commit. So e.g. if we make a change to the measurement code, then we can only use it for commits following it. As an example, there's no (easy) way of testing the performance of commits before the measurement functionality was added. In addition, it'd offer you further functionality, like making the GUI control everything (i.e. the workflow comment above). For example, the GUI could display all the possible commits, and you can select which ones you want to test, hit 'run', and it goes and does the measurements then returns the results.
- can you update the help of `graphhopper.sh` to include all options (including measurement)?

karussell:
> this only works when running server and browser on same host

What do you mean here? You can start the server in the GH root and open measurement.html with the browser directly (no server). Then it should just work

> it's, um, not pretty = ) Also got a few problems with e.g. the graphs

Sure, go ahead for something better ;)

> as far as I can tell, you don't need a server 

You don't need a server if you put the stats in the app itself like with a swing UI. But if you want to display the stats in the browser you cannot load files from local storage so you need to serve them.

Maybe instead of the web server and properties overhead we could visualize the data via a Swing UI directly via freechart like [here](https://github.com/graphhopper/graphhopper/pull/855#issuecomment-261308409).

> if you could have a single instance which runs measurements for every commit (travis?), then you could serve an entire history, and that'd be useful.

That is what we already do [here](https://graphhopper.com/measurement), but travis is not an option as you always run on different machines with different loads.

> As an example, there's no (easy) way of testing the performance of commits before the measurement functionality was added.

That is indeed a problem. But I think if the measurement class stabilizes we could avoid that.

> for now, I don't really see what value this offers the user over and above e.g. what's found here

The advantage is that you can compare many commits for historic purposes.

> The timelines don't make sense

Why do you think so? If you run the measurement roughly at the time of the commit I think this is useful.

> Graphhopper is already a pretty slow repo to clone.

It is improvable, sure. But not the worst. There are integration tests which we can run directly from the repo, but we need the full pbf. Also it will get 'worse' when we include gtfs files. Interesting problem :)

karussell:
> Then you fetch that file, then fetch all the other files defined in it. It's not super pretty, but it removes a dependency on e.g. npm, and is of similar prettiness to using http-server as a directory browser.

I do not understand how this would be an improvement. We already use npm and now we would require a third tool 'python'. I know it is pre-installed on linux&ios but not on windows.

kodonnell:
Hmmm - can you clarify the objective of the measurement suite? I see two specific use cases

- for the developer to use for hunting for improvements, or seeing how their code affects performance, etc. (I.e. they run it on their machine and compare results on their machine.)
- for historical tracking of the entire repo, so that one can see trends or changes that weren't expected (i.e. the developer didn't manually test performance themselves).

I've been focusing on the former - can you clarify?

> What do you mean here

Say I'm on a server, and I run the measurements and then `http-server`. If I then go to my desktop and open it in the browser (e.g. `http://server:8080`) then it'll open the GUI but won't fetch the data as the data URL is hardcoded as `http://localhost:8080` (when it should now point to `http://server:8080`). This seems like it could be a common problem to me - most of my work is done on a server (SSH only so no e.g. browser) and results are viewed from a desktop.

> Sure, go ahead for something better ;)

Fair point! At this stage, it depends on the objective - if it's 'for the developer', then the UI is probably not even required.

> But if you want to display the stats in the browser you cannot load files from local storage so you need to serve them.

Maybe I'm misunderstanding, but I load local files all the time. I do this explicitly at work so non-technical users don't have to e.g. spin up a server to use a web app. That said, there can be some problems with different browsers - off the top of my head Firefox/IE/Safari support it, and Chrome does too (if you allow it). That said, this is only useful in the 'for the developer' case - you wouldn't want to be using it in a production historical tracking app!

> That is what we already do here, but travis is not an option as you always run on different machines with different loads.

Ah, nifty. By 'single instance' I should have said 'single dedicated server'. You can get round the different machine issue (but not different load issue) by rerunning the tests for the entire history, on every single commit. Or just have a dedicated server.

> The advantage is that you can compare many commits for historic purposes.

Can't you do that with both approaches?

> But I think if the measurement class stabilizes we could avoid that.

That's effectively limiting yourself to things which can only be done with the stabilized class ... "we'd like to do X, but we'd lose all of our historic measurements, so we won't". That said, the converse is how you'd test GH separately - if the core is refactored and method X becomes method Y, how would a separate measurement suite know to change from calling X to calling Y? So, maybe not so simple. However, it depends on what the objective is.

> Why do you think so? If you run the measurement roughly at the time of the commit I think this is useful.

Assuming the 'for the developer' use case, then I might get it to run with the last 10 commits. If each commit takes about 1 minute to test, then I'll see a graph with an x range of 11 minutes ... despite the actual commits being days or weeks apart. And, if I'd already run it before (on multiple branches) I might get completely bonkers results. What I'd suggest is

- use the commit date from the commit log
- if a measurement already exists for that commit, append a suffix. Then, in the UI, if multiple are detected for the same commit, this can be used to construct error bars, etc. Unfortunately this means changing the measurement functionality ... which would be bad, as per above.

Two thoughts:

- if the latter, then I'd suggest using a database of some form. Then you get more flexibility as you can e.g. migrate to new schemas. (For example, if wouldn't be too hard to add a commit date to all past measurements, or just set it to null for past ones.)

> You don't need a server if you put the stats in the app itself like with a swing UI ... Maybe instead of the web server and properties overhead we could visualize the data via a Swing UI directly via freechart like here.

I'm not sure what you mean by 'put[ting] the stats in the app itself' - as far as I can see, there has to be measurement data stored somewhere (currently txt files, maybe database as above, etc.), or returned dynamically (as per my comment about a separate repo/tool for evaluating performance).

> Also it will get 'worse' when we include gtfs files. Interesting problem :)

Usually, for something simple (or non-standard), I take the approach of only including source code in git, and having e.g. a 'setup.sh' script which gets all my dependencies first. (This is especially valid for data, e.g. GTFS feeds - though dat is maybe going to change that.) However, most tools already do this for you (e.g. maven, npm, etc.).

> we need the full pbf

You could simply have a `if [ ! -f blah.pdf ]; wget ...; fi` at the start of `graphhopper.sh`. That's (effectively) the approach I usually take. Likewise you could even do the same within test suites run from java.

(As a side thought - while some resources are available online anyway, you could always create a 'graphhopper-resources' repo, in which you put the data, and that way you can pull individual files - possibly from rawgit - of the correct version) only when you need them. And the main git repo stays fast and snappy and easier to use (e.g. diffs only display code diffs, not data diffs).

> I do not understand how this would be an improvement. We already use npm and now we would require a third tool 'python'. I know it is pre-installed on linux&ios but not on windows.

Did you quote the wrong selection? By using local files (as you quoted), there's no need for npm or node or python, and is the most portable solution of all.

karussell:
> I've been focusing on the former - can you clarify?

I would like to have one tool for both. Or why do you think we need two solutions?

> This seems like it could be a common problem to me

To access localhost or whatever host can be made configurable, e.g. via URL parameter or serving the measurement.html also via the server.

> i.e. the developer didn't manually test performance themselves

Tracking changes over time also is important to find issues when merging branches etc. It is 'perf testing' like we do unit and integration tests.

> And the main git repo stays fast and snappy and easier to use

You mean GH repo is hard to use? I never ever had a performance problem to be honest even with my slow connection. The initial clone takes a bit, true. But afterwards it is fine. Also I do not understand how the separate repos make something better: at some point you need to keep those repos in-sync. What if you switch the branch etc? I think it is not that easy :)

> By using local files (as you quoted), there's no need for npm or node or python, and is the most portable solution of all

Which is the most portable solution? Loading the files and put it in a java visualization? I agree but I think html&js is better because then you can use the UI also for the second purpose.

> I'm not sure what you mean by 'put[ting] the stats in the app itself' - as far as I can see, there has to be measurement data stored somewhere 

What I meant is to run the measurement suite and create some in-memory data (or load local measurement files) and display it directly in Java (like with jfreechart)

> Maybe I'm misunderstanding, but I load local files all the time. 

Do you mean I need to click a button and select the files I want? I would like to have something simpler where I just open the file or refresh the site to see updates and that is only possible when serving the measurements files via a server if you want to go with html+js and a standard browser IMO. If not, I'm happy to learn stuff :) !

> What I'd suggest is: use the commit date from the commit log

:+1: 

> if the latter, then I'd suggest using a database of some form. Then you get more flexibility as you can e.g. migrate to new schemas. (For example, if wouldn't be too hard to add a commit date to all past measurements, or just set it to null for past ones.)

I'm not a big fan of databases :) also I do not see how this is not possible with properties files - using defaults if something is missing is easy there too. But again: I just did what suited my needs and I was hoping it could solve not only the historic tracking (2) but also the commit comparison (1) problem one has. 

And also: I'm open towards something more powerful.

kodonnell:
> Or why do you think we need two solutions?

We just have to be careful. E.g. for historical tracking, we'll have millions of measurements, and databases make sense. For local stuff, users don't want to have to fluff round with installing databases, etc., so simple approaches (as currently) make sense. I'll have a think.

> Tracking changes over time also is important

Definitely agree.

> You mean GH repo is hard to use?

Not exactly. However, it's good practice to separate the code that does stuff from dependencies (e.g. other code, or data the code can use). For me, the main reason is to make it clear to the developer what is and what isn't the relevant code (and changes). For example, if I want to see what you've changed with this PR, if I do a `git diff` I get a massive dump about (minified) jquery diffs etc. Those aren't actually changes related to GH (the only change should be version X -> version Y, and not the code for either). You're relying on me figuring this out - if I didn't know better, I'd have to assume that the change in the jquery code was something I had to understand to fully understand your commit. And that would not be fun!

> Also I do not understand how the separate repos make something better: at some point you need to keep those repos in-sync. What if you switch the branch etc? I think it is not that easy :)

To be clear, I'm merely talking about a separate repo as a convenient way of hosting/versioning source data. In other words, a data dump. For example, we currently have [this](https://github.com/graphhopper/map-matching/blob/master/map-data/leipzig_germany.osm.pbf) file in the `map-matching` repo. What I propose to do is put it in a separate repo such as `graphhopper-resources`. Then, in the `map-matching` repo, I'd simply download the file when needed, e.g. `https://github.com/graphhopper/graphhopper-resources/blob/cd388f33bb997031a966bba7266a64dba85dc93a/map-data/leipzig_germany.osm.pbf`. Keeping them in sync is easy - if I want to add new data, then I submit a PR in both. This should happen so rarely that it shouldn't be an issue.

> What I meant is to run the measurement suite and create some in-memory data

Cool, that's what I meant too.

> Which is the most portable solution? Loading the files and put it in a java visualization? I agree but I think html&js is better because then you can use the UI also for the second purpose.

> Do you mean I need to click a button and select the files I want? I would like to have something simpler where I just open the file or refresh the site to see updates and that is only possible when serving the measurements files via a server if you want to go with html+js and a standard browser IMO. If not, I'm happy to learn stuff :) !

Learning time = ) Whack this into an index.html, then open with Firefox.

```
<!doctype html><html><head><meta charset="utf-8"></head><body>
<script src="https://d3js.org/d3.v4.min.js"></script><script>
d3.text("./measurements.txt", function(err, txt) {
    console.log(err ? err : txt)
})
</script></body></html>
```
Console should display an error. Now create a file next to index.html called 'measurements.txt' and add some content. Reload, and the console should show the content.

Note:

- I'm using d3 as it's nice and simple. But it's just a wrapper around XMLHttpRequest.
- iif we're looking for a more-than-developer solution, then I don't think this is the best idea.

> I'm open towards something more powerful.

I'll have a think, but at this stage, I consider other stuff more important (inc. map matching), so will probably work on that.

karussell:
> Learning time = ) Whack this into an index.html, then open with Firefox.

Thanks & works perfectly!

> iif we're looking for a more-than-developer solution, then I don't think this is the best idea.

I only have the dev in mind. Who else should be interested in this?

> Then, in the map-matching repo, I'd simply download the file when needed

That won't happen I think. Because this makes everything more complex and I do not see any benefits. If you download the git repo and you want to contribute you need to run the tests so you need to download the data. I do not see a problem with the current approach nor an advantage with the proposed one. Using npm instead of putting the js files in the repo is of course something we should do.

Furthermore when I picked the pbfs and osm files I took care that this is only as few as possible coverage and I could reuse old areas for new cases.

kodonnell:
> I only have the dev in mind. Who else should be interested in this?

Sorry, I was more meaning that the dev is only likely to use this for a few measurements, so it suits. (And it's easy for them to understand/develop it.) However, if you want a history server with thousands of measurements, I don't think this will suit - partly as you'll be doing thousands of HTTP requests (though I guess you could dynamically fetch them depending on the amount of history they want to view), but also because I think loading local files like this is generally frowned upon for production stuff.

> That won't happen I think.

That's fine with me.

> Using npm instead

Great. Once the deliverable has stabilised (i.e. we know we're not doing a Java swing GUI or some such) I can help with this part if you want.

kodonnell:
> I'll have a think ...

OK, my current thoughts are thus:

- I'm guessing that the measurement suite is unlikely to be used by most contributors, who will be focusing on bug fixes and features as opposed to performance. So, maybe, in the interests of keeping the code base clean (and [not needing to maintain it](https://github.com/graphhopper/map-matching/pull/73#issuecomment-264818693) etc.), we remove it from the main repo?
- the main use (?) is likely to be by core GH contributors when evaluating changes. For this reason, it would seem good if we can make the tests run by default (if they don't already) on every travis build and e.g. cover the last 10 commits - and hope resource doesn't change dramatically on the server while doing them. (See [here](https://github.com/travis-ci/travis-ci/issues/352) for a good discussion.) Then that information is readily available for every commit, and no one needs to run anything manually. This means we're back to having the measurement suite in the code base - unless we can trigger [dependent builds](http://eng.rightscale.com/2015/04/27/dependent-builds-in-travis.html). Note:
    -  we'd want to tidy it up e.g. as in [here](https://github.com/graphhopper/map-matching/pull/73) to make it a bit easier to digest.
    - assuming travis supports caching (?) we can speed it up considerably by [caching the builds](https://github.com/kodonnell/map-matching/blob/9d7865733f30a7978b2d886c8b9afbb4832b3cdd/map-matching.sh#L46). Faster means less exposure to changing resource on the builder.
- there is a secondary use case for historical tracking of performance so e.g. we can easily see "hang on, X has gradually got slower over the last two years - why is that?". It looks like lots of people have this wish (see linked discussion above) - and there are even a few 'solutions' e.g. http://openbenchmarking.org. We could roll our own history-server and GUI ... but if we can use free tools already designed for that purpose, I suspect we'd end up with a better result (and much less maintenance overhead, etc.)

Currently I'm leaning toward probably keeping the measurement stuff in this repo, and just running the measurement suite for e.g. the last 10 commits, as part of Travis. If we can easily use e.g. openbenchmarking, then great - otherwise I'm less inclined to worry about historical tracking (for now).

Thoughts @karussell?

Issue #280, opened 2014-11-04
bicycle routing support for cycleway:left and cycleway:right tag


ratrun:
BikeCommon Flag encoder should add support for the tags "cycleway:left" and "cycleway:right", see http://wiki.openstreetmap.org/wiki/Key:cycleway.

karussell:
Hmmh, the problem is that left and right is country dependent if we want to find out the oneway stuff, right?

karussell:
Or is it identical to the :forward and :backward stuff? #205

ratrun:
I think that the oneway aspect of this issue (values=opposite*) is identical to #205. After making some overpass queries I found out that the usage of #205 (bicycle:backward) is much more common compared to the cycleway:left or  cycleway:right usage with "opposite". I cannot see any country dependent aspect. It does not matter if the car traffic runs on the left or on the right side. 

Besides the oneway aspect of this issue there is the priority handling aspect, which eventually should be handled differently for every kind of bike. Here I see some relation with one of the aspects of #330, where the highway classification tag, if present, should have some influence on the priorisation. E.g. highway=primary together with some cycleway "lane, track ..." tag should not lead to increased preference, especially when comparing to an ordinary highway=residential or highway=unclassified without any cycleway. The same is true if the cycleway is running opposite to a oneway. We should not prefer such a cycleway. I quickly checked the code, I could not find any part where we are rating down the priority in these cases. I know people, which claim to have statistics, which say that these kind of cycleways are extremly dangerous for a cyclist. I believe that they might be right.

karussell:
> It does not matter if the car traffic runs on the left or on the right side. 

Okay, my bad, so it is entirely dependent on the 'tagging direction'

I've fixed #205, will include bicycle:backward now and handle the prioritization after 0.4

karussell:
Hmmh, according to the [wiki](https://wiki.openstreetmap.org/wiki/Key:access#One-way_restrictions) the bicycle:backward has a different meaning.

karussell:
Maybe you have some direct link with current wrong behaviour? (regarding access restriction)

BTW: The cycleway:left/right is not that rare ([see here](http://taginfo.openstreetmap.org/keys/cycleway:right#values)).

ratrun:
[...]

PR #901, opened 2016-12-06
Accept keys "cycleway:left" and "cycleway:right"  with "opposite" values

ratrun:
This change implements access issue for bicycles as discussed in #280.


karussell:
Thanks!

PR #906, opened 2016-12-09
CH-Preparation of Generic Vehicle

boldtrn:
We introduced a small bug with the Generic weighting. When we pass a null Graph on initialization, we try to get the NodeAccess, which results in a null-pointer exception.


karussell:
Thanks!

PR #908, opened 2016-12-13
Update Android build tools

devemux86:


karussell:
Thanks again - BTW: not sure if it is really necessary to upgrade to minor version increases. Maybe to save us some work we do this only once in a while or before a release?

devemux86:
I have thought of that, it's just that Android Studio + Gradle progress rapidly and often require latest versions to play nice.
Still GraphHopper sample is not touched often, so we can change the updates frequency if you want.

Note that I push only the stable version increases (as in Mapsforge & VTM), there are many more in the other development channels!

Issue #885, opened 2016-11-23
bug in enforceHeadingByEdgeId ?


kodonnell:
```
@param incoming if true, incoming edge is unfavored, else outgoing edge
```

Am I correct in that this is not true? As far as I can see, `incoming` is not used, and both the incoming and outgoing edge are unfavoured.

Use is for [map-matching issue 70](https://github.com/graphhopper/map-matching/issues/70)

karussell:
Looks like this is indeed not implemented. We need a unit test for this :)

kodonnell:
> We need a unit test for this :)

I can do a basic one to check it e.g. sets the right edge to unfavoured, but I don't think I understand it well enough to test it while routing. Is that OK?

karussell:
> I don't think I understand it well enough to test it while routing

Please keep asking :) !

See e.g. FastestWeightingTest.testWeightWrongHeading, QueryGraphTest.testEnforceHeading, GraphHopperIT.testMonacoStraightVia, GraphHopperOSMTest.testGetPathsDirectionEnforcement*

kodonnell:
Hmmm, I think I/we may have misinterpreted the intention of this, and that it is only mean to be used at the *start/end* of a route. In in the case 

``` 
     v1    v2 
 A------X------B
```

with nodes A,B and virtual node X, and virtual edges v1 and v2, we'd call enforceHeadingByEdgeId(X, v1, ...) to enforce a heading *left to right* (i.e. X->B). This works by making v1 unfavored *in both directions* and hence v2 is the only option. If we wanted to enforce the reverse direction (X->A), then we'd use enforceHeadingByEdgeId(X, v2, ...). This is the current behaviour and I believe how it is currently used. Therefore I think the `incoming` parameter is completely redundant and can be removed. It might also pay to rename it to something like `unfavorBothDirectionsOnVirtualEdge`, which more clearly reflects its purpose.

The only case where the `incoming` parameter could be useful is if we want to enforce a heading *along* a route, in which case we'd then have to unfavor the same direction on both v1 and v2 (instead of both directions on only one).

Thoughts @karussell?

Cc @stefanholder

karussell:
Yes, this is ugly. In GraphHopper the graph is bidirectional but for every virtual node we create for virtual edges to replace the single real edge. Also we are using this method only via incoming=false

So either we make this method work more generic or refactor it, yes.

stefanholder:
[...]

PR #909, opened 2016-12-13
Refactor unfavoring of virtual edges (#885)

stefanholder:
This addresses issue #885.

Also needed for graphhopper/map-matching#70

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/QueryGraph.java
+++ b/core/src/main/java/com/graphhopper/routing/QueryGraph.java
@@ -141,7 +141,10 @@ public class QueryGraph implements Graph {
             return getElevation(nodeId);
         }
     };
-    private List<VirtualEdgeIteratorState> modifiedEdges = new ArrayList<VirtualEdgeIteratorState>(5);
+
+    // Use LinkedHashSet for predictable iteration order.
+    private final Set<VirtualEdgeIteratorState> unfavoredEdges = new LinkedHashSet<>(5);
+
     private boolean useEdgeExplorerCache = false;
 
     public QueryGraph(Graph graph) {
@@ -407,7 +410,7 @@ public class QueryGraph implements Graph {
 
     /**
      * Set those edges at the virtual node (nodeId) to 'unfavored' that require at least a turn of
-     * 100° from favoredHeading
+     * 100° from favoredHeading.
      * <p>
      *
      * @param nodeId         VirtualNode at which edges get unfavored
@@ -451,11 +454,11 @@ public class QueryGraph implements Graph {
             if (Math.abs(delta) > 1.74) // penalize if a turn of more than 100°
             {
                 edge.setUnfavored(true);
-                modifiedEdges.add(edge);
+                unfavoredEdges.add(edge);
                 //also apply to opposite edge for reverse routing
                 VirtualEdgeIteratorState reverseEdge = virtualEdges.get(virtNodeIDintern * 4 + getPosOfReverseEdge(edgePos));
                 reverseEdge.setUnfavored(true);
-                modifiedEdges.add(reverseEdge);
+                unfavoredEdges.add(reverseEdge);
                 enforcementOccurred = true;
             }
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/graphhopper/routing/QueryGraph.java
```diff
@@ -464,35 +464,43 @@ public boolean enforceHeading(int nodeId, double favoredHeading, boolean incomin
...
-     * @param edgeId   edge to become unfavored
-     * @param incoming if true, incoming edge is unfavored, else outgoing edge
-     * @return boolean indicating if enforcement took place
+     * @param virtualNodeId  virtual node at which edges get unfavored
+     * @param virtualEdgeId  this edge and the reverse virtual edge become unfavored
      */
-    public boolean enforceHeadingByEdgeId(int nodeId, int edgeId, boolean incoming) {
-        if (!isVirtualNode(nodeId))
-            return false;
+    public void unfavorVirtualEdgePair(int virtualNodeId, int virtualEdgeId) {
```
stefanholder: The intention was to make the semantics of enforceHeadingByEdgeId more clear by renaming this method to unfavorVirtualEdgePair.

core/src/main/java/com/graphhopper/routing/QueryGraph.java
```diff
@@ -464,35 +464,43 @@ public boolean enforceHeading(int nodeId, double favoredHeading, boolean incomin
...
 
     /**
      * Removes the 'unfavored' status of all virtual edges.
      */
     public void clearUnfavoredStatus() {
-        for (VirtualEdgeIteratorState edge : modifiedEdges) {
+        for (VirtualEdgeIteratorState edge : unfavoredEdges) {
             edge.setUnfavored(false);
         }
+        unfavoredEdges.clear();
```
stefanholder: Fixed bug.

core/src/main/java/com/graphhopper/routing/QueryGraph.java
```diff
@@ -464,35 +464,43 @@ public boolean enforceHeading(int nodeId, double favoredHeading, boolean incomin
...
         incomingEdge.setUnfavored(true);
-        modifiedEdges.add(incomingEdge);
+        unfavoredEdges.add(incomingEdge);
         reverseEdge.setUnfavored(true);
-        modifiedEdges.add(reverseEdge);
-        return true;
+        unfavoredEdges.add(reverseEdge);
+    }
+
+    public List<VirtualEdgeIteratorState> getUnfavoredVirtualEdges() {
```
stefanholder: Needed for graphhopper/map-matching#70

core/src/main/java/com/graphhopper/routing/QueryGraph.java
karussell: Do you have a test case showing this bug?

core/src/main/java/com/graphhopper/routing/QueryGraph.java
karussell: Can we reduce this to `List<EdgeIteratorState>` or is the implementation necessary?

core/src/main/java/com/graphhopper/routing/QueryGraph.java
stefanholder: The bug was that multiple invocations of `enforceHeadingByEdgeId` added more and more edges to `modifiedEdges`. `clearUnfavoredStatus()` would then go over all previous edges again although their unfavored status had already been cleared.

However, this bug was not observable before because modifiedEdges could not be accessed from outside this class.

core/src/main/java/com/graphhopper/routing/QueryGraph.java
stefanholder: The implementation class should not be necessary for graphhopper/map-matching#70.

One thing I didn't like about `VirtualEdgeIteratorState` is that `detach()` is not supported. If we return `List<VirtualEdgeIteratorState>` here, the user can at least figure out if it is supported. (Unfortunately, he would have to figure this out from the code since it is not explained in the javadoc.)

karussell: Then I think we should still use EdgeIteratorState and if people needs detach we have to implement this :) (in another issue). The less stuff is public the better for refactoring later on.

stefanholder:
>Can we reduce this to `List<EdgeIteratorState>`?

Done.

>Would you mind adding a Testcase that shows the issue described in #885 and that it is solved by your PR?

Done.

core/src/main/java/com/graphhopper/routing/QueryGraph.java
```diff
@@ -141,7 +141,10 @@ public double getEle(int nodeId) {
             return getElevation(nodeId);
         }
     };
-    private List<VirtualEdgeIteratorState> modifiedEdges = new ArrayList<VirtualEdgeIteratorState>(5);
+
+    // Use LinkedHashSet for predictable iteration order.
+    private final Set<VirtualEdgeIteratorState> unfavoredEdges = new LinkedHashSet<>(5);
```
stefanholder: Now using Set instead of List to store unfavoredEdges.

karussell:
@boldtrn please confirm & merge (as here just one author it is safe to 'squash and merge')
@kodonnell can we close #885 then?

stefanholder:
>@boldtrn please confirm & merge (as here just one author it is safe to 'squash and merge')

If both commits are squashed then the test case showing the issue described in #885 is lost.

karussell:
Hmmh, why isn't it added to the latest commit?

stefanholder:
>Hmmh, why isn't it added to the latest commit?

Simply speaking, because adding something and deleting the same in a single commit results in no change. The test case in the first commit tests `enforceHeadingByEdgeId` as it is described in the original javadoc of this method and hence reveals the problems with this method. The second commit changes the contract of the method and the unit tests to match its behavior.

karussell:
Ok, then you mean 'lost' in the sense for historical reasons. Makes sense. @boldtrn

kodonnell:
> @kodonnell can we close #885 then?

Sure. Happy for you to close it when merged, or I can.

core/src/test/java/com/graphhopper/routing/QueryGraphTest.java
```diff
@@ -613,21 +613,46 @@ public void testEnforceHeadingByEdgeId() {
...
         // enforce coming in north
-        queryGraph.enforceHeadingByEdgeId(2, 1, false);
+        queryGraph.enforceHeadingByEdgeId(2, 1, true);
         // test penalized south
-        boolean expect = true;
         VirtualEdgeIteratorState incomingEdge = (VirtualEdgeIteratorState) queryGraph.getEdgeIteratorState(1, 2);
+        VirtualEdgeIteratorState incomingEdgeReverse = (VirtualEdgeIteratorState) queryGraph.getEdgeIteratorState(1, incomingEdge.getBaseNode());
+        boolean expect = true;  // expect incoming edge to be avoided
+        assertEquals(expect, incomingEdge.getBool(EdgeIteratorState.K_UNFAVORED_EDGE, !expect));
+        expect = false;         // expect reverse incoming edge not to be avoided
```
boldtrn: Why not remove `expect = false`. That way it would fail before your PR => Showing the issue. Then you add your PR and therefore fix the failing test case. The order is not important at this point :).

boldtrn:
@stefanholder Thanks for adding a test.

> If both commits are squashed then the test case showing the issue described in #885 is lost.

The reason why I asked you to add a test was to document the issue and make sure that we don't add a similar issue by mistake in the future. See my [comment](https://github.com/graphhopper/graphhopper/pull/909/files/ac385b6918b9001f9ce313f5270ea72ea86506f1#diff-b023b2a2f1f8a34fc135e3905e7b86d3). But maybe I am overthinking this issue? I will approve my review, that's probably somethin @karussell has to decide.

stefanholder:
>as here just one author it is safe to 'squash and merge'

@karussell, squashing a PR via the github UI always changes the author to the person that does the squash merging. I just verified this with a test squash. The reason behind this is that github uses the `git merge --squash` to create the squashed changes and then commits them under the maintainer's name. (See also graphhopper/map-matching#83 for the previous discussion on this topic.)

Note that it's possible to [disable squash merging](https://github.com/blog/2141-squash-your-commits) in the github repository options. Squashing without changing authorship can be done manually using `git rebase -i`.

boldtrn:
Merged, thanks @stefanholder. I will create another PR with your test scenario, to have it in the repo, hope this is OK?

karussell:
> @karussell, squashing a PR via the github UI always changes the author to the person that does the squash merging. I just verified this with a test squash

The author will be still the original author (see 'git log'), still github will show the one who merged it.

karussell:
> Note that it's possible to disable squash merging in the github repository options.

Sure, still this is useful e.g. if the pull requester does not know about this and the PR contains several intermediate commits.

For the future I would suggest the following (where we can disable squashing): we merge the PR even if some commits are 'superfluous' but ask the contributor to squash as much as possible.

WDYT?

> Squashing without changing authorship can be done manually using git rebase -i.

From my experiences with github this has the same effect to the author entry as doing the squash in the UI

devemux86:
> The author will be still the original author (see 'git log'), still github will show the one who merged it.

Doesn't GitHub shows both?
Commits [history](https://github.com/graphhopper/graphhopper/commits/master) has the author's picture (with commiter's smaller picture).
And the commits mention "<author> commited with <committer>".

karussell:
> Doesn't GitHub show both?

That was what I meant. In the git log you only see the original author.

boldtrn:
@stefanholder I hope it was ok that I squashed the PR? Do you think we should add the test in the commit history?

stefanholder:
>@stefanholder I hope it was ok that I squashed the PR? Do you think we should add the test in the commit history?

Yes, it's OK. I think it would have been better to merge both commits so that the failing test would still be in the git history but this is a minor thing.

stefanholder:
>For the future I would suggest the following (where we can disable squashing): we merge the PR even if some commits are 'superfluous' but ask the contributor to squash as much as possible.

In general I think this is a good idea but some contributors might not be familiar with squashing via `git rebase -i` and then doing a forced push. However, it's not that hard and we could provide a guideline how to publish a good contribution.

>Doesn't GitHub show both?
Commits history has the author's picture (with commiter's smaller picture).
And the commits mention "<author> commited with <committer>".

Not for squashed merges via the github UI. I just tested this with a fork of https://github.com/graphhopper/jsprit. After a squashed merge of the last two commits github changes the author to me and sets noreply@github.com as the committer (I used `git log --pretty=full`).




I guess that github sets the person as the author who created the pull request. This would explain why @karussell was set as the committer for graphhopper/map-matching#49.

When I did a rebase merging of the last two commits with github UI, the author stayed the same but the committer was changed to the person doing the rebase merging:





This also happens when manually merging a commit of somebody else into another branch.

PR #910, opened 2016-12-14
Make it possible to add value parser

karussell:
This PR makes it possible to add value parsers e.g. for our current encoders we do not need number parsing and if need this can be added later. Also currently the behaviour is to trigger later added parsers earlier for parsing so that we can use the firstly added date parse as the last that does the parsing if previous parses did return 'invalid'.


core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -472,8 +472,10 @@ protected double getFerrySpeed(ReaderWay way, double unknownSpeed, double shortT
                             shortTripsSpeed = speedEncoder.factor / 2;
                         }
                     } else {
-                        long lastId = way.getNodes().get(way.getNodes().size() - 1);
-                        if (way.getNodes().get(0) != lastId)
+
+                        long lastId = way.getNodes().isEmpty() ? -1 : way.getNodes().get(way.getNodes().size() - 1);
```
boldtrn: Almost created a PR for this :)

boldtrn:
> Also currently the behaviour is to trigger later added parsers earlier for parsing so that we can use the firstly added date parse as the last that does the parsing if previous parses did return 'invalid'.

Good decision, therefore we can also "Override" behavior, like in CSS.

karussell:
(Now merged this PR instead of a cherrypick or rebase. A bit ugly is that the extra merge commit via github UI is not verified.)

PR #911, opened 2016-12-15
Avoid copying graph, switch to CH routing via request parameter

karussell:
This change avoids copying the graph necessary for #912. Copying the graph changes nodes & edge ids which introduces problems when trying to include turn costs for CH. 

TODOs:
 * Investigate if this is faster (it should)
 * Investigate if ch routing is done

### DIFF
--- a/reader-osm/src/test/java/com/graphhopper/routing/RoutingAlgorithmWithOSMIT.java
+++ b/reader-osm/src/test/java/com/graphhopper/routing/RoutingAlgorithmWithOSMIT.java
@@ -45,6 +45,7 @@ import static com.graphhopper.GraphHopperIT.DIR;
 import static com.graphhopper.util.Parameters.Algorithms.ASTAR;
 import static com.graphhopper.util.Parameters.Algorithms.DIJKSTRA_BI;
 import static org.junit.Assert.assertEquals;
+
 import org.junit.Ignore;
 
 /**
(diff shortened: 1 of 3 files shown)


karussell:
Thanks. This cleaning was necessary for an upcoming PR which aims to fix #270 

It removes the usage of copying the graph which would be broken if it contains turn costs. We should document this on 'ghStorage.copyTo(storageCopy)', because in this method we ignore unused edges and copying into a new graph makes the int values of a turn cost storage invalid

reader-osm/src/test/java/com/graphhopper/routing/RoutingAlgorithmWithOSMIT.java
```diff
@@ -139,52 +140,6 @@ public void testBike2_issue432() {
     }
 
     @Test
-    public void testMonacoAllAlgorithmsWithBaseGraph() {
```
karussell: We can remove this as we test this implicitly now that we have just one graph

reader-osm/src/test/java/com/graphhopper/routing/RoutingAlgorithmWithOSMIT.java
```diff
@@ -570,7 +529,7 @@ Graph runAlgo(TestAlgoCollector testCollector, String osmFile,
             throw new RuntimeException("cannot handle " + algoEntry.toString() + ", for " + tmpOneRun
                     + ", file " + osmFile + ", " + ex.getMessage(), ex);
         } finally {
-            // Helper.removeDir(new File(graphFile));
+            Helper.removeDir(new File(graphFile));
```
karussell: Now that we apply different CH preparations (i.e. due to different weightings) we need to get rid of the graph file ... to avoid reimport for a new test file with the same area we could attach the weighting to the graph folder.

reader-osm/src/test/java/com/graphhopper/routing/RoutingAlgorithmWithOSMIT.java
```diff
@@ -527,18 +482,27 @@ public void testNeudrossenfeld() {
     Graph runAlgo(TestAlgoCollector testCollector, String osmFile,
                   String graphFile, List<OneRun> forEveryAlgo, String importVehicles,
                   boolean testAlsoCH, String vehicle, String weightStr, boolean is3D) {
+
+        // for different weightings we need a different storage, otherwise we would need to remove the graph folder
+        // everytime we come with a different weighting
+        // graphFile += weightStr;
```
karussell: seems not necessary

PR #912, opened 2016-12-16
Turncost for Contraction Hierarchies

### LINKED ISSUE #270: Turn Costs for CH

b3nn0:
This is a first try of an implementation of turncost support for contraction hierarchies and should fix #270. This code is highly experimental and certainly needs some work.
What works:
- Basic support for turn costs, based on this paper: https://pdfs.semanticscholar.org/4aa9/65d323ccb8b9e7c71de77f910900ff69a773.pdf
- An edge-based, trun-cost respecting version of DijkstraOneToMany. The idea is based on this pull request: https://github.com/graphhopper/graphhopper/pull/394
However, it has mostly been reimplemented based on this code, as there were too many conflicts when trying to merge the pull request into a current master.

Caveats:
- The implementation only supprts a single p-turn at a node. In rare cases, mutliple p-turns might work, but that is highly dependent on the order of node contraction. To understand the issue, assume the following graph:
grafik
Assume that coming via edge 1, you can only go straight on onto edge 4. Coming from edge 4, you can only turn onto edge 3. The optimal (and only possible) path would be coming from s, taking the p-turn behind edge 4, then taking the p-turn behind edge 3, then going straight-on towards t.
These cases do not work correctly yet and might throw an InvalidStateException

- Limiting U-turns does not yet work correctly. This is usually implemented with code like
`if (prevOrNextEdgeId == edge && noUturnSupport) return Double.POSITIVE_INFINITY `.
However, as we have shortcuts, the search might enter a node through one shortcut and exit it through another, so this check does not trigger. However, both shortcuts might have the same original edge underneath, which will therefore result in a u-turn in the resulting path.

- DijkstraOneToMany does not yet handle source == target routes in edgebased traversal correctly.

-  Integration is not yet done. It is possible to instantiate such an edge-based CH when using the low-level API, but neither the web interface, nor the main GraphHopper class do this yet.

- Performance might be arbitrarily bad. There are quite a few TODO optimize in PrepareContractionHierarchies. It has not yet been tested with larger graphs.

Therefore, the following things need to be done:
- Integration -- should be straight forward.

- Fix DijkstraOneToMany to handle source==target routes in edge-based mode (this is also the cause for the currently failing unit tests).

- CH shortcuts need to be extended by two more ints, which need to represent the original edge ID from the main graph (not only the both skipped edges). This would allow implementing proper u-turn support. Additionally it would make it possible that we do not need to add turn cost info for shortcuts. Right now, whenever a shortcut is inserted, the node's turn cost table needs to be extended, so the shortcut receives the same turn cost info as the underlying original edge. By referencing this original edge, we would be able to omit this, which could result in much better performance (due to smaller turn cost tables).

- The CH witness search needs to be changed. Instead of doing this search with a node-filter that skips the center node (v in a u-v-w contraction), we should include this node in the search and later check if it is part of the witness path.
This should be done because PrepareContractionHierarchies.computeExistingDirectWeight() currently only checks for cost(u,v) + cost(v,w) + cost(bestLoop(v)) - which is the root cause for why no double-p-turns are supported.
By changing the witness search, this search could be used to find the potentially multiple loops and node v and therefore allow multiple p-turns.
However, this also needs to be implemented in Path4CH, which currently also only checks for the single best loop. Instead, Path4CH then needs to do a small local Dijkstra search to find the best combination of loops at v to expand the shortcut (u,w) to (u,v) + loop_at_v_1 + loop_at_v_..n + (v,w).

- Check the performance in different conditions and potentially implement all those TODO optimize, as well as the loop-avoidance strategy from the paper.


karussell:
Thanks a lot for this! This is a big step forward :)

Behind a huge slowdown of the preparation (and also the query time) is probably due to your mentioned superfluous turn cost information that we currently have to add. Can we somehow avoid two new integer fields? I thought about storing the turn cost information per "from-edge", then we could point to this from every edge. The problem then is that we need cannot have turn cost info for the "to-edge".

Despite this I stumbeled over two correctness issues:

A new TurnWeighting is created inside the constructor of PrepareContractionHierarchies. But at query time we need to use the QueryGraph wrapper instead of the graph to properly handle turn costs also for virtual edges&nodes (see QueryGraph.QueryGraphTurnExt). And creating the TurnWeighting at this early time makes this impossible, I think we need to move this into createAlgo somehow.

A second problem I currently do not fully understand is the failing test GraphHopperIT.testCHAndTurnCostsWithFlexmode 

When we add
```java
 req.getHints().put(Routing.EDGE_BASED, "true");
```
for the CH test it should give the same result (1044m) but gives 1176m

I have to further investigate why I get another result (587m) when I run this through the web UI and get the following strange artifact

edge_based_ch

karussell:
Highly likely that we got many problems here due to #991 too

karussell:
Closing in favor of #1247

Issue #845, opened 2016-10-21
Make writes thread-safe


karussell:
The new overlay module should be usable at some point on a per-request method. For that we need as a first step a write-thread-safe handling of graph updates and e.g. apply them permanently.

Potentially we need to break the Java API here to avoid 'flush' for every edge.setXY method and instead flush explicitely only after a certain method call (edge.flush) or automatically after edgeIterator.next calls (which would avoid breaking backward compatibility)

karussell:
This was done in the [change-graph-api branch](https://github.com/graphhopper/graphhopper/commits/change-graph-api). Currently we do not use a per-edge locking but instead use an easier global write lock to make sure that every Dijkstra sees consistent data. When we see that the writes take too long and can't be further optimize we'll need to reinvestigate how to establish per-edge locking, and if it would be faster at all.

michaz:
Roughly, this looks good to me. :-)

I would test it by somehow injecting a fake graph modifier, let that wait, concurrently make a routing request, let the fake graph modifier continue (1), and take care that my routing request doesn't finish before I do (1).

Never done it before, though.

About per-edge locking: I think, no matter how slow global locking is, we need a _theory_ about what changing edges on the fly does to our algorithms. Before that, I think we don't need to investigate performance.

michaz:
To model delays in public transport, we will at some point need to make structural changes to the graph -- changing edge flags will not be enough. 

When a trip is delayed, we must change the to-node of its transfer edge so that it points to the correct point in time.

michaz:
Though it's not clear we have to do any of this in the base graph. Maybe everything delay-related can be in a ``QueryGraph``-like thing which stays wrapped around the base graph.

karussell:
> Though it's not clear we have to do any of this in the base graph.

Yes, we could do this in the QueryGraph. In the base graph disconnecting edges is already possible but adding edges (temporary) to a node leaves gaps in the edge space ... which could be managed by some garbage collector though.

karussell:
Done in #918

karussell:
I stumbled over this project https://github.com/google/thread-weaver (kind of 'stale' status)

PR #918, opened 2016-12-22
Test concurrent graph change api while routing

karussell:
To test #845 better I introduced one new concurrent test. But this looks rather complicated, maybe I did something wrong.

### DIFF
--- a/core/src/test/java/com/graphhopper/GraphHopperAPITest.java
+++ b/core/src/test/java/com/graphhopper/GraphHopperAPITest.java
@@ -17,13 +17,22 @@
  */
 package com.graphhopper;
 
-import com.graphhopper.reader.osm.GraphHopperOSM;
+import com.graphhopper.json.geo.JsonFeature;
 import com.graphhopper.routing.util.EncodingManager;
+import com.graphhopper.storage.Graph;
 import com.graphhopper.storage.GraphBuilder;
 import com.graphhopper.storage.GraphHopperStorage;
 import com.graphhopper.storage.NodeAccess;
+import com.graphhopper.storage.change.ChangeGraphHelper;
+import com.graphhopper.storage.index.LocationIndex;
 import com.graphhopper.util.PointList;
+import com.graphhopper.util.shapes.BBox;
 import org.junit.Test;
+
+import java.util.*;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.junit.Assert.*;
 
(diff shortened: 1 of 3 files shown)


michaz:
This is really tricky. Hold on a bit, I'm onto this.

core/src/test/java/com/graphhopper/GraphHopperAPITest.java
```diff
@@ -115,4 +124,69 @@ public void testNoLoad() {
...
+        Thread thread1 = new Thread() {
+            @Override
+            public void run() {
+                graphHopper.changeGraph(list);
+                checkPointCounter.incrementAndGet();
+            }
+        };
+        thread1.start();
+
+        Thread thread2 = new Thread() {
```
michaz: JUnit (or anything else) doesn't see assertion failures (or any other exceptions) on self-opened threads. When you use an ``ExecutorService`` instead, exceptions are automatically re-thrown in the main thread, and then the calling tool/IDE can actually see the failed assertions.
Right now, it only works because of the final check that the counter is 3, meaning that no thread has crashed.

core/src/test/java/com/graphhopper/GraphHopperAPITest.java
```diff
@@ -115,4 +124,69 @@ public void testNoLoad() {
...
+        Map<String, Object> properties = new HashMap<>();
+        properties.put("speed", 5);
+
+        list.add(new JsonFeature("1", "bbox",
+                new BBox(10.399, 10.4, 42.0, 42.001),
+                null, properties));
+
+        Thread thread1 = new Thread() {
+            @Override
+            public void run() {
```
michaz: Try inserting
```
                    Thread.sleep(1);
```
here, at the beginning of thread1.
The test will fail, because we will not have even started locking the graph yet when the second routing request comes, and it will (correctly) see the old version.
That means that this test isn't really testing a stable situation at the moment.
I found this because I asked myself why we even need thread2: We can just as well send the second routing request on the main thread. But then the test fails, because opening thread2 apparently takes 1 microsecond, and when we are faster (_by not doing that_), we have the same situation as above.

So at the moment, you are only using thread 2 to fix a race condition.

Not a full solution, but: The second routing request should not be sent before we have reason to believe that thread1 has already had enough time to lock the graph.

karussell: The sleep is more a hack and maybe we can replace it using some barriers. I have to investigate how to do that properly. I've introduced two threads so that the second one will be stopped at the lock for sure (or very likely).

What we really want to test is that read and write cannot happen at the same time. Maybe we cannot really unit test this and have to write an integration test. Hmmh.

core/src/test/java/com/graphhopper/GraphHopperAPITest.java
karussell: Ups, right.

core/src/test/java/com/graphhopper/GraphHopperAPITest.java
michaz: I would say we want a test which fails when we comment-out the locking, and succeeds otherwise.

I don't mind your 1-second-sleep. I have no better idea myself at the moment. The problem I described is unrelated to that.

karussell: So you would want a failing test which shows that the route was calculated from a corrupted graph? E.g. calculates parts of the SPT, then the graph updates happens and then the final route weight is different to the weight calculated from the end node of the SPT?

My test makes sure that if there are no locks a different result happens, which is likely still correct. Not sure if forcing the state like mentioned above is easy to reach and at the end more helpful than many parallel requests in an integration test which could show the same issue with a high enough count&pressure.

michaz: OK, maybe let's not mix these two things.

Please insert these lines at the beginning of the ``thread1`` code:

```
                try {
                    Thread.sleep(1);
                } catch (InterruptedException e) {
                }
```

And run.

The test will fail.

How can a test which changes from succeeding to failing just because of a 1-millisecond difference (i.e. "almost by chance") be a good test?

karussell: Ah, now I understand this, I misunderstood this at the first place ... this needs indeed a fix!

karussell:
I made the test more stable with the barrier. Also removed the second thread and use an executor service for better exception handling.

PR #925, opened 2016-12-28
Add .jitpack.yml

michaz:
JitPack, which is super-useful, [tries to](https://jitpack.io/#graphhopper/graphhopper) build GraphHopper with Java 7, because it guesses that from our parent-pom. It doesn't realise that the web submodule needs Java 8.

Let's help it by adding a ``.jitpack.yml``. Then it works.

### DIFF
--- a/.jitpack.yml
+++ b/.jitpack.yml
@@ -1,0 +1,2 @@
+jdk:
+  - oraclejdk8


karussell:
Isn't this a tooling issue which should be solved within the tool itself?

Do we need this due to gtfs lib? As written in [this discussion](https://github.com/conveyal/gtfs-lib/issues/23#issuecomment-263813259) I do not like several external repositories. Also this makes the build more fragile as it then completes only if all repos are online.

michaz:
It enables other people to include our SNAPSHOT builds (and single GitHub commits) as a Maven dependency.

I do it sometimes myself. For instance, I can have a GitHub project, possibly an internal one, which references a certain unreleased GraphHopper version by its commit hash. Super useful.

karussell:
Understand & sounds useful!

Issue #175, opened 2014-03-31
Country-specific speed limits


raunilillemets:
I am using GraphHopper to import OSM data and transform it to graph-based format for a certain program that we are currently using.

As we only work with one country's data at any time, I would like to use an class AreaSpecificCarFlagEncoder extends CarFlagEncoder that would receive the speed limits for a given area in beginning of work of the program.

(More sophisticated) alternative: input for AreaSpecificCarFlagEncoder would be N sets of (WKT geometries and sets of speed limits). AreaSpecificCarFlagEncoder would make sure that the geometries are non-intersecting; and if given way is fully in some area, then the given speed limit is used. If way is not inside any of the given areas, default values are used.
I do not know what would be the best course of action if way is partially in one or more areas. Perhaps it would be most suitable to use default speeds then.

Please comment.

ratrun:
A country specific speed limit implementation should also allow handling of country specific access restrictions. It should provide the possibility to configure tables similar to http://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restrictions. The evaluation of these tables shoud be possible for all kinds of encoders, not just for cars.

ratrun:
I did some investiagtions how this feature could be implemented and found the following library, which does what is needed for reverse geocoding into a country code: 
https://github.com/bencampion/reverse-country-code 
This library does not use OSM data, but rather data from thematicmapping.org . This data is not as precise as the OSM, but I think that it would be sufficient for our purpose. Also the license should be OK.
A totally different appraoch would be to extract the polygons on our own during parsing of the OSM data on the fly. In comparison to using the libraray mentions or a similar library, I see a lot of draw-backs:
1.  It does not work for OSM extracs
2.  It increases the reader complexity
3.  It slows down graph Generation
4.  It will be problematic because from the sea-polygon discussions I know that the country releations will be frequently broken in OSM.

First I thought that the reverse geo-coding task has similarities to the inner/outer city speed limit of issue #230, but after having looked on the code provided by jansoe I came to the conclusion that this is something quite different.
Could you take a look on the library with your GIS expertiese? What approach do you prefer? I made some performance measurements with the lib and came to the conclusion that it is able to look up a lat/lon in an order of magnitute of about 0.1 ms on a 3 years old PC. I think that such performance should be sufficient.

karussell:
Thanks for this idea!

> It does not work for OSM extracs

[...]

Issue #891, opened 2016-12-01
Decide on javadoc guidelines


stefanholder:
Currently, a number of methods is not or only poorly documented with javadoc, which makes it hard for others to contribute. This also applies to the map-matching repository.

I think what we need are is a javadoc convention (see also #774) that applies to all new contributions. I would suggest to adopt the [javadoc rules from the Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s7.3-javadoc-where-required), which describe **where** javadoc is required.

In addition to **where** javadoc is required, we also need some agreement on **how** to write good javadoc comments.

In #774, @karussell commented:
>Also what to comment and what not is important and this is a good read: https://visualstudiomagazine.com/articles/2013/06/01/roc-rocks.aspx
It's important to communicate what the code should be doing. Comments should explain the "why" of the code and stop there.
What comments should never, ever do is explain how the code does its job.

boldtrn:
Yes the code is not commented a lot. I think we should aim for speaking names, short (if possible ;)) methods that can be understood easily. We should break the methods into smaller methods where possible and assign speaking names to the methods, so it becomes clear what happens. This is basically the  approach presented in Clean Code (by Robert Martin). I think most method comments are unnecessary, however some methods are rather complex, these are typically documented already.

Javadoc becomes outdated very fast. Usually with the next refactoring one has to fix the Javadoc anyway.

> In #774, @karussell commented:
+1

stefanholder:
>It's important to communicate what the code should be doing.

I agree. An additional criterion for good javadoc comments is for me that users of the API should understand from javadoc alone how to use the API. Often it is not sufficient to just roughly describe what a method does but to be precise about the preconditions, postconditions and side effects of a method. An example for good javadoc is the java.util package. e.g. [ArrayList](https://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html). It does fully explain how to use the API without giving implementation details.

IMHO, the best mindset for writing good javadoc is [Design by Contract](http://www.javapractices.com/topic/TopicAction.do?Id=194).

>Javadoc becomes outdated very fast.
I would require that javadocs are updated with every pull request by the pull request author. This way javadoc never becomes outdated.

karussell:
> be precise about the preconditions, postconditions and side effects of a method

yes, like information about thread-safety and if stateless.

> > Javadoc becomes outdated very fast.

[...]

Issue #472, opened 2015-07-22
Refactoring to reduce set/getFlags usage


karussell:
The get/setFlags usage is very low level and confusing for new users. Additionally it hurts that the behaviour often cannot be overloaded. For example for #464 I needed a different flags layout for shortcuts which is FlagEncoder-independent and therefor e.g. `edgeState.isForward(flags)` would not work for multiple encoders. The solution in #464 needs to be rolled out for more properties like get/setSpeed etc via just attaching the methods to the edge ala `edgeState.isForward(FlagEncoder)`. A similar problem we had when introducing `edge.getBool` for a different behaviour for virtual edges.

Then I was able to easily throw an exception if `shortcut.getFlags` was called because reading the speed from a shortcut does not make sense.

Of course we could even go further and avoid the `FlagEncoder` argument for every method call and e.g. specify the encoder already for the EdgeExplorer to make it available for all EdgeIterator calls. Which would make the usage very Java-way and user friendly: `edge.getSpeed()` or `edge.isForward()`. 

Another possibility is to do `encoder.getSpeed(edge)` and a check inside if it is a shortcut or not. But this has not the possibility of getting rid of the argument.

We need to take care that such a refactoring does not hurt performance. 

Additionally with this refactoring in place we need to investigate if we can get rid of the EncodingManager withing the BaseGraph, as well as avoiding the expensive loop over the flagEncoders in encodingManager.reverseFlags and instead call this just for the FlagEncoder specified at the first place while creating the EdgeExplorer.

Another possibility to hide this complexity is to do something like:

```java
car = encodingManager.getEncoder("car")
maxSpeed = car.getProperty("max_speed");
maxSpeed.get/set // internally we map this to the correct car max_speed property
```

# Update

The usage is now much simpler in 99% of the cases: edge.get(accessEnc) via #1447. Will keep it open until the 1% is fixed too. Also see #1625.

 - [x] remove FlagEncoder.flagsDefault

devemux86:
:+1: 
This refactoring seems interesting and indeed would make the API very user friendly.

Azbesciak:
Just a question - maybe allow to call `getFlags` on shortcut?
There may be case when you have custom weighting over `PreparationWeighting` which depends also on flags, for example - avoid highways. Such information may be included on base edge, then when creating shortcut (which include that edge) with `or` on this field you would avoid such edge if weighting will discover it.
[...]

PR #926, opened 2016-12-30
Spatial Rule Lookup (e.g. Country Specific Rules)

boldtrn:
This PR intents to solve #175.

The basic data structure is an array, that separates the world (or an area) into tiles, with a given resolution. Each tile contains a reference to a set of rules.

Rules contain a set of polygons in which their rules are valid. Using an array has to disadvantages: (1) covering unused space, e.g. water and (2) inexact resoultion. Currently the default resolution about 20*20km at the equator (0.1 degree). We can achieve exact results when doing a Polygon contains check, which can be configured.

Right now I prepared two set of rules, Austria and Germany, but rules can be written for every Country. All countries are available in the countries.geo.json file. This file is based on the admin levels from [NaturalEarthData](http://www.naturalearthdata.com/) under [public domain](http://www.naturalearthdata.com/about/terms-of-use/). As the file would be 25mb, is reduced it to 10% using DP. Therefore the borders are sometimes slightly inaccurate (should not be too bad, otherwise let me know).

Lookup speed is not too bad. 10 000 000 lookups in an area that is filled approximately half by a polyong, takes between 3-4 seconds on my machine.

TODOs:

- [x] Nr of rules is currently limited to 255, but since we have to create a new rule for every unique set of country borders, the array might be with 20-30 countries, depending on the borders.
- [x] The SpatialLookup is currently created in the web module. It would be better if we could have it as `Storable<LocationIndex>`. That way we could load it, once it was created (even though creation is quite fast). Also I am not sure how to do this (especially with the rule reference array, the lookup array should be easy)?
- [x] How to evaluate these rules? Currently I do this in the DataFlagEncoder, however it might be better to do this in the GenericWeighting? IMHO the DataFlagEncoder is representing the tags of a given source (e.g. OSM), but not business logic. For example we are also not saving the default max speeds in the edge, but calculate it every time in the GenericWeighting
- [ ] Consider using the exact country borders - would slow down the Polygon contains approx. 10 times. Also we have to host this file, as IMHO 25mb should not be put into the repo.
- [x] Differentiate between Polygon and Multipolygon? -> Later
- [ ] Add Javadoc to the Interfaces
- [x] Make spatial index configureable. What rules to load and if it should be used at all.

Let me know what you think. Currently this is still work in progress, so any comments are more than welcome.


karussell:
Thanks - will have a look!

> As the file would be 25mb, is reduced it to 10% using DP

Maybe this is not too critical. How large is the 25mb file if compressed? As git and jar normally does compression and it will be big just on the disc and not in production.

boldtrn:
> How large is the 25mb file if compressed?

2.5mb. You can view the file here: http://bl.ocks.org/d/7280192619e0cc25f609dafe5c0de2cf

I think the borders are not too bad, but we can definetly use the exact borders if this becomes an issue.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1261,4 +1265,13 @@ public void setNonChMaxWaypointDistance(int nonChMaxWaypointDistance) {
         this.nonChMaxWaypointDistance = nonChMaxWaypointDistance;
     }
 
+    public void setSpatialRuleLookup(SpatialRuleLookup spatialRuleLookup) {
```
karussell: Hmmh, I do not like this procedure much. We should not provide a setter here, if this is not an additional thing for GraphHopper (infact GraphHopper.spatialRuleLookup is unused).

Instead the user should/could do `graphHopper.getEncodingManager().getEncoder("generic").setXY` ... or maybe independent on generic and just `graphHopper.getEncodingManager().setXY`?

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -217,6 +224,15 @@ int getAccessValue(ReaderWay way) {
...
         }
 
+        if(accessValue == 0){
+            GHPoint estmCentre = way.getTag("estimated_center", null);
+            if (estmCentre != null) {
+                SpatialRule rule = spatialRuleLookup.lookupRule(estmCentre);
+                accessValue = getEdgeValueForAccess(rule.isAccessible(way, ""));
+            }
+
+        }
```
karussell: Can you explain this case?

core/src/main/java/com/graphhopper/routing/util/spatialrules/Polygon.java
```diff
@@ -0,0 +1,173 @@
...
+package com.graphhopper.routing.util.spatialrules;
+
+import com.graphhopper.util.shapes.GHPoint;
+
+/**
+ * TODO: Implement Shape?
+ *
+ * @author Robin Boldt
+ */
+public class Polygon {
```
karussell: This needs more javadocs

core/src/main/java/com/graphhopper/routing/util/DefaultFlagEncoderFactory.java
```diff
@@ -17,7 +17,15 @@
...
 
+import com.graphhopper.reader.ReaderWay;
+import com.graphhopper.routing.util.spatialrules.*;
+import com.graphhopper.routing.util.spatialrules.countries.GermanySpatialRule;
+import com.graphhopper.util.DouglasPeucker;
 import com.graphhopper.util.PMap;
+import com.graphhopper.util.shapes.BBox;
+
+import java.util.ArrayList;
+import java.util.List;
```
karussell: Not necessary (?)

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Defines rules that are valid for a certain region, e.G. a country.
+ * A rule might be access, max-speed, etc.
+ *
+ * @author Robin Boldt
+ */
+public interface SpatialRule {
+
+    int getMaxSpeed(ReaderWay readerWay, String transportationMode);
```
karussell: return type should be double

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.List;
+
+/**
+ * Defines rules that are valid for a certain region, e.G. a country.
+ * A rule might be access, max-speed, etc.
+ *
+ * @author Robin Boldt
+ */
+public interface SpatialRule {
```
karussell: Hmmh, I'm not sure about the interface. It looks like we'll break implementations as soon as we need a different data from the index. Instead of getCountryIsoA3Name, isAccessible and getMaxSpeed maybe we should be more generic here? (`Double getData("max speed", Double.class)`) hmmh, we'll have to think about this.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleContainer.java
```diff
@@ -0,0 +1,78 @@
...
+                return false;
+            }
+        }
+        return true;
+    }
+
+    @Override
+    public boolean equals(Object o){
+        if(o instanceof SpatialRuleContainer){
+            if(this.represents(((SpatialRuleContainer) o).getRules()))
```
karussell: This is not really correct. Only because X has all rules of Y does not mean the opposite i.e. they are not necessarily equal. But maybe this is not important?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleContainer.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public int size(){
+        return this.rules.size();
+    }
+
+    public SpatialRule first(){
+        return this.rules.iterator().next();
+    }
+
+    public boolean represents(Set<SpatialRule> rules) {
```
karussell: naming it "contains" would make it more explicit (?)

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookup.java
```diff
@@ -0,0 +1,39 @@
...
+ * @author Robin Boldt
+ */
+public interface SpatialRuleLookup {
+
+    SpatialRule lookupRule(double lat, double lon);
+    SpatialRule lookupRule(GHPoint point);
+
+    void addRule(SpatialRule rule);
+
+    SpatialRule getEmptyRule();
```
karussell: I would remove this method and make the EMPTY_RULE constant part of this interface

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookup.java
```diff
@@ -0,0 +1,39 @@
...
+import com.graphhopper.util.shapes.GHPoint;
+
+import java.util.List;
+
+/**
+ * Lookup a rule depending on the location.
+ *
+ * @author Robin Boldt
+ */
+public interface SpatialRuleLookup {
```
karussell: javadocs are missing

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/DefaultSpatialRule.java
```diff
@@ -0,0 +1,78 @@
...
+                return -1;
+        }
+    }
+
+    @Override
+    public AccessValue isAccessible(ReaderWay readerWay, String transportationMode) {
+        String highwayTag = readerWay.getTag("highway", "");
+
+        // As defined in: https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restriction
+        // We tried to find generally forbidden tags
```
karussell: forbidden for a certain transportationMode?

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/DefaultSpatialRule.java
```diff
@@ -0,0 +1,78 @@
...
+package com.graphhopper.routing.util.spatialrules.countries;
+
+import com.graphhopper.reader.ReaderWay;
+import com.graphhopper.routing.util.spatialrules.AbstractSpatialRule;
+import com.graphhopper.routing.util.spatialrules.AccessValue;
+
+/**
+ * Default Implementation for the SpatialRule that contains the current default Values
+ */
+public class DefaultSpatialRule extends AbstractSpatialRule {
```
karussell: Shouldn't this be part of the AbstractSpatialRule, e.g. getCountryIsoA3Name implementation is still missing and so this is not really a default (?)

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GermanySpatialRule.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public int getMaxSpeed(ReaderWay readerWay, String transportationMode) {
+        String highwayTag = readerWay.getTag("highway", "");
+
+        // As defined in: https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxspeed#Motorcar
+        switch (highwayTag) {
+            case "motorway":
+                return Integer.MAX_VALUE;
+            case "trunk":
+                return Integer.MAX_VALUE;
```
karussell: Let us define something for this. E.g. there are 130 limits but I think somewhere I used 140 ... the problem with MAX_VALUE is that this cannot be stored in the encoder

core/src/test/java/com/graphhopper/routing/util/spatialrule/SpatialRuleLookupArrayTest.java
```diff
@@ -0,0 +1,203 @@
...
+        assertEquals(AccessValue.ACCESSIBLE, spatialRuleLookup.lookupRule(48.864716, 2.349014).isAccessible(track, ""));
+        assertEquals(AccessValue.ACCESSIBLE, spatialRuleLookup.lookupRule(48.864716, 2.349014).isAccessible(primary, ""));
+
+        // Vienna
+        assertEquals(AccessValue.ACCESSIBLE, spatialRuleLookup.lookupRule(48.210033, 16.363449).isAccessible(track, ""));
+        assertEquals(AccessValue.ACCESSIBLE, spatialRuleLookup.lookupRule(48.210033, 16.363449).isAccessible(primary, ""));
+        assertEquals(AccessValue.EVENTUALLY_ACCESSIBLE, spatialRuleLookup.lookupRule(48.210033, 16.363449).isAccessible(livingStreet, ""));
+    }
+
+    private Polygon parsePolygonString(String polygonString){
```
karussell: You could use JSONObject in tests to deserialize the string.

core/src/main/java/com/graphhopper/json/geo/GeoJsonPolygon.java
```diff
@@ -0,0 +1,77 @@
...
+import com.graphhopper.util.PointList;
+import com.graphhopper.util.shapes.GHPoint;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Created by robin on 12.12.16.
+ */
+public class GeoJsonPolygon implements Geometry {
```
karussell: Javadocs are missing.

core/src/main/java/com/graphhopper/json/geo/GeoJsonPolygon.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private List<Polygon> polygons = new ArrayList<>();
+
+
+    @Override
+    public String getType() {
+        return null;
```
karussell: Here we should return Polygon or MultiPolygon or create two implementations?

web/src/main/resources/com/graphhopper/http/README.MD
```diff
@@ -0,0 +1,5 @@
+The Polyongs in this folder are from here: https://github.com/datasets/geo-countries and provided under ODC Public Domain Dedication and Licence (PDDL)
+
+The Geojson has roughly 25MB, so I reduced it with http://www.mapshaper.org/ using Douglas Peuckar at 10%. The resulting file is 2.4mb.
+
+The rules are taken from the OSM Wiki: ttps://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restrictions
```
karussell: Should be in the NOTICE.md instead (?)

And why is this and countries.geo.json in package com/graphhopper/http ?

web/src/main/java/com/graphhopper/http/DefaultModule.java
```diff
@@ -52,6 +52,9 @@ public GraphHopper getGraphHopper() {
      */
     protected GraphHopper createGraphHopper(CmdArgs args) {
         GraphHopper tmp = new GraphHopperOSM().forServer().init(args);
+        //TODO We should set the BBox and resolution
+        //TODO, move to a more appropriate place
+        tmp.setSpatialRuleLookup(SpatialRuleLookupBuilder.build());
```
karussell: related to the TODOs :) ... it is suboptimal that it cannot be configured where to load the json from and the resolution ...

web/src/main/java/com/graphhopper/http/SpatialRuleLookupBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+        return SpatialRuleLookupBuilder.build(DEFAULT_BOUNDS, DEFAULT_RESOLUTION, DEFAULT_EXACT);
+    }
+
+    public static SpatialRuleLookup build(BBox bounds, double resolution, boolean exact) {
+        SpatialRuleLookup spatialRuleLookup = new SpatialRuleLookupArray(bounds, resolution, exact);
+        try {
+            GHJson ghJson = new GHJsonBuilder().create();
+            JsonFeatureCollection jsonFeatureCollection = ghJson.fromJson(new FileReader(new File(SpatialRuleLookupBuilder.class.getResource("countries.geo.json").getFile())), JsonFeatureCollection.class);
+
+            // TODO find outer Border of all used features and create SpatialRuleLookupArray onlyo for these Bounds
```
karussell: What do you mean here?

(+minor spelling issues)

web/src/main/java/com/graphhopper/http/SpatialRuleLookupBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+                        if(!geometry.isPolygon())
+                            continue;
+                        spatialRule.setBorders(geometry.asPolygon().getPolygons());
+                        spatialRuleLookup.addRule(spatialRule);
+                        break;
+                    }
+                }
+            }
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
```
karussell: please never catch exceptions and continue silently. Instead rethrow e.g. via `new RuntimeException(e)`

core/src/main/java/com/graphhopper/json/geo/GeoJsonPolygon.java
boldtrn: Added Polygon for now. We definitely have to think about how we handle this in the future.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
boldtrn: Yes, the idea was to first check the ReaderWay if there is an access restriction value on it. If not check the country rules to eventually add a restriction.

core/src/main/java/com/graphhopper/json/geo/Geometry.java
```diff
@@ -17,11 +17,12 @@
  */
 package com.graphhopper.json.geo;
 
+import com.graphhopper.routing.util.spatialrules.Polygon;
 import com.graphhopper.util.PointList;
 import com.graphhopper.util.shapes.GHPoint;
 
 /**
- * This interface is used to define an area or location being used to block or change areas in the graph.
+ * This interface is used to define an area or location in a geojson file.
```
boldtrn: Oups, you'r right :)

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: Makes sense :+1: I think it makes sense to start only with generic for now. We can always extend that.

core/src/main/java/com/graphhopper/routing/util/spatialrules/Polygon.java
boldtrn: Added some, probably still needs more :)

core/src/main/java/com/graphhopper/routing/util/DefaultFlagEncoderFactory.java
boldtrn: Yes, removed.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
boldtrn: Yes that is true. That might be better. Would make it way more flexible to change it.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleContainer.java
boldtrn: I thought about that, This is a rulesEquals and not a contains.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleContainer.java
boldtrn: If X and Y have the same number of rules, than this should be correct :)

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleContainer.java
boldtrn: There was a reason that I extracted that method from the equals method. But I cannot remember it right now.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookup.java
boldtrn: Makes sense. Changed that.

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/DefaultSpatialRule.java
boldtrn: Not sure. The idea was to define default rules for streets that are found frequently in different countries. Adding a default CountryIsoA3Name might now work? I might make sense to make this class abstract as well. The idea was to provide defaults, that why I called it Default. Should we move it to the AbstractSpatialRule?

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/DefaultSpatialRule.java
boldtrn: Yes, currently the SpatialRule only uses car rules, but should be used for other transportation modes as well.

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GermanySpatialRule.java
boldtrn: The idea was to use this as an upper bound. If the Encoder tries to initialize the way with a bigger value we are overwriting it with the Spatial MaxSpeed. This leaves most of the responsibility with the Encoder. Should I change it to 140?

core/src/test/java/com/graphhopper/routing/util/spatialrule/SpatialRuleLookupArrayTest.java
boldtrn: Do you think this would be the better approach? I tried to keep the core clean of JSON.

web/src/main/resources/com/graphhopper/http/README.MD
boldtrn: > Should be in the NOTICE.md instead (?)

Yes thats correct. I wanted to fix this once we have a final location for the countries.geo.json.

> And why is this and countries.geo.json in package com/graphhopper/http ?

Currently I load the countries.geo.json with `SpatialRuleLookupBuilder.class.getResource("countries.geo.json")` and SpatialRuleLookupBuilder is in com.graphhopper.http. I had it in the core first, but wasn't able to read the file in all cases (I think when running the tests with Maven).

web/src/main/java/com/graphhopper/http/DefaultModule.java
boldtrn: Yes, I know :) I should fix this.

web/src/main/java/com/graphhopper/http/SpatialRuleLookupBuilder.java
boldtrn: The idea was to calculate a BBox for all SpatialRules. Initializing an Array for world wide coverage makes no sense If we only have a rule for a subset of the world (e.g., Germany and Austria). Would be quite easy to do and would save quite some storage.

karussell:
I tested this with the landmarks branch and this seems to work (at least the spatial rule part, still something else still does not work ;))

Some minor changes I had to do or would like to have :) (maybe already done, did not review your latest changes too detailed)

 * calculate boundaries from the polygons instead of providing one
 * https://github.com/graphhopper/graphhopper/commit/2c04f4790eb646246fa47fced2fd4f7c24f054cc
 * https://github.com/graphhopper/graphhopper/commit/8dc447d5785953e4513562c954c46ec693663289

boldtrn:
> I tested this with the landmarks branch

Thanks for testing :)

> seems to work (at least the spatial rule part, still something else still does not work there ;))

But unrelated to this PR?

> Some minor changes I had to do or would like to have :)

Thanks for the two commits. I will add these to this branch as well. Calculating the boundaries from the polygons is already on my TODO list as well.

> (maybe already done, did not review your latest changes too detailed)

No, haven't made a lot of progress with this PR in the last weeks. Will do some more soon :).

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleContainer.java
karussell: Yes, you are right, but it is slightly confusing. Why not call the equals of the underlying collection?

boldtrn:
> Some minor changes I had to do or would like to have :) (maybe already done, did not review your latest changes too detailed)

These changes are now committed.

boldtrn:
My last commit is broken. I created a more generic approach for the spatial rule. Is this better? Should I change anything? 

BTW: the hack with the Object and the second method wrapping it into generic was needed in my case. Maybe we find a way around this.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
```diff
@@ -29,12 +29,17 @@
...
+
+    <T> T get(String key, ReaderWay readerWay, String transportationMode, T _default);
 
     List<Polygon> getBorders();
+
     SpatialRule setBorders(List<Polygon> borders);
+
     SpatialRule addBorder(Polygon polygon);
 
     String getCountryIsoA3Name();
```
karussell: I think getCountryIsoA3Name is also too specific e.g. imagine the use case from phil where you have circles around a city to define different areas for the speeds. Maybe we keep everything more specific for now and switch later to a more generic version? The problem I see is that although getObject is already very generic, the parameters could be limited ... or should we remove transportationMode e.g. I can image that sometimes this makes no sense. Maybe the best bet would be `getObject(key, ConfigMap, _default)` and we provide a fast 'converter' from ReaderWay to ConfigMap ... hmmh ... not clear. Similar problem like we have for #472

web/src/main/java/com/graphhopper/http/SpatialRuleLookupBuilder.java
```diff
@@ -54,30 +55,80 @@ public static SpatialRuleLookup build() {
...
                         if (!geometry.isPolygon())
                             continue;
                         spatialRule.setBorders(geometry.asPolygon().getPolygons());
-                        spatialRuleLookup.addRule(spatialRule);
                         break;
                     }
                 }
             }
+
+            Double minLon = null;
```
karussell: Here we could use the same approach like in PathWrapper.calcRouteBBox via
bbox = BBox.createInverse and then bbox.update(lat,lon)?

But not sure if it fits perfectly

karussell:
One more thing: I'm not sure if we should put a dependency inside DataFlagEncoder. Is there an alternative to feed it from outside somehow via the post processing hook, looping over all edges and setSpeed?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookup.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.List;
+
+/**
+ * SpatialRuleLookup defines a container that stores SpatialRules and can lookup
+ * a SpatialRule rule depending on the location.
+ *
+ * @author Robin Boldt
+ */
+public interface SpatialRuleLookup {
```
karussell: Here every method also needs some javadocs. And is visualize important?

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/AustriaSpatialRule.java
```diff
@@ -0,0 +1,56 @@
...
+import com.graphhopper.reader.ReaderWay;
+import com.graphhopper.routing.util.spatialrules.AccessValue;
+import com.graphhopper.routing.util.spatialrules.SpatialRule;
+
+/**
+ * Defines the default rules for Austria roads
+ *
+ * @author Robin Boldt
+ */
+public class AustriaSpatialRule extends DefaultSpatialRule {
```
karussell: Are here some `@Override` missing?

boldtrn:
Thanks for your feedback. I will change the code accordingly.

> Thanks & looks good. Just the question regarding specific vs. generic. Probably we should keep the specific stuff until we have a better impression of more use cases?

Yes, I have the same feeling right now. It seems a bit like premature optimization. Let's start easy and make it more generic when we need it :).

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
boldtrn: Ok, for now I renamed the method to getUniqueName, so we are already more generic at this point. As you said, the rest can be changed once we have a better feeling for the use cases.

web/src/main/java/com/graphhopper/http/SpatialRuleLookupBuilder.java
boldtrn: That's a good idea. I updated the intersection code. Looks way better now.

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/AustriaSpatialRule.java
boldtrn: You are right, added them.

boldtrn:
I changed the the way the Spatial Lookup works a bit. 

When importing, an id is written for every edge. The id references to a spatial rule. In the GenericWeighting  these rules are evaluated.

Currently, the evaluation in the GenericWeighting is not yet perfect.

core/src/test/java/com/graphhopper/routing/weighting/GenericWeightingTest.java
```diff
@@ -108,6 +112,43 @@ public void testCalcTime() {
     }
 
     @Test
+    public void testSpatial() {
```
boldtrn: I don't really like this test. It is rather long for a simple test.

karussell:
I have some notes and will further think about how we should progress here.

E.g. the singleton SpatialRuleRegister won't work pretty and is an unnecessary restriction. In general singleton is considered dangerous and for GraphHopper this cannot be used as we allow (for purpose) multiple instances of everything. Also testing is ugly for this etc.

The EMPTY_RULE_CONTAINER is potential buggy as you can add (by accident) multiple rules (at least I do not see a restriction)

Furthermore there are many javadocs missing, see the discussion in #891. But again, first let us decide on what we really want&need :) !

boldtrn:
> E.g. the singleton SpatialRuleRegister won't work pretty and is an unnecessary restriction.

Ok sure, I can change that.

> The EMPTY_RULE_CONTAINER is potential buggy as you can add (by accident) multiple rules (at least I do not see a restriction)

Will have a look, thanks.

> Furthermore there are many javadocs missing

Yes, this is a bit on purpose. I will do that before we merge this PR. When we constantly change the interface, the documentation might suffer.

>  But again, first let us decide on what we really want&need :) !

Good plan!

boldtrn:
I removed the Singleton. I think this was a good idea. This allows us to easily inject the config to make the CountryRules configureable. The configuring part is not done yet.

boldtrn:
> The EMPTY_RULE_CONTAINER is potential buggy as you can add (by accident) multiple rules (at least I do not see a restriction)

Good find! I updated the EMPTY_RULE_CONTAINER and merged the latest master.

karussell:
It looks better but still there is a bigger issue: the spatial rule index and the spatial rule registry are too loosely coupled. I think the registry stuff should be moved into the index. The problem is that the number of rules define the number of required bits in the encoded value, and so we

 - [ ] have to set the spatial rule index earlier (before passing it to the EncodingManager) and
 - [ ] ensure that the configured rules are always the same number as the loaded once. Otherwise we e.g. add two rules (need two bits) for the import and on load a different config applies and the encoded value requires only 1 bit leading to all strange things. So we either have to return the count in the details or add it to the version or something different to enforce this.  (This is a general problem with the DataFlagEncoder that will also be a problem for the recently introduced configurable height&width&weight - will mention it there too)

Furthermore:

 - [ ] the default for the index should be no rules at all and requiring 0 bits and the code is then skipped the lookups entirely
 - [ ] 'Destroy the SpatialRuleLookup' -> Something seems to be wrong. Because either the 'destroy' should be naturally done from the GC or we should keep it. Also import vs. load with different number of rules will break the encoder. Let me think about this.
 - [ ] SpatialRuleRegister: `public int getIdForName(String name)` should be `public int getId(SpatialRule rule)` instead
 - [ ] SpatialRuleRegister: `Map<Integer, SpatialRule>` -> just use an ArrayList :) ?
 - [x] It should be possible with the config properties to specify a list of countries as JSON files. Let us avoid this for now and only enable this feature via Java API for now.
 - [ ] And what do the access values 0, 5, accessMap.get("no") mean? (DataFlagEncoder.getEdgeValueForAccess)
 - [ ] Optionally move the builder class containing just one method into the index too

boldtrn:
> I think the registry stuff should be moved into the index. The problem is that the number of rules define the number of required bits in the encoded value

True, my idea for this was to completely get rid of the SpatialIndex, once we loaded the OSM file.

>  So we either have to return the count in the details or add it to the version or something different to enforce this.

This is a good idea. So I anyone loads a different graph than configured we can throw an exception.

> the default for the index should be no rules at all and requiring 0 bits and the code is then skipped the lookups entirely

Yes that is true, that is something I wanted to do when making it configureable.

>  'Destroy the SpatialRuleLookup' -> Something seems to be wrong. Because either the 'destroy' should be naturally done from the GC or we should keep it.

Well we pass the object reference to the GenericWeighting, we have to unreference it after the import I think.

>  just use an ArrayList :) ?

Probably, yes :)

> And what do the access values 0, 5, accessMap.get("no") mean? (DataFlagEncoder.getEdgeValueForAccess)

This was introduced when making track unaccessible. So the GenericWheighting tried to estimate how accessible a certain street is. So an accessible street would be 0. 5 was I think eventually accessible.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -405,8 +394,23 @@ public long handleWayTags(ReaderWay way, long allowed, long relationFlags) {
...
+
+                GHPoint estimatedCenter = way.getTag("estimated_center", null);
+                if (estimatedCenter != null) {
+                    SpatialRule rule = spatialRuleLookup.lookupRule(estimatedCenter);
+                    flags = spatialEncoder.setValue(flags, spatialRuleLookup.getSpatialId(rule));
+
+                    // block access if rule says so?
+//                    if (rule.getAccessible(way.getTag("highway"), "", AccessValue.ACCESSIBLE).
+//                            equals(AccessValue.NOT_ACCESSIBLE)) {
+//                        flags = accessEncoder.setValue(flags, NOT_ACCESSIBLE_KEY);
```
karussell: Should we store that access is not possible due to the rule here?

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -793,17 +799,33 @@ public boolean isStoreWidth() {
...
                 "|store_weight=" + storeWeight +
-                "|store_width=" + storeWidth;
+                "|store_width=" + storeWidth +
+                "|spatial_rules=" + spatialRules;
     }
 
     @Override
     public int getVersion() {
-        return 1;
+        return 2;
```
karussell: increase due to access bits changed (3 -> 2)

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleContainer.java
```diff
@@ -17,51 +17,56 @@
...
         return false;
     }
 
-    public SpatialRuleContainer copy(){
+    public SpatialRuleContainer copy() {
         SpatialRuleContainer container = new SpatialRuleContainer();
-        for (SpatialRule rule: this.rules) {
-            container.addRule(rule);
-        }
+        container.addRules(this.rules);
```
karussell: Another probably more explicit way would be to use a constructor which takes another container object instead?

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
```diff
@@ -148,19 +148,6 @@ public void testHighwaySpeed() {
...
-        ReaderWay way = new ReaderWay(1);
-        way.setTag("highway", "secondary");
-        assertEquals(AccessValue.ACCESSIBLE, encoder.getEdgeAccessValue(encoder.handleWayTags(way, encoder.acceptWay(way), 0)));
-
-        way.setTag("vehicle", "destination");
-        assertEquals(AccessValue.EVENTUALLY_ACCESSIBLE, encoder.getEdgeAccessValue(encoder.handleWayTags(way, encoder.acceptWay(way), 0)));
-
-        way.setTag("vehicle", "no");
-        assertEquals(AccessValue.NOT_ACCESSIBLE, encoder.getEdgeAccessValue(encoder.handleWayTags(way, encoder.acceptWay(way), 0)));
-    }
```
karussell: will revert this removal ...

karussell:
Have now pushed a change suggestion.

BTW: The same config is now ensured via the spatial_rules number. The height&width etc config are similarly ensured already.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -793,17 +799,33 @@ public boolean isStoreWidth() {
...
 
+
+    public DataFlagEncoder setSpatialRules(int rules) {
+        this.spatialRules = rules;
+        return this;
+    }
+
+    public DataFlagEncoder setSpatialRuleLookup(SpatialRuleLookup spatialRuleLookup) {
+        if (spatialRuleLookup.size() != spatialRules)
+            throw new IllegalArgumentException("You have to configure the encoder to accept the identical amount of rules " +
```
boldtrn: Should we get the number of spatial rules from the lookup then? But the lookup isn't available when constructing the encoder?

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -793,17 +799,33 @@ public boolean isStoreWidth() {
...
+    }
+
     @Override
     protected String getPropertiesString() {
         return super.getPropertiesString() +
                 "|store_height=" + storeHeight +
                 "|store_weight=" + storeWeight +
-                "|store_width=" + storeWidth;
+                "|store_width=" + storeWidth +
+                "|spatial_rules=" + spatialRules;
```
boldtrn: Should we safe the order and name of the countries? Or would this be too much overhead?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
```diff
@@ -17,21 +17,19 @@
...
  *
  * @author Robin Boldt
  */
 public interface SpatialRule {
 
-    double getMaxSpeed(String highwayTag, double _default);
+    double getMaxSpeed(String highway, double _default);
 
-    AccessValue isAccessible(String highwayTag, String transportationMode, AccessValue _default);
+    AccessValue getAccessible(String highwayTag, String transportationMode, AccessValue _default);
```
boldtrn: getAccessValue or getAccessibility?

web/src/main/java/com/graphhopper/http/DefaultModule.java
```diff
@@ -48,15 +59,33 @@ public GraphHopper getGraphHopper() {
         return graphHopper;
     }
 
+    static SpatialRuleLookup buildIndex(Reader reader, BBox graphBBox) {
+        GHJson ghJson = new GHJsonBuilder().create();
+        JsonFeatureCollection jsonFeatureCollection = ghJson.fromJson(reader, JsonFeatureCollection.class);
+        return new SpatialRuleLookupBuilder().build(
+                Arrays.asList(new GermanySpatialRule(), new AustriaSpatialRule()),
```
boldtrn: Should we make the Countries configureable?

core/src/main/java/com/graphhopper/routing/weighting/GenericWeighting.java
```diff
@@ -124,40 +119,16 @@ public double calcWeight(EdgeIteratorState edgeState, boolean reverse, int prevO
         if (time == Long.MAX_VALUE)
             return Double.POSITIVE_INFINITY;
 
-        int spatialId = gEncoder.getSpatialId(edgeState.getFlags());
```
boldtrn: Did I missed it or is the SpatialIndex actually unused now? We safe the CountryId but don't evaluate it.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
karussell: For now lets keep it simple and a bit unsafer until we have something better :)

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
karussell: Exactly that can be the problem

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
karussell: Ah, yes getAccessValue

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
```diff
@@ -20,20 +20,18 @@
...
 
 /**
- * SpatialRuleLookup implementation using an array as datastructure.
- * Currently limited to 255 rules
- * <p>
+ * SpatialRuleLookup implementation using an array as data structure. Currently limited to 255 ruleContainers
  * The covered area is indexed as tiles, with every tile being "quadratic" having the same degree length on every side.
- * <p>
- * Currently one one rule per Tile is allowed. The later added rule will be added to the tile.
+ * Currently only a limited amount of rule combinations per tile are allowed.
```
karussell: It is not limited to the number of rules per container but to the number of rule combinations. E.g. A,B is one combination then A,B,D is another and so on. So maybe not that clear description?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
```diff
@@ -53,15 +53,34 @@
      * @param exact      if exact it will also perform a polygon contains for border tiles, might fail for small holes
      *                   in the Polygon that are not represented in the tile array.
      */
-    public SpatialRuleLookupArray(BBox bounds, double resolution, boolean exact) {
+    SpatialRuleLookupArray(BBox bounds, double resolution, boolean exact) {
+        if (bounds == null)
+            throw new IllegalArgumentException("BBox cannot be null");
+        if (resolution < 1e-100)
```
karussell: The problem is resolution<=0 is not what we want and maybe we can raise this limit here a bit but not lower it IMO

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
```diff
@@ -74,22 +93,13 @@ private int getNumberOfXGrids() {
...
-            return EMPTY_RULE;
-        }
-        if (lat < bounds.minLat) {
-            return EMPTY_RULE;
-        }
-        if (lat > bounds.maxLat) {
-            return EMPTY_RULE;
-        }
+        if (lon < bounds.minLon || lon > bounds.maxLon || lat < bounds.minLat || lat > bounds.maxLat)
+            return EMPTY;
```
karussell: As it is in SpatialRule you can always write SpatialRule.EMPTY ?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
```diff
@@ -174,35 +181,72 @@ public void addRule(SpatialRule rule) {
                             polygon.contains(center.getLat() + checkDiff, center.getLon() + checkDiff)) {
 
                         if (lookupArray[i][j] == EMPTY_RULE_INDEX) {
-                            lookupArray[i][j] = (byte) ruleIndex;
+                            lookupArray[i][j] = (byte) ruleContainerIndex;
                         } else {
                             // Merge Rules
```
karussell: Extracted already a bit ... it is ok for now IMO

web/src/main/java/com/graphhopper/http/DefaultModule.java
karussell: We could do it with reflection? Maybe later?

core/src/main/java/com/graphhopper/routing/weighting/GenericWeighting.java
karussell: This is true, yes. But we could apply speed and access changes when feeding here: https://github.com/boldtrn/graphhopper/blob/59b7e435f11628507cdbc254d27f7ceb3b31501d/core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java#L408

I would avoid putting the index in the GenericWeighting ... too many things already there and as we could change speed and accessibility when feeding this is also not necessary for now

karussell: Would you mind to add applying the speed and accessibility changes and recover may some tests that I removed? This seems indeed important to me, now that you pointed me to it

karussell:
(BTW: also please have a look at some comments we made at 'outdated' commits here in this PR :))

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
boldtrn: But is it really limited per tile? It is limited for the whole lookup, but not per tile.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
boldtrn: I am perfectly fine with the limit, I was just wondering why it is that number :).

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
boldtrn: Yes, SpatialRule.EMPTY looks good I think. It's just a bit hard to read.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
boldtrn: Yes I saw that. Should be fine.

core/src/main/java/com/graphhopper/routing/weighting/GenericWeighting.java
boldtrn: > Would you mind to add applying the speed and accessibility changes and recover may some tests that I removed? This seems indeed important to me, now that you pointed me to it

So should I put it back into the GenericWeighting? Or should we keep it in the DataFlagEncoder?

karussell: Please for now without GenericWeighting. So like I linked: determine the country and apply a speed reduction if necessary (similar with the access).

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
```diff
@@ -105,10 +115,10 @@ public SpatialRule lookupRule(double lat, double lon) {
             }
         }
 
-        return EMPTY_RULE;
+        return EMPTY;
     }
 
-    protected int getRuleIndex(int xIndex, int yIndex) {
+    protected int getRuleContainerIndex(int xIndex, int yIndex) {
         if (xIndex < 0 || xIndex > lookupArray.length) {
```
karussell: Well, usually ifs for boundary checks are very fast. For now I wouldn't care that much. But we could check the result (in lookupRule) for EMPTY_RULE_INDEX and return EMPTY instead of the call to ruleContainers etc?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
karussell: Ah, yes, it is the whole lookup

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
karussell: I just picked a very small one - maybe we can increase it with some assumption

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
karussell: Please rename it to EMPTY_RULE if you like. I have no preference :)

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
boldtrn: No, that's fine. Was just wondering.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
boldtrn: Yeah, let's just leave it as is. If we see any performance issues we can still tweak?

core/src/main/java/com/graphhopper/routing/weighting/GenericWeighting.java
boldtrn: > Would you mind to add applying the speed and accessibility changes and recover may some tests that I removed? This seems indeed important to me, now that you pointed me to it

Added these back to the DataFlagEncoder.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -296,6 +325,10 @@ public long handleWayTags(ReaderWay way, long allowed, long relationFlags) {
 
             // MAXSPEED
             double maxSpeed = parseSpeed(way.getTag("maxspeed"));
+            if (spatialRuleLookupEnabled() && maxSpeed < 0) {
+                // TODO What if no maxspeed is set, but only forward and backward, and both are higher than the usually allowed?
+                maxSpeed = getSpatialRule(way).getMaxSpeed(way.getTag("highway", ""), maxSpeed);
+            }
             double fwdSpeed = parseSpeed(way.getTag("maxspeed:forward"));
             if (fwdSpeed < 0 || maxSpeed > 0 && maxSpeed < fwdSpeed)
```
boldtrn: Is fwdSpeed > maxSpeed an iillegal/unwanted state? From a mapping perspective it shouldn't matter, right?

boldtrn:
Ok, so I updated this PR. Are we good to merge it or is there anything we should add/change?

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
karussell: It should be allowed, yes.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -418,6 +427,25 @@ public long handleWayTags(ReaderWay way, long allowed, long relationFlags) {
...
+            if (spatialRuleLookup == null)
+                throw new IllegalStateException("This encoder was asked to store spatial IDs for every edge, " +
+                        "but no spatial lookup was specified");
+
+            return true;
+        }
+        return false;
+    }
+
+    private SpatialRule getSpatialRule(ReaderWay way) {
```
karussell: Where is this method used?

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
```diff
@@ -346,7 +345,11 @@ public void testSpatialId() {
         e1.setFlags(encoder.handleWayTags(way, 1, 0));
         e2.setFlags(encoder.handleWayTags(way2, 1, 0));
 
-        assertEquals(encoder.getSpatialId(e1.getFlags()), encoder.getSpatialId(e1.getFlags()));
-        assertEquals(index.getSpatialId(EMPTY), encoder.getSpatialId(e2.getFlags()));
+        assertEquals(index.getSpatialId(new GermanySpatialRule()), encoder.getSpatialId(e1.getFlags()));
+        assertEquals(index.getSpatialId(SpatialRule.EMPTY), encoder.getSpatialId(e2.getFlags()));
+
+        assertEquals(AccessValue.EVENTUALLY_ACCESSIBLE, encoder.getAccessValue(e1.getFlags()));
+        assertEquals(AccessValue.ACCESSIBLE, encoder.getAccessValue(e2.getFlags()));
```
karussell: Would you mind to add a further test where the max_speed value is overwritten by the one stored in the index?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
```diff
@@ -0,0 +1,76 @@
...
+    List<Polygon> getBorders();
+
+    SpatialRule setBorders(List<Polygon> borders);
+
+    SpatialRule addBorder(Polygon polygon);
+
+    /**
+     * Get the unique name for this rule, e.g. the ISO name of the country.
+     */
+    String getUniqueName();
```
karussell: rename to getId?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+        for (SpatialRule spatialRule : rules) {
+            if (!spatialRule.getBorders().isEmpty()) {
+                for (Polygon polygon : spatialRule.getBorders()) {
+                    polygonBounds.update(polygon.getMinLat(), polygon.getMinLon());
+                    polygonBounds.update(polygon.getMaxLat(), polygon.getMaxLon());
+                }
+            }
+        }
+
+        if (polygonBounds.minLon == Double.MAX_VALUE) {
```
karussell: here it is safer to use !polygonBounds.isValid() instead

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    void sout(int stepSize) {
+        for (int i = 0; i < lookupArray.length; i += stepSize) {
+            for (int j = 0; j < lookupArray[0].length; j += stepSize) {
+                System.out.print(lookupArray[i][j]);
+            }
+            System.out.println();
+        }
+    }
```
karussell: As you said this is no longer used/needed: would you mind to remove it?

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GermanySpatialRule.java
karussell: This is fine but should be probably explicitly mentioned in the javadocs of the interface?

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -82,9 +86,13 @@
     private EncodedValue highwayEncoder;
     private EncodedValue transportModeEncoder;
     private EncodedValue accessEncoder;
+    private static int NOT_ACCESSIBLE_KEY;
```
boldtrn: That variable wasn't used :)

boldtrn: Somehow I cannot comment on your comment below.

> Can it be removed?

Yes, but it is used two times in the code. This would most probably lead to duplicated code. So I wouldn't want to do this, only if we have to?

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -296,15 +325,19 @@ public long handleWayTags(ReaderWay way, long allowed, long relationFlags) {
 
             // MAXSPEED
             double maxSpeed = parseSpeed(way.getTag("maxspeed"));
+            if (spatialRuleLookupEnabled() && maxSpeed < 0) {
+                // TODO What if no maxspeed is set, but only forward and backward, and both are higher than the usually allowed?
+                maxSpeed = getSpatialRule(way).getMaxSpeed(way.getTag("highway", ""), maxSpeed);
+            }
             double fwdSpeed = parseSpeed(way.getTag("maxspeed:forward"));
-            if (fwdSpeed < 0 || maxSpeed > 0 && maxSpeed < fwdSpeed)
```
boldtrn: Yes you are right. Added another test.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
boldtrn: Yes, at first I did not want to call it id, but this should be easier to understand and also imply that it should be unique. But I think it makes sense to rename it.

boldtrn:
I updated this PR and added JavaDoc to the Interfaces. I realized that passing a string as TransportationMode is not a good idea and changed it to be an enum (sorry for the last minute change). I feel like it makes more sense like that. This enum could be used in the DataFlagEncoder to differ between different TransportationModes as well.

karussell:
Nice - Thanks!

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
karussell: > Yes, but it is used two times in the code. This would most probably lead to duplicated code. So I wouldn't want to do this, only if we have to?

Ok

core/src/main/java/com/graphhopper/routing/util/spatialrules/Polygon.java
```diff
@@ -0,0 +1,182 @@
...
+ */
+package com.graphhopper.routing.util.spatialrules;
+
+import com.graphhopper.util.shapes.GHPoint;
+
+/**
+ * This class represents a polygon that is defined by a set of points.
+ * Every point i is connected to point i-1 and i+1.
+ * <p>
+ * TODO: Howto design inner rings in the polygon?
```
karussell: Or we do this here #956

Issue #913, opened 2016-12-17
Allow to return turn instructions without street names


boldtrn:
Right now GH returns the translated instruction like "Turn Left onto <Road Name>". However, there are cases when a user only wants to get simple instructions like "Turn Left", "Turn Right". Currently, this is achievable with GH parsing either the String or the sign and create turn instructions using a local TranslationMap. 

Thus, I propose to add a request parameter that allows showing or hiding the street name in the instruction list.

WDYT?

karussell:
What would be a use case behind this?

boldtrn:
Having simpler/shorter turn instructions. One reason would be space. 

For example I want to create a condensed turn instruction list, containing the symbol and the instruction, without the street name.

karussell:
You could combine the sign with the simple instruction string?

boldtrn:
> simple instruction string?

What do you mean by simple instruction string?

Right now an instruction looks like this:
```
distance: 911.477,
sign: -2,
interval: [
19,
27
],
text: "Turn left onto K 3955",
time: 65624
```

devemux86:
I think this happens due to [Instruction.setUseRawName](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/util/Instruction.java#L60) called in e.g. [GraphHopperWeb](https://github.com/graphhopper/graphhopper/blob/master/web/src/main/java/com/graphhopper/http/GraphHopperWeb.java#L150).
That forbids to use properly the [Instruction.getTurnDescription](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/util/Instruction.java#L228) for more customization.

An idea could be to replace the `setUseRawName()` with `setUseRawName(boolean)` allowing users to enable / disable this feature.

karussell:
Ah, yes. In the web UI this is not possible but via Java API it should via calling the translation map without the street name.

devemux86:
Hmm even in directions-api it's not possible as the `setUseRawName()` cannot be undone, or you mean something different?

karussell:
I mean just calling the translationMap.tr("turn_left") etc depending on the returned "sign".

devemux86:
Ah yes already doing that! :slightly_smiling_face:

boldtrn:
> Ah, yes. In the web UI this is not possible

We could add the turn translations to the web translations. This way this could be also done in the web UI.

devemux86:
[...]

PR #927, opened 2016-12-30
Added Exited to the json and parse it accordingly

boldtrn:
This PR adds the `exited` variable to the web api. This allows to differentiate between a used roundabout and an exited roundabout instruction. The change to GraphHopperWeb has to be done in the Java Client as well. I will create a PR there once this is merged.

I think we should also think about deleting the [LEAVE_ROUNDABOUT](https://github.com/graphhopper/graphhopper/blob/de2703820870fa30fb1bb1a514cb26d781464022/core/src/main/java/com/graphhopper/util/Instruction.java#L25) instruction, as it is not used. On the other hand we could use it, but this might break existing cllients.

@devemux86 this should solve the issue?


karussell:
I would avoid (if possible) to make internal stuff like `exited` public. Why is this really necessary?

Regarding the LEAVE_ROUNDABOUT instruction: this is true and I'm not sure how we should handle this. Maybe we look into the JS&Java client and if they would break without a change. For future versions we should make adding instructions possible and return defaults.

devemux86:
@boldtrn yes that could work!

>I would avoid (if possible) to make internal stuff like exited public. Why is this really necessary?

This is supposed to help in #913.

Essentially in web we cannot get instructions without street names (as in Java).
And trying to create them via their signs, have found that (web) Roundabout variables are wrong.

karussell:
> Essentially in web we cannot get instructions without street names (as in Java).

I do not understand why this is useful for a broader range of applications (?)

> And trying to create them via their signs, have found that (web) Roundabout variables are wrong.

What do you mean here?

devemux86:
>I do not understand why this is useful for a broader range of applications (?)

It's convenient if the API is flexible, e.g. text-to-speech has funny results with names. :slightly_smiling_face: 

>What do you mean here?

I had described the situation in #913, specifically see [here](https://github.com/graphhopper/graphhopper/issues/913#issuecomment-269453914).

karussell:
Ok, understood the text-to-speech argument but would like to know more details. How exactly does the workaround work, do you feed the turn instruction to the engine with different locales than the street?

> I had described the situation in #913, specifically see here.

This sounds like a bug which we must fix

devemux86:
>How exactly does the workaround work, do you feed the turn instruction to the engine with different locales than the street?

There actually two cases here:
- Text-to-speech (Google or another) does not exist in all languages, so may need a different or fallback locale
- But even if it does exist, speech of (local) street names is usually not good, except e.g. English TTS with English names

> This sounds like a bug which we must fix

Note that in Java client we are able to use `instruction.setName("")` and disable temporarily the name.
That would be really convenient in web too, not possible right now (see [here](https://github.com/graphhopper/graphhopper/issues/913#issuecomment-267810755)).
So the other solution seems being to recreate the instructions.

boldtrn:
> I would avoid (if possible) to make internal stuff like exited public. Why is this really necessary?

Isn't this the same as LEAVE_ROUNDABOUT? Whereas if exited=false is USE_ROUNDABOUT? Making this public allows for example the java-api-client to behave exactly the same as working with instructions in an offline graph. IMHO, the client should behave the same and if using the same Objects, e.g. Instructions, all properties should be the same. What do you think? Does this make sense?

karussell:
Interesting. If a via point is in the roundabout then exit_number for the "enter roundabout" instruction depends on where the via point is (which is logical) and so we need indeed a separate variable indicating "exited" status. Also this then works already when we have multiple via points in a roundabout

devemux86:
Hmm with `>1` via points in a roundabout, then all `n-1` via points have as instruction "Enter roundabout".
e.g. see [here](https://graphhopper.com/maps/?point=52.512346%2C13.319228&point=52.512088%2C13.321309&point=52.512019%2C13.322071&point=52.512281%2C13.322752&point=52.512708%2C13.32456&locale=en-US&vehicle=car&weighting=fastest&elevation=true&key=2a24e316-61ea-4850-b231-4ef2fe25d229&use_miles=false&layer=Omniscale)

karussell:
Still the exit_number is in the response and IMO correctly set

PR #928, opened 2016-12-30
check for node based traversal in DijkstraOneToMany

kodonnell:
As discussed [here](https://discuss.graphhopper.com/t/turn-cost-support-in-dijkstraonetomany/1353).


karussell:
This is no longer necessary as #1247 is already in a mergeable state.

Issue #649, opened 2016-01-27
Alternative route improvements


karussell:
We've introduced alternative routing via #640 still there are the following open issues:
- [x] make it possible for CH
- [ ] further performance improvements ('Improved Alternative Route Planning' paper), the alternative calculation itself takes only a few percentage of the whole search
- [ ] post processing alternatives to throw away too bad alternatives (e.g. alternatives that are too similar to other alternatives) was necessary in this local query](for final alternative list: remove last alternatives if too much share from previous like necessary [here](http://localhost:8989/?point=49.942825%2C11.576843&point=50.002443%2C11.618557&vehicle=car&alternative_route.max_paths=2&algorithm=alternativeroute) or [here](http://localhost:8989/?point=49.942355%2C11.576908&point=49.482401%2C11.013794&vehicle=car&alternative_route.max_paths=5&algorithm=alternativeroute)
- [ ] can we avoid alternatives with motorways just going back and forth? [local query](http://localhost:8989/?point=48.487486%2C10.118408&point=47.661688%2C10.000305&vehicle=car&alternative.max=3&algorithm=alternativeroute), [query 2](http://localhost:8989/?point=49.942825%2C11.576843&point=50.035753%2C11.681213&vehicle=car&alternative_route.max_paths=3&algorithm=alternativeroute&alternative_route.max_share=0.4)
- [ ] integrate multiple alternatives into GPX with multiple entries -> GPX export in josm works this way
- [ ] test query torture for new log layout (necessary due to refactoring and new log format)
- [ ] why sometimes no alternatives at all? [example](http://localhost:8989/?point=49.942825%2C11.576843&point=50.059563%2C11.67778&vehicle=car&alternative_route.max_paths=5&algorithm=alternativeroute&alternative_route.max_share=0.4)
- [ ] let influence the SPT-counter (see alternative roads to rome post) the selection of a good plateau? e.g. sum of weight_i*count_i for just the plateau (?) but this seems to be too expensive to calculate!

boldtrn:
Good points!

> integrate multiple alternatives into GPX with multiple entries -> GPX export in josm works this way

Do you think it would also make sense to add a parameter like `gpx_export.use_route=ROUTE_NUMBER`? In my experience importing multiple GPX routes does not work as expect, as most navigation devices will just use the first route or try to concat the routes. I think that could be a fast fix for the GPX export.

devemux86:
[...]

PR #929, opened 2017-01-05
Allow exporting alternative routes

boldtrn:
As discussed in #649, exporting alternative_routes is not possible yet. Therefore, I added a parameter to allow exporting a single path of the alternative routes.

WDYT?

### DIFF
--- a/core/src/main/java/com/graphhopper/util/Parameters.java
+++ b/core/src/main/java/com/graphhopper/util/Parameters.java
@@ -17,8 +17,6 @@
  */
 package com.graphhopper.util;
 
-import static com.graphhopper.util.Parameters.Algorithms.ALT_ROUTE;
-
 /**
  * @author Peter Karich
  */
@@ -172,6 +170,6 @@ public class Parameters {
         //TODO rename to gpx.trackname?
         public static final String TRACK_NAME = "trackname";
 
-        public static final String PATH_INDEX = "gpx."+ALT_ROUTE+".path_index";
+        public static final String PATH_INDEX = "gpx." + Parameters.Algorithms.ALT_ROUTE + ".path_index";
     }
 }
--- a/core/src/main/java/com/graphhopper/util/Parameters.java
+++ b/core/src/main/java/com/graphhopper/util/Parameters.java
@@ -17,6 +17,8 @@
  */
 package com.graphhopper.util;
 
+import static com.graphhopper.util.Parameters.Algorithms.ALT_ROUTE;
+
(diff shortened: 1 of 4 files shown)


karussell:
Sorry, that I answer a bit too late ... saw your comment in #649 ... I would prefer to return multiple alternatives in one GPX to save resources as with this approach multiple requests are necessary for multiple alternatives. And to keep the API stable return multiple results **only** if alternatives are requested.

devemux86:
Aren't we talking about an extra option to select the exported path?
Current behavior doesn't change, it's all about provide better API to the user.

karussell:
> Current behavior doesn't change, it's all about provide better API to the user.

Alternatives are currently not allowed for gpx export, so this is an API change and the usage is suboptimal in the explained ways which we could avoid IMO

boldtrn:
> Alternatives are currently not allowed for gpx export, so this is an API change

Yes, that is true, but it does not break any existing behavior.

> the usage is suboptimal 

Actually, I don't think so. There are scenarios where only a single exported path is preferred to multiple exported paths. See for example my comment in #649.

I still think that when trying to export an `alternative_route` without a `path_index` all paths should be exported. But this is something we can implement in the future?

karussell:
The problem is that this export requires an up front request with the JSON response to determine the number of alternatives and so it is stateful and every follow up request issues an expensive alternative route search (or did I misunderstand this?)

karussell:
In this state I would not merge it. There must be something better, like a gpx export with all alternatives. I know many clients would not support this but the client side could pick the required alternative (now in GPX format), ie. there would be just one underlying routing request.

boldtrn:
>  it is stateful and every follow up request issues an expensive alternative route search

This is correct yes.

> There must be something better, like a gpx export with all alternatives

The problem is that when importing the track (or route) most navis don't understand importing alternative routes, so it would not make sense to do this. Maybe we could create n gpx files and zip them, but this is also not a good solution...

> but the client side could pick the required alternative

But this requires xml parsing and writing logic on the client side, which I personally would like to avoid.

karussell:
> But this requires xml parsing and writing logic on the client side, which I personally would like to avoid.

Yes, sure. Both is bad. Another option could be to return multiple GPX strings within a JSON ;)

boldtrn:
> Another option could be to return multiple GPX strings within a JSON ;)

Not sure if I want to do this either ;).

Let's leave it closed for now, maybe we find a better solution in the future.

PR #932, opened 2017-01-12
Improve javadoc comments

stefanholder:
The javadoc changes reflect my learnings about Graphhopper internals when I was implementing graphhopper/map-matching#88.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/QueryGraph.java
+++ b/core/src/main/java/com/graphhopper/routing/QueryGraph.java
@@ -43,6 +43,11 @@ import java.util.*;
  * introducing virtual nodes and edges. It is lightweight in order to be created every time a new
  * query comes in, which makes the behaviour thread safe.
  * <p>
+ * Calling any <tt>lookup</tt> method creates virtual edges between the tower nodes of the existing
+ * graph and new virtual tower nodes. Every virtual node has two adjacent nodes and is connected
+ * to each adjacent nodes via 2 virtual edges with opposite base node / adjacent node encoding.
+ * However, the edge explorer returned by {@link #createEdgeExplorer()} only returns two
+ * virtual edges per virtual node (the ones with correct base node).
  *
  * @author Peter Karich
  */
@@ -57,10 +62,8 @@ public class QueryGraph implements Graph {
     private final GraphExtension wrappedExtension;
     // TODO when spreading it on different threads we need multiple independent explorers
     private final Map<Integer, EdgeExplorer> cacheMap = new HashMap<Integer, EdgeExplorer>(4);
-    /**
-     * Virtual edges are created between existing graph and new virtual tower nodes. For every
-     * virtual node there are 4 edges: base-snap, snap-base, snap-adj, adj-snap.
-     */
+
+    // For every virtual node there are 4 edges: base-snap, snap-base, snap-adj, adj-snap.
     List<VirtualEdgeIteratorState> virtualEdges;
     private List<QueryResult> queryResults;
     /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/graphhopper/routing/QueryGraph.java
```diff
@@ -694,6 +702,9 @@ private boolean isInitialized() {
     }
 
     @Override
+    /**
+     * @see QueryGraph
```
karussell: What do you mean here?

core/src/main/java/com/graphhopper/storage/index/QueryResult.java
```diff
@@ -49,8 +51,10 @@ public QueryResult(double queryLat, double queryLon) {
     }
 
     /**
-     * @return the closest matching node. -1 if nothing found, this should be avoided via a call of
-     * 'isValid'
+     * Returns the closest matching node. This is either a tower node of the base graph
+     * or a virtual node (see also {@link com.graphhopper.routing.QueryGraph#lookup(List)}).
```
karussell: Hmmh, not that happy with this explicit mention of virtual node here. Everytime the QueryGraph and not just the base graph is involved we have to make this distinction e.g. also in all algorithms etc.

core/src/main/java/com/graphhopper/routing/QueryGraph.java
stefanholder: The new class comment also describes what the lookup methods are doing, so I made a reference here.

core/src/main/java/com/graphhopper/storage/index/QueryResult.java
```diff
@@ -49,8 +51,10 @@ public QueryResult(double queryLat, double queryLon) {
...
+     *
+     * @return -1 if nothing found, this should be avoided via a call of 'isValid'
```
karussell: We could throw an exception instead? (more safety, less backward compatibility)

core/src/main/java/com/graphhopper/routing/QueryGraph.java
karussell: But how is createEdgeExplorer related to QueryGraph?

core/src/main/java/com/graphhopper/storage/index/QueryResult.java
stefanholder: But shouldn't the javadoc precisely say what this method returns in which situations? Otherwise developers like me don't know what to expect.

core/src/main/java/com/graphhopper/storage/index/QueryResult.java
stefanholder: Sure but for now I just wanted to improve the javadoc without changing the behavior.

core/src/main/java/com/graphhopper/routing/QueryGraph.java
stefanholder: Sorry, I confused this method with `lookup`. The behavior of the edge explorer is also explained in the class comment:

>However, the edge explorer returned by {@link #createEdgeExplorer()} only returns two virtual edges per virtual node (the ones with correct base node).

core/src/main/java/com/graphhopper/storage/index/QueryResult.java
karussell: The developer shouldn't have to take care about this, but knows in general what the QueryGraph does (virt+real). We can keep it here but in general I would not want to add this separate meaning everywhere.

stefanholder: I agree that we don't have a mention every time that virtual nodes might be involved such as in the routing algorithms. But I think in this case it's important to explain the connection to QueryGraph.lookup since this changes the behavior of this method (a virtual node may then be returned).

Issue #933, opened 2017-01-13
Android: Clean mapped byte buffers fails


devemux86:
On Android the clean of mapped byte buffers fails with:

```java
Caused by: java.lang.RuntimeException: unable to unmap the mapped buffer
   at com.graphhopper.util.Helper.cleanMappedByteBuffer(Helper.java:426)
   at com.graphhopper.storage.MMapDataAccess.clean(MMapDataAccess.java:356)
   at com.graphhopper.storage.MMapDataAccess.close(MMapDataAccess.java:257)
   at com.graphhopper.storage.MMapDataAccess.close(MMapDataAccess.java:249)
   at com.graphhopper.storage.StorableProperties.close(StorableProperties.java:113)
   at com.graphhopper.storage.GraphHopperStorage.close(GraphHopperStorage.java:277)
   at com.graphhopper.GraphHopper.close(GraphHopper.java:1226)
   at com.graphhopper.android.MainActivity.onDestroy(MainActivity.java:170)
   at android.app.Activity.performDestroy(Activity.java:6881)
   at android.app.Instrumentation.callActivityOnDestroy(Instrumentation.java:1153)
   at android.app.ActivityThread.performDestroyActivity(ActivityThread.java:4190)
   at android.app.ActivityThread.handleDestroyActivity(ActivityThread.java:4221) 
   at android.app.ActivityThread.-wrap6(ActivityThread.java) 
   at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1538) 
   at android.os.Handler.dispatchMessage(Handler.java:102) 
   at android.os.Looper.loop(Looper.java:154) 
   at android.app.ActivityThread.main(ActivityThread.java:6119) 
   at java.lang.reflect.Method.invoke(Native Method) 
   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
   at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
Caused by: java.security.PrivilegedActionException: java.lang.NoSuchMethodException: free []
   at java.security.AccessController.doPrivileged(AccessController.java:71)
   at com.graphhopper.util.Helper.cleanMappedByteBuffer(Helper.java:370)
   at com.graphhopper.storage.MMapDataAccess.clean(MMapDataAccess.java:356) 
   at com.graphhopper.storage.MMapDataAccess.close(MMapDataAccess.java:257) 
[...]

PR #934, opened 2017-01-13
Improve cleaning of memory mapped ByteBuffers for JDK9 EA and Android

karussell:
This should pass the JDK9 travis tests once we have JDK9-ea-152 there, see #701.

And regarding the Android part, see #933


karussell:
I merge this issue now, but keep #933 open until it is clear if this is fixed.

PR #936, opened 2017-01-16
Support 'height', 'weight', 'width' road attributes

tmnguyen-swin:
Unrecognized values will be logged and ignored.


core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -327,6 +357,96 @@ public long handleWayTags(ReaderWay way, long allowed, long relationFlags) {
...
+
+        for (String key : keys) {
+            value = way.getTag(key);
+            if (!Helper.isEmpty(value)) break;
+        }
+
+        return value;
+    }
+
+    double stringToTons(String value) {
```
boldtrn: Should we consider other measurement units, like kg?

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -327,6 +357,96 @@ public long handleWayTags(ReaderWay way, long allowed, long relationFlags) {
...
+        } else if (value.endsWith("lbs")) {
+            value = value.substring(0, value.length() - 3);
+            factor = 0.00045359237;
+        }
+
+        return Double.parseDouble(value) * factor;
+    }
+
+    double stringToMeter(String value) {
+        value = value.toLowerCase().replaceAll(" ", "").replaceAll("(meters|meter|mtrs|mtr|mt|m\\.)", "m");
```
boldtrn: According to the [Wiki](http://wiki.openstreetmap.org/wiki/Map_Features/Units) these vales might also happen: `	metre, metres, meter, meters`. Couldn't find them via Taginfo, so not sure if this is relevant.

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
```diff
@@ -24,9 +27,11 @@
     private final EncodingManager encodingManager;
     private final int motorVehicleInt;
 
+    private final double DELTA = 0.1;
+
     public DataFlagEncoderTest() {
         encoder = new DataFlagEncoder();
-        encodingManager = new EncodingManager(encoder);
+        encodingManager = new EncodingManager(Arrays.asList(encoder), 8);
```
boldtrn: Would be great if we could make the number of encoders configureable. Maybe if we pass a heightlimit of 0 via the Configmap to the DataflagEncoder, we don't register the heigthEncoder. WDYT?

core/src/main/java/com/graphhopper/GHRequest.java
```diff
@@ -240,6 +240,10 @@ public HintsMap getHints() {
         return hints;
     }
 
+    public void setHints(HintsMap map) {
```
karussell: Couldn't you use getHints().putAll or something?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1121,6 +1121,10 @@ public ChangeGraphResponse changeGraph(Collection<JsonFeature> collection) {
         }
     }
 
+    public String getBlockedRectangularAreas() {
+        return blockedRectangularAreas;
```
karussell: Do we really need to expose this :) ? I would like to hide it as long as possible as the API for the blocked areas is not really stable and e.g. the String could change to a domain object etc to be more accessible. Also as this is not related to this PR we should move this to another issue?

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
karussell: Wiki says it is tons ... but [here](https://karussell.wordpress.com/2015/11/01/units-in-openstreetmap/) is the real fun with units in OSM ;)

We can contribute our slightly more complex parsing methods from the commercial version.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -327,6 +357,96 @@ public long handleWayTags(ReaderWay way, long allowed, long relationFlags) {
...
+            val = stringToTons(value);
+        } catch (Throwable t) {
+            LOG.warn("Unable to extract tons from malformed road attribute '{}' for way (OSM_ID = {}).", value, way.getId(), t);
+            val = 0.0;
+        }
+
+        return valueEncoder.setDoubleValue(flags, val);
+    }
+
+    private String extractRoadAttribute(ReaderWay way, List<String> keys) {
```
karussell: let us keep it here for now

> Also if two several keys exist, only the last one is returned. Not sure if this is an issue

I think this is not a problem yet. (BTW: the first vs. last?)

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -575,9 +710,26 @@ public ConfigMap readStringMap(PMap weightingMap) {
...
         return cMap;
     }
 
+    private void cloneDoubleAttribute(PMap weightingMap, ConfigMap cMap, String key, double _default) {
+        if (weightingMap.has(key))
+            cMap.put(key, weightingMap.getDouble(key, _default));
+
+    }
+
+    protected void cloneStringAttribute(PMap weightingMap, ConfigMap cMap, String key, String _default) {
```
karussell: Where is this used and why protected access?

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
karussell: Yes, this would be great. 

The real problem is that we are currently limited to 64 bit, where have an idea to make this kind of 'unbounded', but no time :/ ... see the [edge.getData(int) idea here](https://github.com/graphhopper/graphhopper/pull/798#issuecomment-248088079).

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -156,6 +164,18 @@ public int defineWayBits(int index, int shift) {
         carBwdMaxspeedEncoder = new EncodedDoubleValue("car bwd maxspeed", shift, speedBits, speedFactor, 0, maxPossibleSpeed, true);
         shift += carBwdMaxspeedEncoder.getBits();
 
+        /* Value range: [3.0m, 5.4m] */
```
karussell: Here (maybe in a later version) we could do kind some kind of a compression: i.e. assign 

 * 2.5m to index 0
 * 3.0m to index 1
 * 3.1m to index 2

... so we could make it more precise with less bits and for the ones that do not fit we would still round. For this taginfo is useful: http://taginfo.openstreetmap.org/keys/height#values ... it seems 2m is very important for the world wide case. (Or even more sophisticated and area dependent and create statistics in the first parse run and then pick the best compression automatically)

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
tmnguyen-swin: I think the first available value will be picked up. The intention is to have the list of semantically equivalent tags in descending order of preferences.

core/src/main/java/com/graphhopper/GHRequest.java
tmnguyen-swin: Spot on. I actually had the same thought yesterday :) Will remove this method.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
tmnguyen-swin: I have a custom FlagEncoder in my project that subclasses DataFlagEncoder. I needed this method there, so make it 'protected'. The method is actually not used at all in DataFlagEncoder (didn't realize this). Will move the method to my subclass.

core/src/main/java/com/graphhopper/GraphHopper.java
tmnguyen-swin: I need to access 'blockedRectangularAreas' in my subclass of GraphHopper. But I agree with you, this is not related to this PR. Will cancel this change.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
tmnguyen-swin: Yes, the current implementation is quite simple and is based on the value set for the particular area that I have. The value set was obtained from a source independent of OSM that I managed to incorporate into OSM using an internal data ingestion pipeline. As such, the value format is very good (as it is under my control :).  I hope we can have more sophisticated implementation which is applicable to world wide data in later version.

For this particular PR, what I can do is investigating the value range that I can obtain from taginfo (Thanks for the reference - a great website). And see if we need more bits for the encoder and if I can enhance the parsing logic.

tmnguyen-swin: Hmm...The issue with taginfo data is we don't know which values are for road attributes, as opposed to, for instance, building height.

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
tmnguyen-swin: Sorry guys, didn't get the point here. What do you suggest in terms of configurable number of encoders? I understand the current limit of 32 bit or 64 bit for flags.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
tmnguyen-swin: Would be really nice if you can contribute the commercial version here!

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
tmnguyen-swin: I think I leave the current encoders for road attributes as-is, with only one minor improvement in the latest commit. If the value is too big for the encoder, we just log & ignore, and continue parsing the OSM data.

Given this change, I was able to parse Australian OSM with only few exceptions of "too large value".

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
karussell: What I meant is that in some use cases one might not need to disable this feature in order to save bits. But something like this might be interesting for other encoded values too and a different issue could be more appropriated.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
karussell: > we just log & ignore, and continue parsing the OSM data.

If it is bigger than the accepted range we could ignore but maybe it is safer to still include it.

If it is smaller we have to throw an exception or warn the user **and** include it as a restriction (even it is wider)?

karussell:
When look into the responsibility the OSM related parsing and converting should be done in the OSM reader module not in the encoder. But when I move it there we have to convert regardless if it is enabled or not (e.g. using generic or not). Hmmh, interesting ;)

karussell:
But this is worth a separate issue #939

karussell:
Interesting. I was able to push to the road_attributes branch of Transtech/graphhopper ... even a force push. Seems that github has improved this to work together with different repos on the same pull request, nice. Please have a look- hope I didn't messed up something!

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -358,57 +358,43 @@ public long handleWayTags(ReaderWay way, long allowed, long relationFlags) {
...
+            return flags;
         }
-
-        return flags;
     }
 
     private long extractTons(ReaderWay way, long flags, EncodedDoubleValue valueEncoder, List<String> keys) {
-        String value = extractRoadAttribute(way, keys);
-
+        String value = way.getFirstPriorityTag(keys);
```
karussell: This should be the same functionality (do we have a test covering a change like this?)

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -420,11 +406,15 @@ private String extractRoadAttribute(ReaderWay way, List<String> keys) {
         return Double.parseDouble(value) * factor;
     }
 
-    double stringToMeter(String value) {
+    public static double stringToMeter(String value) {
```
karussell: Made it public to reuse it elsewhere ... before I thought putting it into OSMTagParser but this does not work, see #939

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
tmnguyen-swin: Got it now. I've just finished a commit that makes encoders for road attributes configurable. Will merge with Peter's change before push.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
tmnguyen-swin: In many cases, the value is too big for the range as it implies something else rather than road restrictions. For instance, a bridge section over a river has 'height' attribute set to 60m. This value is actually the height from the river ground to the bridge. 

Btw, I tested with germany OSM and there were 77 exceptions due to value being too large.

java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: weight restriction value too large for encoding: 1000000, maxValue:100
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 16, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 24, maxValue:12
java.lang.IllegalArgumentException: weight restriction value too large for encoding: 12000, maxValue:100
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 50, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 50, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 10, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 10, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 10, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 7, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 70, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 16, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 30, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 16, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 9, maxValue:6
java.lang.IllegalArgumentException: weight restriction value too large for encoding: 5000, maxValue:100
java.lang.IllegalArgumentException: height restriction value too large for encoding: 16, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 16, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 16, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: weight restriction value too large for encoding: 9000, maxValue:100
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 100, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 50, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 50, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 10, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 7, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: weight restriction value too large for encoding: 30000, maxValue:100
java.lang.IllegalArgumentException: width restriction value too large for encoding: 7, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 30, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 12, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 10, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 220, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 50, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 100, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 60, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 50, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 50, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 30, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 10, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 100, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: width restriction value too large for encoding: 8, maxValue:6
java.lang.IllegalArgumentException: height restriction value too large for encoding: 60, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 30, maxValue:12
java.lang.IllegalArgumentException: height restriction value too large for encoding: 279, maxValue:12
java.lang.IllegalArgumentException: width restriction value too large for encoding: 50, maxValue:6
tuan@tuan-VirtualBox:/mnt/data/transtech/geo-engine$ cat maven.log | grep "too large" | wc -l
77
tuan@tuan-VirtualBox:/mnt/data/transtech/geo-engine$ cat maven.log | grep "too large" | grep "height" | wc -l
41
tuan@tuan-VirtualBox:/mnt/data/transtech/geo-engine$ cat maven.log | grep "too large" | grep "weight" | wc -l
5
tuan@tuan-VirtualBox:/mnt/data/transtech/geo-engine$ cat maven.log | grep "too large" | grep "width" | wc -l
31

boldtrn: > For instance, a bridge section over a river has 'height' attribute set to 60m.

Does it have a height or maxheight of 60m? Height would be correctly mapped. Maxheight would be wrong. If it is maxheight, we could report an error to OSM.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
tmnguyen-swin: Yes. OSMReaderTest#testRoadAttributes should do the job.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
tmnguyen-swin: That's an interesting question. Let's me have a look at the detailed log file. The error message above ignored the actually tag being used.

tmnguyen-swin: You're right Robin. It's maxheight in both cases I found:
https://www.openstreetmap.org/way/31539875
https://www.openstreetmap.org/way/238785424

tmnguyen-swin:
Guys, I have committed a new change that makes road attribute encoders configurable. Merging fine with master. Let me know if there is anything else you would like me to change.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
boldtrn: Thanks for looking into this. I reported an issue for both ways. Both are very small ways, so probably this is just a mapping issue.

boldtrn: Wow, I think this was the fastest fix I've ever seen on OSM, one problem is already fixed: http://www.openstreetmap.org/note/857064

karussell:
Thanks, this looks good.

Let us (shortly) discuss if we need three configurations or if one e.g. `"store_truck_attributes":true` is simpler (are there meaningful use cases where one would just need e.g. the weight?). Also the name is important, I prefer something with 'store_xy' and how the weighting interprets this value is up to the weighting.

Furthermore the boolean methods should be setXY instead of enableXY

> Btw, I have resolved some conflicts when merging Peter's changes. Hopefully I didn't miss out anything :)

Sorry, I feared that. Force push is never a good idea if one works on the same branch with others :(. So instead I should have made my changes and the rebase/squash should have been done as the last step.

karussell:
Would you mind to just rename the methods? We do not need to debate on configuring and should just change it afterwards when necessary (?)

tmnguyen-swin:
I have renamed the methods and variables as requested. This naming convention makes more sense to me now. Please review and advise.

In relation to choosing single configuration field vs multiple configuration fields for road attributes, IMHO, I would go with multiple fields. That would make it more flexible as you would be able to select the attributes of interest, rather than all or nothing. This is especially true when we are still limited to 64 bit length of the flag & the cost of having all fields is expensive.

karussell:
This looks good now - thanks again!

> This is especially true when we are still limited to 64 bit length of the flag & the cost of having all fields is expensive.

Yeah, this is ugly and needs a fix (better sooner than later :))

PR #938, opened 2017-01-18
fix NullPointerException during Import using the BikeCommonFlagEncoder

tobias74:
As it is stated in the AbstractFlagEncoder, the protected method init() should be called as the last action in the constructor.

I was getting NullPointerExceptions on line 278 in the BikeCommonFlagEncoder resulting in "Problem while parsing file"-Exceptions.
The NPE was caused by the "getConditionalTagInspector" returning null.

Calling the init()-Method at the end of the constructor initializes the ConditionalTagInspector. Now I can import again :-)

Hope this helps :-)

Feedback very welcome!

Grüße aus Stuttgart/München :-)

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -195,6 +195,8 @@ public class BikeCommonFlagEncoder extends AbstractFlagEncoder {
         setCyclingNetworkPreference("deprecated", AVOID_AT_ALL_COSTS.getValue());
 
         setAvoidSpeedLimit(71);
+        
+        init();
     }
 
     @Override


tobias74:
no wait... maybe the BikeCommonFlagEncoder should be abstract? The BikeFlagEncoder inherits from this...

tobias74:
Ok, there is a problem with my solution: The BikeFlagEncoder inherits from BikeCommonFlagEncoder, thus the BikeCommonFlagEncoder must not call init(). It seems as if the BikeCommonFlagEncoder should be abstract.

karussell:
> It seems as if the BikeCommonFlagEncoder should be abstract

Probably yes. What do you think @ratrun ?

ratrun:
I agree. BikeCommonFlagEncoder should be abstract.

tobias74:
Shall I issue a new pull request with the abstract BikeCommonFlagEncoder?

tobias74:
favor pull-request "Make BikeCommonFlagEncoder abstract" over this pr. Closing this.

PR #941, opened 2017-01-23
Make BikeCommonFlagEncoder abstract

tobias74:
Flag-Encoders must call init() as their last action in the constructor. Therefore any FlagEncoder that does not call init() must be abstract.


karussell:
Thanks!

PR #946, opened 2017-02-01
Issue 913 street name

### LINKED ISSUE #913: Allow to return turn instructions without street names

boldtrn:
This PR fixes #913.

If a client requests `translate_instructions=false` only the street name will be put into the text field. The default for this parameter is true, so we won't break existing clients. This change also don't require any client changes.

The downside is that I needed to change the method parameters of toJson. Since InstrucitonList.toJson is used heavily in different tests I wrote a wrapper function so I don't need to change all the tests.

**TODOS**: 

- [ ] Update documentation
- [ ] Update java client
- [ ] Update swagger clients

WDYT?


devemux86:
@boldtrn this is because for not having changes on clients?
A more flexible implementation couldn't be having the street name in an extra field?

Also @karussell, @boldtrn is there any thought to even the java and web clients instructions handling?
Right now they behave differently, if web client could follow the java one, this issue would be solved automatically.

karussell:
> A more flexible implementation couldn't be having the street name in an extra field?

Hmmh, I would also favour a separate street field

> Also @karussell, @boldtrn is there any thought to even the java and web clients instructions handling? Right now they behave differently, if web client could follow the java one, this issue would be solved automatically.

I'm not really in this issue - maybe you describe what you mean with code?

devemux86:
> I'm not really in this issue - maybe you describe what you mean with code?

I'm referring to my [comment](https://github.com/graphhopper/graphhopper/issues/913#issuecomment-267810755), where `GraphHopperWeb` constructs differently the instructions than `GraphHopper`, so we cannot have the same flexibility in client side.

boldtrn:
> A more flexible implementation couldn't be having the street name in an extra field?

Yes we could do that indeed. I did these changes with the Java Client in mind. There we are using the Instruction class of GH core. We are converting the result into the Instruction. The instruction class can only contain the name.

So I decided between putting logic to the client or the server and decided it might be better to use the server in this case.

Should I just add the `street_name` property to the response and let the client decide what to do? If we do that, we could also change the Instruction class to contain the turn description as string. This would allow us to construct an instruction containing the already translated instruction or create it on the fly as we do right now.

> is there any thought to even the java and web clients instructions handling?

Absolutely, I think this is very important. Currently, the only difference is with the raw name, is that correct? With the changes provided in this PR these differences would be solved. But could be also solved by just adding another street_name property.

devemux86:
How about having `GraphHopperWeb` like `GraphHopper`, that is **not** force [Instruction.setUseRawName](https://github.com/graphhopper/graphhopper/blob/master/web/src/main/java/com/graphhopper/http/GraphHopperWeb.java#L153-L155)?
Or at least make it optional for compatibility.
That way we'll have same flexibility in the client, like with offline graphs at instructions modifications.
Then no additional work is needed with extra fields?

A less optimal solution could be the addition of extra street name property.
(but my vote would be on above suggestion) :slightly_smiling_face:

boldtrn:
Ok I updated this PR the logic is now done on the client side. The solution might be not as optimal for the Java client, but I think is way better for other clients that for example use Swagger.

> How about having GraphHopperWeb like GraphHopper, that is not force Instruction.setUseRawName?

I changed that now. Would you mind having a look at [this test](https://github.com/graphhopper/graphhopper/pull/946/files#diff-4539e8d6b8028220a83c1c67cbb29265R152). It shows how getName actually returns the street name but turnDescriptions are translateable.

@karussell Should we remove the translation of the turn description again, since I load the translation map from disk again, as I cannot access the TranslationMapTest.Singleton? It shouldn't slow down the test execution too much.

I think this fixes the differences between GraphHopperWeb and GraphHopper. Can you confirm that @devemux86 ?

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
```diff
@@ -138,6 +140,23 @@ public void testGraphHopperWeb() throws Exception {
...
+        assertTrue(hopper.load(getTestRouteAPIUrl()));
+        GHResponse rsp = hopper.route(new GHRequest(42.554851, 1.536198, 42.510071, 1.548128));
+        assertEquals("Continue onto Carrer Antoni Fiter i Rossell", rsp.getBest().getInstructions().get(2).getName());
+
+        GHRequest request = new GHRequest(42.554851, 1.536198, 42.510071, 1.548128);
+        request.getHints().put("translate_instructions", false);
+        rsp = hopper.route(request);
+        assertEquals("Carrer Antoni Fiter i Rossell", rsp.getBest().getInstructions().get(2).getName());
+        // TODO I think we should remove this later on?
+        TranslationMap tr = new TranslationMap().doImport();
```
devemux86: Couldn't `TranslationMap` be retrieved from [GraphHopper](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/GraphHopper.java#L549)?

devemux86:
@boldtrn :+1: 
Work with instructions can be done now more conveniently in web client too.

boldtrn:
@karussell any news on this PR? Is there anything I should change or does it look good to you?

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
```diff
@@ -138,6 +140,23 @@ public void testGraphHopperWeb() throws Exception {
...
+        // TODO First character is small, this is different from web... Maybe we should return the first char as upper case with getTurnDescription?
+        assertEquals("continue onto Carrer Antoni Fiter i Rossell", rsp.getBest().getInstructions().get(2).getTurnDescription(tr.getWithFallBack(Locale.UK)));
```
karussell: It is a bit strange to fetch translated instructions when instructions are disabled so either we should remove this test or throw an exception?

web/src/main/java/com/graphhopper/http/GraphHopperWeb.java
```diff
@@ -152,7 +154,9 @@ public static PathWrapper createPathWrapper(JSONObject path,
 
                     // The translation is done from the routing service so just use the provided string
                     // instead of creating a combination with sign and name etc
-                    instr.setUseRawName();
+                    // However, you can translate it on the client side by passing translate_insructions=false
```
karussell: As this is *here* the client side this is a bit confusing -> maybe just "you can avoid translating it by passing translate_insructions=false in the hints"?

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
```diff
@@ -138,6 +140,23 @@ public void testGraphHopperWeb() throws Exception {
...
+    public void testNonTranslatedInstructions() {
+        GraphHopperAPI hopper = new GraphHopperWeb();
+        assertTrue(hopper.load(getTestRouteAPIUrl()));
+        GHResponse rsp = hopper.route(new GHRequest(42.554851, 1.536198, 42.510071, 1.548128));
+        assertEquals("Continue onto Carrer Antoni Fiter i Rossell", rsp.getBest().getInstructions().get(2).getName());
+
+        GHRequest request = new GHRequest(42.554851, 1.536198, 42.510071, 1.548128);
+        request.getHints().put("translate_instructions", false);
+        rsp = hopper.route(request);
+        assertEquals("Carrer Antoni Fiter i Rossell", rsp.getBest().getInstructions().get(2).getName());
```
karussell: Hmmh, what are other options to name the parameter? E.g. instructions_use_street_name? As translate_instructions is not that precise (?)

devemux86: How about 'server_instructions' ?

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
devemux86: New parameter is about to turn on/off the server instructions and simultaneously get the street name.
Still user should be allowed to generate full instructions in client like here (not fetch), so an exception seems too strict?

core/src/main/java/com/graphhopper/util/Helper.java
```diff
@@ -437,7 +437,7 @@ public static void cleanHack() {
 
     public static String nf(long no) {
         // I like french localization the most: 123654 will be 123 654 instead
-        // of comma vs. point confusion for english/german guys.
```
boldtrn: Yes, not a problem, will change that.

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
boldtrn: Mhm, not that sure, I think this is exactly the meaning. Maybe we should call it pretranslate_instructions? A client would usually pass this parameter when translating the instructions on the client side. So the client puts either the raw translated name into the instruction name or only the street name. But since we don't set the raw flag we have to translate it on the client side. We only use this parameter on the Java client. I think the server is not evaluating this parameter, since we always return both.

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
boldtrn: Mhm, I think this PR is about making GraphHopper and GraphHopperWeb instructions behave equally. It is not really about showing street names, even though this is the title of the PR :). When fetching non translated instructions we can return the street name. WDYT does this makes sense?

devemux86: >When fetching non translated instructions we can return the street name.

And be allowed to translate the instructions on client side (like mentioned above), is that what you mean?

web/src/main/java/com/graphhopper/http/GraphHopperWeb.java
boldtrn: translate_instructions=false is the default. So I thought it would be better to describe the "special case". Should I change it?

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
karussell: The problem with translate_instructions is that this could be misunderstood with the locale parameter.

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
karussell: Ok, exception is too strict but I wouldn't want this unit test here as this would then define a behaviour that we currently (in this PR) do not care.

web/src/main/java/com/graphhopper/http/GraphHopperWeb.java
karussell: Sorry, I think what I mean is that translate_instruction or 'translate it on the client side' is confusing. We should name it what it does, not what it is supposed to be useful for :)

So what it does IMO is that it returns the raw street name instead of a turn instruction in the name (so that one would need translation on the client side).

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
karussell: Also it is not really about translation it just skips the concatenation of the sign and the street. Maybe turn_sign_instruction=false or instructions_sign_in_name=false? Or similar to what @devemux86 says: instructions_use_server_name=false (and rename the java method too) or instructions_use_raw_name=false?

devemux86: Something with 'raw_name' in it seems suitable to me too.
That way it's connected with the `setUseRawName` which actually controls.

boldtrn: Yes, I like the raw name. So let's go with `instructions_use_raw_instructions`.

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
boldtrn: Ok, so I removed the translation map and only compare the name now. Hope this is fine :)

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
boldtrn: This is where it gets weird. If we use the raw name of the server side, this would mean the street name. getTurnDescription returns only the street name. On the client side using the raw name would mean the pre translated turn instruction that was returned from the server. So after thinking about this a bit more, how about `instructions_use_turn_description`?

boldtrn:
Ok, I updated the PR. Naming is important and I am still not 100% convinced by the current naming. Since this is only relevant for the Java client, it might be changeable more easily. WDYT? I think something like: `instructions_init_with_turn_description` might be a bit better. But then it might get too long?

karussell:
> Since this is only relevant for the Java client, it might be changeable more easily. WDYT?

Ah, yes. This is true. Can we make it a bit shorter though :) ... if not I'll merge :)

boldtrn:
> Can we make it a bit shorter though

Yes, I had the same thought, but I wasn't able to come up with anything. Do you have a proposal?

web/src/main/java/com/graphhopper/http/GraphHopperWeb.java
```diff
@@ -150,9 +152,12 @@ public static PathWrapper createPathWrapper(JSONObject path,
                         instr = new Instruction(sign, text, ia, instPL);
                     }
 
-                    // The translation is done from the routing service so just use the provided string
-                    // instead of creating a combination with sign and name etc
-                    instr.setUseRawName();
+                    // Usually, the translation is done from the routing service so just use the provided string
+                    // instead of creating a combination with sign and name etc.
+                    // This is called the turn description.
+                    // This can be changed by passing <code>instructions_use_turn_description=false</code>.
```
devemux86: This needs to be updated to parameter name.

devemux86:
Could be `instruction_turn_descriptions` or even `turn_descriptions`?

karussell:
> turn_descriptions

I like that. It is not 100% clear but nicely short and we can just refer to this PR for explaining it or change it later on.

web/src/main/java/com/graphhopper/http/GraphHopperWeb.java
boldtrn: Thanks, almost overlooked that one :)

boldtrn:
Ok, that was harder than I expected ;). Changed the parameter to `turn_description` as proposed. Should be good to merge?

karussell:
Thanks!

PR #950, opened 2017-02-11
Remove deprecated methods

### LINKED ISSUE #945: Remove deprecated methods like GraphHopper.setWeighting(s)

karussell:
Fixes #945 

This breaks the old API but was announced earlier via deprecated.

### DIFF
--- a/core/files/changelog.txt
+++ b/core/files/changelog.txt
@@ -1,5 +1,6 @@
 0.9
-    
+    removed deprecated methods in GraphHopper, Path.calcMillis, LocationIndex.findID
+
 0.8
     refactoring to Weighting class, see #807
     removed FlagEncoder from parameters as weighting.getFlagEncoder can and is used
@@ -10,6 +11,7 @@
     moved subnetwork code into own package com.graphhopper.routing.subnetwork
     moved weighting code into own package com.graphhopper.routing.weighting
     code format has changed, so it is important to change your PRs too before you merge master, see discussion #770
+
 0.7        
     added snapped points to output JSON for every path
     the foot routing is now much smoother and only considers safe paths, to use beautiful roads (i.e. prefer hiking routes etc) use the new 'hike' profiles, see #633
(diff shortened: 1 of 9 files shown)


core/files/changelog.txt
```diff
@@ -1,5 +1,6 @@
 0.9
-    
+    removed deprecated methods in GraphHopper, Path.calcMillis, LocationIndex.findID
```
boldtrn: Should we also name the removed methods in GraphHopper?

core/src/main/java/com/graphhopper/storage/index/LocationIndex.java
```diff
@@ -42,12 +42,6 @@
     LocationIndex prepareIndex();
 
     /**
-     * @return the closest node id for the specified geo location (latitude,longitude)
-     * @deprecated will be removed with 0.8 use 'QueryResult findClosest' instead
```
boldtrn: BTW: Why haven't we marked the method with @Deprecated. Maybe someone continued to use them without reading the comment?

karussell: Not sure. For my IDEs it does not matter if I mark them in comment or outside.

core/files/changelog.txt
karussell: ok

karussell:
It should me shown in IntelliJ too: 
deprecated

karussell:
Ah, no. That is only the interface and for any implementation it is not shown!? But this does not change with the explicit annotation. Strange. This smells like a bug to me :)

boldtrn:
> Ah, no. That is only the interface and for any implementation it is not shown!? But this does not change with the explicit annotation. Strange. This smells like a bug to me :)

Ah, I only tried it for one implementation. So in this case I will approve this PR. Makes no sense to block this PR only because of buggy intelliJ implementation ;).

karussell:
BTW: it is not a bug, it is a (configuration) feature ...
deprecated

PR #954, opened 2017-02-20
Fix encoded value issue #947

### LINKED ISSUE #947: encoded value: rounding leads to max value violation
[This code](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/EncodedDoubleValue.java#L56) in EncodedDoubleValue is ugly and we should probably avoid the first 'rounding'. Also it works differently than in EncodedValue.

This leads to an exception when with a precision of 4 the 110km/h is rounded upwards to 112 and violates the maximum allowed value:

```java
   public void testMaxSpeed2() {
       CarFlagEncoder limitedCar = new CarFlagEncoder(5, 4, 0) {
           {
               maxPossibleSpeed = 110;
           }
       };
       EncodingManager em = new EncodingManager(limitedCar);
       way.setTag("highway", "motorway");
       way.setTag("maxspeed", "none");
       long flags = limitedCar.handleWayTags(way, 1, 0);
       assertEquals(110, limitedCar.getSpeed(flags), 0.01);
   }
```

boldtrn:
This PR fixes #947.

In more detail the rounding for checkValue was removed and it is required that maxValue/factor is a natural number.


core/files/changelog.txt
```diff
@@ -1,6 +1,7 @@
 0.9
     
 0.8
+    EncodedDoubleValue requires maxValue/factor to be a natural number #954
```
karussell: The notice should be added to the 0.9 section

core/src/main/java/com/graphhopper/routing/util/EncodedDoubleValue.java
```diff
@@ -31,6 +31,11 @@ public EncodedDoubleValue(String name, int shift, int bits, double factor, long
 
     public EncodedDoubleValue(String name, int shift, int bits, double factor, long defaultValue, int maxValue, boolean allowZero) {
         super(name, shift, bits, factor, defaultValue, maxValue, allowZero);
+        double factorDivision = maxValue / factor;
+        // Interestingly we cannot use modulo here, e.g. 10%00.1 != 0
+        if (factorDivision != (int) factorDivision) {
+            throw new IllegalStateException("MaxValue needs to be divideable by factor without remainder");
```
karussell: divideable -> divisible?

core/src/main/java/com/graphhopper/routing/util/EncodedDoubleValue.java
```diff
@@ -31,6 +31,11 @@ public EncodedDoubleValue(String name, int shift, int bits, double factor, long
 
     public EncodedDoubleValue(String name, int shift, int bits, double factor, long defaultValue, int maxValue, boolean allowZero) {
         super(name, shift, bits, factor, defaultValue, maxValue, allowZero);
+        double factorDivision = maxValue / factor;
+        // Interestingly we cannot use modulo here, e.g. 10%00.1 != 0
```
karussell: This is not unexpected :) http://stackoverflow.com/questions/3227342/modulus-with-doubles-in-java

core/src/main/java/com/graphhopper/routing/util/EncodedDoubleValue.java
karussell: Maybe we can call checkValue somehow here to avoid separate code and include the other situations already in the constructor?

core/src/main/java/com/graphhopper/routing/util/EncodedDoubleValue.java
boldtrn: Thanks for the link. I am aware of the floating point issue, I just thought it would work for these obvious things.

core/src/main/java/com/graphhopper/routing/util/EncodedDoubleValue.java
boldtrn: > divideable -> divisible?

Yes, thank you.

> Maybe we can call checkValue somehow here to avoid separate code and include the other situations already in the constructor?

What do you mean by that?

karussell: I meant that instead of an explicit call we can call the checkValue method in the constructor somehow. So that the other situations are also checked (?)

boldtrn: Yes we can do that, but we still need the current check as well?

karussell: It is good as it is now, because otherwise we would check this for every value which makes no sense.

PR #955, opened 2017-02-20
Improve instructions

boldtrn:
This PR intents to improve the current turn instructions. (#94)

Manly this PR tries to solve the following issues:

- Reduce unnecessary instructions (continue and turns that are actually no turn, but just a street curve).
- Missing instructions if there is no name change

I made a couple of assumptions and changes the way I would like to receive turn instructions. This is very subjective, some people like more, some people like less.

The basic function is:

- Force an instruction if annotations are changing (do we want that?)
- Ignore Continues, unless we are leaving the current street
- If there is no other turn possible, ignore the turn instruction
- If there are other turns, and the turn instruction is something else than a slight turn, show the instruction
- If the instruction is a slight turn, only show it if the other turns contain a slight turn or continue. Otherwise it will be an obvious follow the street situation (this is probably an assumption that others have a different opinion on?)
 - In this case we further check if the ref/name changes or the flags changes

Currently, a lot of tests are failing. I can fix all the failing ITs once, we decided if we want the current behavior like that or not.

**My personal opinion on the state of this PR:**

It is better as the current approach in most situations. However, there are thousands of edge cases we need to consider¹. For example look at the TODOs, there are cases that just happens very infrequent, but they happen and with some I feel like there is no good solution, without making other cases worse. It mostly happens in dense cities with a lot of streets around. Bicycle and Non-City instructions are pretty good IMHO (haven't found bigger issues yet).

**Feedback from others checking the turn instructions for their area would be highly appreciated.**

TODO:

- [x] Shows an unnecessary continue http://localhost:8989/?point=48.004367%2C8.137822&point=48.003878%2C8.137844
- [x] Shows an unnecessary continue on trunks, motorways. Depending on the motorway speed this happens quite frequently.
  - The reason is quite interesting. The trunk is marked with maxspeed = 70, the trunk_link has no maxspeed, but the default is 65. We multiply the maxspeed with .9, but not the default speed. So the trunk_link has the same speed as the trunk -> therefore, we are not sure -> show instruction
  - http://localhost:8989/?point=48.177999%2C8.572013&point=48.17788%2C8.573037
  - http://localhost:8989/?point=47.913318%2C8.097208&point=47.91302%2C8.09805 
  - Possible fix, could be to either slow down _links or mark links in encoder, which would be uneccessary, as we want to switch to the DataFlagEncoder
  - Fixed it by detaching the behavior for continue and slight turns again. If we are not leaving the street, don't show a continue.
- [x] When leaving a street via a CONTINUE instruction. We should check for continue instructions if it seems that we are leaving a street, if yes, show the CONTINUE instruction, or maybe create a special instruction?
  - A good example can be seen here: https://graphhopper.com/maps/?point=49.037889%2C13.118073&point=49.034949%2C13.119403&vehicle=bike
  - Possible fix would be an exit type instruction. This is used for stop following a certain street. This can be a motorway, e.g. switchting to a *_link street. This can be also mean leaving a bigger street via a small street.

- [ ] When turning onto a one-way turn instructions are omitted, as it is the only possible turn. In general this is ok, but when there are two one-ways (e.G. highways) and you need to cross the first one and then turn left on the second, the turn instruction is omitted
  - http://localhost:8989/?point=52.508392%2C13.326287&point=52.509456%2C13.325692
  - Possible fix: when turning into a one-way, show the turn instruction.
  - When thinking about this a bit more, it is kind of correct. Drive straight, until there is only one possible way to turn, then take this way. However, not really nice for the user.

- [x] In this situation, we could show a left turn, not sure: http://localhost:8989/?point=48.769739%2C9.245929&point=48.769707%2C9.245451
  - Same situation, just a couple of hundred meters away - http://localhost:8989/?point=48.770192%2C9.242002&point=48.770174%2C9.241734
  - The reason it happens, is that the street name continues for a very short time, and then switches the street name http://www.openstreetmap.org/way/10578677#map=19/48.76976/9.24587&layers=ND. So it looks like we would follow the street, but actually we are leaving the street.
  - I checked how it looks on Google Maps, no street view available, but the street geometry looks different https://goo.gl/maps/KUfD8fmENwr
   - I checked the satalite pictures, both mappings seem to be ok. Awful situation. Not sure how to fix it without actually seeing the situation.
  - Possible fix could be to check if there is more than one possible turn with the same name, to show the turn instruction?
- [x] Just found the worst case situation for the junction analyzer. Two edges, one is the edge we are turning onto, the other is just a random edge, but both starting and ending at the same node. Therefore, the other edge is not counted as possible turn. See http://localhost:8989/?point=48.815822%2C9.151365&point=48.815409%2C9.152888. The issue is created by this edge http://www.openstreetmap.org/way/7975515.
- [x] Revert the tmp js change. This makes debugging instructions a better experience, but would probably annoy regular users.

¹: Thousands of edge cases sounds like a good case for deep learning.


boldtrn:
So the current state is not too bad, IMHO. I tested a lot of random bike routes and compared the generated instructions with the instructions I would have expected according to the map. I'd say that I agree with > 95%. Are there any examples that I should try? I have tested the examples provided #94, but most of them still behave the same, as they were not intended to be improved in this PR. We should definitely find a solution to the above mentioned **NEW** issue. Not sure if performance is an issue here? Haven't measured my solution, but it obviously needs more resources, than the old solution. It does not feel slower than before (...).

boldtrn:
I updated this PR to show Continue instructions if the situation might not be clear. Sometimes they happen too frequent, but haven't found a better way to filter them.

karussell:
I really like the debugging view where all turns have a marker - nice :) !

Also I think the improvement is indeed there and good and will now look into more examples.

karussell:
I tried a bit within Bavaria, Germany.

## necessary improvement for this PR

 * too few instructions on motorway, no indication that motorway starts and changes like it is the case for the current version http://localhost:8989/?point=50.012758%2C11.325016&point=49.330912%2C11.065464&locale=de&vehicle=car
 * important to guide for this motorway junction http://localhost:8989/?point=49.465419%2C11.213307&point=49.454931%2C11.252661&locale=de&vehicle=car
 * missing 'right' instruction for this junction http://localhost:8989/?point=49.404495%2C11.19524&point=49.393589%2C11.187987&locale=de&vehicle=car

## possible improvements

 * here should be a left turn instruction but bike has this instruction!? http://localhost:8989/?point=49.958285%2C11.491259&point=49.959531%2C11.490026&locale=de&vehicle=foot missing instruction for opposite direction too: http://localhost:8989/?point=49.959489%2C11.490283&point=49.958323%2C11.49098&locale=de&vehicle=foot
 * at the beginning a right instruction is missing, at the end a right instruction is too much, in the beginning a bug is shown: foot is allowed in any direction around the roundabout and we have to tell this in the turn instruction explicitly if in opposite direction http://localhost:8989/?point=49.91849%2C11.561201&point=49.911965%2C11.56899&locale=de&vehicle=foot

## later

 * the right turn is ambiguous http://localhost:8989/?point=49.94869%2C11.514546&point=49.949504%2C11.513452&locale=de&vehicle=foot
 * the left turn is ambiguous http://localhost:8989/?point=49.992208%2C11.45638&point=49.99236%2C11.460114&locale=de&vehicle=foot
 * if turns are <10m away should we consider merging them or how to properly guide a person? http://localhost:8989/?point=49.942248%2C11.545016&point=49.942438%2C11.543326&locale=de&vehicle=foot
 * http://localhost:8989/?point=49.936265%2C11.530414&point=49.9366%2C11.529974&locale=de&vehicle=foot
 * start direction should contain north/west/south/east, finish should contain indication of the side of the road

boldtrn:
@karussell Thanks for the review.

I updated this PR. In order to fix your necessary improvements I had to introduce a new concept: "keep right|left". In theory this should recognize forks, and tell the user if (s)he should go left or right at the fork. This fixes most issues, especially for motorways. Until all clients support the new sign for keeps we could also change the keeps to slight_right|left.

However, this fails sometimes. To make it better I think that I would need the highway type. For example we still miss the right turn here: http://localhost:8989/?point=49.398826%2C11.197751&point=49.39789%2C11.197729&vehicle=car

> at the end a right instruction is too much and in the beginning a bug is shown

This example is very interesting: http://localhost:8989/?point=49.912877%2C11.568802&point=49.911965%2C11.56899&vehicle=foot

Most importantly, why does it show turn right, instead of continue? I debugged a bit. The orientation is calculated for the following points with the following orientation:
```
Orientation:-1.4124066420807577
49.913634357906155,11.568623480989563
49.91265814556162,11.568865624864376

Orientation:0.0
49.91265814556162,11.568865624864376
49.91265814556162,11.568865811128894

Orientation:-1.41231436135386
49.91265814556162,11.568865811128894
49.91196980109481,11.569036649239916
```

The second orientation is calculated for two points that are almost identical. Not sure if GH creates a very small edge in this case? Due to this GH thinks that there is a right turn, with two possible turns, so the right turn is displayed.

> here should be a left turn instruction but bike has this instruction!?

Probably the bike instruction was due to an annotation? Should be fixed with my new commit anyway.

> foot is allowed in any direction around the roundabout and we have to tell this in the turn instruction explicitly if in opposite direction

Haven't touched roundabouts in this PR. Maybe we should create a new issue for this?

> the right turn is ambiguous http://localhost:8989/?point=49.94869%2C11.514546&point=49.949504%2C11.513452&locale=de&vehicle=foot

That's true, but I am not sure if this would be fixable. You could also walk straight ahead, so you are shown the instruction to turn right onto this foot path. From a data perspective this turn instruction is correct. This might be misleading for a user though.

> the left turn is ambiguous

How would you fix that? I think this is the typical example where people turn wrong. GPS is not exact enough, so you have to study the map closely.

> if turns are <10m away should we consider merging them or how to properly guide a walking person?

Shouldn't this be the job of the navigation engine? I think we show the correct turns here. A navigation engine might check if instructions are close together and merge them like: "turn left and after 10 m turn right".

> start direction should contain north/west/south/east, finish should contain an indication which side of the road the destination is

That's a good idea. Maybe put this into a different PR as well?

web/src/main/webapp/js/instructions.js
```diff
@@ -46,7 +46,13 @@ function addInstruction(mapLayer, main, instr, instrIndex, lngLat, useMiles) {
                     setLatLng([lngLat[1], lngLat[0]]).
                     setContent(title).
                     openOn(mapLayer.getMap());
+
         });
+
+        // Debug Turn Instructions more easily
+        L.marker([lngLat[1], lngLat[0]], {
+            draggable: true
+        }).addTo(mapLayer.getRoutingLayer()).bindPopup(title);
```
karussell: Can we always enable turn instruction debugging via an URL parameter `debug_instruction=true` or something? Also printing a marker for every instruction is really helpful (but not for the normal public audience of course ;))

boldtrn:
I've just realized that the improved instructions fail for the DataFlagEncoder. The improved instructions use the getSpeed method, which is not implemented for the DataFlagEncoder. What should we do about that?

boldtrn:
I updated this PR to show the instructions nodes only when passing `debugInstructions=true`. Also also used a smaller icon to make it more usable.

web/src/main/webapp/js/instructions.js
boldtrn: Good idea, added that feature.

karussell:
> What should we do about that?

Probably it is time to refactor something else before this change. Because getSpeed is not something the FlagEncoder can know and only a configured Weighting does. Hmmh.

boldtrn:
> Probably it is time to refactor something else before this change. Because getSpeed is not something the FlagEncoder can know and only a configured Weighting does. Hmmh.

I think we need a different sort of logic for the DataFlagEncoder anyways, as it contains the HighwayType and so on. So I could try and see if I can create some special code block for the DataFlagEncoder?

Not very nice but probably the best choice right now?

boldtrn:
I updated this PR. I added a fix for the DataFlagEncoder. We might have slightly more instructions for the DataFlagEncoder compared to the other Encoders. We might show instructions where they are not necessary, but we don't ignore any instructions. The speed is currently only used to omit instructions that seem to be not relevant.

In a future PR, we should improve the instructions for the DataFlagEncoder using also the highway tag and maybe even calculate the actual speed for every edge. I will open a separate issue for this.

The good thing is we don't have to defer this PR and can use the improved instructions.

core/src/main/java/com/graphhopper/routing/Path.java
```diff
@@ -703,7 +704,16 @@ private boolean surroundingStreetsAreSlowerByFactor(int baseNode, int prevNode,
                 double maxSurroundingSpeed = -1;
                 double tmpSpeed;
                 while (edgeIter.next()) {
-                    tmpSpeed = encoder.getSpeed(edgeIter.getFlags());
+                    if (encoder instanceof DataFlagEncoder) {
+                        tmpSpeed = ((DataFlagEncoder) encoder).getMaxspeed(edgeIter, 0, false);
```
karussell: as we have the weighting in the Path class available, would it be somehow possible to use it here instead? E.g. edgeIter.getDistance/calcMillis?

boldtrn: Might be, yes. I am a bit afraid of rounding issues, as I am comparing these values, but sure, I can try that. OTOH I think we need a different logic for the DataFlagEncoder. We need to consider the highway tag and speed becomes rather unimportant (maybe we only need to check speed with a low priority, whereas for the other encoders speed is important for the instruction generation.

boldtrn:
Currently problematic remains this Route:

http://localhost:8989/?point=49.398826%2C11.197751&point=49.39789%2C11.197729&vehicle=car

But the current version of GH has the same issue, so this is no regression:

https://graphhopper.com/maps/?point=49.398826%2C11.197751&point=49.39789%2C11.197729&vehicle=car

I think we should be aware of these things and try to improve them, but from a data point of view this is hard to solve. With highway tags this would be easy to solve. However, this need a different logic, so I think we should do this separately for the DataFlagEncoder. After some consideration I think we should do the DataFlagEncoder Optimization in a completely different branch. The DataFlagEncoder instructions are improved by this PR, but not as much as the other encoders.

Things I've seen during testing:
- Sometimes we get a keep left, instead of a slight left. Hard to decide since I don't know the actual conditions. http://localhost:8989/?point=49.216686%2C10.884511&point=49.214941%2C10.883374&vehicle=car. Again the turn looks very different on Google maps https://goo.gl/maps/raW74WuqSFF2. Using the satellite view a keep left might actually be not to bad, is there is a left-turn lane.

Something I have changed is to return Continues instead of Keeps, if we are going pretty straight and the other road is at least slightly bend. This removes keeps, that could be misunderstood, if there is a turn close by.

When we merge this, we should also get images for the keeps in the front-end.

karussell:
To avoid breaking the existing clients can we also use keep_left=slight_left just until we have released the more tolerant js&java clients for ~3 months?

boldtrn:
> To avoid breaking the existing clients can we also use keep_left=slight_left just until we have released the more tolerant js&java clients for ~3 months?

Yes sure, not a problem.

boldtrn:
I just changed the code a bit. If there was an annotation, we would not have used the improved instructions, but just showed the old instructions.

Here is one example: http://localhost:8989/?point=49.037889%2C13.118073&point=49.034949%2C13.119403&vehicle=bike

With the old code, this would have generated the following instruction "Continue, small way, unpaved", now we generate this: "Turn slight right, small way, unpaved", which is more accurate.

devemux86:
:+1: 
@boldtrn nice work!

Issue #960, opened 2017-02-23
Too many weights were maxed out. Use a bigger factor


boldtrn:
I was just playing around with the new landmark algorithm. While doing so, I received an interesting log: 
```
landmark 0 -> too many weights were maxed out (1065332/1065333). Use a bigger factor 1.8212685272742498E-4
```

This log was shown twice for all 16 landmarks. The distance heuristic of this weighting is not very good, so my minWeight is rather small. I don't want to change the minWeight, as this might break the a_star. AFAIK this is the only way to change the factor?

The routing for this particular weighting is not faster with landmarks, where it is a lot faster for other weightings I tested.

Is there a way to fix this?

karussell:
Yes, this factor is rather bad in your case. 

> The routing for this particular weighting is not faster with landmarks, where it is a lot faster for other weightings I tested

The factor is indeed approximated via the minWeight value. Instead we more need a maxWeight value for this factor and so I did minWeight*X which seems to fail here. Configuring the factor is a good idea in such cases or some new maxWeight method ...

PR #961, opened 2017-02-23
better landmark docs

karussell:
Please review and also waiting for #960 being merged ...

### DIFF
--- a/docs/core/landmarks.md
+++ b/docs/core/landmarks.md
@@ -10,3 +10,15 @@ for (bidirectional) A* and are done. It is important that although a weight
 **approximation** is used the landmarks algorithm is still not heuristic per
 default. You can make it faster and more heuristical behaving when you set
 the epsilon parameter lower than 1.
+
+For the world wide case we limit the routing in a certain way for landmarks.
+You can only route within the EU, Africa and Asia. The other algorithms are
+not affected from this limitation though. For landmarks this is important
+as we store the weight approximation in a short (two bytes) and for large distances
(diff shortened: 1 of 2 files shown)


docs/core/landmarks.md
```diff
@@ -10,3 +10,15 @@ for (bidirectional) A* and are done. It is important that although a weight
 **approximation** is used the landmarks algorithm is still not heuristic per
 default. You can make it faster and more heuristical behaving when you set
 the epsilon parameter lower than 1.
+
+For the world wide case we limit the routing in a certain way for landmarks.
+You can only route within the EU, Africa and Asia. The other algorithms are
```
boldtrn: For routing requests using the landmark algorithm it is not possible to cross the borders between EU, Africa, and Asia?

Might be a bit more understandable? Or generally stating that you cannot route accross these borders using the landmark algorithm. The above sounds like routing in USA wouldn't be possible at all.

docs/core/landmarks.md
```diff
@@ -10,3 +10,15 @@ for (bidirectional) A* and are done. It is important that although a weight
...
+not affected from this limitation though. For landmarks this is important
```
boldtrn: affected **by**? Not sure though

docs/core/landmarks.md
karussell: Ah, ups. You are right.

karussell:
Will correct the problems and also add the new prepare.lm.threads config possibility.

PR #962, opened 2017-02-23
Renaming si.txt to sl.txt

Klofutar:
Language code for Slovenian is sl.


karussell:
Thanks for pointing to this issue. Will change it to `sl-SI` (with fallback to `sl`), but this requires some changes elsewhere (which I'll do to)

PR #963, opened 2017-02-23
make factor configurable to avoid 'maxed out' problems for e.g. PriorityWeighting

### LINKED ISSUE #960: Too many weights were maxed out. Use a bigger factor

karussell:
Fixes #960


karussell:
> Is there a reason we cannot define the maximum weight for each weighting?

You mean a new method of the Weighting interface? I wanted to avoid this for now :)

> If we prepare good and "bad" weightings, it might be nice to only change the factor for the bad weighting?

What do you mean here?

> I briefly tried your PR. I changed the FastestWeighting to the following (which is not a very clever thing to do ;))

Oh, no ... I had a different intent :)

The idea was to keep the minWeight but specify an absolute value in the config.properties:

prepare.lm.maximum_weight=1.222

> But this is the factor * 2. This should be factor * 2 * PRECISION?

Yes, will fix!

> Maybe we can calculate the average maxedout value and show that here (or max maxedout?)?

We can determine the maximum weight ourselfs and add 50% (maybe in a different PR and if you have interests :) ?). Still this would be just a heuristic approx...

karussell:
Ah, now I understand what you mean here:

> If we prepare good and "bad" weightings, it might be nice to only change the factor for the bad weighting?

This is indeed important. Will think how to handle this.

karussell:
> If we prepare good and "bad" weightings, it might be nice to only change the factor for the bad weighting?

This should be possible via `prepare.lm.weightings=fastest|maximum=123423,shortest|maximum=444`

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -765,20 +782,26 @@ public boolean apply(int nodeId, SPTEntry value) {
...
+                    }
                 }
             });
 
-            if ((double) maxedout.get() / map.size() > 0.1)
-                LOGGER.warn("landmark " + lmIdx + " -> too many weights were maxed out (" + maxedout.get() + "/" + map.size() + "). Use a bigger factor " + lms.factor);
+            if ((double) maxedout.get() / map.size() > 0.1) {
+                LOGGER.warn("landmark " + lmIdx + " (" + nodeAccess.getLatitude(lmNodeId) + "," + nodeAccess.getLongitude(lmNodeId) + "): " +
+                        "too many weights were maxed out (" + maxedout.get() + "/" + map.size() + "). Use a bigger factor than " + lms.factor
+                        + ". For example use the following in the config.properties: weighting=" + weighting.getName() + "|maximum=" + finalMaxWeight.getValue() * 1.2 / PRECISION);
```
boldtrn: Mhm, somehow the proposal does not match here :/. Why do we divide by PRECISION?

When I set `prepare.lm.weightings=fastest|maximum=100` then I get the following log `landmark 15 (45.5060110354277,-73.5529020105206): too many weights were maxed out (294/440). Use a bigger factor than 0.00152587890625. For example use the following in the config.properties: weighting=fastest|maximum=0.004969158898255066`.

karussell: > Why do we divide by PRECISION?

You are right ... I was still thinking in terms of `factor=weight/PRECISION`

karussell: Then the suggestion makes more sense maximum=325 (`2^16*0.004969158898255066`)

boldtrn: Yes, that makes sense. Thank you :). I just tried it with 250, the warning was only printed for one landmark now. When switching to 325 there is no warning anymore.

PR #966, opened 2017-02-24
Updated SpatialRuleLookup initialization

### LINKED ISSUE #965: Cannot correctly build SpatialRuleIndex
I just found an ugly bug with the way we [initialize the SpatialRuleLookupBuilder in the DefaultModule](https://github.com/graphhopper/graphhopper/blob/e085c0b3f3b1e8a3b5942d0938218bec40a357f3/web/src/main/java/com/graphhopper/http/DefaultModule.java#L109). The problem is that at this point the graph is not loaded, therefore we cannot get it. Getting the bounds in advance is probably not possible.

For a quick fix, we could simply define the bbox in the config.

The usage itself is a bit inconvenient as well.
You have to define the location of the json file, the number of spatial rules the encoder contains, (and the BBox).

Maybe it would make more sense, to define the names of the SpatialRules by hand? This would remove the need to define the BBox and the number of SpatialRules.

Currently, the usage SpatialRuleIndex is not documented. When fixing this issue and testing if it works, we should document the process, as well.

WDYT?

boldtrn:
Fixes #965 

Not entirely happy with the fix. Defining the BBox beforehand is unnecessary tedious. The spatial lookup works now with the provided configuration in the config-examples.

### DIFF
--- a/core/src/main/java/com/graphhopper/util/shapes/BBox.java
+++ b/core/src/main/java/com/graphhopper/util/shapes/BBox.java
@@ -308,4 +308,23 @@ public class BBox implements Shape, Cloneable {
 
         return new BBox(minLon, maxLon, minLat, maxLat);
     }
+
+    /**
+     * This method creates a BBox out of a string in format lon1,lon2,lat1,lat2
+     */
+    public static BBox parseBBoxString(String objectAsString) {
+        String[] splittedObject = objectAsString.split(",");
+
+        if (splittedObject.length != 4)
+            throw new IllegalArgumentException("BBox should have 4 parts but was " + objectAsString);
+
+        double minLon = Double.parseDouble(splittedObject[0]);
+        double maxLon = Double.parseDouble(splittedObject[1]);
+
+        double minLat = Double.parseDouble(splittedObject[2]);
+        double maxLat = Double.parseDouble(splittedObject[3]);
+
+        return new BBox(minLon, maxLon, minLat, maxLat);
+    }
+
 }
--- a/web/src/main/java/com/graphhopper/http/DefaultModule.java
+++ b/web/src/main/java/com/graphhopper/http/DefaultModule.java
@@ -27,11 +27,15 @@ import com.graphhopper.reader.osm.GraphHopperOSM;
 import com.graphhopper.routing.lm.LandmarkStorage;
 import com.graphhopper.routing.lm.PrepareLandmarks;
 import com.graphhopper.routing.util.DataFlagEncoder;
+import com.graphhopper.routing.util.FlagEncoder;
+import com.graphhopper.routing.util.FlagEncoderFactory;
+import com.graphhopper.routing.util.spatialrules.SpatialRule;
 import com.graphhopper.routing.util.spatialrules.SpatialRuleLookup;
 import com.graphhopper.routing.util.spatialrules.SpatialRuleLookupBuilder;
 import com.graphhopper.routing.util.spatialrules.countries.AustriaSpatialRule;
 import com.graphhopper.routing.util.spatialrules.countries.GermanySpatialRule;
 import com.graphhopper.util.CmdArgs;
+import com.graphhopper.util.PMap;
 import com.graphhopper.util.Parameters;
 import com.graphhopper.util.TranslationMap;
 import com.graphhopper.util.shapes.BBox;
@@ -43,6 +47,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.util.Arrays;
+import java.util.HashMap;
 
 /**
  * @author Peter Karich
@@ -71,7 +76,7 @@ public class DefaultModule extends AbstractModule {
     static SpatialRuleLookup buildIndex(Reader reader, String rules, BBox bbox) {
         GHJson ghJson = new GHJsonBuilder().create();
         JsonFeatureCollection jsonFeatureCollection = ghJson.fromJson(reader, JsonFeatureCollection.class);
-        return new SpatialRuleLookupBuilder().build(rules, jsonFeatureCollection, bbox, 1, true);
+        return new SpatialRuleLookupBuilder().build(new SpatialRuleListReflectionFactory(rules), jsonFeatureCollection, bbox, 1, true);
     }
 
     /**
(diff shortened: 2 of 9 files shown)


boldtrn:
I will update this PR to either use all SpatialRules. Or to use only specified SpatialRules. The BBox will be calculated from the rules.

core/src/main/java/com/graphhopper/util/shapes/BBox.java
```diff
@@ -308,4 +308,23 @@ public static BBox parseTwoPoints(String objectAsString) {
 
         return new BBox(minLon, maxLon, minLat, maxLat);
     }
+
```
boldtrn: Should we remove this method, since it is not used anymore?

boldtrn:
I updated this PR to use reflections to load generic SpatialRules. This is easier than fiddling out the BBox of the rules and provides a lot of additional flexibility. 

One expected advantage of this approach was to get rid of the tedious definition of graph.flag_encoders=generic|**spatial_rules=X**. I wasn't able to remove this easily. From a user perspective this does not make a lot of sense, as they already provided the list of Spatial Rules, why should one calculate the number of Spatial Rules. The problem is that the DataFlagEncoder is created before the SpatialRuleLookup, including the EncodedValues, which rely on the number of Spatial Rules. One solution could be inject the number of Spatial Rules in the GraphHopper.init method, but I don't like that "solution". Any ideas to fix it? 

BTW: I changed the way of specifying the number of spatial rules. Up to now you needed to specify the number, including the empty spatial rule. I changed it to use the number of spatial rules without the empty rule. Example: If you specified the SpatialRules for Germany and Austria, you had to specify 3 Spatial Rules for the DataFlagEncoder as it also saves the Empty spatial rule. I changed to to 2.

karussell:
In core we cannot use reflection e.g. due to iOS. We have to move the reflection part into the buildIndex method somehow. Also it should be possible to use the full class name including the package name otherwise user have to pollute com.graphopper with their classes. And if no package name is specified we use com.graphhopper....

> The problem is that the DataFlagEncoder is created before the SpatialRuleLookup, including the EncodedValues, which rely on the number of Spatial Rules. 

Indeed, this tricky. The only possibility I see is via setting a custom FlagEncoderFactory e.g. in the DefaultModule class:

```java
final FlagEncoderFactory oldFEF = hopper.getFlagEncoderFactory();
hopper.setFlagEncoderFactory(new FlagEncoderFactory() {
    @Override
    public FlagEncoder createFlagEncoder(String name, PMap configuration) {
        if (name.equals(GENERIC)) {           
            srl = createSpatialRuleLookup();
            configuration.put("spatial_rules", srl.size() - 1);
            return new DataFlagEncoder(configuration).setSpatialRuleLookup(srl);
        }

        return oldFEF.createFlagEncoder(name, configuration);
    }
});
```

> graph.flag_encoders=generic|spatial_rules=X

This still appears in the config but we do not need it anymore, right?

boldtrn:
> In core we cannot use reflection e.g. due to iOS

Ah, wasn't aware of that. I created a Factory in the Web module, that handles the reflection part. The code in the DefaultModule got a bit weird, but should be still okayish?

> Also it should be possible to use the full class name including the package name otherwise user have to pollute com.graphopper with their classes.

That's already possible. You can either define the full package. If you only provide a class name it will fall back to `com.graphhopper.routing.util.spatialrules.countries`. Mostly for convenience :). It made reading the tests rather complicated with the package names cluttering everything.

> Indeed, this tricky. The only possibility I see is via setting a custom FlagEncoderFactory e.g. in the DefaultModule class

I followed you example here, thanks.

> This still appears in the config but we do not need it anymore, right?

With the addition of the custom FlageEncoderFactory this is not needed anymore. Before it was still needed. I removed it now.

core/src/main/java/com/graphhopper/util/shapes/BBox.java
karussell: Do you mean removing parseTwoPoints? Isn't this used for blocking area?

core/src/main/java/com/graphhopper/util/shapes/BBox.java
```diff
@@ -308,4 +308,23 @@ public static BBox parseTwoPoints(String objectAsString) {
...
+            throw new IllegalArgumentException("BBox should have 4 parts but was " + objectAsString);
+
+        double minLon = Double.parseDouble(splittedObject[0]);
+        double maxLon = Double.parseDouble(splittedObject[1]);
+
+        double minLat = Double.parseDouble(splittedObject[2]);
+        double maxLat = Double.parseDouble(splittedObject[3]);
+
+        return new BBox(minLon, maxLon, minLat, maxLat);
+    }
```
karussell: Why do we need this?

web/src/main/java/com/graphhopper/http/DefaultModule.java
```diff
@@ -144,4 +169,38 @@ protected void configure() {
...
+                    // Makes it easy to define a rule as CountrySpatialRule and skip the fqn
+                    if (!rule.contains(".")) {
+                        rule = "com.graphhopper.routing.util.spatialrules.countries." + rule;
+                    }
+                    Object o = Class.forName(rule).newInstance();
+                    if (SpatialRule.class.isAssignableFrom(o.getClass())) {
+                        ruleMap.put(((SpatialRule) o).getId(), (SpatialRule) o);
+                    } else {
+                        String ex = "Cannot find SpatialRule for rule " + rule + " but found " + o.getClass();
+                        logger.error(ex);
```
karussell: why logging and throwing? It will be logged already later on IMO

web/src/main/java/com/graphhopper/http/DefaultModule.java
```diff
@@ -144,4 +169,38 @@ protected void configure() {
             throw new IllegalStateException("Couldn't load graph", ex);
         }
     }
+
+    public static class SpatialRuleListReflectionFactory extends SpatialRuleLookupBuilder.SpatialRuleListFactory {
```
karussell: can we move this out of DefaultModule including the buildIndex method, which I only placed here as it was rather short. Now we should better separate this (?)

michaz:
I followed up a bit on this in branch 'spatial' -- have a look? It tries two things -- not using the super-wide SpatialIndex interface for the super-low-level Landmarks thing, and trying what happens if we use a QuadTree from JTS.

Is there some kind of integration test for DataFlagEncoder, with or without SpatialRules? When I change the examples-config to 'generic', I just get exceptions everywhere in GraphHopperServletIT.

boldtrn:
> Is there some kind of integration test for DataFlagEncoder, with or without SpatialRules?

No, there is no integration test for this, but a unit test, should be the last method in the DataFlagEncoderTest. I usually use the following example (no IT, sorry): https://graphhopper.com/maps/?point=48.685606%2C9.433265&point=48.694919%2C9.455945, without the SpatialIndex the DataFlagEncoder would take the road through the forest, with the SpatialIndex, it takes the detour around the forest. I think we should write an IT for this :).

> not using the super-wide SpatialIndex interface for the super-low-level Landmarks thing

The SpatialIndex was not created for Landmarks. The idea was to have country wide rules for routing.

michaz:
@boldtrn @karussell See 'spatial'. (Commit log). Did this in German because I expect it to be squashed anyway, it's just for you two to read. It's many, many small things.

michaz:
I did a few things in circles, which actually came out again the way you did them. ,-) But it's somewhat more concise in the end, and when (if) we replace your Quadtree by a standard one, the JSON by standard one, and the Geometry by the standard one, it all dissolves. (yay!)

boldtrn:
Let's close this in favor of #987

Issue #236, opened 2014-07-13
Make ford check including checking depth == 0


karussell:
As the [wiki says](http://wiki.openstreetmap.org/wiki/Ford) a ford can have `depth=0` so we should include this in our check for all vehicles. Additionally we could use InstructionAnnotation to warn the user about this possible problem. What about `intermittent=yes` or `seasonal=yes`?

Which will fix e.g. this possible bus route:
![bus-route](https://cloud.githubusercontent.com/assets/129644/3564662/fbaeaa7e-0a91-11e4-9c9c-2fbafb81c5f9.png)

http://graphhopper.com/maps/?point=Kulgera%2C%20Australia&point=Alice%20Springs%2C%20Australia

karussell:
Probably we tag this route segment as potential dangerous (via instruction annotation). Also I got a report [in france](https://graphhopper.com/maps/?point=43.473843%2C6.882076&point=43.473999%2C6.882784&vehicle=foot&elevation=true&layer=TF%20Cycle) that this should be passable.

malenki:
I'd vote for make fords routable with graphhopper. If the weren't passable the one or other way they wouldn't have been mapped as ford. You may warn users for fords (not the way between them), but there are less developed countries where routing with graphhopper will get you nowhere because of fords:
https://graphhopper.com/maps/?point=benje-novosele&point=40.299871%2C20.467808

MarkCupitt:
Suggestion: Maybe in the Turn Instructions, list specifically that it is a Ford and may not be passable ..

scaidermern:
Maybe we should try to give fords just a penalty instead of avoiding them altogether.

pyrog:
@malenki :+1: 
> I'd vote for make fords routable with graphhopper. If the weren't passable the one or other way they wouldn't have been mapped as ford. You may warn users for fords 

@scaidermern :+1: 
> Maybe we should try to give fords just a penalty instead of avoiding them altogether.

See an [example in France](https://graphhopper.com/maps/?point=47.055863,5.993144&point=47.07094,6.014199&point=47.071145,6.014585&point=47.068909,6.041054&vehicle=foot&locale=fr&elevation=true&layer=TF%20Outdoors):  I use intermediates to force the algorithm to use this itinerary.

We could have the same problem with [mountain pass](http://wiki.openstreetmap.org/wiki/Key:mountain_pass) in winter :wink:
:bulb: Add an option to avoid mountain pass, ford, toll…

[...]

PR #973, opened 2017-02-28
Added ford annotation for DataFlagEncoder

boldtrn:
As discussed in #236 fords are not routable with the car profile.

The DataFlagEncoder already allows to route through fords. This PR adds annotations to every ford, notifying users that the route contains a ford. The notification is only seen when looking at the instructions, no actual warning is produced.

Right now every Ford is routable. The question is if we actually want that. In #236 there is a discussion about a couple of attributes, including `depth, seasonal, intermittent`, maybe even more :). These checks are not done by the DataFlagEncoder or the GenericWeighting. Do we want to add that?

TODOS:

- [x] Translate the key `way_contains_ford` or decide on a different key.


core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
```diff
@@ -378,4 +385,28 @@ public void testSpatialId() {
         assertEquals(5, encoder.getMaxspeed(e3, -1, false), .1);
         assertEquals(-1, encoder.getMaxspeed(e4, -1, false), .1);
     }
+
+    private Translation getTranslation(final String msg) {
```
karussell: Will TranslationMapTest.SINGLETON work here?

boldtrn: Yes, thanks.

boldtrn:
ford_gh

I updated the PR with instructions and updated the Test accordingly.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -413,7 +412,7 @@ public long handleWayTags(ReaderWay way, long allowed, long relationFlags) {
             flags = accessEncoder.setValue(flags, getAccessValue(way));
 
 
-            if(spatialRuleLookupEnabled()){
+            if (spatialRuleLookupEnabled()) {
```
karussell: I've seen this in some other PR but forgot to comment: if we make this method public we have to follow java coding rules: `isSpatialRuleLookupEnabled`

karussell:
BTW: with DataFlagEncoder.getTransportModeMap we can even enable/disable fords per request

PR #977, opened 2017-02-28
Fix CH preparation bug

### LINKED ISSUE #976: In very rare cases a shortcut is not added
The problematic code is in PrepareContractionHierarchies.addShortcuts and there in `if() continue NEXT_SC`.

A unit test and a fix will follow soon. Cost me a bunch of hours to figure this out but was definitely worth it.

The bug is rare. I had to use a special combination of encoders, areas and subnetwork config to reproduce it. The problematic network is [here](https://graphhopper.com/maps/?point=52.71237%2C-1.77382&point=52.711746%2C-1.772763) where probably the [one southern road](http://www.openstreetmap.org/way/323870729) has a missing oneway tag.

karussell:
Fixes #976

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ch/PrepareEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/ch/PrepareEncoder.java
@@ -42,20 +42,27 @@ public class PrepareEncoder {
     }
 
     /**
-     * Returns true if flags1 can be overwritten in the edge by flags2 without restricting or
-     * changing the directions of flags1.
-     * <p>
+     * Returns 1 if existingEdgeFlags can be overwritten in the edge by newEdgeFlags without limiting or
+     * changing the directions of existingEdgeFlags. It returns 2 for the same condition AND if the edge
+     * with newEdgeFlags has to be added even if weight is higher than existing edge weight.
+     * <pre>
+     *                    | newEdgeFlags:
+     * existingEdgeFlags  | -> | <- | <->
+     * ->                 |  1 | 0  | 2
+     * <-                 |  0 | 1  | 2
+     * <->                |  0 | 0  | 1
+     * </pre>
      *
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/graphhopper/routing/ch/PrepareEncoder.java
```diff
@@ -42,20 +42,27 @@ public static final long getScBwdDir() {
...
-     * <p>
+     * Returns 1 if existingEdgeFlags can be overwritten in the edge by newEdgeFlags without limiting or
+     * changing the directions of existingEdgeFlags. It returns 2 for the same condition AND if the edge
+     * with newEdgeFlags has to be added even if weight is higher than existing edge weight.
+     * <pre>
+     *                    | newEdgeFlags:
+     * existingEdgeFlags  | -> | <- | <->
+     * ->                 |  1 | 0  | 2
+     * <-                 |  0 | 1  | 2
+     * <->                |  0 | 0  | 1
```
karussell: Interesting that these two cases (with '2' as response) are that raw in reality for shortcuts so that this bug remained so long in the code.

BTW: always **adding** the shortcuts without any updates dramatically slows down the preparation time. So this tuning with canBeOverwritten/getMergeStatus is very important.

karussell:
Uhm, somehow the PR was merged not squashed. The default should be squashing with keeping the commits for different authors separately or sth

karussell:
Also it seems that if I change the merging behavior (once or exceptionally) it is automatically defined as default which is ugly.

boldtrn:
> Also it seems that if I change the merging behavior (once or exceptionally) it is automatically defined as default which is ugly.

Yes, you need to always check. I think for me it even happened that when I changed it for one repo, it was changed for all repos. So this seems to be not repo dependent?

Issue #107, opened 2013-09-17
If no RAM available on import it should not successfully finish


karussell:
```
Exception in thread "PBF Reader" java.lang.OutOfMemoryError: GC overhead limit exceeded
    at java.util.concurrent.LinkedBlockingQueue.put(LinkedBlockingQueue.java:335)
    at com.graphhopper.reader.OSMInputFile.process(OSMInputFile.java:253)
    at com.graphhopper.reader.pbf.PbfDecoder.sendResultsToSink(PbfDecoder.java:106)
    at com.graphhopper.reader.pbf.PbfDecoder.processBlobs(PbfDecoder.java:171)
    at com.graphhopper.reader.pbf.PbfDecoder.run(PbfDecoder.java:184)
    at com.graphhopper.reader.pbf.PbfReader.run(PbfReader.java:49)
    at java.lang.Thread.run(Thread.java:724)
2013-09-17 08:26:08,866 [main] INFO  graphhopper.reader.OSMReaderHelper - finished way processing. nodes: 6359398, osmIdMap.size:32437443, osmIdMap:381MB, osmIdMap.toString:Height:4, entries:247803 totalMB:968, usedMB:931
```

boldtrn:
Is this issue still valid? I just tried it with the latest master and different amounts of RAM.

Either it hung up (I killed it, when there were not updates after the exception for at least a minute) or I received a `Process finished with exit code 1`, which should be OK? 

I tried about 10 different RAM settings, slowly increasing the RAM. Any ideas on how to reproduce this issue?

karussell:
Have not experienced this in a long time. But it could be a similar problem like I stumbled over recently when one landmark thread crashed with: 

```
2017-02-26 16:36:06,143 [fastest_car] ERROR com.graphhopper.routing.lm.LMAlgoFactoryDecorator - Problem while LM preparation fastest_car
com.carrotsearch.hppc.BufferAllocationException: Not enough memory to allocate buffers for rehashing: 67,108,864 -> 134,217,728
        at com.carrotsearch.hppc.IntObjectHashMap.allocateBuffers(IntObjectHashMap.java:1109)
        at com.carrotsearch.hppc.IntObjectHashMap.allocateThenInsertThenRehash(IntObjectHashMap.java:1134)
        at com.carrotsearch.hppc.IntObjectHashMap.put(IntObjectHashMap.java:179)
[...]

PR #980, opened 2017-03-06
Use CompletionService to fail early when a thread crashes

boldtrn:
As discussed in #107 it would be a good idea to switch the Runnable implementations of GraphHopper to implement Callable.

This PR does that for the CH preparation, I can add the same for all other implementations as well, once we decide to follow that implementation.

I tested this with car, ch preparation of fastest,shortest, with prepare.ch.threads=2, 3g of memory, and Baden-Würrtemberg.

With the new algorithm it took 4:13 minutes for the whole process and the preparation itself took 3:01 minutes.

With our current approach it took 4:08 minutes for the whole process and the preparation itself took 2:58 minutes.

I reran the approach of this PR again which took 4:01 minutes for the whole process and the preparation took 2:54 minutes.

I reran our current approach which took 4:04 and 2:53.

I haven't restarted my computer in between or anything, just ran the import. So the measurement might not have been completely unbiased, but it feels like the difference between both approach is not big, if there is a difference at all.

I wrapped the threadPool in an [ExecutorCompletionListener](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorCompletionService.html). This allows to catch Exceptions from any of the threads as soon as they are thrown and therefore stop the import ASAP.

I therefore expanded above described test scenario so that the second ch preparation thread throws an exception in the beginning. Somehow the other thread is not stopped, but finishes. Afterwards the whole process exits. **Edit:** To clarify, the RuntimeException in the Catch block is thrown and propagated up to the DefaultModule. But the other thread is not stopped. So maybe we can force stop the other thread. Once the other thread is finished, the process stops with exit code 1.


boldtrn:
I updated this PR. I check the Thread.interrupted flag every time a log is written during the ch preparation. I thought it might be a good idea to not check the interrupted flag all the time, for performance reasons and for keeping the code clean. So for long running imports it might take a couple of minutes until the process is actually exited after the Exception has been thrown.

WDYT? 
Should we do the same for the other Runnables?

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
```diff
@@ -270,6 +270,11 @@ void contractNodes() {
                         + ", algo:" + prepareAlgo.getMemoryUsageAsString()
                         + ", " + Helper.getMemInfo());
 
+                if(Thread.interrupted()){
```
karussell: Here we should use Thread.isInterrupted instead. Also we should check it more frequently otherwise it will be check just ~4 times per contraction. I expect no slow down here :)

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
```diff
@@ -279,38 +278,42 @@ public void setPreparationThreads(int preparationThreads) {
...
                         properties.put(errorKey, "CH preparation incomplete");
                         prepare.doWork();
                         properties.remove(errorKey);
                         properties.put(CH.PREPARE + "date." + name, Helper.createFormatter().format(new Date()));
                     } catch (Exception ex) {
                         LOGGER.error("Problem while CH preparation " + name, ex);
                         properties.put(errorKey, ex.getMessage());
+                        throw ex;
                     }
+                    return null;
```
karussell: the return should be before the catch (just my taste for the 'flow'). Also if we throw an exception I think there is no need anymore for logging as it should be logged later.

karussell:
Can you try if this works for mmap config too? See the comments here https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java#L82

boldtrn:
> The LM Runnables comes to my mind

Yes exactly. I also find Runnable implementations in the graphhopper.reader.osm.pbf package. PBF*Decoder uses Runnables. So maybe we should change these implementations as well?

> Can you try if this works for mmap config too?

I tried it both for the forced exception and a successful import. Booth worked as expected I think. I changed: `graph.dataaccess=MMAP_STORE_SYNC` (I think this is what you meant?).

Regarding the performance test:
Any idea/requirement on how to do this? I would just run the different imports and compare the while import time and the CH preparation time?

michaz:
Makes much sense, thanks.

I'm changing the title, since this is not so much about ``Runnable`` vs. ``Callable`` than about ``ExecutorService`` vs. ``CompletionService``. (In fact, ``CompletionService`` also supports ``Runnable``.)

karussell:
> PBF*Decoder uses Runnables. So maybe we should change these implementations as well?

better in a separate issue as I expect it to be more complex

> (I think this is what you meant?).

yes

> Regarding the performance test: Any idea/requirement on how to do this? I would just run the different imports and compare the while import time and the CH preparation time?

yes (~German area sized?)

boldtrn:
> Runnable vs. Callable than about ExecutorService vs. CompletionService

The Callable allows to throw Exceptions, whereas the Runnable does not allow to throw Exceptions. This allows us to check if there was an Exception during the preparation.

Compare the Interface: 
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Callable.html
https://docs.oracle.com/javase/7/docs/api/java/lang/Runnable.html

The CompletionService is only a nice helper here to return the first completed Callable. A Callable that thrown an Exception is also considered completed. So you could also write a loop over all Futures and check if they are finished instead of using the CompletionService.

boldtrn:
Ok so I ran a performance test for germany, car, 2 ch threads, 8gb RAM. This time I rebooted between every test to avoid caching issues. No other applications were running (haven't started anything after the reboot, except for the terminal).

```
rm -rf germany-latest.osm-gh/
./graphhopper.sh clean
export JAVA_OPTS="-Xmx8000m -Xms8000m"
./graphhopper.sh web germany-latest.osm.pbf
```

Runs                    | Start CH         | End CH         | Duration CH | Took (shortest, fastest)      |
----------------------- | ------------------ | ------------------ | ----------------- | ---------- |
Callable 1st run   | 13:20:35,629 | 13:55:01,102 |  34:26 min     | 952, 2065 |
Runnable 1st run   | 17:44:37,852 | 18:17:18,299 |  33:40 min     | 925, 1981 |
Callable 2nd run   | 21:41:56,790 | 22:15:21,278 |  33:25 min     | 921, 2004     |
Runnable 2nd run   | 14:27:46,318 | 15:03:05,655 |  35:19 min     | 917, 2119     |

Something that I found very suspicious, but seems to be unrelated to this PR is that fastest took about 2x as long as shortest.

The updates took place at about the same time (fastest was a bit slower), see for example this log:
```
2017-03-18 14:38:17,734 [shortest_car] INFO  com.graphhopper.routing.ch.PrepareContractionHierarchies - 8 608 000, updates:4, nodes: 360 195, shortcuts:6 955 056, dijkstras:166 322 010, t(dijk):520.02, t(period):285.05, t(lazy):62.82, t(neighbor):136.82, meanDegree:4, algo:131MB, totalMB:7437, usedMB:4105
2017-03-18 14:39:17,865 [fastest_car] INFO  com.graphhopper.routing.ch.PrepareContractionHierarchies - 8 608 000, updates:4, nodes: 261 462, shortcuts:5 853 008, dijkstras:149 496 112, t(dijk):586.94, t(period):307.53, t(lazy):62.43, t(neighbor):136.4, meanDegree:3, algo:131MB, totalMB:7437, usedMB:4283
```

But shortest finished way earlier and there were no updates from fastest anymore, see this example:
```
2017-03-18 14:43:03,844 [fastest_car] INFO  com.graphhopper.routing.ch.PrepareContractionHierarchies - took:917, new shortcuts: 6 621 545, prepare|fastest, dijkstras:170231044, t(dijk):800.81, t(period):335.71, t(lazy):124.35, t(neighbor):237.54, meanDegree:1, initSize:8608074, periodic:20, lazy:10, neighbor:20, totalMB:7437, usedMB:4854
2017-03-18 15:03:05,655 [shortest_car] INFO  com.graphhopper.routing.ch.PrepareContractionHierarchies - took:2119, new shortcuts: 8 774 820, prepare|shortest, dijkstras:277688755, t(dijk):1970.76, t(period):346.85, t(lazy):360.77, t(neighbor):1096.08, meanDegree:1, initSize:8608074, periodic:20, lazy:10, neighbor:20, totalMB:7481, usedMB:3602
```

You can see that shortest finished 20 minutes earlier. Whereas the whole CH process took about 35min. Especially weird is that there were no logs in that time. My first thought was that shortest needs longer to calculate and probably introduces more shortcuts than shortest, therefore it might take longer to calculate. But why did the updates happened more or less at the same time?

I tested the same for Baden-Württemberg, where the took log between shortest and fastest was not that big: 150, 206 (btw: I tested this with the Master Branch).

michaz:
> So you could also write a loop over all Futures and check if they are finished instead of using the CompletionService.

But how? I want to get() the first Future which finishes, immediately, so an Exception (if there is one) gets triggered and I can shut down quickly, right? And I don't know which one that is. The CompletionService solves that by using Futures which put themselves into the Queue once they are finished.

I just had a look at ``AbstractExecutorService`` -- if I'm not mistaken, Runnables are wrapped into Callables anyway.

karussell:
@boldtrn Thanks for this work & analysis! So the performance is not an issue and we should just pick the best/cleanest.

With fastest vs. shortest you mean you used the two different weightings for car? The differences can be big, yes. Also if you change the vehicles. We can certainly improve the heuristic to build the node order for CH, but it is very tricky to improve the average 'took' for all the different weightings and graph sizes.

boldtrn:
> But how? I want to get() the first Future which finishes, immediately, so an Exception (if there is one) gets triggered and I can shut down quickly, right? And I don't know which one that is. The CompletionService solves that by using special Futures which put themselves into the Queue once they are finished.

It's actually quite simple. I just changed the code of the CHAlgoFactory to this:
```
    public void prepare(final StorableProperties properties) {
        List<Future> runningPreparations = new ArrayList<>();
        int counter = 0;
        for (final PrepareContractionHierarchies prepare : getPreparations()) {
             ...... the regular code .....
            runningPreparations.add(threadPool.submit(new Callable<Void>() {
                @Override
                public Void call() throws Exception {
                    ...... the regular code .....
                    }
                }
            }));

        }

        threadPool.shutdown();

        try {
            while (true){
                List<Future> toDelete = new ArrayList<>();
                for (Future future : runningPreparations) {
                    if (future.isDone()) {
                        future.get();
                        toDelete.add(future);
                    }
                }

                runningPreparations.removeAll(toDelete);
                if(runningPreparations.isEmpty()){
                    break;
                }
            }
        } catch (Exception e) {
            threadPool.shutdownNow();
        }
    }
```

I am not overly proud of that code but it shows how we could do this ourselves. I use the ExecutorCompletionService only for convenience as I think their code is better than mine and probably better tested :).

> I just had a look at AbstractExecutorService -- if I'm not mistaken, Runnables are wrapped into Callables anyway.

Yes, but you cannot throw a checked Exception in a Runnable, so I would assume that it might not work correctly with Runnable (haven't tested it though).

> @boldtrn Thanks for this work & analysis! So the performance is not an issue and we should just pick the best/cleanest.

I think so, yes.

> With fastest vs. shortest you mean you used the two different weightings for car?

Yes the two standard implementations. I was just wondering that the difference became that big in the end, whereas the regular updates were more or less at the same time. So I wanted to make sure that there is no hidden bug or anything.

michaz:
> I use the ExecutorCompletionService only for convenience as I think their code is better than mine

In particular, it doesn't keep a whole core busy with looping, yes.

michaz:
@boldtrn since you're being empirical already :-), would you mind running another round where you test for ``Thread.currentThread().isInterrupted()`` really often, like, all the time? So that we finish almost immediately when a thread crashes? I'd be interested in whether that matters.

boldtrn:
> In particular, it doesn't keep a whole core busy with looping, yes.

Yes, we could build something like this as well, using a BlockingQueue, but why reinventing the wheel :).

> @boldtrn since you're being empirical already :-) (and probably still have the experiment set up), would you mind running another round where you test for Thread.currentThread().isInterrupted() really often, like, all the time?

[It is already checked very often](https://github.com/graphhopper/graphhopper/pull/980/files#diff-b0eb0b2e6b07deaa542301bbc8904ca9R324). I changed that from my initial approach. As both approaches seem to perform comparably equal I assume calling this method is fast. Should I still test/change anything?

michaz:
> It is already checked very often.

Ah nice, sounds good.

> Should I still test/change anything?

Nope, I think that's it, thanks.

michaz:
> Yes, but you cannot throw a checked Exception in a Runnable, so I would assume that it might not work correctly with Runnable (haven't tested it though).

But this code does not throw checked Exceptions. Otherwise it wouldn't compile with ``Runnable``, since they are checked. Really, the difference between ``Runnable`` and ``Callable`` is about syntax and the type system, not about the multi-threading behavior. Convince me otherwise. :-)

In Java 8, the typical use of both interfaces is as a lambda, where you don't even see the difference very prominently.

> Yes, we could build something like this as well, using a BlockingQueue, but why reinventing the wheel :).

Yes _indeed_, I was just trying to re-visit my point that it is actually kind of what enables the feature we are trying to implement with this ticket, i.e. failing early on an exception, at least very much more so than the difference between ``Runnable`` and ``Callable``, and I wanted to check if we are in agreement about that now, because that did not seem to be the case when I renamed the ticket.

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
```diff
@@ -320,6 +320,10 @@ void contractNodes() {
 
                 prepareGraph.disconnect(vehicleAllTmpExplorer, iter);
             }
+
+            if(Thread.currentThread().isInterrupted()){
```
karussell: This is outside of the main loop, we would need it inside. Best would be directly after the while start to include the cases where `continue` is executed.

boldtrn:
> Convince me otherwise. :-)

When comparing the ThreadPool solution, Runnable vs. Callable, there was no way to check if the Runnable was successful or not (we set CFG variables, which seem to have failed sometimes).

I tested the CompletionService a bit more. You are right, it seems to handle the Exceptions of the Runnable as well, which is really nice, so we wouldn't even have to switch to the Future implementation. Should we stick with Future or should I revert that back to Runnable to make the change even smaller?

> feature we are trying to implement with this ticket, i.e. failing early on an exception, (that, and the periodic checking for isInterrupted())

Actually this PR is about fixing #107, where even though an Exception was thrown in one of the Runnables the import finished successfully, which should not happen.

Failing fast, is "just" a bonus IMHO.

BTW: Something I have seen during tests is that sometimes the process get's stuck. This looks this:

```
2017-03-20 11:57:11,505 [main] INFO  com.graphhopper.routing.ch.CHAlgoFactoryDecorator - 2/2 calling CH prepare.doWork for shortest|car ... (totalMB:360, usedMB:204)
Exception in thread "fastest_car" Exception in thread "shortest_car" 
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "shortest_car"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "fastest_car"
```

It will be stuck like this at least for a very long time (I terminated the process, after some time). Most of the time, the exception will be handled correctly, but sometimes it's just don't. So I am not sure how this might happen, but my idea would be that this might be created by a deadlock due to `properties.put(errorKey, ex.getMessage());`. Put is synchronized (which is good), but maybe the OutOfMemoryError was raised more or less at the same time for both threads. One thread wants to put the errorKey. The other threads wants to put the error key. Due to some reason the first thread might get killed, due to OutOfMemory. If have no idea if this is possible, but since the Exception is not reproducible and haven't occured since, this might have been the reason. But maybe you have a better idea why this might happen?

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
boldtrn: Ok, I moved it to the inner while. I also measured the performance, which has not decreased, see the updated table above.

karussell:
> should I revert that back to Runnable to make the change even smaller?

Yes, please. 

Would you mind to create another PR or changes, where you you do this for the LM decorator?

> So I am not sure how this might happen, but my idea would be that this might be created by a deadlock due to 

How can one thread being stuck inside the properties.put? I think the OOM will free all locks: http://stackoverflow.com/q/2019339/194609

Also could this really be - I see you have removed this line? (As we need this, would you revert this?)

michaz:
> so we wouldn't even have to switch to the Future implementation.

Ah, okay. I didn't see that. If that's possible, all the better.

boldtrn:
> Yes, please.

Ok sure, I reverted back to Runnable.

> Would you mind to create another PR or changes, where you do this for the LM decorator?

Not a problem :)

> How can one thread being stuck inside the properties.put? I think the OOM will free all locks: 

This was just a wild guess. Maybe it happens somewhere else. It seems to be deadlock when the two threads are stopped (OutOfMemory) at the "same" time. Could be something completely different, this is just my guess at this moment.

> Also could this really be - I see you have removed this line? (As we need this, would you revert this?)

Yes, I haven't seen the issue after I removed the line. But since it is not reliably reproducible I cannot say if this fixed it. I added the line back. Where is this line needed? I haven't found any uses of this? It was used to recognize exceptions, but this is handled now via throwing actual Exceptions from the Runnable.

> Ah, okay. I didn't see that. If that's possible, all the better. My impression was that you would have to get() the Future.

Yes, but now I use the wrapped Future of the ExecutorService. In the end it doesn't change that much, but the change of this PR is a bit smaller.

boldtrn:
BTW: I reran the performance measure for Germany again. Seems to be equal to before, so should be good to be merged?

karussell:
:+1: from me. The only (later?) thing is the deadlock which bugs me.

wdyt @michaz?

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
```diff
@@ -279,38 +278,41 @@ public void setPreparationThreads(int preparationThreads) {
...
                         properties.remove(errorKey);
                         properties.put(CH.PREPARE + "date." + name, Helper.createFormatter().format(new Date()));
                     } catch (Exception ex) {
                         LOGGER.error("Problem while CH preparation " + name, ex);
                         properties.put(errorKey, ex.getMessage());
+                        throw ex;
                     }
                 }
-            });
+            }, name);
```
karussell: If we have the name specified here, then we do not need the lines 291+292?

boldtrn: After the thread finishes the Future returns this variable as result. We could write any result that we want, I just figured the name would be nice and could be usable in the future. For example when retrieving the result we could state: "Finished Thread: "+name. This is different from the Thread name.

michaz:
Looks good to me from the outside. :)

PR #982, opened 2017-03-09
MMapDataAccess should synchronize on its ByteBuffers by itself

michaz:
- MMapDataAccess now synchronizes on its ByteBuffers by itself, and is now as thread-safe as the other implementations. (That is, mostly, except for create/close/ensureCapacity and such.)
- Remove the dimension of "synched" or "not synched" from DataAccess.
- Remove tests which tested this dimension.
- Remove some unused code.

Benefits:

- One whole concept less.
- Clarifies that we do read/write locking externally, i.e. on the layer of GraphHopper, not on the layer of DataAccess.
- Cannot accidentally be made incorrect by using more than one (reading) thread without using "_SYNC".

Drawbacks:

- Possibility of being slower when running single-threaded because of the overhead of synchronized{} blocks _as such_, without thread contention. No idea how likely that is.

Counter-argument against drawback:

- That is not as bad as being *generally* thread-safe *except* with certain parameters.

How can I test Android performance?


karussell:
Thanks, this looks better.

> How can I test Android performance?

That is tough. I have not done this properly/scientific. Also there are so many devices where we could test. So I would just route from two identical places a few times and repeat the same for the change or write a bit code to do this in the demo (e.g. when click&pressing some device button). And then @devemux86 could repeat. Or we could see if there are CI's that run Android somehow deterministically where we could include this in the build process.

karussell:
BTW: For the JVM I would exclude any problems due to synchronized. It is VERY fast even with synch. But for dalvik I have no idea. (What a mess Google&Oracle introduced with this ...)

karussell:
This improves the situation for situations on the server side with only a few memory. So although I didn't test concurrent scenario (should be much better) but single threaded is nearly unchanged. So I'll merge and we later see & test different Android devices.

   method  | RAM_STORE | old MMAP | old MMAP_SYNC | new MMAP
---- | ----------------- | ------------- | ---------------------- | -------------
routing| 211 | 254 | 274 | 277
routingCH| 1.9 | 2.0 | 2.8| 2.9
routingLM8| 15 | 18 | 32.5| 33
routingLM4| 16 | 17.3 | 29| 27.6

michaz:
I read this table as "has approximately the same performance as MMAP_SYNC". Right?

devemux86:
Nice results!

michaz:
We should also do an experiment with ThreadLocal<ByteBuffer>, where each thread takes its own mappedByteBuffer.slice(). I'm pretty sure that doesn't use more resources _at all_ and should be closer to "old MMAP" than to "new MMAP".

michaz:
One more thing: When testing this myself, I noted that MMap needs a while to "warm up" (understandable, since the OS only pages in the segments it needs).

Did you discard the first couple of requests? I think that would be fair to MMap.

karussell:
> I read this table as "new MMAP has approximately the same performance as old MMAP_SYNC". Right?

Yes, but _SYNC should be worse than new mmap for multiple threads. How much I have not tested though

> I'm pretty sure that doesn't use more resources at all 

It will need a new Buffer list for every thread, so 0 new resources is not 100% correct ;)

> ThreadLocals are a bit scary, but I think if one googles thoroughly, one can avoid the one or two pitfalls

Currently I do not see a major pain and so I would keep using in-memory stuff as ThreadLocal can be indeed tricky and memory mapped stuff is also a bit ugly at times and you cannot disable swapping :)

And there is only one drawback with in-memory: the GC has a bit problems with huge static memory. But G1 is not that bad and maybe the Shenandoah GC will be better :)

> Did you discard the first couple of requests? I think that would be fair to MMap.

If you use the Measurement class this is already done and should be done for in-memory too as the JVM needs a warmup. But you have to change "forDesktop" to change to MMAP DAType.

Also there is also QueryTorture.java with which you can real queries with many threads.

BTW: I would do perf measurements from scratch these days with jmh

PR #987, opened 2017-03-16
SpatialRuleLookup Refactoring

boldtrn:
I am just creating a second PR here, to discuss @michaz nice refactoring of the SpatialRuleLookup. I cannot really comment on the Landmark part, this should be done by @karussell IMHO. I cannot create this PR on top of my PR as the spatial branch is in the graphhopper repo and my branch is in my gh fork. I would be fine with simply closing my other PR?

I really like the concept of using simple DataTypes as defaults instead of null/empty checks :+1:. Something I don't like is that we cannot create a smaller BBox than provided by the rules anymore.


core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -91,8 +88,27 @@
     private boolean storeWeight = false;
     private boolean storeWidth = false;
     private EncodedValue spatialEncoder;
-    private int spatialRules = 0;
-    private SpatialRuleLookup spatialRuleLookup;
+    private SpatialRuleLookup spatialRuleLookup = new SpatialRuleLookup() {
```
boldtrn: Should we create the SpatialRuleLookup.Empty again?

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
```diff
@@ -378,4 +391,48 @@ public void testSpatialId() {
         assertEquals(5, encoder.getMaxspeed(e3, -1, false), .1);
         assertEquals(-1, encoder.getMaxspeed(e4, -1, false), .1);
     }
+
+    /**
+     * Defines the default rules for German roads
+     *
+     * @author Robin Boldt
+     */
+    private static class TestRule extends DefaultSpatialRule {
```
boldtrn: I think we can cleanup this implementation a lot, or just use the GermanSpatialRule again?

web/src/main/java/com/graphhopper/countries/GermanySpatialRule.java
```diff
@@ -15,17 +15,18 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  */
-package com.graphhopper.routing.util.spatialrules.countries;
+package com.graphhopper.countries;
```
boldtrn: Do we really want this at com.graphhopper? I think it would be better fitting under routing?

web/src/main/java/com/graphhopper/srlarray/SpatialRuleContainer.java
```diff
@@ -15,7 +15,9 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  */
-package com.graphhopper.routing.util.spatialrules;
+package com.graphhopper.srlarray;
```
boldtrn: I really don't like that package name :). Can we rename it?

web/src/main/java/com/graphhopper/srlarray/SpatialRuleContainer.java
```diff
@@ -26,7 +28,7 @@
  */
 class SpatialRuleContainer {
 
-    protected final Set<SpatialRule> rules = new LinkedHashSet<>();
+    final Set<SpatialRule> rules = new LinkedHashSet<>();
```
boldtrn: why do you remove protected?

web/src/main/java/com/graphhopper/srlarray/SpatialRuleLookupArray.java
```diff
@@ -115,7 +131,7 @@ public SpatialRule lookupRule(double lat, double lon) {
         return SpatialRule.EMPTY;
     }
 
-    protected int getRuleContainerIndex(int xIndex, int yIndex) {
+    private int getRuleContainerIndex(int xIndex, int yIndex) {
```
boldtrn: why protected -> private

web/src/test/java/com/graphhopper/srlarray/SpatialRuleLookupArrayTest.java
```diff
@@ -15,14 +17,26 @@
 
     @Test
     public void testSpatialLookup() {
-        SpatialRuleLookupArray lookup = new SpatialRuleLookupArray(new BBox(1, 2, 1, 2), 1, false);
-        SpatialRule germanyRule = new GermanySpatialRule().addBorder(new Polygon(new double[]{1, 1, 2, 2}, new double[]{1, 2, 2, 1}));
-        lookup.addRule(germanyRule);
-        SpatialRule austriaRule = new AustriaSpatialRule().addBorder(new Polygon(new double[]{5, 5, 6, 6}, new double[]{5, 6, 6, 5}));
-        lookup.addRule(austriaRule);
+        List<SpatialRule> spatialRules = new ArrayList<>();
+        SpatialRule germany = new DefaultSpatialRule() {
```
boldtrn: Why aren't we just using the actual implementation?

web/src/test/java/com/graphhopper/srlarray/SpatialRuleLookupArrayTest.java
```diff
@@ -31,23 +45,25 @@ public void testSpatialLookup() {
...
-        assertEquals(AccessValue.EVENTUALLY_ACCESSIBLE, spatialRuleLookup.lookupRule(1.2, 1.7).getAccessValue(null, TransportationMode.MOTOR_VEHICLE, AccessValue.ACCESSIBLE));
+        List<SpatialRule> spatialRules = new ArrayList<>();
+        spatialRules.add(getSpatialRule(new Polygon(new double[]{1, 1, 2, 2}, new double[]{1, 2, 2, 1}), "1"));
+        spatialRules.add(getSpatialRule(new Polygon(new double[]{1, 1, 3.6, 3.6}, new double[]{3, 4, 4, 3}), "2"));
+
+        SpatialRuleLookup spatialRuleLookup = new SpatialRuleLookupArray(spatialRules, 1, false);
+
+        Assert.assertEquals(AccessValue.EVENTUALLY_ACCESSIBLE, spatialRuleLookup.lookupRule(1.2, 1.7).getAccessValue(null, TransportationMode.MOTOR_VEHICLE, AccessValue.ACCESSIBLE));
         assertEquals(AccessValue.EVENTUALLY_ACCESSIBLE, spatialRuleLookup.lookupRule(1.2, 3.7).getAccessValue(null, TransportationMode.MOTOR_VEHICLE, AccessValue.ACCESSIBLE));
-        // Not in the second Polygon anymore, but due to the resolution of 1, this should be still match the rule
```
boldtrn: Why did you remove this test? This test shows that the exact flag actually works.

web/src/test/java/com/graphhopper/srlarray/SpatialRuleLookupArrayTest.java
```diff
@@ -116,7 +141,7 @@ private Polygon parsePolygonString(String polygonString) {
     }
 
     private SpatialRule getSpatialRule(Polygon p, final String name) {
-        SpatialRule rule = new AbstractSpatialRule() {
+        AbstractSpatialRule rule = new AbstractSpatialRule() {
```
boldtrn: why?

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
michaz: It should only contain what's necessary for the test (feel free to reduce). I didn't want to use Germany because I wanted to move Germany out of the core and make it private, to illustrate that the two countries together with their factory are self-contained.

web/src/main/java/com/graphhopper/countries/GermanySpatialRule.java
michaz: It's not in the core, it's in the web module, and it's just an example. I don't know where it will be in the end.

web/src/main/java/com/graphhopper/srlarray/SpatialRuleContainer.java
michaz: Sure.

web/src/main/java/com/graphhopper/srlarray/SpatialRuleContainer.java
michaz: Because the whole class is package-private, so protected on a method doesn't do anything. Also the class doesn't have sub-classes (and shouldn't), so protected doesn't do anything at all.

web/src/main/java/com/graphhopper/srlarray/SpatialRuleLookupArray.java
michaz: Because it's possible. Less public methods -> better. (protected is almost public)

michaz:
> If I only want to route in Berlin, I don't want to create the Lookup for Germany. [...]
> return the Empty SpatialRuleLookup to save storage?

Aha! The Lookup uses memory in proportion to its spatial extent! So maybe *that* is the root of the problem, and we should use a standard implementation from a library, which does not have this property. :-)

michaz:
(Just kidding. That's a different issue, really. But you see my point a bit?)

michaz:
> Der Lookup selbst, inklusive Array wird aktuell nur für den Import benötigt. Danach kann man den wegschmeissen. [...] Wenn du den Graph von der Platte lädst, brauchst du den Lookup gar nicht erzeugen.

Aha! Das sieht man dem ganzen Ding aber nicht an. Es hängt halt am FlagEncoder dran, und dieses hin- und herreichen und kodieren der Rule-Nummer macht den Eindruck, als würde man die SpatialRule auch zur Laufzeit noch mal haben wollen können.

karussell:
Would you mind to use English? I know this is a bit harder if all involved know German but in the issue it is important that all can participate.

> kodieren der Rule-Nummer macht den Eindruck, als würde man die SpatialRule auch zur Laufzeit noch mal haben wollen können

The idea of storing the rule number was that we can use this information in the GenericWeighting e.g. making a "is germany" request. With that the user could e.g. exclude a country from routing or increase a specific property per country etc.

karussell:
>  I cannot really comment on the Landmark part, this should be done by @karussell IMHO

@boldtrn  Yes, please keep the LM changes out in this request. I'll consider the necessary changes in another PR.

michaz:
> The idea of storing the rule number was that we can use this information in the GenericWeighting e.g. making a "is germany" request. With that the user could e.g. exclude a country from routing or increase a specific property per country etc.

This means the DataFlagEncoder can just assign increasing numbers to all different SpatialRules it sees and write that on the link. Voilà, one more thing the SpatialRuleLookup doesn't need to know about. :-)

karussell:
> This means the DataFlagEncoder can just assign increasing numbers to all different SpatialRules it sees and write that on the link. 

The number is usually encoded in only a few bits, so we need to know the necessary bits before allocating the bit ranges.

michaz:
> The number is usually encoded in only a few bits, so we need to know the necessary bits before allocating the bit ranges.

Yeah, no problem. size() is fine.

karussell:
Then I do not understand what you meant which part the SRL doesn't need to know. Did you mean the ID?

michaz:
It could even return the whole (read-only) collection of SpatialRules for the encoder to look at (including the size), that's fine. (Maybe useful, because it needs to know the range of some other attributes which the SpatialRules want to have encoded..)

It's just that SpatialRuleLookup doesn't need the (additional) concept 'id number'. (The rules go in in an arbitrary order anyway.)

michaz:
We're getting to the part where SpatialRuleLookup is a location-indexed-collection<SpatialRule>, is my point. Without any special operations or extra concepts at all.

karussell:
> It's just that SpatialRuleLookup doesn't need the (additional) concept 'id number'. (The rules go in in an arbitrary order anyway.)

Only if we would return a `List<>` (and not just a `Collection<>`) which keeps its order. Hmmh, although we could use the SpatialRule its ID to avoid this requirement.

michaz:
Okay, that depends only on one thing: Does the user want any kind of control over those numbers?

If yes: Then the SpatialRules should have that number in them and return it (as a normal attribute), without any hard requirements on uniqueness or range. DataFlagEncoder can then scan them for the range, and use them.

If no: Then the DataFlagEncoder can just create the ordered List itself when it starts up and go ahead. The numbers then 'belong only to' DataFlagEncoder.

michaz:
*Or* we say we *do* want the SpatialRuleLookup to be always there, also at runtime. But that's an important distinction. :)

karussell:
> If no: Then the DataFlagEncoder can just create the ordered List itself when it starts up and go ahead. The numbers then 'belong only to' DataFlagEncoder.

Yes, it is this case. I do not know what you mean with the 'middle-ground' case.

michaz:
The way it is now.

web/src/main/java/com/graphhopper/countries/GermanySpatialRule.java
boldtrn: Yes I know, but why? This should be a core feature. For example if I want to use it from the MiniGraphUI (just one example, think of any usage that does not need the web module), I would need to add a dependency to the web module. That seems wrong to me.

boldtrn:
> Aha! The Lookup uses memory in proportion to its spatial extent! So maybe that is what's in our way at some points, and we should use a standard implementation from a library, which does not have this property. :-)

Sure, the same would be true for a tree or any data structure. If you only create the tree for Berlin it would be smaller than the tree for Germany (if implemented well, if not it might have the same size).

> Or are we sure we need a raster-based lookup for performance reasons?

No I just wanted to get started fast, that's why I started with a raster. We can always add a spatial tree or anything else. However, my personal believe is that the array will outperform the tree on average. For most requests the array only needs a table lookup¹, only if you need precision, you get a Polygon contains check² for border tiles or tiles that intersect with other countries. The tree might be better memory wise though, but memory shouldn't be the issue at this point. Do you see an issue with the raster-based lookup?

1: O(1)
2: We use the ray casting algorithm, which does some simple arithmetics for every edge, so O([E]), which does not depend on the data structure but on the polygon.

> This means the DataFlagEncoder can just think up increasing numbers for the different SpatialRules it sees, and write that on the link. Voilà, one more thing the SpatialRuleLookup doesn't need to know about. 

No, the SpatialRuleLookupArray builds an index int->Rule internally, we just use the index. It remains stable as long as no one tries to manipulate the list of rules. So this is part of the responsibility of the SpatialRuleLookupArray.

> We're getting to the part where SpatialRuleLookup is some LocationIndexedCollection<SpatialRule>, is my point. Without any special operations or extra concepts at all.

Yes I agree, but where does it break that boundary in your opinion? If you have a Collection you can get the id of one of it's entries, if it's ordered. We simply do the same?

web/src/main/java/com/graphhopper/countries/GermanySpatialRule.java
karussell: Adding the web module for using the SR lookup might not be optimal but putting it into core does not make sense as core is not able to parse the json file. So we could introduce a util module and depend on core&json&osm (?)

boldtrn: Yes, let's keep the JSON parsing in the web module for now. But the implementations of the SpatialRule like Germany, Austria, or the SpatialRuleLookupArray, don't have anything to do with json and should live in the core. Don't you think?

karussell: Probably a matter of taste :)

boldtrn: Not sure. How about using SpatialRules in any non web related project? There is no implementation of the SpatialRuleLookup in the core. This feels wrong to me.

karussell: As I said: in the web module it is probably wrong, and IMO it can be in the core in a separate package.

michaz:
> Sure, the same would be true for a tree or any data structure. If you only create the tree for Berlin it would be smaller than the tree for Germany (if implemented well, if not it might have the same size).

@boldtrn Oh, you mean if we _insert fewer Polygons_ into the index, aka insert fewer SpatialRules into the SpatialRuleLookup. _Then_ sure. My impression has been, all the time, and still is, that the set of _SpatialRules_ is basically user code. And that someone who only wants Berlin will simply not insert SpatialRules for the whole world.

I mean, until recently, the SpatialRules were loaded by config file. Why would the user configure Rules outside his range of interest?

michaz:
Ok, I'm getting there. The idea is that a whole database of countries and their rules is going to be maintained by us, delivered, and be part of the core. By default.

And since the user may still want to route only in Berlin, they should be able to configure a bounding box.

Is that correct?

karussell:
Yes. Also for this specific implementation via arrays a bounding box reduces memory usage even if just one SpatialRule is used.

BTW: isn't this memory reduction the case for a standard library too? As one could intersect the bbox with the polygon(s) before feeding?

michaz:
> BTW: isn't this memory reduction the case for a standard library too? As one could intersect the bbox with the polygon(s) before feeding?

Exactly! That's why I keep saying "it doesn't matter", because in my mind, this happens in the outermost layer, in the self-written main-method of my own GraphHopper instance, which I write for my own server / app / whatever.

Perhaps I was too entrenched in the idea that GraphHopper is a library, and our instance is our instance and the user's instance is the user's instance.

Ok, so we want to truncate the index at the bounding box of the loaded graph once we know it.

karussell:
Ah, now I understood you too :) !

> And since that's not possible because we need the whole thing for loading the graph, we let the user specify it

Yes. Although it now comes to my mind that we process the OSM file two times and we could estimate a bounding box in the first pass and need the lookup in the second if I'm not wrong...

boldtrn:
> @boldtrn Yes, please keep the LM changes out in this request. I'll consider the necessary changes in another PR.

Ok, I removed it.

> @boldtrn Oh, you mean if we insert fewer Polygons into the index, aka insert fewer SpatialRules into the SpatialRuleLookup. Then sure. My impression has been, all the time, and still is, that the set of SpatialRules is basically user code. And that someone who only wants Berlin will simply not insert SpatialRules for the whole world.

Ah ok :). Yes that was the plan. It was meant to be user extendable/changeable, but the default country rules are provided by us for the most common use-cases, as not everyone wants to modify this (I guess).

core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
boldtrn: Since I moved the GermanySpatialRule back to the core, I think it should be fine if we reuse it in the tests again? Or should we create our own implementation again?

boldtrn:
Ok, so I worked a bit on this PR a bit more. I think it should be mergeable now. I allowed a user to simply configure the BBox in the config.properties file. As it makes sense to allow a user to user to use regular routing for example in Berlin without needing to either change a line of code or use unnecessary amounts of RAM.

I moved the SpatialRules back into the core.

I added the getBounds method back to the SpatialRuleLookup, as this made writing tests for different BBoxes easier.

Would you mind to have a look at my changes @michaz?

michaz:
> That's true yes. Should we consider this? I could check possible solutions for this? But maybe in a different PR?

Thanks! I think it would nicely fit in here, would it be ok if you go ahead and I have a look afterwards? (Got to finish the pt thing...) (If it doesn't work out, we can just merge this commit here.)

Suggestion: Maybe try and put the whole loading of the SpatialRules, everything, to a point in time where the graph bbox is already known. So that, conceptually, we are filtering from the outside (only inserting those rules which are interesting). Then, that flow would work exactly the same with another index which doesn't need or want the bbox.

Maybe there's already a hook somewhere which would fit, otherwise we should create one. Right, @karussell ?

karussell:
> Maybe try and put the whole loading of the SpatialRules, everything, to a point in time where the graph bbox is already known

But the problem is (at least if I correctly understood the issue) that we need the bbox before the graph bbox is known. I.e. we need the SpatialRuleLookup to modify access and speed while we are filling the graph. The only solution would be to have a deeper integration with OSMReader where we would calculate the bbox in the first run (determine all involved OSM nodes) and then create the SpatialRuleLookup before the second run starts. But this requirement (1. determine bbox, 2. run again) could make the integration of other data sources complicated.

michaz:
> The only solution would be to have a deeper integration with OSMReader where we would calculate the bbox in the first run (determine all involved OSM nodes) and then create the SpatialRuleLookup before the second run starts. But this requirement (1. determine bbox, 2. run again) could make the integration of other data sources complicated.

All right - then bbox by config file it is. :)

boldtrn:
> The only solution would be to have a deeper integration with OSMReader where we would calculate the bbox in the first run (determine all involved OSM nodes) and then create the SpatialRuleLookup before the second run starts.

Yes and makes it more complicated when using a prepared graph. We have two different ways of finding the BBox.

a) When importing, find out the BBox from the OSM file (or any other source file).
b) When loading a prepared graph we have to get the BBox from the graph.

> All right - then bbox by config file it is. :)

Ok, we can always improve it to automatically determine the BBox in the future :). The DataFlagEncoder is not used in production, yet. So we don't break anything if we change anything in the future.

So is this branch mergeable? Or should I change anything?

karussell:
From my side only one question @michaz: why do you think we do not need to store the number of spatial rules? The problem is if the SRlookup is configured with e.g. two rules and an import is done and then restarted with 3 rules we have a mismatch. 

So we need at least a bit safety (the number check and a stable order e.g. via an ArrayList of rules which is then sorted by the ID or similar)

michaz:
> The problem is if the SRlookup is configured with e.g. two rules and an import is done and then restarted with 3 rules we have a mismatch.

Hmm? I thought we don't want the rules at runtime anymore (I asked, somewhere up there..), and after the import is complete, the rule numbers are only numbers? Isn't that true?

michaz:
> and we have to keep the actual rules stable.

.. which is difficult, because the rules are _code_. Which, to connect to our earlier discussion on the phone (about something else), is probably why the Java serialization mechanism is clunky (because it's difficult).

karussell:
> Hmm? I thought we don't want the rules at runtime anymore (I asked, somewhere up there..)

The runtime usage is still there if users e.g. want to avoid passing the border, this cannot be achieved without feeding the rule IDs. But we can delay this requirement into another issue.

> Or do we say we don't care and use the number of rules as a kind of poor person's :-) checksum?

Yes, kind of. Also I expect that the boundaries are kind of stable and so the number association to the edge remains stable (of course only if the order of the rules/IDs hasn't changed).

michaz:
Thinking some more about this, I think we can and should be principled about this:

I suggested earlier to put the number into the code, like the other rule attributes. Then everything a rule puts into an edge is in one place: the code. Would that still make sense?

Then we say: The rules are only there when creating the graph. We don't even (need to) load them when the graph is loaded-existing. What the rules have written into the edges when the graph was created, is in the edges -- and that's it.

I think re-loading the rules just to check that the rules are the same as they were when the graph was created, but not _really_ checking (but only counting), and also not really _using_ that information except to maybe give an error message -- to me, that would be confusing because I wouldn't understand what we're doing there and why.

michaz:
Or (also very valid to me!) we say, yes, the rules *are* there at run-time, they are always loaded, the ``DataFlagEncoder`` always looks the same, create-time or run-time, and code which has an edge can get the ``SpatialRule`` for it.

karussell:
> Or (also very valid to me!) we say, yes, the rules are there at run-time, they are always loaded, the DataFlagEncoder always looks the same, create-time or run-time, and code which has an edge can get the SpatialRule for it.

That is what I meant&want. Still we want also the feature that some 'static' attributes (e.g. country dependent default speeds or access restrictions) can be overwritten at feeding time. For those there is no reason to do this at runtime as they are static.

michaz:
> Der Lookup selbst, inklusive Array wird aktuell nur für den Import benötigt. Danach kann man den wegschmeissen. [...] Wenn du den Graph von der Platte lädst, brauchst du den Lookup gar nicht erzeugen.

> Or (also very valid to me!) we say, yes, the rules are there at run-time, they are always loaded, 

> That is what I meant&want.

Okay. Good that we sorted this out. :)

> Still we want also the feature that some 'static' attributes (e.g. country dependent default speeds or access restrictions) can be overwritten at feeding time. For those there is no reason to do this at runtime as they are static.

.. but I don't understand that now.

karussell:
> .. but I don't understand that now. Can you give an example / point to code, please?

I'll try :) ... actually there are two requirements:

 1. country specific defaults (like for maxspeeds and access restrictions)
 2. country dependent routing (e.g. avoid Switzerland when you route from Germany to Italy)

Both requirements can be satisfied with the 'runtime SpatialRuleLookup'. But the first requirement can be satisfied also if we detect the country while import and change the speed already there, this makes the routing also a bit faster. So I think that in general we should make it possible to either precalculate stuff while import (faster!) or calculate on the fly while routing (more flexible!).

The usage for the 1. requirement is already done in code [here](https://github.com/graphhopper/graphhopper/pull/987/files#diff-82652cae32e0a3b3bfec84b8ec3e3740R322) and the runtime usage was sketched [here](https://github.com/graphhopper/graphhopper/pull/926/commits/0ff2ce7e80e6674629602efa0840bd4bcfe94108#diff-d5c8ccb9a1670d54256e7dde81d0092d) (although it was done for the access properties there, just imagine you avoid a certain country in the weighting)

Is this more clear now?

michaz:
```
/**
* Returns the rule for a certain id.
* If no such id is registered, return "EMPTY".
*/
public SpatialRule getRuleForId(int id) {
```

_Aha!_ :) Well, then please understand some of my confusion, because there is no such method on ``SpatialRuleLookup`` so this reviewer had little way of knowing that this was intended. :)

boldtrn:
> Aha! :) Well, there is no such method on SpatialRuleLookup so this reviewer had little way of knowing that this was intended, and thought that this was intentionally left out. :)

You are right, for some reason this method only exists in the SpatialRuleLookupArray.

I think the problem right now is that we are not really sure yet, what we want to do. My idea at some point was to separate two concepts:

SpatialRuleLookup, only used for the import. Used to get the SpatialRule for a ReaderWay.
SpatialRuleRegistry, references every rule with a unique id¹. The Registry is available at the Import and Runtime and can provide the SpatialRule for every edge (since the edge has the id). This would also work very well with the DataFlagEncoder concept. The rules would be then evaluated in the Weighting. There was a version of this code when initially creating the SpatialRules. I think the problem was that the Weighting became too slow, so we changed it again, is this correct @karussell?
This is the [commit] (https://github.com/graphhopper/graphhopper/pull/926/commits/0ff2ce7e80e6674629602efa0840bd4bcfe94108) that introduced the Registry. We removed the Registry because it was too losely coupled to the SpatialRuleLookup.

1: This might be completely independet of the ids used internally by the SpatialRuleLookup to store the rules.

karussell:
>  I think the problem was that the Weighting became too slow, so we changed it again, is this correct @karussell?

Just wanted to make it simpler and keep the runtime stuff away from the first PR. Didn't make performance comparisons.

> public SpatialRule getRuleForId(int id) {

Probably this method should be in the DataFlagEncoder as only the encoder will know the integer number/id. So the idea is that the DataFlagEncoder creates a LinkedHashMap and maps the spatialRule.getID() to an `int` counter.

web/src/main/java/com/graphhopper/spatialrules/SpatialRuleLookupBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+                spatialRules.add(spatialRule);
+                for (Polygon polygon : spatialRule.getBorders()) {
+                    polygonBounds.update(polygon.getMinLat(), polygon.getMinLon());
+                    polygonBounds.update(polygon.getMaxLat(), polygon.getMaxLon());
+                }
+            }
+        }
+
+        if (!polygonBounds.isValid()) {
+            // TODO Changed this from Exception to the Empty SpatialRuleLookup, is this ok @michaz?
```
karussell: I think it is ... remove this comment?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
```diff
@@ -79,6 +69,7 @@ public AccessValue getAccessValue(String highwayTag, TransportationMode transpor
             return AccessValue.ACCESSIBLE;
         }
 
+        // TODO: Should we change this to "SpatialRule.EMPTY"?
```
karussell: Yes, would be nice.

karussell:
One more thing: would you mind to update this to the latest master :) ? (There were a few changes&renames)

boldtrn:
> One more thing: would you mind to update this to the latest master :) ? (There were a few changes&renames)

Sure, sorry, I did that yesterday, but was interrupted :). I will review this PR one more time.

boldtrn:
Ok, I had a look at the code and also ran a couple of manual tests, including passing different BBoxes to see if the route result changes. Everything worked fine. I also added some documentation to make it easy for others to get started with SpatialRules.

Issue #993, opened 2017-03-20
Migrate map-matching LocationIndexMatch to graphhopper


kodonnell:
As discussed [here](https://github.com/graphhopper/map-matching/pull/87#discussion_r106920593)

kodonnell:
Closed with #994

PR #221, opened 2014-06-18
Issue 218: Autoexpanding search for nodes around a certain coordinate

JohannesPelzer:
Implementation of autoexpanding search, following @karussell his suggestions in https://github.com/graphhopper/graphhopper/issues/218

PR #994, opened 2017-03-20
moved LocationIndexMatch.findNClosest from map-matching to LocationIndexTree

### LINKED ISSUE #971: make LocationIndexTree.XFirstSearchCheck public
This would be required for https://github.com/graphhopper/map-matching/pull/87

kodonnell:
This is for #993 ~and [map-matching 65](https://github.com/graphhopper/map-matching/issues/65). Comments:~

- ~no real logic changed excepted made maxIterations configurable (from hardcoded value of 2 in map-matching) and throws an error if maxIterations 'truncates' the radius (as discussed in map-matching 65).~ I've added some comments for tidy-ups (that can be made later).
- ~renamed to `findWithinRadius` to better reflect the method (as opposed to the somewhat misleading `findNClosest`).~
- copied tests over from map-matching ~and added one to check the error is thrown for maxIterations truncation.~
- #971 was (I think) prompted by map-matching requirements for XFirstSearchCheck. This is no longer the case, so I'm not sure if we can close #971?
- required by [map-matching 87](https://github.com/graphhopper/map-matching/pull/87)

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
+++ b/core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
@@ -578,8 +578,7 @@ public class LocationIndexTree implements LocationIndex {
      * @param queryLat latitude to search from
      * @param queryLon longitude to search from
      * @param edgeFilter only include edges from edgeFilter
-     * @param radius include all results within this radius (in meters) - assuming it's not
-     *        bounded by maxIterations.
+     * @param radius include all results within this radius (in meters)
      * @param maxIterations if not null, search only a square of (2 * maxIterations - 1) tiles
      *        wide/high (with center tile containing (queryLat, queryLon)). Note that it is
      *        possible that the search radius exceeds this square - in this case, an exception
--- a/core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
+++ b/core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
@@ -37,6 +37,8 @@ import org.slf4j.LoggerFactory;
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 
@@ -71,6 +73,12 @@ public class LocationIndexTree implements LocationIndex {
     private double deltaLon;
     private int initSizeLeafEntries = 4;
     private boolean initialized = false;
+    private static final Comparator<QueryResult> QR_COMPARATOR = new Comparator<QueryResult>() {
+        @Override
+        public int compare(QueryResult o1, QueryResult o2) {
+            return Double.compare(o1.getQueryDistance(), o2.getQueryDistance());
+        }
+    };
     /**
      * If normed distance is smaller than this value the node or edge is 'identical' and the
      * algorithm can stop search.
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
```diff
@@ -564,6 +572,133 @@ protected boolean check(int node, double normedDist, int wayIndex, EdgeIteratorS
...
 
+    /**
+     * Returns all edges that are within the specified radius around the queried position.
+     *
+     * @param queryLat latitude to search from
+     * @param queryLon longitude to search from
+     * @param edgeFilter only include edges from edgeFilter
+     * @param radius include all results within this radius (in meters) - assuming it's not
+     *        bounded by maxIterations.
+     * @param maxIterations if not null, search only a square of (2 * maxIterations - 1) tiles
```
boldtrn: I don't get the purpose of maxIterations and radius. 

> include all results within this radius (in meters) - assuming it's not bounded by maxIterations.

How can I bound the radius with maxIterations, when maxIterations says:

> Note that it is possible that the search radius exceeds this square - in this case, an exception will be thrown.

To me this sounds contradicting. You can bound the radius with maxIterations, but maxIterations has to be at least as big as the radius.

So maxIterations can increase the radius? But why bothering, I could just increase the radius.

BTW: At this point haven't really looked at the code, I just stumbled over the JavaDoc.

**Edit:**
Ok, I am finished with reviewing the code and still don't understand maxIterations.
Is it necessary to have maxIterations? If yes why? And if yes, wouldn't it be easier to have a wrapper method that allows calling this method without maxIterations?

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
```diff
@@ -564,6 +572,133 @@ protected boolean check(int node, double normedDist, int wayIndex, EdgeIteratorS
...
+                throw new IllegalArgumentException("A minimum of " + requiredIterationsForThisRadius
+                        + " iterations are required for a radius of " + radius + "m, but a "
+                        + "maxIterations of " + maxIterations + " was provided. Behaviour is "
+                        + "undefined in this case, so please either decrease maxIterations to "
+                         + requiredIterationsForThisRadius + ", or decrease the radius.");
+            _maxIterations = maxIterations;
+        }
+        
+        final double returnAllResultsWithin = distCalc.calcNormalizedDist(radius);
+        // implement a cheap priority queue via List, sublist and Collections.sort
```
boldtrn: Why aren't we using a [PriorityQueue](https://docs.oracle.com/javase/7/docs/api/java/util/PriorityQueue.html)?
I assume there is a reason, maybe we could add the reason to the comment?

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
```diff
@@ -564,6 +572,133 @@ protected boolean check(int node, double normedDist, int wayIndex, EdgeIteratorS
...
+     *        possible that the search radius exceeds this square - in this case, an exception
+     *        will be thrown.
+     */
+    public List<QueryResult> findWithinRadius(final double queryLat, final double queryLon,
+            final EdgeFilter edgeFilter, final double radius, final Integer maxIterations) {
+
+        // check the max iterations:
+        int requiredIterationsForThisRadius = (int) Math.ceil(radius / this.minResolutionInMeter)
+                + 1; // note that this is a conservative upper bound
+        int _maxIterations;
```
boldtrn: Why do we with _? Wouldn't be a good name for the variable better?

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
```diff
@@ -564,6 +572,133 @@ protected boolean check(int node, double normedDist, int wayIndex, EdgeIteratorS
...
+                        + "maxIterations of " + maxIterations + " was provided. Behaviour is "
+                        + "undefined in this case, so please either decrease maxIterations to "
+                         + requiredIterationsForThisRadius + ", or decrease the radius.");
+            _maxIterations = maxIterations;
+        }
+        
+        final double returnAllResultsWithin = distCalc.calcNormalizedDist(radius);
+        // implement a cheap priority queue via List, sublist and Collections.sort
+        final List<QueryResult> queryResults = new ArrayList<QueryResult>();
+        GHIntHashSet set = new GHIntHashSet();
```
boldtrn: What does the set contain? Needed check the code again to see where it was passed.

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
```diff
@@ -564,6 +572,133 @@ protected boolean check(int node, double normedDist, int wayIndex, EdgeIteratorS
...
+                        @Override
+                        protected double getQueryDistance() {
+                            // do not skip search if distance is 0 or near zero (equalNormedDelta)
+                            return Double.MAX_VALUE;
+                        }
+
+                        @Override
+                        protected boolean check(int node, double normedDist, int wayIndex,
+                                EdgeIteratorState edge, QueryResult.Position pos) {
+                            if (normedDist < returnAllResultsWithin || queryResults.isEmpty()
```
boldtrn: isEmpty check to make sure that at least one result is returned? Should we remove that result later?

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
```diff
@@ -564,6 +572,133 @@ protected boolean check(int node, double normedDist, int wayIndex, EdgeIteratorS
...
+                                    QueryResult qr = queryResults.get(qrIndex);
+                                    // overwrite older queryResults which are potentially more far
+                                    // away than returnAllResultsWithin
+                                    if (qr.getQueryDistance() > returnAllResultsWithin) {
+                                        index = qrIndex;
+                                        break;
+                                    }
+
+                                    // avoid duplicate edges
+                                    if (qr.getClosestEdge().getEdge() == edge.getEdge()) {
```
boldtrn: Why aren't we using a set/map here and filter potential unwanted results after the for loop?

With for loop I mean:
`for (int iteration = 0; iteration < _maxIterations; iteration++) {`

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
kodonnell: > I don't get the purpose of maxIterations and radius.

The purpose is to find everything within `radius`. The `maxIterations` is there to bound to a certain amounts of tiles - as discussed in MapMatching 65. (Personally, I don't think it's required - as long as we always search one tile index 'greater' than the radius - but that's what was in the code - and I'd prefer to keep this a moving of the method as opposed to a rewrite. Yes, I did rewrite it somewhat so that the method actually reflected it's purpose and implementation.)

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
kodonnell: Just my convention - something like `tmpMaxIterations`?

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
kodonnell: No idea - that's how it was done in map-matching, and I was just shifting it across.

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
kodonnell: See third argument of `findNetworkEntries` - I can't remember now, but I believe it's all of the nodes found up to and including that iteration of `findNetworkEntries`.

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
kodonnell: That's how the current code is structured. I agree with you - see my [comments](https://github.com/graphhopper/graphhopper/pull/994/files#diff-bceb63de6e06385d9d74fc703931bff9R635). Maybe it should be removed, as it doesn't behave as described. @karussell - is there a reason we need to keep this? Should we make it optional?

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
kodonnell: I agree (I think) as per my comment about 'clean[ing] up the index tracking business'. However, again, I wanted to port the functionality and tests before tweaking. We can rewrite if you prefer ...

kodonnell:
As a general comment, there's possibly cause for a more thorough rewrite. I feel a lot of the machinery is aligned with the GH use cases, which are to find the nearest candidates to a point. If we want to be fully generic, then we could probably e.g. make better use of the quadtree (as opposed to always looping through leaf nodes - I think that's the term).

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
boldtrn: > The purpose is to find everything within radius. The maxIterations is there to bound to a certain amounts of tiles - as discussed in MapMatching 65.

Sorry, I still don't really get it. Shouldn't this be the default? If we actually want something like this, could we create a boolean like: "includeSurroundingTiles". This would add increase the requiredRadius by one. I think this parameter requires too much knowledge about the actual method. Since this is a public method and is intended to be used by everyone, we should keep it simple.

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
boldtrn: Not sure, as it's not really tmp. I would favor removing maxIterations from the parameter list :). But let's discuss this in the comment above.

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
boldtrn: Ok, thanks. Any idea @karussell?

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
boldtrn: Sorry, I should have written the comment differently. I would just prefer a different name, as set can contain everything.

boldtrn:
> As a general comment, there's possibly cause for a more thorough rewrite.

Yes maybe. Sorry for being so petty about the implementation, as this PR is only about moving the method :).

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
karussell: We have to do that even for the normal version but this parameter allows for even longer searches. See https://github.com/graphhopper/graphhopper/pull/221

I would also prefer to avoid this extra parameter in the method - isn't this also more a static configuration like 'maximum allowed visited nodes' and not per request?

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
karussell: Good question. Lets address in a different issue or via discuss :)

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
karussell: The idea was to have at least one candidate even if outside, so maybe this is not necessary, yes. 

Here again let us try to minimize this issue and create improvement issues separately to make merging this faster.

karussell:
Thanks a lot @kodonnell !

To make merging this a no-brainer would you mind to remove the maxIterations parameter? We can then discuss the improvements or rewrite in a separate issue or in a dev discuss topic

> make better use of the quadtree (as opposed to always looping through leaf nodes

What do you mean here?

kodonnell:
> To make merging this a no-brainer

Done - and to remove the brain completely, I've just reverted to the original map-matching code - I didn't change the name as it'd be wrong too, and it probably makes sense to clarify the purpose before editing it further.

> What do you mean here?

At the moment we're manually iterating through every leaf node (I think). If we've got a radius of 8km, with leaf node size of 1km, then (heuristically) we'd first find the four covering 8km-radius nodes, and then drill down through each using the quadtree until we get to the leaf nodes that contain stuff (which might be e.g. the 4km one in some cases, the 2km in others, etc.). It's a more natural use of a quadtree, and could have better performance in the case where a lot of the search radius is empty (i.e. we know the parent is empty, so no need to check the children). That'd also be especially true if we're traversing the quadtree every time we seach a leaf node. That all said, I've got no idea how the GH quadtree is implemented, so the above might all be baloney = )

core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
kodonnell: I dont know = ) I think I may have confused it with how I interpreted `maxIterations` originally - it's not "iterate through no more than this many tiles" it's "iterate through this many tiles". On second thought, I now think it should really just be making sure that we always include the extra surrounding tiles - which should have nothing to do with the user.

karussell:
> That'd also be especially true if we're traversing the quadtree every time we seach a leaf node.

Aha, makes sense. Still the quadtree implementation is very fast so even if we do this a couple of times without success I guess the improvements will be not really comparable to the routing time.     

> No worries - though I feel like me doing PR's actually takes you guys more time that if you did it yourselves! Hopefully I'll get better

no no :). I'm/we're thankfully for every work being done from others, as the repository activity is nicely/overwhelming increasing :)

kodonnell:
Thanks @karussell = ) Opened #997 for the issues discussed above.

PR #995, opened 2017-03-21
Added ExecutorService for Landmarks

boldtrn:
Similar to #980 I added the Executor Service for landmarks. I also ran a brief Performance test:

Runs                    | Start LM         | End LM         | Duration LM | Took (shortest, fastest)      |
----------------------- | ------------------ | ------------------ | ----------------- | ---------- |
LM Wrapped Runnable | 11:07:14,586 | 11:14:37,499 | 7:23 min | 379,422 |
LM Runnable | 11:29:50,883 | 11:37:39,645 | 7:49 min | 377, 468 |

PR #1000, opened 2017-03-24
Bug when loading prepared graph with LM

boldtrn:
Important Bugfix, when loading an already prepared graph with LM, the server blocked at: completionService.take(), as no thread is created if LM is already prepared.

I introduced this bug with #995. The problem is that when loading a graph with prepared Landmarks that we don't spawn a thread if they are already prepared. With CH we create a new thread.

A test for this scenario would be nice as well, but I wasn't sure where to add something like this?

BTW: This is #1000 🎉🎉🎉


karussell:
Thanks for this and 🎉 ;)

Also the fix looks simple also we could put the load method inside the thread to avoid this difference? (Because at some point I would like to merge this with CH to avoid code dup :))

boldtrn:
> also we could put the load method inside the thread to avoid this difference? 

Not sure right now, what you mean, but sure, go ahead. I'd vote to merge this ASAP, as it breaks the current master, and it breaks very ugly, took me some time debugging, but I won't be able to do this today. So maybe we do this in a future PR or feel free to add your changes here :)?

PR #1001, opened 2017-03-24
Improve Instructions: removed a weak if

boldtrn:
There is an issue with the ORs. We should only accept if both times either the left or the right side are true. Right now, we accept combinations like left, right or right, left.

Haven't found an issue with this yet, but it might happen.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/Path.java
+++ b/core/src/main/java/com/graphhopper/routing/Path.java
@@ -654,8 +654,8 @@ public class Path {
                 return new GHPoint(tmpLat, tmpLon);
             }
 
-            private int returnForcedInstructionOrIgnore(boolean forceInstruction, int sign){
-                if(forceInstruction)
+            private int returnForcedInstructionOrIgnore(boolean forceInstruction, int sign) {
+                if (forceInstruction)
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/graphhopper/routing/Path.java
```diff
@@ -676,23 +676,30 @@ private boolean isLeavingCurrentStreet(long flag, long prevFlag, int baseNode, i
...
                 return false;
             }
 
+            private boolean isTheSameStreet(String name1, long flags1, String name2, long flags2, boolean checkFlag){
+                if (isNameSimilar(name1, name2)) {
+                    if (checkFlag) {
+                        if (flags1 == flags2)
+                            return true;
+                    } else {
+                        return true;
```
karussell: can we simplify this to e.g. if !checkFlags || flags1 == flags2 then return true?

boldtrn: Thanks, yes this make the code a lot better. Now we could even merge the new if with the outer if, but I think this is ok for readability.

devemux86:
Is something else needed here?

PR #1004, opened 2017-03-24
Update routing.md

kugaudo:
The sample requires graphhopper-reader-osm but no such info given anywhere.


karussell:
Thanks!

I'll add a link to this here: https://github.com/graphhopper/graphhopper#maven as otherwise we have to update versions everywhere in the docs. Also I'll add information about the difference when core and when reader-osm

PR #1009, opened 2017-03-25
GTFS from Switzerland requires larger validityids

karussell:
Import of the GTFS for Switzerland works but a few problems occurred:
 
 * https://github.com/conveyal/gtfs-lib/issues/40
 * And the max validityId was too small (What is its purpose?)
 * (query takes long but this is already known)

### DIFF
--- a/reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/PtFlagEncoder.java
+++ b/reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/PtFlagEncoder.java
@@ -40,7 +40,7 @@ class PtFlagEncoder extends AbstractFlagEncoder {
 		shift += time.getBits();
 		transfers = new EncodedValue("transfers", shift, 1, 1.0, 0, 1);
 		shift += transfers.getBits();
-		validityId = new EncodedValue("validityId", shift, 11, 1.0, 0, 2047);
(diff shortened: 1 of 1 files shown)


michaz:
Thanks!

> And the max validityId was too small (What is its purpose?)

Oh. Consider the time range of the schedule, from day 0 to day N.

We encode the days on which a trip "is valid" == takes place as a bit set:

00000110000011... goes only on weekends if day 0 is a Monday.

Each *different* bit set which occurs in a schedule is supposed to get a unique id, a pointer, so that we don't need to save the entire thing in the edge.

This seems to mean there's a lot of different validity patterns. :-/

michaz:
> (query takes long but this is already known)

Yes, this bothers me too. But I'd like to be more feature-complete and more sure about some things before we start optimizing, as long as that seems ok.

karussell:
> Let me know if you strongly disagree. There are some ideas.

No no, I agree here. Just wanted to write it down as the times are significantly slower compared to VBB

karussell:
> Each distinct bit set which occurs in a schedule is supposed to get a unique id, a pointer, so that we don't need to save the entire thing in the edge.

Ah, cool idea. I think I tried a few higher counters like 13 and 14 and 16 but then increased to 20 due to laziness and only a few more bits needed. Maybe we can decrease it later if we find no other data set needs more.

PR #1011, opened 2017-03-26
Create hook for importing OSM elements from something else than a File.

michaz:
On OSM-Saturday at FOSSGIS, I sketched a (GraphHopper-based) routing-plugin for JOSM. Not worth checking out yet, but directly reading the JOSM data model from memory and finding a route from the GUI works like a charm -- in about 120 lines of code.

This is the minimal set of changes I had to make to plug into GraphHopper's OSM-import to have it read the data structures I feed it from JOSM, instead of a file.

(Could be followed up by a refactoring where the PBF and the XML file readers are separated.)

### DIFF
--- a/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMInput.java
+++ b/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMInput.java
@@ -1,0 +1,28 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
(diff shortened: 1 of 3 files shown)


karussell:
Cool :+1: 

Would be really interested in this. Maybe you still release this plugin in a separate repo or on your personal repo? I'm doing so for similar stuff which I see more as sketches :)

michaz:
> Maybe you still release this plugin in a separate repo or on your personal repo?

Definitely.

boldtrn:
Nice, looks pretty elegant :)

Issue #1013, opened 2017-03-27
Instructions should ignore VirtualEdges created during lookup


boldtrn:
When during the lookup phase of a route request a new virtual edge is generated for StartPoint -> FirstRoute, we have to ignore these VirtualEdges. They lead to a wrong number of turns or might be interpreted as turns, which shouldn't be the case. Here is one example case, the requested point is place very close to Edge2:

```
  Edge3      NewVirtualEdge
        \   /
Edge 1   \ /         Edge 2
__________  ______________
```

When routing from NewVirtualEdge via Edge two, the Instructions are Continue -> Turn Sharp Left.

I think we should ignore the newly created VirtualEdges/VirtualNodes in this case?

karussell:
Not sure if this has something to do with this but do you return a weight smaller than 1e-3 in Weighting.calcWeight? Currently we do not check weight boundaries for CH but we could to do so e.g. in CHGraphImpl.setWeight also there you see the WEIGHT_FACTOR=1000

karussell:
Can we close here and is the real underlying reason same as #1022?

boldtrn:
I think so, yes.

karussell:
Done

PR #1014, opened 2017-03-27
Finding the prevEdge from the prevNode

### LINKED ISSUE #1010: Instructions Exception "Couldn't find the edges for..."
We throw an exception when we cannot find the other edges. It happens for some routes that these cannot be found.

We should make the code more robust and maybe just return the calculated sign?

boldtrn:
This issues fixes #1010. The old behavior was created when the algorithm was searching for the current edge and the previous edge. I pass the current edge in the method parameter now, so the current edge is not needed anymore. It makes more sense to search from the previous node to the base node instead of searching from the base node to the previous node in reverse.

I have seen some issues here with VirtualNodes that couldn't be found from the base node correctly. This only happened in combination with a custom weighting and CH, so it is hard to write a test here.


karussell:
> I have seen some issues here with VirtualNodes that couldn't be found from the base node correctly. This only happened in combination with a custom weighting and CH, so it is hard to write a test here.

Ok, let us fix this other issue first, I think you mean #1013 ?

karussell:
Ok, this has not much to do with #1013. Would you mind to try if you can find this bug in a (integration) test against a CH graph? I can guess this should easily occur if you do

./graphhopper.sh measurement core/files/some.osm.gz

karussell:
Strange, I cannot find such an issue although I thought this could happen frequently due to the graph.disconnect call in PrepareContractionHierarchies

karussell:
I thought I had reproduced the issue but this fix doesn't fix the problem that I had, see [this commit](https://github.com/graphhopper/graphhopper/tree/issue_1014).

See the discussion where [this is coming from](https://discuss.graphhopper.com/t/improving-foot-bike-profile-to-show-instructions-when-turning-if-streetname-doesnt-change/1735/8).

karussell:
The reason that this still fails sometimes for via points is tricky. If we have via points we create 3 virtual nodes e.g. A->B and B->C now if for some reason A is on the path B->C the QueryGraph correctly contains the two virtual edges with A leading to a path like B->...->x1->A->x2->...->C

BUT when using the CH graph the query B->C returns a path with some shortcuts that were extracted but using only real edge like `x1->x2` and not the QueryGraph. Now the turn instruction code relies on finding the previous nodes and fails to find A at node x2.

boldtrn:
Thanks a lot for debugging this and finding a test case @karussell.

I suspected the virtual nodes to be the source of the issue as well. I actually wrote some code to [ignore virtual nodes a couple of weeks ago](https://github.com/boldtrn/graphhopper/commit/aca65dcee5f5129017e2a65b1c8bb81faf13ab16) (I think I mentioned it offline). The code does not fix the issue, that why I haven't created a PR. 

But, I just had an idea. And I have to blame myself for writing unnecessarily complicated code here. I cannot so any reason to not store the prevEdge like we do with prevName. I just implemented this ([see my latest commit to your branch issue_1014](https://github.com/graphhopper/graphhopper/commit/aad04aab303c082f082fc91ab88198111a4f88bd)). Currently I simply force the turn sign if the prevEdge is null (which should not happen frequently, if it happens at all).

WDYT?

karussell:
So this means we can avoid the limitation of the CH path extraction (returning real edges)? Nice!

karussell:
Would you mind to update the PR to the latest HEAD?

boldtrn:
> So this means we can avoid the limitation of the CH path extraction (returning real edges)? Nice!

Yes :). Somehow I must have missed that before...

> Would you mind to update the PR to the latest HEAD?

I just updated the PR to the latest master.

PR #1017, opened 2017-03-30
add gtfs reader to script

karussell:


michaz:
Okay.. if this works. :)

I think Maven should know how to build everything and in what order, and if it doesn't, we should tell it. Putting build logic into shell scripts **around** Maven seems wrong.

Is there something in particular which is in the way of that, or is it just something that someone would need to do?

karussell:
> Is there something in particular which is in the way of that, or is it just something that someone would need to do?

Yes, more this :)

And also the problem is that if someone just wants to import OSM she does not need to build shape+web+gtfs module (so reducing recompile times). And also the android module is separated and only triggered via special profile.

michaz:
Okay. And the new ``reader-gtfs`` is necessary here because... later.. somewhere.. the web module is ``assembly:single``d and then its dependencies must already have been ``install``ed, right?

Issue #956, opened 2017-02-20
Use JTS for polygon and geometry related work


karussell:
We should not reinvent the wheel and also JTS will be soon under the EPL or EDL: https://github.com/locationtech/jts

devemux86:
Like a core dependency?

karussell:
Yes, highly likely we need this at some point in the core. Not sure if it is avoidable and/or necessary to avoid it.

boldtrn:
+1 for not reinventing the wheel. I think the lib is already under dual licensing EDL/EPL. Not an expert on the license details. Is this compatible with Apache2 without restricting use cases?

karussell:
> Is this compatible with Apache2 without restricting use cases?

As the one is like MIT it should be compatible, but have not read the details.

karussell:
#1039

Issue #903, opened 2016-12-07
remove unused data from OSM files in repo


kodonnell:
As discussed [here](https://github.com/graphhopper/map-matching/issues/85#issuecomment-265291288) (and following comments) there may be an interest in filtering OSM files to only include relevant data, before adding to the repo. (I'm assuming this is more relevant as an issue here than under map-matching, though I assume it applies to both.)

I definitely recommend it for new OSM files - but filtering existing files will (I think) make the repo bigger, and probably not gain us much?

Objectives for this issue:
- [ ] figure out (and document) how to do the appropriate filtering
- [ ] (optionally) apply it to existing OSM files

Cc @karussell @devemux86

ratrun:
Can you please describe a bit more what you mean by "repo"? Any why would somebody want to integrate such a functionality to graphhopper core? What is the goal (e.g. reduction of graph size,..)? From my perspective you are describing postprocessing steps one might to do and document for specific purposes. But shouldn't such steps be better kept out of scope for the graphhopper project?

kodonnell:
Sorry @ratrun, I should've been clearer - we're not trying to add this functionality to the core (i.e. no code changes), it's just housekeeping. That is, why have full OSM files sitting in the (git) repo, when GH only needs road info, etc.? See the discussion linked above - it'd be nice to be able to add small OSM exports in as unit tests, but currently they're not small as they're full of extra junk. @karussell made the comment that it might be nice to apply this to existing OSM files in graphhoppper/map-matching.

ratrun:
Thanks @kodonnell, now I understand. 
I agree that e.g. buildings could be filtered away, but I would not filter out any ways which potentially could be used. I would not change already checked in files as their pre-decessors would still remain in the git repro. So this idea should be only applied to new files.

kodonnell:
>I agree that e.g. buildings could be filtered away, but I would not filter out any ways which potentially could be used.

Agree. For context, the aforelinked OSM download was of only a few dozen roads, but was 6mb - in large part because of version/author information, but even things like tree species!

> I would not change already checked in files as their pre-decessors would still remain in the git repro. So this idea should be only applied to new files.

There's probably a way to overwrite the history, but that scares me (and will probably cause problems) so yes, I agree.

Nakaner:
I am adding this to this issue because it has the same reason.

[...]

PR #1018, opened 2017-04-04
Public Transit

michaz:
Provides GTFS-based public transit routing.
Please don't squash -- changes to the rest of GraphHopper are in topic-based commits.


michaz:
(I'm figuring out how to batch-insert the copyright header, never mind that.) :-)

core/pom.xml
```diff
@@ -76,11 +76,22 @@
...
             <groupId>org.json</groupId>
             <artifactId>json</artifactId>
             <version>${json.org.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.vividsolutions</groupId>
+            <artifactId>jts</artifactId>
+            <version>1.13</version>
+        </dependency>
```
devemux86: Could use [com.vividsolutions:jts-core:1.14.0](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22com.vividsolutions%22)?

reader-gtfs/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+                <exclusion> 
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-simple</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>13.0.1</version>
```
devemux86: Why not [21.0](http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.google.guava%22%20AND%20a%3A%22guava%22)?

michaz:
@devemux86 Ahh, yes, thanks, very helpful. Let me check.

reader-gtfs/pom.xml
michaz: Indeed, why Guava at all.

devemux86: BTW Guava 21.0 would [require](https://github.com/google/guava/wiki/ReleaseHistory) JDK 1.8?
Don't know if could be an issue with current Android.

michaz: I don't need it anyway. It was just an accidental left-over. Thank you, the dependency police. :-)

core/pom.xml
michaz: Will check, because I need harmonized versions between this place and the GeoJSON serializer. But that one has a newer version, too, so...

BTW are you fine, Android-wise, with jts-core in the GraphHopper core? We were worrying..

devemux86: Already using it with Android, e.g. GraphHopper Android [sample](https://github.com/graphhopper/graphhopper/blob/master/android/app/build.gradle#L52).

reader-gtfs/pom.xml
michaz: On the other hand, I'm using this Conveyal GTFS library, and that one is Java 8 anyway. And my own GTFS import code, too. Eventually, we can pull the non-import-related-GTFS-stuff into another module and make that Java 7 again. I'm already almost prepared for that.

karussell:
Thanks @michaz ! Very nice step by step PR :+1: !

reader-gtfs/pom.xml
karussell: I think the problem is more iOS. Maybe @clns could tell if e.g. jts-core could make a problem or how we could handle this.

clns: Guava looks like its supported by j2objc, it even exists a special Makefile from google that translates it: https://github.com/google/j2objc/tree/master/guava

I think it will require some changes in the iOS port but it's safe to add.

core/pom.xml
michaz: Nope, can't do that at the moment, because com.bedatadriven:jackson-datatype-jts:2.4 uses com.vividsolutions:jts:1.13 and I don't want to mess around with manual exclusion etc.

Can we put this on the to-be-resolved-before-next-release list?

michaz: @clns I think @karussell's comment should have been here. Any strong opinion on JTS?

clns: There are no references for JTS and j2objc, I don't think people tried to translate it or at least they didn't get into troubles (hopefully this is true). I really don't know if it'll work.

karussell:
## Quick start

Just for otheres, [here](https://github.com/graphhopper/graphhopper/tree/8ed20f88e56110cbf4a447423dce204409fef4d9/reader-gtfs#quick-start) is a link to the quick start of public transit routing with GraphHopper.

## Comments

I'll avoid the github review system as this makes my comments too unsorted and a bit messy at the end I think.

I really like the high quality of this PR. How did you managed to get this clean history - via some git magic or copy paste the necessary changes + commit? 

Here are my minor comments as I've not yet fully consumed the 'real work' :)

**1. The performance of the core is unchanged**

Nice :) !

**2. A new JTS dependency is introduced, relates to #956**

Here we need to check our [iOS port](https://github.com/graphhopper/graphhopper-ios). BTW: do you have an xcode setup?

**3. A new jackson dependency is introduced, relates to #773**

Here I'm not sure if we should do this now or in a separate PR as now we have a mixture of json.org, jackson and bson. We should either avoid jackson for now or force jackson for everything immediately after this PR (i.e. fixing #773).

**4. CSV Parser**

I can remember we did this via jackson (in #855) and then I saw another csv reader `com.csvreader.CsvReader`. But now I cannot find the dependency anywhere. How does this work :) ?

**5. GTFS data**

 * The [attribution](https://github.com/graphhopper/graphhopper/pull/1018/commits/321ef1b3cd97090da09cdeecd291997a49f4cae3#diff-fdc4ed1244666190ba851259215a7c11R40) should go into the NOTICE.md file
 * are the new zip files necessary or can we somehow get rid of them or reduce them (I mean this in regard to #903). Or did you already try to optimize this? Why bart.proto and bart.zip?

**6. gtfs-lib**

Can we contribute our changes back to the official repo instead, somehow? Or should we release our own version? I do not like to depend on github and jitpack for a production build additionally to maven central. (Of course jitpack is very convenient for development. Maybe we could solve this via using our internal repository manager that we have control over, not sure.)

**7. DefaultModule is refactored to use annotations**

I'm referring here to @Provides @Singleton etc. Why was this change necessary? I do not like annotations that much (except maybe for serialization). How would I disable a certain binding? E.g. we reuse this setup in the MapMatching and at other places and need to customize this. Also doing this via code looks cleaner and less lines.

E.g. for PtModule we could extend from DefaultModule and overwrite some createGraphHopper method.

**8. javadocs**

We need more javadocs #891 :) to describe the intend of public classes and methods

**9. documentation**

We need some rough documentation of how things work and maybe enhance the existing sketch or create a new one where the classes (Leg, Trip, GraphExplorer, GtfsReader, GtfsStorage, MapDB, NodeType (already in the image), PtTravelTimeWeighting) are mentioned and how they work together, maybe just adding a simple route request in the low-level-api.md with the response and how to find out the line change and 

**10. misc**

> in GraphHopperGtfs class under public static GtfsStorage createGtfsStorage

Why are these helper methods necessary? The underlying call isn't a lot longer but more powerful. And this would be an API we would keep and maintain.

> GraphHopperGtfs.route

Not really related to this PR: but do you have an idea how we can make this method shorter :) ? (of course, this applies to GraphHopper.route too)

> in GraphHopperGtfs class under private List<Trip.Leg> legs

What is this method doing? Kind of instructions calculation for a path?

> in GraphHopperGtfs class -> EARLIEST_DEPARTURE_TIME_HINT = "earliestDepartureTime";

Here we should use the same 'dot and underscore' notation like `pt.earliest_departure_time` and move this to the Parameters class? And for 'walkSpeedKmH' I would remove the 'unit' as we do not specify the unit in other parameters only in the docs.

In general we should prefer underscore for serialized properties too (like in GtfsStorage), although Java coding standard is different but this avoids confusion when we search for something in the storage, properties in the code and properties in the API or URL parameters.

Also in the Weighting currently it is 'pttraveltime' and I would prefer 'pt.travel_time', see 'short_fastest'. Think '.' for a subcategory and '_' if there are multiple words.

Now that we are at naming should we use GTFSXy or keep GtfsXy? If the latter then we should do this for OSMReader etc too

> // graphHopper.close()

Why is the close method not called [here](https://github.com/graphhopper/graphhopper/pull/1018/commits/321ef1b3cd97090da09cdeecd291997a49f4cae3#diff-f2fdf7bb4368e18aeb1fc07502e6386dR38)?

> GtfsHelper.localDateTimeFromDate

There should be something simpler without the intermediate step through SimpleDateFormat

> graph.getEncodingManager().getEncoder("pt")

Should we use an internal String PT = "pt" to allow simpler usage searches?

> GtfsStorage.traveltime

Rename to travelTime and move to GtfsHelper?

> GtfsRealtimeTest.java 

Is real time already supported :) or just the reader? (Maybe here camelCase for GtfsRealTime?)

## Important follow up issues:

 * merge the changes to the UI of PT too
 * If Jackson is introduced here, we need to fix #773
 * Make iOS port running
 * add public transport routing option to measurement to catch regressions early on

## Follow up issues:

 * make public transit routing (not the import!) possible on mobile too. I.e. either remove the MapDB dependency somehow from the core of public transit routing or as a first step try to make it working on Android (maybe it already works even with jdk8 and MapDB)
 * do we need / want a debugging tool for the PT graph? How would it look like - similar to the MiniGraphUI?
 * put changeGraph on GraphHopperAPI interface and remove cast (or some other solution)
 * GTFSXy vs. GtfsXy

karussell:
Please see my recent commits to make the UI working for pt branch but make this optional and with translations.

michaz:
Thanks! I'll start with something important which I'm confident about (I did a lot with Guice on MATSim):

> 7. DefaultModule is refactored to use annotations

The annotations are not the important change here. There is indeed an equivalent way of writing this without annotations, but it is not shorter but longer, using explicit ``Provider<>`` classes).

The important change is that ``createGraphHopper()`` is not called from the ``configure()`` method. The ``configure()`` method must always be light-weight. It may only contain configuration, in a declarative style (like a config file). The only exception is when instances of something were created _before_ starting Guice. Practical reason: In more complicated setups, the ``configure()`` method is called _multiple times_ because Guice thinks it can parse it as often as it wants. In that case, we would produce multiple GraphHoppers.

I can show the annotation-free version of this, but I'm sure you would agree that it's actually uglier. :)

> How would I disable a certain binding? E.g. we reuse this setup in the MapMatching and at other  places and need to customize this.

You can't disable bindings, but with ``Modules.override()``, you can change them, because..

> E.g. for PtModule we could extend from DefaultModule and overwrite some createGraphHopper method.

..Guice has its own concept of overriding bindings from other modules. Modules aren't really meant to be used like this with (additional) Java subclassing.

I would adapt any dependent code to this.

michaz:
> I really like the high quality of this PR.

Thank you! :)

> How did you managed to get this clean history - via some git magic or copy paste the necessary changes + commit?

I copy/pasted the changes to the core individually (they weren't that many, in the end). But *after* that, I juggled a lot with rebase -i to make each commit as I wanted it.

karussell:
> The important change is that createGraphHopper() is not called from the configure() method
> ...
> I can show the annotation-free version of this, but I'm sure you would agree that it's actually uglier. :)

Aha, didn't notice this change - makes sense then!

> I would adapt any dependent code to this.

Ok, will get in touch with you then for this :)

michaz:
> make public transit routing (not the import!) possible on mobile too. I.e. either remove the MapDB dependency somehow from the core of public transit routing or as a first step try to make it working on Android (maybe it already works even with jdk8 and MapDB)

Yes! I want offline-GraphHopper with a Transportr frontend on Android.

devemux86:
It appears that MapDB depends on Java 8 and cannot be used right now on Android, see https://github.com/jankotek/mapdb/issues/768.

michaz:
> It appears that MapDB depends on Java 8 and cannot be used right now on Android

OK, so that's the next thing once we are feature-complete (soon).

michaz:
> 2. A new JTS dependency is introduced, relates to #956
> 3. A new jackson dependency is introduced, relates to #773

Those two are co-dependent. If we decide to adopt the JTS primitives (Point, Polygon, Coordinate ..) as _our_ datatypes, _then_ we should use a JSON library for which there already exists an implementation for serialising those to GeoJSON.

devemux86:
@michaz may I ask why Jackson and not Gson?

michaz:
> @michaz may I ask why Jackson and not Gson?

@devemux86 This is not decided, we can go either way. This is just what's there at the moment.

michaz:
https://github.com/filosganga/geogson

Ah, Ok.

michaz:
> 4. CSV Parser
> I can remember we did this via jackson (in #855) and then I saw another csv reader com.csvreader.CsvReader. But now I cannot find the dependency anywhere. How does this work :) ?

It doesn't. :) There is no CSV parsing in this PR. Alexey's data-driven tests are not in here (yet).

michaz:
> Can we contribute our changes back to the official repo instead, somehow?

Yes. Plan is: Look at the changes again, do a PR if we are still convinced of them, wait, and if there is no release with the changes in them when _we_ want to do a release, we program around it. But only depend on a release version from a release version.

michaz:
> 3. A new jackson dependency is introduced, relates to #773 

> Here I'm not sure if we should do this now or in a separate PR

Me neither. @devemux86 thinks we should use Gson, and indeed the JTS integration seems to be newer there. We could switch to JTS 14, which is good because they switched to a modular packaging there and we would only need the ``core``.

michaz:
> 5. GTFS data
> are the new zip files necessary or can we somehow get rid of them or reduce them (I mean this in regard to #903). Or did you already try to optimize this? Why bart.proto and bart.zip?

It's a real GTFS file with a real GTFS-Realtime file. But you're right, right now I'm inclined to remove _all_ of this and put data-driven tests elsewhere.

karussell:
> mapdb

 For Android I think Google will move forward with proper Java 8 support soon, so this should be okay.
It is a bit sad that they switched to Kotlin. Otherwise if it would be Java a j2objc conversion would be at least something in the possible range.

> thinks we should use Gson, and indeed the JTS integration seems to be newer there. We could switch to JTS 14

I would be okay with both strategies. We are using already jackson internally for other stuff and it is more mainstream I think (but also fatter, more deps and larger, but not 100% sure). On the other side we already use gson for the json parser in the json module.

> Plan is: Look at the changes again, do a PR if we are still convinced of them..

:+1: 

> There is no CSV parsing in this PR

I mean the usage in the tests: https://github.com/graphhopper/graphhopper/pull/1018/files#diff-0662791334af5c9c400dc64430fe506cR7

> But you're right, right now I'm inclined to remove all of this and put data-driven tests elsewhere.

I would be against a separate repo just to save a few MB download, because convenience and we always run 'real' tests too, when we do a change. Of course, I'm okay with a separate maven module (like it is in the `pt` branch)

michaz:
>>in GraphHopperGtfs class under public static GtfsStorage createGtfsStorage

>Why are these helper methods necessary? The underlying call isn't a lot longer but more powerful. And this would be an API we would keep and maintain.

This is a bit tricky. Have a look at ``PtModule``. That's a shopping list of what you need to assemble everything you need, piece by piece. All the ``@Provide`` methods just delegate to a constructor or a factory method, and the helper methods you mentioned are those factory methods.

The Guice recommendation is to make these constructors and factory methods non-public. That is a good point, because then you can change the dependencies without changing the client code. For example you can chop something into smaller pieces and then pass these pieces by constructor.

Put if you do that, you can _only_ construct the thing with Guice. And we don't want Guice in the core. That's where these factory methods come from. I want them because I want object-creation to be single function calls or constructor calls.

It's evil that they are fast-changing public methods, but because ``web`` needs to call them, they need to be ``public``, and because they are object-construction, they must be refactored often.

Some people use an annotation against this, like ``@NonPublic`` or ``@NotAPI`` or something. Don't know.

michaz:
> Some people use an annotation against this, like @NonPublic or @NotAPI or something. Don't know.

Oh, there is another alternative: Make them non-public and use javax.inject.* annotations. Then the ``PtModule`` can slurp them in even though they are non-public. :)

michaz:
>> There is no CSV parsing in this PR

> I mean the usage in the tests: https://github.com/graphhopper/graphhopper/pull/1018/files#diff-0662791334af5c9c400dc64430fe506cR7

Oh. I'm reaching through into gtfs-lib's own dependency there. To have inline-gtfs in code. Because gtfs is CSV. ,-)

michaz:
What is this AppVeyor-thing's problem? I bet something windows-specific, like the GTFS-ZIP behaving differently when unzipped, or CR/LF or some crap like that. It's 2017 and we're still not over that. :-/

karussell:
AppVeyor runs the build on a windows machine. The errors says
```
java.lang.IllegalStateException: Cannot create location index of empty graph!
	at com.graphhopper.RealtimeIT.init(RealtimeIT.java:34)
```

So yes, it could be that it is some LF problem and the import does nothing e.g. because all is on 'one line' or something resulting in an empty graph.

karussell:
Or another simpler reason could be the -Xmx1g setting which clashes with our 2g setting? 

Or another reason: Currently if there is an import failure in the `pt` code then GH does not exit and after restart (as graph cache folder is not clean) it also says: IllegalStateException: Cannot create location index of empty graph! but at com.graphhopper.http.GraphHopperGtfsModule.createLocationIndex(GraphHopperGtfsModule.java:46)

karussell:
Have created #1028, maybe discuss and fix there and see if this fixes the AppVeyors problem

michaz:
Yes, thanks.

michaz:
> We need more javadocs #891 :) to describe the intend of public classes and methods

Hmmm, okay. As opposed to the rest of GraphHopper, there are not that many public classes in this PR. ,-)

> Leg, Trip, GtfsStorage

Okay, will do.

> GraphExplorer, GtfsReader,, MapDB, NodeType (already in the image), PtTravelTimeWeighting

All of these are non-public.

michaz:
>> in GraphHopperGtfs class under private List<Trip.Leg> legs
> What is this method doing? Kind of instructions calculation for a path? 

Yes:

``// We are parsing a string of edges into a hierarchical trip.``

A leg is a segment of a trip.

karussell:
> GraphExplorer, GtfsReader, MapDB, NodeType (already in the image), PtTravelTimeWeighting
> All of these are non-public. Are we sure we want to document private classes and concepts just now?

Ok, then maybe just one question remains for me: how would I do the import and query for PT 'by hands'? (without the GraphHopperGtfs class) I.e. what would be the [low level api](https://github.com/graphhopper/graphhopper/blob/master/docs/core/low-level-api.md#create-and-save-the-graph) approach here?

michaz:
> what would be the low level api approach here?

There is none, because that would make the graph structure public, which I can't do (yet). The diagram is just for informative purposes. There is no public API for doing your own graph traversal, except "changing the code".

michaz:
> what would be the low level api approach here?

The next few possible lower-level API features I can imagine would be:

- Define your own criteria based on edge attributes. Would look a bit different to the rest of GraphHopper because of the multi-criteria approach.

- Explicitly find station nodes, so you can route strictly from station to station.

While of course keeping in mind that we want closer integration with the rest of GraphHopper.

michaz:
The general reason why we need to be more careful here is perhaps this: There is a natural mapping from OSM highways to graph edges, so routing on the graph _is_ routing on OSM, so we can expose routing on the graph. There is no such natural mapping from GTFS trips to graph edges.

michaz:
...and of course: #1023 is very reasonable. ,-) I'll have to think of something there, like hiding the pt-subgraph from the rest of GraphHopper, to prevent errors like accidentally walking on pt edges.

That would be a good idea _anyway_, to have it explicit somewhere that these are distinct subgraphs. For example, I heard it's quite possible to do something like CH on the non-pt-part of the graph.

karussell:
> There is no such natural mapping from GTFS trips to graph edges.

Thanks again for explaing this - this makes sense :+1: !

> in the sense of: afterwards we would have to be more reluctant to change them

I wouldn't worry that much for now - we can stabilize later, i.e. for >= 1.x :) and even then we can brake the API and just upgrade the version to 2.x

> because flag-encoders have to say explicitly that they are interested in an edge, and otherwise it's blocked for them

Yes

boldtrn:
Wow! Haven't had the chance to run the code on my local machine or review it in more detail, but from what I have seen this, this looks very good. Is there a timeline when we want to merge this into the master? 

BTW: Thanks for refactoring the Path class :)

michaz:
>> GraphHopperGtfs.route

> Not really related to this PR: but do you have an idea how we can make this method shorter :)

I _think_ it's possible by using throw-away (request-scoped) private classes for answering requests. I tried something in the last commit.

michaz:
> Here we should use the same 'dot and underscore' notation like pt.earliest_departure_time and move this to the Parameters class?

Okay, done, but now you need to change the JavaScript client again, please. ,-)

michaz:
> Should we use an internal String PT = "pt" to allow simpler usage searches?

Even better, I could remove it completely.

My problem with string constants is that they are

a) not enforceable, so you have to expect other occurrences anyway, so it only gives you false security ;-)
b) makes you _want_ to use it in several places, when the best design would be one where it has to be in only one place, or better, not at all.

So for me, random strings in the code are a reminder that something still needs to be cleaned up, (like in this case - thanks!) and replacing them with constants is like giving up. ,-)

michaz:
>> GtfsStorage.traveltime

> Rename to travelTime and move to GtfsHelper?

Or even better, where they really want to be. Done, thanks.

michaz:
> GtfsRealtimeTest.java
> Is real time already supported :) or just the reader? 

Yes, by now it is. :) 

> (Maybe here camelCase for GtfsRealTime?)

Hmm, dont know..

https://developers.google.com/transit/gtfs-realtime/

michaz:
> There should be something simpler without the intermediate step through SimpleDateFormat

This is going away with #1025.

michaz:
>> // graphHopper.close()

> Why is the close method not called here?

Ah, because it doesn't exist. I don't need to be closed. Thanks -- removed.

michaz:
By the way, my ``GtfsStorage`` really _needs_ to be closed because ``GTFSFeed`` doesn't expose a ``flush``, and closing and reloading it seems wasteful.

Can we add a shutdown hook to ``GHServer.main`` so that ``server.stop()`` and so ``servlet.destroy()`` is called when I quit the JVM, and we then do explicit resource management also on the server?

reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/GtfsStorage.java
```diff
@@ -17,6 +17,11 @@
 
 public class GtfsStorage implements GraphExtension {
 
+	public GtfsStorage() {
+		Runtime.getRuntime().addShutdownHook(new Thread(this::close));
```
karussell: Maybe on the web-side we should call graphhopper.close via shutdownhook for all cases not just `pt`? Or do you have problems with missing close call on non-web cases?

karussell:
> Can we add a shutdown hook to GHServer.main so that server.stop() and in consequence servlet.destroy() are called when I quit the JVM, and we then also close() our stuff on the server?

Yes, should be safe

karussell:
> Okay, done, but now you need to change the JavaScript client again, please. ,-)

should be done now

If the required `pt.earliest_departure_time` parameter is not there or has an illegal format we should throw an IllegalArgumentExc (translates to http code 400) instead the current error (http code 500):

```
at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1949)
	at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
	at java.time.LocalDateTime.parse(LocalDateTime.java:492)
	at java.time.LocalDateTime.parse(LocalDateTime.java:477)
	at com.graphhopper.reader.gtfs.GraphHopperGtfs$RequestHandler.<init>(GraphHopperGtfs.java:96)
	at com.graphhopper.reader.gtfs.GraphHopperGtfs.route(GraphHopperGtfs.java:247)
```

web/src/main/webapp/js/graphhopper/GHRequest.js
```diff
@@ -201,10 +202,22 @@ GHRequest.prototype.createPath = function (url, skipParameters) {
...
+                    var objKey = arr[keyIndex];
+                    url += "&" + encodeURIComponent(key) + "=" + encodeURIComponent(objKey);
+                }
+            } else {
+                // currently on one depth supported
+                var arr = Object.keys(val);
+                for (var keyIndex in arr) {
+                    var objKey = arr[keyIndex];
+                    url += "&" + encodeURIComponent(key) + "." + encodeURIComponent(objKey) + "=" + val[objKey];
+                }
```
karussell: This is ugly and limited to one depth. The intention is clear: use properties with namespaces like `pt.` in the URL like we already do in the config. This will make the transition to JSON POST requests easier.

jsroyal:
Nice doc ! +: @michaz @karussell

PR #1019, opened 2017-04-05
Fixes problem for showing the edge name from shape-file 

shadi-attarha:
I detect why the edge name was not shown when we read information from shape file.
I add the below line to method "addEdge" in class "OSMShapeFileReader"  :

edge.setName(road.getAttribute("name")+"");

and now it shows the edge name when the we read information from shape file.


karussell:
Please have a look on how to create pull requests at github here: https://help.github.com/articles/creating-a-pull-request/ there seems to be something wrong in your setup. Also the way is normally parsed if provided as name tag.

karussell:
This should be already done in OSMShapeFileReader, see line ~280 the for loop "String tag : DIRECT_COPY_TAGS"

PR #1020, opened 2017-04-05
Fixes language problem for shape reader 

shadi-attarha:
Adding "map.put("charset", "****"); " to method "openShapefileDataStore" in class "ShapefileReader"
the problem was solved.
you can replace your used character set which is determined by the .cpg file.


karussell:
I cannot see the mentioned change in the commits. Would you mind to checkout the latest master, apply your change and resubmit this in this or a new pull request?

karussell:
Furthermore would you mind to send me an email so that I can send you the PDF with [our CLA](https://github.com/graphhopper/graphhopper/blob/master/.github/CONTRIBUTING.md#license-agreement)?

Issue #958, opened 2017-02-21
Make Blocking Edges/Areas pluggable like TurnWeighting


boldtrn:
Blocking Edges/Areas is currently only possible with the GenericWeighting. It should be possible to extract that behavior into a pluggable weighting, like the TurnWeighting. This could be used for non ch requests for fastest or shortest weighting.

Implementing this should be straight forward. Move the blocking code from the GenericWeighting into a new class. It might be a good idea to extend the `AbstractAdjustedWeighting`.

asdzxcasd:
I want to get started with graphhopper development, so I'm going to implement this improvement.

I've implemented class `BlockedWeighting` that extends `AbstractAdjustedWeighting`, but I'm doubting about its name. Doesn't "forbidden" have more precise meaning than "blocked", so that the class is named `ForbiddenWeighting`?

boldtrn:
Hi @asdzxcasd, I am happy to hear that you picked up this issue :). I think blocked is a better match than forbidden. For example a road is blocked by an traffic accident or a road is blocked by flooding.

karussell:
Fixed via #1078

Issue #940, opened 2017-01-23
Too many property objects


karussell:
The current state is a bit confusing. We have AlgorithmOptions, PMap (`Map<String, String>`), HintsMap (extends PMap) and ConfigMap (`Map<String, Object>`). Maybe we can merge PMap and ConfigMap somehow and also AlgorithmOptions with HintsMap?

boldtrn:
Would be highly appreciated. I stumbled over that a couple of times as well :).

karussell:
With #1078 ConfigMap is removed

karussell:
With #1934 HintsMap is removed

PR #1029, opened 2017-04-12
Extracted blocking edges/areas from GenericWeighting into BlockedWeighting

asdzxcasd:
Related issue:  #958
I'm not sure about one thing: is it OK to change order of checks in GenericWeighting.calcWeight? When using GenericWeighting with BlockedWeighting now it is firstly checked that edge is not blocked and then other checks from GenericWeighting go. Can it affect performance in any way?

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -898,8 +898,9 @@ public class GraphHopper implements GraphHopperAPI {
 
             cMap = new GraphEdgeIdFinder(graph, locationIndex).parseStringHints(cMap, hintsMap, new DefaultEdgeFilter(encoder));
             GenericWeighting genericWeighting = new GenericWeighting(dataEncoder, cMap);
-            genericWeighting.setGraph(graph);
-            return genericWeighting;
+            BlockedWeighting blockedWeighting = new BlockedWeighting(genericWeighting, cMap);
+            blockedWeighting.setGraph(graph);
+            return blockedWeighting;
         } else if ("shortest".equalsIgnoreCase(weighting)) {
             return new ShortestWeighting(encoder);
         } else if ("fastest".equalsIgnoreCase(weighting) || weighting.isEmpty()) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -898,8 +898,9 @@ public Weighting createWeighting(HintsMap hintsMap, FlagEncoder encoder, Graph g
 
             cMap = new GraphEdgeIdFinder(graph, locationIndex).parseStringHints(cMap, hintsMap, new DefaultEdgeFilter(encoder));
             GenericWeighting genericWeighting = new GenericWeighting(dataEncoder, cMap);
-            genericWeighting.setGraph(graph);
-            return genericWeighting;
+            BlockedWeighting blockedWeighting = new BlockedWeighting(genericWeighting, cMap);
```
boldtrn: We should only use the BlockedWeighting, if we actually block something, otherwise we could only use the GenericWeighting

boldtrn:
I am not sure if you already send us the [CLA](https://www.graphhopper.com/individual-contributor-license-agreement/)?

BTW: This is not mandatory for this PR, but the overall idea for the BlockedWeighting was that we can use this for all Weightings/Encoders, similar to what we do with the `AvoidEdgesWeighting`. If you feel like adding this to this PR or after this PR was merged, this would be a great addition as well.

asdzxcasd:
Thanks for review. I'll try to fix it.

Regarding CLA I'd like to accept it, but I'm not sure how to do it as there is no submit button on that page. Or should I do it in some other way?

karussell:
Thanks & just send this CLA to the info email here: https://www.graphhopper.com/imprint/

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: It furthermore should be wrappable with the FastestWeighting and potentially others, but this might be dependent on #940. Will investigate what is possible.

karussell:
We'll go on with #1078 as no feedback here.

PR #1031, opened 2017-04-20
Refactor Instruction Generation

boldtrn:
Overall this refactoring should increase the performance of the instruction calculation, by reducing the usage of EdgeExplorers, on the other hand I extracted some logic from the InstructionFromEdges class, see https://github.com/boldtrn/graphhopper/commit/88642b4d7d801b41e17b09b41a38bb6bfe971132.

At first I had the InstructionsSurroundingEdges class as inner class, when extracting it, I needed to create a Helper class as there is some shared functionality, used by InstructionsSurroundingEdges and InstructionFromEdges. The shared functionality is about the angle calculation and comparing street names, so no heavy lifting.

@michaz would you mind to have a look at the extract? WDYT, does it makes sense to extract this part of the code?

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
+++ b/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
@@ -1,6 +1,22 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
 package com.graphhopper.routing;
 
-import com.graphhopper.routing.util.DataFlagEncoder;
 import com.graphhopper.routing.util.DefaultEdgeFilter;
 import com.graphhopper.routing.util.FlagEncoder;
 import com.graphhopper.routing.weighting.Weighting;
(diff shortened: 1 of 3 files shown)


michaz:
Looks very good to me. I'm not super-convinced that performance is the issue here, but I'm _totally_ convinced that functions should be ``static``, that members should be ´´final´´, and that refactorings after which more things can be static or final often make things clearer. So +1.

boldtrn:
> Looks very good to me. I'm not super-convinced that performance is the issue here

Thanks a lot for the review! AFAIK, there was a performance degression due to this PR. I will go ahead and merge this PR.

karussell:
Thanks @boldtrn!

Sadly it does not seem to fix the slow down: https://graphhopper.com/measurement/#routingCH

karussell:
perf-instructions

boldtrn:
> Sadly it does not seem to fix the slow down: https://graphhopper.com/measurement/#routingCH

Are you sure that this is connected to instructions? I just had a look at this page: https://graphhopper.com/measurement/#routingCH_no_instr

I can see the same slow down? No_Instr means no instructions?

karussell:
> Are you sure that this is connected to instructions?

Ah, wait. Something is wrong. (And yes "no instr" means instruction calculation is disabled)

karussell:
Where do you see the same slow down for ["routingCH_no_instr_mean"](https://graphhopper.com/measurement/#routingCH_no_instr)?

routingch_no_instr

boldtrn:
> Where do you see the same slow down for "routingCH_no_instr_mean"?

Oops, sorry, I looked at the `routingCH_no_instr_distance_mean`. Let's see what else we can tune for the instruction generation :).

karussell:
I'll profile which line causes the slow down (e.g. via profile snapshot comparison or something)

karussell:
The expensive call is the Math.atan2 in the InstructionsHelper.calculateOrientationDelta which is used in

InstructionsHelper.calculateSign
InstructionsFromEdges.getTurn

I have the feeling that we can reduce the number of (indirect) calls to this method which is called via InstructionsFromEdges.next

karussell:
Probably the loop over calculateSign in InstructionsSurroundingEdges.getOtherContinue is the problem

karussell:
As discussed offline we could try to replace the expensive Math.atan2 with the approximative AngleCalc.atan2 (it is a rather good but very fast approximation). If this is too imprecise for right/left etc detection or angle computation, we could still use it for certain use cases here IMO.

PR #1032, opened 2017-04-20
Correct too small weightings for CH

boldtrn:
As discussed in #1013, setting too small weights for CH leads to weights of 0. This can lead to very weird results. Therefore, I introduced a MIN_WEIGHT, in addition to the MAX_WEIGHT. If a weight is smaller than the min weight, the weight is corrected to the min weight. In addition I added a warning message so that developers can correct their weightings more easily.


karussell:
Hmmh, I do not know if this is the best solution. Detecting it via weighting.getMinWeight would be preferable as then we can throw this at start already. Will have to think about it.

boldtrn:
>  Detecting it via weighting.getMinWeight would be preferable as then we can throw this at start already.

Yes makes sense as well. The problem with getMinWeight is that it depends on the distance of the edge. So we might be able to generate very small weight for edges that are very short (e.g. 1m long). The current FastestWeighting would return `1/140`, which is already kind of close to the `MIN_WEIGHT`. Maybe we should define that getMinWeight for 1m cannot be smaller than `MIN_WEIGHT`, when having CH enabled?

boldtrn:
Any updates on this PR?

karussell:
What if we throw an exception instead of just silently logging it?

boldtrn:
> What if we throw an exception instead of just silently logging it?

Ok sure, we can do this.

boldtrn:
Ok, I updated this PR to throw an exception.

karussell:
Ok, I get now for world wide this problem of a shortcut with weight==0 ... maybe this is a bug in the preparation code as the start and end seems to be identical

```
Caused by: java.lang.IllegalStateException: Setting weights smaller than 0.001 is not allowed in CHGraphImpl#setWeight. You passed: 0.0 for the edge from -21.829727221199732,-42.08242838268939 to -21.829727221199732,-42.08242838268939
	at com.graphhopper.storage.CHGraphImpl.setWeight(CHGraphImpl.java:295)
	at com.graphhopper.storage.CHGraphImpl$CHEdgeIteratorImpl.setWeight(CHGraphImpl.java:484)
	at com.graphhopper.routing.ch.PrepareContractionHierarchies.addShortcuts(PrepareContractionHierarchies.java:560)
	at com.graphhopper.routing.ch.PrepareContractionHierarchies.contractNodes(PrepareContractionHierarchies.java:297)
	at com.graphhopper.routing.ch.PrepareContractionHierarchies.doWork(PrepareContractionHierarchies.java:184)
```

karussell:
Good thing is: this is locally reproducable with just a map extract

karussell:
Ah, the problem is that at this node is a very tiny circle added and all nodes have the exact same coordinate, see here where I have slightly moved the nodes to show the problem:

osm-bug


Fixed this in the data: http://www.openstreetmap.org/changeset/50833637

karussell:
Normally we can handle edges with zero distance, but we obviously cannot handle a ring structure condensed into a single coordinate

boldtrn:
Wow, thanks for the additional information and the nice improvement!

karussell:
This happens a bit too often so for now the warning, then fixing all the problems (if not too many ;)) and maybe enabling the exception again

PR #1033, opened 2017-04-21
Store toll information via DataFlagEncoder

### LINKED ISSUE #643: avoiding toll roads
i was wondering if avoiding toll roads is available with graphhopper, if there is a workaround for it or if/when this option would be available.

thank you

boldtrn:
This PR allows to store toll information in the graph. And avoiding toll roads in the GenericWeighting. It is intended to fix #643.

In this PR I only evaluate ways for the property `toll=yes`. It seems that other also check if there is a `barrier=toll_booth`.

I tried integrating toll booths as well, but I am not sure if the overhead of implementing/maintaining it is worth the benefit? I looked up a couple of toll roads and it seems like all where tagged correct, so toll booth recognition might not be required? I created an overpass-turbo query to find toll_booths that are not on a way with "toll" tagged, but I couldn't find one.

The overpass query looks like this:
```
[bbox:{{bbox}}];
way[!"toll"];node["toll_booth"="yes"](w) -> .booth;
node.booth;
out;
```

BTW: I think the overpass query is correct, I tried it with traffic signals and get results:
```
[bbox:{{bbox}}];
way[!"toll"];node["highway"="traffic_signals"](w) -> .booth;
node.booth;
out;
```

I added the [toll_booth code here](https://github.com/boldtrn/graphhopper/commit/8924dff0e5828e6471a322a94c55b4be13475a97), it's untested, but I assume it could work like that... Testing is not that trivial. I think I would end up extending the test-barriers.xml. Since I couldn't find a live example this might not be worth the effort.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
@@ -913,6 +913,10 @@ public class DataFlagEncoder extends AbstractFlagEncoder {
         cloneDoubleAttribute(weightingMap, cMap, GenericWeighting.WEIGHT_LIMIT, 0d);
         cloneDoubleAttribute(weightingMap, cMap, GenericWeighting.WIDTH_LIMIT, 0d);
 
+        if(weightingMap.has(GenericWeighting.AVOID_TOLL)){
+            cMap.put(GenericWeighting.AVOID_TOLL, weightingMap.getBool(GenericWeighting.AVOID_TOLL, false));
+        }
+
         return cMap;
     }
 
--- a/core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
@@ -85,9 +85,11 @@ public class DataFlagEncoder extends AbstractFlagEncoder {
     private EncodedValue highwayEncoder;
     private EncodedValue transportModeEncoder;
     private EncodedValue accessEncoder;
+    private EncodedValue tollEncoder;
     private boolean storeHeight = false;
     private boolean storeWeight = false;
     private boolean storeWidth = false;
+    private boolean storeToll = false;
     private EncodedValue spatialEncoder;
     private int spatialRules = 0;
     private SpatialRuleLookup spatialRuleLookup;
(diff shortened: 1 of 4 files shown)


karussell:
Thanks! Looks good.

Probably another issue. But now we are slowly getting another problem: The single purpose FlagEncoders are too limited and do not store much but have topic specific code. The DataFlagEncoder stores everything (or could) but also collects code of too many purposes with on&off switches. We need kind of a builder mechanism for the DataFlagEncoder

boldtrn:
Thanks for having a look at this PR.

BTW: Do you have any opinion on the toll_booth part?

> Probably another issue. But now we are slowly getting another problem: The single purpose FlagEncoders are too limited and do not store much but have topic specific code. The DataFlagEncoder stores everything (or could) but also collects code of too many purposes with on&off switches. We need kind of a builder mechanism for the DataFlagEncoder

Yes your are absolutely right. I wasn't even sure if should create this PR as it might get too crowded already. I created a separate issue to discuss this - #1036.

karussell:
> BTW: Do you have any opinion on the toll_booth part?

You mean that we should detect if a network of roads is toll=yes if the entrance nodes have this toll_booth tag? This would be indeed better but complicated and different to how other restrictions work in GH. Also what if some entry nodes are not properly tagged - then the full network gets toll=yes

boldtrn:
> You mean that we should detect if a network of roads is toll=yes if the entrance nodes have this toll_booth tag?

My assumption was that there are probably incorrect tagged roads. So the ways don't contain toll=yes, but there are toll_boots. So I would have simply tagged the underlying way with toll=yes to make sure we tag the road as toll=yes (similar to our current barrier node approach). But I haven't found cases like this yet, see above overpass query.

jsroyal:
@boldtrn What is the status of the this PR?
Also there is some conflict in this PR.

boldtrn:
> @boldtrn What is the status of the this PR?

Well, it's essentially finished. Conflicts should be resolvable easily. I think we want to improve the DataFlagEncoder first, because it gets very crowded. The code should work though.

karussell:
> I think we want to improve the DataFlagEncoder first, because it gets very crowded

Let us merge this and 'fix the DataFlagEncoder' via replacing it with the EncodedValue from #1112 'soonish'

boldtrn:
> Let us merge this and 'fix' the DataFlagEncoder via replacing it with the EncodedValue work from #1112 later

Sure makes sense. I can try to fix the conflicts in the next hours, but I am not sure if I can find the time. Can't promise to resolve the conflicts before Friday though. If you want to merge it now, would it be possible for you to resolve the conflicts? Otherwise I will do it until Friday.

karussell:
Closing for now. Will be easy to implement or recover after #1447 is merged.

Issue #297, opened 2015-01-10
Issue while deploying with WAR-file


drnextgis:
I've built WAR-file from current master and trying to deploy it using Jetty8 in Ubuntu.
When I try to open URL in browser I get message `GraphHopper API offline?` and errors in log file:

```
2015-01-10 20:00:23.706:WARN:oejd.WebAppDeployer:WebAppDeployer is deprecated. Use WebAppProvider
2015-01-10 20:00:23.707:INFO:oejs.Server:jetty-8.1.3.v20120416
2015-01-10 20:00:23.750:INFO:oejw.WebInfConfiguration:Extract jar:file:/var/lib/jetty8/webapps/root.war!/ to /var/cache/jetty8/data/jetty-0.0.0.0-8080-root.war--any-/webapp
2015-01-10 20:00:24.350:INFO:oejw.StandardDescriptorProcessor:NO JSP Support for , did not find org.apache.jasper.servlet.JspServlet
2015-01-10 20:00:24.367:WARN:oejw.StandardDescriptorProcessor:Could not instantiate listener com.graphhopper.http.GuiceServletConfig
java.lang.ClassNotFoundException: com.graphhopper.http.GuiceServletConfig
        at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
        at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:424)
        at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:377)
        at org.eclipse.jetty.server.handler.ContextHandler.loadClass(ContextHandler.java:1471)
        at org.eclipse.jetty.webapp.StandardDescriptorProcessor.visitListener(StandardDescriptorProcessor.java:1839)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #1040, opened 2017-04-25
Move importOrLoad outside of Guice and use addShutdownHook

karussell:
Furthermore I made the createGraphHopper method protected to being able using the same instance with SpatialRuleLookupBuilder

### DIFF
--- a/core/files/changelog.txt
+++ b/core/files/changelog.txt
@@ -1,4 +1,5 @@
 0.9
+    rename of DefaultModule to GraphHopperModule and GHServletModule to GraphHopperServletModule
     EncodedValue uses Math.round(value/factor). This can change the retrieved values for EncodedValues #954
     EncodedDoubleValue and EncodedValue requires maxValue/factor to be a natural number #954
     default base algorithm for all modes is bidirectional A* (except speed mode)
--- a/web/pom.xml
+++ b/web/pom.xml
@@ -55,14 +55,14 @@
         <dependency>
             <groupId>com.google.inject</groupId>
             <artifactId>guice</artifactId>
-            <version>4.0</version>
+            <version>4.1.0</version>
         </dependency>
 
         <!-- necessary to use guice ('@Inject') in servlets -->
         <dependency>
             <groupId>com.google.inject.extensions</groupId>
             <artifactId>guice-servlet</artifactId>
-            <version>4.0</version>
+            <version>4.1.0</version>
         </dependency>
 
         <dependency>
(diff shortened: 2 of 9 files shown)


michaz:
I really wouldn't recommend extending ``DefaultModule``, or designing it to be extended.

I like the ShutdownHook, but did you test this? I faintly remember that getProvider() doesn't work _after_ provisioning, so it would only work the way it is in my code (being called and remembered at configuration time). But I'm not sure.

Using ShutdownHooks should probably mean we should remove the Servlet deployment option by deleting ``GuiceServletConfig``.

karussell:
> I really wouldn't recommend extending DefaultModule, or designing it to be extended.

Ok, then I'll create a static helper or something so that one can reuse the SpatialLookup init code.

> I like the ShutdownHook, but did you test this?

No, will call this earlier then. Should we create an (integration) test for this? 

> Using ShutdownHooks should probably mean we should remove the Servlet deployment option by deleting GuiceServletConfig.

Ah, now I understand the difference. Then I'll keep your idea with the filter for now.

michaz:
> GuiceServletConfig

Speaking of this: This probably doesn't currently work anyway. GHServer and GuiceServletConfig create their modules differently -- they shouldn't, they should create exactly the same Module. Otherwise, we need all tests for both of them separately. ,-)

Or we should delete it.

michaz:
The most principled way would probably if ``GHServer`` would use ``GuiceServletConfig`` to create its stuff, and not use its own way to create an injector.

karussell:
The war:war deployment makes problems anyway #297, maybe we should just drop support for it and remove GuiceServletConfig

karussell:
Making the tests identical was easy via `setUpGuice(server.createModule());` in BaseServletTester

michaz:
> Also this probably means that GHServer isn't really testable (as a Java class) anymore. It leaks resources which are only closed when the JVM quits. Whereas previously, jetty would close everything on stop().

Hmm, we have to remember this now: server.stop() does not close the resources now. With the filter version, it would have. Isn't there a jetty feature like a stop()-hook where we can close, instead of a Runtime-shutdown-hook?

karussell:
> Isn't there a jetty feature like a stop()-hook where we can close, instead of a Runtime-shutdown-hook?

No idea, but our GHServer has a stop method which we should probably better call?

karussell:
Also as this was already merged by me you would have to create a new PR unfortunately...

michaz:
Yes I know. I just wanted to re-emphasize what this means.

- Jetty stops itself on JVM shutdown
- now we also close resources on JVM shutdown
- GHServer.stop results in Jetty.stop
- but Jetty.stop does not result in closing resources.

michaz:
Ok, I'll look for a jetty-stop-listener or something, then we can close the resources there and don't need our own shutdown hook.

PR #1042, opened 2017-04-27
Replaced Math.atan2 with atan2 table lookup implementation. 

boldtrn:
I changed the atan2 implementation of the AngleCalc to a TableLookup version that was provided by @karussell in https://github.com/graphhopper/graphhopper/commit/d64ab7b.

I compared the current master with this version using VisualVM. I used car with ch=fastest. Using the current master, It seems that the `AngleCalc.calcOrientation` was in the Top 10 of CPU time consuming methods of the overall route calculation. With the updated atan2 implementation `AngleCalc.calcOrientation` does not show up in the top 20 CPU time consuming method calls anymore. So it seems to me that this might fix the delay discussed in #1031.

The new atan2 implementation returns slightly different values to `Math.atan2`. I could not find any issues though. I needed to fix two tests. Also I had to add %360 operation, as with the new atan2 implementation we are returning 360 instead of 0. Tests expect 0, so I figured it might be better to do a modulo here instead of changing the tests and maybe introducing a bug in the code.


karussell:
Would you do the comparison with the measurement script? (and an area with size like bavaria or saxonia). Then you can look into concrete numbers like `routingCH.mean` and `routingCH_no_instr.mean` for with and without this change.

I'm not sure if we should use it always, especially since we'd export this slightly wrong value to external consumers (Java/JSON API) for roundabouts (RoundaboutInstruction.setRadian)

> Also I had to add %360 operation, as with the new atan2 implementation we are returning 360 instead of 0

Should we do the modulo operation for the root calcOrientation instead?

boldtrn:
Here are the results of the measurement, I used: `./graphhopper.sh measurement baden-wuerttemberg-latest.osm.pbf`. I ran it two times for every branch, master and speedup. There is a speedup, but it seems like that it is not the 20% we are looking for.

| Runs         | routingCH.mean | routingCH_no_instr.mean  |
| ------------- |:-------------:| -----:|
| speedup 1 | 2.4596033764 | 1.838547689 |
| master 1 | 2.58897116 | 1.7966665841999998 |
| speedup 2 | 2.305940648 | 1.6758486566 |
| master 2 | 2.4005281016 | 1.7010407452 |

> I'm not sure if we should use it always, especially since we'd export this slightly wrong value to external consumers (Java/JSON API) for roundabouts (RoundaboutInstruction.setRadian)

I can create create a parameter for the calcOrientation method like `exact` that is true per default, using a wrapper method, so no existing code gets broken. 

> Should we do the modulo operation for the root calcOrientation instead?

Mhm, might not be that easy. The case where it returns 360, is when calcOrientation returns a value slightly smaller than 0, like -.0000000000000001. That way we call `Math.toDegrees((2 * Math.PI) - .0000000000000001)` which returns 360.

karussell:
> There is a speedup, but it seems like that it is not the 20% we are looking for.

Indeed there is a speedup and I think there are further possibilities.

BTW: strange that you get such different values for the different runs. If you close all other applications and do two runs there shouldn't be a bigger difference than 5%

> Mhm, might not be that easy.

ok

karussell:
I investigated a bit further and when doing the angle calculation for *every* junction in the old code (https://github.com/graphhopper/graphhopper/commit/951b443) it got ~15% slower which is still ~5% faster than the state with this PR but explains a big part of the slow down. We could do some pre-calculation on import for the angle stuff, but for now I think this is ok. 

Maybe you can have a second look if we avoid the loop in getOtherContinue over the edges leading to many tan2 calculations calls? And if not possible we keep it as it is with this PR and just introduce an exact=true parameter for the RoundaboutInstruction.setRadian

BTW: we currently do not cache the fetchWayGeometry calls (also many calls via getPointForOrientationCalculation), but temporarily replacing this with a fixed PointList did not speed up anything
BTW: line 213 looks very similar to the code in getPointForOrientationCalculation

boldtrn:
> BTW: line 213 looks very similar to the code in getPointForOrientationCalculation

Yes, but it does different things. Putting the logic into the `getPointForOrientationCalculation` is not worth the effort IMHO.

> Maybe you can have a second look if we avoid the loop in getOtherContinue over the edges leading to many tan2 calculations calls?

I spent some time thinking about this a bit more. I came up with a [heuristic](https://github.com/boldtrn/graphhopper/commit/235df5a4fc71d43ee5842dfba1c4f27202bb79c9) that is a lot faster, but actually provides worse results, also a couple of tests fail. The idea is to use the current edge and use its point for orientation calculation as reference. Then we take the point for orientation calculation of the other edge and project it on a straight line to be equal distance than the other point. We then compare the distance between the points and the distance of the point to the base, if the points are closer to each other than to the base, this could be a continue. It might be easier to just read the code ;). So if we need speed, this could be used, but I wouldn't recommend it, as the results are worse. 

I think the best would be to keep it as is and simply add the exact flag.

boldtrn:
I forgot to add that the heuristic got about 10% speedup for ch with measurement and Baden-Württemberg, but before investing more time we should remeasure this as I only did one rin. Maybe somebody has a good idea on how to improve it, as the speedup might be worth it.

karussell:
> Putting the logic into the getPointForOrientationCalculation is not worth the effort IMHO.

ok

> I think the best would be to keep it as is and simply add the exact flag.

:+1:

boldtrn:
Ok, I added the exact flag, to make sure that we are actually doing the right thing, I reran the performance measures.

Master: routingCH.mean=2.426111127 routingCH_no_instr.mean=1.7078601832
PR: routingCH.mean=2.2701361754000002 routingCH_no_instr.mean=1.7059984101999999

karussell:
Thanks!

PR #1044, opened 2017-04-27
switch from json.org to jackson - minimal version

michaz:
Replaces json.org with jackson. Basically 1:1, without making much use of its features yet.

This does _not_ mean we can't switch to Gson. On the contrary, those two are very similar, and at the moment it would be search+replace, I think.

### DIFF
--- a/web/src/main/java/com/graphhopper/http/GHBaseServlet.java
+++ b/web/src/main/java/com/graphhopper/http/GHBaseServlet.java
@@ -17,10 +17,11 @@
  */
 package com.graphhopper.http;
 
-import com.graphhopper.GHRequest;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.graphhopper.routing.util.HintsMap;
-import org.json.JSONException;
-import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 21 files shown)


karussell:
Thanks! And +1 for jackson for now ...

web/src/main/java/com/graphhopper/http/GHBaseServlet.java
```diff
@@ -65,33 +71,33 @@ protected void writeJson(HttpServletRequest req, HttpServletResponse res, JSONOb
...
+                writeResponse(res, callbackName + "(" + objectMapper.writeValueAsString(json) /*.toString(2)*/ + ")");
             else
-                writeResponse(res, callbackName + "(" + json.toString(0) + ")");
+                writeResponse(res, callbackName + "(" + objectMapper.writeValueAsString(json) /*.toString(0)*/ + ")");
 
         } else {
             res.setContentType("application/json");
             if (debug)
-                writeResponse(res, json.toString(2));
+                writeResponse(res, objectMapper.writeValueAsString(json) /*.toString(2)*/);
```
karussell: Here we would need an additional "pretty" objectMapper: 

ObjectMapper prettyOM = new ObjectMapper();
prettyOM.enable(SerializationFeature.INDENT_OUTPUT);

michaz: Ah, this can also be done "by request".

michaz: The super-advanced DI-way would probably be to bind() a request-scoped ``ObjectWriter`` which is debug or non-debug, and then inject that configured ``ObjectWriter`` everywhere..

michaz: Apparently, this is _the_ use case. From the doc:

> Construction of new ObjectReaders and ObjectWriters is a very light-weight operation so it is usually appropriate to create these on per-call basis, as needed, for configuring things like optional indentation of JSON.

karussell:
Thanks, I will merge. (this time I can squash I guess?)

karussell:
This then fixes #773 I think (?) and would you mind to do similar work for the map matching :) ?

michaz:
> Thanks, I will merge. (this time I can squash I guess?)
Yes please.

karussell:
just fyi: I just stumbled over [a possibility](https://stackoverflow.com/a/33578478/194609) which avoids a jackson dependency in the core module.

Issue #1025, opened 2017-04-10
Be aware of time zones


michaz:
We currently use ``LocalTime`` and pretend the issue does not exist. Should be easy to change, since GTFS feeds are supposed to contain the time zone to which the ``LocalTime`` is local.

michaz:
API change: time parameters are now expected to be ISO-8601 strings, like this: ``2013-05-30T23:38:23.085Z``, to be parsed into a ``java.time.instant``.

PR #1046, opened 2017-04-29
Time-zone awareness for pt #1025

michaz:


michaz:
API change: time parameters are now expected to be ISO-8601 strings, like this: ``2013-05-30T23:38:23.085Z``, to be parsed into a ``java.time.Instant``.

michaz:
Note that this is not so clear a decision. When you query ``bahn.de`` or (especially) ``lufthansa.com``, times always mean local times relative to where the trip starts, and how they would appear in the schedule.

However, we are aiming at a multi-modal "ground" route planner, so we resolve physics-time to schedule-time when we enter the pt-subgraph, and the rest of the world doesn't know about schedule-time.

Helping the user with finding out which time he/she probably means when doing a query is left for another layer.

michaz:
When we do another query type - an explicit station-to-station query -, we can add an option to search by local time, because stations have an (authoritative) local time from gtfs.

michaz:
> BTW: why los angeles as timezone?

Well, because the gtfs-test-agency is there.

Issue #1050, opened 2017-05-02
move gson dep to jackson


karussell:
(Use jackson also for the map matching component instead of json.org)

michaz:
Done.

PR #1052, opened 2017-05-02
Replace gson with Jackson; remove custom geometry data structures #1050

michaz:
Removed (only) those classes which were directly meant as deserialized forms of GeoJson. The rest is still there.


karussell:
Partly fixes #1050

PR #1053, opened 2017-05-02
Remove remaining traces of removed WAR deployment support

michaz:


karussell:
Ah, thanks, related to #297

karussell:
(Full in the sense of: make peace not war ;))

Issue #1041, opened 2017-04-26
Upgrade jetty


karussell:
Upgrade of jetty from 9.3 to 9.4 resulted in broken image files as it now somehow automatically redirects to /index.html ... will revert https://github.com/graphhopper/graphhopper/commit/e8ff9a9398973cba6e03a65efcc1af5e4fa62bbd

boldtrn:
There seems to be something wrong with the current jetty configuration anyway. When I start the server with the current master, I get exception for not found images, like the favicon or the picture generic vehicle.

```
WARN  org.eclipse.jetty.servlet.ServletHandler - /favicon.ico
java.lang.NullPointerException
	at com.graphhopper.http.GHBaseServlet.writeJsonError(GHBaseServlet.java:94)
	at com.graphhopper.http.InvalidRequestServlet.service(InvalidRequestServlet.java:30)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:848)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1772)
	at com.graphhopper.http.IPFilter.doFilter(IPFilter.java:60)
	at com.graphhopper.http.CORSFilter.doFilter(CORSFilter.java:38)
	at com.graphhopper.http.HeadFilter.doFilter(HeadFilter.java:37)
	at com.google.inject.servlet.ManagedFilterPipeline.dispatch(ManagedFilterPipeline.java:120)
	at com.google.inject.servlet.GuiceFilter.doFilter(GuiceFilter.java:135)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1759)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:582)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1180)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:512)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1112)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141)
	at org.eclipse.jetty.server.handler.HandlerList.handle(HandlerList.java:52)
[...]

PR #1055, opened 2017-05-03
Upgrade Jetty. #1041

michaz:
```
    /* ------------------------------------------------------------ */
    /**
     * @param redirectWelcome
     *            If true, welcome files are redirected rather than forwarded to.
     *            redirection is always used if the ResourceHandler is not scoped by
     *            a ContextHandler
     */
    public void setRedirectWelcome(boolean redirectWelcome)
    {
        _resourceService.setRedirectWelcome(redirectWelcome);
    }
```

### DIFF
--- a/web/pom.xml
+++ b/web/pom.xml
@@ -16,7 +16,7 @@
         <version>0.9-SNAPSHOT</version>
     </parent>
     <properties>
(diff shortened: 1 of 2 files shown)


karussell:
Cool, thanks - works as advertised :) !

PR #1057, opened 2017-05-05
flat url parameters fixes #1034

### LINKED ISSUE #1034: Webclient can't handle '.' in parameter names, so pt parameter names remain without '.'
Following up on #1018 - the attempted fix didn't work. It would only work _once_, but just moving one of the markers would break it. I reverted the parameter renaming and the change in the web client.

If we still want to rename the parameters to 'pt.*', we have to try again.

karussell:
tries to fix #1034


michaz:
Ah, please don't do global reformattings in such pull requests. That's bad for merging.

michaz:
Ok, works for me! Note that I have not researched the point in question, i.e. I don't know if it's standard or good practice to do build hierarchical parameter objects like this. I guess you did. ,-)

karussell:
> Ah, please don't do global reformattings in such pull requests. That's bad for merging.

Oh. Do you mean stuff like [here](https://github.com/graphhopper/graphhopper/pull/1057/files#diff-754799f1d31c4863491e5d3cb1207843R243)?

> i.e. I don't know if it's standard or good practice to do build hierarchical parameter objects like this. I guess you did. ,-)

I didn't, but I know that we need this when switching to JSON requests.

michaz:
> Oh. Do you mean stuff like here?

Yeah, but on second thought I wasn't even sure if I didn't do that myself. ,-)

michaz:
> I didn't, but I know that we need this when switching to JSON requests.

Oh, interesting.. What are the considerations about switching to JSON requests? I know little about this, but my understanding was that http-GET always works with resource paths and not with request bodies, so that would make our routing requests POSTs? Is that a good thing?

karussell:
> I know little about this, but my understanding was that http-GET always works with resource paths and not with request bodies, so that would make our routing requests POSTs? Is that a good thing?

Yes. Still we would support GET too. A GET request would work either via our "lazy json" formatting with **readable** parameters ala pt.xy=ab instead of the request body `{"pt":{"xy": "ab"}`. Another way is to use an URL-encoded JSON as **unreadable** URL parameter like `q={%22pt%22:{%22xy%22:%20%22ab%22}}` ... e.g. ElasticSearch does this. Not sure if our readable parameter stuff is really necessary.

Issue #194, opened 2014-05-02
Engine warm up


devemux86:
Τhe 1st route calculation is significantly slower than the next ones, which perform faster and seem to getting quicker over time.
This is more observable at Android where we're dealing with more limited resources.

A proposed way to overcome this initial behavior is a random query through the area, performed at the background after loading the graph.

PR #1059, opened 2017-05-08
Issue #194 added method for engine warm up and test

theaaron123:
#194

### DIFF
--- a/core/src/main/java/com/graphhopper/util/EngineWarmUp.java
+++ b/core/src/main/java/com/graphhopper/util/EngineWarmUp.java
@@ -1,26 +1,51 @@
 package com.graphhopper.util;
 
+import com.graphhopper.GHRequest;
+import com.graphhopper.GraphHopper;
 import com.graphhopper.storage.GraphHopperStorage;
-import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
 
 public class EngineWarmUp {
     /**
      * 'Warm up' virtual machine so initial route calculation is not slower than consequent. To be used
      * on a less powerful machine like Android or Raspberry Pi.
      */
-    public static void warmUp(GraphHopperStorage ghStorage, int iterations) {
+    public static void warmUp(GraphHopper graphHopper, int iterations) {
+        GraphHopperStorage ghStorage = graphHopper.getGraphHopperStorage();
         if (ghStorage == null)
             throw new IllegalStateException("Do a successful call to load or importOrLoad before routing");
         if (ghStorage.isClosed())
             throw new IllegalStateException("You need to create a new GraphHopper instance as it is already closed");
-        if(ghStorage.isCHPossible() == false)
-            throw new IllegalStateException("Cannot warm up non CH network");
 
-        BreadthFirstSearch bfs = new BreadthFirstSearch();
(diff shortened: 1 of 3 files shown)


karussell:
Thanks!  As a start this is good to go. Technically I would just have it into a separate class to avoid growing GraphHopper.java even more. And also removing the test ... as this is not really a test and not really testable I fear.

And also idea is simple&good but would fail for certain areas. So as a second step we need something more graph based (e.g. a BFS per subnetwork) that warms up the whole graph. Surely this would need special care as it would be forbidden for non-CH requests. Maybe then we could use BFS and skip certain nodes randomly then.

karussell:
Thanks, this looks nice! 

Sorry for my trouble and my incompleteness at the first comment: would you mind to do three more changes (should be the last :) !!):

 * make it possible via CH with your old query method but instead of lat,lon pick random node ids and catch exceptions and retry if e.g. different subnetworks (I think a warmup should never fail with an exception?)
 * limit BFS search to max_visited_nodes

core/src/main/java/com/graphhopper/util/EngineWarmUp.java
```diff
@@ -1,26 +1,51 @@
...
+    }
+
+    private static void warmUpNonCHSubNetwork(GraphHopper graphHopper, int iterations) {
+        GraphHopperStorage ghStorage = graphHopper.getGraphHopperStorage();
+
+        BreadthFirstSearch bfs = new BreadthFirstSearch();
         for (int i = 0; i < iterations; i++) {
-            int startNode = rand.nextInt(ghStorage.getBaseGraph().getNodes() + 1);
+            int startNode = ThreadLocalRandom.current().nextInt(0, graphHopper.getMaxVisitedNodes() + 1);
             bfs.start(ghStorage.getBaseGraph().createEdgeExplorer(), startNode);
```
karussell: The previous nextInt code is correct. The BFS needs to stop after graphHopper.getMaxVisitedNodes() has reached i.e. something like:

```
// it is save to create this once and reuse for others as we have no different states or threads involved:
EdgeExplorer explorer = ghStorage.getBaseGraph().createEdgeExplorer()
for loop {
  // fresh instance necessary to reset counter every time
  new BreadthFirstSearch() {
     int counter = 0;
     public boolean goFurther(int nodeId) {
        counter++;
        return counter < maxVisitedNodes;
     }
  }.start(edgeExplorer, startNode)
}
```

This is not entirely correct as we can visit certain nodes multiple times but we would have to call `bitset.getCardinality` on every call which is too expensive and so not necessary for this 'warmup' purpose.

karussell:
Thanks - would you mind to add catching all exceptions?

karussell:
Didn't mean to rethrow ... the problem could be that you pick one point from the mainland and one from the island with no connection. So just catch the exception and log it. But no counter & no rethrow necessary - I also do not know of any example where a 2nd request should succeed if the previous one failed.

theaaron123:
@karussell  sorry, log by printing the stack trace, creating an instance of logger, or by throwing with an error message?

karussell:
creating an instance of Logger via slf4j LoggerFactory

PR #1066, opened 2017-05-16
merge PMap and ConfigMap

karussell:
The only major change is that [PMap now extends ConfigMap](https://github.com/graphhopper/graphhopper/pull/1066/files#diff-2089a5d49b9bef4726a1a85bb1e12c7dL30), which makes it compatible and less annoying to handle this. This is an improvement regarding #940 and also PMap was renamed to StringConfigMap.

TODOs:

 - [ ] cleanup GraphHopper.createWeighting

Normally I would say we should disregard the `Map<String,String>` (StringConfigMap) and always prefer the ConfigMap but the problem for a further cleanup is that e.g. GHRequest uses String values and e.g. the weighting is later converted in the GraphHopper class into the `Weighting`-**Object** which is then put into the AlgorithmOptions. So we use the string as a key for a "`Weighting`"-factory (similar for the vehicle and even a bit trickier for the `algorithm`-String) and this conversion is part of the `core`-module and e.g. not separated out into the web as e.g. it should be easy to use via Android too. We'll have to think further about this.

### DIFF
--- a/core/src/main/java/com/graphhopper/GHResponse.java
+++ b/core/src/main/java/com/graphhopper/GHResponse.java
@@ -17,7 +17,7 @@
  */
 package com.graphhopper;
 
-import com.graphhopper.util.PMap;
+import com.graphhopper.util.StringConfigMap;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -30,7 +30,7 @@ import java.util.List;
  */
 public class GHResponse {
     private final List<Throwable> errors = new ArrayList<Throwable>(4);
-    private final PMap hintsMap = new PMap();
+    private final StringConfigMap hintsMap = new StringConfigMap();
     private final List<PathWrapper> pathWrappers = new ArrayList<PathWrapper>(5);
     private String debugInfo = "";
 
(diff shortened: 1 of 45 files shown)


karussell:
Hmmh, probably not a good idea. Only if we could guarantee that objects are only read otherwise it could be that a property is unintended 'serialized' into a string.

karussell:
Ok, a better approach is to deprecate the PMap usage and do "HintsMap extends ConfigMap". This implies that all values are objects and not strings, improving the situation on the Java side for future usage. E.g. we could just parse a JSON and use this as ConfigMap.

The problem then is that this breaks Java API compatibility:

 * put("xy", "true") => ClassCastException for getBoolean("xy", true); // the stored "true" cannot be casted to a Boolean
 * put("xy", new Date()) => ClassCastException for get("xy", "17.05.2017"); // the stored Date is not a String

karussell:
I would say that we accept the downsides. What is your view on this @michaz? E.g. when I do this then many gtfs tests fail, but making them pass via explicit `date.toString` or even getting the real object without parsing (and parsing is moved to the web module) should not be hard.

Here is the change (currently failing): https://github.com/graphhopper/graphhopper/commits/hintsmap

PR #1068, opened 2017-05-24
docker-compose file

legraina:
This pull request propose a docker-compose file for running graphhopper. It eases the modification of environment variables. The graphhopper script has also been modified to manage path to a map or to donwlod this map at the right place if necessary. Finally, I've also modified the README.md to add the docker-compose instructions.


karussell:
Thanks!

I know that the docker file cannot be moved but as the root folder is already a bit crowded - can we move docker compose file somewhere else? 

>  graph-hopper:
>        container_name: disp-graph-hopper

I would prefer graphhopper here without `-`

And would you mind to send [an email to us](https://www.graphhopper.com/imprint/)? (just that I can send you our [CLA](https://github.com/graphhopper/graphhopper/blob/master/.github/CONTRIBUTING.md#license-agreement) to sign easily via signaturit)

legraina:
Ok I will change the name to graphhopper and perhaps osm-graphhopper for the container_name.
Where should I move the docker-compose ?

karussell:
> Ok I will change the name to graphhopper and perhaps osm-graphhopper for the container_name

ok

> Where should I move the docker-compose 

probably to `/core/files`? (we should move this to `/files` at some point :))

legraina:
Hi, sorry for the delays ... I've moved the docker-file in core/files/

karussell:
Thanks!

Why are the changes in the graphhopper.sh script necessary? Are you sure this is working as I think the DATADIR often misses the '$' sign (?)

legraina:
Yuo are right, I've added 2 $ signs to DATADIR.

karussell:
Where does the DATADIR come from? Did you test this with and without docker?

If the DATADIR is not specified then `cd $DATADIR` will make problems, i.e. the way without docker would be broken.

legraina:
Yes, for sure. There is no need to define it. The script computes this directory:
````bash
DATADIR=$(dirname "${FILE}")
````
The goal is to be able to give the directory where the osm file needs to be saved, e.g. 'data/europe_germany_berlin.pbf'

karussell:
Sorry, didn't see this.

karussell:
Have merged it.

So the DATADIR is only necessary if we download from geofabrik? 

And how would I be able to change it?

legraina:
What do you mean by changing it ? The datadir is used for the geofabrik. For the others, it shouldn't change anything as you were already able to give /data/europe_germany_berlin.ghz without any issue.

karussell:
Ah, got it. So if the PBF is not present it tries to download, but before it was only possible to say europe_germany.pbf and not data/europe_germany.pbf

legraina:
You can still do europe_germany.pbf .

PR #1070, opened 2017-05-26
Added IT for DataFlagEncoder+SpatialRuleLookup

### LINKED ISSUE #988: create IT for DataFlagEncoder & SpatialRuleLookup
See the discussion: https://github.com/graphhopper/graphhopper/pull/966#issuecomment-285901672

boldtrn:
Fixes #988.

Since we init the SpatialRuleLookup in the web module I needed to create an IT, including starting jetty, which is not ideal, but right now there is no other way (?) to make sure everything works.

I used Bayreuth North, since it is covered by the German SpatialRule and contains enough small tracks that blocked by the SpatialRule to make sure it is enforced.


boldtrn:
For some reason the tests fail on travis, but work locally. The server does not start on travis. Maybe because starting the generic vehicle takes more ram or maybe because bayreuth is too large?

karussell:
I think the problem is this:

> 1) Error in custom provider, java.lang.RuntimeException: java.io.FileNotFoundException: ../web/src/main/resources/com/graphhopper/spatialrules/countries_full.geo.json (No such file or directory)

boldtrn:
> I think the problem is this:

Ouch, thanks for finding this :).

PR #1071, opened 2017-05-26
Updated translations

boldtrn:
I updated some of the translations.

I removed the `!` from the German and English finish translation. We don't use a `!` in the other instructions, why do we do this for the last one, seems a bit inconsistent?

I changed the German translation for `continue` to `dem Straßenverlauf folgen` and `continue_onto` to `dem Straßenverlauf von %1$s folgen`. These seem to better describe what is happening for the user.

I switched the translation for `finish`. I changed the case, since other translations usually start with a lower case, so I switched it for the finish instruction as well. Also I changed it to `arrive at destination`. I wasn't sure if `destination reached` would be a better translation. I choose to use `arrive at destination` as this seems to be the common phrase used by other routing engines. Usually they also add the name of the road, which could be a nice addition for use as well?

So it would be like `arrive at main road`.

Let me know if you are OK with the changes or if I should change anything. Translations are always very subjective, so let me know if you have a different opinion.


karussell:
Thanks :+1: 

> We don't use a ! in the other instructions, why do we do this for the last one, seems a bit inconsistent?

Historical reasons (!!) ;)

> Usually they also add the name of the road, which could be a nice addition for use as well?

And and also the side but lets do this in another change.

Issue #1048, opened 2017-04-29
Instructions: Missing instruction when turning onto road with separated one way lanes


boldtrn:
We do not show turn instructions if the turn we are currently performing is the only valid turn. This is very useful in a lot of situations to remove unnecessary and misleading instructions. In rare cases this leads to missing instructions. Here is an example:

https://graphhopper.com/maps/?point=52.508392%2C13.326287&point=52.509456%2C13.325692

boldtrn:
Should we instead of checking all reachable roads, check all surrounding roads? We will see more instructions, but maybe it's better to show more instructions than missing one?

This is also important for generating a u-turn instructions as currently proposed in #1073.

karussell:
Probably we should take into account oneways like here, but considering all ways might be not so good. Assume there is a junction or roundabout with foot-only ways which could produce confusing instructions for car.

boldtrn:
Yes I would only use roads that are accessible by that encoder.

boldtrn:
@karussell in #1252 provided some more examples where this happens. Relevant for this issue are only the first two red circles.

With the improved instructions I would have expected instructions at the red circles in [this route](https://graphhopper.com/maps/?point=42.507034%2C1.531906&point=42.509581%2C1.532421&layer=OpenStreetMap) in Andorra:

 1. left
 2. right

![missing-instructions](https://user-images.githubusercontent.com/129644/34416877-e382327e-ebf5-11e7-999c-300be3398f6b.png)

Issue #446, opened 2015-06-16
Turn Restriction with one or more ways as a via member


mapsg:
Hello to community
I was using GraphHopper for routing when I noticed that the  Turn Restrictions with one or more ways as a via member do not supported from the routing engine. That's right ? Are you planning to support this feature?
A typical example
http://www.openstreetmap.org/relation/442887

karussell:
Yes, currently only nodes are allowed for the 'via' of the turn restriction. As these kind of turn restrictions do not occur very often we will create additional nodes and ways in a preprocessing to support this.

BTW: in this specific example just forbid the left turn and all is fine ;)

devemux86:
See also http://wiki.openstreetmap.org/wiki/Michigan_Lefts

matkoniecz:
Currently there are more than 50 000 `via` ways (2% of all `via` roles for turn restrictions). See https://taginfo.openstreetmap.org/relations/restriction#roles

easbar:
>  We would have to add turn restrictions to B1 (can only go from C4 to D2),

Ok this would force going from C4 to D2 and we can prevent the u-turn this way (using the artificial edge), but at the same time we cannot go straight (C4-B1-C2) anymore ?!

This problem is also discussed here: 

* https://github.com/valhalla/valhalla/issues/224
* https://github.com/valhalla/valhalla/issues/1801

karussell:
> but at the same time we cannot go straight (C4-B1-C2) anymore ?!

I think the B1-C2 edge is not a oneway one (my explanation was a bit unclear)

easbar:
> I think the B1-C2 edge is not a oneway one

Ok, but you said `We would have to add turn restrictions to B1 (can only go from C4 to D2)` and if we do this the turn restrictions prevent us from going C4-B1-C2 so we can no longer drive on the road normally (go straight on our way from C4 to C1).

karussell:
Oh, I think what I meant is: `add a turn restriction to B1 (can not go from C2->B1->D2)` so C4-B1-C2-C1 should still be possible. I hope this works now :) ?

easbar:
Ok, to summarize:
```
    A1
    | 
C1--C2--B1--C4
    | /
D1--D2--D3
    | 
    E1
```

- we introduced an artificial node B1 and edges C4->B1 and B1->C2
- added an artificial edge (oneway) B1->D2
[...]

PR #1073, opened 2017-05-26
Add U-Turns

### LINKED ISSUE #1064: Instructions: Add U-Turn Instruction
Currently there is no u-turn instruction for GraphHopper. 

There are two types of u-turns:
1: Being on a big road with separated lanes, e.g. https://graphhopper.com/maps/?point=52.508875%2C13.326448&point=52.50896%2C13.326555
2: Being on a normal road. This usually happens when there is a via point on that road or when navigating and taking a wrong turn, e.g. https://graphhopper.com/maps/?point=52.97255%2C12.879045&point=52.975069%2C12.881727&point=52.972576%2C12.879045

Handling case 1 won't be too easy and would probably require #1007.
Handling case 2 should be doable with checking the orientation and/or checking if the prev edge id is equal to the next edge id.

So this issue is about adding U-Turn instructions at all and provide them for case 2.

boldtrn:
This PR fixes #1064. This PR is adding U-Turn instructions. 

We recognize 3 types of U-Turns:
- Heading is opposite to the orientation of the first instruction. Convert CONTINUE => U-TURN
- At waypoints, if we have to turn around at the waypoint. Convert CONTINUE => U-TURN
- When performing two turn in the same direction in a very short distance. The street before and after the turn have to have the same name and the orientation has to be more or less opposite. This converts two TURNS => One U-TURN.

http://localhost:8989/?point=48.005027%2C7.997961&point=48.005874%2C7.999538&heading=230
initial-u-turn
http://localhost:8989/?point=48.017509%2C8.008572&point=48.019012%2C8.008116&point=48.018319%2C8.010836&vehicle=bike
u-turn-wp
http://localhost:8989/?point=47.996983%2C7.841899&point=47.996987%2C7.842087
example-u-turn

The third method performs pretty well for cars and fixes most of the u-turns. One issue is that we require two turn instructions. In some cases we don't create a turn instruction because there is no other possible turn, in this case we cannot convert the instruction to a u-turn, e.g., [here](http://localhost:8989/?point=48.005874%2C7.833906&point=48.005738%2C7.83374).

I also required that the road to u-turn on is a one way. I have seen some cases where a u-turn was created that is not a real u-turn. Here are some examples in Baden-Württemberg. I found these examples using measurement, with foot and printed the location of every u-turn. So these examples are not very frequent, but they exist:
- http://localhost:8989/?point=48.870569%2C8.646058&point=48.870509%2C8.645768
- http://localhost:8989/?point=47.881974%2C7.824572&point=47.882046%2C7.824427
- http://localhost:8989/?point=47.954715%2C7.777317&point=47.954693%2C7.777017
- http://localhost:8989/?point=48.893573%2C10.004784&point=48.893471%2C10.005176&vehicle=foot
- http://localhost:8989/?point=49.769659%2C9.469453&point=49.769784%2C9.469732

When only showing u-turns for one-way streets the samples above don't produce turn instructions. When thinking about this a bit more I think usually u-turns are only relevant for major roads that are usually one-ways separated from each other with designated u-turn lanes or u-turning at traffic lights. So this limitation makes sense for me. 

TODO:

- [x] Add U-turn icon
- [x] Decide if we want different signs for left and right u-turns
- [x] Decide if we want a different translation for the different u-turns


karussell:
A heading per instruction is not a bad idea as e.g. the very first continue after every via or start point could need something like "go north" which the client then could at least create on its own.

But before exposing this into the public JSON response lets thing twice or more about the "if" and "how" :)

> While doing that I realized that in most cases GraphHopper won't actually lead you back on the same edge that you were coming

What do you mean here? With a via point it is not that unlikely (?)

boldtrn:
> What do you mean here? With a via point it is not that unlikely (?)

Yes exactly, the typical case would be a via-point. But GrapHopper calculates the instructions independently for every path between waypoints (see PathMerger). So Having a route a->b->c, GH calculates instructions for a->b and b->c. Therefore, it is not easy to create a u-turn instruction at b. One solution could be to pass some contextual information about the last path (a->b) into the creation of the next path (b->c). Another idea could be to merge the instructions by adding a u-turn if necessary. This could be done by adding the orientation for the last part of the edge in every last instruction and the orientation for every first instruction, so the merger could see if they are opposite and add a u-turn. If they are 90° the merger could add a regular turn instruction (might be the case when placing a waypoint on an intersection.

> A heading per instruction is not a bad idea as e.g. the very first continue after every via or start point could need something like "go north" which the client then could at least create on its own.

Yes, so I only added it for the first instruction after a waypoint, but we could also do this for every instruction (not sure if that would be needed).

boldtrn:
After thinking about this a bit more, we could also create a u-turn instruction as the first instruction if the first heading is going into the opposite way of the requested heading. This should be actually quite easy as well. This could be done in the PathMerger or in a new class "InstructionMerger"?

karussell:
Yes it could be done in PathMerger or if necessary in a new class.

devemux86:
Another idea could be to post-process the instructions after their generation.
Then knowing them all, can do such things as skip / merge / u-turn?

boldtrn:
Ok, I updated the code. 

I create a u-turn instruction for waypoints, this means if we are going from `a->b->a`, there will be u-turn instruction at `b`. I do a heading check, if the difference between the two headings is very small (currently I have 1°, which could be further improved).

I also added an initial u-turn if there was a heading for the first point and if the heading is pointing in the other direction than the heading of the first edge. (Diff > 170°)

After giving this some more thought I am not sure if it is a good idea to do this on the server side. Think about a user is parked in his garage and plans a route. The heading might be completely off at this point. Once the route is calculated he wants to start the navigation and starts riding. The first instruction would be "turn around", which is a bit weird. Probably we should add some logic on the client side for this, as the client knows the initial heading of the continue instruction.

boldtrn:
Something I just came up with could solve case 1). But this would require extra computation power. We could add the starting and end heading of every instruction. Essentially this would require 2 azimut calculations per instruction. This would allow us to check the instruction list for short instructions with repeating turn signs, like it would appear for u-turns in case 1).

E.g.

```
>----- Turn right
                |
<----- Turn right
```

boldtrn:
We also need an icon for the front end here.

devemux86:
Note that in your example, being parked and planning a route, the GPS could return `0°` location bearing, which essentially won't participate in route initial direction.
(GPS bearing would become accurate when device moves, not when it's stationary)

boldtrn:
> Note that in your example, being parked and planning a route, the GPS could return 0° location bearing, which essentially won't participate in route initial direction.
(GPS bearing would become accurate when device moves, not when it's stationary)

Yes, I had the same thought, see my comment above:

> After giving this some more thought I am not sure if it is a good idea to calculate an initial u-turn on the server side. Think about a user is parked in his garage and plans a route. The heading might be completely off at this point. Once the route is calculated he wants to start the navigation and starts riding. The first instruction would be "turn around", which is a bit weird. Probably we should add some logic on the client side for this, as the client knows the initial heading of the continue instruction.

So I think we should remove it again. Let me know what you think.

karussell:
Sorry, I'm a bit late with this ... would you mind updating the PR? probably just recreate the translation files.

> Probably we should add some logic on the client side for this, as the client knows the initial heading of the continue instruction

The heading variable is already something where the client can and should pass this information - or is this already solved (?)

> Essentially this would require 2 azimut calculations per instruction. 

Isn't one of those calculations already done in the instruction creation code?

boldtrn:
Ok, so I updated this PR.

> The heading variable is already something where the client can and should pass this information - or is this already solved (?)

Yes, this is how I implemented it :). However, I am not sure if it wouldn't simply work better to do this on the client side. Here we could also check if the user is moving or not and all that.

We could also add this to GH though. We could add a `speed` or `is_moving` variable. It the user is moving, we could create a bigger penalty for a u-turn, whereas when being parked, performing a u-turn is often easier. 

I am really not sure what we should do here? WDYT @devemux86, would you see this on the client or server side, and why?

> Isn't one of those calculations already done in the instruction creation code?

Yes, well we could add the heading for every instruction and not just for via instructions which would make this easier in the end. Probably the best would be to solve this case during the instruction creation and not in a post processing step. Will try if it's possible.

karussell:
> We could add a speed or is_moving variable

Why would this be necessary? Why not just omit the heading if it is unknown? (and then no u-turn instruction is created)

> would you see this on the client or server side, and why?

What exactly do you want to see on the client side?

> Probably the best would be to solve this case during the instruction creation and not in a post processing step. Will try if it's possible.

Sounds good!

devemux86:
> What exactly do you want to see on the client side?

Indeed what's the actual use case or do we complicate things?
GPS bearing exists after some device movement, at that point the instructions were already calculated.

boldtrn:
> Why would this be necessary? Why not just omit the heading if it is unknown? (and then no u-turn instruction is created)

Ok, maybe I was thinking too complicated then :). Let's leave it as it is currently.

> What exactly do you want to see on the client side?

The initial U-turn when being parked parallel to a street and needing to start in the other direction as currently facing. I was just thinking that it might be easier/better to do this on the client side.

devemux86:
> The initial U-turn when being parked parallel to a street and needing to start in the other direction as currently facing. I was just thinking that it might be easier/better to do this on the client side.

That could very easily fail by GPS accuracy.. :slightly_frowning_face: 
GPS can shows us some meters away in other direction, GraphHopper will retrieve that as start point, snap position to reverse lane and we'll have unwanted routing.

boldtrn:
Ok, I updated this PR. It contains 3 features:

- U-turns for Via points
- U-turns for start if heading points in the opposite direction
- U turns for turning between two separated one-way lane roads that belong to the same road, if the connection between the two is short (<35m) and is only one edge. This only works if there are instructions, like two left turns. The second instruction will be ignored and the first will be converted to a u-turn. This could be further improved when we solve #1048.

Let me know what you think.

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
```diff
@@ -196,19 +204,42 @@ public void next(EdgeIteratorState edge, int index, int prevEdgeId) {
...
+                    Note: The current approach only works if the short connector is only 1 edge, as we check the orientation,
+                    and name of the edge before the current edge. Could be extended to a second edge, however I haven't
+                    any example where we would have needed 2 edges.
+                  */
+                if (prevInstruction.getDistance() < MAX_U_TURN_DISTANCE && (sign < 0) == (prevInstruction.getSign() < 0) &&
+                        (Math.abs(sign) == Instruction.TURN_SLIGHT_RIGHT || Math.abs(sign) == Instruction.TURN_RIGHT || Math.abs(sign) == Instruction.TURN_SHARP_RIGHT) &&
+                        (Math.abs(prevInstruction.getSign()) == Instruction.TURN_SLIGHT_RIGHT || Math.abs(prevInstruction.getSign()) == Instruction.TURN_RIGHT || Math.abs(prevInstruction.getSign()) == Instruction.TURN_SHARP_RIGHT) &&
+                        InstructionsHelper.isNameSimilar(doublePrevName, name) &&
+                        // We only divide by 1.9 instead of 2, since the orientation calculation is not very exact
+                        (Math.abs(doublePrevOrientation - prevOrientation) % Math.PI > (Math.PI / 2) * .9) &&
```
karussell: Comment says '1.9' but code says '2'?

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
```diff
@@ -196,19 +204,42 @@ public void next(EdgeIteratorState edge, int index, int prevEdgeId) {
...
                     .setExited();
 
+            doublePrevName = prevName;
             prevName = name;
             prevAnnotation = annotation;
 
         } else {
             int sign = getTurn(edge, baseNode, prevNode, adjNode, annotation, name);
 
             if (sign != Instruction.IGNORE) {
```
karussell: Not really related, but we should move all instruction related code into a new package and make this constant package protected?

karussell:
And what kind of problems do you see with the current state? (Just to be sure I understand all caveats properly :))

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
boldtrn: Oops, yes I changed the logic again, but forgot to delete the comment :(

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
boldtrn: Ok, moving into a new package makes sense, will create a new issue for this. Not sure about the package protected, as I think this is needed for the clients as well?

boldtrn:
> And what kind of problems do you see with the current state? (Just to be sure I understand all caveats properly :))

IMHO this PR works already pretty good and I feel quite confident to merge it.

It's just that sometimes a u-turn instruction is missing, because we miss the second turn. This happens quite often for u-turns as they appear when turning around on a 2-lane separated road. It's not a big problem, but I would believe that some people would feel confused.

I planned a couple of routes with this PR and checked if the u-turns appeared correctly, which worked good. Give it a try :).

karussell:
Thanks, will do!

Will still have to wait ~2 week before merging, and then also merge #1075 (due to the same reasons)

karussell:
FYI: I think the instruction thing is just one side of the problem with those u-turns, in general we should be able to add the u-turn penalty. See [this example](https://graphhopper.com/maps/?point=49.930257%2C11.586864&point=49.934097%2C11.588559&locale=en&vehicle=car&ch.disable=true) and this is related to #446 where we completely want to exclude these u-turns (that then have a way as 'via' member)

devemux86:
BTW couldn't have left & right u-turns like all other "turn" instructions?

karussell:
What do you mean?

devemux86:
U-turns can be left or right depending on left/right-hand traffic rules. Wondering if we could have that information in instruction sign, instead of a common id (like with roundabouts).

karussell:
Ah, yes. For me this would make sense.

boldtrn:
> U-turns can be left or right depending on left/right-hand traffic rules. Wondering if we could have that information in instruction sign, instead of a common id (like with roundabouts).

The issue with this, that we cannot know the direction of a u-turn, created after a via-point or due to the heading, as they don't have a direction. The u-turns that are created inside `InstructionsFromEdges` have a direction and we could add a direction to them. Furthermore we could find out the direction using the SpatialRuleIndex to check if the u-turn happens in a country with left-hand traffic, if yes the u-turn usually goes to the right, otherwise it usually goes to the left. But I am not sure if this will work and if the overhead justifies this?

devemux86:
It would be a nice to have feature as u-turns (and roundabouts) come in response as generic sign / symbol / instruction.
Or to overcome that could need clients to be aware of country traffic rules where routing is performed, though that wouldn't be optimal for cross-country routes.

devemux86:
BTW I understand the technical difficulties, just wondering if extra improvements are possible. :slightly_smiling_face:

boldtrn:
> BTW I understand the technical difficulties, just wondering if extra improvements are possible. :slightly_smiling_face:

As I said, I could try to check the location using the SpatialRuleIndex to check the usual traffic side in that country. This would solve the roundabout and u-turn issue.

boldtrn:
I merged the current master into this PR. Should be mergeable now? Or should we still wait?

karussell:
What is the performance impact on the instructions? Would you mind to measure this?

boldtrn:
> What is the performance impact on the instructions? Would you mind to measure this?

The current master showed `routingCH.mean=2.60|259` (two different measurements), this pr showed `routingCH.mean=2.69|2.65|2.61` (3 different measurements), so there an impact, but it's <3%.

boldtrn:
I just slightly updated this PR. We are now more generous when it comes to slightly different orientations of the roads before and after the turn.

One example I found where the orientation change was too big for the previous check can be found [here](https://graphhopper.com/maps/?point=48.995213%2C8.383271&point=48.995333%2C8.38303&locale=en-au&vehicle=car&weighting=fastest&elevation=false&debug=true&use_miles=false&layer=OpenStreetMap). I don't think that we increased our false positive rate here (showing u-truns where there are none), because we are also assuming, that there has to be two very close turns in the same direction and that the street names have to be similar before and after. I could not find any false positive so far.

Will continue testing bike and foot and report if I find any issues.

boldtrn:
Ok, this was a bit too generous I think. [This turn](http://localhost:8989/?point=49.119104%2C8.783264&point=49.119276%2C8.782952) was recognized as u-turn.

boldtrn:
Let's wait with this PR a bit. I think I should do some more manual testing with different encoders to fine-tune this feature. Will do this soon.

boldtrn:
I updated this PR. Now I am quite happy with the results. When running measurement this PR was even a bit faster than the current master (well it shouldn't actually be faster, but the performance impact is not that huge). I updated the first comment to contain the relevant information.

karussell:
Should we differentiate left and right U-turns (and adapt the signs accordingly) like @devemux86 mentioned?

boldtrn:
We could, but I am not sure if the cost use factor makes sense.

The benefit would be to be able to create u turns in a certain direction. But I am not sure how often u-turns would be done in the "wrong" direction.

The downside of this would be that we need a spatial rule index to check if we are in a left hand traffic country. 

An alternative would be to differentiate between turn around and u-turn. U turns would have a direction and would be calculated in the instruction class. Turn around would be calculated for via points in the path merger.

I think differentiating between u turns and turn around would be a nice addition. WDYT?

boldtrn:
Btw I think "turn around" and "u turn" is a synonym, so we might need a different name :).

devemux86:
> The downside of this would be that we need a spatial rule index to check if we are in a left hand traffic country.

That would be a benefit for other cases in the library too.
Like mentioned above, if having that information, could use it also in roundabouts.

BTW is there the thought to allow setting the [rule of the road](https://en.wikipedia.org/wiki/Left-_and_right-hand_traffic) externally by the apps, instead to calculate it inside the library, or it's better to have it always inside?

boldtrn:
> Like mentioned above, if having that information, could use it also in roundabouts.

Yes true :). I am still not 100% sure if this wouldn't be easier done on the client side? Also the benefit for roundabouts is not that huge IMHO. 

I think for roundabouts it would interesting to calculate the exiting angle (we got that already, maybe we need a bit more info) and create a turn instruction image on demand. For example if we know that we exit 80° to the left and the we go right around (might be possible to calculate from the edge geometry) we could generate a perfectly matching image. Might be easy to do with on demand SVG creation? Just a crazy thought :), but that would be really cool.

> BTW is there the thought to allow setting the rule of the road externally by the apps, instead to calculate it inside the library, or it's better to have it always inside?

Isn't it enough to have it on the client side? I think besides the direction of roundabouts the changes for the routing are not that big and this could be easily done on the client side?

devemux86:
> Isn't it enough to have it on the client side?

Client can calculate or simply request users to set the rule of the road for their region.
But then need to declate it to any routing library, so that library can generate the proper instructions.

boldtrn:
I just pushed a change. We now have `U_TURN_LEFT`, `U_TURN_RIGHT` (these are calculated for u-turns that use an intermediate short connecting road in between), and `U_TURN_UNKNOWN` (which would be used for generic, please turn around instructions, e.g. at waypoints etc. where GH does not calculate the actual turn movement).

Icon wise we could now use directed u-turn icons for the specified u-turns. For the unknown we could also create a more generic turn around icone, for example like [this](https://github.com/mapbox/directions-icons/blob/master/src/png/dark/direction_roundabout.png).

WDYT?

devemux86:
> For the unknown we could also create a more generic turn around icone, for example like [this](https://github.com/mapbox/directions-icons/blob/master/src/png/dark/direction_roundabout.png).

That icon is for right-hand traffic roundabouts, not u-turns.
For a generic u-turn without known direction, the usual icon without the arrow could work.

boldtrn:
> That icon is for right-hand traffic roundabouts, not u-turns.
For a generic u-turn without known direction, the usual icon without the arrow could work.

Yes, this was just an example :). We could also just turn around the continue straight arrow, so that it points in the other direction. Another option would be a u-turn symbol without an arrow head, yes.

boldtrn:
I just updated this PR and resolved the merge conflict. This should be mergeable now?

boldtrn:
> I think afterwards we should compare measurement with 3 points, currently there are just 2 and then the PathMerger is probably not activated.

Thanks for merging this :). I couldn't follow your comment, sorry. What do you mean by that?

karussell:
In the measurement class we use 2 point for querying so there won't be any u-turn calculation. If we use 3 points we could better see the performance impact of this change.

boldtrn:
> In the measurement class we use 2 point for querying so there won't be any u-turn calculation. If we use 3 points we could better see the performance impact of this change.

Ah, ok, now I understand what you mean by that. There are still u-turns, but no "via-point u-turns". Sounds good, should I create a PR that creates multi point routes?

karussell:
Just mean to do the perf comparison in this case. Not sure if such a thing is necessary in the long run.

PR #1074, opened 2017-05-26
web js: fix array handling in method flatParameter

### LINKED ISSUE #1072: Issue with Parameter Flatting and Heading
Currently, there is a bug in JS. If you pass a heading, it is converted to an array in the `url.js#mergeParamIntoObject`. In `GHRequest#flatParameter` we recognize this as array but we run into an infinite loop, because we always pass an array into the flatParameter method again. When fixing this, the parameter gets flatted in the form `heading.0=123`. Which is not recognized by the server. 

I am not entirely sure what we should do about this. Obviously the issue was introduced due to the flattening. We could create a special rule to flat the heading differently or we could try to change the handling of the heading parameter on the server side.

The bug was introduced here: https://github.com/graphhopper/graphhopper/commit/02a10a96493b0a3918fea28cbc5efd6c3e043b8d#diff-fc1dc9e1722e4b003037f95e5f87921aR223

I think the line should be:
`url += this.flatParameter(key + "."+ keyIndex, objKey);`

But then we would pass `heading.0=123` to the server, which is not recognized.

karussell:
Fixes #1072 

Added a few tests and the fix seems to work


boldtrn:
I had the same idea to fix it. But wasn't the idea of the flat parameter method to allow parameters like pt.abc=123. If we would have arrays with pt['abc'] the parameter would be flattened to pt?

karussell:
> I had the same idea to fix it. But wasn't the idea of the flat parameter method to allow parameters like pt.abc=123?

Yes, this still works. If you specify multiple pt.abc then you create an array. See the tests.

>  If we would have arrays with pt['abc'] the parameter would be flattened to pt.

What do you mean here?

boldtrn:
> Yes, this still works. If you specify multiple pt.abc then you create an array. See the tests.

You are right :+1:. Yes this looks good now.

PR #1075, opened 2017-05-29
Enable keeps

boldtrn:
This PR enables keeps, since our clients should reliable enough now.

I currently don't really like the naming "keep left onto ROAD_NAME". I think in this case only on would be better, but this would require a special case for keeps, so I am not sure if this is worth the change, whereas the German translation with `auf` would be fine.

We also need a symbol for the front-end.


devemux86:
Well until lanes support! :slightly_smiling_face:

karussell:
> This PR enables keeps, since our clients should reliable enough now

We have to postpone this minor change as only recent versions of the clients are reliable enough.

boldtrn:
Ok, sure, no problem :).

karussell:
Ok, we can merge this here now. @boldtrn would you mind update to the latest master? Sorry for the hassle ...

boldtrn:
> Ok, we can merge this here now. @boldtrn would you mind update to the latest master? Sorry for the hassle ...

Done :)

karussell:
I think we also need images for keep :) ! 

https://graphhopper.com/maps/?point=49.464164%2C11.248755&point=49.460203%2C11.245108

boldtrn:
> I think we also need images for keep :) !

Yes I know, but I don't know where the GraphHopper images come from or how to produce them? For [Kurviger](https://kurviger.de/en?point=49.464164%2C11.248755&point=49.460203%2C11.245108) we use [this image](https://kurviger.de/img/mapbox/keep-right.png). I like it and it's CC0. We will also need a u-turn image.

devemux86:
> We will also need a u-turn image.

Note: this would probably need to be without arrow, since the instruction is generic "Make a U-turn".

karussell:
> but I don't know where the GraphHopper images come from or how to produce them

From [here](https://github.com/graphhopper/graphhopper/tree/master/web/src/main/webapp/img) created via inkscape from existing arrows (as svg converted to png)

karussell:
Have added the necessary images

karussell:
Do you have a (quick) example where a U turn is created?

boldtrn:
Any inner city one way primary or secondary should work. Place one waypoint at one side the other at the other side or look at the PathTest example.

boldtrn:
> Do you have a (quick) example where a U turn is created?

Sorry I just briefly commented yesterday from my mobile device. [This](https://graphhopper.com/maps/?point=48.640637%2C9.330654&point=48.640537%2C9.330933) would be one example. Finding them is actually quite easy, as they are usually generated in cities on a bigger primaries or secondaries.

PR #1078, opened 2017-05-29
Block area setting for all weightings

karussell:
This is a replacement for #1029 (as no feedback and want to have it in 0.9) ie. a fix for #958. Furthermore it does some cleanup in DataFlagEncoder, allows the block area option for all weightings and removes ConfigMap to improve #940 too.

It is not yet complete although all tests pass: the setGraph method is removed but necessary. I think a graph creation with the block area setting in the config will fail. We should add a test for this.


core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -847,24 +847,38 @@ public String toString() {
...
+        Result result = new Result();
+        // create per request object
+        result.speedArray = getHighwaySpeedMap(map);
+        return result;
+    }
 
-        cloneDoubleAttribute(weightingMap, cMap, GenericWeighting.HEIGHT_LIMIT, 0d);
-        cloneDoubleAttribute(weightingMap, cMap, GenericWeighting.WEIGHT_LIMIT, 0d);
-        cloneDoubleAttribute(weightingMap, cMap, GenericWeighting.WIDTH_LIMIT, 0d);
+    public class Result {
```
karussell: I like how we can keep the speedArray knowledge local to the DataFlagEncoder

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
```diff
@@ -130,54 +122,87 @@ public void fillEdgeIDs(GHIntHashSet edgeIds, Geometry geometry, EdgeFilter filt
...
                 } else {
                     throw new IllegalArgumentException(objectAsString + " at index " + i + " need to be defined as lat,lon "
                             + "or as a circle lat,lon,radius or rectangular lat1,lon1,lat2,lon2");
                 }
             }
         }
+        return blockArea;
+    }
+
+    public static class BlockArea {
```
karussell: Same here: the logic and difference of shapes and edgeIds is all local to this class and no longer in the Weighting

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
boldtrn: Why is this called result? Wouldn't be Config or RoadConfig or something similar be a better matching name?

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -847,24 +847,38 @@ public String toString() {
...
+        private double[] speedArray;
 
-        return cMap;
-    }
+        public double getSpeed(EdgeIteratorState edgeState) {
```
boldtrn: I would expect this method to check the speed encoder of the edge as well, if there is no speed stored, we return the value of speedArray.

core/src/main/java/com/graphhopper/routing/weighting/BlockAreaWeighting.java
```diff
@@ -0,0 +1,56 @@
...
+import com.graphhopper.routing.util.HintsMap;
+import com.graphhopper.storage.Graph;
+import com.graphhopper.storage.GraphEdgeIdFinder;
+import com.graphhopper.storage.index.LocationIndex;
+import com.graphhopper.util.EdgeIteratorState;
+
+/**
+ * This weighting is a wrapper for every weighting to support block_area
+ */
+public class BlockAreaWeighting implements Weighting {
```
boldtrn: Is there are reason to not use the `AbstractAdjustedWeighting`?

core/src/main/java/com/graphhopper/routing/weighting/GenericWeighting.java
```diff
@@ -48,40 +40,30 @@
     protected final long headingPenaltyMillis;
     protected final double maxSpeed;
     protected final DataFlagEncoder gEncoder;
-    protected final double[] speedArray;
+    protected final DataFlagEncoder.Result result;
     protected final int accessType;
-    protected final int eventuallAccessiblePenalty = 10;
+    protected final int eventualAccessiblePenalty = 10;
```
boldtrn: I think naming this `uncertainAccessiblePenalty`or something like this would be better. Eventual is a false friend to the German eventuell. I think this was introduced by me some time ago :), but since I just saw it, we should fix it.

core/src/main/java/com/graphhopper/routing/weighting/GenericWeighting.java
```diff
@@ -48,40 +40,30 @@
...
-        speedArray = gEncoder.getHighwaySpeedMap(cMap.getMap("highways", Double.class));
-        double tmpSpeed = 0;
-        for (double speed : speedArray) {
-            if (speed > tmpSpeed)
-                tmpSpeed = speed;
-        }
-        if (tmpSpeed > encoder.getMaxPossibleSpeed())
-            throw new IllegalArgumentException("Speed bigger than maximum speed: " + tmpSpeed + " > " + encoder.getMaxPossibleSpeed());
+        result = encoder.createResult(hintsMap);
+        double biggestSpecifiedSpeed = result.getMaxspeed();
```
boldtrn: Maybe we should rename the getMaxspeed method also to biggestSpecifiedSpeed?

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
boldtrn: This is nice :+1:

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
karussell: The naming is indeed suboptimal it should be more like RequestConfig ... as this class is probably always used via DataFlagEncoder.XY the naming could be done with this association in mind

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
karussell: > check the speed encoder of the edge as well

What do you mean here? There is no speed for DataFlagEncoder. Only a max speed value.

core/src/main/java/com/graphhopper/routing/weighting/BlockAreaWeighting.java
karussell: :+1: :)

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
boldtrn: Yes sorry, this is what I meant. I would have expected the same result as when calling getSpeed in a different Encoder.

karussell: So, what you mean is that if there is no request value for the speed we should use the default maxspeed value of the highway?

boldtrn: This is just what I would have expected when reading the method head. The getSpeed method of the other encoders reads the speed that is saved in the edge. So here I would have expected that there is a check, if maxSpeed stored -> return maxSpeed, else return speed for highway type.

karussell: I think we need to adjust the createResult method and use `Math.min(requestSpeedValue, storedSpeedValue)`. So the provided request parameters should be able to reduce the stored maxspeed values.

karussell: Let us do this in a later issue as this was so before

boldtrn: Ok no problem, it always bothered me ;).

karussell: BTW: this seems to be already implemented in GenericWeighting.calcMillis

karussell:
The problem with the default block_area is that we would have to lazy load the GraphEdge finding stuff too and not only the setGraph. Because the weighting should be read very early but the graph and locationIndex might be still uninitialized. So I have decided to remove this for now.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
boldtrn: I know, I am just used to have the getSpeed in the FlageEncoder ;).

karussell:
> Still a bit unfortunate.

Yes. But better a smaller feature now than a bigger or more features later (?)

PR #1079, opened 2017-05-30
Improve SpatialRuleLookup

### LINKED ISSUE #1077: SpatialRuleLookupArray.lookupRule throws exception
if exact == true and isBorderTile then the empty rule container passes  [this if clause](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java#L108-L113) and we throw an exception for rule.getBorders() [here](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupArray.java#L116)

Is there a `return SpatialRule.EMPTY;` missing in the mentioned if clause?

boldtrn:
Fixes #1077.

I extended the changes made by @karussell, mostly by adding tests and also fixing the creation of landmarks. There was an issue that the map.geo.json file was not formatted correctly.

The only thing that is missing right now is an IT that tests the execution of `GraphHopperModule.createGraphHopper -> createSpatialRule`. Not sure how we want to do this, though?

Other from that the change looks good and mergeable to me.


karussell:
I have updated the branch...

web/src/test/java/com/graphhopper/http/GraphHopperLandmarksIT.java
```diff
@@ -20,49 +20,77 @@
...
+    @Before
+    public void setUp() {
         Helper.removeDir(new File(DIR));
-        shutdownJetty(true);
     }
 
-    @Before
-    public void setUp() {
+    @Test
+    public void testSimpleQuery() throws Exception {
```
boldtrn: This query is really nothing special, we can remove it and save some time by only using your method, I don't think it adds more value than yours.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -108,8 +108,8 @@ public String toString() {
             }
         };
 
-        // later make edge base working! Not really necessary as when adding turn costs while routing we can still
-        // use the node based traversal as this is a smaller weight approximation
+        // Edge based is not really necessary because when adding turn costs while routing we can still
+        // use the node based traversal as this is a smaller weight approximation and will still produce correct results
```
boldtrn: Not sure if I get your comment right. You are stating that during the creation of LM the traversal mode is not relevant, only during routing, as we only forbid certain turns?

boldtrn:
I updated the IT, so it only uses the belarus osm file. I moved your working LM query to a different method and only kept the disconnected graphs test in the existing method. Tests already need quite long to run, so we don't need to import LM two times. The branch looks good to me now and I think we can merge it.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
karussell: Yes. We can rephrase this of course...

karussell:
Thanks!

karussell:
> Tests already need quite long to run, so we don't need to import LM two times. 

Nice idea!

PR #1081, opened 2017-05-30
Update GraphHopper in Android and dependencies

devemux86:


karussell:
Thanks!

(We should then also use 1.7.25 for the slf4j in the pom.xml ...)

devemux86:
> (We should then also use 1.7.25 for the slf4j in the pom.xml ...)

Indeed, I added the change in this PR.

PR #1085, opened 2017-06-02
avoid mixing CH & LM for now

### LINKED ISSUE #1082: Mixing of CH and LM should be avoided
If LM and CH are enabled but LM is not allowed being disabled per request then LMAlgoFactoryDecorator.getDecoratedAlgorithmFactory does not return the defaultFactory even for CH routing requests. This means that we use the currently slower LM+CH combination for this configuration.

karussell:
fixes #1082

### DIFF
--- a/web/src/test/java/com/graphhopper/http/GraphHopperLandmarksIT.java
+++ b/web/src/test/java/com/graphhopper/http/GraphHopperLandmarksIT.java
@@ -18,6 +18,7 @@
 package com.graphhopper.http;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.graphhopper.GraphHopper;
 import com.graphhopper.util.CmdArgs;
 import com.graphhopper.util.Helper;
 import org.junit.AfterClass;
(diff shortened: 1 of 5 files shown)


web/src/test/java/com/graphhopper/http/GraphHopperLandmarksIT.java
```diff
@@ -70,19 +72,19 @@ public void testSimpleQuery() throws Exception {
...
-        JsonNode json = query("point=55.99022,29.129734&point=56.007787,29.208355", 400);
+        JsonNode json = query("point=55.99022,29.129734&point=56.007787,29.208355&ch.disable=true", 400);
         JsonNode errorJson = json.get("message");
         assertTrue(errorJson.toString(), errorJson.toString().contains("Different subnetworks"));
 
-        // without lm it should work
-        json = query("point=55.99022,29.129734&point=56.007787,29.208355&lm.disable=true", 200);
+        // without landmarks it should work
+        GraphHopper hopper = getInstance(GraphHopper.class);
+        hopper.getLMFactoryDecorator().setDisablingAllowed(true);
```
karussell: Important that the simple case has LM disallowed==false

PR #1091, opened 2017-06-17
Add Average Speed as PathDetails to Response

### LINKED ISSUE #439: Add speed, time and more informations to instructions

boldtrn:
Fixes #439.

I added the possibility to view details of the calculated as proposed in #439. It should be very easy to extend so we can add different details. 

The calculation of the PathDetails is a bit similar to Instructions and maybe in the future we might find that it makes sense to merge them, but not now.

I haven't measured the performance yet, but I can do this if you like the general approach.

Simplification is disabled when adding PathDetails right now.

Right now I don't like two things:
- The data structure, not an issue right now, but maybe if we output osm ids for continental requests? Also see the conversion in the SimpleRouteSerializer. I think, wrapping the PathDetails in a Map again, might be more efficient. Overall it is a bit too much map usage IMHO, but it seems like the right data structure ;).
- The initialization happens in the GraphHopper class, right now not an issue, but once we have more details, we might want to move this somewhere else. 

Future improvements:
Change the way the Instructions are generated so that we generate the intervals for the instructions in the PathMerger. Then we could try to simplify points between PathDetails/Instructions without simplifying over the bounds of any of the two.


core/src/main/java/com/graphhopper/util/details/PathDetailsFromEdges.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+        // We need this in the finish() step
+        adjNode = edge.getAdjNode();
+    }
+
+    private int findMatchingPointIndex(int baseNode, int pointIndex) {
+        lat = nodeAccess.getLat(baseNode);
+        lng = nodeAccess.getLon(baseNode);
+        for (; pointIndex < points.size(); pointIndex++) {
+            if (lat == points.getLat(pointIndex) && lng == points.getLon(pointIndex)) {
```
devemux86: Is it safe comparing coordinates without epsilon?

boldtrn: It should be, since they come from the same source? I could add a delta, though. But it's good that you remind me, I wanted to check how the algorithm behaves for loops etc.

karussell:
Thanks!

> Check how the algorithm behaves for loops, dead ends, etc.

Maybe just run 100K routes for Germany via the Measurement class :) ?

karussell:
> Also see the conversion in the SimpleRouteSerializer

It might be interesting here or in a separate issue to get rid of this class entirely as we now could let jackson do this serialization

boldtrn:
> Maybe just run 100K routes for Germany via the Measurement class :) ?

Ah, I meant if the correct nodes are matched, just because at a crossing of the route it could be point=50 on the first way but point 270 on the way back, with waypoints we start the counting from the beginning for every path, but after every waypoint a new path starts, but we need to consider the points from the start to the end of the path to calculate the correct index, so maybe we have to remember the last index or something like that (not sure if this is understandable, not easy to explain without code)

For example a route like this: https://graphhopper.com/maps/?point=49.368066%2C9.703674&point=49.396452%2C9.753628&point=49.375332%2C9.75071&point=49.38606%2C9.683933

When comparing only coordinates we might produce incorrect results when the routes cross.

> It might be interesting here or in a separate issue to get rid of this class entirely as we now could let jackson do this serialization 

:+1:

karussell:
Sorry didn't yet look into the source code. Why do we need coordinate comparison?

boldtrn:
> Sorry didn't yet look into the source code. Why do we need coordinate comparison?

Not sure if we need it, I used this to calculate the interval, but now that you mention it, we could probably also count the number of points per edge and just sum it up.

karussell:
Yes, please :). To make this then more secure we need probably the same tests that you mention but matching is only good if we have no other choice

boldtrn:
I updated the code, it's a lot better now, compared to coordinate comparing. I added a rather ugly js debugging code to check if the intervals are set correctly, see the attached screenshot:

path_details

boldtrn:
I updated this PR so that the PathDetails are now stored similar to the instructions, so every PathDetail knows the number of points it contains. The benefit here is that when we run the simplification we don't have to update every PathDetail, but only the ones that are actually changed. 

When serializing I run a similar approach as the instruction to create actual intervals.

I saw a "minor issue" close to waypoints. Sometime the start of the PathDetails move slightly when the waypoints are placed close to the beginning of an edge, where the PathDetails would change. Not sure what's the reason, but I think it could have something to do with VirtualEdges, but haven't checked, since I am not sure if it's worth the trouble, see the attached screenshots.

pathdetails4
pathdetails5

The route is in Baden-Württemberg: http://localhost:8989/?point=48.596365%2C8.634396&point=48.586398%2C8.618098&point=48.578155%2C8.609955&details.average_speed=true&debug_details=true

[This is the OSM way](http://www.openstreetmap.org/way/23456459).

karussell:
Thanks! Will have a look

> I saw a "minor issue" close to waypoints

I think it is important to understand the issue to judge if we should fix it or not.

web/src/main/java/com/graphhopper/http/SimpleRouteSerializer.java
```diff
@@ -100,6 +101,16 @@ private String getMessage(Throwable t) {
...
                     jsonPath.put("legs", ar.getLegs());
 
+                    List<PathDetails> details = ar.getPathDetails();
+                    if (!details.isEmpty()) {
+                        //TODO I don't really like the conversion, should we keep it like that in the PathWrapper?
+                        Map<String, Map<Object, List<int[]>>> pathDetailsMap = new HashMap<>();
+                        for (PathDetails pathDetails : details) {
+                            pathDetailsMap.put(pathDetails.getName(), pathDetails.getPathDetailsMap());
+                        }
+                        jsonPath.put("details", pathDetailsMap);
```
karussell: Can't we use a Map when returning from ar.getPathDetails so that we could write jsonPath.put("details", ar.getPathDetails())? Or annotate PathDetails so that we get the JSON output that we want via jackson.

core/src/main/java/com/graphhopper/util/details/PathDetailsFromEdges.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * This class calculates the PathDetails, similar to the instruction calculation
+ *
+ * @author Robin Boldt
+ */
+public class PathDetailsFromEdges implements Path.EdgeVisitor{
+
+    int i;
+    PathDetailsCalculator calc;
```
karussell: Would you mind to move everything inside the loop? This reads confusing.

core/src/main/java/com/graphhopper/util/details/PathDetailsFromEdges.java
```diff
@@ -0,0 +1,66 @@
...
+        this.calculators = calculators;
+    }
+
+    @Override
+    public void next(EdgeIteratorState edge, int index, int prevEdgeId) {
+        for (i = 0; i < calculators.size(); i++) {
+            calc = calculators.get(i);
+            if (calc.edgeIsDifferentToLastEdge(edge)) {
+                details.get(i).endInterval(numberOfPoints);
+                details.get(i).startInterval(calc.getCurrentValue());
```
karussell: It reads a bit strange that the index is used for calculators and details. Can we somehow assign one calculator to the PathDetails?

Also can we avoid stateful stuff like with the .reset call? (if possible at all)

core/src/main/java/com/graphhopper/util/details/PathDetails.java
```diff
@@ -0,0 +1,129 @@
...
+            pathDetails.add(this.currentDetail);
+        }
+        isOpen = false;
+    }
+
+    public List<Detail> getDetails() {
+        return pathDetails;
+    }
+
+    public void merge(PathDetails pD) {
```
karussell: Why do we need the merge method?

core/src/main/java/com/graphhopper/util/details/AverageSpeedDetails.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public Object getCurrentValue() {
+        return curAvgSpeed;
+    }
+
+    @Override
+    public String getName() {
+        return "average-speed";
```
karussell: Potentially we should use the same name `average_speed` like we use for the parameter in the URL?

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -86,8 +95,8 @@ public void doWork(PathWrapper altRsp, List<Path> paths, Translation tr) {
                     }
 
                     for (Instruction i : il) {
+                        origPoints += i.getPoints().size();
```
karussell: Was this a bug in the previous state of code?

core/src/main/java/com/graphhopper/routing/Path.java
```diff
@@ -372,6 +371,20 @@ public InstructionList calcInstructions(final Translation tr) {
...
 
+    /**
+     * @param calculators All calculators to be considered for this route
+     * @return List of PathDetails for this Path
+     */
+    public List<PathDetails> calcDetails(final List<PathDetailsCalculator> calculators) {
+        List<PathDetails> details = new ArrayList<>(calculators.size());
+        for (PathDetailsCalculator calc : calculators) {
+            details.add(new PathDetails(calc.getName()));
+            calc.reset();
```
karussell: Why do we have to call reset for the initialization?

core/src/main/java/com/graphhopper/PathWrapper.java
```diff
@@ -243,6 +246,23 @@ public void setInstructions(InstructionList instructions) {
...
+            this.pathDetails = details;
+        }else{
+            if(this.pathDetails.size() != details.size()){
+                throw new IllegalStateException("Details have to be the same size");
+            }
+            for (int i = 0; i < details.size(); i++) {
+                this.pathDetails.get(i).merge(details.get(i));
+            }
+        }
+    }
```
karussell: See below: what is a 'merge'? And why not just adding PathDetails like the method says it?

Also instead of EMPTY_LIST and if+else I would prefer without the if and always do the same.

web/src/main/java/com/graphhopper/http/SimpleRouteSerializer.java
boldtrn: Yes, makes sense :+1:

core/src/main/java/com/graphhopper/util/details/PathDetailsFromEdges.java
boldtrn: Don't you think that it makes sense to have calculator per detail? Essentially every calculator handles one PathDetail. So if you want to add a PathDetail in the Response, you have to pass the calculator to the PathMerger.

karussell: But why should we use two independent lists? Why not e.g. setting the calculator as variable of PathDetails?

boldtrn: > Also can we avoid stateful stuff like with the .reset call? (if possible at all)

I don't like the reset call either. Mhm, we could create a new calculator for every Path. However, I think it makes sense in this case that the Calculator has a state and that we don't have to keep the state outside of the calculator. This allows us to handle complex states, using multiple variables in the calculator, but we don't have to care about it from the outside. WDYT?

karussell: Why not create fresh calculators for every Path?

boldtrn: > Why not create fresh calculators for every Path?

On the other hand, why creating the overhead :)?

I think we could create a better setup here. In the GraphHopper.java class we could create a PathDetailsFactory that is initialized with the GHRequest. It reads the request and can create fresh calculators on request?

core/src/main/java/com/graphhopper/util/details/PathDetails.java
boldtrn: To merge the PathDetails for every Path. We could also do this completely in the PathMerger class. At first I created it because PathDetails were a map and it made a lot of sense then. With the current code it could be removed and done in the PathMerger.

core/src/main/java/com/graphhopper/util/PathMerger.java
boldtrn: Yes. Well it was only used for logging, so it was not really bad, also it only appeared when having multiple Paths. I thought about creating an issue for it but then I thought it's not worth to create an issue for this.

boldtrn:
> I think it is important to understand the issue to judge if we should fix it or not.

Ok, so I tried to dig into this issue a bit. I compared these two routes:
http://localhost:8989/?point=48.58802%2C8.618581&point=48.586491%2C8.617648&point=48.580355%2C8.606071&details.average_speed=true&debug_details=true
http://localhost:8989/?point=48.58802%2C8.618581&point=48.586469%2C8.61798&point=48.580355%2C8.606071&details.average_speed=true&debug_details=true

I understood the issue now. There is one point in the resulting points that is reserved for the via point that cannot be calculated via sumarizing points of edges but has to be plainly added if there is a via point.

boldtrn:
I updated this PR. I added a way to simplify the Path for PathDetails+Instructions. So the simplification keeps the points in between the different Instructions and PathDetails. 

Optically, when using the Debugging view, the results look correct. I can also see that there is a difference of the response size if I pass `way_point_max_distance=0` in the URL (getting bigger, as no actual simplification is triggered). The location of the average speed as well as instructions nodes look correct.

I would have expected that the GHResponse contains the same amount of points as before if it only contains instructions. This is not true right now. We have less points, I am not exactly sure if this indicates an issue or not. I haven't found the reason yet.

This is also the reason why the tests are failing right now.

karussell:
Cool - thanks!

> This is not true right now. We have less points, I am not exactly sure if this indicates an issue or not. I haven't found the reason yet.

There might be bugs or sub-optimalities before but maybe you find the time to compare one test to see where the difference is coming from exactly?

BTW:  I'd really love to have this already there :D This would make debugging the routing so much easier.

core/src/main/java/com/graphhopper/util/Instruction.java
```diff
@@ -141,6 +141,10 @@ public PointList getPoints() {
         return points;
     }
 
+    public void setPoints(PointList points) {
+        this.points = points;
```
karussell: Can we avoid this setter?

boldtrn: Not easily. The issue is that we have to change the points, once the instruction is created. Changing this might be complicated, but not necessarily impossible.

Once the instructions and PathDetails are calculated, we do the simplification. When simplifying, we need to change the points. I think changing this would require quite a lot overhead.

boldtrn:
> BTW: I'd really love to have this already there :D This would make debugging the routing so much easier.

Well I mean it already works, so you could use this branch for debugging.

 There might be bugs or sub-optimalities before but maybe you find the time to compare one test to see where the difference is coming from exactly?

Sure, I will debug the code and check where the differences might come from.

boldtrn:
Ok, I further debugged the issue. I turned out that I simplified the points between the start of the current instruction, to the start of the next instruction, whereas the old code simplified the points from the start of the instruction to the end of the instruction.

The tests could be easily fixed by simply adding a -1 in the PathSimplify class. On the other hand I am currently thinking, why we shouldn't keep it like that. I cannot see a downside right now, can you? This produces smaller results while maintaining the important points. We have to run a couple of tests and think about this very well though.

WDYT?

karussell:
Thanks for investigating! and this seems reasonable. Will investigate and think about this.

core/src/main/java/com/graphhopper/util/details/PathDetails.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Holds the details for a path
+ *
+ * @author Robin Boldt
+ */
+public class PathDetails {
```
karussell: If we associate the PathDetails with the PDCalc I think we then only need a `List<Detail>` in the PDCalc and this class here gets kind of a builder for this `List<Detail>` which would also better separate the concerns IMO. Do you know what I mean :) ?

core/src/main/java/com/graphhopper/util/details/PathDetailsCalculator.java
```diff
@@ -0,0 +1,38 @@
...
+import com.graphhopper.util.EdgeIteratorState;
+
+/**
+ * Calculate details for a path
+ *
+ * @author Robin Boldt
+ */
+public interface PathDetailsCalculator {
+
+    boolean edgeIsDifferentToLastEdge(EdgeIteratorState edge);
```
karussell: Java naming: edgeIsDifferentToLastEdge -> isEdgeDifferentToLastEdge

karussell:
> I cannot see a downside right now, can you? This produces smaller results while maintaining the important points. We have to run a couple of tests and think about this very well though.

ok

> BTW: Interestingly, the altitude of the Result in one test changed. Still not to check what happened here :)

here again the real causes would be important to understand :)

core/src/main/java/com/graphhopper/util/details/PathDetails.java
```diff
@@ -0,0 +1,135 @@
...
+        this.currentDetail = new Detail(value);
+        isOpen = true;
+    }
+
+    /**
+     * Ending intervals multiple times is safe, we only write the interval if it was opened.
+     * <p>
+     * Writes the interval to the pathDetails
+     *
+     * @param numberOfPoints Length of the PathDetail
```
boldtrn: Because we end all intervals in the finish() call of the EdgeVisitor. It was just convenient. Could be solved differently.

core/src/main/java/com/graphhopper/util/details/PathDetails.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public void merge(PathDetails pD) {
+        if (!this.name.equals(pD.getName())) {
+            throw new IllegalArgumentException("Only PathDetails with the same name can be merged");
+        }
+        List<Detail> otherDetails = pD.getDetails();
+
+        // Make sure that pathdetails are merged correctly at waypoints
+        if(!this.pathDetails.isEmpty() && !otherDetails.isEmpty()){
+            Detail lastDetail = pathDetails.get(pathDetails.size()-1);
```
boldtrn: Sorry I can add them. IntelliJ colors them differently so I started to not write the `this.` anymore :). You are right I should probably add them.

core/src/main/java/com/graphhopper/util/details/PathDetails.java
```diff
@@ -0,0 +1,135 @@
...
+
+        for (Detail detail : this.pathDetails) {
+            List<int[]> detailIntervals;
+            if (detailsMap.containsKey(detail.value)) {
+                detailIntervals = detailsMap.get(detail.value);
+            } else {
+                detailIntervals = new ArrayList<>();
+            }
+            detailIntervals.add(new int[]{pointer, pointer + detail.numberOfPoints});
+            detailsMap.put(detail.value, detailIntervals);
```
boldtrn: Yes :). Putting an existing element should be O(1), so I didn't bother. Can change this though.

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private int getLengthIL(InstructionList il) {
+        return il.size();
+    }
+
+    private int getLengthPD(PathDetails pd) {
+        return pd.getDetails().size();
+    }
+
+    private int getLength(Object o) {
```
boldtrn: Yes I had the same idea. Not sure if it's worth the troubles. Should I try if it's possible?

web/src/main/webapp/js/instructions.js
```diff
@@ -91,6 +89,27 @@ module.exports.create = function (mapLayer, path, urlForHistory, request) {
...
+                        icon: L.icon({
+                            iconUrl: './img/marker-small-blue.png',
+                            iconSize: [15, 15]
+                        }),
+                        draggable: true
+                    }).addTo(mapLayer.getRoutingLayer()).bindPopup(detailKey+":"+k);
+                }
+            }
+        }
+    }
```
boldtrn: Does it matter, since it's only used for debugging? I can try, but I think it was easier to also have the text etc.

core/src/main/java/com/graphhopper/PathWrapper.java
```diff
@@ -243,6 +244,26 @@ public void setInstructions(InstructionList instructions) {
...
+    public void addPathDetails(List<PathDetails> details) {
+        if (!this.pathDetails.isEmpty() && !details.isEmpty() && this.pathDetails.size() != details.size()) {
+            throw new IllegalStateException("Details have to be the same size");
+        }
+        for (int i = 0; i < details.size(); i++) {
+            PathDetails pd;
+            if (this.pathDetails.containsKey(details.get(i).getName())) {
+                pd = this.pathDetails.get(details.get(i).getName());
+                pd.merge(details.get(i));
+            } else {
```
boldtrn: Why would this be the case?

core/src/main/java/com/graphhopper/util/details/PathDetails.java
karussell: I mean: why is it allowed to end an interval twice? When could this happen except when this is a bug and should be thrown as an exception?

core/src/main/java/com/graphhopper/util/details/PathDetails.java
karussell: > IntelliJ colors them differently so I started to not write the this

for me it was just a bit confusing that you sometimes use with and sometimes without `this.` resulting in a different coloring when it should not :)

core/src/main/java/com/graphhopper/util/details/PathDetails.java
karussell: (depends on the `Map` implementation ;))

core/src/main/java/com/graphhopper/util/PathSimplification.java
karussell: Maybe in a separate PR (?) But the goal should be to make 'instructions' identical to any other path details.

web/src/main/webapp/js/instructions.js
karussell: Hmmh, we either can have a compact way for path details which won't be a GeoJSON (current state) but probably it would be nice to have something which is easy to display directly as a GeoJSON. Similar to what is possible for the route geometry with points_encoded. But probably also for a different PR

core/src/main/java/com/graphhopper/PathWrapper.java
karussell: Ups no, forget about it :) but I think there will be either all merge or all put operations?

BTW: here is also no need to call pathDetails.put for the merge case

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -102,17 +115,23 @@ public void doWork(PathWrapper altRsp, List<Path> paths, Translation tr) {
...
                     douglasPeucker.simplify(tmpPoints);
                 }
+                */
                 fullPoints.add(tmpPoints);
             }
+            if ((calcPoints || enableInstructions) && calculatorFactory != null) {
+                altRsp.addPathDetails(path.calcDetails(calculatorFactory));
+            }
 
             allFound = allFound && path.isFound();
```
boldtrn: Sure, just had it to test a couple of things

boldtrn:
> here again the real causes would be important to understand :)

Just debugged this. Before, we called this after we simplified the route, now we call this before we simplify the route :). More points, result in more elevation diff. Not sure if this is good, but might be a bit more accurate?

>I cannot see a downside right now, can you? This produces smaller results while maintaining the important points. We have to run a couple of tests and think about this very well though.

@karussell what is your opinion here? Should we do this?

core/src/main/java/com/graphhopper/util/details/PathDetailsCalculator.java
boldtrn: Woops, you are right, fixed it.

core/src/main/java/com/graphhopper/util/details/AverageSpeedDetails.java
```diff
@@ -0,0 +1,60 @@
...
+    private double curAvgSpeed = -1;
+
+    public AverageSpeedDetails(FlagEncoder encoder) {
+        this.encoder = encoder;
+    }
+
+    @Override
+    public boolean edgeIsDifferentToLastEdge(EdgeIteratorState edge) {
+        if(encoder.getSpeed(edge.getFlags()) != curAvgSpeed){
+            curAvgSpeed = encoder.getSpeed(edge.getFlags());
```
boldtrn: Yes, that's true :+1:

core/src/main/java/com/graphhopper/PathWrapper.java
boldtrn: > here is also no need to call pathDetails.put for the merge case

No, if the element does not exist, we have to put it. If it exists, then we have to merge it? I removed the unnecessary put for the merge case.

core/src/main/java/com/graphhopper/routing/Path.java
```diff
@@ -372,6 +372,24 @@ public InstructionList calcInstructions(final Translation tr) {
...
+     */
+    public List<PathDetails> calcDetails(final PathDetailsCalculatorFactory calculatorFactory) {
+        List<PathDetailsCalculator> calculators = calculatorFactory.createPathDetailsCalculator();
+        if(calculators.isEmpty()){
+            return Collections.EMPTY_LIST;
+        }
+        List<PathDetails> details = new ArrayList<>(calculators.size());
+        for (PathDetailsCalculator calc : calculators) {
+            details.add(new PathDetails(calc.getName()));
+            calc.reset();
```
boldtrn: Good idea, will do this :)

core/src/main/java/com/graphhopper/util/details/PathDetails.java
boldtrn: In `PathDetailsFromEdges.finish()` we end every interval. I think this is the correct behavior. Don't you think?

core/src/main/java/com/graphhopper/util/details/PathDetails.java
boldtrn: Oh ok, well it doesn't matter in intelliJ, will try to think about this from now on.

core/src/main/java/com/graphhopper/util/details/PathDetailsFromEdges.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public void next(EdgeIteratorState edge, int index, int prevEdgeId) {
+        for (int i = 0; i < calculators.size(); i++) {
+            PathDetailsCalculator calc = calculators.get(i);
+            if (calc.edgeIsDifferentToLastEdge(edge)) {
+                details.get(i).endInterval(numberOfPoints);
+                details.get(i).startInterval(calc.getCurrentValue());
+                numberOfPoints = 0;
+            }
+        }
```
boldtrn: I think handling the PathDetails inside the PathDetailsCalculator is a good idea, I just changed that, the code looks cleaner due to that :+1:. I am not yet sure about the DetailsBuilder. Could you explain why we would benefit from that? The current PathDetails are already kind of a Builder that build a Map in the end.

web/src/main/webapp/js/instructions.js
boldtrn: But we would build this ourselves in JavaScript? If yes, then let's create an issue for this and think about it later :). If we would want to change the PathDetails structure in the response, we should discuss it before merging this PR?

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -0,0 +1,195 @@
...
+        // The offset of already included points
+        int[] offset = new int[toSimplify.size()];
+        int[] endIntervals = new int[toSimplify.size()];
+        // All start at 0
+        int[] startIntervals = new int[toSimplify.size()];
+
+        int nonConflictingStart;
+        int nonConflictingEnd;
+        int toSimplifyIndex;
+        int toShiftIndex;
```
boldtrn: Ok, makes sense :)

core/src/main/java/com/graphhopper/util/PathSimplification.java
boldtrn: Created a new issue for this #1121

core/src/main/java/com/graphhopper/util/details/PathDetails.java
boldtrn: As mentioned below, I cannot see the benefit or maybe I don't understand the concept. Would you mind explaining this to me?

karussell: PathDetails currently does two things: holds a List (pathDetails) and manages start&end of intervals. I know this is a common pattern in GH but we should step-by-step migrate to a cleaner separated approach (model properties & builder).

So here I was proposing to move out the `List<Details>` and rename PathDetails to e.g. PathDetailsBuilder (that produces the `List<Details>` on calling `finish`?) or something.

core/src/main/java/com/graphhopper/util/details/PathDetails.java
karussell: Yes, this is correct. My question is related to a different thing: we should throw an exception inside endInterval if !isOpen instead of silently accepting it.

web/src/main/webapp/js/instructions.js
karussell: Ok. Thought about this. The response is fine, maybe we add later a GeoJSON compatible conversion as possibility. But what I would like to change is the way we request the details, it should not require 2 parameters and instead we should either have a semicolon-based list like `details=average_speed;something_else` like we allow for `blocking_area` or just `details=average_speed&details=something_else` like e.g. for the point array.

core/src/main/java/com/graphhopper/util/details/PathDetailsFromEdges.java
karussell: See my comment above

karussell:
Uhm, did I really lost all the comments I made. WTF github ...

core/src/main/java/com/graphhopper/util/details/PathDetails.java
boldtrn: I think you are right. I was always thinking about something like a ford PathDetails. My idea was that it gets only opened when there is a ford, which is wrong, it would be mostly open with ford=false.

boldtrn: Ah, I just found another case, this also happens on the first call of `public void next(EdgeIteratorState edge, int index, int prevEdgeId) {` in `PathDetailsFromEdges`. We first end the interval before starting it.

web/src/main/webapp/js/instructions.js
boldtrn: What do you mean by 2 parameters?

Requesting the details works like this: `path.average_speed=true`. The reason for this is, that it works well with putting this into the hints. If we want something like that, we would need to add it as an actual field to the GHRequest. I thought this might not be worth the overhead, I can do this thought if you like?

boldtrn:
> What exactly do you mean with 'this'? Some elevation approximization?

Oh sorry, I should have clarified this a bit more :). I was talking about `PathMerger.calcAscendDescend`. This was called after we simplified the route. Now it's called before we simplify the route.

boldtrn:
> Yes, sounds better :)

Ok, I updated the simplification. I tested the actual impact on the response size, but due to the encoded points there is no real impact on the response size.

boldtrn:
For some reason I cannot reply to this comment anymore, so I will reply here.

> Understood. So the setter is more explicit ... it might be uglier but more efficient to do an inplace modification of the PointList - we have to compare to understand how much this would be

Right now we use the whole PointList for the simplification and ignore the points of the Instructions, we essentially just set them afterwards, so the setter made a lot of sense for me.

Wouldn't it make more sense to remove the PointList from the Instruction and only store the number of points?

There are only 3 uses, that I could find, of the points in an Instruction. There are two usages in the GTFS module, not sure if we could remove them, but probably yes. The `InstructionList.find` method. Should be possible to do this otherwise, maybe providing the PointList for the whole route as a parameter, if we know the number of points in the Instruction it should be easy to get the actual points.

BTW: This would be also against our decision to move the Instructions closer to the PathDetails, don't you think?

karussell:
> Wouldn't it make more sense to remove the PointList from the Instruction and only store the number of points?

Yes, or keeping it as a convenient API but propagate to the full PointList?

> BTW: This would be also against our decision to move the Instructions closer to the PathDetails, don't you think?

You mean removing the pointlist?

web/src/main/webapp/js/instructions.js
karussell: > What do you mean by 2 parameters?

You wrote above that I need two parameters details.average_speed=true and debug_details=true? 

And I would like to have it as a main parameter and array, so without any prefix and without the 'true': `details=average_speed`. The JS client then needs a slight adjustment to skip 'details' and not just 'point' at a certain place.

core/src/main/java/com/graphhopper/util/details/PathDetails.java
karussell: > We first end the interval before starting it

Why would we do this? (Just trying to understand it.) For me it is not logically to end something twice and so it is safer to throw an exception in this case.

web/src/main/webapp/js/instructions.js
boldtrn: I updated this PR, you can now request path details like: `path_details=average_speed&path_details=something_else`. The parameter `debug_details=true` is only used locally to show the results of the path details similar to the parameter `debug_instructions=true`, that shows turn instructions. This parameter is still required :).

boldtrn:
> Yes, or keeping it as a convenient API but propagate to the full PointList?

Yes makes sense.

> You mean removing the pointlist?

No, this is fine and actually beneficial for that IMHO. I was talking about your request to create a logic to update the Instruction PointList in place.

core/src/main/java/com/graphhopper/util/Parameters.java
```diff
@@ -191,9 +191,9 @@
      */
     public static final class DETAILS {
 
-        private static final String DETAILS_PREFIX = "details.";
```
karussell: Ok, I think we should keep the variable 'DETAILS_PREFIX' as we do similar for the others.

The actual string `path_details` vs. `details` is harder to decide as path_details is longer but would be align with the Java class. I'm still preferring just `details` as this is directly attached to a path and also it is not clear if the class PathDetails remains or if it is replaced by `List<Details>` and `PathDetailsBuilder`. But the most important thing is: if we name the parameter `x` then the json entry should be also `x`

karussell:
> I was talking about PathMerger.calcAscendDescend. This was called after we simplified the route. Now it's called before we simplify the route.

Yes, this sounds like an improvement :)

> I was talking about your request to create a logic to update the Instruction PointList in place

I see. But if you'll be successful to use 'instructions.pointlist' just as a getter then the setter will disappear?

core/src/main/java/com/graphhopper/util/Parameters.java
karussell: Also regarding the debugging I would just use `debug=true` (and not debug_instruction=true and debug_details=true etc) and then show **all** path details. If this is too much in the UI one can always pick just a few or one path detail or we can later think to improve this.

boldtrn: > Ok, I think we should keep the variable 'DETAILS_PREFIX' as we do similar for the others.

Do I understand you right, that I should revert the change then? Or just change `path_details` to `details`? So that a request might look like `details=average_speed`. I updated this PR use `details=average_speed`.

> But the most important thing is: if we name the parameter x then the json entry should be also x

I agree 100%, I missed that we have the return type named as details.

> Also regarding the debugging I would just use debug=true

This would conflict with the debug parameter of the API. There we also got a debug parameter that formats the json output. Not that this would be an issue, just wanted to mention it. Changed this as well.

boldtrn:
> I see. But if you'll be successful to use 'instructions.pointlist' just as a getter then the setter will disappear?

Yes, maybe :). I am not yet sure how we will be able to provide the convenience getter, at some point we have to insert a PointList. Maybe it would make sense at this point to have a builder for PathDetails and a Builder for Instructions, which might offer the same interface. The instructions will be built after the simplification as well as the PathDetails. At this point the information stored in them is fixed and everything can be final.

core/src/main/java/com/graphhopper/util/Parameters.java
karussell: > Do I understand you right, that I should revert the change then

Yes, the variable should be named DETAILS_PREFIX

> Not that this would be an issue, just wanted to mention it

Yes, I know. I do not see an issue with this as we can always change UI debug parameters IMO

karussell: and the dot is missing, so yes, a revert is best: DETAILS_PREFIX="details." or are you using this elsewhere where the dot is bad?

core/src/main/java/com/graphhopper/util/details/PathDetails.java
boldtrn: Look at this loop:
```
for (PathDetailsCalculator calc : calculators) {
            if (calc.isEdgeDifferentToLastEdge(edge)) {
                calc.getPathDetails().endInterval(numberOfPoints);
                calc.getPathDetails().startInterval(calc.getCurrentValue());
                numberOfPoints = 0;
            }
        }
``` 

We do this for every edge, for the first edge, the `if (calc.isEdgeDifferentToLastEdge(edge))` is true. We haven't opened any pathDetail yet, but trying to end the interval before starting a new one. I think this doesn't hurt, but makes the code easier to read.

core/src/main/java/com/graphhopper/util/details/PathDetails.java
boldtrn: Ok, I just created the PathDetailsBuilder that creates PathDetails. We cannot simply use the List<Details>. PathDetails also have a name, contain extra logic to merge PathDetails, and logic for the Json serialization.

core/src/main/java/com/graphhopper/util/Parameters.java
boldtrn: Mhm, I think we have to differentiate between two different concepts here. Either we use the old way of requesting the PathDetails or the new array style that was proposed by you.

> Yes, the variable should be named DETAILS_PREFIX
> and the dot is missing, so yes, a revert is best: DETAILS_PREFIX="details." or are you using this elsewhere where the dot is bad?

The usage of the variable just completely changed, it's not used as a prefix anymore. Right now this is used for the request in the form `detail=average_speed`. If we go back to the old request style we would write `details.average_speed=true`. Then we would add the dot and rename it again to DETAILS_PREFIX. Let me know which style you prefer.

boldtrn:
I just went over the code, updated the Javadoc and improved some minor things. Should be good to merge now, except we want to change anything like the Parameters.

core/src/main/java/com/graphhopper/util/details/PathDetails.java
karussell: > PathDetails also have a name, contain extra logic to merge PathDetails, and logic for the Json serialization.

The logic for the serialization should go into the json serializer, but yes, the name could make this extra class necessary (merge method could go into the builder as a static function or so). Hmmh, I find it suboptimal that inside PathDetails there is a variable pathDetails. Also now we have a PathDetailsCalculator and a PathDetailsBuilder :(

karussell: Isn't the structure a `Map<String, List<PathDetail>>` and not `List<PathDetails>`? With a Map we naturally avoid duplicates with the same name (the `merge` method belongs more to managing the Map) and avoid the additional PathDetail**s** class for just "name and `List<PathDetail>`"

web/src/main/java/com/graphhopper/http/GraphHopperServlet.java
```diff
@@ -112,6 +113,8 @@ public void doGet(HttpServletRequest httpReq, HttpServletResponse httpRes) throw
                     throw new IllegalArgumentException("If you pass " + POINT_HINT + ", you need to pass a hint for every point, empty hints will be ignored");
                 }
 
+                List<String> pathDetails = new ArrayList<String>(Arrays.asList(getParams(httpReq, PATH_DETAILS)));
```
karussell: minor: the wrapper `new ArrayList` should not be necessary. If IntelliJ warns you just do `Arrays.<String>asList`?

core/src/main/java/com/graphhopper/util/Parameters.java
karussell: > Either we use the old way of requesting the PathDetails or the new array style that was proposed by you.

Uups, you are right.

web/src/main/webapp/js/instructions.js
```diff
@@ -43,19 +43,17 @@ function addInstruction(mapLayer, main, instr, instrIndex, lngLat, useMiles, deb
...
 
-        if(debugInstructions){
+        if (debug) {
             // Debug Turn Instructions more easily
             L.marker([lngLat[1], lngLat[0]], {
                 icon: L.icon({
-                    iconUrl: './img/marker-small-blue.png',
-                    iconSize: [15, 15]
+                    iconUrl: './img/marker-small-red.png',
+                    // Mad the instructions icon a bit bigger, as they are placed before the path_details
```
karussell: Mad->Made

web/src/main/webapp/js/tools/url.js
```diff
@@ -36,7 +36,7 @@ function mergeParamIntoObject(res, key, value) {
     var objectIndex = key.indexOf(".");
     if(objectIndex < 0) {
         // force always array for some keys even if just one parameter
-        if (typeof res[key] === "undefined" && key !== "heading" && key !== "point") {
+        if (typeof res[key] === "undefined" && key !== "heading" && key !== "point" && key !== "path_details") {
```
karussell: path_details->details

web/src/main/webapp/js/instructions.js
karussell: The response format is something like?

```json
"details": {
 "average_speed": {
     25: [[15, 30], [40, 45]],
     40: [[0, 15], [30, 40]],
  }
}
```
I'm just confused that the array access is detailObj[detailKey]["details"]; and not detailObj["details"][detailKey];

core/src/main/java/com/graphhopper/util/details/PathDetails.java
karussell: Ok, understood. I think the underlying reason is also that not all path detail values start at index 0

karussell:
Code is better than 1K words ... so I've did a quick refactoring regarding the question of `PathDetails` vs. `List<PathDetail>` and created this branch: https://github.com/graphhopper/graphhopper/commit/2a384fd1a1fb5e1d99e7dd6e854ad8db39c67ad5

Where I also merged the logic for 'calculator'&'builder' and was able to use a Map.Entry instead of the extra PathDetails class. Also some inconvenient differences between InstructionList and PathDetails are no longer available and we can avoid the private getLength helpers and some minor duplicate checks are no longer necessary.

Do you think this makes more sense or is this too complicated?

The remaining task in this commit is the JSON serialization. We could avoid this special serialization method if we would store the PathDetail information in the form of List<intervals>. I would leave this and the decision to you (if you find this approach better at all :))

What I stumbled over is that in GraphHopperWeb the deserialization for the path details is missing (not sure if needed in this PR).

web/src/main/java/com/graphhopper/http/GraphHopperServlet.java
boldtrn: Oops, yes you are right :)

web/src/main/webapp/js/instructions.js
boldtrn: Good point, I am not sure why it was that way :) or why I changed that, but now it is in that format.

```
details: {
   average_speed: {
      45.0: [[68,73], ...
```

boldtrn:
> Code is better than 1K words ... so I've did a quick refactoring regarding the question of PathDetails vs. List<PathDetail> and created this branch: 2a384fd

Nice, thanks for the improvement! I merged your changes into this PR.

> Do you think this makes more sense or is this too complicated?

Mhm, hard to say. For me the Calculator and the Builder had two different purposes in the beginning, but you are right, with the latest changes they became more and more similar, so I think it makes sense to merge them. Also I think it's better to have a native List compared to a model class, if no special methods are needed. This case is a bit borderline, but since I like that the PathSimplification became a lot easier and PathDetails are now already similar to the InstructionList this is a big benefit and I think this makes therefore a lot of sense.

> The remaining task in this commit is the JSON serialization

I just did this and the code is working again. I just reused the old JSON serialization code and use it in the SimpleRouteSerializer. Not particularly elegant, but works.

> We could avoid this special serialization method if we would store the PathDetail information in the form of List. I would leave this and the decision to you (if you find this approach better at all :))

I am not sure if I understand you correct here? How would you store this in a list for the serialization? 

> What I stumbled over is that in GraphHopperWeb the deserialization for the path details is missing (not sure if needed in this PR).

You are right, I haven't thought about that yet. This brings us to an interesting point. In GraphHopperWeb we rebuild the GHResponse from json. However, the way we store the PathDetails in the GHResponse is not really useful to actually use them as we store `List<PathDetail>`. This only stores number of points, but no actual coordinates like the Instructions. Maybe we have to store them in a different way in the GHResponse. Maybe we should store them as "Map<String, Map<Object, List<int[]>>>" or maybe as a convenient model class, like the InstructionList, the PathDetailsList?

karussell:
> > We could avoid this special serialization method if we would store the PathDetail information in the form of List. I would leave this and the decision to you (if you find this approach better at all :))

> I am not sure if I understand you correct here? How would you store this in a list for the serialization?

I mean, can we hold the information in Java so that we need no explicit converter method, maybe just some jackson annotation?

boldtrn:
> I mean, can we hold the information in Java so that we need no explicit converter method, maybe just some jackson annotation?

Well, we could store it as: `Map<String, Map<Object, List<int[]>>>`?

> Then jackson would automatically do the deserialization for us. Maybe you meant already the same when you wrote about PathDetailsList, but potentially we can live without a separate domain model.

Yes and no. It would be helpful I think. For example this method would be very useful for PathDetails as well, taken from InstructionList `public Instruction find(double lat, double lon, double maxDistance)`, essentially to find the current PathDetail. Could be also put into a helper method though and might not be worth the overhead. Maybe this is also something for a later PR?

karussell:
> Well, we could store it as: Map<String, Map<Object, List<int[]>>>?

Hmmh, this wouldn't make the handling in Java easier. But we can add some more domain models, where the naming would be unclear to me. 

When I thought about this a bit more: the `Object` is a bit unsafe for floating point stuff we should do some rounding. Or just using the current model as the JSON output:

```json
"details": {
   "average_speed": [[45.0, 5], [25.0, 10], [45.01, 3]]
}
```

Where the array is always in the form of `[value, number_of_points]` which would be the PathDetail content. This saves a bit text (no double brackets, the value can be written directly as number and just one number for the interval) but repeats the value if it occurs multiple times in a path, still the rounding of the value is up to the client (see 45.0 vs. 45.01)


ok, the real json where the deserialization would already work, would be very noisy:

```json
"details": {
   "average_speed": [
       {"value": 45.0, "points": 5}, 
       {"value": 25.0, "points": 10}, 
       {"value": 45.01, "points": 3}]
}
```

> Maybe this is also something for a later PR?

Ok, maybe until then can we hide the PathDetail(s) related class a bit?

karussell:
I think in general this is a good solution and implemented this in [this commit](https://github.com/graphhopper/graphhopper/commit/7f5b936bd5c9b1d0f944b3318329b823c77f91f3).

(the JS part is not yet adapted and also the GraphHopperWeb part is not yet tested but should theoretically work this way without any custom code :))

@boldtrn would love to get your feedback on this :)

boldtrn:
Sorry, I must have missed your comment above.

> "details": {
   "average_speed": [[45.0, 5], [25.0, 10], [45.01, 3]]
}

I thought about this as well. I think this is an excellent solution. It might take a bit more space, but makes the handling a lot easier, we can also skip the complex MapCreation, which is a big plus IMHO. On the other hand, for the user it would be easier to have actual point references instead of the number of points, however that way it's easier to generate them. It is pretty easy to generate the references on the client side, I showed that in the js debug.

I added your commit to this PR and changed some smaller things and added a test for GraphHopperWeb.

However, I did not understand how to use the `PathDetailDeserializer`. I simply copied the code into GraphHopperWeb, which I currently assume is not the right way to do it, but I couldn't make it work with the Deserializer. Is there a better way to deserialize PathDetails?

karussell:
> On the other hand, for the user it would be easier to have actual point references instead of the number of points

Yeah, later on we could serialize the path details as GeoJSON.

> However, I did not understand how to use the PathDetailDeserializer

Will have a look on how to do this.

karussell:
Hmmh, just a partial parsing process is indeed not really simple, even reusing just the code lines is hard due to the Json exceptions. So we should migrate the custom serialization in a later issue when we move to

```
GHResponse res = objectMapper.readValue(str, GHResponse.class);
```

> and added a test for GraphHopperWeb

Thanks! I'll have a last look now and merge it...

karussell:
Ok, I think we have to change the name 'PathDetail.numberOfPoints' to 'PathDetail.length'. So if there is a pointlist of size 3 the pathdetail.length is 2 but the number of involved points is 3, and the currently used value would be 2.

Now regarding your comment:

> https://github.com/graphhopper/graphhopper/pull/1091#issuecomment-312922063

Is adding the via point really correct? I mean, maybe before this was a bug? Lets assume I have two paths with a pointlist of size 2 for both. Now when I merge them the pointlist size is 4 and so should be the PathDetail.length (?)

karussell:
Or is the via point somehow not part of the last PathDetail of the first path **and** also not part of the first PathDetail of the second path? This would be ugly.

karussell:
Regarding the json representation: what we need to find is a generic representation of a sparse array in JSON or better an efficient representation of an array with many duplicates. I've looked around and found not really a standard for this (json itself does specify this). We should still think about changing it to `[[index, value], ...]` instead of `[[value, length], ...]`. Then we would need again a container class ala PathDetail**s** or PathDetailList with custom (de)serialization as the previous entry is necessary in the deserialization code. BTW: having it as `[["index": value], ...]` is a bit better key-value representation but has the drawbacks (client side) of a necessary string-to-int conversion of the key and a slightly worse handling of the map compared to an array. Before setting this public format into stone it is really to have this clear.

Regarding the via issue: the PointList in one instruction is indeed `[from1,from2)` but this is wrong and it should be `[from1, from2]` ie. including the next start point, which makes merging PointLists harder but it is more meaningful and then the length of a ViaInstruction with one point would be 0 (size-1) and not 1. This is not really important to fix as we deprecate the getPointList later anyway.

We need a test that fails without the via point adding in the merge method.

karussell:
Doh, this is hard. Having no length makes the length of the last PathDetail object unspecified and everything slightly harder to handle. So is it `[[index, length, value], ...]` vs. `[[length, value], ...]`? And in Java we have the choice between implicit length through follow-up element of the list or explicit length but data duplication e.g. if length change of one element every follow-up element has to change its index. So we'll need the PathDetail**s** class back and maybe call it PathDetailList to avoid mix-up.

boldtrn:
I just created a comparison between three different formats:

- 1) The current format `[value, length]`
- 2) An array containing `[from, value]`
- 3) An array containing `[from, to, value]`

I tested this for the following route:
http://localhost:8989/?point=49.174522%2C9.349365&point=49.007249%2C9.494934&point=48.469279%2C10.096436&point=48.210032%2C9.332886&point=47.829752%2C8.833008&point=48.416442%2C8.566589&point=48.603858%2C9.467468&point=48.785152%2C10.074463&point=48.91167%2C8.602295&details=average_speed

This route is ~750 km long.

I returned **5** PathDetails, every time the `average_speed`, but 4 copies of it and compared the response sizes, not gziped.

The size of 1) is 88.2 kb. The size of 2) is 92.7 kb. The size of 3) is 102.4 kb. I attached the according files as zip if you want to have a look at the implementation.

[3routes.zip](https://github.com/graphhopper/graphhopper/files/1217196/3routes.zip)

karussell:
Thanks! Is this for car or motorcycle? Would you mind to create the same route for bike or foot?

Just locally gzipped them and the sizes are a bit closer: 1. 30.9kb, 2. 31.7kb, 2. 32.6kb so the maximum difference is under 6%. My minor preference is towards 3. The argument is: we already have a compact format and with 1 this would be a bit harder to use. Later we could introduce a details_format parameter that accepts "geojson", "compact" or "default" or so.

boldtrn:
> Thanks! Is this for car or motorcycle?

This was car.

I attached the files for bike, the difference here is less: 237,7kb; 240,4kb; 246,4kb. There are no? speed changes for footrouting, as everything is 5km/h.

> My minor preference is towards 3

I am fine with number 3, currently I cannot see any downside besides that it takes more space in the response. The implementation is also pretty straightforward if we only use the format in the response but actually keep the Java model the same. I therefore introduced a new inner class for PathDetail:

```
public class PathDetail {
  
......

    public Response asResponse(int startReference){
        return new Response(this.value, startReference, startReference + this.numberOfPoints);
    }

    public class Response{

        public final Object value;
        public final int from;
        public final int to;

...
    }
}
``` 

Serialization and Deserialization can be done by Jackson, but in the end we need to convert the `List<PathDetail>` to `List<PathDetail.Response>` or the other way around for deserialization. BTW: This was just the fastest way to implement this, I haven't thought about this enough to state that this is the best way, but it's not too bad.

[issue_439_bike.zip](https://github.com/graphhopper/graphhopper/files/1217300/issue_439_bike.zip)

karussell:
> The implementation is also pretty straightforward if we only use the format in the response but actually keep the Java model the same

Hmmh, this looks ugly, especially when we want to avoid special serialization classes. I already refactored everything to use first+last in PathDetail (getLength=last-first) ... but of course the simplification code gets a bit more challenging. The benefits would be that json and java are identical and the client usage in java would be also simpler.

karussell:
> I attached the files for bike, the difference here is less: 237,7kb; 240,4kb; 246,4kb.

Nice, so lets do it this way!

boldtrn:
> Hmmh, this looks ugly, especially when we want to avoid special serialization classes. I already refactored everything to use first+last in PathDetail (getLength=last-first) ... but of course the simplification code gets a bit more challenging.

The more challenging simplification was the reason why I didn't do that. Also this makes the handling with multiple paths more complicated. But sure all this is solvable :). If you already did it, let's use it :).

karussell:
> Also this makes the handling with multiple paths more complicated

why?

boldtrn:
> why?

Since we reference the PointList, we have to remember the size of the overall PointList. Or we have to change the reference in the merge method.

karussell:
The merge method should be easier as we can just call pathDetails.setLast(otherDetail.getLast()) but I didn't get the simplification working easily. Would you have a look?

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -98,69 +94,57 @@ public PointList simplify() {
             if (toSimplifyIndex >= 0 && simplificationPossible) {
                 // Only simplify if there is more than one point
                 if (nonConflictingEnd - nonConflictingStart > 1) {
-                    // Simplify
                     int removed = douglasPeucker.simplify(pointList, nonConflictingStart, nonConflictingEnd);
                     if (removed > 0) {
                         for (int i = 0; i < toSimplify.size(); i++) {
-                            reduceNumberOfPoints(toSimplify.get(i), offset[i], removed, startIntervals[i], endIntervals[i] - removed);
```
karussell: Here I didn't understood what nonConflictingXY means ... I think we have to call this for all PathDetails even if no simplification is possible, as we need to change the pathDetail.first variable for all entries following a simplified PathDetail.

boldtrn: The nonConflictingXY describes the index from where to where we can simplify without potentially remove a point that we might have to reference later.

> I think we have to call this for all PathDetails even if no simplification is possible, as we need to change the pathDetail.first variable for all entries following a simplified PathDetail.

Yes this is true, since we don't store the length anymore, but the indexes we have to propagate our change trough all PathDetails, which is not very nice. That's btw the main reason why I decided earlier to only use length instead of indexes.

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -98,69 +94,57 @@ public PointList simplify() {
...
+                throw new IllegalStateException("toShiftIndex cannot be negative");
+
+            int length = getLength(toSimplify.get(toShiftIndex), offset[toShiftIndex]);
             startIntervals[toShiftIndex] += length;
             offset[toShiftIndex]++;
-            if (offset[toShiftIndex] >= toSimplify.get(toShiftIndex).size()) {
-                endReached = true;
-            }
+            if (offset[toShiftIndex] >= toSimplify.get(toShiftIndex).size())
+                break;
```
boldtrn: should we just `return pointList`?

core/src/main/java/com/graphhopper/util/PathSimplification.java
boldtrn: > Yes this is true, since we don't store the length anymore, but the indexes we have to propagate our change trough all PathDetails, which is not very nice. That's btw the main reason why I decided earlier to only use length instead of indexes.

Ok, I just checked, we actually don't necessarily have to do this. Right now the Simplification is completely based on the length and does not use the index of the toSimplify elements at all, since the length of elements is fixed until they are actually simplified, it still works.

If you check how the startInterval and endInterval are managed, you will see that we don't use the index, but only the length of the elements. However, it looks a bit like "magic" and you don't understand it on first sight, which I generally don't like. On the other hand the simplification is a bit more complex anyway so it might be ok, that the code is not completely understandable by just skimming over it.

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -98,6 +98,16 @@ public PointList simplify() {
                     if (removed > 0) {
                         for (int i = 0; i < toSimplify.size(); i++) {
                             reduceLength(toSimplify.get(i), offset[i], startIntervals[i], endIntervals[i] - removed);
+                            /*
```
boldtrn: As you can see, the tests succeed even if we don't do this. Also the result on the map with debug=true looks the same to me in both cases, so this seems unnecessary. As described in my other comment and I would remove it before we merge this. I just left it there so you can compare it.

core/src/main/java/com/graphhopper/util/PathSimplification.java
karussell: If possible I try to return at the end

core/src/main/java/com/graphhopper/util/PathSimplification.java
karussell: Ah, I think I have now fully understood the complexity, so when there are n PathDetails we do n simplifications and roughly n²/2 updates to the PathDetail object. The only alternative would be to sort the PathDetails by first index, create a dependency graph and then being to only update the PathDetails that are necessary

karussell: But this is probably not really necessary as a first version (and probably even later)

core/src/main/java/com/graphhopper/util/PathSimplification.java
boldtrn: Ok, I am fine with that :)

core/src/main/java/com/graphhopper/util/PathSimplification.java
boldtrn: > Ah, I think I have now fully understood the complexity, so when there are n PathDetails we do n simplifications

No, we only have one PointList that we simplify. We only update the PahtDetails and Instructions. That's why I would like to remove the Points from the Instructions :).

> roughly n²/2 updates to the PathDetail object

No, as I said, we don't need to propagate the changes. Everything works with this code commented out. At first I thought it would be necessary though.

> The only alternative would be to sort the PathDetails by first index, create a dependency graph and then being to only update the PathDetails that are necessary

Instead of doing the simplififcation per PathDetails I tried to do it one run, which makes the process a bit more complicated but saves us a lot of computation (I think).

karussell:
There is stilll a problem in the simplification (I think it is not sufficient to reduceLength only for cases with `removed>0`). I've added a general check and in GraphHopperServletIT there will be e.g. this problem:

```
java.lang.IllegalStateException: PathDetail list average_speed is inconsistent due to entries 60.0 [30, 32] vs. 50.0 [38, 39]
	at com.graphhopper.util.PathSimplification.simplify(PathSimplification.java:135)
```

It is a bit strange that when you try this in the Measurement class against Germany with my minor changes there will be no cases for this problem.

boldtrn:
Thanks for finding this. It seems that this issue is fixed by uncommenting the loop which I thought that it would be unecessary.

Apparently, the loop that I thought would not be needed is necessary :). When I uncomment the code, the glitch is fixed. I also found another issue, when a connection was not found, we should not calculate the PathDetails. Now the measurement for Baden-Württemberg and Bayern works.

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -99,16 +99,13 @@ public PointList simplify() {
                     if (removed > 0) {
                         for (int i = 0; i < listsToSimplify.size(); i++) {
                             reduceLength(listsToSimplify.get(i), offset[i], startIntervals[i], endIntervals[i] - removed);
-                            /*
-                            // TODO This is not needed, I left it here temporarily, has to be removed before merging
-                            if(listsToSimplify.get(i).get(0) instanceof PathDetail){
+                            if (listsToSimplify.get(i).get(0) instanceof PathDetail) {
```
karussell: Why should it be necessary only for PathDetails?

core/src/main/java/com/graphhopper/routing/Path.java
```diff
@@ -377,7 +377,7 @@ public InstructionList calcInstructions(final Translation tr) {
      * @return List of PathDetails for this Path
      */
     public Map<String, List<PathDetail>> calcDetails(final PathDetailsBuilderFactory pathBuilderFactory, int previousIndex) {
-        if (pathBuilderFactory == null)
+        if (!isFound() || edgeIds.isEmpty() || pathBuilderFactory == null)
```
karussell: Hmmh, we should add an integration test and not calling PathMerger at all or wdyt?

karussell:
Thanks!

Another minor thing I noticed is that `maxspeed` is used as OSM tag and not `max_speed`. So either we break OSM style for the path details naming (which I would slightly prefer) or we use the ugly `averagespeed` instead of `average_speed` now.

core/src/main/java/com/graphhopper/util/PathSimplification.java
karussell: Ah, because for the remaining PathDetails we need to update the indices and instructions have the pointlist stored and do not need an update if the length changes

core/src/main/java/com/graphhopper/routing/Path.java
karussell: Also understand this now ;) and will add an integration test.

I think the `edgeIds.isEmpty()` should not be here as a PathDetail could still be included even if the path has no length. So we have to do similar things that we do in calcInstructions when the path is empty but the finish instruction should be there.

core/src/main/java/com/graphhopper/util/PathSimplification.java
boldtrn: Yes exactly :+1:

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -67,41 +67,42 @@ public PointList simplify() {
...
                 // Only simplify if there is more than one point
                 if (nonConflictingEnd - nonConflictingStart > 1) {
                     int removed = douglasPeucker.simplify(pointList, nonConflictingStart, nonConflictingEnd);
                     if (removed > 0) {
                         for (int i = 0; i < listsToSimplify.size(); i++) {
-                            reduceLength(listsToSimplify.get(i), offset[i], startIntervals[i], endIntervals[i] - removed);
-                            if (listsToSimplify.get(i).get(0) instanceof PathDetail) {
-                                List<PathDetail> pathDetails = listsToSimplify.get(i);
-                                for (int j = offset[i] + 1; j < pathDetails.size(); j++) {
+                            List<PathDetail> pathDetails = listsToSimplify.get(i);
```
boldtrn: This looks wrong, it can be either InstructionList or a list of PathDetail. It works, but I think we should not leave it like that. WDYT?

boldtrn:
> Another minor thing I noticed is that maxspeed is used as OSM tag and not max_speed. So either we break OSM style for the path details naming (which I would slightly prefer) or we use the ugly averagespeed instead of average_speed now.

Mhm, hard to say. For osm users `maxspeed` makes probably more sense. For everyone else `max_speed` makes probably more sense. My first intuition would be to go with `max_speed`, as this is the way we did it with essentially every other parameter, but I don't know a strong argument for either option.

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -98,12 +98,12 @@ public PointList simplify() {
...
                         for (int i = 0; i < listsToSimplify.size(); i++) {
-                            List<PathDetail> pathDetails = listsToSimplify.get(i);
+                            List pathDetails = listsToSimplify.get(i);
                             reduceLength(pathDetails, offsets[i], startIntervals[i], endIntervals[i] - removed);
-                            // instructions do
+                            // This is not needed for Instructions, as they don't contain references, but PointLists
                             if (pathDetails.get(0) instanceof PathDetail) {
                                 for (int j = offsets[i] + 1; j < pathDetails.size(); j++) {
-                                    PathDetail pd = pathDetails.get(j);
+                                    PathDetail pd = (PathDetail) pathDetails.get(j);
```
karussell: ups, thanks!

karussell:
Thanks @boldtrn for all the work - this is a nice new feature!

PR #1095, opened 2017-06-21
Broken link fixed

jsroyal:


karussell:
It is fixed on master already. Will pick it from there.

PR #1101, opened 2017-06-29
Updated Stopover texts for German and English

boldtrn:
As discussed in the forum, I changed stopover to waypoint and "Zwischenziel" to "Wegpunkt". I haven't touched the other languages. Should we change anything for the other languages?


boldtrn:
Mhm, why did Travis said that this build was successfull if you look at the logs it failed...?

karussell:
> Mhm, why did Travis said that this build was successfull if you look at the logs it failed...?

Do you mean the problems reported in #1102? There should be two builds, one for jdk8 and one for 9 and only jdk8 is required to succeed I think ... uhm, but it looks like jdk8 is missing!?

karussell:
Ah, no. We have both IMO:

```
matrix:
  fast_finish: true
  include:
    - jdk: oraclejdk9
....
jdk:
  - oraclejdk8
```

boldtrn:
Sorry for the late reply, somehow I missed the update here.

Look at [this travis build](https://travis-ci.org/graphhopper/graphhopper/builds/248323163?utm_source=github_status&utm_medium=notification), the maven test failed but travis reported test completed successful.

karussell:
The problems are fixed via #1102, see https://travis-ci.org/graphhopper/graphhopper/builds/249286195

boldtrn:
> The problems are fixed in #1102

Not sure if this was the issue. The build I referenced failed, because I forgot to change a String in a test. So the fail was correct. Travis did not show that it failed, but you can see that it failed in the console output.

karussell:
I think so, as the build with jdk8 was not triggered and only jdk9 was run but this is allowed to failed (for now)

boldtrn:
> I think so, as the build with jdk8 was not triggered and only jdk9 was run but this is allowed to failed (for now)

Ah ok, sorry I missed that. Now it makes sense, thanks for explaining.

PR #1108, opened 2017-07-05
Dropwizardifying

michaz:
Main conceptual changes (web module only):

* ~~Guice Modules are now wrapped by Dropwizard/Guice Bundles~~
* GraphHopper is wrapped by a GraphHopperService (something which can be started and stopped, managed by Dropwizard/Jetty), so that no heavy lifting is done at the time the Injector is created.
* ~~Servlets access things which are constructed by GraphHopper via a Provider, because now the Injector with the Servlets is not created by us (in "user space"), but by the framework, and that means that the things which GraphHopper creates aren't there yet (because GraphHopperService is not yet started), so we have to access them lazily. (This is because Servlets are always Singletons, created at startup, according to specs.~~ The issue disappears for Dropwizard/Jersey Resource classes, which can simply be (by default, even?) created per-request, so issues of initialization order don't arise.)
* The Configuration (Dropwizard concept) only wraps the CmdArgs for now. This is where to continue! I think half of GraphHopper's initialization/configuration/"manager" code will dissolve if we replace this by a hierarchy of configuration objects (which can include factories), which are automatically assembled from .yml (or anything else Jackson can parse), but can also be put together by hand, in Java.
* The custom web integration test code is gone. Seems to work fine with the Dropwizard/Jersey client and test support classes, out of the box.
* For the other web-services (map-matching, isochrones, etc.), no sub-classing or anything is required. You just compose Bundles.
* TODO: Much of the custom Jetty configuration code is gone. Don't know how much of that is really missing/breaking anything. But maybe we could rather check if we can remove even more (IP filters etc.), as maybe that doesn't really belong in the web application but should be done by some outside proxy, or something.


michaz:
Next steps: Reap the returns from using a framework, by simplifying, simplifying, simplifying, and using standard-based mechanisms for configuration, serialization, and _everything_ which isn't routing itself.

michaz:
(Except the bit-fiddling data-structures which make us fast.) :-)

michaz:
((But only the ones where we *know* this to be true.))

karussell:
Cool & thanks!

> The Configuration (Dropwizard concept) only wraps the CmdArgs for now. This is where to continue! 

My preferred way here is to go via a custom json (is this possible)? Because having a yml for service config and another json for vehicle profiles seems suboptimal to me (especially if different formats).

michaz:
> My preferred way here is to go via a custom json (is this possible)? Because having a yml for service config and another json for vehicle profiles seems suboptimal to me (especially if different formats).

Not sure, but to me it looks like in the end it doesn't matter and it will take anything Jackson can parse (including JSON, Yaml and who knows what else).

karussell:
Good - makes sense!

web/src/test/java/com/graphhopper/http/BaseServletTester.java
```diff
@@ -1,179 +0,0 @@
...
-import java.io.IOException;
-import java.net.HttpURLConnection;
-import java.util.concurrent.TimeUnit;
-
-import static org.junit.Assert.assertEquals;
-
-/**
- * @author Peter Karich
- */
-public class BaseServletTester {
```
karussell: Great that we can get rid of it :) !

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
```diff
@@ -39,27 +40,30 @@
...
+
+    @AfterClass
+    public static void cleanUp() {
+        Helper.removeDir(new File(DIR));
     }
 
     @Test
     public void testBasicQuery() throws Exception {
-        JsonNode json = query("point=42.554851,1.536198&point=42.510071,1.548128", 200);
+        JsonNode json = app.client().target("http://localhost:8080/route?point=42.554851,1.536198&point=42.510071,1.548128").request().buildGet().invoke().readEntity(JsonNode.class);
```
karussell: Can we try to use a test server port which does not collide when running the normal thing? Because often I have a GH server running and still want to run the tests

web/src/test/java/log4j.xml
```diff
@@ -1,25 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE log4j:configuration SYSTEM "log4j.dtd">
-<log4j:configuration>
```
karussell: I'm not sure if we can easily remove this. In production it is very important to have async logging as otherwise routing requests could be delayed if disc is busy. This is especially true under high load and when we fetch GBs of new road network data.

web/src/main/java/com/graphhopper/http/GraphHopperConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public CmdArgs cmdArgs = new CmdArgs();
+
+    public GraphHopperConfiguration() {
+        final DefaultServerFactory serverFactory = new DefaultServerFactory();
+        // Move Jersey out of the way -- static assets (the web client)
+        // and the API root cannot _both_ be mapped to "/".
+        // We don't use Jersey services yet (instead: pure Servlets),
+        // but once we do, we have to think of something.
+        serverFactory.setJerseyRootPath("/api/");
```
michaz: Hmmmm, but we want index.html at the root, right?

michaz: Perhaps we can put a redirect to /webapp/index.html on / or something..

karussell: As I would prefer to break the demo UI instead of breaking any API out there, I'm more for a change there like having /demo or similar.

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
karussell: BTW: sometimes I run multiple tests in parallel and therefor I had this port start-trial-error loop in the BaseServletTester :)

web/src/main/java/com/graphhopper/http/GraphHopperModule.java
```diff
@@ -62,110 +42,39 @@ public GraphHopperModule(CmdArgs args) {
     protected void configure() {
         install(new CmdArgsModule(args));
```
michaz: Same instance! Passed down over many levels. Way too complicated, but I can't simplify all at once. ,-)

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
michaz: Yeeeeeeah, also it looked a bit like it was starting many servers but shutting down only one. 

I see the point, but please also see the transparency of just writing down the port number which is tested. And of testing the real server application with its real default port number / real way of configuring its port number.

web/src/main/java/com/graphhopper/http/GraphHopperConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.graphhopper.http;
+
+import com.graphhopper.util.CmdArgs;
+import io.dropwizard.Configuration;
+import io.dropwizard.server.DefaultServerFactory;
+
+public class GraphHopperConfiguration extends Configuration {
+
+    public CmdArgs cmdArgs = new CmdArgs();
```
michaz: No, not yet, because there's downstream code (in the other services in the other repos) which currently insists on replacing it.

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
karussell: The multiple port approach probably was a bit too much, but we need test and production separation for multiple reasons. As this is easily achievable via a fixed test port and also transparent I do not see a big pain here. Furthermore the default production port should be the old 8989.

web/src/main/java/com/graphhopper/http/ChangeGraphServlet.java
```diff
@@ -1,87 +0,0 @@
...
-import javax.servlet.http.HttpServletResponse;
-import java.io.IOException;
-import java.io.InputStreamReader;
-
-/**
- * This class defines a new endpoint to submit access and speed changes to the graph.
- *
- * @author Peter Karich
- */
-public class ChangeGraphServlet extends GHBaseServlet {
```
karussell: wow - nice that this just goes away :)

michaz:
Currently trying to get rid of Guice, because I already found what I think is a bug in the third-party Dropwizard-Guice-integration.

Dropwizard already [contains](https://javaee.github.io/hk2/) a full ``javax.inject`` implementation, so we should rather use that.

karussell:
This would be great. 

> (If at all.)

This would be even better. Setting instances directly in the 'resources' should be possible once there are no servlets anymore I think.

michaz:
Let's let this rest for a while, I'll use and test it for developing my next few features and see how it behaves.

michaz:
Okay, the YAML vs. JSON thing is not an issue, and the reading of our property file is also fine.

But our command line options are a problem, like ``datareader.file=berlin-latest.osm.pbf``.

Dropwizard uses a config file first, and then patches that from system properties, like ``-Ddw.graphhopper.datareader.file=berlin-latest.osm.pbf`` (with the dw prefix hardcoded).

How important is it to be exactly backwards-compatible to how we currently call the JAR?

karussell:
> How important is it to be exactly backwards-compatible to how we currently call the JAR?

Not very. But I would try to avoid the 'dw' property - is this possible somehow? Maybe just -Dgh.datareader.file=berlin-latest.osm.pbf ?

michaz:
Oh, I see we do the system property patching thing internally, anyway! So if we just advertise that way of passing command line parameters, ``-Dgraphhopper.datareader.file=berlin-latest.osm.pbf``? It already works that way, right?

karussell:
> It already works that way, right?

Hmmh, yes, it should. We could introduce some abbreviation so that only "-Dgh." needs to be added and moved from program to JVM system options

michaz:
> We could introduce some abbreviation so that only "-Dgh." needs to be added and moved from program to JVM system options

That would be _yet another_ method. ``graphhopper`` exists and seems to work. Just tried it.

web/src/test/java/log4j.xml
michaz: Dropwizard uses Logback, not log4j -- I hope it as reasonable defaults (I should think so). This file was in a test directory anyway -- can't do anything in production.

web/src/test/java/com/graphhopper/http/GraphHopperServletIT.java
michaz: > Furthermore the default production port should be the old 8989.

Okay, that would go into the default config file, which we should provide. (Probably migrate the example properties file which we have now.)

michaz:
Okay, I would call this almost finished, but it needs plenty of testing how it behaves on our servers. Maybe you can give it a try on a test machine at some point?

Oh, by the way, JSONP is gone right now. We need that, right? ,-)

karussell:
> Oh, by the way, JSONP is gone right now. We need that, right? ,-)

I'm not sure. It is not documented for the isochrone API, so no one should use it. But if it fails we have a bit damaged experience&trust. The problem is: if something works, even if not documented, it could be used and is kind of part of the API contract.

karussell:
Ah, forget the most important aspect: thanks for this work :) ! And will have a look into this in the next days and see if something needs to be tuned

michaz:
> But if it fails we have a bit damaged experience&trust.

Hmhmmm. Maybe we can find something which we can just wrap around it in our installation, like a proxy. Maybe even nginx can do it?

karussell:
Yes, or could jackson do it somehow? https://stackoverflow.com/questions/20069088/deserialize-jsonp-with-jackson

config-example.yml
```diff
@@ -0,0 +1,140 @@
...
+  ##### Vehicles #####
+
+
+  # Possible options: car,foot,bike,bike2,mtb,racingbike,motorcycle (comma separated)
+  # bike2 takes elevation data into account (like up-hill is slower than down-hill) and requires enabling graph.elevation.provider below
+  graph.flag_encoders: car
+
+
+  # Enable turn restrictions for car or motorcycle.
+  # graph.flag_encoders: car|turn_costs=true
```
boldtrn: In Yaml, we might run into issues when using special characters like | or =. Did you try if this works or if we need to wrap the whole string ins quotes?

core/src/main/java/com/graphhopper/util/shapes/GHPoint.java
```diff
@@ -35,28 +35,27 @@ public GHPoint(double lat, double lon) {
...
+            double fromLat = Double.parseDouble(fromStrs[0]);
+            double fromLon = Double.parseDouble(fromStrs[1]);
+            if (lonLatOrder) {
+                return new GHPoint(fromLon, fromLat);
+            } else {
                 return new GHPoint(fromLat, fromLon);
-            } catch (Exception ex) {
             }
+        } else {
+            throw new IllegalArgumentException(str);
```
boldtrn: Should we add a description like length!=2?

config-example.yml
michaz: Haha. Nice one. Doesn't work, of course.

michaz: This is kind of a bridge technology anyway -- at one point, we should migrate it to a hierarchical config without self-invented encoding of structure.

I'll think about a suggestion what to do in the meantime.

michaz: Works. (deleted a comment where I say it doesn't work, but did something stupid there)

core/src/main/java/com/graphhopper/util/shapes/GHPoint.java
michaz: Maybe. I do such things incrementally and think, hey, an Exception is better than null (what was there before).

config-example.yml
michaz: JSON should work, too (without changing anything).

michaz:
The shell script will need some attention.

Main class is now GraphHopperApplication. It has a command-line interface provided by Dropwizard, (so it could replace the shell script in principle). I put in an "import" command to demonstrate. The shell script could use that. Benefit: "import" and similar things are accessible via the runnable JAR, and the main config file can be reused.

config-example.yml
karussell: Ah, cool :+1:

boldtrn: Thanks for checking that :)

core/src/main/java/com/graphhopper/util/shapes/GHPoint.java
boldtrn: True :)

karussell:
> Maybe even the 'tools' project could go away

I'm not sure how the class loader exactly works but it might get problematic when a user installs headles jdk and the swing class dependencies from the tools module are in the web module too. I do not see other issues, so if this works we can merge the tools module.

devemux86:
> The Android people (@devemux86?) would be fine with entering at the 'core' level, I'm sure. They don't want the logger etc., so they're probably not using the 'tools' project now, anyway.

Only the `core` module is needed on mobile.

> Don't know if "command line import without web dependencies" is an important thing to keep?

On desktop the `tools` module is useful for creating graphs (if not using desktop Cruiser).
Nothing that cannot be re-implemented by a developer for not having also web dependencies.

michaz:
Dropwizard convention is to put in a cute Ascii-art banner which goes into the log on startup. :) I used a train-themed one for our railway-customer specific app. Maybe the community can start finding one for the main distribution. :)

```
     ___                        _ __    _        _  _               _ __     _ __
    / __|      _ _    __ _     | '_ \  | |_     | || |     ___     | '_ \   | '_ \    ___       _ _
   | (_ |     | '_|  / _` |    | .__/  | ' \    | __ |    / _ \    | .__/   | .__/   / -_)     | '_|
    \___|    _|_|_   \__,_|    |_|__   |_||_|   |_||_|    \___/    |_|__    |_|__    \___|    _|_|_
  _|"""""| _|"""""| _|"""""| _|"""""| _|"""""| _|"""""| _|"""""| _|"""""| _|"""""| _|"""""| _|"""""|
  "`-0-0-' "`-0-0-' "`-0-0-' "`-0-0-' "`-0-0-' "`-0-0-' "`-0-0-' "`-0-0-' "`-0-0-' "`-0-0-' "`-0-0-'
```

karussell:
Haha, nice one!

karussell:
I'd use the simple [ivrit style](http://www.patorjk.com/software/taag/#p=display&h=2&v=0&f=Ivrit&t=graphhopper) plus maybe something else (addable later):
```
                        _     _                                 
   __ _ _ __ __ _ _ __ | |__ | |__   ___  _ __  _ __   ___ _ __ 
  / _` | '__/ _` | '_ \| '_ \| '_ \ / _ \| '_ \| '_ \ / _ \ '__|
 | (_| | | | (_| | |_) | | | | | | | (_) | |_) | |_) |  __/ |   
  \__, |_|  \__,_| .__/|_| |_|_| |_|\___/| .__/| .__/ \___|_|   
  |___/          |_|                     |_|   |_|              
```

tools/pom.xml
```diff
@@ -49,6 +49,13 @@
             <version>${log4j.version}</version>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+            <scope>test</scope>
+        </dependency>
```
michaz: Ah. I stumbled over this because dropwizard has its "dropwizard-testing", which includes a junit version.

Why do we have dependencies in the parent pom? Do they belong there? Parents often do "dependency management", where they say to children "**if** you need dependency A, then use version a.b". But ultimately only the modules by themselves can know what dependencies they have.

karussell: > Do they belong there? 

I do not know for sure. I just didn't thought of a case where we won't need junit and wanted to reduce code a bit.

> where they say to children "if you need dependency A, then use version a.b" (to resolve conflicts with siblings).

ok

karussell:
Thanks for the merge from master @michaz ! Would you mind to look into the build failure for JDK9?

"Caused by: java.net.BindException: Address already in use"

As JDK9 is released now we do no longer ignore build failures for this case.

I'll now try to move this PR forward and start with a real world performance comparison and so the recent merge makes this much easier :)

michaz:
Whoa! Yeah, those are two problems, a) first test fails because class not found (will fix, probably something about stricter dependency resolution in Java 9 or something), and b) Server does not get cleaned up on exception, so the rest of the tests fail.

Will do.

michaz:
> test fails because class not found (will fix, probably something about stricter dependency resolution in Java 9 or something)

Fixed in 415d1ac

michaz:
I don't get what AppVeyor's problem is, though.

karussell:
> I don't get what AppVeyor's problem is, though.

Yes, this looks strange:

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.4.1:single (make-assembly) on project graphhopper-web: Failed to create assembly: Error creating assembly archive with-dep: Problem creating jar: jar:file:/C:/projects/graphhopper/web/target/graphhopper-web-0.10-SNAPSHOT.jar!/assets/img/foot.svg: JAR entry assets/img/foot.svg not found in C:\projects\graphhopper\web\target\graphhopper-web-0.10-SNAPSHOT.jar -> [Help 1]

karussell:
How will I use this to do just the import? Or is this just for web why did you remove Import.java?

karussell:
Ah, ok. I have to use the web module as it seems.

karussell:
Had to use maven shade to avoid a strange error ala:

```
Failed to parse configuration at: server.applicationConnectors.[0]; Could not resolve type id 'http' as a subtype of [simple type, class io.dropwizard.jetty.ConnectorFactory]: known type ids = [] (for POJO property 'applicationConnectors''
```

see https://github.com/dropwizard/dropwizard/issues/2159#issuecomment-331080609

I cannot make the command line arguments ala `-Dgraphhopper.datareader.file` working, it always says "You need to specify an OSM file." and I have to use the config file, which is suboptimal for scripting.

michaz:
> How will I use this to do just the import? Or is this just for web why did you remove Import.java?

Yes. I think "web" should, in the future, be called "app" and contain everything stand-alone, no matter if command line or web-server. Because you need infrastructure (e.g. parsers) for both of them, and ideally they can be the same.
"core" would then be a pure library.

michaz:
> Now I cannot make the command line arguments ala -Dgraphhopper.datareader.file working, it always says "You need to specify an OSM file." and I have to use the config file, which is suboptimal for scripting.

Ah, will investigate, I thought this would work..

karussell:
There is also a possibility to specify commands after the jar file: http://www.dropwizard.io/0.9.1/docs/manual/core.html#man-core-commands 

Some notes/questions:

 * the logging format for errors is different to info: `ERROR [2018-01-31 06:57:01,228] com.graphhopper.http.resources.RouteResource: point=37.438` vs. `<requested_IP> - - [31/Jan/2018:06:57:01 +0000] "GET /route?point=37.438541`
 * And should we keep the current logging format or switch to the standard dropwizard? I slightly prefer to keep it if not too much effort - are there some advantages we would get for "standard dropwizard"?
 * also there is always a 0 at the end of each log line. Is this the time it takes per request and is a rounded integer?
 * Something is wrong. I cannot get the system to accept more than 10 queries per second. Will investigate.
 * can we avoid the healthcheck warning at start up?
 * should we use endpoint /maps instead of /webapp for UI demo?
 * the graphhopper banner for CLI is not fully printed, one part of the second P is missing somehow
 * Do you get the same problem regarding the configuration parsing and were able to fix it via maven shade or am I doing it wrong?

michaz:
`for every request there seems to be an additional log line ala <requested_IP> - - [31/Jan/2018:06:57:01 +0000] "GET /route?point=37.438541 and not just INFO .. or ERROR ..

Yes. From what I've heard, it is a standard "access log" format which also things like NGINX produce, and which is analysed by many standard tools. I'm sure it can be switched off.

`And should we keep the current logging format or switch to the standard dropwizard? I slightly prefer to keep it if not too much effort - are there some advantages we would get for "standard dropwizard"?

Oh, this is Logback, the format can be configured in the config file.

> It looks that we are blocking IPs for the application properly but not for admin. So either we should block it for both or disable the admin service to avoid security problems for people using this feature

Here I'd strongly suggest to either turn off admin service, or do nothing. I think we should remove IP filtering from the open source version. I have the rather strong opinion that a specialised web service which uses a framework should just leave the defaults of that framework like they are. Because this is the least surprise for a user who just Googles "Dropwizard" to look how it works.

I think the reasonable expectation is that a webservice is just a webservice, and that things like logging, security, etc. are details managed by proxies or containers.

michaz:
> can we avoid the healthcheck warning at start up?

Yes, by adding a healtheck. :-) They say they deliebrately put that in to force people to look at healthchecks.

michaz:
> Something is wrong. I cannot get the system to accept more than 10 queries per second. Will investigate.

Now _that_ is irritating. I'd be happy to learn more..

karussell:
> Something is wrong. I cannot get the system to accept more than 10 queries per second. Will investigate.

This looks much better now :). I think this is a measurement bug when there are errors. Now I was able to reduce the errors for the test instance and still use a realistic scenario and I'm getting 650-950 requests/sec for 10 workers (not that bad as we skips routes <100km)

The interesting thing is that dropwizard seems to better handle more workers. For workers>5 the throughput is often measurable better and much less differences there between two identical runs ("more predictable"). To avoid log and disc related slow down I piped the logs into /dev/null for both. Maybe there are other problems in the measurement procedure or differences in the jetty config.

For future reference: I'm using world wide car data only with 30GB RAM (graph is 28gb) and real world logs from production (mostly with CH) and execute the following on a different server very close to the GH server (ping <0.25ms):

```
./graphhopper.sh torture logfile=~/some.logs.gz workers=10 'baseurl=http://host:8989/route?' maxqueries=30000 timeout=1000
```

karussell:
Also found how to specify parameters via command line which is a bit ugly due to the long prefix "-Ddw.graphhopper." and because escaping the dot is required:

```bash
java -Ddw.graphhopper.datareader\\.file=some.pbf -jar gh.jar server config.yml
```

My preferrence seems only possible via [explicit parameters as Command](http://www.dropwizard.io/1.2.2/dropwizard-core/apidocs/io/dropwizard/cli/Command.html):

```bash
java -jar gh.jar server config.yml -datareader.file=some.pbf
```

But then we would need to duplicate configuration information in the configuration object and in the command object.

michaz:
> Now I cannot make the command line arguments ala -Dgraphhopper.datareader.file working, it always says "You need to specify an OSM file." and I have to use the config file, which is suboptimal for scripting.

For me, this works:

```
java -jar -Dgraphhopper.datareader.file=wurst web/target/graphhopper-web-0.10-SNAPSHOT.jar server config-example.yml
```

Also with maven shade, as per current commit. Sure you didn't mistype?

karussell:
Strange. It works now - thanks!

> Also with maven shade, as per current commit. Sure you didn't mistype?

Works now without any change.

karussell:
@michaz it would be better to create a new branch for new features here. I know we are very late to merge this but new features and new dependencies should not be added as they could make previous testing obsolete. Also I would like to discuss this first, e.g. why adding this to the core and why not e.g. hibernate validator etc.

michaz:
Ok sure. Thing is, I'm actually using this stuff for my test instance, which means I'm already kind-of maintaining three different states.. Not sure if I can handle a fourth one. :-/ But I see the point.

karussell:
Oh, why three states :D ? Next thing after full release is this, so it should not take long or I can use the commit at shade plugin.

michaz:
> Oh, why three states :D ?

``master``, ``dropwizardify``, ``trimet``. I currently do a merge-chain every time I commit a new domain-level feature. ,-) Not catastrophic, but not optimal either.

karussell:
> I'm sure it can be switched off. Though I'd rather suggest to leave it on and remove our own (similar) line.

The problem is that we need some more information like used algorithm etc. But yes, if we can use some kind of standard this would be an additional improvement.

So now that we have 0.10 released I would like to merge this and before merging we need to fix only some minors

 - [ ] switch back to commit 72ef9c1 
 - [ ] adapt graphhopper.sh (I'll do this. We then take arguments and forward them directly to the java app but we need a tiny wrapper to download the pbfs, potentially install maven and use some good defaults)
 - [ ] additionally to -Dgraphhopper allow -Dgh (or only allow the latter?)
 - [ ] avoid double logging, see above
 - [ ] turn off admin service in config by default for security reasons
 - [ ] adapt docs to document new config options or provide a backward compatible way
 - [ ] add healthcheck to avoid warning at start up
 - [ ] use endpoint /maps instead of /webapp for UI demo
 - [ ] the graphhopper banner for CLI is not fully printed, one part of the second P is missing
 
Later we should think about

 - a converter from old properties to new json/yml format?
 - remove IPFilter but we should avoid being a target like memcache ;)

michaz:
> switch back to commit

Sure, but let's branch off another PR from there. Not revert anything here. Just making sure.

karussell:
Thanks for #1314 - will have a look and close this here

Issue #978, opened 2017-03-01
Introduce new getData method to avoid 64bit limit of getFlags


karussell:
Open questions

 - int getData(int index) or maybe byte getData(int index) or `ByteRef getData(int index)` like lucene does similar? `int getData` would be easy to implement and not that bad.
 - How to do this with #472
 - Or should we implement a list of GraphExtension? This would allow more dynamically sized data and optional compression
 - Or just a separate DataAccess nicely integrated in GHStorage

See something already in this branch https://github.com/graphhopper/graphhopper/compare/intsref_refactoring

Best would be if we can avoid getData method and use higher level methods instead.

~~getFlags would be still necessary as it is the only 'column' where reversing is applied (#728)~~

Not part of this issue is to change the data type `long` to `byte[]` in the OSMReader IMO.

karussell:
This is done via #1447

Issue #728, opened 2016-05-29
Reversing flags is expensive


karussell:
When reversing the flags this is an expensive operation as also flag encoders that are not used for one specific traversal have to be properly swapped. Instead we could define one bit (e.g. first) which we use as indication for "forward storage" or "reverse storage", then on access of direction dependent properties like speed we can access the bit ranges dependent on this `bit0`:

``` java
class SomeFlagEncoder {...
 double getSpeed(flags) {
  if(isBit0Set(flags)) // if bit0 is 1 this defines 'forward'
     return forwardSpeedEncoder.getDoubleValue(flags);
  return reverseSpeedEncoder.getDoubleValue(flags);
}}
```

This way we can reduce the whole EncodingManager.reverseFlags method to one bit operation flipping the first bit, avoiding the loop **and** avoiding all swap operations (i.e. unrelated and even related). But introducing the `isBit0Set` for all direction dependent properties.

We could even avoid to store the direction with this one bit. Instead we could use the nodeA and nodeB as direction indication (e.g. `nodeA>nodeB` defines as 'forward') and store the flags accordingly. The only problem then would be that one would need to create the flags either in forward or in backward direction dependent on this low level storage which would be too ugly or we still need one potential `reverseFlags` call at least at writing time. Also the `bit0` solution is edge independent and would work without #472.

Take into account shortcuts of CHGraph. See `EdgeAccess.reverseFlags` where we swap the two accessibility bits and now would need to read the forward&backward property dependent on `bit0`

karussell:
Fixed via #1447

PR #1112, opened 2017-07-13
(WIP) EncodingManager refactoring

### LINKED ISSUE #472: Refactoring to reduce set/getFlags usage

karussell:
This is a first sketch of a new approach to improve the flexibility for the edge property storage and it fixes #472, #728, #978 and probably some more issues. 

For the moment just look at the nice [FastestCarWeighting](https://github.com/graphhopper/graphhopper/blob/8dd658c5ae48e8a690a90ed6c9135d4ae7d7479a/core/src/main/java/com/graphhopper/routing/weighting/FastestCarWeighting.java) and how we fetch the values in the constructor that are necessary  in the calcWeight method, really easy and understandable:

```java
double speed = edgeState.get(averageSpeed);
```

These values are composed upfront and are responsible just for a value (and optionally a reverse value), have a look into [OSMReaderTest.testEncodedValueBasedEncodingManager](https://github.com/graphhopper/graphhopper/pull/1112/files#diff-100e0c53430e5948905a11dcb0821360R812) for a complete test parsing an OSM file.

So all in all I was able to remove the necessity of the FlagEncoder classes. So just EncodingManager and a list of EncodedValue is necessary. And the best: this already works within the existing system and the new EncodingManager is compatible with the old.

I've created a EncodedValueFactory to create some default EncodedValues like maxspeed, surface, 'car access' etc so that useful Weightings can be created.

The main changes in this PR:

 1. `long flags` was replaced with a IntsRef that is not limited to 64bits (later on we could use this for all edge properties like distance and pointers, also there is a minor advantage that we could point directly into the storage backing array instead of copying but this is for a later PR)
 2. new directed EncodedValue that are able to store forward and backward values, no need for expensive reverseFlags method, much simpler usage in Weighting implementations, no need for flag encoders but still backward compatible to them

main goals of this PR:

 * edge storage should allow more than 64bits per edge
 * easier storing of edge properties -> EncodedValue methods
 * EncodingManager will be EncodedValues based
 * avoid reverseFlags

non-goals:

 * remove FlagEncoders
 * fix PathExtract
 * flexible storage for relation and node
 * solve problem with two bits for access restriction versus speed==0

Further work items:

 - rename MappedXY to more common DictionaryXY?
 - use new EncodedValue classes in old FlagEncoder classes: DataFlagEncoder, CarFlagEncoder, ...
 - merge master into this branch
 - what to do with the unfavoredEdge used only in QueryGraph -> do we need a 0-bit EncodedValue used just as a 'marker'?
 - use two EncodedValues (weight and access) for shortcuts in CHGraphImpl -> advantage would be that we would be able to store weight with higher precision
 - version of every EncodedValue to make loading graph more secure?
 - avoid Weighting.getFlagEncoder
 - replace AbstractFlagEncoder and FlagEncoder with something less ugly, separate turn and node flags (Profile == collection of EncodedValue or ComposedEncodedValues)
 - maybe rename all XY08 into XYOld or XY09 instead? Add a deprecation notice and removal with 1.x
 - remove all the TODOs and ugly hacks
 - copyProperties can be done via array copy and setting just the node indices afterwards
 - define TagParser order
 - OSMReader is currently doing duplicate work to calculate bits of one OSM way, instead it should calculate the bits once and just array copy to the other edges using edge.get&setData?
 - InstructionAnnotation getAnnotation -> remove this or replace by a ComposedEncodedValue that can warn about private, tool, ferry or ford access?
 - separate TurnWeighting handling -> how to do  storage of OSM-node and OSM-relation information?
 - ~~merge new EncodingManager with old~~
 - ~~try a real world example and parse OSM with a car profile and query via the new weighting mechanism~~
 - ~~use IntRefs instead of int flags~~
 - ~~a directed DoubleProperty and BooleanProperty class that holds two values (one value for a direction) that can be easily 'reversed' -> ComposedEncodedValue holds multiple encoded values? support get and getReversed directly for every encodedValue?~~
 - ~~should we introduce a PropertyFilter so that a maxspeed is not parsed and calculated if the highway tag is not accepted ('rail')?~~
 - ~~move out EncodedValue.parse into something else to make support of non-OSM easier~~

Later:

 - allow cross-integer space for EncodedValue otherwise we can get bad constellations e.g. if EncodedValue has two directions and requires e.g. 17 bits
 - we could introduce a "reverse" EdgeExplorer or `iter.setBaseNode(node, reverse=true)` and then avoid the reverse parameter in Weighting.calcWeight as well as the if clauses to fetch e.g. the forward or the reverse speed. I.e. one edge is directed API-wise and undirected storage-wise. We should do this undirected vs. directed stuff in a later step, see also https://github.com/graphhopper/graphhopper/pull/1112#issuecomment-322522398
 - store distance as an EncodedValue
 - But in general we need length dependent mechanism somehow
 - reimplement the CarFlagEncoder as a ComposedEncodedValue?
 - PathExtract should be more separated (call extract explicitly after calcPath and potentially rename calcPath to calcShortestPathTree)
 - solve problem with two bits for access restriction versus speed==0; what to do if a reverse read-out of a oneway edge: exception vs. silently ok? how would we do the exception - via a new setting "read of default value results in exception" or somehow associating the `access`-EncodedValue with e.g. the `max_speed`-EncValue?


core/src/main/java/com/graphhopper/routing/PathExtract.java
```diff
@@ -0,0 +1,100 @@
...
+                };
+            }
+
+            roundaboutAccess = new BoolAccessor() {
+                @Override
+                public boolean get(EdgeIteratorState edge) {
+                    return encoder.isBool(edge.getFlags(), FlagEncoder.K_ROUNDABOUT);
+                }
+            };
+        } catch (Exception ex) {
```
boldtrn: I am struggling with this a bit. Why do we need the try/catch, especially with a generic exception? Has it something to do if there is a encoder stored in the weighting? 

Wouldn't it be better to be able to read the state without creating an exception?

core/src/main/java/com/graphhopper/routing/lm/LandmarkSuggestion.java
```diff
@@ -40,7 +40,7 @@ public BBox getBox() {
      * to specify an explicit bounding box. TODO: support GeoJSON instead.
      */
     public static final LandmarkSuggestion readLandmarks(String file, LocationIndex locationIndex) throws IOException {
-        // landmarks should be suited for all vehicles
+        // landmarks should be suited for all profiles
```
boldtrn: What about PT :)?

core/src/main/java/com/graphhopper/routing/profiles/BitEncodedValue.java
```diff
@@ -0,0 +1,54 @@
+package com.graphhopper.routing.profiles;
+
+/**
+ * This class provides easy access to just one bit.
+ */
+public final class BitEncodedValue extends IntEncodedValue {
```
boldtrn: Nice, GH missed a Boolean Encoded Value, not sure if we should call it like this instead of Bit?

boldtrn:
Wow, this change looks impressive! Thanks!

I haven't found the time to give this a full in-depth review yet, I only touched the tip of the iceberg so far. My main concern until now is the "ugliness" that you asked to ignore. I would prefer a way to get the state without provoking an exception.

I think using exceptions as a "if replacement" should be avoided if possible. If not then it is what it is :).

core/src/main/java/com/graphhopper/routing/PathExtract.java
karussell: Do not care about it this much. It will&must disappear. Just wanted to make this backward compatible and working

core/src/main/java/com/graphhopper/routing/lm/LandmarkSuggestion.java
karussell: You mean public transit? In its current form A* should work and so also landmarks but not sure if it works nor if it makes it faster

core/src/main/java/com/graphhopper/routing/profiles/BitEncodedValue.java
karussell: I made the difference here because boolean implies 1 byte and double implies 8 bytes. For Integer I just didn't find a better word yet.

karussell:
> I think using exceptions as a "if replacement" should be avoided if possible.

Yes, of course. But if statements wouldn't make it more beautiful. We need a different solution, but I'll improve here only if I can solve the followings things before: 0. tuning everything a bit more 1. having a better API for some places (e.g. simple setüget using and 'reverse' stuff) and 2. trying to import a real data set to see if this is actually fast enough and we can model something more realistic.

core/src/main/java/com/graphhopper/routing/profiles/EncodingManager.java
```diff
@@ -0,0 +1,229 @@
...
+        });
+        this.mockEncoder = fetchEdgeEncoders().get(0);
+
+        this.parser = parser;
+        // Everything is int-based: the dataIndex, the EncodedValue hierarchy with the 'int'-value and the offset
+        // TODO should we use a long or byte-based approach instead?
+        this.extendedDataSize = Math.min(1, extendedDataSize / 4) * 4;
+    }
+
+    public EncodingManager add(TagParser parser) {
```
boldtrn: should be only allowed when not initialized?

core/src/main/java/com/graphhopper/routing/profiles/EncodingManager.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /**
+     * Size of arbitrary storage per edge and in bytes
+     */
+    public int getExtendedDataSize() {
+        return extendedDataSize;
+    }
+
+    // TODO should we add convenient getters like getStringProperty etc?
```
boldtrn: Would make it more convenient IMHO.

core/src/main/java/com/graphhopper/routing/profiles/IntEncodedValue.java
```diff
@@ -0,0 +1,148 @@
...
+            throw new IllegalArgumentException("negative value for " + name + " not allowed! " + value);
+    }
+
+    /**
+     * This method 'merges' the specified integer value with the specified 'flags' to return a value that can
+     * be stored.
+     *
+     * @return the storable format that can be read via fromStorageFormatToInt
+     */
+    public final int toStorageFormat(boolean reverse, int flags, int value) {
```
boldtrn: Why are flags int and not long?

core/src/main/java/com/graphhopper/routing/profiles/IntEncodedValue.java
```diff
@@ -0,0 +1,148 @@
...
+        checkValue(value);
+        return uncheckToStorageFormat(reverse, flags, value);
+    }
+
+    final int uncheckToStorageFormat(boolean reverse, int flags, int value) {
```
boldtrn: How about unchecked?

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.List;
+
+public class MappedDecimalEncodedValue extends IntEncodedValue {
+    private final int toValueMap[];
+    private final IntIntHashMap toStorageMap;
+    private final double precision;
+
+    /**
+     * TODO should we really use precision here or use something like the already used 'factor'?
```
boldtrn: Mhm, precision is the commonly used word for Double, so this does not sound wrong to me.

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
```diff
@@ -0,0 +1,53 @@
...
+    public MappedDecimalEncodedValue(String name, List<Double> values, double precision, Double defaultValue, boolean store2DirectedValues) {
+        super(name, (int) Long.highestOneBit(values.size()), -1, store2DirectedValues);
+
+        this.precision = precision;
+        // store int-int mapping
+        toValueMap = new int[values.size()];
+        toStorageMap = new IntIntHashMap(values.size());
+
+        int index = 0;
+        for (double val : values) {
```
boldtrn: Not that it would change a lot, but should we sort the list of Doubles first? This would also allow us to accept Collections in general. On the other hand the current List approach might be a bit more robust against adding new elements at the end if no additional bit is required (not sure if this is actually a use case).

boldtrn: Ah BTW, I think we should also check if any double occurs twice.

core/src/main/java/com/graphhopper/routing/profiles/StringEncodedValue.java
```diff
@@ -0,0 +1,46 @@
...
+ * to restore this information.
+ */
+public final class StringEncodedValue extends IntEncodedValue {
+    private final String[] map;
+
+    public StringEncodedValue(String name, List<String> values, String defaultValue) {
+        super(name, (int) Long.highestOneBit(values.size()));
+
+        // we want to use binarySearch so we need to sort the list
+        // TODO should we simply use a separate Map<String, Int>?
```
boldtrn: I would think Map might be faster for larger collections and maybe even easier to use?

core/src/main/java/com/graphhopper/routing/profiles/TagParser.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.graphhopper.reader.ReaderWay;
+import com.graphhopper.util.EdgeIteratorState;
+
+public interface TagParser {
+
+    String getName();
+
+    // TODO Every tag parser has an EncodedValue associated but except for convenient usage in EncodingManager we currently do not need this method
+    EncodedValue getEncodedValue();
```
boldtrn: If it doesn't hurt but makes the code easier :+1:

core/src/main/java/com/graphhopper/routing/profiles/TagParserFactory.java
```diff
@@ -0,0 +1,141 @@
+package com.graphhopper.routing.profiles;
+
+import com.graphhopper.reader.ReaderWay;
+import com.graphhopper.routing.util.AbstractFlagEncoder;
+import com.graphhopper.util.EdgeIteratorState;
+
+public class TagParserFactory {
```
boldtrn: Isn't this factory a bit of an overkill? Wouldn't it be easier to put this into classes?

core/src/main/java/com/graphhopper/routing/PathExtract.java
boldtrn: Ok :)

core/src/main/java/com/graphhopper/routing/lm/LandmarkSuggestion.java
boldtrn: Yes, I was just wondering as it's stating "suited" meaning, it's a good idea to combine PT with LM. I am not sure if this is true, so I was asking.

core/src/main/java/com/graphhopper/routing/profiles/BitEncodedValue.java
boldtrn: Ah I meant if we should call it BooleanEncodedValue instead of bit?

boldtrn:
> We need a different solution, but I'll improve here only if I can solve the followings things before: 0. tuning everything a bit more 1. having a better API for some places (e.g. simple setüget using and 'reverse' stuff) and 2. trying to import a real data set to see if this is actually fast enough and we can model something more realistic.

Sounds good and makes sense. I just saw it when browsing through the code and thought I might add my two cents.

core/src/main/java/com/graphhopper/routing/profiles/BitEncodedValue.java
karussell: A boolean usually uses 1byte (roughly) in the JVM, so I wanted to differentiate here as we really use just 1 bit, but not that important and we can change to Bool or Boolean

core/src/main/java/com/graphhopper/routing/profiles/IntEncodedValue.java
karussell: Currently it is an array of ints as the underlying RAMDataAccess is an array of ints and so this is the most efficient. But I'm thinking of using a byte array instead, which would be required if the GH node or way IDs exceed the 2^32 integer boundaries. The best would be if we could encapsulate this a bit with something like the IntRefs class and pass this in this method, which would also avoid the need of the public getOffset method.

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
karussell: Duplicates checking makes sense. Regarding the sorting: using Collection instead of a List is already possible now.

core/src/main/java/com/graphhopper/routing/profiles/StringEncodedValue.java
karussell: I bet a map is only faster for relative big collections (>30 entries) ... we should benchmark to make this argument ;). "Easier to use" is not an argument here as we can hide it in this class IMO.

core/src/main/java/com/graphhopper/routing/profiles/TagParserFactory.java
karussell: What do you mean with 'put this into classes'? Creating a RoundaboutTagParser and a HighwayTagParser etc? (This would introduce a lot more code/overhead IMO :)) 

I'm not super happy with it now but for me this factory was just convenient to avoid creating the same parsers in tests again and again.

karussell:
I was able to fix a few things (regarding highway filter and oneway storage) so that full Germany is now properly imported (in 6min) and possible to route via car. The following config.properties is necessary

```
graph.flag_encoders=car
prepare.ch.weightings=no
prepare.min_network_size=0
prepare.min_one_way_network_size=0
```

and the following parameters are necessary and currently no bidirectional algorithm will work:

`vehicle=weighting&weighting=fastest2&algorithm=astar`

Such an uni-dir route request from south to nord takes ~6sec. Not too bad. Only a smaller fraction (~25%) of astar.runAlgo seems to be related to the new 'dynamic' FastestCarWeighting which could be probably further reduced via reusing the IntsRef for the next edges.

Somehow the instructions have no street name associated. Will investigate.

The applyWayTags is currently very lightweight and accounts to 0% of CPU time in the profiler.

karussell:
When working on *"we could introduce a 'reverse' EdgeExplorer or iter.setBaseNode(node, reverse=true)"* the problem is that e.g. for the LocationIndex we get EdgeIteratorStates from graph exploration where the direction does not really matter and the only thing that is required is that in at least one direction it should be accessible by a certain vehicle, so we have the DefaultEdgeFilter. But this would not be easy with the new EncodedValues if we do not allow to grab the 'forward' and the 'reverse' access EncodedValue at for one EdgeIteratoreState. So we could either allow edge.getReverse(access) or another way could be to create two EdgeIteratorStates and use them somehow in QueryResult.

karussell:
> So we could either allow edge.getReverse(access)

I decided for this way.

It seems that we can replace most occurrences `long flags` with `IntsRef ints` ... but then we stumble in a bigger problem when copying the IntsRef into another graph as we cannot easily 'reverse' the IntsRef in case the nodes A and B have opposite order in the target graph compared to the source graph. So we must either find a way to reverse the IntsRef (I dislike this: the storage should no longer know about these specifics!) or we must get rid of the assumption nodeA<nodeB and always read the position of the nodes first. In current master we can reverse the flags easily as we already have a reverse method necessary to iterate through an edge in a different order, which is now done via a reverse variable in graph delegated to the EncodedValue implementations.

karussell:
> we must get rid of the assumption nodeA<nodeB and always read the position of the nodes first

That was possible, will have to fix one single unit test. Performance seems also nearly uneffected (after a second run)

karussell:
@boldtrn have commited a change that just compiles but does not work. Will try to make it working tomorrow or in the next days. What you could do already is a review and when most unit tests work doing a 'real' refactoring for the instructions code and the tunnel/bridge stuff. So that it uses the global EncodedValues and will work for all encoders.

karussell:
We need to store the state of the EncodingManager and restore it without the need to configure it. In theory this is now simpler than before because we just store the name, bit size and type of the used EncodedValues instead of creating a custom FlagEncoderFactory and even someone without the code will be able to use the graph directly or can easier uprade GH or apply changes to some storage config!

The problem is that I would like to have this in the GraphHopperStorage, which is in `core` and there we currently use just the jackson annotations dependency. Therefore I'm currently thinking about putting the jackson-databind dependency plus just the GHJson implementation in yet another but independent module (maybe just as a `provided` dependency) or directly use it in the `core` module (except for the iOS port this should be no problem)

core/src/main/java/com/graphhopper/routing/profiles/TagsParser.java
```diff
@@ -0,0 +1,44 @@
...
+package com.graphhopper.routing.profiles;
+
+import com.graphhopper.reader.ReaderWay;
+import com.graphhopper.storage.IntsRef;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+
+public class TagsParser {
```
boldtrn: Can we either rename this class or the TagParser Interface? It's a bit hard to differentiate between the two?

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -20,99 +20,311 @@
...
+            if (surface) {
+                List<String> surfaces = Arrays.asList("_default", "paved", "asphalt", "cobblestone", "cobblestone:flattened", "sett", "concrete",
+                        "concrete:lanes", "concrete:plates", "paving_stones", "paving_stones:30", "unpaved", "compacted"
+                        , "dirt", "earth", "fine_gravel", "grass", "grass_paver", "gravel", "ground", "ice", "metal"
+                        , "mud", "pebblestone", "salt", "sand", "wood");
+                add(TagParserFactory.createSurface(new StringEncodedValue(TagParserFactory.SURFACE, surfaces, "_default")));
+            }
+
+            if (carMaxSpeed) {
+                // TODO NOW do we really need this for instructions?
```
boldtrn: No, but it's a possible improvement for instructions.

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -20,99 +20,311 @@
...
+                DecimalEncodedValue ev = new DecimalEncodedValue(TagParserFactory.CAR_MAX_SPEED, 7, 0, 2, false);
```
boldtrn: Would this be a duplicate of
`add(TagParserFactory.Car.createMaxSpeed(new DecimalEncodedValue(TagParserFactory.CAR_MAX_SPEED, 5, 0, 5, false), filter)).`? I would assume that one is the maxspeed for cars and the other the tagged max speed? But why is it car max speed bot htimes? Or is this just not implemented yet :)?

boldtrn: Ah, ok just read the next line :). `add(TagParserFactory.Car.createAverageSpeed(new DecimalEncodedValue(TagParserFactory.CAR_AVERAGE_SPEED, 5, 0, 5, false), speedMap)).`

I guess the idea woul dbe to check if both speed would be compatible and if yes, only store it once. In this case the number of bits and the factor would not match.

core/src/main/java/com/graphhopper/routing/profiles/BooleanEncodedValue.java
```diff
@@ -0,0 +1,66 @@
...
+                flags = flags | (1 << bwdShift);
+            ref.ints[bwdDataIndex + ref.offset] = flags;
+
+        } else {
+            int flags = ref.ints[fwdDataIndex + ref.offset];
+            flags &= ~fwdMask;
+            if (value)
+                flags = flags | (1 << fwdShift);
+            ref.ints[fwdDataIndex + ref.offset] = flags;
+        }
```
boldtrn: The if and else branch look pretty similar, maybe we can combine them in one function?

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
boldtrn: Ah ok the `OSMReaderTest#testEncodedValueBasedEncodingManager` actually fails because of this. `Cannot add parser car.max_speed. Already existing: car.max_speed`

boldtrn: After removing this line there is another exception: `java.lang.RuntimeException: Graph after reading OSM must not be empty. Read 12 items and 0 locations`

boldtrn: Ok, the issue seems to be that the access TagParser is not yet used in the EncodingManager. Currently the EM only checks other FlagEncoders.

core/src/main/java/com/graphhopper/routing/profiles/BooleanEncodedValue.java
karussell: This is a very perf critical section and avoiding code duplication seems only possible with some additional variables

boldtrn: It's only a couple of lines, so it's not a big issue anyway. I thought we might be able to rewrite the code to be something like:

```
        if (store2DirectedValues && reverse) {
            _setBool(bwdDataIndex, bwdMask, value, bwdShift)
        } else {
            _setBool(fwdDataIndex, fwdMask, value, fwdShift)
        }
```

karussell: I wouldn't risc this for now

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
karussell: Will have a look into OSMReaderTest to see what you mean

karussell:
In a local trial I've moved the GHJson implementation into a separate module so that the dependency is reader-json -> core -> json. @boldtrn Do you think this would make sense?  See my comment above why I think this is unavoidable.

(Maybe we could get rid of reader-json afterwards. For Android we could exclude this dependency and use a gson one?)

devemux86:
> For Android we could exclude the json-module dependency and use a gson one?

Maybe could decide one json library to avoid multiple dependencies.
i.e. having in all modules or other repositories either Jackson or Gson?
I think now there is Jackson everywhere.

karussell:
Just thought if gson is bundled on Android already somehow, then we would not need an additional dependency.

devemux86:
In Android there is only the simple [org.json](https://developer.android.com/reference/org/json/package-summary.html) package, Gson remains an external library.

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
karussell: For me the testEncodedValueBasedEncodingManager passes. What do you mean with "the access TagParser is not yet used in the EncodingManager"?

karussell:
Have pushed my work regarding the `json` module extraction. If you agree I would extract this change into a different PR and try to get rid of `reader-json` before.

Then, in this PR, we can use this *"inject into graph"*-approach or without injection and create a `GHJson`-object in the EncodingManager, so that it is not necessary to explicitly create it, but could be changed if required (like for iOS).

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
boldtrn: Interesting, maybe IntelliJ did something unexpected here. When I checked out the branch and tried the test, it failed for me. I just updated the branch to use your latest changes and then it wouldn't even build anymore. Running `mvn clean` fixed the issue, probably something changed that IntelliJ couldn't cope with. The test works now for me as well.

> What do you mean with "the access TagParser is not yet used in the EncodingManager"?

When I debugged the test to see what was happening, the EncodingManager didn't use the access tag parser in `acceptWay`. Now I does, but I can see that you didn't add it afterwards. I am not exactly sure what happened here. Sorry for the troubles.

boldtrn:
> In a local trial I've moved the GHJson implementation into a separate module so that the dependency is reader-json -> core -> json (via jackson). @boldtrn Do you think this would make sense?

Yes I think this makes sense. Json is planned to become a key/core role in GraphHopper. Having a json module in the core makes sense to me. 

When we think about the future, it might make sense to think about how we could allow different formats like yaml. If you look at other projects, a lot started with json and after some time also added yaml, just because writing yaml configurations is a lot more convenient. Maybe with this perspective in mind it would make sense to deserialize the GHJson into a "ConfigDataTransferObject", which might be just a Map<String,Object>, similar to what we would get from Jackson. But I feel like this might be too much right now, so maybe this would be something for a future change, once this PR is merged?

In addition to the above I also feel a bit uncertain if we really want to pass GHJson to low level classes like GraphHopperStorage. I think GraphHopperStorage should not know about json. Maybe we can handle the Json on the level of the "GraphHopper" class and not propagating it too deep into the lower level? WDYT?

karussell:
> a lot started with json and after some time also added yaml, just because writing yaml configurations is a lot more convenient. 

This is just a config change in Jackson if you use databind.

> I think GraphHopperStorage should not know about json. Maybe we can handle the Json on the level of the "GraphHopper" class and not propagating it too deep into the lower level? 

This is possible but I'm not sure if necessary as we also know in the storage how to store edges and nodes, so why should we not able to store json properties without the help of others in the storage?

core/src/main/java/com/graphhopper/routing/profiles/TagParserFactory.java
```diff
@@ -0,0 +1,495 @@
...
+        {
+            add("yes");
+            add("true");
+            add("1");
+        }
+    };
+
+    public static final String ROUNDABOUT = "roundabout";
+    public static final String ROAD_CLASS = "road_class";
+    public static final String ROAD_ENVIRONMENT = "road_environment";
```
boldtrn: IMHO the word `road_environment` is not ideal here. From the name I would assume that it's about the environment of the road, like city, forest, desert, etc.

Naming it is hard though. Maybe `road_structure` or `road_attributes`?

core/src/main/java/com/graphhopper/routing/profiles/TagParserFactory.java
```diff
@@ -0,0 +1,495 @@
...
+    public static final String MAX_WIDTH = "max_width";
+    public static final String SPATIAL_RULE_ID = "spatial_rule_id";
+    public static final String CURVATURE = "curvature";
+    public static final String CAR_ACCESS = "car.access";
+    public static final String CAR_MAX_SPEED = "car.max_speed";
+    public static final String CAR_AVERAGE_SPEED = "car.average_speed";
+    public static final String BIKE_ACCESS = "bike.access";
+    public static final String BIKE_AVERAGE_SPEED = "bike.average_speed";
+    public static final String FOOT_ACCESS = "foot.access";
+    public static final String FOOT_AVERAGE_SPEED = "foot.average_speed";
```
boldtrn: Would it be better to have these keys in a different class? When using them at a later stage, by referencing `TagParserFactory.BIKE_ACCESS`, I don't want to create a new TagParser but use an existing EncodedValue. 

For example 
`this.averageSpeed = em.getDecimalEncodedValue(TagParserFactory.CAR_AVERAGE_SPEED);` in FastestCarWeighting.

At this point I want to read average speeds and not create anything, but I have to use the Factory to get the key.

boldtrn:
> This is possible but I'm not sure if necessary as we also know in the storage how to store edges and nodes, so why should we not able to store json properties without the help of others in the storage?

Maybe I misinterpreted the idea here. I thought that the idea would be that the Storage would configure itself from the passed json. If it's only about storing a json file, then I don't see an issue.

I just pushed a change to this branch that enables the Elevation Interpolation (or at least the test for the Elevation Interpolation are succeeding :)). I had to change some minor things, where I am a bit unsure if you would like that, will add comments in the code.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -1003,6 +1003,7 @@ public int getMaxId() {
         public boolean next() {
             while (true) {
                 edgeId++;
+                edgeRowCache = null;
```
boldtrn: Not sure if this was intended or not, but when iterating over the edges, the edgeRowCache was not invalidated, which could lead to incorrect IntRefs.

core/src/main/java/com/graphhopper/routing/profiles/EncodedValueLookup.java
```diff
@@ -11,4 +11,6 @@
     DecimalEncodedValue getDecimalEncodedValue(String key);
 
     StringEncodedValue getStringEncodedValue(String key);
+
+    boolean containsEncodedValue(String key);
```
boldtrn: I needed a way to check if the EncodingManager supports/contains a encodedValue. Not sure if we should call this method supports or if there is a different way of doing this?

core/src/main/java/com/graphhopper/reader/dem/TunnelElevationInterpolator.java
```diff
@@ -29,12 +29,13 @@
...
-        super(storage, dataFlagEncoder);
+    public TunnelElevationInterpolator(GraphHopperStorage storage, StringEncodedValue roadEnvironmentEnc) {
+        super(storage, roadEnvironmentEnc);
     }
 
     @Override
     protected boolean isInterpolatableEdge(EdgeIteratorState edge) {
-        return dataFlagEncoder.isTunnel(edge.getData());
+        //TODO We should not hardcode this string here
+        return edge.get(roadEnvironmentEnc).equals("tunnel");
```
boldtrn: For now I just hardcoded this. But we should have a way to access these properties.

core/src/main/java/com/graphhopper/reader/dem/AbstractEdgeElevationInterpolator.java
```diff
@@ -86,6 +86,9 @@ private void interpolateElevationsOfTowerNodes() {
 
         while (edge.next()) {
             final int edgeId = edge.getEdge();
+            if(edgeId == 6){
+                System.out.println(6);
+            }
```
boldtrn: Sorry, already removed this :flushed:

core/src/main/java/com/graphhopper/routing/profiles/TagParserFactory.java
karussell: Especially this name is hard, yes. But attributes is not what is meant ... and structure also not my preference ;) ... let us think again :)

core/src/main/java/com/graphhopper/routing/profiles/TagParserFactory.java
karussell: Yes the place is incorrect especially since the keys belong to the core and the TagParserFactory to the `osm` module. Probably Parameters would be a better place but let us first solve the harder problems and write this down as a todo item.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
karussell: Nice finding. I think I have just done this in the normal iterator not in the "all iterator"

core/src/main/java/com/graphhopper/reader/dem/TunnelElevationInterpolator.java
karussell: The way it should work is something like

```java
constructor {
   tunnelIndex = roadEnvironmentEnc.indexOf("tunnel")
   // is this too ugly?
   roadEnvironmentEnc = (IntEncodedValue) roadEnvironmentEnc;
}
isInterpolatableEdge {
   return edge.get(roadEnvironmentEnc) == tunnelIndex;
}
```

core/src/main/java/com/graphhopper/routing/profiles/EncodedValueLookup.java
karussell: Not sure about the naming. The implementation in the AbstractFlagEncoder could be done with the encodedValueLookup

karussell:
> I just pushed a change to this branch that enables the Elevation Interpolation

Cool :+1: 

> Maybe I misinterpreted the idea here. I thought that the idea would be that the Storage would configure itself from the passed json. If it's only about storing a json file, then I don't see an issue.

It is both: we load&store the json and we would configure the EncodingManager from this file.

core/src/main/java/com/graphhopper/reader/dem/TunnelElevationInterpolator.java
boldtrn: Ah yes makes sense.

boldtrn: > The way it should work is something like

Could you explain why? The roadEnvironmentEnc is a StringEncodedValue. The get method would therefore return a String? The tests also succeeded like that.

Or was your comment about having a short cut and saving some computation time by not getting the stored string but the index? This would then require a new method for the EdgeIteratorState, like `getRaw`?

core/src/main/java/com/graphhopper/routing/profiles/EncodedValueLookup.java
boldtrn: How about `supportsEncodedValue`?

boldtrn: > The implementation in the AbstractFlagEncoder could be done with the encodedValueLookup

The issue with this is that we would return the values that are stored in the EncodingManager, but maybe not with/because of this encoder. I am not sure if this is important. For example:

```
EncodingManager em = new EncodingManager.Builder().addGlobalEncodedValues().addAllFlagEncoders("car,bike,foot").build();
CarFlagEncoder encoder = (CarFlagEncoder) em.getEncoder("car");

encoder.containsEncodedValue(TagParserFactory.BIKE_ACCESS) // => is true
```

For me this seems a bit weird. But maybe it seems weird to me because I am still thinking too much about the "old" system?

karussell: This is not a big issue as we will deprecate the FlagEncoder or at least how they work at the moment :)

core/src/main/java/com/graphhopper/reader/dem/TunnelElevationInterpolator.java
karussell: If you convert it into an IntEncodedValue the raw integer is returned (as `edge.get(IntEncodedValue)` is called) and no expensive string comparison is necessary and so it can be used e.g. in the calcWeight method.

> Or was your comment about

The comment regarding 'too ugly' is about the cast and so this usage with better perf is a bit too hidden.

>  This would then require a new method for the EdgeIteratorState, like getRaw?

I would not introduce a new method for EdgeIteratorState, instead we could think about roadEnvironmentEnc.getIntEncodedValue instead of the cast?

core/src/main/java/com/graphhopper/routing/profiles/EncodedValueLookup.java
karussell: > How about supportsEncodedValue?

Best would be just "supports(String key)" but this clashes with the existing stuff for the FlagEncoder ... but maybe there is no clash as the naming is quite different.

boldtrn: Yes there is a conflict with `EncodingManager#supports(String encoder)`.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
```diff
@@ -236,7 +236,9 @@ public void parse(IntsRef ints, ReaderWay way) {
         // Note: if you update this list you have to update the method getAccessValue too
         List<String> accessList = Arrays.asList(
                 //"designated", "permissive", "customers", "delivery",
-                "yes", "destination", "private", "no"
+                "yes", "destination",
+                //"private", <= this currently conflicts with SpatialRule.Access, as this only allows yes, conditional, and no
```
boldtrn: I think this shouldn't be an issue? This did not work with the SpatialRule.Access, as this only contains 3 entries. The returned integer was used in the DataFlagEncoderTest like this:

```
idx = accessClassEnc.getInt(false, ints);
assertEquals(SpatialRule.Access.NO, SpatialRule.Access.values()[idx]);
```

Is this the expected behaviour?

boldtrn:
How should we handle maximum values? For example think about the maxspeed. If the tag would contain a maxspeed of 1000 (see `DataFlagEncoderTest#testLargeMaxspeed`). I see three options:
- throw an exception as we do right now (I don't really like this option)
- handle it graceful in the EncodedValue:
Instead of:
```
if (value > maxValue * factor)
   throw new IllegalArgumentException(getName() + " value " + value + " too large for encoding. maxValue:" + maxValue * factor);
```
We could do:
```
if (value > maxValue * factor)
   value = maxValue * factor
```
- Make it possible to retrieve the maxValue and decide per TagParser. We would therefore need a method to get the maxValue from the EncodedValue. For maxspeed we could for example fix the the value and log a warning. For something else we might want to thrown an exception, etc.

core/src/main/java/com/graphhopper/routing/util/DataFlagEncoder.java
karussell: Both should not depend on each other as there are much more than 2 conditional restrictions we need a mapping here.

karussell:
> Make it possible to retrieve the maxValue and decide per TagParser

This would be my preferred choice, as otherwise the EncodedValue would do parts of the logic. Per default we should throw an exception.

core/src/main/java/com/graphhopper/routing/profiles/EncodedValueLookup.java
karussell: I know but the naming of an encoder is different (always?) to the one of the EncodedValue? At the end "supports encoder" is a usage we should deprecate

core/src/main/java/com/graphhopper/reader/dem/TunnelElevationInterpolator.java
boldtrn: The easiest solution I found was to simply cast the StringEncodedValue to IntEncodedValue. I just pushed that change. I haven't measured performance, but I think the cast should be alright?

core/src/main/java/com/graphhopper/reader/dem/BridgeElevationInterpolator.java
```diff
@@ -29,13 +30,16 @@
...
         super(storage, roadEnvironmentEnc);
+        //TODO We should not hardcode this string here
+        bridgeRef = roadEnvironmentEnc.indexOf("bridge");
     }
 
     @Override
     protected boolean isInterpolatableEdge(EdgeIteratorState edge) {
-        //TODO We should not hardcode this string here
-        return edge.get(roadEnvironmentEnc).equals("bridge");
+        return edge.get((IntEncodedValue) roadEnvironmentEnc) == bridgeRef;
```
karussell: It is important to move the cast into the constructor to avoid performance issues.

aramzl:
Hi @karussell, @boldtrn,

I created a pull request (#1269) to add the lane information based on this encoder refactoring. The target branch for the request is to this branch `encoder_refactoring`.
I have some comments or remarks about the implementation:

1. TagParserFactory is already big and it could be split up, e.g. by car, foot, etc.
2. It is hard to tell when you have to update the FlagEncoder and TagParsers and how long will we have both on the code
3. In the class `InstructionsFromEdges` the encodedValues should be grouped and not passed every one of them in the constructor, every time a new one is added the signature has to be changed.
4. To map also the `turn:lanes:backward` I need more than 8 bytes, but currently is not allowed, so I only implemented support for `turn:lanes` and `turn:lanes:forward`
5. The number of required bytes for the `EncodedValue` depends on the order of processing of the TagParsers i.e. if we need 21 + 28 + 5 = 3 bytes are needed but if we do 21 + 5 + 28 = 2 only are needed. I used a LinkedList but a general solution like ordering by required bytes is needed.

Please take a look at the PR, any comments are appreciated.

aramzl:
I added a test to proof the last point (5.) called `EncodedValueTest`

aramzl:
Also the branch has to be rebased soon because there are a lot of changes in master. =)

karussell:
Thanks a lot for taking the time to rewrite it and reading though this PR - will have a closer look into your PR! (but probably not this and the next week I fear)

> TagParserFactory is already big and it could be split up, e.g. by car, foot, etc.

Yes. But I'm not sure if we should split it by profile or else, as IMO splitting by vehicle profile does not make much sense, where do you put the max speed? max_speed is of course car related but can be useful for bike too (avoid high speed roads) or similar to max_width which is not only relevant for trucks but also cargo bikes. So the current structure with subclasses is suboptimal and I would like to find a better structure. Any ideas?

> It is hard to tell when you have to update the FlagEncoder and TagParsers and how long will we have both on the code

Good point. FlagEncoder will disappear when we found alternative barrier and turn cost handling, which are big tasks (but unfortunately not the only big tasks - will have to revisit this PR in the next week)

> I used a LinkedList but a general solution like ordering by required bytes is needed

Yes, this can be done in a future optimization.

> I need more than 8 bytes, but currently is not allowed

Hmmh this should be easy to incease in this PR

> Also the branch has to be rebased soon because there are a lot of changes in master. =)

Yes, big issue. But this gets prio 2 this year, directly after prio 1 issue #1108 :)

HendrikLeuschner:
We are currently looking into replacing our own edge information storages on OpenRouteService with this ansatz here. As we do have some rather spezialized storages (e.g. mountain bike trail difficulty or wheelchair), one key question is how easily these can be added to the new framework.
I guess it is not your intention to have all these niche storages in the core, so is there a way (or the intention to implement a way) to extend the existing encoders easily with subclasses or -methods?

karussell:
> one key question is how easily these can be added to the new framework

This should be really easy if the information is edge-based, if not let me know :)

The first step is: forget about FlagEncoders (at least in the first place as we do not need them anymore but they are still working). Instead think about EncodedValues that we can put&get into the storage with the help of the EncodingManager. So create your custom EncodedValue (potentially also with a TagParser) and add it to the EncodingManager:

encodingManager.add(tagParser or encodedValue).

Then use the encoded values in your custom Weighting e.g. see [this](https://github.com/graphhopper/graphhopper/blob/8dd658c5ae48e8a690a90ed6c9135d4ae7d7479a/core/src/main/java/com/graphhopper/routing/weighting/FastestCarWeighting.java). Then fetching and setting is easy:

edge.get(encValue) and edge.set(encValue, intVal)

> As we do have some rather spezialized storages (e.g. mountain bike trail difficulty or wheelchair)

I think we could add also "special case"-EncodedValues to the core as we don't have to enable them but users could then configure their bike profile to consider this information. The main problem is how we structure these hundreds of different EncodedValues and parsers. Currently we have a single & huge TagParserFactory only, maybe we create different files or implement a service loader mechanism

HendrikLeuschner:
Thanks for the reply. I suppose a service which runs through various specialized tagparsers would be a good idea. This way modular expansion of the storage is easier (I guess). We will start refactoring our stuff into this new logic. Do you have an estimate on the merge into master of a first version?

karussell:
> Do you have an estimate on the merge into master of a first version?

No, but it will be the next big one after the dropwizardification is merged. There are too many unresolved things: do we handle relation and node information similar and what about performance?

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -238,102 +252,70 @@ public long handleNodeTags(ReaderNode node) {
...
             throw new IllegalArgumentException("Speed cannot be negative or NaN: " + speed
-                    + ", flags:" + BitUtil.LITTLE.toBitString(flags));
+                    + ", flags:" + BitUtil.LITTLE.toBitString(ints));
 
-        if (speed < speedEncoder.factor / 2)
-            return setLowSpeed(flags, speed, false);
+        if (speed < speedFactor / 2) {
+            averageSpeedEnc.setDecimal(false, ints, 0);
+            accessEnc.setBool(false, ints, false);
+            // accessEnc.setBool(true, ints, false);
```
boldtrn: Why was this commented out? I just removed this comment and this fixed `CarFlagEncoderTest#testSetSpeed0_issue367`. Also it seems reasonable that if setting speed ~= 0 that access is false, except for Encoders with speeds per direction.

This was commented out [here](https://github.com/graphhopper/graphhopper/commit/8dd658c5ae48e8a690a90ed6c9135d4ae7d7479a). The number of failing tests did not change (except for `CarFlagEncoderTest#testSetSpeed0_issue367`).

karussell: speed 0 vs. accessibility is a problem, because they usually depend on each other but not always (traffic jam). If we would have one speed value per direction we wouldn't have this problem but this would reduce available bits. Though problem.

boldtrn: > Lets assume we have one speed value per direction (2 per edge) then the previous state was correct

Are you sure that the previous state was correct? The previous state always set the forward access to false and the backward access got commented out, but was in there before. Shouldn't we first try not to change any concepts (ideally leaving most tests as they are) and just get it working before we start changing concepts?

boldtrn:
How should we handle Priorities? For example `BikeFlagEncoderTest#testCalcPriority` currently fails, because we now use the `PriorityEncodedValue` instead of `FlagEncoder#getDouble`, in `FlagEncoder#getDouble` we did [`priorityWayEncoder.getValue(flags) / BEST.getValue()`](https://github.com/graphhopper/graphhopper/blob/69878067b7084579a0f9662997055a2d3b8cf2a0/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java#L659). For now I set the factor of the DecimalEncodedValue to 1/Priority.Best, which works. Did you plan to do it differently?

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
karussell: Yes, avoiding changing concepts for the deprecated FlagEncoder API makes 100% sense.

karussell:
I have not yet thought about priorities. But the storage layer should just store the value and not doing some normalization or similar. So storing this via `priority/BEST` might one way or to avoid thinking about this we might think about a normalization factor in the EncodedValue or a new EncodedValue class or something good :)

HendrikLeuschner:
We at openrouteservice are looking forward to this pull request so we can adapt our code. We would like to help out to make these changes possible asap. Is there a part of the code you need help implementing @karussell @boldtrn ?

karussell:
Great to hear this. My current 'internal' - probably confusing - TODO list is:

 * move a few things into separate PRs, like I've done with #1291 already
 * probably revert the PathExtract separation
 * think about how to solve the problem regarding too many EncodedValue parsers in one factory class
 * implement a similar mechanism for relation flags too (could also go into a separate PR). Same for node-based information
 * performance tuning, especially since the getData method creates new object. Instead we could create only one IntsRef per EdgeExplorer for traversal or even less.
 * should we keep FlagEncoders? We need to keep the concept of vehicle profiles but more like a simple list of EncodedValues and possible weightings and not as special classes. Not sure.
 * remaining issues at the top of this PR (only some of them are critical)
 * solve problem with two bits for access restriction versus speed==0 (probably also in a separate PR)
 * make all tests pass :)

(will update this list once I'll invest more time into this after we merge #1314 really soonish)

HendrikLeuschner:
We would start working on 
> think about how to solve the problem regarding too many EncodedValue parsers in one factory class
In ORS, we solve this problem by employing one superclass that iterates over all (enabled) Parsers which are thematically structured. This means we have e.g. a parser for tollways, a parser for trail difficulty, a parser for hgv tags etc. It should then be possible to configure which attributes are enabled during preprocessing and query time. If this sounds reasonable to you @karussell , we will get going.

karussell:
Sounds good. Do you have a pointer into this class?

HendrikLeuschner:
Sure.
These are the different builders: [Builders](https://github.com/GIScience/openrouteservice/tree/master/openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/storages/builders)
They are employed via our custom [ORSOSMReader ](https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/ORSOSMReader.java) which in turn calls the [Processingcontext](https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/src/main/java/heigit/ors/routing/graphhopper/extensions/GraphProcessContext.java), which is a custom class, where the processWay function calls the builders of all enabled processors.
Currently we have changed the normal OSMReader so that it contains an empty method onProcessWay that is extended in our custom ORSOSMReader. We would choose a similar approach where we employ the calls in the Reader?

HendrikLeuschner:
> I do not understand why a new layer (GraphProcessContext) and the changes to the OSMReader is necessary, also the naming is unspecific.
Sorry, I was not clear on this. This is how we do it right now. I will ditch both the Reader and the GraphProcessContext and implement everything in your existing classes + the new classes for separate TagParsers.

karussell:
Ah, no problem. Sounds great, thanks :) !

HendrikLeuschner:
I've rewritten the structure a bit. All TagParsers (though I have not added all yet) now reside in their respective own class: [Tagparsers](https://github.com/HendrikLeuschner/graphhopper/tree/encoder_refactoring/core/src/main/java/com/graphhopper/routing/profiles/tagparsers)
The needed attributes (like surfaces list) are integrated to ensure good modularity. I did not 
The [TagParserFactory](https://github.com/HendrikLeuschner/graphhopper/blob/encoder_refactoring/core/src/main/java/com/graphhopper/routing/profiles/TagParserFactory.java) now does not define profiles anymore but only creates single modules. The overview over enabled parsers and the profile definitions are now inside [encodingmanager](https://github.com/HendrikLeuschner/graphhopper/blob/encoder_refactoring/core/src/main/java/com/graphhopper/routing/util/EncodingManager.java) via a simple list of which parsers should be enabled. The Encodingmanager also checks whether a parser has already been added and has methods to add bikeParsers, footParsers and carParsers. I don't really like putting the profiles in here, I think they should be moved to a different class or configuration file.
So far I find this structure much easier to understand and modify, but I wanted to get your opinion(s) before I continue further.

HendrikLeuschner:
> Now ỳou could remove most of the methods in TagParserFactory or maybe the whole class?
Yes, this is now mostly done. Entire thing is of course still a work in progress.

>  Is such a check necessary if we have the duplicate check for the EncodedValue objects already?
No, you are right.

I fixed some things and added most of the remaining parsers, except for a few where I was not yet sure how to handle it (namely SpatialRuleId). Cleaned up some of the files too. I will continue to work on this.

HendrikLeuschner:
Hi @karussell , are there in principal different encoded values for all profiles, meaning e.g. bike and bike2 have different encodedValues for average_speed and access etc? Including different tag parsers?
Running into this problem while running the tests. One of them fails while trying to find the ev bike2.access for the Bike2WeightFlagEncoder.
If they are shared, there is a need to distinguish the parsers too, possibly without duplicating them.

karussell:
> If they are shared, there is a need to distinguish the parsers too, possibly without duplicating them

To make it backward compatible average_speed and access values cannot be shared, yes. E.g. bike2 modifies the speed due to elevation and racing bike is faster, but also access is different as mtb can use paths with a more difficult grade for example. Although I can imagine that making the access shared across bikes should be possible as one can use the Weighting to just avoid these paths (but the space requirement is very low and so this optimization should be done later)

In the future one could further try to calculate average_speed dynamically from other generic values but especially for non-CH use cases a precomputed value for both is faster while query.

HendrikLeuschner:
I have not yet created all EncodedValues for all profiles, e.g. MotorcycleFlagEncoder. Therefore I have more failing tests than before (Without my changes its 142, with them its 216). I don't know when or if I will have the time to do add every remaining profile. Would you like me to stage a pull request anyway? Sorry for all the questions.

karussell:
> Therefore I have more failing tests than before (Without my changes its 142, with them its 216). 

I'm not sure if I understand this but a PR will make it clear probably.

> I don't know when or if I will have the time to do add every remaining profile.

Sure

> Would you like me to stage a pull request anyway? 

Yes, please. You can add [WIP]

> Sorry for all the questions.

No need for that really :)

HendrikLeuschner:
Hi,

any update on this so far? 
Best, Hendrik

karussell:
I'm currently in the process of defining what should go in this PR and what not. 

Here are the main goals of this PR:

 * edge storage should allow more than 64bit 
 * easier storing of edge properties
 * EncodingManager will be EncodedValues based
 * avoid reverseFlags

non-goals:

 * remove FlagEncoders
 * fix PathExtract
 * flexible storage for relation and node
 * solve problem with two bits for access restriction versus speed==0

And so I will remove the AccessValue refactoring and create a separate PR, see this separate branch: https://github.com/graphhopper/graphhopper/tree/access_refactoring

Furthermore the json refactoring won't happen in this PR: https://github.com/graphhopper/graphhopper/pull/1291#issuecomment-386907844 and we'll use a string based EncodedValue properties storage somehow.

Now I'll try to merge your PR and rebase the changes against the access_refactoring branch. As a first step after this bigger change I'll try to fix the performance problem (currently we are 2x slower) and in parallel we could work on making more tests green.

HendrikLeuschner:
Currently osm tags are only processed before edges are created via parse(IntsRef ints, ReaderWay way). For some of our information we need a parser that is called after edge creation, such as parse(EdgeIteratorState edge, ReaderWay way).
Do you think we could include something like this in the osm reader after edge creation?

On a different node, the optimal case for us would be to have a possibility of defining TagParsers on the fly even from outside the graphhopper repo. This should be some sort of method like TagParserFactory.add(TagParser p) or equivalent in EncodingManager, so that we can maintain our own parsers. I think this would be optimal for other projects built on graphhopper too.

karussell:
This PR is too big. I'll split it into separate ones:

 * access_refactoring
 * move from long to IntsRef and avoid that performance will suffer, maybe fetch ints for the whole edge at once
 * EncodingManager will be EncodedValues based and we avoid calling reverseFlags for every edge access

karussell:
Closing in favor of #1447

PR #1120, opened 2017-07-30
Allow Foot, Hike, and Bike to enter Piers.

boldtrn:
As discussed in the [forum](https://discuss.graphhopper.com/t/adding-a-new-way-type-to-foot/2223), GraphHopper does not use piers for foot routing.

Piers are important to consider, for example to access ferries. Usually ferries also allow transportation of bikes, so I allowed piers also for bicycles.


boldtrn:
The failing tests were quite interesting. Since there seems to be a pier in our Monaco tests data, some details like visited nodes or ids changed. Fixed that. No actual routing result changed.

karussell:
Nice - thanks a lot!

PR #1124, opened 2017-08-01
Check access restrictions for ways with highway==null

### LINKED ISSUE #1119: Check restrictions for platform
In FootFlagEncoder (also Hike), we check `way.hasTag("railway", "platform")`, if yes we accept it for foot routing, but we don't check if there are possible restrictions tagged like access=no. Not sure if this is relevant but there might be reasons to still check this?

### LINKED ISSUE #1122: issue with private=access on ferries
We got a bug report via email:

> I'm curious how access=private is handled in Graphhopper because when
trying to use your engine on OpenStreetMap I was routed via a ferry
(which at that time was unspecified). However, after adding a tag saying
it is access=private I am still routed over it.
> See: https://www.openstreetmap.org/directions?engine=graphhopper_bicycle&route=64.4893%2C19.3299%3B64.5037%2C19.2600#map=13/64.4941/19.2954
> and the ferry way: https://www.openstreetmap.org/way/221099681

boldtrn:
Fixes #1122. Fixes #1119.

I added a check for restrictions, if highway is not set. If any restrictions apply, we don't accept the way.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
@@ -200,18 +200,26 @@ public class FootFlagEncoder extends AbstractFlagEncoder {
     public long acceptWay(ReaderWay way) {
         String highwayValue = way.getTag("highway");
         if (highwayValue == null) {
+            long acceptPotentially = 0;
+
             if (way.hasTag("route", ferries)) {
                 String footTag = way.getTag("foot");
                 if (footTag == null || "yes".equals(footTag))
-                    return acceptBit | ferryBit;
+                    acceptPotentially = acceptBit | ferryBit;
             }
 
             // special case not for all acceptedRailways, only platform
             if (way.hasTag("railway", "platform"))
-                return acceptBit;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
```diff
@@ -200,18 +200,26 @@ public long getTurnFlags(boolean restricted, double costs) {
     public long acceptWay(ReaderWay way) {
         String highwayValue = way.getTag("highway");
         if (highwayValue == null) {
+            long acceptPotentially = 0;
+
             if (way.hasTag("route", ferries)) {
                 String footTag = way.getTag("foot");
                 if (footTag == null || "yes".equals(footTag))
-                    return acceptBit | ferryBit;
+                    acceptPotentially = acceptBit | ferryBit;
```
karussell: It does not look right as we do only "return acceptBit" at the end.

boldtrn: Oops, you are right. I also added a test for the FootFlagEncoder.

PR #1126, opened 2017-08-04
Equal Ferry Speeds between Encoders

### LINKED ISSUE #1116: Ferry Duration Times Inconsistent for Cars, Bicycles and Pedestrian
I've been adding in ferry duration times in OSM for ferry routes in Washington State. Testing out the results on Graphhopper, I noticed that the router gives different times on the same ferry route for cars, bicycles and pedestrians. For example the Mukilteo-Clinton Ferry (5901929)1 shows the following results using the graphhopper.com/maps:

* Car 4.42km will take 44min
* Bicycle 4.42km will take 26min
* Pedestrian 4.42km will take 29min

The actual duration of the ferry route is 20 minutes which should be the same for all modes of transportation.

boldtrn:
Fixes #1116. 

I removed the parameters with individual ferry speeds for the method getFerrySpeed. The speeds are now stored in the AbstractFlagEncoder.

The FootFlagEncoder, however, does not not allow to store speeds > 10 km/h. So I changed the max speed for the FootFlagEncoder to 15 km/h (still 4 bits). If the stored speed speed equals 15 km/h I return 20 km/h for the speed instead. This is the short_trip_ferry_speed. 

I increased the versions for Foot, as well as HikeEncoder, not for the others, as I didn't change anything with the EncodedValues for the others.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
@@ -469,8 +469,8 @@ public abstract class AbstractFlagEncoder implements FlagEncoder, TurnCostEncode
                             return getMaxSpeed();
                         }
                         // If the speed is lower than the speed we can store, we have to set it to the minSpeed, but > 0
-                        if (Math.round(calculatedTripSpeed) < speedEncoder.factor) {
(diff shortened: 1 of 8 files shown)


karussell:
Looks good - perfect also that you thought about increasing the versions :+1: !

PR #1129, opened 2017-08-04
graphhopper.sh: let env decide $CONFIG

justinmk2:
If there is concern about a spurious CONFIG in the environment (because the name is so generic)  could look for GH_CONFIG instead:

    CONFIG=${GH_CONFIG:-config.properties}

### DIFF
--- a/graphhopper.sh
+++ b/graphhopper.sh
@@ -16,8 +16,8 @@ if [ "$bit64" != "" ]; then
 fi
 echo "## using java $vers from $JAVA_HOME"
 
-CONFIG=config.properties
-if [ ! -f "config.properties" ]; then
+CONFIG=${CONFIG:-config.properties}
+if [ ! -f "$CONFIG" ]; then
   cp config-example.properties $CONFIG
 fi
 


karussell:
Nice & Thanks! So this bash magic uses either $CONFIG or if empty it uses config.properties? This would be nice in other places as well :) !

Regarding first time contribution: would you sign your commits or [contact us](https://github.com/graphhopper/graphhopper/blob/master/.github/CONTRIBUTING.md#license-agreement) so that we can send you the CLA?

karussell:
Will be fixed with a slightly different notation like ${CONFIG:?config.properties} in #1314

justinmk2:
Great. I did send a CLA request and never got a reply.

CLA shouldn't be required for such trivial changes anyway.

karussell:
Ah, did you send the CLA one week ago? I replied to it with a question to which PR this email belongs to :) ... so now I know and keep the CLA for future contributions.

Sorry for this ugly first contribution experience - I hope you still enjoy graphhopper and would love to see your second PR :)

> CLA shouldn't be required for such trivial changes anyway

That is a tough thing. If you correct a misspelled word, ok. But as soon as it could be interpreted 'creative act' this is not 100% sure.

karussell:
(Ups, there was no CLA attached, will send the request now ..)

justinmk2:
Well, it's up to you, but not even the FSF requires CLA for minor changes (measured by number of lines, not "creative").

FWIW Joyent/nodejs and many other projects have discarded their CLAs: https://nodejs.org/en/blog/uncategorized/notes-from-the-road/#easier-contribution

karussell:
We don't use the CLA to give us the opportunity to relicense the project. Our aim is the same like described here https://blog.chef.io/2009/08/11/why-we-chose-the-apache-license/ which is very similar to other foundations like Apache - they would never use a different license and still require the CLA.

> Well, it's up to you, but not even the FSF requires CLA for minor changes (measured by number of lines, not "creative").

IMO this is not as easy

karussell:
But this CLA stuff is of course debatable because if you are contributing it could be seen as implicitly accepting that the contribution is under the same license as the project. And if we then even include a note regarding the projects license in the text that everyone sees who does a pull request we would be safer. At least for such 'small' changes (according to our lawyer ;))

justinmk2:
> Our aim is the same like described here https://blog.chef.io/2009/08/11/why-we-chose-the-apache-license/ which is very similar to other foundations like Apache - they would never use a different license and still require the CLA.

That post is from 2009, around the time when lots of projects (including nodejs) tried the CLA thing.

Chef no longer requires a CLA: https://blog.chef.io/2016/09/19/introducing-developer-certificate-of-origin/

karussell:
Interesting, thanks! 

Which line number is the FSF using?

karussell:
Cool. We'll adopt this way, especially with the electronic signatures for the commits this is easier.

karussell:
Difficult topic, really. The "chef" procedure is partially discussed in #1208 and sometimes relies on "higher" git knowledge (signing commits, rebase if necessary multiple commits) and the name and email in clear text. Signing the CLA is a no-brainer for everyone as we do this via signaturit.com and we associate the github nickname with those PDFs in the background.

So I'm really not sure if this would lower the entry barrier.

karussell:
This comment sums it up pretty good: https://github.com/chef/chef-rfc/pull/213#issuecomment-239296547

PR #1135, opened 2017-08-09
Added documentation for turn restrictions

boldtrn:
Due to the latest [forum request](https://discuss.graphhopper.com/t/turns-restrictions/2275) about turn restrictions, I realized that we don't have a documentation explaining how to use turn restrictions, so I thought it might make sense to add a documentation.

Let me know if I should add anything.


boldtrn:
Oh no, something went wrong here with my branches, will create a new PR...

PR #1136, opened 2017-08-09
Added documentation on how to use turn restrictions

boldtrn:
Due to the latest [forum request](https://discuss.graphhopper.com/t/turns-restrictions/2275) about turn restrictions, I realized that we don't have a documentation explaining how to use turn restrictions, so I thought it might make sense to add a documentation.

Let me know if I should add anything.


docs/core/turn-restrictions.md
```diff
@@ -0,0 +1,19 @@
+# How to work with Turn Restrictions
+
+GraphHopper supports [turn restrictions](http://wiki.openstreetmap.org/wiki/Relation:restriction).
+Currently, it is not possible to use turn restrictions with the speed mode (contraction hierarchies),
+but it's possible using the flexible mode (landmarks) and without any post-import optimization.
```
karussell: can we call it flexible and hybrid mode and maybe linking to [the explanation](https://github.com/graphhopper/graphhopper#technical-overview)? Technically the landmarks algo is flexible but I avoid calling it flexible mode :)

docs/core/turn-restrictions.md
```diff
@@ -0,0 +1,19 @@
...
+Turn restrictions are crucial for correct vehicle navigation and help to avoid forbidden turn.
```
karussell: **a** forbidden turn or forbidden turn**s**?

docs/core/turn-restrictions.md
```diff
@@ -0,0 +1,19 @@
...
+![turn without turn restrictions](./images/turn-restrictions-wrong.png)
+
+With turn restrictions:
+
+![turn with turn restrictions](./images/turn-restrictions-correct.png)
+
+Turn restrictions have to be enabled on a vehicle basis. To enable it for one vehicle add
+`|turn_costs=true` in the config, for example: `graph.flag_encoders=car|turn_costs=true`.
+Turn restrictions are not available for every vehicle as they have low relevance
+for some vehicles like `foot`.
```
karussell: I would also add: if you have the speed mode enabled on import (e.g. prepare.ch.weightings=fastest) you'll need to disable the speed mode per request via `&ch.disable=true`

docs/core/turn-restrictions.md
boldtrn: Sure, sorry, I mixed up the terms :)

boldtrn:
Thanks for the comments, I updated the PR.

PR #1138, opened 2017-08-17
PathDetails issue when placing the waypoint exactly on a PathDetail change

boldtrn:
I found the issue here. If we place a waypoint at exactly the position where the PathDetails change, the PathDetails are not merged. The index we passed to the PathDetail generation was too big by 1 index. So this was not necessarily a Roundtrip Issue.


boldtrn:
The trick here was again the +1 for via points. Since our merge method corrected most issues at via points by merging the last and first PathDetail of adjacent Paths, if they have the same value. This value only pops up if the last PathDetail of one Path has a different value than the first PathDetail of the next Path.

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -153,6 +153,8 @@ public static void merge(List<PathDetail> pathDetails, List<PathDetail> otherDet
         // Make sure that the PathDetail list is merged correctly at via points
         if (!pathDetails.isEmpty() && !otherDetails.isEmpty()) {
             PathDetail lastDetail = pathDetails.get(pathDetails.size() - 1);
+            // Add +1 for the via point
+            lastDetail.setLast(lastDetail.getLast()+1);
```
karussell: Hmmh, this feels wrong and we should probably skip adding the via points opposed to my recent comment in #1139?

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -99,6 +99,8 @@ public void doWork(PathWrapper altRsp, List<Path> paths, Translation tr) {
                     if (pathIndex + 1 < paths.size()) {
                         ViaInstruction newInstr = new ViaInstruction(fullInstructions.get(fullInstructions.size() - 1));
                         newInstr.setViaCount(pathIndex + 1);
+                        // Similar to the FinishInstruction, ViaInstructions have lenght=0
```
boldtrn: We could also enforce this during the Finish Instruction creation [here](https://github.com/graphhopper/graphhopper/blob/091c16e5f78acf9cb32ce2b3e181cac9cae228c5/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java#L230)

core/src/main/java/com/graphhopper/util/PointList.java
```diff
@@ -242,14 +242,23 @@ public void add(GHPoint point) {
     }
 
     public void add(PointList points) {
-        int newSize = size + points.getSize();
+        this.add(points, 0, points.size());
+    }
+
+    public void add(PointList points, int from, int to) {
```
boldtrn: I added this method to have a better performance compared to first copy the PointList and then add it.

boldtrn:
I updated this PR to remove the duplicated point in the PointList

core/src/main/java/com/graphhopper/util/PointList.java
karussell: Looks good. Another idea would be removeLast via size--

core/src/main/java/com/graphhopper/util/PathMerger.java
karussell: I do not fully understand it. The ViaInstruction and FinishInstruction both should have one point, why is setting to an empty list necessary? And why do we still need this viaPosition/viaCount workaround?

boldtrn: As I understood it, they should have no points. If they have no points, we create intervals like [9,9]. If there is one point, we create an interval [9,10]. If we leave the point, we have to create a workaround for the simplification as well as for the serialization.

boldtrn: I just updated the PR to show an alternative approach. Now the Instructions contain one point, but we ignore these points during the simplification and serialization. I think it looks a bit cleaner like this.

core/src/main/java/com/graphhopper/util/PointList.java
boldtrn: You are right, might be the better solution. I did it like this, because at first I wanted to remove the first point of the PointList, which turned out to be the worse approach :). Once it was implemented doing the other war around worked easy. Should we switch to the `removeLast` approach?

core/src/main/java/com/graphhopper/util/PathMerger.java
karussell: > If there is one point, we create an interval [9,10]

But that is two points?

> If we leave the point, we have to create a workaround for the simplification as well as for the serialization.

Why? Simplification should be disabled for <3 points.

core/src/main/java/com/graphhopper/util/PointList.java
karussell: I'm fine with the current state

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -141,7 +141,12 @@ public PointList simplify() {
     private int getLength(Object o, int index) {
         if (o instanceof InstructionList) {
             // we do not store the last point of an instruction
-            int size = ((InstructionList) o).get(index).getPoints().size();
+            Instruction instruction = ((InstructionList) o).get(index);
+            int size;
+            if(instruction instanceof ViaInstruction)
+                size = 0;
+            else
+                size = instruction.getPoints().size();
```
karussell: Here I would expect 'getPoints.size-1'

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -99,8 +99,6 @@ public void doWork(PathWrapper altRsp, List<Path> paths, Translation tr) {
                     if (pathIndex + 1 < paths.size()) {
                         ViaInstruction newInstr = new ViaInstruction(fullInstructions.get(fullInstructions.size() - 1));
                         newInstr.setViaCount(pathIndex + 1);
```
karussell: can we remove setViaCount as getViaCount is unused?

core/src/main/java/com/graphhopper/util/PathSimplification.java
boldtrn: For the ViaInstruction it shouldn't matter as it should be only one point anyway. For the other case it should be the size IMHO. Could you explain why it should be size-1?

core/src/main/java/com/graphhopper/util/PathMerger.java
boldtrn: > can we remove setViaCount as getViaCount is unused?

The viaCount is used in the `getTurnDescription` mehtod of the ViaInstruction.

core/src/main/java/com/graphhopper/util/PathMerger.java
boldtrn: > But that is two points?

No. The second point ref of the interval is equal to the first point of the next interval. Intervals are generated by `[ START_REF, START_REF + POIN_COUNT ]` for all Instructions,
expect the FinshInstruction where we artificially set the point count to 0 in the SimpleRouteSerializer. This PR enforces this behavior for all ViaInstructions and FinishInstructions.

Due to that they don't have their own point in the points array. That's why we have to create the workaround with lenght=0 in the PathSimplification.

core/src/main/java/com/graphhopper/util/PathSimplification.java
karussell: Consistency. All path details currently work this way: you specify first and last and the length is last-first. Now imagine an instruction 0-1-2-3-4, where e.g. at 3 there is the via point, then the first instruction includes point 3, i.e. it is `interval:[0,3]` and therefor the length is 3-0=3, although the pointlist.size is 4.

core/src/main/java/com/graphhopper/util/PathMerger.java
karussell: Ah, ok. Makes sense. Is it possible to move this to the constructor?

core/src/main/java/com/graphhopper/util/PathMerger.java
karussell: Hmmh. Let me summarize how it **should** be: every instruction has a first point and a last one, obviously the last point is the first point of the next instruction and the json would be `interval:[first_ref,last_ref]` and `length=last_ref-first_ref`. For Finish and ViaInstruction the result is length=0 (due to first==last) and pointlist.size==1 which makes still sense to me.

Everything which contradicts to this should be fixed e.g. the current ViaInstruction `interval` etc. and/or make the Instructions like the JSON output (and also PathDetails).

boldtrn:
I just updated this PR. The latest push is only WIP, and it's not even building. I just commited to show you the current approach I am taking. Essentially I tried to remove everything point related and move it up to the InstructionList, so the InstructionList contains the whole PointList for a Path and can get Points for a single Instruction. The Instruction only stores references. This breaks the current API, but I think it's a nice improvement :).

Let me know what you think or if we should follow a different Path.

core/src/main/java/com/graphhopper/util/FinishInstruction.java
```diff
@@ -21,30 +21,10 @@
...
-
-    public FinishInstruction(String name, PointAccess pointAccess, int node) {
-        this(name, pointAccess.getLatitude(node), pointAccess.getLongitude(node),
-                pointAccess.is3D() ? pointAccess.getElevation(node) : Double.NaN);
-    }
 
-    public FinishInstruction(PointAccess pointAccess, int node) {
-        this(pointAccess.getLatitude(node), pointAccess.getLongitude(node),
-                pointAccess.is3D() ? pointAccess.getElevation(node) : Double.NaN);
+    public FinishInstruction(int pointRef) {
```
karussell: Is pointRef == last == first for the FinishInstruction?

core/src/main/java/com/graphhopper/util/Instruction.java
```diff
@@ -40,24 +39,30 @@
...
      * The points, distances and times have exactly the same count. The last point of this
      * instruction is not duplicated here and should be in the next one.
      */
-    public Instruction(int sign, String name, InstructionAnnotation ia, PointList pl) {
+    public Instruction(int sign, String name, InstructionAnnotation ia, int first) {
         this.sign = sign;
         this.name = name;
-        this.points = pl;
         this.annotation = ia;
+        this.first = first;
```
karussell: Can we make first final?

core/src/main/java/com/graphhopper/util/Instruction.java
```diff
@@ -197,45 +136,31 @@ public String toString() {
...
+    public void setLast(int last) {
+        this.last = last;
+    }
 
-        double lat = points.getLatitude(0);
-        double lon = points.getLongitude(0);
-        return AC.calcAzimuth(lat, lon, nextLat, nextLon);
+    public int getLast() {
+        if (last < first)
+            throw new IllegalStateException("last cannot be smaller than first");
```
karussell: checkOne can be used or is unused?

core/src/main/java/com/graphhopper/util/InstructionList.java
```diff
@@ -334,4 +424,15 @@ public Instruction find(double lat, double lon, double maxDistance) {
         return get(foundInstruction);
     }
 
+    public void setPoints(PointList points) {
+        this.points = points;
+    }
+
+    public PointList getPoints(Instruction instruction) {
```
karussell: Maybe a different name or some javadocs are important to see the difference between this method and getPoints()?

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -90,7 +90,7 @@ public void doWork(PathWrapper altRsp, List<Path> paths, Translation tr) {
             fullDistance += path.getDistance();
             fullWeight += path.getWeight();
             if (enableInstructions) {
-                InstructionList il = path.calcInstructions(tr);
+                InstructionList il = path.calcInstructions(tr, origPoints);
```
karussell: A better name now for origPoints might be origPointCount (?) as I initially thought it is the full PointList that must be injected into every InstructionList ...

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -98,14 +98,21 @@ public PointList simplify() {
...
+                                int first = 0;
+                                int last = 0;
+                                if (o instanceof PathDetail) {
+                                    first = ((PathDetail) o).getFirst();
+                                    last = ((PathDetail) o).getLast();
                                 }
+                                if (o instanceof Instruction) {
+                                    first = ((Instruction) o).getFirst();
+                                    last = ((Instruction) o).getLast();
+                                }
```
karussell: here an `if&else if` might make sense with and else that throws an IllegalStateException as the list is of an unknown type.

core/src/main/java/com/graphhopper/util/ViaInstruction.java
```diff
@@ -23,23 +23,17 @@
...
     }
 
-    public int getViaCount() {
-        if (viaPosition < 0)
-            throw new IllegalStateException("Uninitialized via count in instruction " + getName());
-
-        return viaPosition;
-    }
-
     public void setViaCount(int count) {
```
karussell: this can be removed now (?)

core/src/main/java/com/graphhopper/util/FinishInstruction.java
boldtrn: Yes

core/src/main/java/com/graphhopper/util/Instruction.java
boldtrn: No, as we change the first reference during the simplification. I would have liked that as well.

core/src/main/java/com/graphhopper/util/Instruction.java
boldtrn: :+1:

core/src/main/java/com/graphhopper/util/InstructionList.java
boldtrn: Yes certainly, it's still pretty raw and I am trying to get everything working, as it might involve bigger changes at some points in the code (not 100% sure yet).

core/src/main/java/com/graphhopper/util/PathSimplification.java
boldtrn: I actually want the Instruction and PathDetail implement the same interface, that all provide getLast, getFirst, getLength, etc. so we could solve most of this without any instanceof.

karussell: For now I would have no problem with instanceof.

core/src/main/java/com/graphhopper/util/InstructionList.java
```diff
@@ -429,10 +436,52 @@ public void setPoints(PointList points) {
     }
 
     public PointList getPoints(Instruction instruction) {
+        if (instruction == null)
+            return PointList.EMPTY;
+        if (instruction.getLength() == 0)
+            // Copy does not include the last point
+            return this.points.copy(instruction.getFirst(), instruction.getLast() + 1);
         return this.points.copy(instruction.getFirst(), instruction.getLast());
```
karussell: We should always include the last point, even if we break backward compatibility.

core/src/main/java/com/graphhopper/util/InstructionList.java
```diff
@@ -429,10 +436,52 @@ public void setPoints(PointList points) {
...
     }
 
     public PointList getPoints() {
         return this.points;
     }
+
+    /**
+     * Appends the insturctionList to this InstructionList, from instruction index 0 to toIndex
```
karussell: minor typo: instructionList

core/src/main/java/com/graphhopper/util/InstructionList.java
```diff
@@ -417,9 +410,23 @@ public Instruction find(double lat, double lon, double maxDistance) {
...
+
+        int foundInstruction = -1;
+
+        for (int instructionIndex = 0; instructionIndex < size(); instructionIndex++) {
+            Instruction instruction = get(instructionIndex);
+            if (instruction.getFirst() <= foundPointIndex && instruction.getLast() > foundPointIndex) {
+                foundInstruction = instructionIndex;
+                break;
+            }
+        }
```
karussell: Is this logic change of introducing two for-loops necessary or to make it easier?

core/src/main/java/com/graphhopper/util/ViaInstruction.java
karussell: Ups, I see :) ... we still could pass it via the constructor and make it final (?)

boldtrn:
> Yes, I think this is more consistent in a few places and already align with the JSON API - thanks!

Thanks for the feedback, I updated the PR. I got everything working except for the gtfs and web module so far. The web module depends on the gtfs module, so I couldn't build it without fixing the gtfs module.

@michaz, maybe you can give me a hint on how to fix the issues in the `TripFromLabel` class? Especially tricky for me was the method `InstructionList getInstructions(Translation tr, List<Trip.Leg> legs)`.

Why is there a special handling for the last leg [here](https://github.com/graphhopper/graphhopper/blob/e051df2c02f23d943af0906b38bdda5413268655/reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/TripFromLabel.java#L146)? Because of the FinishInstruction?

Why do we use `instructions.size()-2` [here](https://github.com/graphhopper/graphhopper/blob/e051df2c02f23d943af0906b38bdda5413268655/reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/TripFromLabel.java#L157)? I would have assumed that it would be `instructions.size()-1`, meaning the last instruction, but -2 means the instruction before the last instruction. Is this because the last instruction in this case would be a stop at a station, but we stay in the same train, so we just extend the points of the instruction before instead of creating a new instruction?

What happens [here](https://github.com/graphhopper/graphhopper/blob/e051df2c02f23d943af0906b38bdda5413268655/reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/TripFromLabel.java#L168) with the replace? Are the points of the `arrivalInstruction` the same as the points of the instruction that is replaced?

michaz:
> Why is there a special handling for the last leg here? Because of the FinishInstruction?

Yes.

> Is this because the last instruction in this case would be a stop at a station, but we stay in the same train, so we just extend the points of the instruction before instead of creating a new instruction?

Yes.

> What happens here with the replace? Are the points of the arrivalInstruction the same as the points of the instruction that is replaced?

No. When we are "inSameVehicleAsPrevious", the _previous_ trip already gave an arrival instruction, but we don't want it, so we overwrite it. We only really arrive _now_.

Those are two trips which, according to the schedule, are _technically_ two trips with two trip numbers, but the user does not know this and doesn't care, so we patch the instructions so that they look like one trip.

boldtrn:
Thanks @michaz. 

> Is this because the last instruction in this case would be a stop at a station, but we stay in the same train, so we just extend the points of the instruction before instead of creating a new instruction?
Yes.

Why is the station instruction needed then? Is this working with the current code? Right now the serialization and deserialization (and probably a lot more code as well) assumes that instructions follow each other and are not mixed.

> No. When we are "inSameVehicleAsPrevious", the previous trip already gave an arrival instruction, but we don't want it, so we overwrite it. We only really arrive now.

Would this be the same example as the train switches it's name example as above?

michaz:
> Is this working with the current code?

The test is: GraphHopperGtfsIT.testBlockTrips, and last time I checked, it looked fine in the JS GUI. Maybe have a look? What is supposed to happen in the isSameVehicleAsPrevious-case is just this:

For these Legs:

```
Trip 1 A -> B
Trip 2 B -> C (inSameVehicleAsPrevious == true)
```

create not these instructions

```
Enter at A
Get off at B
Enter at B
Get off at C
```

but these

```
Enter at A
Get off at C
```

... with those instructions covering all the points of the ride. In the normal order.

reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/TripFromLabel.java
```diff
@@ -297,7 +335,7 @@ void finish() {
                     edges(path).map(edgeLabel -> edgeLabel.edgeIteratorState).collect(Collectors.toList()),
                     lineStringFromEdges(path),
                     edges(path).mapToDouble(edgeLabel -> edgeLabel.distance).sum(),
-                    instructions.stream().collect(Collectors.toCollection(() -> new InstructionList(tr))),
```
boldtrn: Was there a reason for this?

boldtrn:
Thanks for the clarification @michaz, I understood it slightly different.

> but these
Enter at A
Get off at C

This should be fine, and thanks for pointing me towards the test.

I finally got the GTFS tests to succeed. Can you live with the changes @michaz? Unfortunately, I wasn't able to do this with a lower impact, as I needed to change more than I would have expected, but maybe there would be a better way to do this?

The PR is still WIP BTW :).

reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/TripFromLabel.java
michaz: You're right, that was left over from something. I'm merging this line (for a start :-) ) to master, thanks.

michaz:
> I finally got the GTFS tests to succeed. Can you live with the changes @michaz? Unfortunately, I wasn't able to do this with a lower impact, as I needed to change more than I would have expected, but maybe there would be a better way to do this?

Looks good to me. I mean, yes, it's more code, but on the other hand, ``Instruction`` now contains what it claims to contain, namely indices, not points.

The super-cleanest thing of all would probably to have a builder which allows creating instructions with the previous syntax, but the new result. But not necessary, from my point.

boldtrn:
This PR has been stale for quite a while now. The biggest problem when converting Instructions to contain only indices in contrast to a PointList, it becomes more expensive to work with the points of an instruction. The code became more complicated and there was a measurable slowdown. This might be improvable with updating the InstructionList code, but for now it's probably not worth the effort. In the future it would be still nice to see this feature merged into master.

michaz:
Hmm, I thought it was ready to merge, wasn't it? I don't even remember, it was so long ago. But look at how much writing and thinking was put in by all of us.

I don't even remember what the problem was which it was solving, but _if_ it was solving something substantial, like allowing a non-realistic corner case which was previously not working, please re-consider merging it.

boldtrn:
> I don't even remember what the problem was which it was solving, but if it was solving something substantial, like allowing a corner case which was previously not working in some surprising way, please re-consider merging it.

Yes parts of this PR (the bug part) has been merged in a separate PR #1240.

The new feature part of this PR was to remove the PointList from the Instructions and only store start and end indexes, similar to the PathDetails. In the long term this could allow to remove the difference between PathDetails and Instructions and to declare instructions just a PathDetail. Overall  I was also hoping to remove some of the overhead generated by splitting the geometry into the instructions and afterwards merging them again. The problem with this is that it didn't feel good in the code, it made many things more complicated, for which I created some helper methods to hide the more complicated parts. Overall it is harder to work with indexes than with PointLists as you need to update the global PointList.

The PR was "working" in a sense that the response looked correct. It was a bit slower than our current approach. Because the InstructionList used many rather "expensive" calls that required us to extract a part of the global PointList. I think this could have been improved and it might become even a bit faster than our current approach.

IMHO there was still some work to do to get this to a mergeable state and I am not sure if the required work justifies the benefit? If you want to I can reopen this PR, update it to the current master and do some more tests?

michaz:
Ohhhhkay, thanks for explaining. That seems fine then. I was only looking at the headline, and that looked worrying to me. Like, when I use a seemingly-general data structure in a certain totally-allowed-looking way, it won't work. Because I'm notoriously the person whose use case includes exactly that way. ;-)

michaz:
(Or do we still have that?)

boldtrn:
> Like, when I use a seemingly-general data structure in a certain totally-allowed-looking way, it won't work.

Do you mean if we still see issues with PathDetails? AFAIK they work fine and I am not aware of any bugs related to them that would prevent anyone from using them or implementing own PathDetails.

michaz:
Okay cool, thanks!

PR #1141, opened 2017-08-21
Updated the Weighting documentation to reflect the current state of GH

boldtrn:
I updated the weighting documentation, as the current code examples were outdated.

### DIFF
--- a/docs/core/weighting.md
+++ b/docs/core/weighting.md
@@ -1,39 +1,46 @@
 ## Weighting
 
-In order to create a custom Weighting you need to do the following:
+A weighting allows to create a "weight" for an edge. The weight of an edge reflects the cost of travelling along this edge. 
+All implementations of [RoutingAlgorithm](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/RoutingAlgorithm.java)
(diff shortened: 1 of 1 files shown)


docs/core/weighting.md
```diff
@@ -1,39 +1,46 @@
 ## Weighting
 
-In order to create a custom Weighting you need to do the following:
+A weighting allows to create a "weight" for an edge. The weight of an edge reflects the cost of travelling along this edge. 
+All implementations of [RoutingAlgorithm](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/RoutingAlgorithm.java)
+in GraphHopper calculate the shortest path, which means the path with the [lowest overall cost from A to B](https://en.wikipedia.org/wiki/Shortest_path_problem). 
+The definition of the cost function is up to you, so you can create a cost function (we call this weighting) for the fastest path.
```
karussell: Yeah, maybe we should rename the Weighting class to CostFunction and calcWeight to calcCost :)

boldtrn: Maybe :), I just thought it makes sense to name it here like that, so if anyone searches for cost function GraphHopper this document would be found.

PR #1142, opened 2017-08-22
Add more PathDetails

### LINKED ISSUE #1140: add more path details
already possible:

 - [x] street_name (important for navigation)
 - [x] time (nice for ETA predictions e.g. if parts of the route were already travelled)
 - [x] edge_id or traversal_key

later when more stuff is stored as edge properties:

 - [x] max_speed (very important for navigation)
 - [x] surface
 - [x] highway type ("road_class")
 - [x] ferry,toll,.. ("road_environment"?)
 - [x] weight, height, ...
 - [ ] inner vs. outer city

boldtrn:
Fixes #1140.

I added all PathDetails that were requested in #1140. In order to add the time, I used the Weighting. I don't especially like this, as it adds an additional dependency, but I think it's the best solution for this problem.

### DIFF
--- a/core/src/main/java/com/graphhopper/util/details/TimeDetails.java
+++ b/core/src/main/java/com/graphhopper/util/details/TimeDetails.java
@@ -1,0 +1,59 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.util.details;
+
+import com.graphhopper.routing.util.FlagEncoder;
+import com.graphhopper.routing.weighting.Weighting;
+import com.graphhopper.util.EdgeIteratorState;
+
+import static com.graphhopper.util.Parameters.DETAILS.TIME;
+
+/**
+ * Calculate the time segments for a Path
+ *
+ * @author Robin Boldt
+ */
+public class TimeDetails extends AbstractPathDetailsBuilder {
+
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/graphhopper/util/details/TimeDetails.java
```diff
@@ -0,0 +1,59 @@
...
+        super(TIME);
+        this.encoder = encoder;
+        this.weighting = weighting;
+    }
+
+    @Override
+    public boolean isEdgeDifferentToLastEdge(EdgeIteratorState edge) {
+        if (edge.getEdge() != edgeId) {
+            edgeId = edge.getEdge();
+            time = weighting.calcMillis(edge, encoder.isBackward(edge.getFlags()), -1);
```
boldtrn: Since we don't have the reverse Flag here, I used this "hack" to determine the direction of the flag. Not sure if this is valid or if we should do it different?

karussell: The edges in the Path should be already properly directed so reverse should be always false for the path details?

core/src/test/java/com/graphhopper/routing/PathTest.java
```diff
@@ -309,20 +310,60 @@ public void testCalcDetails() {
...
+        assertEquals(10.0, averageSpeedDetails.get(2).getValue());
+        assertEquals(45.0, averageSpeedDetails.get(3).getValue());
+
+        assertEquals(0, averageSpeedDetails.get(0).getFirst());
+        assertEquals(1, averageSpeedDetails.get(1).getFirst());
+        assertEquals(2, averageSpeedDetails.get(2).getFirst());
+        assertEquals(3, averageSpeedDetails.get(3).getFirst());
+        assertEquals(4, averageSpeedDetails.get(3).getLast());
+
+        List<PathDetail> streetNameDetails = details.get(STREET_NAME);
```
karussell: We should have different test methods for the different path details (and maybe reuse the common parts)

web/src/main/java/com/graphhopper/http/GraphHopperServletModule.java
```diff
@@ -140,6 +140,10 @@ public void serialize(PathDetail value, JsonGenerator gen, SerializerProvider se
                 gen.writeNumber((Double) value.getValue());
             else if (value.getValue() instanceof Long)
                 gen.writeNumber((Long) value.getValue());
+            else if (value.getValue() instanceof Integer)
+                gen.writeNumber((Integer) value.getValue());
+            else if (value.getValue() instanceof Integer)
+                gen.writeNumber((Integer) value.getValue());
```
karussell: I think this is somehow duplicated?

boldtrn: Oops, you are right :)

Issue #786, opened 2016-09-01
Round trip heading


devemux86:
To set heading for round trips (with code), the only way seems to be via [request hints](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/template/RoundTripRoutingTemplate.java#L76).

It would be useful to use also the `GHRequest` API for headings, like we do with regular routing.

(forum discussion [here](https://discuss.graphhopper.com/t/round-trip-heading/986))

PR #1158, opened 2017-09-18
Allows setting initial heading for round trips via GHRequest.

ammagamma:
Implements suggestions in #786.
For now both using a request hint and using the GHRequest API is supported to set the initial heading for roundtrips. Setting it via GHRequest takes precedence.


ammagamma:
Shall we remove/deprecate/undocument the possibility to set the heading via a request hint ?

karussell:
Thanks! Yeah I think we might remove RoundTrip.HEADING and related or what do you think @devemux86 ?

devemux86:
> I think we might remove RoundTrip.HEADING and related or what do you think @devemux86 ?

Probably yes, unless support backwards compatibility or allow both API and hints for convenience.

ammagamma:
Ok, I removed RoundTrip.HEADING in the last commit. Within Java one now needs to use `favoredHeadings` instead of the `HintsMap` to set the initial heading of a round trip.
However, the server API is basically unchanged: Setting a heading is done by passing e.g. `heading=123` as request parameter. In `GraphHopperServlet#doGet` the heading is added to `favoredHeadings` of the GHRequest, which is now used by `RoundTripRoutingTemplate`. Before this PR this was the same except that `RoundTripRoutingTemplate` did not take the heading from `favoredHeadings` but parsed the request parameters again looking for the 'heading' request parameter. Consequently the documentation in api-doc.md does not need to be changed.

karussell:
This is awesome & cleaner - thanks!

core/src/main/java/com/graphhopper/routing/template/RoundTripRoutingTemplate.java
```diff
@@ -100,6 +100,14 @@ public RoundTripRoutingTemplate(GHRequest request, GHResponse ghRsp, LocationInd
         return queryResults;
     }
 
+    private double getInitialHeading() {
+        if (ghRequest.hasFavoredHeading(0)) {
+            return ghRequest.getFavoredHeading(0);
+        } else {
+            return Double.NaN;
+        }
+    }
```
karussell: Do we need this here? Shouldn't ghRequest.getFavoredHeading(0) return NaN already in those cases?

ammagamma: Without this check we get `IndexOutOfBoundsException` in case no heading was used because then the favoredHeadings list is empty. Of course it would be possible to change `GHRequest#getFavoredHeading` and make a similar check there, it is otherwise only used in `ViaRoutingTemplate`.

karussell: ah, I see there is a first condition against the size in the hasFavoredHeading method ... maybe we remove this first check and always fill the `favoredHeadings` List? But to be honest: I do not see how this list stays empty :)

ammagamma: Oh ok you are right in GHRequest it is made sure that `favoredHeadings always has same size as `points`, and `favoredHeadings` is initialized with NaN entries.

karussell:
Cool, thanks!

Issue #1162, opened 2017-09-25
Investigate Path Details memory usage and performance


karussell:
Currently there is a measurable slow down of nearly 10% when instructions are enabled. (In the Measurement class we also enable the PathDetails calculation if withInstructions=true).

So we should look if everything is good when looking into the memory and speed via a profiler. Also see if the slow down disappears if path details are disabled.

karussell:
I'm pretty sure that there is no slow down if we disable path details (we still have to check), but I would like to be sure that when we enable them the overhead is as low as possible.

boldtrn:
Ah OK, so this is only when we enable PathDetails? 10% for CH sounds actually not too unreasonable to me, when using a couple of PathDetails. The simplification becomes more complicated, which I guess should be a big part of the slow down. The more PathDetails are requested, the more complicated the simplification becomes. I think I would have expected 5% slow down, just as a gut feeling. For example the edgeId PathDetails, should add a lot more simplification runs.

karussell:
Hmmh, even if I disable path details the same slow down happens, this should not be. Can you confirm this?

boldtrn:
I compared the current master to a state before the PathDetails and indeed the routingCH.mean is about 10% higher in the current master.

I did some profiling and found that the method `PointList#copy` takes too much CPU time. The method is called frequently in `PathSimplification#reduceLength`. This call is removed in #1138, [here](https://github.com/graphhopper/graphhopper/pull/1138/files#diff-0948f2a2464f0901289e39074f54c493L158).

Also the method DouglasPeucker#subSimplify and DouglasPeucker#compressNew take up a lot of cpu time. When comparing it to the older state of GraphHopper, the simplify algorithm took a lot of cpu time before, so this is probably OK.

So I think we could improve the performance having a more efficient `PointList#copy` method. I will try this.

karussell:
We can close this via #1171

PR #1171, opened 2017-10-03
Improve PathSimplification Performance and Memory Overhead

boldtrn:
As reported in #1162, there are issues with the performance and memory usage.

One issue of memory usage was found in the PathList#copy method. 

I profiled the current master and found that even with 3051eca5683ed9bf1446419fd9e0c4052c33a5a5 a lot of memory is used due to PathList#copy and PathList#add. This PR tries to solve this to solve this issue by creating [shallow copies](https://en.wikipedia.org/wiki/Object_copying#Shallow_copy) of the PathList instead of deep copies. I cannot find the same memory consumption with this PR.

The issue with the shallow copies is that we might see incorrect results if we change the original PointList. For example we have a PointList with 3 points, A,B,C and we create a Shallow copy of this PointList, referencing point A, B. If we now change the underlying PointList, for example reverse it, then the shallow copy references to the points C,B instead of A,B.

We cannot simply lock the PointList, once we create a shallow copy, since we create shallow copies during the PathSimplification after each simplification (see [PathSimplification#reduceLengh](https://github.com/graphhopper/graphhopper/blob/20d6e67bf114543ea6feaabf5510510a14eecc8d/core/src/main/java/com/graphhopper/util/PathSimplification.java#L160), but change the original PointList with each simplification. It might be possible to lock the original PointList once we merge #1138. A partial lock might work, but I am not sure if this becomes too much of an overhead?

I benchmarked the performance of this PR with Apache Bench and the following command: `ab -n 1000 -c 20 "http://localhost:8989/route?point=49.214009%2C8.723145&point=49.203243%2C10.151367&point=47.755945%2C9.695435&point=48.991032%2C9.613037&point=48.026672%2C7.973328&type=json&locale=en-AU&vehicle=car&weighting=fastest&elevation=false&key="`

So it sampled 1000 requests and 20 concurrent requests. The used route is about 750km long.

Master:
```
Concurrency Level:      20
Time taken for tests:   14.157 seconds
Complete requests:      1000
Failed requests:        6
   (Connect: 0, Receive: 0, Length: 6, Exceptions: 0)
Total transferred:      55914988 bytes
HTML transferred:       55505994 bytes
Requests per second:    70.64 [#/sec] (mean)
Time per request:       283.139 [ms] (mean)
Time per request:       14.157 [ms] (mean, across all concurrent requests)
Transfer rate:          3857.08 [Kbytes/sec] received
```

This PR:
```
Concurrency Level:      20
Time taken for tests:   13.346 seconds
Complete requests:      1000
Failed requests:        12
   (Connect: 0, Receive: 0, Length: 12, Exceptions: 0)
Total transferred:      55914976 bytes
HTML transferred:       55505988 bytes
Requests per second:    74.93 [#/sec] (mean)
Time per request:       266.920 [ms] (mean)
Time per request:       13.346 [ms] (mean, across all concurrent requests)
Transfer rate:          4091.46 [Kbytes/sec] received
```

For some reason `./graphhopper.sh measurement` does not produce reproducible results for me. I have not found the reason for this but it might differ +-5% per measurement. So sometimes it seemed this PR would improve the performance, sometimes it seems it does not. With Apache Bench it seems to improve the performance.


karussell:
> The issue with the shallow copies is that we might see incorrect results if we change the original PointList. 

This is ugly. But unavoidable IMO. To make it a bit more safe it might be interesting to have an explicit makeReadOnly method after the simplification? So while simplification there is some risk for our internal development but afterwards everything is fine for the end user? To make this more obvious we could rename the method to shallowUnsafeCopy?

karussell:
I can confirm that this improves the introduced instructions overhead reported in #1162.

But I can use Measurement. Also I found that the slow down was introduced in commit 

11efb7ae6289a66b7a34644b2b40e12bb49de40f Add Average Speed as PathDetails to Response (#1091)

~3.6ms vs. 4.1ms

Now with this change it is back to 3.7ms but of course only if I delete the newly introduced lines in Measurement:

```java
if (withInstructions)
    req.setPathDetails(Arrays.asList(Parameters.DETAILS.AVERAGE_SPEED));
```

boldtrn:
> if (withInstructions)
    req.setPathDetails(Arrays.asList(Parameters.DETAILS.AVERAGE_SPEED));

Ah, this explains a lot :). I forgot that we add PathDetails in der Measurement and didn't change that line.

> To make it a bit more safe it might be interesting to have an explicit makeReadOnly method after the simplification?

Yes I think a write lock would be a good addition here. Will add this.

> To make this more obvious we could rename the method to shallowUnsafeCopy?

Yes, makes sense.

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -142,6 +142,8 @@ public void doWork(PathWrapper altRsp, List<Path> paths, Translation tr) {
         if (allFound && simplifyResponse && (calcPoints || enableInstructions)) {
             PathSimplification ps = new PathSimplification(altRsp, douglasPeucker, enableInstructions);
             ps.simplify();
+            // Make sure that the instruction references are not broken
+            altRsp.getPoints().makeImmutable();
```
boldtrn: I wasn't sure if it would be better to put this in the simplification, but then decided it makes more sense here, so it's not hidden and it's more obvious to others what happens here.

core/src/test/java/com/graphhopper/util/PointListTest.java
```diff
@@ -110,4 +111,46 @@ public void testCopy_issue1166() {
...
+        PointList pl3 = pl2.shallowCopySafe(0, 100);
+        // If we create a safe shallow copy of pl2, we have to make pl1 immutable
+        assertTrue(pl1.isImmutable());
+        assertEquals(100, pl3.size());
+        for (int i = 0; i < pl3.size(); i++) {
+            assertEquals(pl2.getLon(i), pl3.getLon(i), .01);
+        }
+
+        PointList pl4 = pl1.shallowCopyUnsafe(0, pl1.size());
+        assertTrue(pl1.equals(pl4));
```
boldtrn: Is this correct? Yes it contains the same points, but it is not really equal as one is only a shallow copy.

boldtrn:
Ok, at first I implemented the `equals` function, that it would also check if the mutability of both PointLists is equal. This made some tests fail, as assertEquals failed. For now I reverted the change but I am not sure, if equals should check for the mutability of the object as well?

core/src/test/java/com/graphhopper/util/PointListTest.java
karussell: I would say, yes.

core/src/main/java/com/graphhopper/util/PathMerger.java
karussell: Good question :)

I would tend to put it in PathSimplication as then all uses of it are already safe. 

BTW: what happens if we have an immutable PointList and pass it to PathSimplification - will this fail?

core/src/main/java/com/graphhopper/util/PointList.java
```diff
@@ -138,12 +138,13 @@ public boolean is3D() {
         }
     };
     private final static DistanceCalc3D distCalc3D = Helper.DIST_3D;
-    private static String ERR_MSG = "Tried to access PointList with too big index!";
+    protected static String ERR_MSG = "Tried to access PointList with too big index!";
```
karussell: just `final static` is better here. `protected` is too open

core/src/main/java/com/graphhopper/util/PointList.java
```diff
@@ -216,6 +220,9 @@ public void add(double lat, double lon) {
     }
 
     public void add(double lat, double lon, double ele) {
+        if (this.isImmutable()) {
```
karussell: To reduce redundancy I would move this into a ensureImmutability method

core/src/main/java/com/graphhopper/util/PathMerger.java
boldtrn: Ok, I can change this. 

> BTW: what happens if we have an immutable PointList and pass it to PathSimplification - will this fail?

Haven't tried it, but most probably, yes. Right now I haven't added a way to unlock a PointList. Once Locked it stays locked. You will have to create a deep copy to have the same PointList unlocked, which kind of makes sense to me, as we promised that the safeShallowCopy would be actually safe.

core/src/main/java/com/graphhopper/util/PointList.java
boldtrn: Ah yes, makes sense, thanks :)

core/src/main/java/com/graphhopper/util/PointList.java
```diff
@@ -449,14 +468,41 @@ public PointList copy(int from, int end) {
...
+
+    /**
+     * Create a shallow copy of this Pointlist from from to end, excluding end.
+     * <p>
+     * This method makes this object immutable, therefore the shallow copy is guaranteed to always contain the correct
+     * references.
+     * <p>
+     * If you know what your are doing, you can also use {@link PointList#shallowCopyUnsafe(int, int)}.
+     */
+    public PointList shallowCopySafe(final int from, final int end) {
```
karussell: Do we need the safe method? Or maybe use unsafe as a parameter?

core/src/main/java/com/graphhopper/util/ShallowImmutablePointList.java
```diff
@@ -0,0 +1,205 @@
...
+            sb.append(getLatitude(i));
+            sb.append(',');
+            sb.append(getLongitude(i));
+            if (is3D) {
+                sb.append(',');
+                sb.append(getElevation(i));
+            }
+            sb.append(')');
+        }
+        return sb.toString();
```
karussell: Is this duplication necessary?

core/src/main/java/com/graphhopper/util/PointList.java
boldtrn: Would be also possible. I liked the idea of two different methods here, but doesn't really matter and keeps the duplication low. I can merge the two methods.

core/src/main/java/com/graphhopper/util/ShallowImmutablePointList.java
boldtrn: I just spotted an issue here, thanks. The method should run from `from` to `end` and not from 0 to size. I can try to reduce the code duplicates a bit :).

core/src/main/java/com/graphhopper/util/PointList.java
karussell: I do not like the name with safe+unsafe (not exactly clear to which word the safe belongs) and shallowCopy (or createShallowCopy?) is nicer

core/src/main/java/com/graphhopper/util/PointList.java
boldtrn: Thanks, I name the method `ensureMutability` as we are making sure that the object is mutable, I think it is a bit clearer that way?

karussell:
Thanks. One final safety improvement we should discuss: should we make both PointList classes final? Currently one could tweak the PointList endlessly which could introduce methods with which the shallow copy has problems etc

boldtrn:
> One final safety improvement we should discuss: should we make both PointList classes final? 

Would this work? The ShallowCopy extends the PointList. I think this would be forbidden if we set the PointList final?

karussell:
Ah, ok. That is still a bit ugly. Hmmh.

boldtrn:
> Ah, ok. That is still a bit ugly. Hmmh.

I added a hint in the class comment of the PointList. I think if someone tries to extend this class he/she would at least read the class comment. Probably there is not much we can do, if someone incorrectly extends any class of GH, it might break?

karussell:
> Probably there is not much we can do, if someone incorrectly extends any class of GH, it might break?

If some classes are not explicitly intended for this we should make them final (I know this is not the case elsewhere in the code base but we should improve step by step :)).

There is the following trick that we could try: create a _PointList class that is package protected and that does what the current PointList does, then create the two **final** classes: `PointList extends _PointList` and `ShallowCopyPointList extends _PointList`. Would this work?

boldtrn:
> If some classes are not explicitly intended for this we should make them final (I know this is not the case elsewhere in the code base but we should improve step by step :)).

But wouldn't it be better to offer users the flexibility to extend and change GraphHopper very flexible? Or what is the reason to forbid this explicitly?

> There is the following trick that we could try: create a _PointList class that is package protected and that does what the current PointList does, then create the two final classes: PointList extends _PointList and ShallowCopyPointList extends _PointList. Would this work?

It could work like this, yes. But this makes usages of the PointList class outside of the package ugly. If we would keep the code as it is, we could not simply use the ShallowCopy instead of the PointList and we cannot set the datatype to _PointList as this class is package protected. So we would have to set the PointList of the Instructions to ShallowImmutablePointList and change all interfaces they use together. So maybe we can trick Java to accept this by casting the ShallowCopy to a PointList. Personally, I would prefer to avoid this.

karussell:
> But this makes usages of the PointList class outside of the package ...

Ehem, yes ;) ... so this won't work. The only solution left would be to break the API and introduce a PointList interface, but this would be too ugly. Let us keep it this way and make at least the shallow class final.

>  But wouldn't it be better to offer users the flexibility to extend and change GraphHopper very flexible? Or what is the reason to forbid this explicitly?

We can open up our API if we know there really is a good use case. Otherwise if we allow broad usage we later cannot easily migrate to other stuff or implementation OR we have to break this public API.

boldtrn:
> Otherwise if we allow broad usage we later cannot easily migrate to other stuff or implementation OR we have to break this public API.

Ah, now I understand :). Ok, makes sense.

> Let us keep it this way and make at least the shallow class final.

Just pushed the change :).

PR #1176, opened 2017-10-05
LM cut regions on Import

boldtrn:
There was a bug (or missing feature :)) in the Import class. If you create a graph with LM and import, the regions/continents were not separated. If you import a Graph via the GraphHopperModule this works. 

I moved as much code as possible into GraphHopperOSM and the json specific part into reader-json. I think having it like this makes the code a lot cleaner.


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1183,7 +1183,7 @@ protected void prepareCH() {
      * For landmarks it is required to always call this method: either it creates the landmark data or it loads it.
      */
     protected void loadOrPrepareLM() {
-        boolean tmpPrepare = lmFactoryDecorator.isEnabled();
+        boolean tmpPrepare = lmFactoryDecorator.isEnabled() && !lmFactoryDecorator.getPreparations().isEmpty();
```
boldtrn: Not sure if the isEmpty is necessary, but we had that check in the GraphHopperModule before, so I thought it would make sense to add it here as well.

reader-osm/src/main/java/com/graphhopper/reader/osm/GraphHopperOSM.java
```diff
@@ -44,4 +61,32 @@ public GraphHopperOSM setOSMFile(String osmFileStr) {
         super.setDataReaderFile(osmFileStr);
         return this;
     }
+
+    @Override
+    protected void loadOrPrepareLM() {
+        if (!getLMFactoryDecorator().isEnabled() || getLMFactoryDecorator().getPreparations().isEmpty())
+            return;
+
+        if (landmarkSplittingFeatureCollection != null && !landmarkSplittingFeatureCollection.getFeatures().isEmpty()) {
```
boldtrn: This should not break any existing code. As the landmarkSplittingFeatureCollection will be null by default.

karussell:
Aha - thanks :) !

A bit related issue: having no json in the core turns out to make too many things too ugly. In #1112 I have created a new `json` module (that is used from `core`) that comes with GHJson and a jackson implementation and could be replaced for Android or iOS.

The reason for that was that I replaced the StorableProperties with a JSON format, that could then be similar to a new config.json. Otherwise we had to increase self-invented format and parsing and stuff like we already do for car|turn_costs=true. Afterwards I was able to just do `encodingManager = json.parse(stream, EncodingManager.class)` without writing (de)serialization code. And if the jackson implementation is not available in `core` we would not be able to load the graph which is a core functionality IMO or we had to create another module that combines `core` and `json`.

All the geometry related code should *not* be in the `json` module to keep it really minimal, but could be moved from the `reader-json` module into `web` to avoid confusion about `json` vs. `reader-json`.

WDYT?

boldtrn:
> encodingManager = json.parse(stream, EncodingManager.class)

This is pretty cool :). 

> All the geometry related code should not be in the json module to keep it really minimal, but could be moved from the reader-json module into web to avoid confusion about json vs. reader-json.

What exactly would you specify as "geometry related code"?

Generally I am against moving anything that is not strictly web-related into the web module. It just makes it hard for anyone that might not want to use the web module with GraphHopper. For example the same would be true for this PR. We could not use "geometry related code" (I am still not 100% sure what this means :)) in the Import class.

Regarding this PR: Should we merge this or should I change anything here?

Issue #240, opened 2014-07-28
Improve query time for CH


karussell:
Some possibilities to make CH search faster:
- [x] separate list of edges for CH and normal edges #339
- [x] implement stall on demand #240
- [ ] do a `chgraph.optimize` after preparation to make shortcuts more compact. Shortcuts for one node should come after each other. And as we have called `disconnect` one edge is only used from one node.
- [x] try sorting the graph (see #233 and #12)
- [ ] tune parameters of preparation, see also #1225
- [ ]  Cache optimizing the PriorityQueue #489
- [ ] special case if just distance&time is necessary, then we do not need to call Path.extract but have the distance&time store in every shortcut
- [ ] combine with landmarks (currently A* makes CH slower => try same approach we do for edge based CH where we use something similar but not same as usual A* heuristic)
- [ ] try stall on demand for edge-based CH
- [ ] decouple base graph edge weights from CH #1067

karussell:
After discussing with @kodonnell ([here](https://github.com/graphhopper/map-matching/pull/71#issuecomment-251380373)) it is likely that we traverse unnecessary nodes in CH. Probably 'stall on demand' is the solution. See the overly long shortest path tree of the CH algorithm where the green dots are start&end:

![ch-snippet](https://cloud.githubusercontent.com/assets/129644/19469396/767d18a6-951b-11e6-87a4-8be7b68bbcca.png)

kodonnell:
Wow, that's surprising for such close query points! Though I guess it kind of makes sense - by removing nodes it means that your average remaining node has more edges (I think) and hence more to explore.

I might be able to justify working on these optimisations for my employer (as we've got to route a *lot* of records). With that in mind, what do you think is best to work on - largest gain in performance vs time to implement (for someone who's still not super familiar with CH etc.)? Stall on demand sounds like it gives quite significant performance gains, so maybe that?

karussell:
> Stall on demand sounds like it gives quite significant performance gains, so maybe that?

Not the simplest one but sounds promising. Depending on your requirements e.g. if you just need distance there are other optimization possible and nice to have where the boost can be of same significance.

karussell:
Also there are different algorithms which use CH as a preprocessing step (like transit node routing) which kind of convert the routing into a "table lookup process" and have a significant preprocessing overhead but the query time is or can be orders of magnitude faster (according to the papers)

kodonnell:
> Depending on your requirements

[...]

PR #1179, opened 2017-10-06
Implements a simple version of stall-on-demand for CH queries.

ammagamma:
This is a simple version of stall-on-demand that leads to a 1.5x speed up of CH queries on the osm map for Germany. Also see discussion in: https://github.com/graphhopper/graphhopper/issues/240.
I put the stall-on-demand version of the bidirectional ch dijkstra into a separate class, because PrepareContractionHierarchies already seems very large (?). I thought it would be better to leave it up to configuration whether stall-on-demand is used or not. I also added some measurement tests for stall on demand in the measurement class. Not sure if this is really needed, but for testing it is helpful at the moment. I might test some of the more aggressive versions of stall on demand if I find the time.


tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -413,8 +472,8 @@ else if (!rsp.getErrors().get(0).getMessage().toLowerCase().contains("not found"
         count -= failedCount.get();
 
         // if using none-bidirectional algorithm make sure you exclude CH routing
-        final String algoStr = ch ? Algorithms.DIJKSTRA_BI : Algorithms.ASTAR_BI;
-        put(prefix + ".guessed_algorithm", algoStr);
+        String guessed_algorithm = ch ? ("".equals(algoStr) ? DIJKSTRA_BI : algoStr) : Algorithms.ASTAR_BI;
```
ammagamma: Getting the guessed algorithm string seems rather messy (?) ;)

karussell:
Wow, cool - will try!

karussell:
Wow, really nice & thanks a lot! I can confirm this is a huge improvement from 3.9ms to 2.7ms in my case (measurement of Germany & car)

core/src/main/java/com/graphhopper/routing/DijkstraBidirectionRef.java
```diff
@@ -42,8 +42,8 @@
     protected SPTEntry currFrom;
     protected SPTEntry currTo;
     protected PathBidirRef bestPath;
-    private PriorityQueue<SPTEntry> pqOpenSetFrom;
-    private PriorityQueue<SPTEntry> pqOpenSetTo;
+    protected PriorityQueue<SPTEntry> pqOpenSetFrom;
+    protected PriorityQueue<SPTEntry> pqOpenSetTo;
```
karussell: Can we make it package protected instead? As it is not final this is a bit unsafe to allow subclasses access to it.

core/src/main/java/com/graphhopper/util/Parameters.java
```diff
@@ -33,6 +33,10 @@
          */
         public static final String DIJKSTRA_BI = "dijkstrabi";
         /**
+         * Bidirectional Dijkstra making use of stall-on-demand optimization
+         */
+        public static final String DIJKSTRA_BI_SOD = "dijkstrabisod";
```
karussell: A new internal parameter is ok for tests and comparison etc, but I wouldn't make this official in the Parameters class as I think we have to make SOD the default :)

tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -322,9 +327,62 @@ private void compareRouting(final GraphHopper hopper, String vehicle, int count)
...
+                    logger.error("Error for SOD but not for original" + locStr);
+                    continue;
+                }
+            }
+            String infoStr = " weight:" + sodRsp.getBest().getRouteWeight() + ", original: " + stdRsp.getBest().getRouteWeight()
+                    + " distance:" + sodRsp.getBest().getDistance() + ", original: " + stdRsp.getBest().getDistance()
+                    + " time:" + Helper.round2(sodRsp.getBest().getTime() / 1000) + ", original: " + Helper.round2(stdRsp.getBest().getTime() / 1000)
+                    + " points:" + sodRsp.getBest().getPoints().size() + ", original: " + stdRsp.getBest().getPoints().size();
+
+            if (Math.abs(1 - sodRsp.getBest().getRouteWeight() / stdRsp.getBest().getRouteWeight()) > 0.000001)
```
karussell: Ah, really nice that you did this test too!

core/src/main/java/com/graphhopper/routing/DijkstraBidirectionRef.java
```diff
@@ -157,8 +157,8 @@ public boolean finished() {
         return currFrom.weight + currTo.weight >= bestPath.getWeight();
     }
 
-    void fillEdges(SPTEntry currEdge, PriorityQueue<SPTEntry> prioQueue,
-                   IntObjectMap<SPTEntry> bestWeightMap, EdgeExplorer explorer, boolean reverse) {
+    protected void fillEdges(SPTEntry currEdge, PriorityQueue<SPTEntry> prioQueue,
```
karussell: This should not be necessary IMO

karussell:
BTW: for disabled instructions the differences are even bigger: 2.9ms vs. 1.5ms!

devemux86:
Seem impressive results!

boldtrn:
Wow this sounds very impressive! If I understand we only stall a node when we are sure that the path via this node is worse than a previously found path. [We do this by checking the current weight + the weight to one of the next edges is higher than the previously calculated best weight](https://github.com/graphhopper/graphhopper/pull/1179/files#diff-5400334343d76e757c16a0bcc75896e4R70).

Could it happen that we due to this incorrectly stall a node and might not find the best path anymore?
```
Assume this is a part of a graph. We are calculating the shortest path from some node X that is to the left to node D.
The weight to Node C is 5
The weight to Node A is 4
The weight for the edge between A and C is 2.
If we now check if A is stallable, we check 4+2 > 5 => Ok, A is stallable
The weight of the edge A-B and B-C might be 1, but the weight of the edge C-D might be 5, but we stalled A, and cannot find the best path anymore.
------  C
     /      \
--  A - B - D 
```   

Probably this situation is impossible because of the hierarchy and due to this it is guaranteed that we always find the best path?

core/src/main/java/com/graphhopper/routing/DijkstraBidirectionRef.java
ammagamma: If we want to use package private and still use this method (and `pqOpenSetFrom/To`) we need to move `DijkstraBidirectionCHWithSOD` to the same package as `DijkstraBidirectionRef`. In this case I would also move `DijkstraBidirectionCH` and `AStarBidirectionCH` from `PrepareContractionHierarchies` to `com.graphhopper.routing`. I would find this cleaner anyway ;)

karussell: Yes, please ;)

ammagamma:
Thanks for testing everyone!
@boldtrn I am not sure if I understand your example completely. Checking if A is stallable happens when settling A during the bidirectional CH Dijkstra. A can be stalled if there is an incoming edge coming from a node such that the tentative distance of the neighboring node plus the edge weight is smaller than the tentative distance of A (5+2 would have to be smaller than 4 in your example).
But you are right, the stalling optimization is specific to the query algorithm in contraction hierarchies (and some other hierarchical approaches). It is guaranteed that the shortest path is always found. For example if a node that lies on a shortest path gets stalled it will be discovered by the backward search. It gets more complicated when one tries to stall not only the current node, but also other ones that are reached from the stalled node (aggressive stalling). In this case one might have to un-stall nodes later, but this does not apply for the simple version that is used here.

boldtrn:
> But you are right, the stalling optimization is specific to the query algorithm in contraction hierarchies (and some other hierarchical approaches). It is guaranteed that the shortest path is always found.

Thanks for the answer. I thought that this would work due to the hierarchical properties of CH, but I couldn't proof it myself :). 

> For example if a node that lies on a shortest path gets stalled it will be discovered by the backward search.

Ah that's interesting, so the stalling only happens for one direction. That makes sense.

core/src/main/java/com/graphhopper/routing/DijkstraBidirectionCH.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    protected Path createAndInitPath() {
+        bestPath = new Path4CH(graph, graph.getBaseGraph(), weighting);
+        return bestPath;
+    }
+
+    @Override
+    public String getName() {
+        return "dijkstrabinosod|ch";
```
ammagamma: If the version with stall on demand becomes the default, maybe change the class names as well ?

karussell: Yes

karussell: Should we use a more readable name dijkstrabi_no_sod?

karussell: or even dijkstrabi|ch|no_sod?

tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -148,10 +148,8 @@ protected DataReader importData() throws IOException {
                 CHGraph lg = g.getGraph(CHGraph.class, weighting);
                 fillAllowedEdges(lg.getAllEdges(), allowedEdges);
                 printMiscUnitPerfTests(lg, isCH, encoder, count * 100, allowedEdges);
-                printTimeOfRouteQuery(hopper, isCH, isLM, count, "routingCH", vehicleStr, true, -1, "");
-                printTimeOfRouteQuery(hopper, isCH, isLM, count, "routingCH_sod", vehicleStr, true, -1, DIJKSTRA_BI_SOD);
-                printTimeOfRouteQuery(hopper, isCH, isLM, count, "routingCH_no_instr", vehicleStr, false, -1, "");
-                printTimeOfRouteQuery(hopper, isCH, isLM, count, "routingCH_sod_no_instr", vehicleStr, false, -1, DIJKSTRA_BI_SOD);
+                printTimeOfRouteQuery(hopper, isCH, isLM, count, "routingCH", vehicleStr, true, -1);
+                printTimeOfRouteQuery(hopper, isCH, isLM, count, "routingCH_no_instr", vehicleStr, false, -1);
```
karussell: also here I would test 1. with SOD 2. without SOD and 3. with SOD and no instructions instead of running two queries per measurement

core/src/main/java/com/graphhopper/routing/DijkstraBidirectionCH.java
ammagamma: sure

ammagamma:
some integration tests are failing now that sod is used by default, will have a look ;)

core/src/main/java/com/graphhopper/routing/DijkstraBidirectionCH.java
```diff
@@ -0,0 +1,93 @@
...
+import com.graphhopper.storage.Graph;
+import com.graphhopper.storage.SPTEntry;
+import com.graphhopper.util.EdgeExplorer;
+import com.graphhopper.util.EdgeIterator;
+
+/**
+ * Uses a very simple version of stall-on-demand (SOD) for CH queries to prevent exploring nodes that can not be part
+ * of a shortest path. When a node that is about to be settled is stallable it is not expanded, but no further search
+ * for neighboring stallable nodes is performed.
+ */
```
karussell: Feel free to add an `@author` tag here.

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
```diff
@@ -46,6 +46,7 @@
  * @author Peter Karich
  */
 public class PrepareContractionHierarchies extends AbstractAlgoPreparation implements RoutingAlgorithmFactory {
+    public static final String STALL_ON_DEMAND = "stall_on_demand";
```
karussell: Should this better go into the Parameters class?

karussell: Or should we probably hide this config and do not even make it public? (At least it needs the `ch.` prefix)

ammagamma: it was in Parameters.Algorithms before :) Maybe Parameters.CH would be a good place for it ?

ammagamma: I think it is useful to be able to deactivate it for example in the Measurement class

karussell: :D ... oh ... so yes, lets keep it here but reduce visibility? I currently do not see a case where it would make sense to disable stalling except for debugging and performance measurements. And then we could use the pure string without introducing some public usable constants.

ammagamma: I usually prefer constants so I can search for usages, but you probably want to keep the public API as small as possible to reduce compatibility problems in future versions I guess. In this case yes I would make the constant private or remove it and just use the string value where it is needed.

karussell: Yes, exactly. The only solution would be package protected but would not help much here.

ammagamma:
Looks like I am too used to IntelliJ's organize imports feature ...

boldtrn:
Great PR, thanks for the awesome work. Just to make sure, this does not change the graph generation in any way, meaning, if I imported a graph before that PR, this graph is still perfectly valid and functional? From looking at the code, I would assume that this should be the case, just wanted to make sure.

ammagamma:
Yes a previous graph should still work, the graph pre-processing is the same and only the query algorithm is changed.

PR #1180, opened 2017-10-06
Fix 'undefined' bug by checking location input fields

oschlueter:
This PR proposes a fix for #272 

What I changed:
* leave point param empty if location input is empty aswell
* trigger location input update when deleting "via point" (necessary because location input update was implicitly caused by a page reload due to routing along 'undefined' locations)

How I tested:
1. open localhost:8989 and click on the car button
this updates the URL but leaves the location inputs empty
2. open localhost:8989, add two via points and remove the via points 
this wouldn't work with the first change and without the second change: the URL would be updated but the UI components would remain until the next route update occurs
3. open localhost:8989, (optionally) add via points,  set a _start_ point anywhere and drag it around
this leaves all empty location inputs empty (instead of filling them with 'undefined')
4. open localhost:8989, (optionally) add via points,  set an _end_ point anywhere and drag it around
this leaves all empty location inputs empty (instead of filling them with 'undefined')


karussell:
Cool & thanks! This is a nice contribution regarding a very old issue :)

karussell:
This works good, thanks again!

Issue #581, opened 2015-11-27
Route: More info for "Connections between locations not found"


michaelkrog:
If I make a route between many points but one of them are not possible to make a route to, then I get a RuntimeException with the message "Connections between locations not found" from https://github.com/graphhopper/graphhopper/blob/9d07f4fb300726aa2386587fbd52408b0fbdccbb/core/src/main/java/com/graphhopper/util/PathMerger.java#L142.

I miss some info about the actual point(s) that causes the error, so Im able to fix the data if that is the problem.

karussell:
Indeed this should be improved!

karussell:
Related to #776

theaaron123:
Has this been resolved [here](https://github.com/graphhopper/graphhopper/blob/826342c3b42b1f6402367e9aceb1cd7816a4f6c4/core/src/main/java/com/graphhopper/util/PathMerger.java#L131)?

karussell:
No, still the points are missing. BTW: @michaelkrog are you interested in the point index or actual lat,lon?

oschlueter:
Which points should be included in that collection? 

Say we have a graph like this:
```
a <--> b

       c <--> d
```
Further assume we'd like to travel from point `a` to point `d`, hence the problem is not a single node unconnected to the rest but a fifty-fifty split situation. 

The `Path` instance passed to `PathMerger.doWork` doesn't seem to contain any information regarding any specific points which are causing the error. Additionally it seems to me `fromNode` and `endNode` are set to `-1` if no path is found which would render calculations after the fact impossible. 

Thus would you have all points of the underlying graph be added to the collection?

boldtrn:
> The Path instance passed to PathMerger.doWork doesn't seem to contain any information regarding any specific points which are causing the error. Additionally it seems to me fromNode and endNode are set to -1 if no path is found which would render calculations after the fact impossible.

I haven't debugged this, but I would assume that you know for which of the paths you cannot find a connection.

Assume we want to find a way from Berlin via Cologne to Munich. GraphHopper would calculate two paths Berlin->Cologne and Cologne->Munich. If we see that the Path Cologne->Munich is empty, then we should be able to create an exception stating from point 1 to point 2.

karussell:
This is fixed now.

PR #1183, opened 2017-10-08
added information about missing connections when merging paths

oschlueter:
This PR proposes a fix for #581 based on [this comment by Robin](https://github.com/graphhopper/graphhopper/issues/581#issuecomment-334612177).

### DIFF
--- a/core/src/main/java/com/graphhopper/util/PathMerger.java
+++ b/core/src/main/java/com/graphhopper/util/PathMerger.java
@@ -23,9 +23,7 @@ import com.graphhopper.util.details.PathDetail;
 import com.graphhopper.util.details.PathDetailsBuilderFactory;
 import com.graphhopper.util.exceptions.ConnectionNotFoundException;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 
 /**
  * This class merges multiple {@link Path} objects into one continues object that
@@ -130,7 +128,8 @@ public class PathMerger {
             altRsp.setInstructions(fullInstructions);
 
         if (!allFound) {
-            altRsp.addError(new ConnectionNotFoundException("Connection between locations not found", Collections.<String, Object>emptyMap()));
+            altRsp.addError(new ConnectionNotFoundException("Connection between locations not found: " +
+                    unconnectedListString(paths), unconnectedInfo(altRsp, paths)));
         }
 
         altRsp.setDescription(description).
@@ -182,5 +181,54 @@ public class PathMerger {
         }
         rsp.setAscend(ascendMeters);
         rsp.setDescend(descendMeters);
+    }
+
+    private String unconnectedListString(List<Path> paths) {
+        StringBuilder sb = new StringBuilder().append("[");
+
+        for (int n = 0; n < paths.size(); ++n) {
+            if (!paths.get(n).isFound())
+                sb.append(String.format("(%d,%d), ", n, n+1));
+        }
+
+        if (sb.length() < 2)
+            throw new IllegalArgumentException("list doesn't have unavailable paths");
+
+        sb.setLength(sb.length() - 2); // delete last occurrence of  ", "
+        sb.append("]");
+
+        return sb.toString();
+    }
+
+    private Map<String, Object> unconnectedInfo(PathWrapper altRsp, List<Path> paths) {
+        PointList waypoints = altRsp.getWaypoints();
+
+        ArrayList<Object> idxFrom = new ArrayList<>();
+        ArrayList<Object> latFrom = new ArrayList<>();
+        ArrayList<Object> lonFrom = new ArrayList<>();
+        ArrayList<Object> idxTo = new ArrayList<>();
+        ArrayList<Object> latTo = new ArrayList<>();
+        ArrayList<Object> lonTo = new ArrayList<>();
+
+        for (int pathIndex = 0; pathIndex < paths.size(); ++pathIndex) {
+            if (!paths.get(pathIndex).isFound()) {
+                idxFrom.add(pathIndex);
+                latFrom.add(waypoints.getLat(pathIndex));
+                lonFrom.add(waypoints.getLon(pathIndex));
+                idxTo.add(pathIndex + 1);
+                latTo.add(waypoints.getLat(pathIndex + 1));
+                lonTo.add(waypoints.getLon(pathIndex + 1));
+            }
+        }
+
+        Map<String, Object> info = new HashMap<>();
+        info.put("idxFrom", idxFrom);
+        info.put("latFrom", latFrom);
+        info.put("lonFrom", lonFrom);
+        info.put("idxTo", idxTo);
+        info.put("latTo", latTo);
+        info.put("lonTo", lonTo);
+
+        return info;
     }
 }
--- a/core/src/test/java/com/graphhopper/util/PathMergerTest.java
+++ b/core/src/test/java/com/graphhopper/util/PathMergerTest.java
@@ -1,0 +1,149 @@
+package com.graphhopper.util;
+
+import com.graphhopper.GHRequest;
+import com.graphhopper.GHResponse;
+import com.graphhopper.routing.AlgorithmOptions;
+import com.graphhopper.routing.Path;
+import com.graphhopper.routing.QueryGraph;
+import com.graphhopper.routing.RoutingAlgorithmFactorySimple;
+import com.graphhopper.routing.template.ViaRoutingTemplate;
+import com.graphhopper.routing.util.CarFlagEncoder;
+import com.graphhopper.routing.util.EncodingManager;
+import com.graphhopper.routing.util.FlagEncoder;
+import com.graphhopper.routing.util.TraversalMode;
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -183,4 +182,53 @@ private void calcAscendDescend(final PathWrapper rsp, final PointList pointList)
...
+    private String unconnectedListString(List<Path> paths) {
+        StringBuilder sb = new StringBuilder().append("[");
+
+        for (int n = 0; n < paths.size(); ++n) {
+            if (!paths.get(n).isFound())
+                sb.append(String.format("(%d,%d), ", n, n+1));
+        }
+
+        if (sb.length() < 2)
+            throw new IllegalArgumentException("list doesn't have unavailable paths");
```
boldtrn: Maybe a better exception would be: "All provided paths are found" or "The paths don't contain an unfound path"?

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -130,7 +128,8 @@ public void doWork(PathWrapper altRsp, List<Path> paths, Translation tr) {
             altRsp.setInstructions(fullInstructions);
 
         if (!allFound) {
-            altRsp.addError(new ConnectionNotFoundException("Connection between locations not found", Collections.<String, Object>emptyMap()));
+            altRsp.addError(new ConnectionNotFoundException("Connection between locations not found: " +
```
boldtrn: Since we now know which locations/points create this issue, wouldn't it be better to state "Connection between point1 and point2 (, and point 5 and point 6) not found"? IMHO the (1,2) look a bit like coordinates (at least I didn't understand the error message on first sight).

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -183,4 +182,53 @@ private void calcAscendDescend(final PathWrapper rsp, final PointList pointList)
...
+        sb.setLength(sb.length() - 2); // delete last occurrence of  ", "
+        sb.append("]");
+
+        return sb.toString();
+    }
+
+    private Map<String, Object> unconnectedInfo(PathWrapper altRsp, List<Path> paths) {
+        PointList waypoints = altRsp.getWaypoints();
+
+        ArrayList<Object> idxFrom = new ArrayList<>();
```
boldtrn: Do we need 6 ArrayLists here? Wouldn't it be easier to have everything as map so that the json result looks nice as well.

```
details = {
  "connection-not-found-1": {
    "idFrom": 1,
    ...
  },
  "connection-not-found-2": {
    "idFrom": 2
    .....
  }
}
```

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -183,4 +182,53 @@ private void calcAscendDescend(final PathWrapper rsp, final PointList pointList)
         rsp.setAscend(ascendMeters);
         rsp.setDescend(descendMeters);
     }
-}
\ No newline at end of file
+
+    private String unconnectedListString(List<Path> paths) {
```
boldtrn: Could we move these two methods into the ConnectionNotFoundException class to hide that logic? This would help to keep the PathMerger class a bit slimmer.

So we might have a new Constructor for this class accepting a List<Path> and a PointList?

boldtrn:
Thanks for creating this PR. Since you are a first time contributor, and I don't know if you already did this, but [could you send us a signed CLA via email](https://github.com/graphhopper/graphhopper/blob/master/.github/CONTRIBUTING.md#license-agreement)?

karussell:
> but could you send us a signed CLA via email?

Yes, already done. We should make this somehow better transparent. I'm currently thinking about dropping the requirement of such a CLA and instead sign every commit and have a "first contribution commit" which adds the nick name or something into the contributor file. Like the linux kernel community is doing this.

boldtrn:
> We should make this somehow better transparent. I'm currently thinking about dropping the requirement of such a CLA and instead sign every commit and have a "first contribution commit" which adds the nick name or something into the contributor file. Like the linux kernel community is doing this.

I like the way Google is handling their CLAs. They have a bot, not sure if they use [this opensource clabot](https://github.com/clabot/clabot). But maybe we should discuss this somewhere else :).

oschlueter:
Thank you for the feedback! 

I've adjusted the code accordingly and refactored recurring string values into constants.

core/src/main/java/com/graphhopper/util/exceptions/ConnectionNotFoundException.java
```diff
@@ -25,7 +30,75 @@
...
+    private static final String FMT = "point %d and point %d";
+    private static final String JOIN = ", and ";
+    private static final String OUTRO = " not found";
+
     public ConnectionNotFoundException(String var1, Map<String, Object> details) {
         super(var1, details);
     }
+
+    public ConnectionNotFoundException(List<Path> paths, PointList waypoints) {
+        super(unconnectedListString(paths), unconnectedInfo(paths, waypoints));
```
boldtrn: Wouldn't it be better to call `this(unconnectedListString(paths), unconnectedInfo(paths, waypoints));` in this case?

core/src/main/java/com/graphhopper/util/exceptions/ConnectionNotFoundException.java
```diff
@@ -25,7 +30,75 @@
...
+                Map<String, Number> infoEntry = new HashMap<>();
+
+                infoEntry.put(Keys.idFrom.name(), pathIndex);
+                infoEntry.put(Keys.latFrom.name(), waypoints.getLat(pathIndex));
+                infoEntry.put(Keys.lonFrom.name(), waypoints.getLon(pathIndex));
+                infoEntry.put(Keys.idTo.name(), pathIndex + 1);
+                infoEntry.put(Keys.latTo.name(), waypoints.getLat(pathIndex + 1));
+                infoEntry.put(Keys.lonTo.name(), waypoints.getLon(pathIndex + 1));
+
+                info.put("connection-not-found-" + info.size(), infoEntry);
```
boldtrn: I know this was my own idea :), but maybe it would be better to create json like:
```
connection-not-found: [
{...},
{...}
]
```

WDYT @karussell?

core/src/main/java/com/graphhopper/util/exceptions/ConnectionNotFoundException.java
oschlueter: Yes absolutely, that was an oversight.

core/src/main/java/com/graphhopper/util/exceptions/ConnectionNotFoundException.java
oschlueter: No sweat, API design needs to be thought through :-)

When I'm back from vacation and you guys have decided on your format of choice I can adjust it.

karussell: Sorry for the delay will have a look in the next time!

core/src/main/java/com/graphhopper/util/exceptions/ConnectionNotFoundException.java
```diff
@@ -25,7 +30,75 @@
...
+     *   },
+     *   "connection-not-found-1": {
+     *     "idFrom": 4
+     *     ....
+     *   }
+     * }
+     *
+     * @author Oliver Schlüter
+     */
+    public enum Keys {
```
karussell: This should not be part of the public Java API

core/src/main/java/com/graphhopper/util/exceptions/ConnectionNotFoundException.java
```diff
@@ -25,7 +30,75 @@
  * @author Peter Karich
  */
 public class ConnectionNotFoundException extends DetailedIllegalArgumentException {
+    private static final String INTRO = "Connection between ";
+    private static final String FMT = "point %d and point %d";
+    private static final String JOIN = ", and ";
+    private static final String OUTRO = " not found";
```
karussell: Why define them here explicitely?

oschlueter:
Wouldn't it be more useful to see all problems? I'm thinking about a compiler only showing one error each time I start a compilation...

core/src/main/java/com/graphhopper/util/exceptions/ConnectionNotFoundException.java
oschlueter: I'm using `INTRO` and `JOIN` multiple times (lines 54 and 57 respectively) hence I figured to store the strings in class variables. `FMT` and `OUTRO` exist to handle the string creation in `unconnectedListString`  consistently (i.e. no magic string components). 

How would you prefer this to be handled?

karussell:
We can print all but it should not be required for the implementation to do this, e.g. if the first problem is reported we can stop wasting CPU resources to calculate further.

karussell:
What about this PR? As @boldtrn already stated we have to be strict here with regards to the changed JSON output. 

What do you think about this: 

> I would like to propose that we use the same JSON format that we currently use for the Matrix API

PR #1182, opened 2017-10-08
(WIP) include junction=circular with additional edge flag

oschlueter:
This PR aims towards including the junction=circular tag (see #1149).

Image of circular junction

### approach
The approach is based upon Robin's idea of internally handling circular junctions as roundabouts. In order to accomplish this the PR 
* adds checks for `junction=circular` in all `FlagEncoder`s
* adds the edge flag `K_CIRCULAR_JUNCTION` to render instructions different from roundabout instructions
* adds i18n entries 'circular_junction_exit' and 'circular_junction_exit_onto'

### open tasks
(Potentially incomplete) list of remaining tasks required for this PR to be merged
- [ ] in the grand scheme of things consider implications of adding an edge flag (possible showstopper)
- [ ] in [i18n spreadsheet](https://t.co/f086oJXAEI) add `circular_junction_exit` and `circular_junction_exit_onto` 
- [ ] in client-hc create test checking if `GraphHopperWeb#createPathWrapper` sets `circularJunction` boolean from JSON
- [ ] in core review `InstructionsFromEdges#next` if it's sufficient to rely on encoders to set `FlagEncoder.K_ROUNDABOUT` if `FlagEncoder.K_CIRCULAR_JUNCTION` is set
- [ ] in core review `AbstractGraphStorageTester#test8BytesFlags` if it's ok to reduce the `speedBits` to `28` for the two `TmpCarFlagEncoder`s (due to the upper limit of available bits)
- [ ] in core review `GraphHopperOSMTest#testMultipleCHPreparationsInParallel` if it's ok to remove a vehicle
- [ ] in core review `GraphHopperOSMTest#testMultipleLMPreparationsInParallel` if it's ok to remove a vehicle
- [ ] in core create test checking if `RoundaboutInstruction.circularJunction` is `true` if instruction is built for ways tagged `junction=circular`
- [ ] in core analyze in `MotorcycleFlagEncoderTest#testCircularJunction` why `FlagEncoder.K_CIRCULAR_JUNCTION` is overwritten. Maybe some bit positions are assigned multiple times? (_**broken test**_)
- [ ] in core review `EncodingManagerTest#testWrongEncoders` why the amount of requested bits is lower than the amount checked for (_**broken test**_)

### alternative
A less intrusive change might be to simply add `junction=circular` as condition for `isRoundabout` in all `FlagEncoder`s (using regular roundabout textual instructions):
```
boolean isRoundabout = way.hasTag("junction", "roundabout");
[...]

Issue #1149, opened 2017-09-01
Use junction=circular


boldtrn:
Similar to junction=roundabout there is a tag called `junction=circular`. It is not very commonly used, but from time to time. I think for us the right way to handle this would be to set it as roundabout, as the only difference is the right of way.

[Overpass Query](http://overpass-turbo.eu/s/roI).
[GH Maps Example](https://graphhopper.com/maps/?point=51.310258%2C3.121136&point=51.309742%2C3.120579)
[Google Maps Route](https://goo.gl/maps/qBJPbBM8CfQ2)
[OSM Wiki](https://wiki.openstreetmap.org/wiki/Tag:junction%3Dcircular)

Since they are slightly different from roundabouts, some people might be mislead by the roundabout instruction?

WDYT?

karussell:
Tags under 1K are not worth the effort IMO: https://taginfo.openstreetmap.org/search?q=junction%3Dcircular

Why are they no roundabouts?

boldtrn:
> Why are they no roundabouts?

Because of a different right of way. For roundabouts the cars inside the roundabout have the right of way, whereas the cars outside of the roundabout have to yield. There are some examples where this is not true.

> Tags under 1K are not worth the effort IMO

Yes maybe, especially since most of them seem to be turning areas :). On the other hand, for us internally it would be easy to match all tags with either junction=roundabout or junction=circular to roundabouts.

devemux86:
> On the other hand, for us internally it would be easy to match all tags with either junction=roundabout or junction=circular to roundabouts.

If that's possible, then any such enhancement would improve routing responses and user experience.

oschlueter:
Based on the suggestions made here I've submitted a _work-in-progress_ PR as basis for further discussions. The details are in the PR description. 

Would love to hear your feedback!

karussell:
Fixed via #1185

PR #1185, opened 2017-10-09
consider junction=circular for roundabout flag

oschlueter:
This PR implements Robin's proposal to include the junction=circular tag (see #1149) and is a reduced version of my previous PR #1182.

circular junction as roundabout


karussell:
Thanks. There is no need to increase the flag encoder version as the edge flags layout stays the same, ie the roundabout bit is at the same index etc

PR #1186, opened 2017-10-11
consider hints in batch mode

oblonski:
@karussell there is this batchIT testBike(..) which does not work as expected anymore. Would you mind to have a brief look at it?


oblonski:
I ran it with a local matrix server and it worked.

PR #1190, opened 2017-10-17
Enable PMD and Checkstyle for Code Climate

efueger:


karussell:
We are using now a more specific config via #1195

Issue #1184, opened 2017-10-09
Tests are slow since Findbugs


boldtrn:
I really like the addition of findbugs, but the tests take 4:30 min to run on my computer, without findbugs it's < 3 min. Would it make sense to only run findbugs on travis and remove it from the default test cycle? Maybe we can add it to a different phase?

karussell:
Yes, sounds good!

karussell:
Can you try out https://github.com/spotbugs/spotbugs-maven-plugin/ if it is faster?

boldtrn:
I have just seen that there are also statistical code analysis tools available in the Github marketplace. All of them are free for open source projects. Most also add code coverage and security checks. I have no experience with any of these tools, but maybe it would be also interesting to add one of them?

- https://github.com/marketplace/codacy
- https://github.com/marketplace/codebeat
- https://github.com/marketplace/better-code-hub
- https://github.com/marketplace/code-climate

They run checks before a PR can be merged, similar to Travis.

> Can you try out https://github.com/spotbugs/spotbugs-maven-plugin/ if it is faster?

Sure, I can try this, but don't you think it would be enough to only run them before merging the PR?

karussell:
> but don't you think it would be enough to only run them before merging the PR?

Yes, this would be enough. But also we should use this new findbugs and hope it is faster ;)

boldtrn:
> But also we should use this new findbugs and hope it is faster ;)

I just tried the spotbugs plugin and I wasn't able to use the maxRank config, so it always failed the build because of some minor issues. I used the same config as the findbugs plugin, which is also documented like this.The spotbugs maven plugin is not yet released as a stable version but only as a RC and is a fork of the findbugs plugin, so I am not sure how big the performance boost will be? I think it will be released as a first stable version in shortly, see [this issue](https://github.com/spotbugs/spotbugs/issues/8).

karussell:
Fixed via #1190 and #1191?

boldtrn:
Yes should be fixed, thanks :).

PR #1191, opened 2017-10-18
Don't run findbugs and checkstyle as default during the ITs

boldtrn:
As mentioned in #1184 findbugs and checkstyle slow down the test execution. I therefore removed the from the verify maven phase. They are still executed on travis, so if we introduce a massive issue in a PR travis should fail and report about it.

BTW: Findbugs seems to have issues with the Java 9 build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ before_install:
 install: true
 
 script:
-  - mvn clean test verify -B
+  - mvn clean test verify checkstyle:checkstyle findbugs:findbugs -B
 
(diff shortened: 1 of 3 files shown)


boldtrn:
This PR should be good to merge now. The speed of `mvn clean test verify` is no back to normal. On demand you can run `mvn clean test verify checkstyle:check findbugs:check`  locally to execute findbugs and checkstyle. These will be executed on travis as well.

karussell:
Thanks!

PR #1194, opened 2017-10-19
Fixed Exponential Notation for GPX export

boldtrn:
[Kurviger](https://kurviger.de/en) received a [bug report](https://groups.google.com/d/msg/kurviger/b4TdQQSzkT8/nTGrpmTwAQAJ) that there can be an issue with the GPX export if coordinates are close to 0. The default formatting of Java for values like `0.000123` is `1.23E-4` which makes sense for readability, but some programs like Garmin Basecamp have issues importing these numbers.

This PR fixes this issue. I tried to stay as close to the current behaviour of GH as possible, but removing the exponential notation. For coordinates like 15.0 it prints 15.0 but not just 15 or 15.000000. Also values like 0.1 are printed as 0.1.

### DIFF
--- a/core/src/main/java/com/graphhopper/util/Helper.java
+++ b/core/src/main/java/com/graphhopper/util/Helper.java
@@ -63,7 +63,6 @@ public class Helper {
     private static final float DEGREE_FACTOR = Integer.MAX_VALUE / 400f;
     // milli meter is a bit extreme but we have integers
     private static final float ELE_FACTOR = 1000f;
-    private static final DecimalFormat DECIMAL_FORMAT = new DecimalFormat("#");
 
     private Helper() {
     }
@@ -483,13 +482,6 @@ public class Helper {
 
     public static final double round6(double value) {
         return Math.round(value * 1e6) / 1e6;
-    }
-
-    public static final String print6(double value) {
-        DECIMAL_FORMAT.setMinimumFractionDigits(1);
-        DECIMAL_FORMAT.setMaximumFractionDigits(6);
-        DECIMAL_FORMAT.setMinimumIntegerDigits(1);
-        return DECIMAL_FORMAT.format(value);
     }
 
     public static final double round4(double value) {
--- a/core/src/main/java/com/graphhopper/util/Helper.java
+++ b/core/src/main/java/com/graphhopper/util/Helper.java
@@ -37,6 +37,7 @@ import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.text.DateFormat;
+import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.text.SimpleDateFormat;
 import java.util.*;
@@ -62,6 +63,7 @@ public class Helper {
     private static final float DEGREE_FACTOR = Integer.MAX_VALUE / 400f;
     // milli meter is a bit extreme but we have integers
     private static final float ELE_FACTOR = 1000f;
+    private static final DecimalFormat DECIMAL_FORMAT = new DecimalFormat("#");
 
     private Helper() {
     }
@@ -481,6 +483,13 @@ public class Helper {
 
     public static final double round6(double value) {
         return Math.round(value * 1e6) / 1e6;
+    }
+
+    public static final String print6(double value) {
+        DECIMAL_FORMAT.setMinimumFractionDigits(1);
+        DECIMAL_FORMAT.setMaximumFractionDigits(6);
+        DECIMAL_FORMAT.setMinimumIntegerDigits(1);
+        return DECIMAL_FORMAT.format(value);
     }
 
     public static final double round4(double value) {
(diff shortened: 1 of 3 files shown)


karussell:
Hmmh, I would like to avoid this complexity just because other programs have bugs. E.g. according to [this SO](https://stackoverflow.com/questions/4387170/decimalformat-formatdouble-in-different-threads) the thread safety is not guaranteed.

karussell:
What about String.format("%.0f", priceValue) or BigDecimal.toPlainString()?

boldtrn:
> Hmmh, I would like to avoid this complexity just because other programs have bugs.

I did some more research about this. The GPX Schema [defines lats and longs](http://www.topografix.com/gpx/1/1/#type_latitudeType) to be xsd:decimal. [xsd:decimal does not allow scientific notation](http://books.xmlschemata.org/relaxng/ch19-77057.html). A lot of GPS programs accept the scientific notification, probably because most String to Number converters accept the scientific notation.

> What about String.format

The issue with String.format is that it is rather inflexible when it comes to different numbers of digits. We could build a formatting string per input or accept that all values have six digits after the decimal point, like: `1.000000`. I would not mind, but this would change the values that we currently create in the GPX and also the resulting GPX track might be a bit bigger. Alternatively, we could count the number of required digits, and create a formatting string per double, but this seems to be a bit of an overhead to me.

[Here is the Java doc](https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html#dndec):
> The number of digits in the result for the fractional part of m or a is equal to the precision. If the precision is not specified then the default value is 6. If the precision is less than the number of digits which would appear after the decimal point in the string returned by Float.toString(float) or Double.toString(double) respectively, then the value will be rounded using the round half up algorithm. Otherwise, zeros may be appended to reach the precision. For a canonical representation of the value, use Float.toString(float) or Double.toString(double) as appropriate.

```
BigDecimal bigDecimal = new BigDecimal(0.1);
System.out.println(bigDecimal.toPlainString());

Prints: 0.1000000000000000055511151231257827021181583404541015625
``` 

``` 
BigDecimal bigDecimal = new BigDecimal(0.1);
bigDecimal = bigDecimal.round(new MathContext(6));
System.out.println(bigDecimal.toPlainString());

Prints: 0.100000
```

> E.g. according to this SO the thread safety is not guaranteed.

How about, we create the number formatting object inside the print method? I wasn't aware of thread issues here. So we could change the code to:

```
    public static final String print6(double value) {
        DecimalFormat decimalFormat = new DecimalFormat("#");
        decimalFormat.setMinimumFractionDigits(1);
        decimalFormat.setMaximumFractionDigits(6);
        decimalFormat.setMinimumIntegerDigits(1);
        return decimalFormat.format(value);
    }
```

karussell:
Ok, if the spec says this and it is indeed a but. Thanks for pointing this out.

> we create the number formatting object inside the print method? 

It probably would be better to create one formatter per request and reuse it `print6(DecimalFormat, double)`. I'm not sure about the perf implications if such a formatter is created per method call.

boldtrn:
> It probably would be better to create one formatter per request and reuse it print6(DecimalFormat, double). I'm not sure about the perf implications if such a formatter is created per method call.

The constructor looks a bit complicated, so there might be a performance implication, when calling the constructor often. Since I cannot see a benefit of the function in the Helper class when only calling the formatter and only using it on file, I decided the easiest would be to only do this in the InstructionList.

boldtrn:
How should we handle the code climate issues? Should we change the allowed line length? Should I break the lines?

karussell:
Hmmh, this check style stuff is overwhelming and to be honest: not so useful - or what do you think? (we could deactivate the not so useful warnings for now)

karussell:
BTW: looks good now :)

boldtrn:
> Hmmh, this check style stuff is overwhelming and to be honest: not so useful - or what do you think? (we could deactivate the not so useful warnings for now)

Personally I don't think enforcing a line length is a very good style check. Often breaking a line makes it harder to read compared to the longer version. Maybe we can change the max line length to 180-200 chars, which would be definitely too long and should be broken down.

On the other hand, I liked the Javadoc style check. I think we should enforce it. If you touch the code, you could as well write the Javadoc comment without having too much extra work. This would lead to improved documentation over time.

karussell:
> Maybe we can change the max line length to 180-200 chars, which would be definitely too long and should be broken down.

Good idea

> If you touch the code, you could as well write the Javadoc comment without having too much extra work

I do not think that every method needs a javadocs. Our public API (public classes and methods) surely need one (can this be specified?), but we should reduce this number to a minimum and make code mostly self explanatory via naming and tests.

boldtrn:
> I do not think that every method needs a javadocs. Our public API (public classes and methods) surely need one (can this be specified?), but we should reduce this number to a minimum and make code mostly self explanatory via naming and tests.

I just did some research about this. It appears that this is not easily possible. We could limit the necessity for JavaDoc only for public methods or classes that match a certain regex, but I think this might not be what we are looking for. [Here](http://checkstyle.sourceforge.net/config_javadoc.html) is the docu for this.

PR #1198, opened 2017-10-22
some random improvements

ammagamma:
just some minor changes that I thought were useful when browsing the code base

### DIFF
--- a/core/src/test/java/com/graphhopper/storage/AbstractGraphStorageTester.java
+++ b/core/src/test/java/com/graphhopper/storage/AbstractGraphStorageTester.java
@@ -36,7 +36,7 @@ import static org.junit.Assert.*;
 
 /**
  * Abstract test class to be extended for implementations of the Graph interface. Graphs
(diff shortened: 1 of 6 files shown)


core/src/test/java/com/graphhopper/storage/AbstractGraphStorageTester.java
```diff
@@ -36,7 +36,7 @@
 
 /**
  * Abstract test class to be extended for implementations of the Graph interface. Graphs
- * implementing GraphStorage should extend GraphStorageTest instead.
+ * implementing GraphStorage should extend {@link GraphHopperStorageTest} instead.
```
ammagamma: here is no GraphStorageTest, but even with this change this javadoc seems outdated ?

karussell: I think this is correct with your change. There were some implementations like for neo4j that used this abstract test, but all tests like GraphHopperStorageCHTest should not use it directly

PR #1201, opened 2017-10-25
Maximum Nodes Exceeded Exception

### LINKED ISSUE #1188: Create a GHException for "maximum nodes exceeded"
To make it easy for the frontend to parse the "maximum nodes exceeded" exception, it would be nice to have this exception as as GHException also returning its class.

boldtrn:
This fixes #1188. I added some more exceptions to GH to make it easier for the client to show appropriate help to the user.

Something interesting I found along the way was that in the GraphHopperServlet, we did not catch all RuntimeExceptions but only IllegalArgumentExceptions. Is this desired? I changed this to catch RuntimeExceptions. I would understand the argument, that users should not be notified about exceptions other than IllegalArgument, because it's not their fault. The API returned a 500 html document, maybe we catch all RuntimeExceptions and return a generic json error then?

I also changed a couple of other exceptions a bit. I think the overall goal should be to convert the most common exceptions into GHExceptions so the client can improve the parsing of these exceptions.


karussell:
>  Is this desired? I changed this to catch RuntimeExceptions

Other exceptions are errors that needs to be reported and should use http code=500. But if an html is served this could be fixed :)

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -549,7 +549,10 @@ boolean initActiveLandmarks(int fromNode, int toNode, int[] activeLandmarkIndice
         if (subnetworkFrom <= UNCLEAR_SUBNETWORK || subnetworkTo <= UNCLEAR_SUBNETWORK)
             return false;
         if (subnetworkFrom != subnetworkTo) {
-            throw new ConnectionNotFoundException("Connection between locations not found. Different subnetworks " + subnetworkFrom + " vs. " + subnetworkTo, new HashMap<String, Object>());
+            Map<String, Object> details = new HashMap<String, Object>();
+            details.put("subnetwork-from", subnetworkFrom);
+            details.put("subnetwork-to", subnetworkTo);
```
karussell: For json we should always use underscore

core/src/main/java/com/graphhopper/routing/template/ViaRoutingTemplate.java
```diff
@@ -119,8 +120,14 @@ public ViaRoutingTemplate(GHRequest ghRequest, GHResponse ghRsp, LocationIndex l
 
             int idx = 0;
             for (Path path : tmpPathList) {
-                if (path.getTime() < 0)
-                    throw new RuntimeException("Time was negative " + path.getTime() + " for index " + idx + ". Please report as bug and include:" + ghRequest);
+                if (path.getTime() < 0) {
+                    Map<String, Object> details = new HashMap<>();
+                    details.put("time", path.getTime());
+                    details.put("from", queryResults.get(idx).getSnappedPoint());
+                    details.put("to", queryResults.get(idx + 1).getSnappedPoint());
```
karussell: All these details could be included in the message, there is no need for an automated way to read this as this is a bug and should not occur. Also will it still result in a 500 http code? This is very important.

core/src/main/java/com/graphhopper/routing/template/ViaRoutingTemplate.java
```diff
@@ -132,8 +139,12 @@ public ViaRoutingTemplate(GHRequest ghRequest, GHResponse ghRsp, LocationIndex l
             // reset all direction enforcements in queryGraph to avoid influencing next path
             queryGraph.clearUnfavoredStatus();
 
-            if (algo.getVisitedNodes() >= algoOpts.getMaxVisitedNodes())
-                throw new IllegalArgumentException("No path found due to maximum nodes exceeded " + algoOpts.getMaxVisitedNodes());
+            if (algo.getVisitedNodes() >= algoOpts.getMaxVisitedNodes()) {
+                Map<String, Object> details = new HashMap<>();
+                details.put("max-visited-nodes", algoOpts.getMaxVisitedNodes());
+                throw new MaximumNodesExceededException("No path found due to maximum nodes exceeded", details);
```
karussell: Why would we need this changed? What is the use case that a client can better parse the value?

web/src/main/java/com/graphhopper/http/GraphHopperServlet.java
```diff
@@ -144,7 +145,7 @@ public void doGet(HttpServletRequest httpReq, HttpServletResponse httpRes) throw
                         put(WAY_POINT_MAX_DISTANCE, minPathPrecision);
 
                 ghRsp = graphHopper.route(request);
-            } catch (IllegalArgumentException ex) {
+            } catch (RuntimeException ex) {
```
karussell: As explained: without further work this change is not possible as for all other exceptions we have to throw http error == 500

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
karussell: Why would we need the subnetwork IDs parsable - what would be the use case?

boldtrn: > Why would we need the subnetwork IDs parsable - what would be the use case?

The idea was that it would be possible to differentiate between the "different subnetwork" exception and the general "connection not found" exception. Both have different reasons and I think it is important for the client to be able to understand the difference and tell the user about it. But you are right, I think it might actually be better to have a different Exception class here. I know, we are adding a massive number of classes here, but I think there is a huge benefit for the client here, making it easier for users to help themselves, by receiving a useful, translated error message. Will add a new class here :).

core/src/main/java/com/graphhopper/routing/template/ViaRoutingTemplate.java
boldtrn: I think I overdid it with the map and details a bit :). But the different class, yes to make it easier for the client to recognize this exception and react accordingly.

boldtrn: I actually think the number of maximum nodes is not really interesting for the client, so I removed it completely, WDYT?

web/src/main/java/com/graphhopper/http/GraphHopperServlet.java
boldtrn: Thanks, I added additional handling here. Would this be enough or should it work differently?

karussell: Would you ensure this via an integration test (illegal arg => 400, trigger exception => 500). Triggering an exception is probably only possible via injecting a 'broken' GraphHopper class...

core/src/main/java/com/graphhopper/routing/template/ViaRoutingTemplate.java
karussell: Yes, good idea. This is confusing to the client. Still maybe explain more what went wrong like "Path cannot be found as too many nodes ("junctions") already processed. This could happen if the locations are too far away" or something?

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
karussell: > I think it is important for the client to be able to understand the difference and tell the user about it

Ok, maybe the reason property is a good compromise if users want more details and we do not need to introduce new exception classes.


> In this case, the different subnetwork warning is created during our import

Where would we throw this in the import?

boldtrn: > Where would we throw this in the import?

Sorry I think I described this a bit weird :). What I meant is that we create the separation during the import, in contrast to blocked areas.

core/src/main/java/com/graphhopper/routing/template/ViaRoutingTemplate.java
boldtrn: > Still maybe explain more what went wrong like 

Good idea :+1:

web/src/main/java/com/graphhopper/http/GraphHopperServlet.java
boldtrn: Yes makes sense. I looked a bit more into this and this might be a bit complicated. Injecting a GraphHopper class is a surprising overhead. I had to do a couple of rather dirty things in the test class, but I could not find a better way to do this. I did not want to change anything in the production code. The issue is that we cannot extend the method `GraphHopperModule#createGraphHopper`, because of the `@Provides` annotation. We could create a wrapper method to make this possible, but I am not sure if we should do this just for a test?

karussell: You should be able to use

install(Modules.override(new GraphHopperModule(cmdArgs)).with(new TestModule(cmdArgs)));

and then put another createGraphHopper in the TestModule

boldtrn: > install(Modules.override(new GraphHopperModule(cmdArgs)).with(new TestModule(cmdArgs)));

I had a look at this, but this would end up in the same amount of code duplication, wouldn't it? Or maybe I didn't understand this correctly?

karussell: No, you just need to implement createGraphHopper not all the other methods as they are provided from the GraphHopperModule

boldtrn: Yes, but I cannot override the createGraphHopper method because of the @Provide annotation or do I miss something?

karussell: You cannot but because of the Modules.override the provider (if existent) of the test modules are called and not the ones in GraphHopperModule. So use a provide annotation and the GraphHopper return type.

core/src/main/java/com/graphhopper/util/exceptions/ConnectionNotFoundException.java
```diff
@@ -28,4 +29,29 @@
...
+    public ConnectionNotFoundException(String var1, Map<String, Object> details, Reason reason) {
+        super(var1, details);
+
+        if (details == null) {
+            details = new HashMap<>();
+        }
+        details.put("reason", reason.toString());
+    }
+
+    public enum Reason {
```
karussell: Why a separate class?

web/src/test/java/com/graphhopper/http/BaseServletTester.java
```diff
@@ -109,6 +109,10 @@ private void bootJetty(CmdArgs args, int retryCount) {
         }
     }
 
+    GHServer getServer(CmdArgs args) {
```
karussell: I think this is not necessary as you can call BaseServletTester.setUpGuice(modules)

core/src/main/java/com/graphhopper/util/exceptions/ConnectionNotFoundException.java
boldtrn: Yes maybe I over engineered here a bit. I think my idea was that we could have different reasons and this would make it easier always return the same reason from different locations. Should I just make it a String?

karussell: Yes, for now a string seems good. Is the Map as parameter in the constructor the same way like we do it for other exceptions? Or should we make the map internal & final and allow changing it via `exception.set(key, value)`?

web/src/test/java/com/graphhopper/http/BaseServletTester.java
karussell: Here I meant that you can customize the modules etc via extending the modules, there is no need to customize GHServer, nor to create this method. I think you can search for TestModule and how to utilize this (this would then also better work with dropwizard IMO)

core/src/main/java/com/graphhopper/util/exceptions/ConnectionNotFoundException.java
boldtrn: Ok let's change it to a String :).

web/src/test/java/com/graphhopper/http/BaseServletTester.java
boldtrn: Ah thanks. I think I now understood :)

web/src/test/java/com/graphhopper/http/BaseServletTester.java
```diff
@@ -69,9 +69,6 @@ public void setUpGuice(Module... modules) {
      * This method will start jetty with andorra area loaded as OSM.
      */
     public void setUpJetty(CmdArgs args) {
-        if (injector != null)
-            throw new UnsupportedOperationException("Do not call guice before");
-
```
boldtrn: Had to remove this, otherwise it wouldn't have been possible to call setupGuice.

karussell: It should be possible. I'll have a look.

web/src/test/java/com/graphhopper/http/BaseServletTester.java
karussell: I think, there is no TestModule. Probably have created it in another project instead of calling `setUpGuice(server.createModule());`

web/src/test/java/com/graphhopper/http/BaseServletTester.java
boldtrn: Ah, nice, that makes sense, thanks :).

karussell:
Did some minor simplifications [here](https://github.com/graphhopper/graphhopper/compare/issue_1201). Is this ok for you?

boldtrn:
> Did some minor simplifications here. Is this ok for you?

Thanks. My only issue with the changes is explained [here](https://github.com/graphhopper/graphhopper/commit/9c568c255783b60825c54df1b44895bf7ca8ded0#r27729592).

Issue #1157, opened 2017-09-18
Gradle build error when building in android studio


kkalera:
Hi, I wanted to try out the android demo app but I'm getting the following error when building:
```

Information:Gradle tasks [:app:assembleDebug]
Error:Uncaught translation error: com.android.dx.cf.code.SimException: signature-polymorphic method called without --min-sdk-version >= 26
Error:1 error; aborting
Error:Execution failed for task ':app:transformClassesWithDexForDebug'.
> com.android.build.api.transform.TransformException: java.lang.RuntimeException: java.lang.RuntimeException: Unable to pre-dex 'C:\Users\Kalera\.gradle\caches\modules-2\files-2.1\com.graphhopper\graphhopper-core\0.9.0\963667d1bcedee2656b9891aa6069a95e2e7b4cf\graphhopper-core-0.9.0.jar' to 'C:\Users\Kalera\Desktop\graphhopper-master\android\app\build\intermediates\transforms\dex\debug\folders\1000\10\graphhopper-core-0.9.0_1faaa0ecde7189cc3a7428820cb9955817078d04'
Information:BUILD FAILED
Information:Total time: 6.872 secs
Information:3 errors
Information:0 warnings
Information:See complete output in console
```

I have tried switching the version from 0.10-SNAPSHOT to 0.9 but without any luck.
Cleaning and rebuilding the project also do not solve the issue.

karussell:
See this issue comment: https://github.com/graphhopper/graphhopper/issues/1144#issuecomment-324585805

If you know the underlying reason and how to fix this - would love to hear about it.

ChristophKaser:
I also encountered this error and think I have found the cause of it.
Apparently, the newest version of the android build tools do not like the call to MethodHandle.invoke in com.graphhopper.util.Helper#cleanMappedByteBuffer

If I comment out the code depending on Constants.JRE_IS_MINIMUM_JAVA9, I can use the built jar in an android project. If I leave it in, it results in the error above.

I think changing the code to use "old-school" reflection would fix the issue.

karussell:
Interesting! Thanks for the update!

> I think changing the code to use "old-school" reflection would fix the issue.

What do you mean here?

ChristophKaser:
I meant instead of this:
` MethodHandle unmapper = MethodHandles.lookup().findVirtual(unsafeClass, "invokeCleaner",
                                MethodType.methodType(void.class, ByteBuffer.class));`
Using something like this:
[...]

PR #1205, opened 2017-11-01
Android: Gradle 4, Android plugin 3, SDK 27

devemux86:
Android has released new major versions in all tools, so we should advance the sample too.

~But first the PR demonstrates immediately the current failed use of GraphHopper on Android, as described in #1157, if not commenting the mentioned code.~
EDIT: that should be fixed via #1207.


android/build.gradle
```diff
@@ -1,15 +1,17 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     repositories {
+        google()
```
karussell: Is this necessary :) ?

devemux86: `google()` is a now the formal repository for Android related libraries, like plugins, support libs, etc. ([here](https://dl.google.com/dl/android/maven2/index.html) are the contents).

devemux86:
> BTW: Does it work with the mmap fix now?

Yes I tested it, all seem working ok again!

karussell:
Cool, thanks!

karussell:
One last comment: would you mind signing the commit? (if not possible yet, no problem & no hurry, we can do this later then)

devemux86:
Signed and force pushed, should be ok now?

karussell:
Cool, thanks!

Issue #1208, opened 2017-11-02
Contributing via signed commits


karussell:
As already mentioned in #1207 I would like to change the signing process for the CLAs. The current process requires printing&signing&scanning&emailing us or submitting to a third party like signaturit.com. 

When we instead switch to a "signed commit"-based procedure we can improve the visibility & transparency who has signed and therefore 'commit rights'. Also we increase the number of signed & verified commits in general improving "trust" into open source code further IMO.

Necessary steps for the contributor will be:

 1. change [CONTRIBUTORS.md](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTORS.md) and add a line in alphabetical order with his/her github nickname stating the CLA agreement, optionally adding contribution details and company or website.
 2. [sign the commit](https://help.github.com/articles/signing-commits-using-gpg/)
 3. make sure the commit contains a valid Email

This still give us the guarantee that a certain user agrees to the CLA and also contact information via the email. Would this be an improvement?

devemux86:
As [repositories](https://github.com/mapsforge) administrator, I find the loss of PR commits squash (mentioned in #1207) as disadvantage.
That would lead to excessive repository history and difficulty to read a PR if spreading to many commits.

BTW GitHub nicknames better to be links to the actual web pages?

karussell:
> That would lead to excessive repository history and difficulty to read a PR if spreading to many commits.

This does not happen if the contributors rebase their commits as then multiple commits could end up in the master better reflecting the history compared to a single squashed commit. This has the disadvantage of the mentioned steeper learning curve but also has two advantages: clearer history of PR itself and identical commits like in the master branch

> GitHub nicknames better to be links to the actual web pages?

Optionally yes, also supporting company or similar

karussell:
Would be interested in an opinion from @michaz @boldtrn @oblonski @ratrun and @stefanholder (just picked a few contributors regarding the different scenarios like external company, single dev etc :) ).

ratrun:
The requirement to provide a valid e-mail would be a no-go for me.

karussell:
The problem without an email is the lack of identity and another personal argument <send via email> ;)

karussell:
[...]

PR #1212, opened 2017-11-04
Docker conventions and thoughts

legraina:
I've modified the Dockerfile and the docker-compose to use Docker conventions.
You now have an entrypoint in the Dockerfile and only a command in the docker-compose.yml file.

What's more, I think you should create your own docker image on docker hub (https://docs.docker.com/docker-hub/repos/#searching-for-images) and also perhaps create automatic builds (https://docs.docker.com/docker-hub/github/#github-service-hooks). I could do it, but I think you should create your own ids and organization on docker hub. I could also provide help if necessary.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -38,6 +38,7 @@ Here is an overview:
  * JohannesPelzer, improved GPX information and various other things
  * karussell, one of the core developers
  * khuebner, initial turn costs support
+ * legraina, improved docker for dockerhub
  * lmar, improved instructions
  * michaz, one of the core developers
  * mprins, improvements for travis CI and regarding JDK9 #806
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,21 @@
 FROM openjdk:8-jdk
 
-RUN mkdir -p /data
-RUN mkdir -p /graphhopper
+ENV JETTY_PORT 11111
+ENV JAVA_OPTS "-server -Xconcurrentio -Xmx1g -Xms1g -XX:+UseG1GC -XX:MetaspaceSize=100M"
+
+RUN mkdir -p /data && \
+    mkdir -p /graphhopper
 
 COPY . /graphhopper/
 
 WORKDIR /graphhopper
 
 RUN ./graphhopper.sh buildweb
+
+VOLUME [ "/data" ]
+
+EXPOSE 11111
+
+ENTRYPOINT [ "./graphhopper.sh", "web" ]
+
+CMD [ "/data/europe_germany_berlin.pbf" ]
--- a/core/files/docker-compose.yml
+++ b/core/files/docker-compose.yml
@@ -10,7 +10,4 @@ services:
             JAVA_OPTS: "-server -Xconcurrentio -Xmx1g -Xms1g -XX:+UseG1GC -XX:MetaspaceSize=100M"
         ports:
             - "11111:11111"
-        entrypoint:
-            - ./graphhopper.sh
-            - web
-            - ../data/europe_germany_berlin.pbf
+        command: /data/europe_germany_berlin.pbf


karussell:
Thanks! What is the improvement here? (sorry I'm kind of a docker newb ;))

> but I think you should create your own ids and organization on docker hub. I could also provide help if necessary

Can we deploy a new docker image everytime travis runs successfully or for every tagged release or something?

legraina:
The improvement is that most of the people are using an entrypoint and then a command: when the docker starts, it run the entrypoint + command. In the entrypoint (here ./graphhopper web), you put the basics that normally a user won't owerride, the command is used by the final user to put its options (here the map that the server will run).
Most of the common docker image are using an entrypoint that run a script docker-entrypoint.sh (nginx, redis, mongod, ...), and it's now the common way of building a docker image. The docker-entrypoint.sh run all commands that need to be executed each time the docker image is started.

We could also declare some of the environment variable in the Dockerfile and override them if needed (as the JETTY_PORT or the JAVA_OPTS), but I do not know what should be the basic behavior.

You will deploy a new image, every time travis run correctly. The image that corresponds to the master branch will automatically be tagged with latest (no specific tag is needed), and for the stable, you will just add the number of the version. You could also add a tag docker on every stable branch that needs to deploy. In travis, you should also be able to recover the name of the branch.

With these improvements, people will be able to run graphhopper without downloading the git repo. You will just need in the dockerfile FROM graphhopper/graphhopper or in the docker-compose to use image: graphhopper/graphhopper

I hope that it's clearer.

Dexyon:
I love the idea of having you guys maintaining your own DockerHub account and push out some images. Github/ Docker Hub is if full of people/projects who have done it for their own in the past and I think it will be good to publish your own. Even just for centralising all discussions and best practices in a single place which is trusted by people. 

As @legraina already mentioned, there are enough people who are willing to lend a helping hand. I would not mind helping out a bit on the docker (or Kubernetes) side. We have been using GH for about two years in a Docker / Kubernetes set-up!

Some comments on the Dockerfile. I personally always tend to use a fixed version of Java SDK's base images. We are now using `openjdk:8-jdk`. This is ok if you want to keep all your images based upon the latest sources. But maybe consider using a fixed version. I agree that this is more a personal preference, since I love to keep control over which exact base images are used.

legraina:
I have updated the Dockerfile to have all the necessary information. The docker can be run with:
````bash
docker build -t graphhopper . 
docker run -v /absolute/path/to/data/:/data -p 80:11111 graphhopper
````

karussell:
Would you state "agrees to the project's CLA" also in the contributors file?

Also I need just two (or one) commits, currently there are 9. Another possibility would be to create a new pull request with just the CLA and 1 signed commit that I would merge. Then I could squash&merge on my own and no more complication. (see #1208)

legraina:
I think it's correctly done now. Do you need any help for creating the Docker image on docker hub ?

karussell:
Thanks!

> Do you need any help for creating the docker image on DockerHub ?

We would need to find an automated way to avoid any manual triggers. Is this somehow possible after e.g. a github release/tag?

legraina:
It can been done really easily on DockerHub. You don't need any tag, you just need to grant  access on DockerHub to all your git repositories. I can create everything and then transfer ownership. Is graphhopper a great name for the organization and the docker image ?

karussell:
I think I've done something that publishes the latest stuff :) ?
https://hub.docker.com/r/graphhopper/graphhopper/builds/

legraina:
That's exactly what you need. Nothing else. What you can also do is following the other branches that correspond to previous releases (it can be done in DockerHub), you can also add the tag stable to the latest stable version. Finally, you can add a markdown to the README.md for the docker, for example: [Docker Build Status](https://hub.docker.com/r/graphhopper/graphhopper/). And you are done 👍  Many thanks :)

legraina:
I've seen that you have added the stable flag, however I've realized that the corresponding Dockerfile (for the branch 0.9) is not working the same way. I think you should add the stable tag when the branch 1.0 will be pushed.

karussell:
There will no 1.0 in the near future or do you mean 0.10? Maybe I can modify 0.9 and use the latest Dockerfile?

Dexyon:
The dockerfile looks fine. I usually try to take a more specific version of the Java JDK instead of relying on the latest tag here. But that is just my opinion. Latest and greatest is also a good choice :)

karussell:
Would you mind to have a look into #1250 @legraina?

Issue #384, opened 2015-03-25
Use high precision elevation data


karussell:
Currently we use SRTM or CGIAR both having 90m resolution but a new data set released which has 30m resolution. Still the CGIAR data is not updated and I'm not sure if using 'pure' SRTM is wise. 

BTW: Instead of interpolating via surrounding tiles - maybe we should interpolate while creating an edge with the ele data?

Also it looks it is not yet fully released ... in April 2015 it will release for EU+US and other parts will follow this year.

Data sources:
- http://www2.jpl.nasa.gov/srtm/ -> http://dds.cr.usgs.gov/srtm/version2_1/SRTM30/
- New srtm data [now void filled](https://lpdaac.usgs.gov/nasa_shuttle_radar_topography_mission_srtm_global_1_arc_second_data_released_over_middle_east). See the [SRTM user guide](https://lpdaac.usgs.gov/sites/default/files/public/measures/docs/NASA_SRTM_V3.pdf)
- new and open data set from EU [here](http://www.eea.europa.eu/data-and-maps/data/eu-dem). There seems to be more recent data [here](https://spacedata.copernicus.eu/web/cscda/news/-/asset_publisher/NQThjzBp3dRm/content/copernicus-dem-a-new-digital-elevation-model-for-the-copernicus-programme?_101_INSTANCE_NQThjzBp3dRm_redirect=%2Fweb%2Fcscda%2Fnews).
- new and open data set from [UK](https://environmentagency.blog.gov.uk/2015/09/18/laser-surveys-light-up-open-data/)
- new and open data set from Sweden, license is CC-By (ftp://download-open.lantmateriet.se/GSD-Hojddata%20grid%2050%20plus/nh/Stockholm/Sweref%2099%20TM/geotiff/) but there is an older data set which becomes public in ~2020, see section 2.2. of (ftp://download-open.lantmateriet.se/GSD-Hojddata%20grid%2050%20plus/hdb/Stockholm/Sweref%2099%20TM/ascii-tabell/)
 - http://ned.usgs.gov/faq.html#DATA
 - [OpenDEM](https://opendem.info/)

Interesting tool to merge these different data sets [here](https://gitlab.com/lucadelu/OpenDEM_Generator)

NopMap:
It is not a good idea to use SRTM. The holes and errors in mountainous regions are terrible. Best example is this lake in a valley which has an elevation difference of 400m according to SRTM.
[...]

PR #1217, opened 2017-11-10
Use multiple elevation data sources

boldtrn:
I did some more research about different elevation data sources with the goal to provide elevation data for countries that lie outside of the area covered by SRTM.

During my research I found some global or almost global datasets:
- [Aster Gdem](https://asterweb.jpl.nasa.gov/gdem.asp)
- [ALOS AW3D30](http://www.eorc.jaxa.jp/ALOS/en/aw3d30/)
- [GTOPO30](https://lta.cr.usgs.gov/GTOPO30)
- [GMTED2010](https://lta.cr.usgs.gov/GMTED2010)
- [Viewfinderpanoramas](http://viewfinderpanoramas.org/Coverage%20map%20viewfinderpanoramas_org3.htm)

Aster Gdem and AW3D30 both have the issue that they have big holes of missing data, e.g. see the coverage map of [AW3D30](http://www.eorc.jaxa.jp/ALOS/en/aw3d30/l_map.htm) (you can see that the holes in the SRTM covered area have been filled with SRTM data).

GTOPO30 has a resolution of 30 arc seconds, which translates to a grid of roughly 1*1 km, which is simply too coarse, [as some valleys do not show up in this data](https://pubs.usgs.gov/of/2011/1073/pdf/of2011-1073.pdf).

GMTED2010 has a resolution of 7.5 arc seconds, which is roughly a grid of 250*250 m and is available between latitudes of 84°N and 70°S.

The 3 arc second data of Viewfinderpanoramas is often described as one of the best datasets available. I am still not 100% sure about the license of Viewfinderpanoramas, will need to investigate more.

Therefore, the interesting choices for us would be GMTED2010 or Viewfinderpanoramas. Due to the licensing uncertainty, I started with an implementation using the GMTED2010 elevation data.

As expected the elevation calculation with GMTED for areas covered by SRTM are worse than the CGIAR data, but the data is good enough to produce a rough elevation diagram. If we would further smooth the elevation values, the results could be pretty good.

I added a `MultiSourceElevationProvider` that returns results from different ElevationProviders depending on the requested coordinates. Currently it returns CGIAR data inside of the SRTM covered area and GMTED data outside of the SRTM covered area.

Top GMTED, Bottom CGIAR from [GH maps](https://graphhopper.com/maps/?point=48.549796%2C9.38344&point=48.535164%2C9.407773)
gmtedneuffen
ghneuffen

Top GMTED, Bottom CGIAR from [GH maps](https://graphhopper.com/maps/?point=48.643345%2C9.649944&point=48.585408%2C9.765215)
gmtedauendorf
ghauendorf

Top GMTED, Bottom CGIAR from [GH maps](https://graphhopper.com/maps/?point=47.942337%2C7.867756&point=47.826871%2C7.942429&locale=en-AU&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale)
gmtedfreiburg
ghfreiburg

Top GMTED, Bottom CGIAR from [GH maps](https://graphhopper.com/maps/?point=48.021391%2C8.885193&point=47.942337%2C8.759365&point=47.927041%2C8.577404&point=47.955444%2C8.4902). This route follows the donau valley, with the more coarse elevation data, it happened more often that a small part of a road ended up "on top of a hill", you can see the spikes, but they appear for the CGIAR data as well as the GMTED data. The spikes look bad, but are only ~40-50m (see the heights on the left, the route itself is quite flat).
gmteddonau
ghdonau

A routing accross the SRTM borders. Top GMTED, Mid [GH maps](https://graphhopper.com/maps/?point=59.391975%2C13.540649&point=60.683939%2C13.705444), Bottom the MultiSourceElevationProvider.
gmtedsweden
ghsweden
multisweden

Also a short route with the [Multi Source Provider](https://graphhopper.com/maps/?point=59.965666%2C13.970146&point=59.999759%2C13.953323&point=60.022668%2C13.905945), showing the switch of resolution. The slider has been moved to the position where the data switches. The left side has a lot more details (CGIAR), the right side is more coarse (GMTED):
multiswitch

Overall I really like the multiple source approach. At this point I cannot see any disadvantage compared to our current approach, but it offers the benefit of adding elevation data outside of the SRTM covered area. Using the same approach we could also add high precision elevation data, that is often only available for smaller areas, also see #384.


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -576,10 +576,15 @@ public GraphHopper init(CmdArgs args) {
...
             tmpProvider = cgiarProvider;
+        } else if (eleProviderStr.equalsIgnoreCase("gmted")) {
+            tmpProvider = new GMTEDProvider();
+        } else if (eleProviderStr.equalsIgnoreCase("multi")) {
+            tmpProvider = new MultiSourceElevationProvider();
         }
 
         tmpProvider.setCalcMean(eleCalcMean);
-        tmpProvider.setCacheDir(new File(cacheDirStr));
+        if (!cacheDirStr.isEmpty())
```
boldtrn: Currently the default cacheDir was "" (the project root) and not "/tmp/[provider-name]". I changed that, but I am not sure if this might create issues for users of non unix systems?

core/src/main/java/com/graphhopper/reader/dem/CGIARProvider.java
```diff
@@ -80,84 +66,31 @@ public static void main(String[] args) {
 
         // 337.0
         System.out.println(provider.getEle(49.949784, 11.57517));
-        // 453.0
+        // 466.0
```
boldtrn: These values seem to have changed, maybe the CGIAR data changed over time, maybe there was a CGIAR update? This happens for me with the current master as well, so it wasn't introduced by a change in this PR.

core/src/main/java/com/graphhopper/reader/dem/GMTEDProvider.java
```diff
@@ -0,0 +1,261 @@
...
+                "/tmp/gmted",
+                "GraphHopper GMTEDReader");
+    }
+
+    public static void main(String[] args) {
+        GMTEDProvider provider = new GMTEDProvider();
+
+        System.out.println(provider.getEle(46, -20));
+
+        // 337.0 (339)
```
boldtrn: The first number is the SRTM/CGIAR result for this coordinate, the result of GMTED is in braces.

core/src/main/java/com/graphhopper/reader/dem/HeightTile.java
```diff
@@ -36,21 +36,27 @@
...
+    private final int verticalDegree;
     private final double lowerBound;
-    private final double higherBound;
+    private final double lonHigherBound;
+    private final double latHigherBound;
     private DataAccess heights;
     private boolean calcMean;
 
-    public HeightTile(int minLat, int minLon, int width, double precision, int degree) {
+    public HeightTile(int minLat, int minLon, int width, int height, double precision, int horizontalDegree, int verticalDegree) {
```
boldtrn: The GMTED tifs are not quadratic, the same is true for other elevation sources as well, so it is quite important that we can handle non quadratic files.

karussell:
Wow, cool & thanks!

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: Yeah, we should keep it to project root

karussell: It looks like we use /tmp/cgiar already so `/tmp/<provider-name>` makes sense.

core/src/main/java/com/graphhopper/reader/dem/CGIARProvider.java
```diff
@@ -80,85 +66,33 @@ public static void main(String[] args) {
...
-        return this;
-    }
-
-    @Override
     public double getEle(double lat, double lon) {
-        // no data we can avoid the trouble
-        if (lat > 60 || lat < -60)
+        // Return fast, if there is no data available
+        // See https://www2.jpl.nasa.gov/srtm/faq.html
+        if (lat > 60 || lat < -56)
```
karussell: Nice finding!

core/src/main/java/com/graphhopper/reader/dem/MultiSourceElevationProvider.java
```diff
@@ -0,0 +1,115 @@
...
+        return gmtedProvider.getEle(lat, lon);
+    }
+
+    public CGIARProvider getCgiarProvider() {
+        return cgiarProvider;
+    }
+
+    public GMTEDProvider getGmtedProvider() {
+        return gmtedProvider;
+    }
```
karussell: Do we need those public methods?

core/src/test/java/com/graphhopper/reader/dem/MultiSourceElevationProviderTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.net.SocketTimeoutException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * @author Peter Karich
```
karussell: Not 100% correct ;)

core/src/main/java/com/graphhopper/reader/dem/MultiSourceElevationProvider.java
boldtrn: Not necessarily anymore. I first thought it would be handy to change settings for one provider but not the other, but I am not sure if anyone would want to use this?

core/src/test/java/com/graphhopper/reader/dem/MultiSourceElevationProviderTest.java
boldtrn: Woops, sorry :)

core/src/main/java/com/graphhopper/reader/dem/MultiSourceElevationProvider.java
karussell: Let us remove them for now until there is a use case :)

boldtrn:
Ok, I tried to squash this PR after merging in the master, but apparently, the merging information gets lost due to squashing. Now all the changes of the master pollute this PR as well. I can try to create a new PR extracting all the my changes from the change log by hand.

karussell:
Hmmh, not sure what you mean. What are you doing to squash the changes?

My steps would be squash into a single commit, push force into the branch so that it appears here, then we merge it into master.

boldtrn:
> What are you doing to squash the changes?

```
git rebase -i HEAD~n (n=number of commits)
# Then I squash by changing pick to s and update the commit message to make sense.
git push -f origin [branch-name]
```

> My steps would be squash into a single commit, push force into the branch so that it appears here, then we merge it into master.

Yes, this is not possible if you need to merge the master, especially not if there are merge conflicts, because the information about the merge will be lost. Have a look at the changes in this PR, it shows every change made in the master since I created this PR.

I think my mistake was to squash the merge commit as well. But since merging the main.js file is not reasonable, I recreated it after merging. This requires another commit. But I cannot squash this commit, as the important merge information for git seems to get lost.

BTW: These are some of the issues I mentioned in #1208. IMHO there are too many issues with squashing locally.

karussell:
> Yes, this is not possible if you need to merge the master

Yes, if you intend to squash things you should not do this. In those case people rebase the branch to the master, not merging the master into the branch.

Or you do it the lazy 'reset' way: https://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit but then you need to recreate the feature branch

> IMHO there are too many issues with squashing locally

They would have the same issues remotely if they'd do it that way so we need to find a simple method that always works...

boldtrn:
> Yes, if you intend to squash things you should not do this.

But doesn't this unnecessarily complicates the whole development process? It should be always possible to update a PR from the master branch. Also what happens with merge conflicts that cannot be solved easily like the main.js file?

> we need to find a simple method that always works...

Yes absolutely.

karussell:
>But doesn't this unnecessarily complicates the whole development process?

You do it via rebase, when I have more time I'll dig into it. But as github has an automated way to do this server side there is a simple method :D

> we need to find a simple method that always works

The reset method in the link above always works btw ...

boldtrn:
> The reset method in the link above always works btw ...

Ok, I can try this method for the next PR then. 

> But as github has an automated way to do this server side there is a simple method :D

I know, and I really like the GitHub way :).

Issue #553, opened 2015-10-30
Exampel for bad bicycle routing in Bad Hindelang (Germany)


bkuisle:
Hi,
I'm very new here an I hope I'm right!
I'm a very heavy user of graphhopper as a (bicycle) biker and mostly very satisfied about this router. 
But now I have planned a route for next weekend and found this:

https://graphhopper.com/maps/?point=47.505802%2C10.367285&point=47.506128%2C10.368208&locale=de-DE&vehicle=bike&weighting=fastest&elevation=true&layer=Omniscale

All streets are residential.

Regards
Bernhard Kuisle (Alias OSM b*k*)

karussell:
Thanks (again :)) and the reason could be some problems with elevation see #384 and #517 and the discussion [here](https://discuss.graphhopper.com/t/current-configuration-of-online-demo/218) where a user of GH tries to improve elevation. But this is important to get fixed!

ratrun:
This is the same thing as analysed in #330 and the reason for https://discuss.graphhopper.com/t/elevationweighting-implementation-and-bike2/213. Even if you have 100% right elevation data there always will be findings like this.
@bkuisle: You can try the MTB bike or the racebike profile: These profiles do not weight any elevation data. If you know this segment, please let us know how steep it is in real. According to our data and mathematics we use it is a 10 meters ascend on 82 meters. For such an ascend our calculated speed is about 6km/h. The longer segement has 12 meters ascend. But ss the ascend is much lower, the calculated speed is faster.

karussell:
@bkuisle do you have any real elevation or speed data for us here?

bkuisle:
Hi, thanks for asking.
Yes, I rode it with my bike in Nov 2015 equipped with my ciclomaster bicyle computer to measure the elavations (not absolut, but relatively exact)  and the distances. I startet at the crossing Karl-Haffner-Str. - Marktstraße via Badstraße (as proposed by Graphhopper) to crossing Zillenbachstraße-Schützenstraße and back the direct way.
![graphhopper - routing in hindelang - ciclomaster data](https://cloud.githubusercontent.com/assets/15415692/12446325/a6eaf554-bf69-11e5-90d5-ad94c97d899b.jpg)

Attached are the datas in an screenshot from an excel table.

karussell:
[...]

PR #1220, opened 2017-11-15
Road Elevation Smoothing

### LINKED ISSUE #517: Smoothing of elevation
Discussed in #209 but should happen on import and GPL code is really a no go and we cannot reuse the core method from this issue - we cannot even look into it.

The smoothing should happen on import e.g. could be here #507

Relates to #384

boldtrn:
This PR fixes/approaches #517. 

The problem with GH currently is that the elevation data contains spikes and the overall calculated ascend and descend are sometimes wrong by a factor of two or more.

Because of the increased calculation of ascend and descend, the distance can be wrong as well, because GH calculates the distance of a road using lat, lon, and elevation.

This PR proposes a rather simple approach. Before adding an edge to the graph, we smooth it's elevation, removing most spikes. Take the geometry of the edge and for every "pillar" but not "tower" node of the geometry, look at the surrounding points and calculate the average of these points.

**TODO**:
~~I am not sure if it would be better to rate points that are farther away with a lower weight?~~ I tried this briefly and this led to worse results. Probably because outliers will have a greater impact.

Here are some comparisons:

[Schauinsland](https://graphhopper.com/maps/?point=47.977512%2C7.848358&point=47.830213%2C7.940369&waypoint_max_distance=0)
GH Maps 29.46km ↗1891m ↘1515m
1220schauinsland-gh

PR 28.96km ↗1208m ↘828m
1220schauinsland-pr

[Google Maps](https://goo.gl/maps/Sa9Ajp177m92) 28.0km ↗902m ↘542m
1220schauinsland-gm

[From Grabenstetten to St. Johan](https://graphhopper.com/maps/?point=48.52462%2C9.451933&point=48.457555%2C9.3437&waypoint_max_distance=0).

GH Maps 15.72km ↗907m ↘863m
1220johann-gh

PR 15.37km ↗491m ↘446m
1220johann-pr

[Google Maps](https://goo.gl/maps/mQURguMq42Q2) 15.3 km ↗328m ↘277m
1220johann-gm

[Neuffen](https://graphhopper.com/maps/?point=48.548433%2C9.386015&point=48.535136%2C9.407644&waypoint_max_distance=0).

GH Maps 5.15km ↗613m ↘332m
1220neuffen-gh

PR 4.86km ↗360m ↘82m
1220neuffen-pr

[Google Maps](https://goo.gl/maps/zGsdJ8gPUx52) 4.8km ↗293m ↘14m
1220neuffen-gm

I also tried routes #330 and #533. It seems like both issues are already fixed because there has been a CGIAR data update recently. However I showed how the smoothing works in these cases with a screenshot below.

Routes from #330. 
[Herrnstein](https://graphhopper.com/maps/?point=47.8947%2C16.1056&point=47.900074%2C16.106408&vehicle=bike)
1220herrnstein-unsmoothed
1220herrnstein-smoothed

Routes from #553.

[Bad Hindelang](https://graphhopper.com/maps/?point=47.505802%2C10.367285&point=47.506128%2C10.368208&locale=de-DE&vehicle=bike&weighting=fastest&elevation=true&layer=Omniscale)
1220badhindelang-unsmoothed
1220badhindelang-smoothed


boldtrn:
I just tried to apply a [LOESS](https://en.wikipedia.org/wiki/Local_regression) regression instead of a "simple average" approach. For most cases this produced comparable or slightly better results than the average function, but sometimes it completely failed resulting in elevation values of 0, for example for [this route](https://graphhopper.com/maps/?point=48.52462%2C9.451933&point=48.457555%2C9.3437&locale=en-AU&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale), I just commited it so that interested users could try this approach or maybe there is some room for tuning the LoessInterpolator? Unless we find a way to fix these issues, I would remove the Loess approach for the final PR again.

boldtrn:
I updated this PR to also smooth the Tower Nodes, I am not 100% sure yet, if this is a good thing to do or not. Most cases improved due to that, but a few became worse. Maybe it might be better to use a high precision elevation data source and only smooth pillar nodes?

boldtrn:
I updated this PR. The smoothing happens now in the OSMReader.

For two reasons:
1. If we change the elevation, we have to update the FlagEncoders like Bike2Weight after the elevation was updated.
2. The elevation spikes actually lead to a wrong distance. One example would be [this](https://graphhopper.com/maps/?point=48.548745%2C9.385672&point=48.53596%2C9.4067&locale=en-AU&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale) route. GraphHopper Maps states it is 4.91km, after the smoothing it is 4.65km. [Google Maps](https://goo.gl/maps/nyivqKXb4g42) 4.6km. [Here](https://wego.here.com/directions/mix/L1250,-72639-Neuffen,-Germany:48.54873,9.38564/L1250,-72639-Neuffen,-Germany:loc-dmVyc2lvbj0xO3RpdGxlPUwxMjUwO2xhdD00OC41MzU5Njtsb249OS40MDY3O3N0cmVldD1MMTI1MDtjaXR5PU5ldWZmZW47cG9zdGFsQ29kZT03MjYzOTtjb3VudHJ5PURFVTtzdGF0ZUNvZGU9QmFkZW4tV3VydHRlbWJlcmc7Y291bnR5PUVzc2xpbmdlbjtjYXRlZ29yeUlkPXN0cmVldC1zcXVhcmU7c291cmNlU3lzdGVtPWludGVybmFs?map=48.54234,9.39658,16,normal&msg=B307%20Bayrischzell) 4.6km.

Currently it is designed as an "opt-in" feature, so you have to specify `graph.elevation.smooth_elevation=true`.

karussell:
Cool, thanks! So it is no longer a WIP?

boldtrn:
> Cool, thanks! So it is no longer a WIP?

I need to go through the code one more time. It is working, I just want to clean up a couple of minor things and need to fix some tests, etc. Feel free to try it though :). I think it should be ready for review tomorrow :+1:.

boldtrn:
> Cool, thanks! So it is no longer a WIP?

I updated and cleaned up this PR. I think this should be good to merge now.

reader-shp/src/main/java/com/graphhopper/reader/shp/OSMShapeFileReader.java
```diff
@@ -246,6 +246,12 @@ public DataReader setWayPointMaxDistance(double wayPointMaxDistance) {
         return this;
     }
 
+    @Override
+    public DataReader setSmoothElevation(boolean smoothElevation) {
+        // TODO implement elevation smoothing for shape files
+        return this;
+    }
```
boldtrn: Do we need elevation smoothing for shape files?

karussell: Can be done later

core/src/main/java/com/graphhopper/reader/dem/RoadElevationInterpolator.java
```diff
@@ -0,0 +1,48 @@
...
+            for (int i = 1; i < geometry.size() - 1; i++) {
+                int start = i - MAX_SEARCH_RADIUS < 0 ? 0 : i - MAX_SEARCH_RADIUS;
+                // +1 because we check for "< end"
+                int end = i + MAX_SEARCH_RADIUS + 1 >= geometry.size() ? geometry.size() : i + MAX_SEARCH_RADIUS + 1;
+                double sum = 0;
+                int counter = 0;
+                for (int j = start; j < end; j++) {
+                    // We skip points that are too far away, important for motorways
+                    if (MAX_SEARCH_DISTANCE > Helper.DIST_PLANE.calcDist(geometry.getLat(i), geometry.getLon(i), geometry.getLat(j), geometry.getLon(j))) {
+                        sum += geometry.getEle(j);
```
karussell: We could even let the distance influence the average: https://en.wikipedia.org/wiki/Weighted_arithmetic_mean

core/src/main/java/com/graphhopper/reader/dem/RoadElevationInterpolator.java
```diff
@@ -0,0 +1,48 @@
...
+public class RoadElevationInterpolator {
+
+    private final static int MIN_GEOMETRY_SIZE = 3;
+    // The max amount of points to go left and right
+    private final static int MAX_SEARCH_RADIUS = 8;
+    // If the point is farther then this, we stop averaging
+    private final static int MAX_SEARCH_DISTANCE = 180;
+
+    public static PointList smoothElevation(PointList geometry) {
+        if (geometry.size() >= MIN_GEOMETRY_SIZE) {
```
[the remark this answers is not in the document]
boldtrn: Yes true, I had it there because I was experimenting with different min geometry sizes to see if there is a difference. Also when I played with the LOESS regression, this required at least 7? points to work.

core/src/main/java/com/graphhopper/reader/dem/RoadElevationInterpolator.java
```diff
@@ -0,0 +1,48 @@
...
+            for (int i = 1; i < geometry.size() - 1; i++) {
+                int start = i - MAX_SEARCH_RADIUS < 0 ? 0 : i - MAX_SEARCH_RADIUS;
+                // +1 because we check for "< end"
+                int end = i + MAX_SEARCH_RADIUS + 1 >= geometry.size() ? geometry.size() : i + MAX_SEARCH_RADIUS + 1;
```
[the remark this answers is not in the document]
boldtrn: Points are not evenly scattered. For highways one point might be far away. You are right about the naming, I will improve this, thanks.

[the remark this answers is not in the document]
karussell: But isn't the search distance the only thing we want?

core/src/main/java/com/graphhopper/reader/dem/RoadElevationInterpolator.java
boldtrn: What I tried so far was to weight closer points as more important. This made the results worse, see my comment (copied it below). What we could try is to weight closer points with a lower weight. But this sounds wrong to me, but I could try it if you think this would make sense?

> I am not sure if it would be better to rate points that are farther away with a lower weight? I tried this briefly and this led to worse results. Probably because outliers will have a greater impact.

karussell: Ah, ok. Did you apply the changed formula ala `(x*ele1+y*ele2)/(x+y)`?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -554,6 +555,7 @@ public GraphHopper init(CmdArgs args) {
 
         // elevation
         String eleProviderStr = args.get("graph.elevation.provider", "noop").toLowerCase();
+        this.smoothElevation = args.getBool("graph.elevation.smooth_elevation", false);
```
karussell: also I'd slightly prefer to avoid "elevation" wording duplication and instead use `graph.elevation.smoothing`?

core/src/main/java/com/graphhopper/reader/dem/RoadElevationInterpolator.java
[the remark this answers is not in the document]
boldtrn: > But isn't the search distance the only thing we want?

Good point :). I will have to try and compare the results :+1:.

core/src/main/java/com/graphhopper/reader/dem/RoadElevationInterpolator.java
boldtrn: Yes, but I didn't use the actual distance from the points but onĺy the array index. I can do more experiments using the weighted arithmetic mean if you want?

core/src/main/java/com/graphhopper/GraphHopper.java
boldtrn: Yes makes sense, I will change this :).

core/src/main/java/com/graphhopper/reader/dem/RoadElevationInterpolator.java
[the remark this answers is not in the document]
boldtrn: I did that. I think this makes indeed more sense. Overall the results weren't huge in my tests. I think we might need to improve this over time if we find cases where the smoothing is not ideal.

karussell:
There seems to be a test failure (or is this unrelated?)

  GraphHopperWebIT.testRetrieveOnlyStreetname:123 expected:<[Turn right onto B 24]6> but was:<[Continue onto B 9]6>

boldtrn:
> There seems to be a test failure (or is this unrelated?)

This seems to be unrelated, but I fixed it anyway. The routing on the server changed, not sure why, but [the route](https://graphhopper.com/maps/?point=52.261434%2C13.485718&point=52.399067%2C13.469238&locale=en-AU&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=OpenStreetMap) looks OK to me.

karussell:
Ah, ok. Didn't notice we are testing here against the server - thanks!

PR #1225, opened 2017-11-19
Extracts new class NodeContractor from PrepareContractionHierarchies.

ammagamma:
This PR extracts a new class `NodeContractor` from `PrepareContractionHierarchies`.
The idea is this:

* `NodeContractor` is responsible for searching for witness paths and adding required
shortcut edges for a single given node that shall be contracted

* `PrepareContractionHierarchies` is responsible for determining the order
in which nodes are contracted and provides the interface to start the
contraction of a graph

This separation of concerns should make it easier to extend contraction hierarchies for
new features and applications (like turncosts, dynamic ch or transit node routing). 
It also makes it easier to test the contraction of a single node and hopefully increases
readability.

I tried to make the code changes easy to follow in the single commits. In the first commits
basically no real code changes are made except some reordering, renaming and delegation. I also 
stumbled over a few minor issues in the existing `PrepareContractionHierarchies` which should now
be fixed:

1) In `setOrigEdgeCount()` an `IllegalStateException` was thrown that gives information about 'graph.max'
I think this was supposed to return the number of all edges in the graph (original + shortcuts),
but `ghStorage.getAllEdges().getMaxId()` returns only the number of original edges, which is redundant
to 'max' in the message. 

2) The `tmpDegreeCounter` variable in `contractNodes` does not really count the degree of the node,
because it is only increased for each incoming (but not outgoing) edge. This is a bit confusing
when reading the code, but otherwise is not very crucial, because the degree calculation is 
only used to estimate the `meanDegree` of the graph, which in turn is used to estimate a good upper
bound of the `maxVisitedNodes` for the local witness path searches. I tried to document this a bit,
but maybe there is a better name than 'degree' ?

3) When calling `doWork() `in `PrepareContractionHierarchies` the priority for all nodes is calculated
in `prepareNodes()`. However, at this point `meanDegree` is zero, and so in `findShortcuts`
a `maxVisitedNodes` of 0 is set on `prepareAlgo`. This means that no witness paths can be found
and all shortcuts are counted (but not added).
For example this can be seen in `PrepareContractionHierarchiesTest#testAddShortcuts`.
When `contractNodes()` is called `meanDegree` is finally set to a non-zero value and when actually
inserting the shortcuts the witness path searches work as expected.
This was an easy fix, but required adjusting a few tests, because `PrepareContractionHierarchies`
now finds a different (potentially better) node ordering.

4) when calculating `maxVisitedNodes` from `meanDegree` it was cast like this:
`(int) meanDegree * 100`, so `maxVisitedNodes` was 0 if `meanDegree` < 1
I changed this to `(int) (meanDegree * 100)`, which yields a different (probably better) node
contraction order.

The last two fixes affect the node ordering when contracting the graph.
I ran the measurement class on the latest OSM data for Hessen and the performance
seems to be similar before and after the refactoring. This version introduces a few (~1-2%) less shortcuts, takes longer to prepare the graph (~10%) and leads to faster queries (for example routingCH.sum was reduced by 5%). All this seems plausible given that in the prior version `maxVisitedNodes` was sometimes zero, i.e. the witness path searches were cancelled quickly and therefore more shortcuts were introduced. I did not make exhaustive measurements (different maps, multiple runs etc.) though. 
In the new version the node ordering should be closer to what was intended, but maybe the parameter
tuning in `PrepareContractionHierarchies` needs to be adjusted. 

One thing I am not sure of is if we need `NodeContractor#initFromGraph`, because so far it is always
called directly after the constructor. However, if we call the constructor and change the graph
afterwards we need to call `initFromGraph()` to make `NodeContractor` aware of the changes. Alternatively,
we could call `initFromGraph()` inside `NodeContractor` as needed (for example in `contractNode()` etc.).

Finally, I stumbled over a weird behavior of `AllCHEdgesIterator` where the skippedEdge values are equal
although they should be different, see `NodeContractorTest#testContractNode_directed_shortcutRequired`.
I have not looked into this, any idea what the reason for this could be ?


karussell:
Thanks, this looks nice, will try what this means for bigger areas.

> In the new version the node ordering should be closer to what was intended, but maybe the parameter tuning in PrepareContractionHierarchies needs to be adjusted.

Yeah, this is really time intense but might be worth trying different periodicUpdatesPercentage, lastNodesLazyUpdatePercentage and neighborUpdatePercentage

> Finally, I stumbled over a weird behavior of AllCHEdgesIterator where the skippedEdge values are equal although they should be different

Hmmh, this might be a bug where we need to skip invalid edges (maybe edges from CH disconnection)

ammagamma:
> Yeah, this is really time intense but might be worth trying different periodicUpdatesPercentage, lastNodesLazyUpdatePercentage and neighborUpdatePercentage

Also it might be worth trying different linear combinations in calculatePriority

> Hmmh, this might be a bug where we need to skip invalid edges (maybe edges from CH disconnection)

It looks like `AllEdgesCHIterator.getSkippedEdge1/2` always returns 0. What do you mean 'we need to skip invalid edges' ? You mean this behavor is intentional or something needs to be changed ? In the unit tests I would have liked to get all shortcut edges directly from the graph after the contraction to be able if the shortcuts I am expecting have been added (with the right weight, distance and skippedEdge1/2).

karussell:
> It looks like AllEdgesCHIterator.getSkippedEdge1/2 always returns 0

hmmh, I need to refresh my memory here :) but yes, this might be a bug. 

> What do you mean 'we need to skip invalid edges' ? 

When we call prepareGraph.disconnect in the CH preparation there will be edges afterwards that are no longer used from the graph, but are still present in the edge storage. And we skip them (maybe here is a bug).

BTW: In graph.optimize we remove unused nodes coming from PrepareRoutingSubnetworks but we do not remove unused edges yet, as removing nodes is already a too complicated thing.

ammagamma:
I ran a few more performance tests with the OSM data for all of Germany. Its not easy to draw clear conclusions (without doing many many runs), but it looks like the 'fixed' node ordering slows down the prepare CH time (`prepare.ch.time` in the `Measurement` class) by a few (5-10) percent. The query time does not seem to be improved significantly, although about 1.5% less shortcuts are introduced. 
Therefore I have reverted the fixes to the node ordering, and simply left a comment about the issues for now. I would postpone fixing them until somebody takes the time to re-optimize the parameters in PrepareContractionHierarchies.

karussell:
Thanks, this makes sense. Have added a note in #240.

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
```diff
@@ -432,398 +463,22 @@ int calculatePriority(int v) {
...
-            }
-        } else {
-            throw new IllegalArgumentException("Algorithm " + opts.getAlgorithm() + " not supported for Contraction Hierarchies. Try with ch.disable=true");
-        }
-
-        algo.setMaxVisitedNodes(opts.getMaxVisitedNodes());
-        algo.setEdgeFilter(levelFilter);
-        return algo;
+    private int getMaxVisitedNodesEstimate() {
+        // todo: we return 0 here if meanDegree is < 0, which is not really what we want, but changing this changes
```
karussell: did you mean "we return 0 here if meanDegree is < 1" instead of <0?

ammagamma: sure ;)

karussell:
Sorry I forgot about this. Do you think I can squash & merge it or do you want to preserve some commits separated?

ammagamma:
You can squash & merge, I will keep this branch in my fork. Not sure about these new code climate issues though, for example maximum number of parameters=4, maximum method length=25 lines, maximum file length=250 lines. I think they will be rather difficult to realize especially in test code and will for sure require many changes in existing code. I am nowhere against strong separation of concerns and short methods, but I don't think such strict rules are the best way to achieve this. I think it is better to let the programmer(s) decide (via code reviews for example ;).

karussell:
Yeah, the codeclimate stuff is rather restrictive and for some things I find it valuable but for the most things not really. Probably we should remove it for now.

PR #1226, opened 2017-11-19
Makes AllCHEdgesIteratorImpl read/write skipped edges to/from shortcuts DataAccess

ammagamma:
get/setSkippedEdge1/2 did not really work so far, but was not used anywhere either.
Also see discussion in #1225.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/CHGraphImpl.java
+++ b/core/src/main/java/com/graphhopper/storage/CHGraphImpl.java
@@ -599,18 +599,25 @@ public class CHGraphImpl implements CHGraph, Storable<CHGraph> {
 
         @Override
         public final void setSkippedEdges(int edge1, int edge2) {
-            baseGraph.edges.setInt(edgePointer + S_SKIP_EDGE1, edge1);
-            baseGraph.edges.setInt(edgePointer + S_SKIP_EDGE2, edge2);
+            checkShortcut(true, "setSkippedEdges");
+            if (EdgeIterator.Edge.isValid(edge1) != EdgeIterator.Edge.isValid(edge2)) {
+                throw new IllegalStateException("Skipped edges of a shortcut needs "
+                        + "to be both valid or invalid but they were not " + edge1 + ", " + edge2);
+            }
+            shortcuts.setInt(edgePointer + S_SKIP_EDGE1, edge1);
+            shortcuts.setInt(edgePointer + S_SKIP_EDGE2, edge2);
         }
 
         @Override
         public final int getSkippedEdge1() {
-            return baseGraph.edges.getInt(edgePointer + S_SKIP_EDGE1);
+            checkShortcut(true, "getSkippedEdge1");
+            return shortcuts.getInt(edgePointer + S_SKIP_EDGE1);
         }
 
         @Override
         public final int getSkippedEdge2() {
-            return baseGraph.edges.getInt(edgePointer + S_SKIP_EDGE2);
+            checkShortcut(true, "getSkippedEdge2");
+            return shortcuts.getInt(edgePointer + S_SKIP_EDGE2);
         }
 
         @Override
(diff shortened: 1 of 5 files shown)


karussell:
Making setSkippedEdges more expensive is ok, but get might be not worth. Are there measurable slowdowns?

core/src/main/java/com/graphhopper/storage/CHGraphImpl.java
```diff
@@ -599,18 +599,25 @@ public int getMaxId() {
...
         public final void setSkippedEdges(int edge1, int edge2) {
-            baseGraph.edges.setInt(edgePointer + S_SKIP_EDGE1, edge1);
-            baseGraph.edges.setInt(edgePointer + S_SKIP_EDGE2, edge2);
+            checkShortcut(true, "setSkippedEdges");
+            if (EdgeIterator.Edge.isValid(edge1) != EdgeIterator.Edge.isValid(edge2)) {
+                throw new IllegalStateException("Skipped edges of a shortcut needs "
+                        + "to be both valid or invalid but they were not " + edge1 + ", " + edge2);
+            }
+            shortcuts.setInt(edgePointer + S_SKIP_EDGE1, edge1);
+            shortcuts.setInt(edgePointer + S_SKIP_EDGE2, edge2);
```
karussell: Doah, what an ugly bug!

karussell:
Ah, ok. As it isn't used there are no current perf implications (at least for our code)

karussell:
The codeclimate issues are really strange (they report stuff as new which cannot be new)

karussell:
@ammagamma would you mind verifying your commits? See discussion #1208 (Not that important right now but would like to migrate to a fully verified git history in the long term)

ammagamma:
> Ah, ok. As it isn't used there are no current perf implications (at least for our code)

No, no performance implications for our code, but even foreign code most likely did not use it because the code was wrong (?). Does it even make sense to use the `S_SKIP_EDGE1/2` offset in `baseGraph.edges` as it was done before or does this accidentally overwrite some other values ?

> The codeclimate issues are really strange (they report stuff as new which cannot be new)

To be honest for some of them I do not even know what to do, like this one:
> Javadoc comment at column 17 has parse error. Details: mismatched input ':' expecting <eof> while parsing JAVADOC</eof>

Is there any way to run code climate locally on my machine ? Or am I supposed to fix code and push changes until it is green on github ? For example for SonarQube there are very nice IntelliJ/Eclipse plugins (sonar lint) that show you any errors while editing the code. However, in my experience the problem with these tools is that you often get many false positives and in the worst case the code gets worse just so some 'issues' are fixed ;)

>verifying your commits

 ok

karussell:
> Is there any way to run code climate locally on my machine ?

Running findbugs and stylecheck is similar ... not sure if we keep the codeclimate thing, we are really trying this out only

karussell:
One last thing: would you mind trying to [squash your changes](https://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit)?

ammagamma:
on this same branch ? it will change the git hashes

karussell:
Yes, then push force (which one should never do on a branch shared with others!)

With this method we can merge the verified commits into the master and then they are still verified. See the last 2 merge commits: https://github.com/graphhopper/graphhopper/commits/master

As explained in the other issue, this is a bit more work for the contributor (and we need to figure out if the hassle is worth it) but has the advantages of 

1. a cleaner history making it easier in other branches
2. a fully verified history of the master branch with the commits directly from the PRs

ammagamma:
ok yes, I was just concerned because I try never to change any history on published branches. this one is not really 'shared' with others, but still published. ok, will squash and force push.

karussell:
Êven if shared with others this is not really a problem as long as the other parties have no changes and do a git pull :D ... I once lost my changes due to another one push forced into the same branch.

ammagamma:
ok, now the two old previous commits are gone, but at least the conversation in the pr is unchanged

karussell:
Thanks!

karussell:
Again, this "verify stuff" with git squash is kind of an experiment and can fail if contributing is too ugly or other disadvantages.

PR #1228, opened 2017-11-20
Importing GMTED2010 elevation data and MultiSourceElevationProvider

boldtrn:
This PR is a clone of #1217. 

You can now import GMTED2010 elevation data, which has roughly a 250m resoultion. You can mix SRTM data with GMTED data by using the new MultiSourceElevationProvider by using GMTED data outside of the SRTM area.

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/dem/CGIARProvider.java
+++ b/core/src/main/java/com/graphhopper/reader/dem/CGIARProvider.java
@@ -17,26 +17,17 @@
  */
 package com.graphhopper.reader.dem;
 
-import com.graphhopper.storage.DAType;
 import com.graphhopper.storage.DataAccess;
-import com.graphhopper.storage.Directory;
-import com.graphhopper.storage.GHDirectory;
-import com.graphhopper.util.Downloader;
 import com.graphhopper.util.Helper;
 import org.apache.xmlgraphics.image.codec.tiff.TIFFDecodeParam;
 import org.apache.xmlgraphics.image.codec.tiff.TIFFImageDecoder;
 import org.apache.xmlgraphics.image.codec.util.SeekableStream;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.awt.image.Raster;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.SocketTimeoutException;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/graphhopper/reader/dem/MultiSourceElevationProvider.java
```diff
@@ -0,0 +1,108 @@
...
+        this.gmtedProvider = gmtedProvider;
+    }
+
+    public MultiSourceElevationProvider() {
+        this(new CGIARProvider(), new GMTEDProvider());
+    }
+
+    @Override
+    public double getEle(double lat, double lon) {
+        // Sometimes the cgiar data north of 59.999 equals 0
```
boldtrn: I just found some examples where this happens, so I decided to use GMTED data slightly before 60° which seems to fix the issue.

core/src/main/java/com/graphhopper/reader/dem/CGIARProvider.java
```diff
@@ -80,85 +66,33 @@ public static void main(String[] args) {
...
-        return this;
-    }
-
     @Override
     public double getEle(double lat, double lon) {
-        // no data we can avoid the trouble
-        if (lat > 60 || lat < -60)
+        // Return fast, if there is no data available
+        // See https://www2.jpl.nasa.gov/srtm/faq.html
+        if (lat > 60 || lat < -56)
```
karussell: Don't we need the same 59.999 border like in the multi ele provider?

karussell: Ah, maybe we need here if lat>=60 and in the multi ele provider `if(lat < 60)`?

boldtrn: > Don't we need the same 59.999 border like in the multi ele provider?

Shouldn't really matter, as outside of these borders the values will be 0 anyway. Also I think that the > 59.999 is not necessarily true globally. I tried a couple of routes in Sweden and some showed this effect.

> Ah, maybe we need here if lat>=60 and in the multi ele provider if(lat < 60)?

True this would make more sense. But the values were for lats like 59.999632 and not 60.

boldtrn:
I updated this PR to your comments, should be good to merge?

karussell:
One question: if multi is used how do I specify a separate directory for gmted and cgiar? My first idea would be to use `graph.elevation.<provider name>.cache_dir` - wdyt?

boldtrn:
> One question: if multi is used how do I specify a separate directory for gmted and cgiar?

Yes this is a good idea. I think we should make it possible to configure the providers used in the MultiElevationProvider. Currently for the global provider there is only the GMTED provider, but the SRTM provider could be specified as srtm, cgiar, or srtmgl1. In the future the global provider could be changed to ALOS data or a similar provider.
 - `graph.elevation.multi.srtm=cgiar`
 - `graph.elevation.multi.global=gmted`
 - `graph.elevation.multi.srtm.cache_dir=~/srtm/`
 - `graph.elevation.multi.global.cache_dir=~/gmted/`

This follows the naming scheme I proposed [here](https://github.com/graphhopper/graphhopper/pull/1235/files#diff-10a05bb2835502ef4ab61dbc99bba02eR30).

Issue #1131, opened 2017-08-05
Instructions: parse turn:lanes to improve instructions


boldtrn:
There is valuable instruction information stored the [turn:lanes information in OSM](https://wiki.openstreetmap.org/wiki/Key:turn#Turning_indications_per_lane). It would be interesting to see if we can improve the turn instruction by parsing them in a post-processing step and maybe add them as annotation or something like that?

I wouldn't go as far and propose to actually support turn lane information in the response, which might be nice in the future, but probably not now.

aramzl:
Hi I just added this information to graphhopper, hopping to create a pull-request. 
Can anyone help me out on that? I need permissions
@karussell 
@boldtrn

aramzl:
I figured out ;)

PR #1232, opened 2017-11-22
Added lane information based on the osm tag turn:lanes

aramzl:
lane_info=true in the encoder string to enable it, it requrires 8
bytes_for_flags.

@karussell 
@boldtrn

Issue #1131


boldtrn:
Thank you very much for your contribution and sorry for the late reply. This PR looks very impressive. Having turn lane support would be great.

Since turn lanes are not important for the routing itself and are only required for the TurnInstruction generation, I would probably put them in a DataAccess, similar to the NameIndex. As you have seen yourself, it requires a lot of bits, and bits are very precious :). Using a DataAccess we could even think about just storing the whole turn:lane information and not needing to encode it (I am not sure if we cover every possible turn lane combination?).

WDYT @karussell, could we merge something like this, using DataAccess before #1112 or do we have to wait until #1112 is finished?

Regarding the PR itself, you included the tag `turn:lanes`, but not `turn:lanes:forward` and `turn:lanes:backward`. I think this will make the code a bit more complicated, but a lot of turn lanes are tagged like this, so I think it would be required to have that information.

aramzl:
@boldtrn Thanks for the comments. I saw the turn:lanes:forward and turn:lanes:backward I wanted to make a second pull request including them. It will take more memory (more bits are needed).

boldtrn:
> @boldtrn Thanks for the comments. I saw the turn:lanes:forward and turn:lanes:backward I wanted to make a second pull request including them. It will take more memory (more bits are needed).

There is a way to get around this by using a DataAccess, similar to the NameIndex or TurnCostExtension. Let's wait for Feedback from @karussell before deciding our next steps :).

karussell:
Thanks @aramzl for this useful and big feature!

I'm not sure if we should merge such a complex thing before #1112 as this will highly likely needs a bigger refactoring. If you would have time it would be really interesting to see if you could base your PR on the mentioned branch. This could be also used as a test case if everything what we need is already possible

karussell:
Thanks @aramzl for this useful and big feature!

I'm not sure if we should merge such a complex thing before #1112 as this will highly likely needs a bigger refactoring. If you would have time it would be really interesting to see if you could base your PR on the mentioned branch. This could be also used as a test case if everything what we need is already possible

aramzl:
I would try to redo it based on #1112 and create another pull request.

karussell:
This is awesome! Please note that this is cutting edge, and could potentially use lots of your time ... we really appreciate your feedback e.g. if this feels better from a programming perspective.

karussell:
Will close here, then resurrect #1269 (include the tests from here, maybe wait until #1548 is merged)

Sorry that this takes so long @aramzl. Now that #1447 was merged this is a much better time to do your change. Have marked #1269 for 0.13

PR #1235, opened 2017-11-29
Use SRTM 1 arc second elevation data

### LINKED ISSUE #384: Use high precision elevation data

boldtrn:
This PR partially fixes #384.

With this PR we add the functionality to use SRTM 1 arc second data. I did a couple of tests so far it looks a lot better than the CGIAR data. It would be interesting to see it working together with #1220. Both PRs seem to change the data in the same direction, with both combined I think we would see a very smooth and nice looking elevation diagram. Also this would probably improve calculated routes a lot.

At first I wanted to use the [ALOS data](http://www.eorc.jaxa.jp/ALOS/en/aw3d30/), but decided against it after reading more reviews about the ALOS data. In general it is considered to be one of the most detailed data sources, but it has some major flaws in some areas:
- voids (there is a void filled version in the SRTM area)
- Issues with the [post-processing](http://www.imagico.de/map/alos.php). ALOS (as well as ASTER GDEM) is based on images which have to be preprocessed. It struggles with clouds, water, and buildings. So sometimes the elevation is a mix building height and street height. Sometimes there are elevation changes on a water surface, etc. 
- Some data corrections appear to be incorrect or insufficient. This was mentioned in some reviews ([1](http://www.imagico.de/map/alos.php, [2](https://www.preprints.org/manuscript/201708.0081/v1/download)). For example citing [2]:
> However, while analyzing the results, it has become apparent that it also contains
a widespread elevation anomaly. We conclude that this anomaly is the result of uncompensated
sensor noise and the data processing algorithm (downsampling of the higher resolution data).

On the other hand some studies found that the AW3D30 data [has less errors](https://www.int-arch-photogramm-remote-sens-spatial-inf-sci.net/XLI-B4/149/2016/isprs-archives-XLI-B4-149-2016.pdf) than SRTM data.

I know that SRTM has similar issues especially with water surfaces being not perfectly even. Combining AW3D30, SRTMGL1, and GMTED, in the MultiElevationProvider is probably our best bet on the long term.

Here are a couple of comparison (I used the same as in #1220 ).

[Schauinsland](https://graphhopper.com/maps/?point=47.977512%2C7.848358&point=47.830213%2C7.940369&waypoint_max_distance=0)
GH Maps 29.46km ↗1891m ↘1515m
1220schauinsland-gh

PR 29.08km ↗1479m ↘1120m
1220schauinsland-pr

[From Grabenstetten to St. Johan](https://graphhopper.com/maps/?point=48.52462%2C9.451933&point=48.457555%2C9.3437&waypoint_max_distance=0).

GH Maps 15.72km ↗907m ↘863m
1220johann-gh

PR 15.45km ↗642m ↘595m
1235johann-pr

[Neuffen](https://graphhopper.com/maps/?point=48.548433%2C9.386015&point=48.535136%2C9.407644&waypoint_max_distance=0).

GH Maps 5.15km ↗613m ↘332m
1220neuffen-gh

PR 4.89km ↗393m ↘113m
1220neuffen-pr

[Herrnstein](https://graphhopper.com/maps/?point=47.8947%2C16.1056&point=47.900074%2C16.106408&vehicle=bike&waypoint_max_distance=0)
1220herrnstein-unsmoothed
1235fahrrad1pr

[Bad Hindelang](https://graphhopper.com/maps/?point=47.505802%2C10.367285&point=47.506128%2C10.368208vehicle=bike&waypoint_max_distance=0)
1220badhindelang-unsmoothed
1235fahrrad2pr


core/src/main/java/com/graphhopper/reader/dem/SRTMGL1Provider.java
```diff
@@ -0,0 +1,155 @@
...
+import java.io.*;
+
+/**
+ * SRTMGL1 contains elevation data for most of the world with 1 arc second (~30m) accuracy.
+ * We use the mirror of OpenTopography, as the official SRTMGL1 download requires authentication.
+ * http://opentopo.sdsc.edu/raster?opentopoID=OTSRTM.082015.4326.1
+ * <p>
+ * TODO: In addition, NSF requires oral acknowledgment during all news media interviews.
+ * TODO: This is weird, we have to check if really want to use OpenTopography then, see: http://www.opentopography.org/citations
+ * <p>
```
boldtrn: I find this very weird. Maybe we find a different source for the SRTM data.

> In addition, NSF requires oral acknowledgment during all news media interviews

These seems a bit much for just providing a download server :).

core/src/test/java/com/graphhopper/reader/dem/SRTMGL1ProviderTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Robin Boldt
+ */
+public class SRTMGL1ProviderTest {
```
karussell: I wonder: when we added a test isn't there also a file required under core/files/xy.hgt? Otherwise fresh tests won't run without internet or could be slow (?)

core/src/main/java/com/graphhopper/reader/dem/SRTMGL1Provider.java
```diff
@@ -0,0 +1,155 @@
...
+    private String getEastString(int lon) {
+        if (lon < 0) {
+            return "W";
+        }
+        return "E";
+    }
+
+    @Override
+    public String toString() {
+        return "srtmgl1";
```
karussell: Is there a reason we use lower case here and upper case in the other? I would prefer having the same case for all even if this means changing the old case.

core/src/main/java/com/graphhopper/reader/dem/SRTMGL1Provider.java
karussell: Indeed a bit strange.

Other question: how much larger is the full world wide data on disc. (CGIAR is 70GB)

karussell:
One minor question: did you do the screenshots without douglas peucker? (do not hesitate to create a PR that disables douglas peucker for debug=true :))

core/src/main/java/com/graphhopper/reader/dem/SRTMGL1Provider.java
boldtrn: > Is there a reason we use lower case here and upper case in the other?

Oh sorry I missed that. The GMTED provider is also lowercase. I can change both. I just thought it should be the same as the config value.

karussell: Makes more sense if they have the config value. Would this break some tests or usage scenarios?

core/src/test/java/com/graphhopper/reader/dem/SRTMGL1ProviderTest.java
boldtrn: All elevation provider (except the SRTMProvider) tests don't actually return the height of a coordinate, they only return the name of the file they are going to download. I haven't seen that there are files for the srtm provider. Should we add files for the SRTMGL1 provider as well? (note, the files are 25mb large and per default not zipped). CGIAR files are also around 30mb. GMTED are >200mb. Not sure if it's easy to create small dummy files?

This is actually something that bothers me as well. I was thinking about adding a test, but to put @Ignore at the test, so we can enable the test if we change something about the elevation provider, but don't use it most of the time. WDYT?

karussell: There are files at core/files/*hgt.zip

> Should we add files for the SRTMGL1 provider as well?

Ok, probably not :)

The idea with the `@Ignore` seems interesting but I think there is something in combination with maven for executing only a specific set of tests

> BTW: Our API/client tests probably fail if there is no internet as well (haven't tested that yet I think)?

Yes, we should include an option to exclude it or just run it on travis?

boldtrn:
> One minor question: did you do the screenshots without douglas peucker?

Yes all screenshots were taken with `waypoint_max_distance=0`. I also added the parameter in the sample requests links above.

> Other question: how much larger is the full world wide data on disc. (CGIAR is 70GB)

Good question. So every .hgt file is 25mb large. There are about 14000 files. So the overall size should be at around 340 gb. I haven't tested it though. According to [this link](https://www.usna.edu/Users/oceano/pguth/website/research/srtm_grid/guth_hpc_sc06.pdf) it's 318 gb.

BTW, according to [this link](https://mediatum.ub.tum.de/doc/1369138/file.pdf), ASTER GDEM is about 560gb.

core/src/main/java/com/graphhopper/reader/dem/SRTMGL1Provider.java
boldtrn: > Would this break some tests or usage scenarios?

I don't think so. I actually think the toString method is only used to do a System.out:
```
logger.info(this.toString() + " Elevation Provider, from: " + baseUrl + ", to: " + cacheDir + ", as: " + daType);
```

boldtrn:
> and do not hesitate to create a PR that disables douglas peucker for debug=true :)

I thought about this a bit. I am not sure if this is wanted/expected behavior. Think about you have a request and want to inspect it, you set debug=true and suddenly you receive more points. I think this would be unexpected. WDYT?

karussell:
> Think about you have a request and want to inspect it, you set debug=true and suddenly you receive more points. I think this would be unexpected.

You are probably right. And as avoiding douglas peucker is already possible we should avoid too many things on debug=true :)

karussell:
Looks like a good improvement. 

Can we reduce the code in the GraphHopper class a bit and why is the RAM increase necessary? (because of the tests with tiles of 30 degree resolution?)

boldtrn:
> Oh, this would be ugly. We need to have the memory really on a very low level to detect memory issues very early.

Sorry I should have mentioned that the last commit changing the memory limit was intended to show that the tests work, but Travis timed out. As discussed earlier, I will @Ignore these tests and set the memory limit back to the original value.

> Can we reduce the code in the GraphHopper class a bit

Sure :).

> why is the RAM increase necessary? (because of the tests with tiles of 30 degree resolution?)

Yes.

Sorry I should have mentioned that this PR wasn't considered mergeable after yesterday's commit :(.

karussell:
> Sorry I should have mentioned that this PR wasn't considered mergeable after yesterday's commit :(. > BTW: Will commit tomorrow, sorry for the delay :(

Really no problem ... all fine and please no hurry :)

boldtrn:
@karussell do you remember why we read the width of the tile in the SRTM provider (see [here](https://github.com/graphhopper/graphhopper/blob/1b1c22c7e4cfc5ebe4c97da7a6b34ba5ad7d4e3b/core/src/main/java/com/graphhopper/reader/dem/SRTMProvider.java#L232)): `int width = (int) (Math.sqrt(heights.getHeader(WIDTH_BYTE_INDEX)) + 0.5);` and in the CGIAR provider we just assume that the width would be the default (see [here](https://github.com/graphhopper/graphhopper/blob/1b1c22c7e4cfc5ebe4c97da7a6b34ba5ad7d4e3b/core/src/main/java/com/graphhopper/reader/dem/CGIARProvider.java#L109)). This is one major difference between SRTM and TIFF handling that makes it complicated to extract some of the logic into a common super class. 

[This](https://lpdaac.usgs.gov/sites/default/files/public/measures/docs/NASA_SRTM_V3.pdf) document describes the number of rows and colums for the SRTM files so it seems reasonalbe to me to assume that all files use the same width?

core/src/main/java/com/graphhopper/reader/dem/AbstractElevationProvider.java
```diff
@@ -1,14 +1,14 @@
...
- *  GraphHopper GmbH licenses this file to you under the Apache License, 
- *  Version 2.0 (the "License"); you may not use this file except in 
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
  *  compliance with the License. You may obtain a copy of the License at
- * 
+ *
  *       http://www.apache.org/licenses/LICENSE-2.0
- *
```
boldtrn: I just updated IntelliJ, they added formatting for Copyright Statements (I mean who hasn't been waiting for this great feature). Even if I disable formatting for copyright statements and for java doc IntelliJ still messes this up...

Before merging I could just overwrite all copyright statements that got changed by the autoformatting?

karussell:
I think we read&store the width because of the different resolutions (30 vs. 90). If you click blame you see this commit: https://github.com/graphhopper/graphhopper/commit/2e60c80beb19f46c14d9d3b5cccc084b873e4944

core/src/main/java/com/graphhopper/reader/dem/AbstractElevationProvider.java
karussell: sure

boldtrn:
> I think we read&store the width because of the different resolutions (30 vs. 90). If you click blame you see this commit

Ah ok makes sense. Since we have two different providers for this now, we could think about reverting this and define the width before. I think this would make the code a bit nicer and we might be able to abstract more and reduce the code duplication.

I updated the PR adding some more tests that check that the borders between the HeightTiles are aligned correctly.

BTW: A very handy tool to check elevation is the [elevation-service](http://www.liedman.net/elevation-service/) by @perliedman.

core/src/main/java/com/graphhopper/reader/dem/AbstractTiffElevationProvider.java
```diff
@@ -1,14 +1,14 @@
 /*
  *  Licensed to GraphHopper GmbH under one or more contributor
- *  license agreements. See the NOTICE file distributed with this work for 
+ *  license agreements. See the NOTICE file distributed with this work for
```
boldtrn: I tried to copy the copyright information from a different class in GraphHopper and I think that the applied formatting is actually correct and that the formatting before has been wrong, so I would keep it that way.

boldtrn:
I think this PR should be good to merge now. If we want to change the width handling of the SRTM provider I could further try to further melt down the code.

Something worth checking would be also if there is a major slow down due to the getFileName method. This method uses String.format and builds Strings with '+' instead of using StringBuilder. I would assume that the jit should be able to properly optimize these methods. If there is a performance issue I would revisit this method and try to improve it's performance.

karussell:
> I think this would make the code a bit nicer and we might be able to abstract more and reduce the code duplication.

Sure, let's write this down in another issue

> This method uses String.format and builds Strings with '+' instead of using StringBuilder

Yes, I think it is necessary to avoid String.format in this method (but you can easily prove me wrong via comparing this for world wide import ;)). Please see CGIARProvider where I wrote that String.format makes problems (btw: I do not think that strings and pluses are a problem due to the JIT you mentioned).

core/src/main/java/com/graphhopper/reader/dem/SRTMGL1Provider.java
kannes: I build a GeoTIFF earlier this year for fun (from the raw tiles, no hole fixing etc!): http://hannes.enjoys.it/blog/2017/01/one-srtmgl1-geotiff-to-rule-them-all/

The result was ~90GB without overviews, ~120GB with. I used `-co PREDICTOR 2 -co COMPRESS=DEFLATE -co TILED=YES` which from my understanding was ideal.
https://www.datenatlas.de/geodata/public/srtmgl1/

boldtrn: Thanks for the details @kannes. I think the newer SRTMGL1 are already hole/void fixed with ASTER GDEM data? 

See - https://lpdaac.usgs.gov/dataset_discovery/measures/measures_products_table/srtmgl1_v003: 
> Voids in the Version 3.0 products have been filled with ASTER Global Digital Elevation Model (GDEM) Version 2.0, the Global Multi-resolution Terrain Elevation Data 2010 (GMTED2010), and the National Elevation Dataset (NED)

boldtrn:
> Yes, I think it is necessary to avoid String.format in this method (but you can easily prove me wrong via comparing this for world wide import ;)).

String.format is definitely slower than using regular concatenation, so I just changed it. Even if the difference might be small, having a faster import is always better.

I think this PR should be mergeable now. Let me know if I should change anything?

boldtrn:
I did a minor change here. I kept the CGIAR provider as default provider in the MultiProvider so that we don't break any existing code. Over time it might make sense to set the SRTMGL1Provider as the default provider.

core/src/main/java/com/graphhopper/reader/dem/AbstractTiffElevationProvider.java
```diff
@@ -37,91 +32,122 @@
...
     /**
-     * Creating temporary files can take a long time as we need to unpack tiff as well as to fill
-     * our DataAccess object, so this option can be used to disable the default clear mechanism via
-     * specifying 'false'.
+     * Return true if the coordinates are outside of the supported area
      */
-    public void setAutoRemoveTemporaryFiles(boolean autoRemoveTemporary) {
-        this.autoRemoveTemporary = autoRemoveTemporary;
-    }
+    abstract boolean outsideSupportedArea(double lat, double lon);
```
karussell: I know this is not a setter but do we need Java code style here: isOutsideSupportedArea?

core/src/main/java/com/graphhopper/reader/dem/SRTMGL1Provider.java
karussell: What do you think about a new section below the software for data licenses [here](https://github.com/graphhopper/graphhopper/blob/master/NOTICE.md) (and explicitly mention what is used by default)?

I would add stuff we have [here](https://graphhopper.com/api/1/docs/attribution/#data) and also gtfs.

Maybe via a table

|source | license | used as default | included in repo |
|---------|-----------|---------|------|
|OSM for road network data | [ODBL](https://www.openstreetmap.org/copyright) | yes | yes
| GTFS NRW | TODO link | no | yes
| SRTM elevation | TODO link | no | yes
| CGIAR elevation | TODO link | no | yes
| SRTMGL1 elevation | TODO link | no | no
|OpenTopography mirror for SRTMGL1 | TODO  link | no | no

core/src/main/java/com/graphhopper/reader/dem/AbstractTiffElevationProvider.java
boldtrn: Makes sense :). I changed the method name.

boldtrn:
Merged this PR after updating the method name

karussell:
Thanks, will be deployed in the next days!

PR #1213, opened 2017-11-07
Use Policeman Tools

boldtrn:
As mentioned in #1210, it would be nice to use the policeman tools to avoid locale issues.

I just pasted the default configuration of the policeman tools. I changed it so it does only run on travis or by using `mvn forbiddenapis:check`. It finds a lot of issues for the core so far.

For example:
```
[ERROR] Forbidden method invocation: java.lang.String#toLowerCase() [Uses default locale]
[ERROR]   in com.graphhopper.routing.util.EncodingManager (EncodingManager.java:111)

if (!encoderList.equals(encoderList.toLowerCase()))
```

The fix that was proposed [here](http://blog.thetaphi.de/2012/07/default-locales-default-charsets-and.html), is to rewrite the above line to:
```
if (!encoderList.equals(encoderList.toLowerCase(Locale.ROOT)))
```

> As mentioned before, you can pass a locale to String.toLowerCase(), so the naive approach would be to tell Java, that we are in the US or using the English language: String.toLowerCase(Locale.US) or String.toLowerCase(Locale.ENGLISH). This produces identical results but is still not consistent. What happens if the US government decides to lowercase/uppercase like in Turkey? -- OK, don’t use Locale.US (this is also too US-centric). Locale.ENGLISH is fine and very generic, but languages also change over the years (who knows?), but we want to have it language invariant! If you are using Java 6, there is a much better constant: Locale.ROOT -- You should use this constant for our lowercase example: String.toLowerCase(Locale.ROOT).

Should I go ahead and try to fix all the found issues by the policeman tools?

The full list of issues for the core is:
```
[ERROR] Forbidden method invocation: java.lang.String#toLowerCase() [Uses default locale]
[ERROR]   in com.graphhopper.routing.util.EncodingManager (EncodingManager.java:119)
[ERROR] Forbidden method invocation: java.lang.String#format(java.lang.String,java.lang.Object[]) [Uses default locale]
[ERROR]   in com.graphhopper.routing.util.EncodingManager (EncodingManager.java:190)
[ERROR] Forbidden method invocation: java.lang.String#format(java.lang.String,java.lang.Object[]) [Uses default locale]
[ERROR]   in com.graphhopper.routing.util.EncodingManager (EncodingManager.java:196)
[ERROR] Forbidden method invocation: java.lang.String#format(java.lang.String,java.lang.Object[]) [Uses default locale]
[ERROR]   in com.graphhopper.routing.util.EncodingManager (EncodingManager.java:202)
[...]

Issue #1210, opened 2017-11-04
RealtimeIt#testExtraTrip fails for some local time zones


boldtrn:
I just ran the tests for the current master using the time one `UTC+10:00` and the test `RealtimeIt#testExtraTrip` failed. If I change my time zone to Shanghai `UTC+08:00` it works. If I change my time zone to Perth `AWST` (same time as `UTC+08:00` right now), it fails.

Maybe @michaz you have an idea what might be happening here? In the test we define the time zone, but maybe the server itself runs on a different time zone?

michaz:
Haha, nice. I probably left some debug or mock thing in there which uses the "current" time zone. Good that we are a global team. :)

michaz:
(Will look at it today)

karussell:
As we stumbled over this recently #1209 maybe we should use something like the policeman tools?

https://github.com/policeman-tools/forbidden-apis/wiki
http://blog.thetaphi.de/2012/07/default-locales-default-charsets-and.html

boldtrn:
> maybe we should use something like the policeman tools?

I second that! I think this is a good idea :). I created a PR #1213 testing the policeman tools.

boldtrn:
Any updated on this @michaz? Could we ignore this test until the problem is solved?

michaz:
Just a second..

michaz:
That was a really helpful ticket, thanks for reminding me!

boldtrn:
Thanks so much for fixing this @michaz!

PR #1236, opened 2017-11-30
Use Policeman Tools

boldtrn:
This is a clone of #1213.

As mentioned in #1210, it would be nice to use the policeman tools to avoid locale issues.

I just pasted the default configuration of the policeman tools. I changed it so it does only run on travis or by using `mvn forbiddenapis:check`. It finds a lot of issues for the core so far.


karussell:
Thanks!

Why is there a change to main.js?

boldtrn:
> Why is there a change to main.js?

Because it changed in the master branch. When merging it will create a conflict. The easiest solution for me is to simply regenerate the main.js, that's why it detects a change. When squashing a commit, that has been updated from the master, this happens and I don't know how to get around this.

karussell:
> Because it changed in the master branch. When merging it will create a conflict.

You should rebase it after that change so that this wouldn't happen. Or just exclude any changes of main.js from the commit and it should be easy to merge.

> When squashing a commit, that has been updated from the master,

What do you mean here?

boldtrn:
> You should rebase it after that change so that this wouldn't happen. Or just exclude any changes of main.js from the commit and it should be easy to merge.

The issue is that if you want to update the changes from the master (fetch and merge), need to resolve conflicts and then squash, git does not understand that this was a merge conflict resolution. Or at least I don't know how to do it better, as I think there should be a way :).

PR #1237, opened 2017-12-01
initial version to open source the Isochrone API

karussell:


willcohen:
This code has been working quite well for us so far, thanks so much! Do all five of these steps on the to-do need to happen before this can be merged into the project? How much information do you want documented? It seems like [ReachServlet](https://github.com/graphhopper/graphhopper/blob/7c127c25fc8bdc64277353198263ff9f418fc6f3/web/src/main/java/com/graphhopper/http/ReachServlet.java) is an excellent contained example of how to use and generate a full isochrone.

karussell:
> Do all five of these steps on the to-do need to happen before this can be merged into the project? 

Not finally decided

> How much information do you want documented?

Also not sure yet. But I want some more javadocs, [docs in Java](https://github.com/graphhopper/graphhopper/blob/master/docs/core/routing.md) and for the web API.

karussell:
Have created #1333 with work from @michaz.

PR #1240, opened 2017-12-09
Fix PathDetails exception for empty EdgeIdDs list

boldtrn:
If you request a route with two points, with start and end being the same snapped_point, the resulting path has only one point and no edges.

This resulted in an exception:
```
java.lang.IllegalStateException: PathDetails time must not be empty
at com.graphhopper.PathWrapper.addPathDetails(PathWrapper.java:258)
```

This PR proposes a possible solution for this by allowing to set empty PathDetails in the PathWrapper.

Due to this change the resulting PathDetails will be empty:
```
details: {
average_speed: [ ]
},
```

### DIFF
--- a/core/src/main/java/com/graphhopper/util/PathSimplification.java
+++ b/core/src/main/java/com/graphhopper/util/PathSimplification.java
@@ -1,14 +1,14 @@
 /*
  *  Licensed to GraphHopper GmbH under one or more contributor
- *  license agreements. See the NOTICE file distributed with this work for
+ *  license agreements. See the NOTICE file distributed with this work for 
  *  additional information regarding copyright ownership.
- *
- *  GraphHopper GmbH licenses this file to you under the Apache License,
- *  Version 2.0 (the "License"); you may not use this file except in
+ * 
+ *  GraphHopper GmbH licenses this file to you under the Apache License, 
+ *  Version 2.0 (the "License"); you may not use this file except in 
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -58,7 +59,7 @@ public PathSimplification(PathWrapper pathWrapper, DouglasPeucker douglasPeucker
     }
 
     public PointList simplify() {
-        if (listsToSimplify.isEmpty() || pointList.isEmpty())
+        if (listsToSimplify.isEmpty() || pointList.isEmpty() || pointList.size() <= 2)
```
karussell: pointList.isEmpty is not necessary anymore

boldtrn:
> Is there a fundamental difference to a via-point with zero length and just a single point?

Yes, for via points we know the value of the edge and if we merge the PathDetails this case is automatically resolved. For empty paths we don't know the value of the edge and cannot create a PathDetail like `[0,0,value]`.

karussell:
> For empty paths we don't know the value of the edge and cannot create a PathDetail like [0,0,value].

I thought about how it should be and IF both points are on an edge this is solvable. But when I'm think about this deeper, the problem is if both points are exactly on a tower node so that e.g. street_name value is not decidable, so an empty path details is the only solution.

PR #1244, opened 2017-12-13
Add a force flag to download a file without any user input

legraina:
I have added a -f flag in the graphhopper.sh script to be able to run it correctly in a docker. The goal is to remove any input of the user, especially when trying to download a new file.
We can now run `./graphhopper.sh -f import europe_germany_berlin.pbf` without having:
```bash
File not found 'europe_germany_berlin.pbf'. Press ENTER to get it from: http://download.geofabrik.de/europe/germany/berlin-latest.osm.pbf
Press CTRL+C if you do not have enough disc space or you don't want to download several MB.
```

### DIFF
--- a/graphhopper.sh
+++ b/graphhopper.sh
@@ -1,6 +1,52 @@
 #!/bin/bash
-(set -o igncr) 2>/dev/null && set -o igncr; # this comment is required for handling Windows cr/lf 
+(set -o igncr) 2>/dev/null && set -o igncr; # this comment is required for handling Windows cr/lf
 # See StackOverflow answer http://stackoverflow.com/a/14607651
+
+function printBashUsage {
+  echo "Usage:"
+  echo "-h | --help: display this message"
+  echo "-f | --force-download: force the download of the data file if needed"
+  echo "-d | --datareader.file: path to the map file or name of the file to download"
+  echo "-a | --action: must be one the following actions:"
+  echo "     --action import      creates the graphhopper files used for later (faster) starts"
+  echo "     --action web         starts a local server for user access at localhost:8989 and API access at localhost:8989/route"
+  echo "     --action build       creates the graphhopper JAR (without the web module)"
+  echo "     --action buildweb    creates the graphhopper JAR (with the web module)"
+  echo "     --action clean       removes all JARs, necessary if you need to use the latest source (e.g. after switching the branch etc)"
+  echo "     --action measurement does performance analysis of the current source version via random routes (Measurement class)"
+  echo "     --action torture     can be used to test real world routes via feeding graphhopper logs into a GraphHopper system (Torture class)"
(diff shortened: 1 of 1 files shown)


karussell:
Thanks!

I regret a bit that we have such a limiting (but very simple) command line format and the force flag would be useful but does not really fit into the current schema. Also we plan to use dropwizard via #1108 that will require yet another slightly different parameter format, so I would postpone this for a bit.

legraina:
Ok, so no need to try to fix the current script and be sure that all checks pass ? If you like to use something like that for the moment, I can make a version that is compatible with the previous versions and where you can add new flags. As you like.

karussell:
Hmmh, I'm a bit undecided here as I've not looked yet very deep in the dropwizard change and I do not want to waste anyones time :)

My initial thought was to use the same parameter style for the script like we use for direct java call but this is probably not possible nor standard with bash parameter style.

karussell:
The best would be something in the style of the java program like
```
./graphhopper.sh  --action web --datareader.file some.pbf
```

But maybe having this as a start with just `-f` is already good. Hmmh, but "force" could mean anything, maybe we should make the parameter more descriptive at the beginning and introduce short later on?

legraina:
Yes, I can totally do that for the moment (it would be pretty easy). I would also keep the previous settings to remain backwards compatible.

legraina:
I've added all these flags and also kept the positional arguments:
-h|--help
-f|--force-download
-a|--action
-d|--datareader.file
Does it seem good for you ?

karussell:
Does not look too bad :) thanks!

>  -f|--force-download

can we do -fd here and for datareader.file the -f?

karussell:
(Shouldn't sound negative regarding your work, sorry. Just still a bit hesitating to introduce yet another parameter format ...)

karussell:
But maybe we can also iterate here after we merge the dropwizardifying

legraina:
I've done the modifications. And no worry if you don't want :) I was doing these modifications to be able to use more easily the docker image (you can like that download and import any pbf file directly from the docker image, otherwise it's more tricky).

karussell:
Thanks again! (Give me a few days over the holidays and I can better judge :))

karussell:
I'm currently playing with [fish](https://fishshell.com/) and there is an interesting auto complete feature where the parameters and its description are shown. This would be cool to support - do you know if this is easy :) ?

legraina:
From what I've seen, you cannot do that. To be able to have a nice completion in bash you need to load first the possible completion (https://debian-administration.org/article/317/An_introduction_to_bash_completion_part_2), and then you have access to an auto-completion. It means that you will have to 'install' graphhopper.

karussell:
As we change the command line handling due to #1314 we have to rewrite the script a bit. The work you did here seems really useful and I'll try to see how we can use it to make a cleaner scripting possible.

karussell:
Please have a look at #1319. Will close here for now.

PR #1264, opened 2018-01-09
Force min speed value for short ferries

### LINKED ISSUE #1256: Too small duration for ferry
[This ferry](http://www.openstreetmap.org/directions?engine=graphhopper_car&route=52.0163%2C4.6358%3B51.9010%2C4.6856) should not be used and probably instead [this alternative](https://graphhopper.com/maps/?point=52.0163%2C4.6358&point=51.985488%2C4.752359&point=51.901%2C4.6856). See [osm way](http://www.openstreetmap.org/way/7301827)

boldtrn:
Fixes #1256.

For cars, this feature makes only sense if a speed factor < 5 is used, so I am not sure if it makes sense? See my comment in #1256.


karussell:
Thanks!

> For cars, this feature makes only sense if a speed factor < 5 is used, so I am not sure if it makes sense? See my comment in #1256.

If customized this makes sense for non-customized versions we have to live with this I fear

PR #1267, opened 2018-01-15
Add geocoding to the hc client

### LINKED ISSUE #1221: Add Geocoding to HC Client
Currently, if one wants to use the GraphHopper routing and geocoding API, a user either has to only the swagger client (no points encoded etc.) or has to mix the hc client with the swagger client. I think it would be a nice addition to the hc client to be able to use geocoding as well.

boldtrn:
This PR fixes #1221.

I reused the model classes that we created in the geocoding converter to easily convert json received via the API to Java objects. Probably the effort of hosting them in an independent repository and add a dependency from gh and the geocoding converter is not worth it. Also I changed one or two small details like the name of the classes and removed a constructor that used AbstractAddress, which we don't need here.


client-hc/src/main/java/com/graphhopper/api/GraphHopperGeocoding.java
```diff
@@ -0,0 +1,133 @@
...
+
+        if (request.isReverse()) {
+            url += "reverse=true";
+            url += "&point=" + request.getLat() + "," + request.getLon();
+        } else {
+            url += "reverse=false";
+            url += "&q=" + request.getQuery();
+            if (!Double.isNaN(request.getLat()))
+                url += "&point=" + request.getLat() + "," + request.getLon();
+        }
```
karussell: I would help here the user a bit: if reverse than throw an exception if the lat or lon is invalid. And if non-reverse throw exc if query is empty. Also adding the point can be moved outside of the if+else IMO to have just one line

client-hc/src/main/java/com/graphhopper/api/model/GHGeocodingRequest.java
```diff
@@ -0,0 +1,99 @@
...
+        return reverse;
+    }
+
+    public double getLat() {
+        return lat;
+    }
+
+    public double getLon() {
+        return lon;
+    }
```
karussell: Not sure if important: we could reuse the GHPoint class here too

boldtrn: Ok, sure, I can do this. Not sure if it's a big benefit. Should I change it?

client-hc/src/test/java/com/graphhopper/api/GraphHopperGeocodingIT.java
```diff
@@ -0,0 +1,61 @@
...
+import java.net.SocketTimeoutException;
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Robin Boldt
+ */
+public class GraphHopperGeocodingIT {
+
+    public static final String KEY = "614b8305-b4db-48c9-bf4a-40de90919939";
```
karussell: We should avoid duplicating this info and reuse GraphHopperWebIT.KEY

client-hc/src/main/java/com/graphhopper/api/model/GHGeocodingRequest.java
karussell: Is it necessary for serialization?

karussell:
Also this name would need a change https://github.com/graphhopper/graphhopper/blob/master/client-hc/pom.xml#L28

client-hc/src/test/java/com/graphhopper/api/GraphHopperGeocodingIT.java
boldtrn: Thanks, I missed that :+1:

client-hc/src/main/java/com/graphhopper/api/model/GHGeocodingRequest.java
boldtrn: No, we only serialize the response from the server.

client-hc/src/main/java/com/graphhopper/api/GraphHopperGeocoding.java
boldtrn: Thanks, good idea!

client-hc/src/main/java/com/graphhopper/api/model/GHGeocodingRequest.java
boldtrn: I haven't changed that so far, let me know if I should

boldtrn:
I updated this PR also adding some more documentation to make it easier for users to use methods that are publicly available.

client-hc/src/main/java/com/graphhopper/api/model/GHGeocodingRequest.java
karussell: Ok, yes, please change it to use GHPoint

karussell: This way we can just check against null instead of the NaN checks

karussell: Or just point.isValid

boldtrn: Good idea, I updated the code :)

karussell:
Thanks!

PR #1268, opened 2018-01-15
Load only last versions for OSM ways

ChernetsovNG:
Hello. I made the changes so that when loading data from OSM the graph was built only for ways of the latest versions.
I have the special case when in a .pbf file saved the history of modifications OSM data. In result, we have many versions for most lines. When building a graph, there are many lines in it that correspond to one osm ID (in my case maximum 37 versions of one line). To solve this problem I made this changes. 
Please, consider them for inclusion in the GraphHopper.


karussell:
As I can read you now parse a third time just because to determine the version? We can't do this as this will slow down the import process for a feature not important for almost all users.

Also I do not understand why this is necessary at all - why not preprocess the file on your own and feed the result to GraphHopper?

karussell:
Related discussion: https://discuss.graphhopper.com/t/2438

ChernetsovNG:
Unfortunately, I couldn't find how to preprocess osm or pbf file to leave only the latest versions of objects in it. "Standard" programs like osmosis can't do this. I can try to add this functionality in osmosis, but there need to learn how to do it...

Maybe, we can add a flag to the GraphHopper class, indicating that we need to process OSM file to delete the old versions, i.e. read it a third time (and by default, not do this)?

karussell:
Can we jump to the discussion website and would you mind to let me know what one use case behind this would be? Why not downloading stuff from geofabrik or extract from planet.pbf?

karussell:
The use case can be satisfied otherwise. Discussion: https://discuss.graphhopper.com/t/2438

PR #1269, opened 2018-01-16
Lane info (2. version) based on encoder refactoring

aramzl:


karussell:
Thanks! In #1232 you had some tests, is the usage like this still valid?

aramzl:
@karussell hi, yes the tests and usage is the same, just the implementation is different.

PR #1272, opened 2018-01-19
Avoid access=destination for bike and foot as well

### LINKED ISSUE #733: Add support for restriction value 'destination' for car

boldtrn:
This PR should fix #733. It adds support for access=destination tags.

I unfortunately needed to not only change the priority but also the speed. Some cases might have been solved by the priority only, but all 3 cases by @ftrebien, reported [here](https://github.com/graphhopper/graphhopper/issues/733#issuecomment-357673369), were not properly solved by only changing the Priority. 

If you look at the examples, the generated detour is sometimes huge, e.g. 500m instead of 80m.

BTW: The changed test in GraphHopperIT is actually a nice improvement. It misses an unnecessary continue instruction. I debugged it and it seems like there was a small road with access=destination (it's not there in the current OSM data anymore).  The small road with access=destination had the same name like the current road and the same flags. Now that we change the flags for roads with access=destination, the instruction get's ignored [here](https://github.com/graphhopper/graphhopper/blob/d1d60828ce4493c799990b5a817272d101406af4/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java#L369).


core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
```diff
@@ -331,6 +333,10 @@ protected int handlePriority(ReaderWay way, int priorityFromRelation) {
         else
             weightToPrioMap.put(110d, priorityFromRelation);
 
+        if (isWayAccessDestination(way))
```
boldtrn: We could merge this and the above `isWayAccessDestination`. But then we would change the Priority outside of the `handlePriority` method, which I don't like that much. I can change it if you would prefer that?

karussell:
> If you look at the examples, the generated detour is sometimes huge, e.g. 500m instead of 80m.

I'm not sure if anybody sane would do such a detour with a bike, if the area is just "destination only". Ugly. Maybe we should handle them just as "pushing sections"?

boldtrn:
> I'm not sure if anybody sane would do such a detour with a bike, if the area is just "destination only".

In theory "destination only" should be handled by a routing engine as exactly that. Only route over a road with destination only, if this is the only road that leads to the destination, otherwise use a different road. There might be examples where driving through an "destination only" area could result in a fine or in inconvenience for users following the route.

But, I looked at the tags in overpass and saw that there are many incorrectly tagged roads in Germany. Instead of motor_vehicle=destination, they contain just access=destination. This is often true for roads that might be very well suited for cycling.

Now I am not sure anymore if we should enforce the destination tag, as pushing section, avoid it, or don't do anything (current behavior). If we start avoiding it or use it as a pushing section, this could raise awarness of this problem in OSM and maybe slowly lead to improved data quality. Maybe your proposal to use it as pushing section might be a good solution, maybe enforcing an instruction, "get off the bike, access destination".

If you look at the examples in #733 on streetview: [example1 (should probably be tagged with access=customer)](https://goo.gl/maps/7yggvgTARSq) [example2 (should maybe be tagged as private?)](https://goo.gl/maps/cYsYqvpDexK2), [example3 (should maybe be tagged access=private)](https://goo.gl/maps/9PitfcfnGNJ2); I would not want to drive through any of them with a bicycle. But I am also not sure if they are tagged correctly. @ftrebien, are you sure that the examples are tagged correctly or could they be tagged differently, like access=private|customer?

karussell:
> This is often true for roads that might be very well suited for cycling.

Yes, exactly. Similar to the "default"-problem of barriers and fords

ftrebien:
@boldtrn In my examples, the restrictions really apply to all modes of transportation. I'm not sure if access=* has been used with this strict meaning in Germany, but that's what's been written in the English wiki for many years. It seems that the German wiki recently was [edited](https://wiki.openstreetmap.org/w/index.php?title=DE%3AKey%3Aaccess&type=revision&diff=1481427&oldid=1467258) to suggest using motor_vehicle=destination, copying from a previous [edit](https://wiki.openstreetmap.org/w/index.php?title=Key%3Aaccess&type=revision&diff=1216057&oldid=1208601) in the English wiki on 2015 (in fact this was first [written](https://wiki.openstreetmap.org/w/index.php?title=DE%3AKey%3Aaccess&type=revision&diff=629090&oldid=626407) in the German wiki, 4 months earlier). So there may have been some confusion between the two tags. Still, if the router implements the correct semantic, that will encourage mappers to fix the data.

boldtrn:
> Examples 2 and 1 are indeed bicycle/foot/motor_vehicle=destination because everyone can access the area if they are going there for the service (shopping at the mall or using government services).

If you go there for a service, this means you are a customer :).

> access=customers was born in combination with amenities, so its usage with the highway tag is still somewhat lower than access=destination.

Yes but both examples are driveways or parking of an amenity so I think this is correct?

>  It is quite possible that some (let's hope not most) of ways were incorrectly tagged before those dates and haven't been fixed since.

Unfortunately not. There are many tagging issues in Germany. This could have a massive negative impact on bicycle routing. 

I think it might be a good idea to convert access=destination to pushing sections, this could improve the tagging in OSM. At first this would lead to problems though, until most cycleways are fixed.

Also I would be ok to apply the the speed reduction + avoidance for ways that are tagged with `bicycle=destination`. There are more ways than I would have expected with this tagging scheme, so it would improve the routing. (same for `foot=destination`).

ftrebien:
@boldtm You are right, but that still makes such data [incorrect](https://forum.openstreetmap.org/viewtopic.php?id=57874) in Germany. Recent discussions on the subject, for example [this discussion on private streets](https://forum.openstreetmap.org/viewtopic.php?id=59723), assume that those tags are being used correctly. So, well, it depends on whether GraphHopper hopes to help encourage mappers to improve map data or simply continue to use the tag values incorrectly (as [most routers](https://forum.openstreetmap.org/viewtopic.php?pid=659576#p659576) do too). Perhaps another approach for this standstill would be to slowly decrease the speed on access/bicycle=destination so that small detours are not affected initially.

ftrebien:
@boldtrn On whether _destination_ or _customers_ would be more correct, there is [disagreement](https://wiki.openstreetmap.org/wiki/Talk:Proposed_features/customers) on the need for the new _customers_ value.

boldtrn:
> You are right, but that still makes such data incorrect.

Yes using access=destination where motor_vehicle=destination is signed, is tagged incorrect. 

> So, well, it depends on whether GraphHopper hopes to help encourage mappers to improve map data or simply continue to use the tag values incorrectly

Yes, I should have explained, "convert access=destination to pushing sections" means slowing down these streets. GraphHopper then assumes that you have to push your bike on these ways, which is slower than riding your bike, and also results in an instruction "get off your bike".

core/src/main/java/com/graphhopper/routing/util/FootFlagEncoder.java
karussell: This looks good to me.

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -690,6 +690,22 @@ protected double applyMaxSpeed(ReaderWay way, double speed) {
         return speed;
     }
 
+    /**
+     * Checks if a road is access destricted. For example, if we have a case with
+     * access=destination and [encoder_vehicle]=yes, this method returns false
```
karussell: Is this really true? Would it be possible to add destination to the intendedValues?

But let's clarify if this issue is really worth the effort. Using pushing sections for bike in these cases is ok, but for foot "destination" does not make much sense. And then only bike should be changed.

boldtrn: Mhm, not sure if destination is a intended value? In general we want to avoid roads with access=destination?

I think the example was about a road that has for example access=destination and bicycle=yes, so bicycle would be the more specific rule and therefore more important. So we would see this road as a regular road for bicycles, but as access restricted for cars.

> but for foot "destination" does not make much sense

You mean because of incorrectly tagged roads? I think there are not many roads that actually use foot=destination.

karussell:
I'm not sure if this makes sense for foot. Is there a sign explicitly for foot? If not let's do this just for bike.

boldtrn:
> I'm not sure if this makes sense for foot. Is there a sign explicitly for foot? If not let's do this just for bike.

I also think that the relevance for foot is not that high. In theory there could be situations where foot=destination. I checked a sample on overpass and [compared it with mapillary](https://www.mapillary.com/map/im/KB1_u3OA-sypYXRr5gJ8UQ). I don't see why there would be foot=destination on this footpath.

karussell:
I think we should not merge this. See #1831.

boldtrn:
Yes, I agree with you. I think there are too many tagging issues here. Should we close here?

PR #1276, opened 2018-01-23
Properly do turn costs update

### LINKED ISSUE #1107: calling "addTurnInfo()" twice won't update turn costs
When using the function [addTurnInfo(int fromEdge, int viaNode, int toEdge, long turnFlags)](https://github.com/graphhopper/graphhopper/blob/0a2ae97c087d8152b78430885fd52cc427b5b6c7/core/src/main/java/com/graphhopper/storage/TurnCostExtension.java#L113)  a second time on the same edges, it doesn't update the turn costs.
When trying to receive them by calling [getTurnCostFlags(int edgeFrom, int nodeVia, int edgeTo)](https://github.com/graphhopper/graphhopper/blob/0a2ae97c087d8152b78430885fd52cc427b5b6c7/core/src/main/java/com/graphhopper/storage/TurnCostExtension.java#L153) you'll get the first cost entry - not the 'updated' one.

Add the following test to EdgeBasedRoutingAlgorithmTest shows that effect.

```
@Test
public void testTurnCostsUpdateBug() {
    final GraphHopperStorage g = createStorage(createEncodingManager(false));
    initGraph(g);
    TurnCostExtension tcs = (TurnCostExtension) g.getExtension();
    long tflags = carEncoder.getTurnFlags(false, 1);
    tcs.addTurnInfo(getEdge(g, 5, 2).getEdge(), 2, getEdge(g, 2, 3).getEdge(), tflags);

    tflags = carEncoder.getTurnFlags(false, 2);
    tcs.addTurnInfo(getEdge(g, 5, 2).getEdge(), 2, getEdge(g, 2, 3).getEdge(), tflags);

    assertEquals(tflags, tcs.getTurnCostFlags(getEdge(g, 5, 2).getEdge(), 2, getEdge(g, 2, 3).getEdge()));
}
```

i know this problem could be solved by simply using the correct turnCosts in the first place, but in some cases this might be the easier way.

This hack is fixing the problem by iterating all turnCostEntries and using the maximum.
```
     private long nextCostFlags(int edgeFrom, int nodeVia, int edgeTo) {
         int turnCostIndex = nodeAccess.getAdditionalNodeField(nodeVia);
+        int cost = -1;
         int i = 0;
         for (; i < 1000; i++) {
             if (turnCostIndex == NO_TURN_ENTRY)
@@ -168,7 +168,8 @@ public class TurnCostExtension implements GraphExtension {
             long turnCostPtr = (long) turnCostIndex * turnCostsEntryBytes;
             if (edgeFrom == turnCosts.getInt(turnCostPtr + TC_FROM)) {
                 if (edgeTo == turnCosts.getInt(turnCostPtr + TC_TO))
-                    return turnCosts.getInt(turnCostPtr + TC_FLAGS);
+                    if (cost < turnCosts.getInt(turnCostPtr + TC_FLAGS))
+                       cost = turnCosts.getInt(turnCostPtr + TC_FLAGS);
             }
 
             int nextTurnCostIndex = turnCosts.getInt(turnCostPtr + TC_NEXT);
@@ -177,6 +178,9 @@ public class TurnCostExtension implements GraphExtension {
 
             turnCostIndex = nextTurnCostIndex;
         }
+               if (cost >= 0)
+                       return cost;
+        
         // so many turn restrictions on one node? here is something wrong
         if (i >= 1000)
             throw new IllegalStateException("something went wrong: there seems to be no end of the turn cost-list!?");
        return EMPTY_FLAGS;
    }
```

I guess the clean solution would be an update mechanism when adding the turnCosts.

Nakaner:
Add a new method to TurnCostExtension which enables users to define their preferred merge strategy (overwrite vs. bitwise OR vs. …). These changes require JDK 8.

#1107

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/TurnCostExtension.java
+++ b/core/src/main/java/com/graphhopper/storage/TurnCostExtension.java
@@ -17,6 +17,8 @@
  */
 package com.graphhopper.storage;
 
+import java.util.function.BiFunction;
+
 import com.graphhopper.util.EdgeIterator;
 
 /**
@@ -109,16 +111,45 @@ public class TurnCostExtension implements GraphExtension {
     /**
      * This method adds a new entry which is a turn restriction or cost information via the
      * turnFlags.
+     *
+     * It appends a new entry and does not overwrite existing entries of the same turn.
      */
     public void addTurnInfo(int fromEdge, int viaNode, int toEdge, long turnFlags) {
         // no need to store turn information
-        if (turnFlags == EMPTY_FLAGS)
+        if (turnFlags == EMPTY_FLAGS) {
             return;
+        }
+        BiFunction<Long, Long, Long> overwriteFunc = (oldFlags, newFlags) -> {
+            return newFlags;
+        };
+        setAndMergeTurnInfo(fromEdge, viaNode, toEdge, turnFlags, overwriteFunc);
+    }
 
+    /**
+     * Add a new turn cost entry.
+     * @param fromEdge edge ID
+     * @param viaNode node ID
+     * @param toEdge edge ID
+     * @param turnFlags flags to be written
+     * @param mergeFunction function which merges the old and the new flags as you prefer.
+     * This function takes two arguments – the old flags are the first, the new flags the
+     * second argument.
+     *
+     * See EdgeBasedRoutingAlgorithmTest for examples.
+     */
+    public void setAndMergeTurnInfo(int fromEdge, int viaNode, int toEdge, long turnFlags,
+            BiFunction<Long, Long, Long> mergeFunction) {
+        // no need to store turn information
+        if (turnFlags == EMPTY_FLAGS) {
+            return;
+        }
         // append
         int newEntryIndex = turnCostsCount;
-        turnCostsCount++;
         ensureTurnCostIndex(newEntryIndex);
+        boolean oldEntryFound = false;
+        long existingFlags = turnFlags;
+        long newFlags = turnFlags;
+        int next = NO_TURN_ENTRY;
 
         // determine if we already have an cost entry for this node
         int previousEntryIndex = nodeAccess.getAdditionalNodeField(viaNode);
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/graphhopper/storage/TurnCostExtension.java
```diff
@@ -17,6 +17,8 @@
  */
 package com.graphhopper.storage;
 
+import java.util.function.BiFunction;
```
devemux86: Note: on Android the `java.util.function` package is [available](https://developer.android.com/studio/write/java8-support.html) from API level 24 or higher.

karussell:
Thanks for the patch! Currently it is not planned to switch to JDK8 for core (otherwise harder for Android and iOS) - can you avoid this? Furthermore we need a [signed CLA](https://github.com/graphhopper/graphhopper/blob/master/.github/CONTRIBUTING.md#license-agreement) from new contributors, just [send us an email](https://www.graphhopper.com/imprint/).

karussell:
Also did you compare speed for import with and without this patch (e.g. for Germany)? I expect no changes, but if there are many turn costs associated (e.g. if we add costs for every junction) we might run into troubles and then a different approach is needed like a sorted storage where we could stop the loop earlier.

core/src/main/java/com/graphhopper/storage/TurnCostExtension.java
```diff
@@ -109,16 +111,45 @@ public boolean loadExisting() {
...
+     * @param toEdge edge ID
+     * @param turnFlags flags to be written
+     * @param mergeFunction function which merges the old and the new flags as you prefer.
+     * This function takes two arguments – the old flags are the first, the new flags the
+     * second argument.
+     *
+     * See EdgeBasedRoutingAlgorithmTest for examples.
+     */
+    public void setAndMergeTurnInfo(int fromEdge, int viaNode, int toEdge, long turnFlags,
+            BiFunction<Long, Long, Long> mergeFunction) {
```
boldtrn: Maybe instead of passing a function here we could convert this to an Enum that specifies the used mode like `MergeMode.OVERWRITE` and `MergeMode.OR`, that way we can add new modes in the future and still continue using JDK 7?

Nakaner:
I pushed two commits today which should address your comments.

Results from a performance test importing Germany with following settings:

```
graph.flag_encoders=car
graph.flag_encoders=car|turn_costs=true
prepare.ch.weightings=no
graph.dataaccess=RAM_STORE
```

The master needs 7 minutes and 6 seconds, my branch needs the same time (or a few seconds more but it seems to be within the range of variation of the measurements).

karussell:
Thanks a lot for adressing the commits - looks better now and almost mergable. 

One question, identical to the comment from @boldtrn: why can't we just use a boolean instead of the function as parameter? I.e. do the merge if `merge==true` and otherwise do an overwrite.

Nakaner:
@karussell wrote:
> One question, identical to the comment from @boldtrn: why can't we just use a boolean instead of the function as parameter? I.e. do the merge if merge==true and otherwise do an overwrite.

I wanted to keep it flexible. A bitwise OR is the right solution if you write the flags of one encoder after another. But it is not the right operation on the old and new flags if you just want to overwrite the flags of one encoder or do other – maybe strange – things.

karussell:
We should introduce an additional layer of abstraction only if it is really necessary. Here I do not see why a boolean is not sufficient - can you explain?

karussell:
Would you mind to do this tiny change so that we can include it in 0.10 release (hope to get some initial version out in the next days)

karussell:
Thanks for the change!

Why should keep `checkExistingEntries==false` as you said the import was similarly fast? If is is false (per default) it would not fix #1107 and if it is always true we can remove it from the method.

Nakaner:
@karussell wrote:
> Why should keep the option checkExistingEntries==false as you said the import was similarly fast?

It was a rough test and variation was within the measurement accuracy after two or three iterations.

> If this parameter is false (per default) it would not fix #1107 and if it is always true we can remove it from the method.

#1107 is caused by a lack of documentation. If a user of the library does not have a look on the actual implementation, he might think that `addTurnInfo` will look for existing entries. I tried to keep this behaviour to be fully backwards-compatible but added an alternative to ease the usage of the library. I myself can use Graphhopper 0.9 because I found a simple way to mitigate this problem (switch to for loops – the loop over the nodes is now the outer one and the inner one iterates over the encoders).

Do with the pull request what you want. Git gives you the ability to pick out of this branch what you are interested in. I offered you a patch with unit tests and documentation and I am tired to continue this discussion.

karussell:
> he might think that addTurnInfo will look for existing entries

We should look for existing entries. This is no documentation issue IMO, there should be no feature "add without looking for existing entries" - at least I do not see why.

> I tried to keep this behaviour to be fully backwards-compatible 

why?

> I offered you a patch with unit tests and documentation and I am tired to continue this discussion.

Oh, wow. I really appreciate your work and time but this is a bit disturbing to me. Open source lives from these discussion! Do you think I'm doing this for wasting your time?

karussell:
I will merge this: https://github.com/graphhopper/graphhopper/commit/bc91215a0b3314482b7dda310f703c84f20d6d36

Please @Nakaner let me know if you are ok with this. I also added the possibility to clear a certain turn information with the new method.

Nakaner:
I am ok with that.

karussell:
Thanks again for your contribution - merged now!

PR #1278, opened 2018-01-30
Added Extent and OsmKey to Geocoding

boldtrn:
As mentioned by @devemux86 in #1221 the geocoding of the client-hc misses the osm key. While inspecting the Geocoding result I also saw that the extent was also missing, so I added that as well.

BTW: The extent is not documented in the API and I think the format is a bit weird (similar to Photon though). Should we add documentation about this feature in the API doc?

### DIFF
--- a/client-hc/src/main/java/com/graphhopper/api/model/GHGeocodingEntry.java
+++ b/client-hc/src/main/java/com/graphhopper/api/model/GHGeocodingEntry.java
@@ -20,6 +20,7 @@ package com.graphhopper.api.model;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.graphhopper.util.shapes.BBox;
 
 /**
  * Contains the results of a geocoding request.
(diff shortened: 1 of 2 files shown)


boldtrn:
> Is the extent also available through the geocoder-converter?

Not yet :).

Nominatim returns a [`boundingbox`](https://nominatim.openstreetmap.org/search/bodensee?format=json&addressdetails=1&limit=1&polygon_svg=1) which seems similar to the extent.

```
boundingbox: [
"47.4780623",
"47.8175234",
"8.8593204",
"9.7538633",
],
``` 

OpenCageData returns bounds, which could be converted to the extent.

```
"bounds": {
        "northeast": {
          "lat": 48.1799218,
          "lng": 11.5961796
        },
        "southwest": {
          "lat": 48.1399218,
          "lng": 11.5561796
        }
      }
```

Should we open a ticket for the geocoding converter?

karussell:
> Should we open a ticket for the geocoding converter?

Yes

client-hc/src/main/java/com/graphhopper/api/model/GHGeocodingEntry.java
```diff
@@ -173,6 +189,32 @@ public void setOsmValue(String osmValue) {
...
+        return this.extent;
+    }
+
+    @JsonProperty
+    public void setExtent(Double[] extent) {
+        if (extent == null || extent.length == 0) {
+            return;
+        }
+        if (extent.length == 4) {
+            // Extend is in Left, Top, Right, Bottom; which is very uncommon, Photon uses the same
```
karussell: Has nothing to do with uncommon. There are many different orders - see bbox javadoc

boldtrn: I have just seen that this is the "two points" format, but in a GeoJson style, which switches lat, lon. This is indeed similar to how GraphHopper formats the BBox in Geojson. Should I remove the comment?

PR #1279, opened 2018-01-31
Added ascend, descend, weight, and description to the client-hc

### LINKED ISSUE #1233: ascent is not available when using GraphHopperWeb
There is something like

pathWrapper.setAscend(path.get("ascend").asDouble());

missing. And also for other properties like descend

boldtrn:
Fixes #1233.

I added ascend, descend, weight, and description to the client-hc.

Because description only appears for alternative_route, I had to write a test for alternative_route as well, with ch.disabled. Not sure if we want this? I tried to create a very short route < 1km.


karussell:
Thanks!

> Because description only appears for alternative_route

Ah, this can be indeed calculated for all cases (but if no alternatives it might be useless?) and it is very likely improvable :)

> Not sure if we want this?

Wanted :) !

PR #1281, opened 2018-02-08
Show instructions for turns onto oneway streets

### LINKED ISSUE #1048: Instructions: Missing instruction when turning onto road with separated one way lanes

boldtrn:
Fixes #1048.

The idea of this PR is that we show an instruction, even if the turn is the only viable turn, if there is another potential (but forbidden) turn and if the turn is not a "follow road/slight turn".

The idea is that we don't see instructions when entering motorways, because these turns are usually only slight turns, but that we see instructions when turning onto a separated lane oneway as described in #1048.

One situation would be [this](https://graphhopper.com/maps/?point=-33.824437%2C151.187925&point=-33.824415%2C151.188177), but there are plenty.

So far the cases I looked at, look good. So far it feels like a good improvement. But I didn't perform too many tests yet.


client-hc/src/test/java/com/graphhopper/api/GraphHopperWebIT.java
```diff
@@ -67,9 +67,9 @@ public void testSimpleRoute() {
 
     @Test
     public void testAlternativeRoute() {
-        // https://graphhopper.com/maps/?point=52.044124%2C10.378346&point=52.043847%2C10.381994&algorithm=alternative_route&ch.disable=true
+        // https://graphhopper.com/maps/?point=52.044098%2C10.378368&point=52.043847%2C10.381994&algorithm=alternative_route&ch.disable=true
```
boldtrn: Here the server did not produce alternative routes anymore, I had to move the waypoint a couple of meters.

karussell: Hmmh, strange it should still produce alternatives: https://graphhopper.com/maps/?point=52.044124%2C10.378346&point=52.043847%2C10.381994&algorithm=alternative_route&ch.disable=true

boldtrn: Mhm, maybe it was a one time issue? Should I revert the change?

boldtrn: I checked it on GH Maps and it didn't produce an alternative route.

core/src/main/java/com/graphhopper/routing/InstructionsSurroundingEdges.java
```diff
@@ -89,6 +89,15 @@ public int nrOfPossibleTurns() {
         return 1 + reachableEdges.size();
     }
 
+    /**
+     * Calculates the Number of surrounding edges, including the current turn.
+     * This includes oneways in the wrong direction that are no possible turn.
+     */
```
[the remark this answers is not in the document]
karussell: For me "turns" is not really defined and if I think about a definition it will be combination of start and destination edge and then possibleTurns is not correct and should be possibleEdges

client-hc/src/test/java/com/graphhopper/api/GraphHopperWebIT.java
karussell: locally it seems to fail - got the same problem in my branch: https://github.com/graphhopper/graphhopper/commits/json_refactoring

karussell: This is weird as it is okay, when I run it locally.

karussell: Yes, when rerunning the failing jobs now they become green, i.e. they failed temporary due to some data update.

boldtrn: Yes I had the same impression :). Should I revert this change as it doesn't really belong into this PR?

core/src/main/java/com/graphhopper/routing/InstructionsSurroundingEdges.java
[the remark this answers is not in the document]
boldtrn: You are right, it makes reading the code of the instruction generation a bit better understandable I think, but it's not really defined. How about outgoingEdges?

Now we need a way to differentiate the two cases like allowed/passable/traversableOutgoingEdges and allOutgoingEdges? I struggle a bit with possibleEdges, I don't think it describes the situation well enough to understand it without reading the documentation (the same was true for surrounding edges ;)).

client-hc/src/test/java/com/graphhopper/api/GraphHopperWebIT.java
karussell: ok

core/src/main/java/com/graphhopper/routing/InstructionsSurroundingEdges.java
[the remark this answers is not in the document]
karussell: all and allowedOutgoingEdges sounds good. And yes surrounding edges is also unclear. But naming is really hard :)

karussell:
Thanks!

Issue #1007, opened 2017-03-25
Merge/Contract close instructions


boldtrn:
The instruction generation does not know much about the context of the route, it knows the last and the next edge. Also it considers surrounding edges, but not the whole path.

When knowing the whole path we can check instructions that are close to each other if one instruction might make the second instruction redundant (or vice versa).

For example roundabouts:
Some roundabouts have an extra lane, meant to be used for shortcutting the first exit of the roundabout. If we don't take the shortcut we oftimes create "KEEP_LEFT" followed by the roundabout instruction. http://localhost:8000/?point=48.680077%2C9.367062&point=48.679121%2C9.366644
The same might happen when leaving a roundabout. I think it would be obvious to everyone that when leaving the roundabout via the 2nd exit, the shortcut should not be taken.

Also see @karussell comment in #955:

> * if turns are <10m away should we consider merging them or how to properly guide a walking person? 
   * http://localhost:8989/?point=49.942248%2C11.545016&point=49.942438%2C11.543326&locale=de&vehicle=foot
   * http://localhost:8989/?point=49.936265%2C11.530414&point=49.9366%2C11.529974&locale=de&vehicle=foot

I think we have to collect a couple of use-cases where merging/contracting Instrucions makes sense, to decide what steps we should take.

devemux86:
Probably that handling is more appropriate in app level (where the requirements vary) than being forced by the library?
So an option to turn it on / off could be useful via e.g. request hint.

karussell:
Maybe @devemux86 is right. Also you @boldtrn mentioned that you would prefer this (in another issue)?

boldtrn:
> Probably that handling is more appropriate in app level (where the requirements vary) than being forced by the library?

> Also you @boldtrn mentioned that you would prefer this (in another issue)?

Yes, generally speaking I think this is still true. But there might be situations where we always want to contract/remove turn instructions.

So my idea was to have a correction loop after creating the instructions that uses the knowledge about the graph + the context of all instructions. In this loop we could for example remove the [unnecessary instruction at roundabouts.](http://localhost:8989/?point=48.680077%2C9.367062&point=48.679121%2C9.366644) It would be wrong to remove all slight_turns that are closer than 50m to a roundabout instruction (that's what I could do on app level), but it might be good to remove all of the slight turn instructions before a roundabout that do not include a street name change.

[...]

PR #1284, opened 2018-02-12
Merge consecutive continue instructions

### LINKED ISSUE #1282: Instructions: Merge multiple close Continue Instructions
Sometimes at more complicated intersections, there might be multiple continues in a very short succession (<50m apart). Creating multiple instructions does not really help the wayfinding, but is rather distracting IMHO.

Here is one [example](https://graphhopper.com/maps/?point=-33.788943%2C151.127254&point=-33.787342%2C151.125891&debug=true). Both continues on it's own are correct. But having them in this case, might not help.

I propose to merge two successive continues if they are close (e.g. < 50m) and if they use the same street name. I have been at this intersection multiple times with the [Kurviger app](https://play.google.com/store/apps/details?id=gr.talent.kurviger.pro) and the two continues make it hard to understand that one has to turn right shortly afterwards.

gh-continue
gh-continue-2

boldtrn:
This PR fixes #1282.

In addition to the sample showed in #1282, I ran the measurements for BW and print every occurence where we merged an instruction and found 5 occurrences (they each appeared multiple times), there are probably many more that are not used as frequently during the routing.

After some more tests I think that 40m (maybe even 35m) would be a good distance to merge instructions. At least for motor vehicles this is a very short distance and creating multiple continue instructions doesn't add much value IMHO.

Here are the examples from BW:
- http://localhost:8989/?point=49.140688%2C10.031462&point=49.139158%2C10.025861&debug=true
- http://localhost:8989/?point=48.595155%2C8.86808&point=48.595666%2C8.86682&debug=true
- http://localhost:8989/?point=49.134188%2C10.069313&point=49.134329%2C10.06382&debug=true
- http://localhost:8989/?point=49.459784%2C8.491632&point=49.461667%2C8.490028&debug=true
- http://localhost:8989/?point=49.005525%2C8.368433&point=49.004564%2C8.36633&debug=true

Interestingly, most don't seem to occur for GH maps. Probably because of the improved ETAs.

Furthermore I moved the `updateInstructionsWithContext` method to the InstructionList. I think it fits there better and is also better testable.


core/src/main/java/com/graphhopper/util/InstructionList.java
```diff
@@ -342,4 +342,86 @@ public Instruction find(double lat, double lon, double maxDistance) {
...
+        Instruction nextInstruction;
+
+        final int MERGE_CONTINUE_INSTRUCTIONS_DISTANCE = 40;
+
+        for (int i = 0; i < this.size() - 1; i++) {
+            instruction = this.get(i);
+            nextInstruction = this.get(i + 1);
+
+            if (i == 0 && !Double.isNaN(favoredHeading) && instruction.getExtraInfoJSON().containsKey("heading")) {
+                double heading = (double) instruction.getExtraInfoJSON().get("heading");
```
karussell: I do not like using something internally that we have implemented for the JSON export. Instead we should use normal Java variables with compile time checks etc or implement this in a different way.

Probably we should create a java variable for the (last_)heading and also think about creating a JSON serializer for the Instruction class to replace InstructionList.createJson

boldtrn: Actually this is not really part of this PR, this was added in #1073, where this parameter was added for this and for the JSON, [here](https://github.com/graphhopper/graphhopper/pull/1073/files#diff-e5132d18a6606ee07a176bff362f8ccfR129). IIRC, the reason why I implemented it this way instead of a regular property was that we only use the heading for the first continue instruction. We add the `last_heading` to all FinishInstructions. This allows us to compare the first and last heading.

We could introduce a new Instruction subclass for the initial Instruction and store the heading property there. We could store the last_heading in the FinishInstruction. We could also add the heading to every instruction, but I don't think that this would make much sense.

Should we move this to a separate issue?

karussell: Ok, I see that this is already used. Let us get this fixed first in another PR as this is too unclean and afterwards proceed here?

Regarding the different headings: why couldn't we always use `heading` why do we need `last_heading`? The finish instruction would then get the heading of the very last edge.

karussell:
Is this mostly fixed now due to #1844?

boldtrn:
I wouldn't say it is fixed, but there are less occurrences. I just rechecked the 5 example mentioned in my initial issue description, 2 of the 5 still exist.

https://graphhopper.com/maps/?point=49.134188%2C10.069313&point=49.134329%2C10.06382&debug=true
https://graphhopper.com/maps/?point=49.005525%2C8.368433&point=49.004564%2C8.36633&debug=true

I still think, we should merge some close instructions (also see #1007). IMHO, having multiple instructions, within 50m for motor vehicles should be avoided, where possible. Probably even instructions within 100m could be considered for merging. 

I think generally there are two cases:
- Cases where the client should merge or handle these instructions, like "turn left, then turn in 80m turn right". These would be actual individual turns.
- Cases where surrounding context around the instructions indicate that one instruction would be enough. I think a couple of close continue instructions are one of the best examples for this. But I think there are also other cases, for example two very close keep right instructions, could indicate, that a single keep right might be enough.

boldtrn:
Since this PR is quite old and rather outdated, I would be fine with closing it but keeping the initial issue and maybe add some of the discussions of this PR into the initial issue, so we can look into this at some other time?

karussell:
Ok, yes. Could you close and link from the initial issue to this discussion here?

PR #1286, opened 2018-02-14
Improve release management

### LINKED ISSUE #1285: Improved release cycle
Our aim should be more frequent releases. The best thing would be if we could tag certain commits and everything else is done from e.g. travis. A new release is then also automatically created on github.

This is possible with this procedure (completely without the release plugin) stolen from [here](https://axelfontaine.com/blog/dead-burried.html):

 * use `${revision}` in version field for all pom.xml files and use 
`<revision>0.10-SNAPSHOT</revision>` in properties as default
 * make it possible for travis to deploy to maven central if a commit is tagged. If we make gpg working this should be easy, see e.g. [here](https://dracoblue.net/dev/uploading-snapshots-and-releases-to-maven-central-with-travis/).
 * specify `-Drevision=0.9.$TRAVIS_BUILD_NUMBER` when deploying to maven central to overwrite the snapshot. Or we could use the git tag of the commit: `-Drevision=$TRAVIS_TAG`!
 
It was a bit tricky but it [seems to work](https://travis-ci.org/graphhopper/graphhopper/jobs/341437512).

karussell:
This PR automatically creates maven releases if a commit is tagged and fixes #1285 :) ! A first test was already *automatically* released as 0.10.alphaX on maven central.

It also changes a bit our travis handling uses the default way to support jdk9 and removes our old stuff.

Tasks:

 - [x] deploy Android demo too. I'm unsure if we can mix android and java on travis. Not sure if this is a blocker
 - [x] use `-Drevision=$TRAVIS_TAG`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,9 +42,10 @@ after_success:
   # deploy snapshot artifacts to sonatype and if tagged deploy the release to maven central
   - if [ "$TRAVIS_JDK_VERSION" == "oraclejdk8" ] &&
        [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
-        mvn deploy --settings core/files/settings.xml -DskipTests=true -B;
+        travis_wait mvn deploy --settings core/files/settings.xml -DskipTests=true -B;
     elif [ "$TRAVIS_JDK_VERSION" == "oraclejdk8" ]; then
-        mvn -P release --settings core/files/settings.xml -DskipTests=true -Drevision=0.10.alpha$TRAVIS_BUILD_NUMBER -B deploy;
+        echo "release to maven central";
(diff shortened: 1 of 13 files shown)


boldtrn:
> if a commit is tagged

What do you mean with is tagged and how would the tagging work?

If I understand the PR correctly we will deploy a new version for every commit into master? We create the version on Travis with `Drevision=0.10.alpha$TRAVIS_BUILD_NUMBER`?

karussell:
> What do you mean with is tagged and how would the tagging work?

You would go to any git commit and do `git tag <version>` and then we'll get this as release on maven central.

> If I understand the PR correctly we will deploy a new version for every commit into master?

We only deploy snapshots for every commit as usual.

karussell:
It looks like mixing Android and Java is possible but jdk9 won't work at the moment https://github.com/travis-ci/travis-ci/issues/8513

PR #1291, opened 2018-02-20
Add JSON dependency to core

karussell:
As discussed [here](https://discuss.graphhopper.com/t/json-jackson-dependency-in-gh-core/2807) introducing the JSON dependency in core could make several things less clunky, especially since it seems we need this in #1112 in the core as we use the json format instead of the hacky binary+text mix of StorableProperties. (It will not happen for 0.10)


karussell:
If JTS is able to [read and write geojson on its own](https://github.com/graphhopper/graphhopper/issues/1297) we probably don't need to add jackson to the core.

willcohen:
I should note that JTS also broke itself into a few modules, so [jts-core](https://search.maven.org/#artifactdetails%7Corg.locationtech.jts%7Cjts-core%7C1.15.0%7Cbundle) is the main replacement, and then there's also [jts-io-common](https://search.maven.org/#artifactdetails%7Corg.locationtech.jts.io%7Cjts-io-common%7C1.15.0%7Cjar).

karussell:
Thanks for highlighting this! I think we will then better stay with jackson instead of using the inactive json-simple. (https://github.com/conveyal/gtfs-lib/issues/97#issuecomment-370471610)

karussell:
Really hard decision, but I think for now we close this and try to keep JSON out of the core. Still unsure.

PR #1294, opened 2018-02-21
simplify also when instructions disabled

### LINKED ISSUE #1293: With instructions=false no simplification is applied
If someone wants just the geometry it is ok to say `instructions=false` but then the geometry simplification can be applied. In fact without instructions (and no path details) it should be much more efficient to simplify. But [we do nothing](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/util/PathSimplification.java#L47) if instructions are disabled.

karussell:
fixes #1293


karussell:
Would you mind to have a look into this @boldtrn :D ?

PR #1298, opened 2018-02-22
Bringing isochrone_open_source branch up-to-date

willcohen:
Re #1237, this:

1. Brings the PR branch up-to-date with current master and its changes to versioning since Dec
2. Checks off the first of the five todos, renaming the 'reach' packages to 'isochrone'


karussell:
Thanks, LGTM. Do you think we should rename `Reachability` into `Isochrone(Algorithm)`?

karussell:
Any preferences here @willcohen?

willcohen:
I think the renames make sense. Apologies for the delay -- as I was giving the refactor a shot I also noticed that the ContourBuilder says it's an adaptation of org.opentripplanner.common.geometry.DelaunayIsolineBuilder, which is LGPL, which means that even if JTS and the other pieces get their license issues resolved, the broader issue will still remain once isochrones get merged in. Does it make sense to try to base this feature off of a different algorithm instead? [Sextante](http://ojwiki.soldin.de/index.php?title=Sextante) is a little old but has a bunch of MIT-licensed GIS algorithms, including some [Isolines from TIN](https://joinup.ec.europa.eu/svn/sextante/soft/sextante_lib/trunk/algorithms/src/es/unex/sextante/tin/linearIsolinesFromTin/). The canonical home for Sextante is [here](http://code.google.com/p/sextante), but I just haven't gotten through adapting this to JTS's object types yet.

karussell:
LGPL is not a huge problem for now if we document which parts are which license in the NOTICE file. We can solve this later step by step. E.g. in 0.8 we were free of LGPL and we might be free of LGPL in 0.11 again, but this is not a big problem on the server side and on mobile only Android is affected due to the "merging".

> Does it make sense to try to base this feature off of a different algorithm instead?

This would be really cool, also to see other advantages (and also disadvantages). Still I would prefer this in a separate PR to avoid further delay.

willcohen:
Okay -- I think I got the rest of the reach -> isochrone renames with this latest commit

karussell:
Thanks!

PR #1300, opened 2018-02-26
Fix download issue if SRTM files misses '.' before 'hgt'

### LINKED ISSUE #1274: SRTM import is broken
Currently we are unable to build graphs with height profiles for the whole world and for North America.

The srtm height profile file names at http://srtm.kurviger.de/SRTM3/North_America/index.html sometimes do have the format:
N56W071hgt.zip
and not
N56W071.hgt.zip
The last format is assumed in com.graphhopper.reader.dem.SRTMProvider. 

This line fails:
downloader.downloadFile(zippedURL, file.getAbsolutePath());
with zippedURL being: 
http://srtm.kurviger.de/SRTM3//North_America/N56W071.hgt.zip
which should be:
http://srtm.kurviger.de/SRTM3/North_America/N56W071hgt.zip

I have tried to update core/src/main/resources/com/graphhopper/reader/dem/North_America_names.txt with the correct names but this does not solve the problem. The filenames are parsed at the init routine of the SRTMProvider and converted to coordinates. For downloading the coordinates are converted to an url.

It might be a temporary problem at srtm.kurviger.de.

boldtrn:
Fixes #1274. When we see a FileNotFoundException, we now try to remove the '.' and try to download the file again.

I also added a test, but since this actually requires to download the file, I ignored the test.


boldtrn:
> Wouldn't this change mean that we potentially always try the wrong name first? If we make this change at all we should test only once and make it a permanent change.

Sorry for the late reply. What do you mean here?

We only try the wrong name once on the server. We download the file and store it using the right now. [Here](https://github.com/boldtrn/graphhopper/blob/086dd8518f0eee29c3d87cb90c645027908a03e4/core/src/main/java/com/graphhopper/reader/dem/AbstractSRTMElevationProvider.java#L175) you can see that we have a fixed file name locally and only change the remote url.

karussell:
If there is a different naming schema it will try the wrong name first, then fail, then download it and try again with the wrong file name etc

boldtrn:
> try again with the wrong file name etc

I don't think so, because we store the file locally with the correct naming scheme?

karussell:
> I don't think so, because we store the file locally with the correct naming scheme?

Imagine there are two zip files: A and B

We request A.hgt.zip and learn that this does not exist, then download Ahgt.zip. The next file will be B.hgt.zip and not Bhgt.zip IMO

devemux86:
Note that files in a [server](https://dds.cr.usgs.gov/srtm/version2_1/SRTM3/North_America/) have both name schemes.
So shouldn't try always with the right name and if fail try with the orher one?

karussell:
Ah, that it is ugly and brings light into this issue - thanks a lot for pointing this out!

PR #1303, opened 2018-03-07
(WIP) Isochrone documentation

willcohen:
Signed-off-by: Will Cohen <willcohen@users.noreply.github.com>

Re #1237 and a first pass at documentation:
- Adds annotations to the process used in IsochroneServlet to docs/isochrone/java.md, replacing the request parameters with variables at the top. Is this the right approach? It seems like sticking as close as possible to what's actually being used on the backend is important.
- Describes the parameters for /isochrone in docs/web/api-doc.md, based on the [directions API docs](https://github.com/graphhopper/directions-api-java-client/blob/master/client/src/main/java/com/graphhopper/directions/api/client/api/IsochroneApi.java).
- Does not currently have javadocs. The algorithm itself is taken from OTP, and the meat of the isochrone generation itself is spread among the multiple components. Where would you like to see more information?

### DIFF
--- a/docs/isochrone/java.md
+++ b/docs/isochrone/java.md
@@ -7,19 +7,6 @@ To create an isochrone in Java code:
 
 You'll first need to build off an existing Graphhopper instance for [routing](/../core/routing.md).
 
-```java
-// create one GraphHopper instance
-GraphHopper hopper = new GraphHopperOSM().forServer();
-hopper.setDataReaderFile(osmFile);
-// where to store graphhopper files?
-hopper.setGraphHopperLocation(graphFolder);
-hopper.setEncodingManager(new EncodingManager("car"));
-
-// now this can take minutes if it imports or a few seconds for loading
-// of course this is dependent on the area you import
-hopper.importOrLoad();
-```
-
 Next, compute the isochrone itself.
 ```java
 
--- a/docs/isochrone/java.md
+++ b/docs/isochrone/java.md
@@ -20,115 +20,29 @@ hopper.setEncodingManager(new EncodingManager("car"));
 hopper.importOrLoad();
 ```
 
-Next, you'll need to set a range of variable options for isochrone generation.
-```java
-String vehicle = "car";
-int buckets = 1L;
-boolean reverseFlow = false;
-String queryStr = "point";
-String resultStr = "polygon";
-double distanceInMeter = 10000;
-long timeLimitInSeconds = 600L;
-// bigger raster distance => bigger raster => less points => stranger buffer results, but faster
-double rasterDistance = 0.75;
-// bigger buffer distance => less holes, lower means less points!
-double bufferDistance = 0.003;
-// precision of the 'circles'
-int quadrantSegments = 3;
-
-```
-
 Next, compute the isochrone itself.
 ```java
-// parse the query point and get an encoder
 
-GHPoint query = GHPoint.parse(queryStr);
-
+// get encoder from GraphHopper instance
 EncodingManager encodingManager = hopper.getEncodingManager();
-if (!encodingManager.supports(vehicle)) {
-    throw new IllegalArgumentException("vehicle not supported:" + vehicle);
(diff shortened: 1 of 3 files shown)


docs/isochrone/java.md
```diff
@@ -0,0 +1,134 @@
...
+        propMap.put("bucket", index);
+        geometryMap.put("type", "Polygon");
+        // we have no holes => embed in yet another list
+        geometryMap.put("coordinates", Collections.singletonList(polygon));
+        index++;
+    }
+    calcRes = polyList;
+} else {
+    throw new IllegalArgumentException("type not supported:" + resultStr);
+}
```
karussell: I'm not sure if we should duplicate the whole servlet code. Why not show really only a simple time limit example that returns the point lists and link to the full code in the servlet?

docs/web/api-doc.md
```diff
@@ -219,3 +219,24 @@ HTTP error code | Reason
 500             | Internal server error. It is strongly recommended to send us the message and the link to it, as it is very likely a bug in our system.
 501             | Only a special list of vehicles is supported
 400             | Something was wrong in your request
+
+## Isochrone
+
+In addition to routing, the URL path to obtain an isochrone is `/isochrone`.
```
karussell: Hmmh, when I read this again I think the original routing description should exactly the opposite:

> The URL of the local instance is http://localhost:8989
> The end point to obtain the route is /route

?

docs/web/api-doc.md
```diff
@@ -219,3 +219,24 @@ HTTP error code | Reason
...
+reverse_flow                | false   | If false the flow goes from point to the polygon, if true the flow goes from the polygon inside to the point. Example usage for false: *How many potential customer can be reached within 30min travel time from your store* vs. true: *How many customers can reach your store within 30min travel time.* (optional, default to false)
+point                       |         | Specify the start coordinate (required)
+result                      | polygon | Can be "pointlist" or "polygon".
+distance_limit              | -1      | Specify which distance the vehicle should travel. In meter. (optional, default to -1)
+time_limit                  | 600     | Specify which time the vehicle should travel. In seconds. (optional, default to 600)
+isochrone.raster_distance   | 0.75    | bigger raster distance => bigger raster => less points => stranger buffer results, but faster
+isochrone.buffer_distance   | 0.003   | bigger buffer distance => less holes, lower means less points!
+isochrone.quadrant_segments | 3       | precision of the 'circles'
```
karussell: are those `isochrone.xy` parameters still used somewhere :) ?

araichev:
I suggest clarifying the ``buckets`` parameter in the [v1 isochrone docs](https://graphhopper.com/api/1/docs/isochrone/) to include the fact that **equal** time/distance intervals are used for the subisochrones, e.g.

"Number by which to divide the given ``time_limit`` to create ``buckets`` nested isochrones of time intervals ``time_limit/buckets``, ``time_limit/(buckets - 1)``,..., ``time_limit``. Applies analogously to ``distance_limit``. "

willcohen:
Apologies for the delay on this. Rebased and switched to master.

Latest commit simplifies the example and removes a lot of the extraneous servlet code. Is this streamlined enough or should it be even closer to what's listed in [IsochroneTest](https://github.com/graphhopper/graphhopper/blob/master/isochrone/src/test/java/com/graphhopper/isochrone/algorithm/IsochroneTest.java)?

Additionally, in cleaning up the API reference, it's looking like the "weighting" parameter is no longer used, since [weightingStr](https://github.com/graphhopper/graphhopper/blob/9e2765c3e7233a963bb1f8da8afdfb46f939417d/web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java#L57) never gets called again. Is it correct to take it out of the API docs?

karussell:
Thanks a lot!

> Is it correct to take it out of the API docs?

No, I think this is a bug - thanks for catching. See #1416

Issue #1280, opened 2018-02-03
Unify builk customization of graph properties like speed and access


karussell:
See my comment [here](https://github.com/graphhopper/graphhopper/issues/1214#issuecomment-362801969), ie. we have three approaches:

 * spatial rules: requires to build an index up front and is best used via iterating through all edges. So usage will be mostly on import time
 * change graph API: currently does not support polygons and should be used for rather small areas only like for a construction or longer traffic jam
 * block_area can be used per request but has bugs like #1214 and #1127 (the change graph API highly likely has similar problems)

boldtrn:
How do you propose to unify the code?

Change graph and block area both use the `GraphEdgeIdFinder`. So the core logic seems to be the same. Maybe these two could become more similar by passing an encoded json into the block area?

The spatial rules have to exist before the edge is created, therefore it cannot use the GraphEdgeIdFinder. I am not sure how we could merge the logic? It might be possible to create a "ChangeGraphSpatialRuleIndex". Then we wouldn't have to update the edges anymore, but could check the properties in the Weighting. Therefore changes would be faster, but routing would be slower.

PR #1306, opened 2018-03-07
Polygon for blocked areas

andreylh:
This PR allows block areas to support polygons (discuss [here](https://discuss.graphhopper.com/t/polygon-for-blocked-areas/1464) and [here](https://discuss.graphhopper.com/t/restricted-routing-zones-specified-by-a-polygon/2318/10))

How this works (it's in Portuguese, "Centro" means "Center"):
poligono

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
+++ b/core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
@@ -27,6 +27,7 @@ import com.graphhopper.util.PointList;
 import com.graphhopper.util.shapes.BBox;
 import com.graphhopper.util.shapes.Circle;
 import com.graphhopper.util.shapes.GHPoint;
+import com.graphhopper.util.shapes.Polygon;
 import com.graphhopper.util.shapes.Shape;
 import com.vividsolutions.jts.geom.*;
 
@@ -77,12 +78,20 @@ public class GraphEdgeIdFinder {
         if (shape.contains(qr.getSnappedPoint().lat, qr.getSnappedPoint().lon))
             edgeIds.add(qr.getClosestEdge().getEdge());
 
+        final boolean isPolygon = shape instanceof Polygon;
+
         BreadthFirstSearch bfs = new BreadthFirstSearch() {
             final NodeAccess na = graph.getNodeAccess();
             final Shape localShape = shape;
 
             @Override
             protected boolean goFurther(int nodeId) {
+                if (isPolygon) {
+                    BBox bbox = localShape.getBounds();
+                    double lat = na.getLatitude(nodeId);
+                    double lon = na.getLongitude(nodeId);
+                    return lat <= bbox.maxLat && lat >= bbox.minLat && lon <= bbox.maxLon && lon >= bbox.minLon;
+                }
                 return localShape.contains(na.getLatitude(nodeId), na.getLongitude(nodeId));
             }
 
@@ -136,7 +145,10 @@ public class GraphEdgeIdFinder {
             for (int i = 0; i < blockedCircularAreasArr.length; i++) {
                 String objectAsString = blockedCircularAreasArr[i];
                 String[] splittedObject = objectAsString.split(innerObjSep);
-                if (splittedObject.length == 4) {
+                if (splittedObject.length > 4) {
+                    final com.graphhopper.util.shapes.Polygon polygon = Polygon.parsePoints(objectAsString);
+                    findEdgesInShape(blockArea.blockedEdges, polygon, filter);
+                } else if (splittedObject.length == 4) {
                     final BBox bbox = BBox.parseTwoPoints(objectAsString);
                     if (bbox.calculateArea() > useEdgeIdsUntilAreaSize)
                         blockArea.add(bbox);
--- a/core/src/main/java/com/graphhopper/util/shapes/Polygon.java
+++ b/core/src/main/java/com/graphhopper/util/shapes/Polygon.java
@@ -1,0 +1,193 @@
+package com.graphhopper.util.shapes;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Polygon implements Shape {
+
+    private final double[] lat;
+    private final double[] lon;
+
+    private double minLat;
+    private double minLon;
+    private double maxLat;
+    private double maxLon;
+
+    private final double epsilon;
+    private static final double GROWN_FACTOR = 0.003;
+
+    public Polygon(double[] lat, double[] lon) {
+        if (lat.length != lon.length) {
+            throw new IllegalArgumentException(
+                    "Points must be of equal length but was " + lat.length + " vs. " + lon.length);
+        }
+        if (lat.length == 0) {
+            throw new IllegalArgumentException("Points must not be empty");
+        }
+        this.lat = lat;
+        this.lon = lon;
+
+        for (int i = 0; i < lat.length; i++) {
+            if (i == 0) {
+                minLat = lat[i];
+                maxLat = lat[i];
+                minLon = lon[i];
+                maxLon = lon[i];
+            } else {
+                if (lat[i] < minLat) {
+                    minLat = lat[i];
+                } else if (lat[i] > maxLat) {
+                    maxLat = lat[i];
+                }
+                if (lon[i] < minLon) {
+                    minLon = lon[i];
+                } else if (lon[i] > maxLon) {
+                    maxLon = lon[i];
+                }
+            }
+        }
+
+        minLat -= GROWN_FACTOR;
+        minLon -= GROWN_FACTOR;
+        maxLat += GROWN_FACTOR;
+        maxLon += GROWN_FACTOR;
+
+        epsilon = (maxLat - minLat) / 10;
+    }
+
+    @Override
+    public boolean intersect(Shape o) {
+        // TODO Auto-generated method stub
+        return false;
+    }
+
+    @Override
+    public boolean contains(double lat, double lon) {
+        if (lat < minLat || lat > maxLat || lon < minLon || lon > maxLon) {
+            return false;
+        }
+
+        double rayStartLat = maxLat - (minLat / 2);
+        double rayStartLon = minLon - epsilon;
+
+        boolean inside = false;
+        int len = this.lat.length;
+        for (int i = 0; i < len; i++) {
+            if (edgesAreIntersecting(rayStartLon, rayStartLat, lon, lat, this.lon[i], this.lat[i],
+                    this.lon[(i + 1) % len], this.lat[(i + 1) % len]))
+                inside = !inside;
+        }
+        return inside;
+    }
+
+    private boolean edgesAreIntersecting(double v1x1, double v1y1, double v1x2, double v1y2, double v2x1, double v2y1,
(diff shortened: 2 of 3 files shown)


karussell:
Thanks a lot for this! We'll have a look, would you mind to send us [your CLA](https://github.com/graphhopper/graphhopper/blob/master/.github/CONTRIBUTING.md#license-agreement)? Please also go ahead and post this PR into the referenced discussion threads.

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -0,0 +1,193 @@
+package com.graphhopper.util.shapes;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Polygon implements Shape {
```
karussell: We should avoid another Polygon class here. The problem here is probably that the current one is in the wrong package? com.graphhopper.routing.util.spatialrules

Or should we use the polygon from JTS?

andreylh: I implemented a new ```Polygon``` class because:
 - My class has the ```GROWN_FACTOR``` constant, which is used to increase bounds of ```Polygon```
 - The existing ```Polygon``` class doesn't implement ```Shape``` interface, so I cannot pass it to ```findEdgesInShape``` method.
 - The package com.graphhopper.util.shapes seem correct to me since it already has others geometric shapes.

I'm not sure if I can apply this changes to the existing ```Polygon``` class and use. If I can, please let me know and I do the changes.

I don't know if ```Polygon``` from JTS library has all features that I need, so I'm going to study the library if I can use it.

karussell: > I'm going to study the library if I can use it.

This would be the preferred option

> If I can, please let me know and I do the changes.

Yes, if JTS is not possible please apply the changes to the existing class and e.g. use a default GROW_FACTOR of 1

boldtrn: > The existing Polygon class doesn't implement Shape interface

I didn't implement the Shape interface because at that point we did not need all the methods supported by the shape interface like intersects etc. If using JTS does not work, feel free to extend the existing class :).

andreylh: So, I did the changes using JTS library and it works. The problem was the performance. Polygon class is almost 10x faster than JTS.

Maybe I did something wrong, I do not know. Anyway, I'm going to commit the changes using Polygon class.

karussell: This is interesting thanks for letting us know. So which in-built methods did you use from JTS or where did the speed loss came from?

karussell: btw: applying a (speed or access) change permanently via polgon would be also nice and we currently investigate to make this more streamlined in #1280

andreylh:
@karussell, I did the changes using JTS library and it works. The problem was the performance. Polygon class is almost 10x faster than JTS.

Maybe I did something wrong, I do not know. Anyway, I committed the changes using Polygon class.

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
andreylh: @karussell I did not investigate where the performance problem is, just did a test with 484 requests. Using JTS the time was approximately 464 seconds versus 42 of the Polygon class

You can see the [changes using JTS solution](https://github.com/andreylh/graphhopper/commit/00157d0f7e08f0282676a879f1a2def65d63738d) in my forked repo

karussell: Thanks! What is this method call: GeometryFactory.createPointFromInternalCoord ?

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -69,6 +68,11 @@ public Polygon(double[] lat, double[] lon) {
             }
         }
 
+        minLat -= GROWN_FACTOR;
+        minLon -= GROWN_FACTOR;
+        maxLat += GROWN_FACTOR;
+        maxLon += GROWN_FACTOR;
```
karussell: Here it would be better if we can keep the default constructor and create a second one like

public Polygon(double[] lat, double[] lon, double growFactor)

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -92,6 +96,11 @@ public boolean contains(GHPoint point) {
         return contains(point.lat, point.lon);
     }
 
+    @Override
+    public boolean intersect(Shape o) {
+        return false;
```
karussell: Here it should throw an UnsupportedException and not return a potentially wrong result.

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -192,4 +221,21 @@ public double getMaxLat() {
...
+
+        for (int j = 0; j < arr.length; j++) {
+            if (j % 2 == 0) {
+                lats[j / 2] = Double.parseDouble(arr[j]);
+            } else {
+                lons[(j - 1) / 2] = Double.parseDouble(arr[j]);
+            }
+        }
+
+        return new Polygon(lats, lons);
```
karussell: @boldtrn do you think "lat,lon,lat2,lon2,..." is a good format? At least we should check that the count is dividable by 2

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
andreylh: It's to transform Coordinate into Geometry (Point) to be abble to pass into contains() method

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
boldtrn: Mhm, good question. My issue with this is that it is really hard to read and write for humans. It is easy for computers though. We should definitely check if the length is a multiple of 2. Maybe a geojson like format would be better? e.g. [[lon,lat],[lon2,lat2]]... But then we don't do this for BBox. Maybe we should do this for BBox and Polygon?

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
```diff
@@ -92,7 +101,7 @@ protected boolean checkAdjacent(EdgeIteratorState edge) {
                     edgeIds.add(edge.getEdge());
                     return true;
                 }
-                return false;
+                return isPolygon;
```
boldtrn: Why do we return isPolygon here?

boldtrn:
Really nice approach @andreylh, thanks for providing this PR. Would you mind to add a test that shows that block Polygon works?

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
karussell: When I think about this, the only format that makes sense would be like it is now. But this shows again the ugliness of the URL request approach that we currently have and we should invest more into 1112 to being able to request via JSON where we then could easily use GeoJSON for something like this.

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
andreylh: When the shape is polygon I need to check if is adjacent even if lat/lon is outside of shape bounds. What can I do is return true only if lat/lon is inside of bbox of the polygon. What do you think?

andreylh:
@boldtrn thank you very much. Sure, I can do that.

andreylh:
@karussell @boldtrn Is there still something I need to change?

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -192,4 +224,26 @@ public double getMaxLat() {
...
+
+    @Override
+    public String toString() {
+        return "polygon (" + lat.length + " points)";
+    }
+
+    public static Polygon parsePoints(String pointsStr, double growFactor) {
+        String[] arr = pointsStr.split(",");
+
+        double[] lats = new double[arr.length /2];
```
karussell: @andreylh we need a small length. E.g. if arr.length%2==1 throw IllegalArgumentException("incorrect polygon specified")

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -192,4 +224,26 @@ public double getMaxLat() {
...
+        double[] lons = new double[arr.length /2];
+
+        for (int j = 0; j < arr.length; j++) {
+            if (j % 2 == 0) {
+                lats[j / 2] = Double.parseDouble(arr[j]);
+            } else {
+                lons[(j - 1) / 2] = Double.parseDouble(arr[j]);
```
karussell: here `j/2+1` makes more sense as index into the lons array IMO

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
karussell: I do not fully understand this. Do you mean with "even if lat/lon is outside of shape bounds" that adjacent edges are checked because later edges could be again inside the polygon?

We discussed this with @michaz and instead of this "blurry" workaround that will still fail for certain cases, we should probably implement fetching the edgeIds from the LocationIndexTree instead of from the Graph. Of course this would be out of the scope of this PR but then we should `return false` here? Or should we keep it as a compromise?

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
michaz: @karussell Can you give an example of a ``j`` where it is clear how it would make more sense?

karussell: Oh, probably I misunderstood this index, sorry for the bubble

karussell:
@andreylh I've added two comments (regarding if arr.length%2==1 and `return isPolygon`)

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
andreylh: @karussell You're right. Adjacent edges could be outside of the polygon but inside of BBox, this is why I need return true.

At the moment I see three options. We could:
 - Keep returning true;
 - Return false only if the adjacent node is outside of BBox (same as goFuther() method);
 - Increase depth of BFS. But it's difficult to determine a value.

karussell: For the moment I would prefer the second option (same as goFurther to reduce the candidates meaningful) and later we need to properly fix this via implementing a new method for LocationIndexTree (`tree.traverseEdges(callHookPerEdge, shapeAsFilter)`). This way we would also not being stuck in a potential subnetwork (assume 2 islands in a shape then we would traverse only within the 1 island) and we do not need a 'middle' entry point that could be somewhere far away from the road network

PR #1314, opened 2018-03-18
Dropwizardify - really now

michaz:
- [x]  switch back to commit 72ef9c1
- [x]  adapt graphhopper.sh #1319
- [x] ~~additionally to -Dgraphhopper allow -Dgh (or only allow the latter?)~~ no, we use the graphhopper.sh script for shorter and commonly used parameters
- [x]  avoid double logging, see above
- [x]  turn off admin service in config by default for security reasons
- [x]  adapt docs to document new config options or provide a backward compatible way
- [x]  add healthcheck to avoid warning at start up
- [x]  use endpoint /maps instead of /webapp for UI demo
- [x]  the graphhopper banner for CLI is not fully printed, one part of the second P is missing


devemux86:
Seen some new javax dependencies, were they tested in GraphHopper's all supported platforms (e.g. Android) ?

michaz:
@devemux86 Where? Maybe I'm mixing something up, but I only see them in the client and in the web module... At least in this branch.

devemux86:
In client-hc pom.xml there are 2 new dependencies.

michaz:
> In client-hc pom.xml there are 2 new dependencies.

Ah yes, had to look into my own commit message to remember what was up there.. Java 9 wants those dependencies declared, since those APIs are not part of the regular JRE anymore. Modularization and such.

Would be very surprised if this were a problem on Android, but.. how do I test that again?

michaz:
(I hope at one point we can automate this..)

devemux86:
Android sample (which can be tested on device) currently uses only the core module.
With @boldtrn we use the `client-hc` in [Kurviger](https://kurviger.de/en), so I'd need to check specifically the branch to see how it plays.

michaz:
> additionally to -Dgraphhopper allow -Dgh (or only allow the latter?)

@karussell Can I leave that one to you, please? Currently, this works through the _existing_ mechanism in ``CmdArgs``, I didn't touch any of this.

I think I wouldn't change it, to be honest. ``graphhopper`` is already a thing (and it's what the block in the ``.yml`` is called), ``gh`` is not.

In the midterm, we should migrate to a new, real, hierarchical configuration -- all of this is just a backport of what's there.

michaz:
_Or_ we do it like you say, rename the ``.yml`` block and the environment variable prefix to ``gh``, and reserve ``graphhopper`` for the future _new_ config?

karussell:
The solution for my "long parameter problem" is to create abbreviation arguments like I've did in #1319 which uses ideas from #1129 and #1244 to allow overwriting parameters like config, datareader.file and improves this scripting approach a lot IMO. 

It still keeps backward compatibility with the simple "./graphhopper.sh <action> <file>" approach.

I'll only add a few "short parameters" for easier usage of torture, measurement and gtfs and will then merge with this branch if you are ok.

karussell:
@michaz I forgot that squashing removes authorship of @legraina. May I rewrite this commit to attach authorship in this commit and push force or do you have changes already downstream?

karussell:
I thought this problem is now solved: https://blog.github.com/2018-01-29-commit-together-with-co-authors/

karussell:
Fixed some more issues. The only remaining issue is "Class path contains multiple SLF4J bindings" but we can also merge first and fix later. WTDY @michaz ?

reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/GraphHopperGtfs.java
```diff
@@ -341,6 +339,10 @@ public static GraphHopperStorage createOrLoad(GHDirectory directory, EncodingMan
             for (int i = 0; i < id; i++) {
                 new GtfsReader("gtfs_" + i, graphHopperStorage, gtfsStorage, ptFlagEncoder, walkNetworkIndex).readGraph();
             }
+            // This currently needs to happen as the last step, since we cannot add new nodes after this step.
+            // This means that disconnected parts of the transit network are removed as well.
+            // If we don't want that, we need to think of something.
+            new PrepareRoutingSubnetworks(graphHopperStorage, Collections.singletonList(ptFlagEncoder)).doWork();
```
karussell: Strange, why does it revert your change @michaz although I merged master? 

https://github.com/graphhopper/graphhopper/commit/506489399629519804b44ff1cf5058b79db46cf5

ammagamma:
I used to have different run configurations in IntelliJ were I was passing graphhopper properties as program arguments, like datareader.file etc., now I need to have these in config.yml, right ? I would at least find it useful to have some kind of naming pattern or dedicated folder in .gitignore where I can put different config.yml files if it needs to be done this way, or what is your approach for this ? For example I want to run GraphHopperApplication from the IDE for debugging with certain program arguments, but to support different settings I need multiple config.yml files ?

michaz:
Yes, the idea is that those are your run configurations. I'd suggest git-ignoring those files locally -- we shouldn't assume a user _doesn't_ want to commit a configuration.

ammagamma:
Ok, but then we should remove config.yml from .gitignore ? https://github.com/graphhopper/graphhopper/blob/2c10f76a4dac00b82ad9ed68d1b57ad411d9a58f/.gitignore#L8

ammagamma:
hmm no nevermind, in this case it would be too likely that one accidentally pushes some custom config I guess

michaz:
> Ok, but then we should remove config.yml from .gitignore ?
I agree. Reason: Git is distributed -- what the user is commiting to might just as well be their own project repository where they commit their configuration to. ``.gitignore`` shoudln't be biased towards ``origin/master``. :-)
But maybe in this case the benefits outweigh the philosophy, don't know.

michaz:
Actually, I would go one step further, rename ``config-example.yml`` to ``config.yml``, remove it from ``.gitignore`` and change it so that it actually runs when run from the project root, i.e. without the commented-out OSM filename.

ammagamma:
I don't think I have a strong opinion here, but why not. Only thing is that it should be dedicated, maybe even a little 'migration guide' for dropwizard to make it easier to switch. I would still find it useful to have something like `local/` in .gitignore. I am using .git/info/exclude where I added `local/` with some folders for maps, config files etc.

karussell:
I have it working with just one config.yml like before and e.g. the following **VM** options:

```
-Xmx4g -Xms4g -Dgraphhopper.graph.flag_encoders=car,foot -Dgraphhopper.datareader.file=berlin-latest.osm.pbf -Dgraphhopper.graph.location=graph-cache
```

karussell:
Regarding the .gitignore there is also this possibility: https://stackoverflow.com/q/7335420/194609

Not sure if this applies in your case.

ammagamma:
Ok using the VM options works (and overwrites the settings from config.yml). The link you posted is the global version of using `.git/info/exclude`, so for me these are enough options to choose from :)

Issue #1311, opened 2018-03-16
NPE for public transit


karussell:
See discussion here: https://discuss.graphhopper.com/t/2913

`{"vehicle":"pt","weighting":"short_fastest","point":["42.4448765,-76.4808143","42.4301429,-76.5082167"],"points_encoded":false,"pt.arrive_by":false,"ch.disable":true,"pt.walk_speed":3,"pt.earliest_departure_time":"2018-03-15T14:49:15.303Z","pt.profile":true,"pt.limit_solutions":6}`

```
java.lang.NullPointerException\n\tat com.graphhopper.reader.gtfs.Label$1$1.next(Label.java:120)\n\tat 
com.graphhopper.reader.gtfs.Label$1$1.next(Label.java:105)\n\tat 
java.lang.Iterable.forEach(Iterable.java:74)\n\tat 
com.graphhopper.reader.gtfs.TripFromLabel.getTrip(TripFromLabel.java:130)\n\tat 
com.graphhopper.reader.gtfs.GraphHopperGtfs$RequestHandler.parseSolutionsAndAddToResponse(GraphHopperGtfs.java:219)\n\tat 
```

michaz:
Argh, that's a tough one that goes all the way down to graph construction.
I somehow manage to create a graph which has an edge from A to B, and an edge _with the same number_ from B to C. Reproducibly. In the datastore.

Tracing this now.

michaz:
Apparently, during graph constructing, we cannot add transit _after_ having removed disconnected network components.

I'm doing that as the _last_ step now instead, which is _probably OK in most cases_.

Closing this, as the NPE is definitely gone, and will keep an eye on whether the change matters in unexpected ways.

karussell:
Thanks!

> Apparently, during graph constructing, we cannot add transit after having removed disconnected network components.

This should be allowed or did you mean that adding transits should not be allowed after this step? If that is the case you could call `graphHopperStorage.freeze()` to ensure this.

michaz:
No, this: #1313

PR #1316, opened 2018-03-19
Graph should be consistent also after subnetwork removal

### LINKED ISSUE #1313: Broken graph after removing subnetworks, then adding nodes
Adding edges (or, at least, adding edges to new nodes) to a graph after having removed disconnected subnetworks results in a corrupted graph:

https://github.com/graphhopper/graphhopper/commit/7f82541a3d391d53d7992df3333e48cb6d0a7a4a

karussell:
Fixes #1313. Related to #1311


karussell:
This fix is not sufficient. We need to account for a large nodeCount: https://github.com/graphhopper/graphhopper/commit/01d2bc10d80a33e1cb2d2ea9635e36232e86ce3f

PR #1319, opened 2018-03-23
Adapt graphhopper script for dropwizard branch

karussell:
And replaces #1244 
 
 - [x] support for gtfs
 - [x] ~~some more short parameters (test measurement and torture)~~ doesn't work at all. we should fix in different PR
 - [x] ~~convert all unknown parameters into `-Dgraphhopper.<unknown>` instead of printing the error?~~ Problem is that this could hide compatibility issues between this and the old arguments. We can do this in one of the next releases as for now other parameters are possible via the old `export GH_WEB_OPTS=xy` approach
 - [x] ~make old approach working: `export JETTY_PORT=8990;./graphhopper.sh web ~/Downloads/berlin-latest.osm.pbf` even when port is removed from config~ use `-port` command line option instead
 - [x]  ~even if not configured the admin stuff is started!~ no problem when we bind to localhost
 - [x] ~test IPFilter~ can be remove since we bind to localhost

Let us also document the old vs. the new way of configuring something:

|old | new
|-----|--------
|export JETTY_PORT=8989| --port 8989
|export GH_FOREGROUND=false| -d

### DIFF
--- a/graphhopper.sh
+++ b/graphhopper.sh
@@ -225,10 +225,13 @@ echo "## now $ACTION. JAVA_OPTS=$JAVA_OPTS"
 if [ "$ACTION" = "web" ]; then
   export MAVEN_OPTS="$MAVEN_OPTS $JAVA_OPTS"
 
-  RC_BASE=./web/src/main/webapp
+  if [ $JETTY_PORT != "" ]; then
+    GH_WEB_OPTS="$GH_WEB_OPTS -Ddw.server.applicationConnectors[0].port=$JETTY_PORT"
+  fi
+
   if [ "$RUN_BACKGROUND" = "true" ]; then
     exec "$JAVA" $JAVA_OPTS -Dgraphhopper.datareader.file="$OSM_FILE" -Dgraphhopper.graph.location="$GRAPH" \
-         -Dgraphhopper.jetty.resourcebase="$RC_BASE" -Dgraphhopper.jetty.port=$JETTY_PORT -Dgraphhopper.jetty.host=$JETTY_HOST \
+         -Dgraphhopper.jetty.resourcebase="$RC_BASE" -Dgraphhopper.jetty.host=$JETTY_HOST \
          $GH_WEB_OPTS \
          -jar "$JAR" server $CONFIG <&- &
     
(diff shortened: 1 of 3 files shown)


graphhopper.sh
```diff
@@ -2,68 +2,97 @@
...
 
 if [ "$ACTION" = "" ]; then
- echo "## action $ACTION not found. try" 
- printUsage
+ echo "## action $ACTION not found!"
+ printBashUsage
+fi
+
+# default init, https://stackoverflow.com/a/28085062/194609
+: "${CONFIG:=config.yml}"
```
justinmk2: `:` is a easy-to-miss synonym for `true`. This form is clearer:

    CONFIG=${CONFIG:-config.yml}

and doesn't require studying a stackoverflow post.

legraina: Nice update.
I was reading again the flags, and I have perhaps others suggestions:
- -r for --datareader.file
- -d for --run-backgound or --detach (which is really common for different software)
- -f for --force-download (it can still be -fd!)

karussell: @legraina Thanks! For `-r` I'm not sure. I think -f for "file" is not uncommon. Maybe -i or -in for input as we have -o for output ...

> and doesn't require studying a stackoverflow post.

@justinmk2 the other method requires typing the variable twice ;) (maybe useful later for long variables) and to be honest the `:-` is not 100% clear too and the `:=` smells more intutive

karussell:
Have updated also with look to comments from @legraina. For me this seems to be ok for now and we should try this a bit and improve step by step, e.g. some actions do not work at all but this would require some more class similar to the import action.

Using GTFS is also possible now with this script 

E.g. the short form looks like:

```
export JAVA_OPTS="-Xmx10g -Xms10g"; ./graphhopper.sh -a web -i berlin-latest.osm.pbf -ig ~/Downloads/gtfs.zip -p pt
```

long form:

```
export JAVA_OPTS="-Xmx10g -Xms10g"; ./graphhopper.sh --action web --input berlin-latest.osm.pbf --gtfs gtfs.zip --profiles pt
```

(the test failure is something else I tried to fix in another commit but obviously it wasn't that successful)

graphhopper.sh
```diff
@@ -2,68 +2,105 @@
...
+  echo "     --action measurement does performance analysis of the current source version via random routes (Measurement class)"
+  echo "     --action torture     can be used to test real world routes via feeding graphhopper logs into a GraphHopper system (Torture class)"
+  echo "-i | --input:       path to the input map file or name of the file to download"
+  echo "-ig| --gtfs:        path to the input file for public transit routing (GTFS format)"
+  echo "-o | --graph-cache: directory for graph cache output"
+  echo "-p | --profiles:    specify a comma separated list of vehicle profiles"
+  echo "-c | --config:      specify the application configuration"
+  echo "-host:              specify to which host the service should be bound"
+  echo "-port:              start web server at specific port"
+  echo "-jar:               specify the jar file (useful if you want to reuse this script for custom builds)"
```
karussell: or should we better do something like `-h | --host` and `-p | --port` (clashes with `--profiles`) and `-j | --jar`?

legraina: Yes, it will be nice. I think you should use -p for port, and perhaps just --profile without a short version.

michaz:
Looks good to me.. Nice to see that the script could get shorter.

graphhopper.sh
karussell: ok. Maybe `-v` for `--vehicle-profiles` but yes, `--profiles` is also ok

karussell: -h for --host would clash with -h for help :)

karussell: And -v should be used for version. So I currently prefer to use --host and --port and --jar

PR #1307, opened 2018-03-14
update block_area to support polygons.  eg  39.9423088 116.3507785,39…

wxgang168:
….9451463 116.3599607,39.9469345 116.3653061,39.9480079 116.4186783,39.9480736 116.4249213,39.9468058 116.4271589,39.9147197 116.4283642,39.9094016 116.4293085,39.8995464 116.4300157,39.8967059 116.4370689,39.8775988 116.4394502,39.8699284 116.4371174,39.8692423 116.4353268,39.8699470 116.4153372,39.8698448 116.4114116,39.8710281 116.4088730,39.8685412 116.3627049,39.8672065 116.3532638,39.8673983 116.3445800,39.8688574 116.3424818,39.8754838 116.3432543,39.8954190 116.3428464,39.8961248 116.3442369,39.8971442 116.3502925,39.9020172 116.3510337,39.9203199 116.3504961,39.9407134 116.3497543,39.9423088 116.3507785

PR #1320, opened 2018-03-26
Modify the block_area module to support polygons.

wxgang168:


karussell:
As already previously noted in #1307: this is supported via #1306.

Can you explain what is different and what improvements we would get here? You already commented this in #1307 but I did not understand the benefits. Furthermore without new tests we cannot merge new features.

PR #1332, opened 2018-04-10
Rename getMaxId

### LINKED ISSUE #1321: Method name AllEdgesIterator.getMaxId() does not describe intention
The comment says:
```
the maximum edgeId in the graph which is roughly the number of edges
```

I think "roughly" refers to the possibility of edges being marked for deletion.
I also think this is the reason why this method was named like it was, and not e.g. ``size()`` or ``nEdges()``.

However, the maximum edgeId in the graph is precisely what this method does _not_ return. It returns maximum edgeId + 1. This is relied upon in ``QueryGraph``, where ``getMaxId()`` is in fact the first edge id ``QueryGraph`` thinks it can use.

Hence, when I implement ``Graph``, take the method's name literally (which I did), and use a ``QueryGraph`` on it, I get a duplicate edge id.

Would recomment ``size()`` or ``length()``, see ``java.util.BitSet``.

karussell:
Fixes #1321

### DIFF
--- a/reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/GraphHopperGtfs.java
+++ b/reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/GraphHopperGtfs.java
@@ -165,7 +165,7 @@ public final class GraphHopperGtfs implements GraphHopperAPI {
         private void substitutePointWithVirtualNode(int index, boolean reverse, GHPoint ghPoint, ArrayList<QueryResult> allQueryResults) {
             final GraphExplorer graphExplorer = new GraphExplorer(queryGraph, weighting, flagEncoder, gtfsStorage, realtimeFeed, reverse, extraEdges, true);
             int nextNodeId = graphWithExtraEdges.getNodes() + 2 + index; // FIXME: A number bigger than the number of nodes QueryGraph adds
-            int nextEdgeId = graphWithExtraEdges.getAllEdges().getMaxId() + 100; // FIXME: A number bigger than the number of edges QueryGraph adds
+            int nextEdgeId = graphWithExtraEdges.getAllEdges().length() + 100; // FIXME: A number bigger than the number of edges QueryGraph adds
 
             final List<Label> stationNodes = findStationNodes(graphExplorer, allQueryResults.get(index).getClosestNode(), reverse);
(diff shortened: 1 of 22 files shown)


reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/GraphHopperGtfs.java
```diff
@@ -165,7 +165,7 @@ GHResponse route() {
         private void substitutePointWithVirtualNode(int index, boolean reverse, GHPoint ghPoint, ArrayList<QueryResult> allQueryResults) {
             final GraphExplorer graphExplorer = new GraphExplorer(queryGraph, weighting, flagEncoder, gtfsStorage, realtimeFeed, reverse, extraEdges, true);
             int nextNodeId = graphWithExtraEdges.getNodes() + 2 + index; // FIXME: A number bigger than the number of nodes QueryGraph adds
-            int nextEdgeId = graphWithExtraEdges.getAllEdges().getMaxId() + 100; // FIXME: A number bigger than the number of edges QueryGraph adds
+            int nextEdgeId = graphWithExtraEdges.getAllEdges().length() + 100; // FIXME: A number bigger than the number of edges QueryGraph adds
```
karussell: Could we somehow know the number of edges QueryGraph adds in this place?

reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/RealtimeFeed.java
```diff
@@ -113,7 +113,7 @@ public static RealtimeFeed fromProtobuf(Graph graph, GtfsStorage staticGtfs, PtF
         final LinkedList<VirtualEdgeIteratorState> additionalEdges = new LinkedList<>();
         final Graph overlayGraph = new Graph() {
             int nNodes = 0;
-            int firstEdge = graph.getAllEdges().getMaxId()+1;
+            int firstEdge = graph.getAllEdges().length()+1;
```
karussell: Is +1 required here?

michaz: Ah, no. If ``maxId`` were true, it would have been required. :)

PR #1336, opened 2018-04-11
Some minor refactoring picked from turncost_ch branch.

ammagamma:
* Fixes some typos and whitespace
* Adds visited nodes sum to via routing template debug info.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
+++ b/core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
@@ -220,7 +220,7 @@ public class PrepareContractionHierarchies extends AbstractAlgoPreparation imple
             }
         };
 
-        maxLevel = prepareGraph.getNodes() + 1;
+        maxLevel = prepareGraph.getNodes();
         vehicleAllExplorer = prepareGraph.createEdgeExplorer(allFilter);
         vehicleAllTmpExplorer = prepareGraph.createEdgeExplorer(allFilter);
         calcPrioAllExplorer = prepareGraph.createEdgeExplorer(accessWithLevelFilter);
@@ -259,8 +259,7 @@ public class PrepareContractionHierarchies extends AbstractAlgoPreparation imple
         // graph contraction parameters, because it affects the node contraction order.
         meanDegree = prepareGraph.getAllEdges().length() / prepareGraph.getNodes();
         initSize = sortedNodes.getSize();
-        // todo: why do we start counting levels with 1 ??
-        int level = 1;
+        int level = 0;
         long counter = 0;
         long logSize = Math.round(Math.max(10, initSize / 100d * logMessagesPercentage));
         if (logMessagesPercentage == 0)
@@ -388,7 +387,7 @@ public class PrepareContractionHierarchies extends AbstractAlgoPreparation imple
         close();
     }
 
-    public void close() {
+    private void close() {
         nodeContractor.close();
         sortedNodes = null;
         oldPriorities = null;
--- a/core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
+++ b/core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
@@ -28,6 +28,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.*;
 
+import static com.graphhopper.util.Helper.nf;
 import static com.graphhopper.util.Parameters.Algorithms.ASTAR_BI;
 import static com.graphhopper.util.Parameters.Algorithms.DIJKSTRA_BI;
 
@@ -72,6 +73,7 @@ public class PrepareContractionHierarchies extends AbstractAlgoPreparation imple
     private double periodTime;
     private double lazyTime;
     private double neighborTime;
+    private int initSize;
 
     public PrepareContractionHierarchies(Directory dir, GraphHopperStorage ghStorage, CHGraph chGraph,
                                          Weighting weighting, TraversalMode traversalMode) {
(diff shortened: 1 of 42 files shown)


core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
```diff
@@ -239,10 +258,11 @@ private void contractNodes() {
         // no witness path can be found. this is not really what we want, but changing it requires re-optimizing the
         // graph contraction parameters, because it affects the node contraction order.
         meanDegree = prepareGraph.getAllEdges().length() / prepareGraph.getNodes();
+        initSize = sortedNodes.getSize();
+        // todo: why do we start counting levels with 1 ??
```
karussell: I think I was using it to distinguish uninitialized nodes from nodes with level=0 but now that we initialize all nodes with maxLevel at start this is probably no longer necessary.

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
```diff
@@ -364,25 +377,15 @@ private void contractNodes() {
...
-                + ", initSize:" + initSize
-                + ", periodic:" + periodicUpdatesPercentage
-                + ", lazy:" + lastNodesLazyUpdatePercentage
-                + ", neighbor:" + neighborUpdatePercentage
-                + ", " + Helper.getMemInfo());
+        logStats(counter, updateCounter);
+
+        // Preparation works only once so we can release temporary data.
+        // The preparation object itself has to be intact to create the algorithm.
+        close();
```
karussell: Should we make close private now?

core/src/main/java/com/graphhopper/routing/lm/PrepareLandmarks.java
```diff
@@ -122,12 +122,10 @@ public boolean loadExisting() {
...
     @Override
-    public void doWork() {
-        super.doWork();
-
+    public void doSpecificWork() {
         StopWatch sw = new StopWatch().start();
-        LOGGER.info("Start calculating " + lms.getLandmarkCount() + " landmarks, default active lms:"
-                + defaultActiveLandmarks + ", weighting:" + lms.getLmSelectionWeighting() + ", " + Helper.getMemInfo());
+        LOGGER.info("Start calculating {} landmarks, default active lms: {}, weighting: {}, {}",
+                lms.getLandmarkCount(), defaultActiveLandmarks, lms.getLmSelectionWeighting(), Helper.getMemInfo());
```
karussell: To be honest: I found the previous logging statement more clear :)

The "{}" notation is used to avoid string concatening if we disable logging for this class, but usually the advantage is only for massive logging like we could get in the REST server but even there we would not deactivate logging because it is valuable to have them for analyzation

ammagamma: I was not trying to improve performance here, but think {} is more readable, but that's probably because I am used to it. I will revert it, don't have a strong opinion here.

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
ammagamma: I think that would be better yes. btw @Nakaner mentioned it would sometimes be useful to have some protected hooks, for example at cleanUp() in GraphHopper.java https://media.ccc.de/v/2018-5257-eisenbahnrouting_mit_graphhopper#t=739 (around 12min)

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
ammagamma: Ok I would suggest starting the count at 0 then, this would have saved me some nerves the first time I did not realize it starts at 1 :)

core/src/main/java/com/graphhopper/routing/lm/PrepareLandmarks.java
karussell: The string itself is more clear. But counting the brackets {} {} {} to know which property goes where makes it harder for me. Also not a strong just a minor opinion :)

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
karussell: Yes, but we should not do this in general. We should start as closed as possible as we otherwise provide a broader API that is harder to evolve. Also inheritance should be less used and replaced with composition. All this means we should still provide hooks, but well thought out and only if there is a 'proper' use case. E.g. the GraphHopper class with all the "hooks" is simple but not that good IMO and needs refactoring at some point.

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
karussell: Oh, ok. :+1:

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
```diff
@@ -156,14 +158,31 @@ public PrepareContractionHierarchies setContractedNodes(double nodesContracted)
     }
 
     @Override
-    public void doWork() {
+    public void doSpecificWork() {
```
boldtrn: I don't understand this rename :). Could you explain the background and why doSpecificWork is better than doWork (BTW: I think both names are not especially good at explaining what happens in this method)?

karussell: The reason is highly likely that now it is not required to call the super method (so more methods but cleaner design IMO)

ammagamma: Yes the idea was to be able to just override `doSpecificWork` instead of overriding `doWork` and calling `super.doWork` in the overriding method. This way AbstractAlgoPreparation is only dealing with making sure a preparation is only run once and the subclasses are only concerned with preparing the algorithm, which I think is cleaner. Regarding the names I think `doWork` isn't so bad ? Tbh I was more concerned with `PrepareContractionHierarchies` because it is not really a noun, but now I got used to it :) I would have preferred something like `ContractionHierarchiesPreparer` and then maybe naming the `doWork()` method `prepare()` instead (and maybe `doPrepare` instead of `doSpecificWork`. Then again this class not only 'prepares' the graph, but also creates the Algorithm (its a RoutingAlgorithmFactory). Anyway, I think these are all rather minor issues and I do not insist on my change if you dont like it.

karussell: I have no preferences and the noun-argument incl. prepare sounds good.

PR #1338, opened 2018-04-13
Don't merge: How to store a custom attribute, read/write

michaz:
This is just a request for review of a use case. I want to store a custom edge attribute in the ``Graph``, and I want to use that attribute in a ``PathDetailsBuilder``. Am I doing it right?

What I particularly would like to know:

- Do I need a whole ``FlagEncoder`` to store one attribute? I am asking because the idea behind ``FlagEncoder`` really seems to be specifically **vehicle type** or **mode of transport**. (How do I know this? Because a broken vehicle icon appears in the web client.) Is there a different public API for what I am trying to do?

Things I do not require feedback about:

- that the ``FlagEncoder`` and the ``PathDetailsBuilder`` are hacked into ``GraphHopperManaged``
- that you are not seeing where I actually set the attribute -- just assume my setter on my custom ``FlagEncoder`` is used at some point during preprocessing

### DIFF
--- a/web/src/main/java/com/graphhopper/http/GraphHopperManaged.java
+++ b/web/src/main/java/com/graphhopper/http/GraphHopperManaged.java
@@ -20,15 +20,26 @@ package com.graphhopper.http;
 
 import com.graphhopper.GraphHopper;
 import com.graphhopper.reader.osm.GraphHopperOSM;
+import com.graphhopper.routing.util.DefaultFlagEncoderFactory;
+import com.graphhopper.routing.util.FlagEncoder;
+import com.graphhopper.routing.util.FlagEncoderFactory;
+import com.graphhopper.routing.weighting.Weighting;
 import com.graphhopper.spatialrules.SpatialRuleLookupHelper;
 import com.graphhopper.util.CmdArgs;
+import com.graphhopper.util.EdgeIteratorState;
+import com.graphhopper.util.PMap;
 import com.graphhopper.util.Parameters;
+import com.graphhopper.util.details.AbstractPathDetailsBuilder;
+import com.graphhopper.util.details.PathDetailsBuilder;
+import com.graphhopper.util.details.PathDetailsBuilderFactory;
 import io.dropwizard.lifecycle.Managed;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import com.michaz.OriginalDirectionFlagEncoder;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import java.util.List;
 
 @Singleton
 public class GraphHopperManaged implements Managed {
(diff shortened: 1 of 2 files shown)


web/src/main/java/com/graphhopper/http/GraphHopperManaged.java
```diff
@@ -41,8 +52,44 @@ public GraphHopperManaged(CmdArgs configuration) {
...
+            public List<PathDetailsBuilder> createPathDetailsBuilders(List<String> requestedPathDetails, FlagEncoder encoder, Weighting weighting) {
+                // request-scoped
+                OriginalDirectionFlagEncoder originalDirectionFlagEncoder = (OriginalDirectionFlagEncoder) graphHopper.getGraphHopperStorage().getEncodingManager().getEncoder("original-direction");
+                List<PathDetailsBuilder> pathDetailsBuilders = super.createPathDetailsBuilders(requestedPathDetails, encoder, weighting);
+                pathDetailsBuilders.add(new AbstractPathDetailsBuilder("r5_edge_id") {
+                    private int edgeId = -1;
+
+                    @Override
+                    public boolean isEdgeDifferentToLastEdge(EdgeIteratorState edge) {
+                        if (edge.getEdge() != edgeId) {
```
boldtrn: This won't work. You need to calculate your magic edgeId before comparing the two ids. This case will probably almost never pop up though.

boldtrn: So you need something like:
```
int newEdgeId = edgeId = edge.getEdge() * 2 + (originalDirectionFlagEncoder.isOriginalDirection(edge.getFlags()) ? 0 : 1);
if(newEdgeId!=edgeId)
```

karussell: Or use a different value for the currentValue or even always execute this code (ie. always return true) as we can assume that we loop through a list of edges and always call a new edge ... can or should we assume this?

michaz:
Sorry, I still think this is a misunderstanding. I am storing a property in the flags. Internally. That I can store whatever I want externally, in my own data structures, is clear.

The thing that I am trying to store internally is the direction bit I am inventing here. I already found out that I need that, and it is, of course, a direction-dependent property. So my use case is: Store an internal, direction dependent property. This is what I am trying to do here.

Where are the ``setBool``, ``setLong`` methods with which I can easily set a property? Do you mean those in ``AbstractFlagEncoder``? Those throw an exception if I use them for a flag they don't know, so I have to override them anyway, so how is that easier than writing my own method, which I can give a descriptive name?

karussell:
> That I can store whatever I want externally, in my own data structures, is clear.

If you have something externally then this does not depend on the direction so you would need some bit like you have implemented.

> So my use case is: Store an internal, direction dependent property. This is what I am trying to do here.

Then the procedure via two variables should be used. See reverseSpeedEncoder in Bike2WeightFlagEncoder. (You do not need an extra bit)

> so how is that easier than writing my own method, which I can give a descriptive name?

Uhm, this is a really good question :) 

They will be useful if you implement a Weighting that is used by multiple encoders that are not in the same inheritance hierarchy, e.g. to fetch max speed or similar.

PR #1348, opened 2018-04-21
added fullscreen button

karussell:
You can already try it on https://graphhopper.com/maps/

PR #1356, opened 2018-05-04
Changed fullscreen button size

boldtrn:
This is just a minor follow-up PR for #1348. I changed the size and stroke-width of the fullscreen button. The border was indeed specified correctly, not sure why it rendered incorrectly before.

In order to look good, I needed to actually change the fullscreen images and not just css properties. Not sure if you like it?

new-fullscreen-button


karussell:
Now the border looks different. Why are there changes to the css necessary? Currently it looks like

image

boldtrn:
> Now the border looks different

Mhm, haven't changed anything about the border. Did you try to move the map? I found that if there is a non uniform background the border looks different, because it's opaque. 

> Why are there changes to the css necessary?

The image is now smaller than the whole button, the css move the image to the center of the button and sets a white background.

``` 
background-color: white;
background-position: 50% 50%;
```

The other two changes are only formatting and I can revert them?

karussell:
> In order to look good

Do you mean to increase the border? Then I would close here as I think it is ok as it is now. As the intention of the button is to show that the map size gets bigger. Or maybe we should introduce a completely different look like an arrow.

boldtrn:
> Or maybe we should introduce a completely different look like an arrow.

That would be a nice as well.

> Do you mean to increase the border?

No, but IMHO the icon is too big. It doesn't really match the look of the other icons. I haven't changed anything about the border, I only changed the icon.

In your screenshot the border looks different, because the background is blue water. The border is partly invisible. So it looks darker, if you move the map to land, the border looks lighter again.

But yeah, this is very subjective, feel free to close this PR.

karussell:
Ok, thanks! I have tried this again and I cannot reproduce my problems. So I'll merge this after testing a few more browser - sorry for the big delay :)

karussell:
It is now live on graphhopper.com/maps :)

PR #1358, opened 2018-05-05
Update Leaflet to 1.3.1, enable autoPan on markers

fbonzon:
autoPan allows to move the markers beyond map edge when it approaches map viewport edges. New feature of Leaflet 1.3.

Localize date picker

Upgrade other dependencies

Tested at: https://gh.bikewm.com/


web/src/main/resources/assets/index.html
```diff
@@ -27,14 +27,14 @@
...
-        <link rel="stylesheet" href="css/leaflet.css?v=1.0.3" />
+        <link rel="stylesheet" href="css/leaflet.css?v=1.3.1" />
         <link rel="stylesheet" href="css/leaflet.elevation-0.0.4.css" />
         <link rel="stylesheet" href="css/leaflet_numbered_markers.css" />
         <link rel="stylesheet" href="css/leaflet.contextmenu.css" />
         <link rel="stylesheet" href="css/leaflet.loading.css?v=0.1.24" />
         <link rel="stylesheet" href="css/ui-lightness/jquery-ui.min.css" />
         <link rel="stylesheet" href="css/flatpickr.min.css">
-        <script type="text/javascript" src="js/main.js?v=0.9.1"></script>
+        <script type="text/javascript" src="js/main.js?v=1.0.0"></script>
```
boldtrn: This should probably be 0.10.0?

fbonzon:
Done.

How can I enable the public transport mode, or the new path details display in GraphHopper maps UI ? I didn't find a way to make the date field show up. I want to verify that the flatpickr upgrade I do here in package.json didn't break the date picker field.

fbonzon:
Verified date picker still works. I also added its localization, so it uses same language as the GraphHopper Maps UI:

screen shot 2018-05-06 at 03 22 33

PR #1362, opened 2018-05-06
Fix missing vertical bar for elevation

fbonzon:
It was not showing up and in a wrong color, due to a bad `.leaflet-overlay-pane` selector in CSS, which was removed in https://github.com/MrMufflon/Leaflet.Elevation/commit/94a5d5ee7e0708e1ea2399c1a95c37ab0e18b316.

I delete also here the unused themes, and keep only the white-theme.

screen shot 2018-05-06 at 08 28 56

Demo currently at: https://gh.bikewm.com/


karussell:
Thanks

Issue #1351, opened 2018-04-25
Make tools classes like measurement and querytorture working via graphhopper.sh script


karussell:
We don't want the tools dependency in web (would be ugly due to swing etc) but also don't want dropwizard in tools (?). We can solve this via creating a tools-jar-with-dependencies.

PR #1366, opened 2018-05-10
Adds some new functionality to Measurement class.

ammagamma:
This PR adds some new parameters to the Measurement class:

1.  `measurement.clean` can be used to delete an existing graph when running the measurement, default is false

2. `measurement.summaryfile` can be used to specify a file name and at the end of the measurement a single log line with selected properties of the measurement is written to the specified file. If no file is specified nothing is done.

3. `measurement.repeats` can be used to run the measurement multiple times

Using all of these parameters we can get a table of the (hard-coded) selected properties of the different runs for example like this:

    #graph.nodes graph.edges measurement.seed prepare.ch.time prepare.ch.shortcuts routing.distance_mean routing.mean routing.visited_nodes_mean routingCH.distance_mean routingCH.mean routingCH.visited_nodes_mean propertyFile
           32678       48788              123            1536                24389              14566.23         3.77                    5827.71                14021.56           0.96                       111.69 measurement2018-05-10_09_57_11.properties
           32678       48788              123            1152                24389              14566.23         3.67                    5827.71                14021.56           0.93                       111.69 measurement2018-05-10_09_57_36.properties
           32678       48788              123            1196                24389              14566.23         3.42                    5827.71                14021.56           0.93                       111.69 measurement2018-05-10_09_57_59.properties
           32678       48788              123            1136                24389              14566.23         3.49                    5827.71                14021.56           1.05                       111.69 measurement2018-05-10_09_58_21.properties

This is useful to check how stable the measurement data is across multiple runs and even more when we want to run graph preparations (like LM or CH) with different parameters (one still needs to add the parameters that should be varied, like number of lazy updates in CH etc.


tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -510,4 +525,75 @@ private void store(FileWriter fileWriter, String comment) throws IOException {
...
         fileWriter.flush();
     }
+
+    /**
+     * Writes a selection of measurement results to a single line in
+     * a file. Each run of the measurement class will append a new line.
+     */
+    private void writeSummary(String summaryLocation, String propLocation) {
+        // choose properties that should be in summary here
+        String[] properties = {
```
ammagamma: Here we can choose the properties that are written to the summary

boldtrn: Would it be possible to add a timestamp? I think this could really help when looking at historical data, for example write to the same summary over several weeks/days and easily spot when the run was executed. WDYT?

boldtrn:
Nice, I really like the idea. The repeat feature is a great help IMHO. Running measurements multiple times always bothered me.

Regarding the summary file, do you think it would make sense to write this as csv? This would only require to add a `,` between every value. This would make it easy to open the file with for example excel and probably easier to read. Also it might be easier to run an automatic evaluation on this (if we would want to do this).

tools/src/main/java/com/graphhopper/tools/Measurement.java
ammagamma: right now we get an 'indirect' timestamp via the .properties file name, but yeah I think adding a proper timestamp is a good idea. will do.

ammagamma:
> make sense to write this as csv?
I just checked and one can specify the [separator for gnuplot](http://gnuplot.sourceforge.net/docs_4.2/node173.html) also using `set datafile separator ","`.

ammagamma:
It looks like this now:

    #graph.nodes, graph.edges, measurement.seed, prepare.ch.time, prepare.ch.shortcuts, routing.distance_mean, routing.mean, routing.visited_nodes_mean, routingCH.distance_mean, routingCH.mean, routingCH.visited_nodes_mean, measurement.timestamp, propertyFile
           32678,       48788,              123,            1875,                24389,              14566.23,         4.27,                    5827.71,                14021.56,           1.16,                       111.69,   2018-05-10_11_33_42, measurement2018-05-10_11_33_42.properties

So it is both human readable and can be parsed like a csv file.

ammagamma:
Slightly off-topic: Why are routing.distance_mean and routingCH.distance_mean different ?

karussell:
Looks good to me

btw: how are you currently executing Measurement without #1351 :) ?

> Slightly off-topic: Why are routing.distance_mean and routingCH.distance_mean different ?

The request count is different and so it is not using the same total distance

ammagamma:
> btw: how are you currently executing Measurement without #1351 :) ?

I am simply starting the main either using run configuration in IntelliJ or using `java -cp <tools.jar>` with these or similar command line arguments:
`datareader.file=europe_germany.pbf measurement.clean=true measurement.summaryfile=summary.dat measurement.repeats=5 prepare.ch.weightings=fastest prepare.lm.weightings=no graph.flag_encoders=car prepare.min_network_size=10000 prepare.min_oneway_network_size=10000 -Xmx=10g -Xms=10g`

> The request count is different and so it is not using the same total distance

ok right

tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -510,4 +528,79 @@ private void store(FileWriter fileWriter, String comment) throws IOException {
...
+                CH.PREPARE + "shortcuts",
+                "routing.distance_mean",
+                "routing.mean",
+                "routing.visited_nodes_mean",
+                "routingCH.distance_mean",
+                "routingCH.mean",
+                "routingCH.visited_nodes_mean",
+                "measurement.timestamp"
+        };
+        try {
```
karussell: Can we use `try (FileWriter writer = new FileWriter)` to use automated writer.close call?

ammagamma: sure

ammagamma:
build fails because of a missing api key ?

karussell:
If you update to master this should not happen anymore. The API key was likely misused and so I changed it and hide it now: https://github.com/graphhopper/graphhopper/commit/50e975bd079fe22058052574fb8e82c3e83c27e9

karussell:
Hmmh, ugly. I should additionally exclude the client-hc from the build as otherwise the build does not run through locally without the key parameter

ammagamma:
Hmm still getting error with most recent master merged in (80d345ad8b154d3b28f9deb837b2816ce115bae2)

``` bash
testSimpleRoute(com.graphhopper.api.GraphHopperWebIT)  Time elapsed: 0.006 sec  <<< ERROR!
java.lang.IllegalStateException: Key cannot be empty
	at com.graphhopper.api.GraphHopperWebIT.setUp(GraphHopperWebIT.java:38)
```

karussell:
It looks like $API_KEY stays empty whereas in the master branch it is ok. Ah, this is the reason:

> Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository

karussell:
Can you try this fix: https://github.com/graphhopper/graphhopper/commit/c02ee49cbc366ba6e27f7c0673fccde88d787723?

ammagamma:
appveyor built successfully, but travis seems to be running forever (>3h) ?

karussell:
Again my fault - but good news is that I was able to push the fix directly to your branch - got the git magic right ;)

PR #1367, opened 2018-05-10
Does some cleanup of contraction hierarchies and prepares for turn costs.

ammagamma:
In this PR some refactoring is done for the contraction hierarchies graph preparation. Some of these changes are preliminary actions to support turncosts for contraction hierarchies. This is a follow up of #1366.

- node priorities are now represented by floats instead of ints and GHTreeMapComposed is adjusted accordingly. This should not really change anything, but is advantageous for edge-based CH.

- IgnoreNodeFilter is moved to a separate class

- the mean degree estimation used to limit the maximum number of visited nodes for the local Dijkstras in CH and the node priority calculation is now done in NodeContractor instead of in PrepareContractionHierarchies

- usage of stopwatches in PrepareContractionHierarchies is cleaned up and the statistics logging during the contraction is somewhat improved

- NodeContractor is replaced by a new interface that right now only has one implementation NodeBasedNodeContractor. Some methods that are also useful for the edged-based case are moved to an abstract parent class AbstractNodeContractor.

- BaseGraph has a new method debugPrint()

I checked the performance with the measurement and could not find any performance regression.


core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
```diff
@@ -397,97 +367,58 @@ public long getDijkstraCount() {
         return nodeContractor.getDijkstraCount();
     }
 
-    public int getShortcuts() {
+    public long getShortcuts() {
```
karussell: oh, did you already reached this limit?

core/src/main/java/com/graphhopper/coll/GHTreeMapComposed.java
```diff
@@ -34,39 +35,60 @@
...
+                    int key1 = bitUtil.getIntLow(o1);
+                    int key2 = bitUtil.getIntLow(o2);
+                    if (key1 == key2) return 0;
+                    return key1 < key2 ? -1 : 1;
+                }
+                float f1 = Float.intBitsToFloat(value1);
+                float f2 = Float.intBitsToFloat(value2);
+                return Float.compare(f1, f2);
+            }
+        });
```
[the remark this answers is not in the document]
ammagamma: Yes, I tried using some fixed precision, but this turned out to be rather instable, because depending on the metric one uses its hard to know the required range beforehand and sometimes the values become very small or very large. I still think it should be possible, but then again I do not think switching to float hurts and its a lot easier to work with.

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
ammagamma: No, luckily not, but I think its better to be prepared ;)

ammagamma:
Yes, I will update and check the performance again.

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
karussell: Ah, ok. 

btw: before shortcuts I expect that the edgeIds count will introduce a problem ;( so far we are at 250 million edges and <200 million shortscuts which is ~4x away from the unsigned int limit...

ammagamma: Ok. For edge-based CH the number of shortcuts is typically larger than the number of edges, I am still working on reducing the number of shortcuts, but at the moment there are about 3x as many shortcuts than edges for Germany (30 vs 10million). Note that the shortcuts are (so far) unidirected while most of the edges are bidirectional.

ammagamma: But you are right that the `getShortcuts()` method is only a simple counter while if this overflows there will be other problems too (edge ids are integers).

ammagamma:
> Is this change neutral performance-wise?

I ran the measurement 15 times for master and this branch for Germany. I got:
master: 519+-5s (about 5min40s)
this:      529+-7s (about 5min50s)
and on average master is about 1.5% faster. Query times were equal within error bounds and the number of shortcuts was exactly the same on every run.
The difference in CH preparation time might be because the custom comparator in `GHTreeMapComposed` is slower (two float comparisons instead of one long), but I did not further investigate, because the difference seems rather small to me. What do you think ?

PR #1376, opened 2018-05-21
Improvement of landmark weight precision, issue 1080

### LINKED ISSUE #1080: more precise landmark routing
Currently we use 1 short to store the forward and 1 for the backward weight of the landmarks heuristic. We could improve precision if we would store the forward weight and the delta instead and then use e.g. 20 bits for the forward weight and only 12 for the delta (as the delta is usually small). This way we should be able to avoid most of the 'too many weights were maxed out' warnings for world wide preparation.

Related to #969

baumboi:
I changed the storage of landmark heuristics from two shorts (one for the forward and one for the backward weight) to one integer with 18 bits storing the backward weight and 14 bits storing the difference between theese values. Therefore the maximum value which can be stored rises from 65534 (=2^16-2) to 262142 (=2^18-2). You can also use other bit combinations like 20/12 which would allow to store values of up to 1048574 (=2^20-2) at the expense of reducing the range of the delta value.
However, the new getToWeight function may be slightly slower since it has to get both, the backward and delta weight, and add them together.

I also added some tests to prove the correctness of the landmark storage. Please note that testSetGetWeight and testDelta will only work with the current bit combination 18/14 and need to be edited if you change it.

Fixes #1080


core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -797,25 +826,31 @@ public boolean apply(int nodeId, SPTEntry value) {
...
-                        + ". For example use the following in the config.properties: weighting=" + weighting.getName() + "|maximum=" + finalMaxWeight.getValue() * 1.2);
+                if (from){
+                    LOGGER.warn("landmark " + lmIdx + " (" + nodeAccess.getLatitude(lmNodeId) + "," + nodeAccess.getLongitude(lmNodeId) + "): " +
+                            "too many backward weights were maxed out (" + maxedout.get() + "/" + map.size() + "). Use a bigger factor than " + lms.factor
+                            + ". For example use the following in the config.properties: weighting=" + weighting.getName() + "|maximum=" + finalMaxWeight.getValue() * 1.2);
+                }else{
+                    LOGGER.warn("landmark " + lmIdx + " (" + nodeAccess.getLatitude(lmNodeId) + "," + nodeAccess.getLongitude(lmNodeId) + "): " +
+                            "too many delta weights were maxed out (" + maxedout.get() + "/" + map.size() + "). Use a bigger factor than " + lms.factor
+                            + ". For example use the following in the config.properties: weighting=" + weighting.getName() + "|maximum=" + finalMaxWeight.getValue() * 1.2);
+                }
```
karussell: Here probably it can make sense to use the short "if-then-else" form within the string like

LOGGER.warn("blup" + (from? "too many backward weights" : "too many delta weights")

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -497,40 +494,72 @@ int getFromWeight(int landmarkIndex, int node) {
...
     int getToWeight(int landmarkIndex, int node) {
-        int res = (int) landmarkWeightDA.getShort((long) node * LM_ROW_LENGTH + landmarkIndex * 4 + TO_OFFSET)
-                & 0x0000FFFF;
-        assert res >= 0 : "Negative to weight " + res + ", landmark index:" + landmarkIndex + ", node:" + node;
-        if (res == SHORT_INFINITY)
-            return SHORT_MAX;
+        int res = landmarkWeightDA.getInt((long) node * LM_ROW_LENGTH + landmarkIndex * 4);
+
+        //the right bits of "res" store the backward value
+        int from = res & FROM_WEIGHT_INF;
```
karussell: This statement could be moved after the if to avoid this in case of the specific delta (?)

core/src/test/java/com/graphhopper/routing/lm/LandmarkStorageTest.java
```diff
@@ -228,4 +240,66 @@ public BBox getBounds() {
...
+
+        ghStorage.edge(3, 10, distance, true);
+        ghStorage.edge(10, 11, distance, true);
+        ghStorage.edge(11, 12, distance, true);
+        ghStorage.edge(12, 13, distance, true);
+        ghStorage.edge(13, 14, distance, true);
+        ghStorage.edge(14, 15, distance, true);
+
+        LandmarkStorage storage = new LandmarkStorage(ghStorage, new RAMDirectory(), new FastestWeighting(encoder), 2);
+        storage.createLandmarks(); //landmarks: 15, 9
```
karussell: Instead of a comment it would be better to assert the actual landmark node ids (?)

core/src/test/java/com/graphhopper/routing/lm/LandmarkStorageTest.java
```diff
@@ -228,4 +240,66 @@ public BBox getBounds() {
...
+        ghStorage.edge(1, 7, distance, true);
+        ghStorage.edge(7, 8, distance, true);
+        ghStorage.edge(8, 9, distance, true);
+
+        ghStorage.edge(6, 10, distance, true);
+        ghStorage.edge(10, 11, distance, true);
+        ghStorage.edge(11, 12, distance, true);
+
+        LandmarkStorage storage = new LandmarkStorage(ghStorage, new RAMDirectory(), new FastestWeighting(encoder), 2);
+        storage.createLandmarks(); //landmarks: 12, 9
```
karussell: Here we should assert something :)

The logging statement cannot be checked automatically but maybe the maxed out value and the actual landmark nodeIDs?

core/src/test/java/com/graphhopper/routing/lm/LandmarkStorageTest.java
```diff
@@ -89,21 +89,33 @@ public void testSetGetWeight() {
...
+        2^18 = 262144, use -1 for infinity and -2 for maximum
+        As the backward weight reaches a too high value it will use the maximum instead of 0 or infinity*/
+        lms.setWeight(0, 0, 16, Math.pow(2,18), true);
+        assertEquals((int) Math.pow(2,18) - 2, lms.getFromWeight(0, 0));
+        lms.setWeight(0, 0, 16, 999999, true);
+        assertEquals((int) Math.pow(2,18) - 2, lms.getFromWeight(0, 0));
+
+        /* FROM_WEIGHT_BITS = 18 --> remaining bits: 32-18 = 14
+        The delta value is signed and will therefore go from -2^(14-1) to 2^(14-1).
+        Now 2^13-1 is used for infinity, 2^13-2 as maximum and -2^13 as minimum.
```
karussell: Ah, thanks for the reminder that delta can be pos+neg unlike the forward value :+1:

baumboi:
I updated the files at the lines you marked :)

> Did you test the performance and average precision improvement e.g. for German area?

Hm I don't really know how to test this

baumboi:
I ran both versions twice for Germany and got these results:

old:
prepare.lm.time=342045
prepare.lm.time=343976
routingLM4.mean=78.3398418488
routingLM4.mean=75.99576999119999

new:
prepare.lm.time=355327
prepare.lm.time=350337
routingLM4.mean=67.7551219
routingLM4.mean=68.0320597584

This makes on average 343 s for the old version and 352 s for the new one to prepare the landmarks, which is about 3% slower. However, the routing is about 12% faster, taking only 68 ms instead on 77 ms.

karussell:
This sounds really good. Will try. Why do you think could it be over 10% faster? Because instead of fetching two `shorts` from DataAccess we fetch just one `int`?

baumboi:
> This sounds really good. Will try. Why do you think could it be over 10% faster? Because instead of fetching two shorts from DataAccess we fetch just one int?

I'm not sure if this is the reason, because the new version also fetches two `int`s (1 for `getFromWeight` and 1 for `getToWeight` seperatly). Maybe the convertation from `short` to `int` takes some time and fetching an `int` is faster than fetching a `short`. Both functions have pretty much the same operations as before, `getToWeight` is even a bit longer, so I'm surprised that it's such a big improvement.

boldtrn:
> Why do you think could it be over 10% faster?

Is this improvement reproducible? Could be just a random anomaly?

ammagamma:
> Is this improvement reproducible? Could be just a random anomaly?

You can run the measurement multiple (maybe 10 or so) times and print a table of the results with the new flags in the measurement class: https://github.com/graphhopper/graphhopper/pull/1366 to get a better picture.

baumboi:
I ran both another 8 times with these results:

old:
routingLM4.mean=79.8826709984
routingLM4.mean=79.7894196064
routingLM4.mean=72.8586727184
routingLM4.mean=82.17932466479999
routingLM4.mean=79.23005568800001
routingLM4.mean=80.3595968472
routingLM4.mean=78.9029193504
routingLM4.mean=71.5291511544

new:
routingLM4.mean=76.31359076640001
routingLM4.mean=64.7755447112
routingLM4.mean=81.274656928
routingLM4.mean=81.1376535216
routingLM4.mean=72.4896060752
routingLM4.mean=70.0979023784
routingLM4.mean=72.3667225016
routingLM4.mean=70.8192831368

Now the values vary much more than before. On average I got 78 ms for the old and 74 ms for the new which makes a (much more realistic) improvement of about 5%. Looks like I had two lucky rounds before^^

karussell:
I merged the master but it seems that also before the merge the tests were failing. Can you run the tests and see what went wrong? (Either command line `mvn clean test` or IntelliJ->select all modules->right click->Run all tests, the failing test class is com.graphhopper.routing.RoutingAlgorithmWithOSMIT)

karussell:
Sorry this took a bit too long here to merge, but we really needed to test this out on a large scale setup and compare this to the very recent versions.

Now, this is a massive improvement in terms of visited nodes and therefore speed and maximum possible travel distance for continental requests with ch.disable=true.

I'll try to merge this fast ... I'll just include a version number so that previous data imports are not possible to load.

karussell:
@baumboi do you have an updated version where the tests run through?

karussell:
Thanks a lot!

> I don't really know why this is the case. It works perfectly in my repository in IntelliJ

Will fix in this branch.

karussell:
I have now fixed the compiling issue, but it seems it creates much more of those warnings for the test cases now with the limitation of the factor to 0.1

Couldn't we increase the PRECISION now to 18 or even more bits?

karussell:
The warning regarding "maxed out" values is somehow wrong.

Is the following correct?

 * for maxed out `delta` values one should increase the maximum to make deltas smaller
 * for maxed out `from` values one should decrease the maximum (i.e. increase the factor)

Maybe this is also the reason PRECISION can stay at 16.

karussell:
> but have at least as many weights maxed out as before (probably even more because of the quite small size of the delta values)

Thanks, makes sense!

> I can't really say which of these options will perform better, but i could do some tests (e.g. for Germany)

No problem. Currently doing a world wide import again and will report about PRECISION=16

> I think for both cases we need to decrease the precision (--> increasing the factor), in order to reduce the value we store.

Thanks for the clarification, will fix this in the code

boldtrn:
Thanks for the nice improvement @baumboi. 

> No problem. Currently doing a world wide import again and will report about PRECISION=16

Did you find out anything about the precision @karussell?

If I understand this correctly, there is no need to adjust an already set maximum in the config? But due to the increased precision the A* heuristic will work better and also this will reduce the Likelihood of rounding errors when two almost similar weights are stored, because a better precision and lower factor?

karussell:
Yes, 16 works better in the sense that we have less maxed out weights leading to a performance improvement as less nodes will be visited due to the improved heuristic

boldtrn:
Thanks for reporting your findings :)

Issue #1137, opened 2017-08-09
avoid manual parsing in GraphHopperWeb


karussell:
Make it possible to just write:

GHResponse res = objectMapper.readValue(str, GHResponse.class);

and this will also simplify or remove the code in SimpleRouteSerializer.

(We need further custom serialize code for the polyline encoding stuff)

michaz:
What is the general plan for the client? [Elsewhere](https://github.com/graphhopper/graphhopper/issues/1104) I was suggesting to remove it, but only because it's duplicated in another project.

In general, it can of course be a good idea to have the (same-language) client in the same repository, precisely because serialization code and representations can be shared. Dropwizard recommends it. It should go into a different JAR, of course, so rather than being in "web", it should be in a "client" module. Client code can then include the client JAR and that's it..

So maybe we move the "hand-coded" client over here? In general?

karussell:
Yes, this might be a good solution to move the hand-coded into a new `client` maven module. The only problem is that we would either have to split the hand-coded matrix part and keep it externally or put it here, which would mix the open source with the API which I would like to avoid to reduce confusion.

karussell:
> which would mix the open source with the API which I would like to avoid to reduce confusion.

Hmmh, this is kind of 'violated' already as the GraphHopperWeb points to the commercial offering per default. So moving the hc client into this repo sounds good to me and we can finally avoid duplicate code.

michaz:
Related: What is the plan for the Swagger spec? Is it just a side-project, or do we want to make it more authoritative by reviewing it, adding documentation and perhaps eventually generate the official API docs from it?

karussell:
> Is it just a side-project

Definitely not. Why do you think so?

> Related: What is the plan for the Swagger spec?

The swagger spec and clients generated are already working. Also there is a bit documentation for this already inside the spec. But a few bigger issues remain (limitation of either the codegen or swagger spec itself) like https://github.com/graphhopper/directions-api-clients/issues/1

michaz:
> Definitely not. Why do you think so?

Just asking. For example, it does not generate the official API documentation, but duplicates it.

michaz:
> So moving the hc client into this repo sounds good to me and we can finally avoid duplicate code.

I could do that on the Dropwizard-PR. Would fit nicely there because of close connections to "doing more with Jackson" and "sharing representations".

karussell:
[...]

PR #1378, opened 2018-05-25
Split off api

michaz:
This does three things about sub-modules of which 1.) and 2.) are necessary to cleanly finish the across-repository Dropwizardification, and 3.) happened on the way:

1.) Split off ``web-bundle`` from ``web``. This is minor and probably nobody will notice. Problem is that ``web`` builds into a shaded JAR (of which we don't use the shading), and such JARs are ugly to import from other projects, but we want to import the web services from other projects (to combine them to a single server), so the web services must go out of the module in which they are tied together.

2.) Split off ``api`` from ``core``. This is, I think, a major win, also for our mobile platforms, because it means that the ``client-hc`` now doesn't pull in the entire ``core`` as a dependency, which it was doing before. I simply moved all the things needed by the client _and_ the server _and_ the core, transitively, until I could remove ``core`` from ``client-hc``.

3.) Then, I also started moving some of the serialization and deserialization code to ``api``, see #1137.   But doing so requires a Jackson dependency in ``api``, which means we now also have it (transitively) in ``core``, in contrast to what the latest discussion in #1291 says. Read on:

- It is _possible_ to _not_ put serialization / deserialization code into an ``api`` module. But then, such code would have to go into _yet another_ module. Not framing this as "what dependencies do I need" but as "what does the module specify", it would mean that ``api`` contains the Java data classes and ``representation`` (or something) contains the wire format (JSON), of which there could be several. Again, this is _possible_, but the question is whether that would be practical, or maybe a bit purist. Actually, since GraphHopper is both a library and a REST service, it makes sense to say that our Java objects and our JSON representations are 1:1 and should be close together, see below.

- Allowing _on top of that_ the JSON _library_ to be replaced would be totally enterprisey and over-engineered. We use Dropwizard, and Dropwizard uses Jackson, and the whole idea is to remove technical/boring/distracting stuff like serialization from code that is supposed to be about routing, and in the age of frameworks, that is made _harder_ by wrapping it in custom interfaces, not easier.

- We are still not using JSON (de-)serialization in the ``core``, so we _can_ split off the additional ``representation`` module any time if we must. Actually, this was totally easy to do, without the custom JSON facade: ``GraphHopper`` core methods can just be passed deserialized objects like ``JsonFeatureCollection``, not JSON maps or filenames. Voilà. This is the start of another (different) discussion: Is GraphHopper ``core`` a _library_, or is it supposed to be exactly like ``web`` but without the web?
  - If the former, then it can stay like this: the user of the library can just de-serialize the JSON themselves (with a single call to Jackson, or whatever), or build the object in code, and pass the object, that's how pure libraries are supposed to be, and we can remove the Jackson dependency from ``api`` if we want, and from then on, ``GraphHopper`` only knows about Java objects, not about JSON or strings or files (except for ``GraphHopperStorage``).
  - If the latter, we can just close the deal and say, fine, ``core`` includes Jackson now, and it can read JSON files by itself, and that's it.


karussell:
Thanks! Looks like a huge change :)

I saw you moved some helper code into the memory mapping file, @devemux86 do you think this works with all the different Android version? 

> which means we now also have it (transitively) in core, in contrast to what the latest discussion in #1291 says

Thanks for explaining this in-depth. This decision also makes sense and let's separate if there is a problem or if we have something else than JSON like pbf, kryo or whatever.

> Is GraphHopper core a library, or is it supposed to be exactly like web but without the web?

I think more a library but at the same time when doing some work for #1112 I was in the need to serialize EncodedValues as "something" and if we use JSON there everything is easier compared to string or properties but then the core&storage really requires JSON. Probably the best would be if the Java library contains JSON at some point :)

michaz:
> and let's separate if there is a problem or if we have something else than JSON like pbf, kryo or whatever.

Exactly! I think a good model for a Java-library-which-is-also-a-REST-service which is allowed to evolve is: At each point in time, there is _one_ mapping of Things-which-go-over-the-internet to Java objects which we think is _current_ and the _best_, and the code for _that_ one goes directly next to the Java objects. Things for "other formats" or "backward-compatibility" or "developer-sugar" go somewhere else, in pre-processors or filters or converters or something.

devemux86:
@karussell apparently you mean [this](https://github.com/graphhopper/graphhopper/pull/1378/files#diff-71).
If that code is accessed again with the same way probably would work on Android, but @michaz can I ask why the change?
Since it involves utility methods which handle special cases, would be preferred in some Helper class.
If need to be only in core, then can create a new Helper class there?

Regarding json, if I may express an opinion and if it's possible in the API, I'd vote for a separate module, so that core can remain pure Java.
i.e. with offline graphs there is no need for json (?), with only Java API the routing is fine.

michaz:
> If that code is accessed again with the same way probably would work on Android, but @michaz can I ask why the change?

@devemux86 Because it was used by ``StorableProperties`` and _only_ by ``StorableProperties`` (and its test), so I figured it should go there. Maybe I don't get the rationale of the ``Helper`` class -- isn't it that things which _don't_ work on Android should _not_ go there but into ``Helper7``?

devemux86:
No it's not a binary selection, Android or not. Even in Android unfortunately there are several paths in those methods, depending the Android version. :slightly_frowning_face:

michaz:
Sorry, I'm new to this, I have no idea what this is about. :-) I just see a class that's called ``Helper`` (pretty innocent name for something which is about preventing breaking platform compatibility :-) ), and I saw a method which is only used in one other place, and moved it to that place.

Why doesn't that work and what should I pay attention to?

devemux86:
I didn't say it won't work!
Without testing GraphHopper Android sample on Android Emulator with various Android versions, cannot be sure for anything. :slightly_smiling_face:

michaz:
Aha! So ``Helper`` is defined as "where things that maybe cause problems on Android go", correct? Got it.

But that's just for.. information purposes, right? I mean, suppose it works with Android, then ``StorableProperties`` works with Android. Suppose it doesn't, then it doesn't. But there's no automagic going on which replaces it with an alternative that _does_ work on Android, right?

So either way, it will be as before. :-)

karussell:
Sorry, maybe also my fault because Helper7 is indeed for the compatibility issues and not Helper. So let's try .... and sorry for the bubbles.

michaz:
Arrrrrrgh! :-)

Issue #1380, opened 2018-05-27
Reduce usages of DefaultEdgeFilter(encoder, true, false)


ammagamma:
We often do something like 
``` java
graph.createEdgeExplorer(new DefaultEdgeFilter(encoder, false, true));
```
(about 35 times in the current master). I've done this myself quite a few times and everytime I need to check if I got the flags right. Therefore I would suggest to add two or three static factory methods to `DefaultEdgeFilter`: `DefaultEdgeFilter.inEdges(FlagEncoder)`, `DefaultEdgeFilter.outEdges(FlagEncoder)` and optionally `DefaultEdgeFilter.allEdges(FlagEncoder)`. The last one is what the default constructor does, but would be more readable in my opinion.

karussell:
Yes, please - this is a good idea. This would also solve the ugliness in #1112 where I swapped the booleans (as fwd,bwd is more intuitive) but this would brake all usages without being noticed.

PR #1381, opened 2018-05-27
Reduce usages of fwd/bwd flags in DefaultEdgeFilter constructor

ammagamma:
#1380 

Replacing the constructor calls in other branches can be done quite easily using this regex search&replace in IntelliJ: 
replace

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
+++ b/core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
@@ -27,18 +27,18 @@ public class DefaultEdgeFilter implements EdgeFilter {
     private final boolean fwd;
     private FlagEncoder encoder;
 
-    protected DefaultEdgeFilter(FlagEncoder encoder, boolean bwd, boolean fwd) {
+    protected DefaultEdgeFilter(FlagEncoder encoder, boolean fwd, boolean bwd) {
         this.encoder = encoder;
         this.bwd = bwd;
         this.fwd = fwd;
     }
 
     public static DefaultEdgeFilter outEdges(FlagEncoder flagEncoder) {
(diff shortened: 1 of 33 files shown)


ammagamma:
Should we replace the default constructor calls with `DefaultEdgeFilter.allEdges` as well ? Maybe there should be a better name for it to distinguish it from `EdgeFilter.ALL_EDGES` ?

core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
```diff
@@ -40,6 +40,18 @@ public DefaultEdgeFilter(FlagEncoder encoder, boolean bwd, boolean fwd) {
         this.fwd = fwd;
     }
```
karussell: Should we remove this or even all constructors? IMO the order bwd+fwd is unintuitive

karussell:
> Should we replace the default constructor calls with DefaultEdgeFilter.allEdges as well ? 

IMO yes, so we have no mixed constructor+factory usage. At least the bwd+fwd constructor

> Maybe there should be a better name for it to distinguish it from EdgeFilter.ALL_EDGES ?

IMO this is fine - or maybe add a javadoc comment what it does?

core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
ammagamma: I removed the default constructor and made the other one protected. We could make it private, too and instead add a getter for the flag encoder (this is currently needed for the PrepEdgeFilter in PrepareRoutingSubnetworks).

ammagamma: Changed the oder bwd+fwd too

Issue #1389, opened 2018-06-01
WebHelper.errorResponse (currently: RouteResource.errorResponse())


Nakaner:
This issue refers to the current master branch (with all the changes related to Dropwizard).

I don't fully understand the thoughts behind recent moves of 
`WebHelper.errorResponse` (currently: `RouteResource.errorResponse()`).

@michaz moved this and other static helper method from the `RouteResource` to the `WebHelper` class in commit 69cfd03 because "map matching re-uses them" (quote of the commit message). A few weeks later he moved them back with the comment "Move web service helper methods back to route resource; clean dependencies" in commit 9e2765c and made the method private.

This breaks the map matching module. He tried to fix it in commit [4bee777](https://github.com/graphhopper/map-matching/commit/4bee7776611025a90b789f2a387a0a61f9cb164a) of the map matching module but that fix makes map-matching only to compile properly but it is still broken. (That's not covered by a unit test ;-)) If you send broken GPX data to the map matching API and request a XML response, you will get your error message in JSON format but valid input data would result in a valid XML response.

The reason: Yesterday the public method `com.graphhopper.http.WebHelper.errorResponse()` (part of graphhopper-web-api) was moved to `com.graphhopper.resources.RouteResource.errorResponse()` in commit [9e2765c](https://github.com/graphhopper/graphhopper/commit/9e2765c3e7233a963bb1f8da8afdfb46f939417d) and made private. This change reverted the changes of commit [69cfd03](https://github.com/graphhopper/graphhopper/commit/69cfd035b0a2255f074056016b7d245b246c0435) which moved it in the opposite direction.

I think that such a method is useful for anyone who wants to build a GraphHopper like API – be it just the map matching module or an obscure routing engine. :)

michaz:
I'm moving those methods back and forth because I'm looking for a compromise between no code duplication and smallest surface area. (I don't want to publish lots of public "helper" methods which I'm then not supposed to change anymore.)

That said, _all_ of those methods are refactoring artifacts. Ideally, they shouldn't exist. So no, please don't consider them public library methods which other people building things should use.

Thanks for pointing out that map-matching currently doesn't return XML formatted errors (I think I noticed that while I was doing that). Same thing here: I'm looking for something with the least amount of code, extra ``else`` branches for uncommon cases etc.

Let's be fair, the XML error output [isn't even specified](https://graphhopper.com/api/1/docs/map-matching/).

karussell:
[...]

PR #1390, opened 2018-06-02
Exception Mapping

michaz:
Based on #1389, moved the creation of exception responses to where it belongs (in Dropwizard's opinion): To an exception mapper and a message body writer, and a regular JSON serializer for the JSON case. _Those_ can easily be shared by other Repos (map-matching, matrix).

Dropwizard/Jersey want the switch for content types to be _mainly_ done based on the header, not on a query param, but other things are also possible: "re-mapping" or "normalizing" those things in a Filter is what's recommended. (Jersey has an example where the same is done for filename extensions. That's where I copied this from.)

The explanation how it works is in ``GraphHopperBundle``, where it's plugged together:

```
        // If the "?format=gpx" parameter is present, sets a corresponding media type header
        environment.jersey().register(new FormatGPXFilter());

        // Together, these two take care that MultiExceptions thrown from RouteResource
        // come out as JSON or GPX, depending on the media type
        environment.jersey().register(new MultiExceptionMapper());
        environment.jersey().register(new MultiExceptionGPXMessageBodyWriter());
```


michaz:
..and the only thing which complicates doing the same thing for the regular case (the successful response) is that they are parameterized (by "do we want instructions?" and such). Not sure how that is supposed to be done yet.

michaz:
If we like this, we can codify in the same way the GraphHopper-convention (correct me if I'm wrong) that IllegalArgumentExceptions should correspond to status 4xx. Then we don't have to wrap them in WebApplicationExceptions (as before) or MultiExceptions (as now) but can throw them directly, even from the library (if it's only one).

Nakaner:
I lack the in-depth knowledge of Dropwizard etc. to review the code but I checked out your branch and tested it.

Is it intended that the `type` parameter of the query string is ignored and only the HTTP `Accept` header field (defaults to JSON) matters? Valid requests would return GPX data if `type=gpx` is represent but no `Accept` header field.

`curl  -H "Accept: application/gpx+xml" "localhost:8989/route?point=40.0,30.0&point=20.0,30.0&vehicle=car&format=gpx"` results in an XML error response but if I omit the HTTP `Accept` header field, JSON is returned.

The HTTP API documentation don't mention at all that you need to send a header.

michaz:
> Is it intended that the type parameter of the query string is ignored

Err.. no. Hold on..

michaz:
Ahh, look, it's supposed to be `&type=gpx`, not `&format=gpx`. Then it works.

Were you primed by my class name ``FormatGPXFilter`` maybe? No idea why I called it that, will change.

karussell:
> IllegalArgumentExceptions should correspond to status 400

:+1:

PR #1406, opened 2018-06-20
Made point hint matching stricter and avoid problems for short street names

### LINKED ISSUE #1404: point hint problem
The snapping due to the point hint with [this original request](https://graphhopper.com/maps/?point=52.50977,13.371971&point=52.509842,13.369761&point_hint=Ben%20Gurion&point_hint=)  does not work, but without white spaces [it works](https://graphhopper.com/maps/?point=52.50977,13.371971&point=52.509842,13.369761&point_hint=BenGurion&point_hint=)

If there is a space the too short name "ben" will be removed. If it is a dash, the dash will be removed resulting in the necessary "bengurion".

karussell:
The biggest change is that with this PR we require now the input to be just the street and not the full address to improve matching. Was able to increase the JARO_WINKLER_ACCEPT_FACTOR from under 0.8 to 0.9

Also it:

 * Fixes #1404
 * ignores words with length<2 instead length<3 and a certain set of "street" words
 * does not ignore words that could make the difference like parts of the street ("ben")
 * introduces a rewriteMap so that the filter can be improved for other cases and languages without a code change
 * a clean up and include some ignored tests
 * made JaroWinkler instance static (tiny memory usage reduction)


boldtrn:
> we require now the input to be just the street and not the full address to improve matching

This is a really good change! It was quite hard to get the matching right in some cases with full addresses.

core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -46,34 +48,56 @@
...
+        put("avenue", "");
+        put("avenida", "");
+        put("rd.", "");
+        put("road", "");
+        put("str.", "");
+        put("str", "");
+        put("straße", "");
+        put("strasse", "");
+        put("st.", ""); // saint vs. street
+        put("street", "");
```
boldtrn: Rewriting to an empty string could be problematic in some more exotic cases. But I think the alternatives also introduce potential errors, so this is probably the best approach.

I miss a couple of strings here (not sure if they get abbreviated though, so maybe it's better not to add them?):
strada (e.g. Strada Provinciale 209)
via (e.g. Via Fosso Ancaiano)
calle (Calle el Dr. Mabuse)
plaza (Plaza de la Bámbola)

Some English ones:
lane
ln. (short form of lane)
place
pl.
track
tr.
pathway
path
crescend
cres
cres.
parade
pde

I found a great resource for street abbreviations in English [here](http://meteor.aihw.gov.au/content/index.phtml/itemId/270020). An even longer list can be found [here](http://meteor.aihw.gov.au/content/index.phtml/itemId/429387).

karussell: Yes :)
We could even extract them from OSM:
https://github.com/komoot/photon/pull/46/files#diff-751e12c3a9da7a1987a479baab0ff97c
https://github.com/karussell/languator

But probably we should not overcomplicate things here and improve this just a bit.

Another question is how we handle those common words if they are part of the street name like with "Hauptstraße" we can't just remove them because e.g. `st` could be a part of a street name without the street-meaning

boldtrn: > Another question is how we handle those common words if they are part of the street name like with "Hauptstraße" we can't just remove them because e.g. st could be a part of a street name without the street-meaning

Yes I think we cannot really do anything about these without introducing bigger errors as well.

> But probably we should not overcomplicate things here and improve this just a bit.

I know that English speakers love abbreviations, so I think fixing the English abbreviations is quite important. The names I mentioned above are pretty common, the ones in the first link are quite common as well, the second link contains more uncommon ones as well. But Google Maps loves to abbreviate them in english speaking countries as well, most abbreviations are rather short like Drive and Dr, so the dr will be removed anyway. So I think if anyone tries to combine the point_hint with GMaps Geocoding, this will become an issue.

core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -46,34 +45,61 @@
...
+                "av.", "avenue", "avenida",
+                "calle",
+                "rd.", "road",
+                "ln.", "lane",
+                "pl.", "place", "plaza",
+                "str.", "str", "straße", "strasse", "st.", "street", "strada",
+                "sq.", "square",
+                "tr.", "track",
+                "via")) {
+            put(remove, "");
```
karussell: Do you think this is now sufficient and we could add new stuff if required later? I think adding all these words does not necessarily improve the result. We probably could try an automated way to find areas where our snapping is not optimal now. E.g. pick a location and 4 close streets and pick one street to see if the others get rejected or similar.

boldtrn: Ok sure, we can always add more later and see how we go :). 

Personally, I would add:
crescend, cres, cres., parade, pde, pde., boulevard, bvd, bvd.

These are really common ones. In GB there are many Close, Cl, which could make sense to add as well. Probably we just need to see how we go.

karussell: Have added a few as requested :)

PR #1407, opened 2018-06-20
a new IllegalArgumentException mapper

karussell:
Implemented a mapping from an IllegalArgumentException to the usual error JSON (or GPX) after a suggestion from @michaz in #1390. This PR

 * makes throwing Bad requests in resources simpler
 * should fix downstream exception handling: https://github.com/graphhopper/map-matching/issues/119
 * and the EncodingManager no longer necessary in RouteResource

We cannot throw away the MultiException (as I understand it) because of potential other cases when we want an error code != 400 and also because we use it in one place where we collect multiple message which is not possible with an IllegalArgumentException.


core/files/findbugs-exclude.xml
```diff
@@ -2,4 +2,9 @@
 <FindBugsFilter>
     <!-- documentation about findbug's rule match clauses can be found here:
          http://findbugs.sourceforge.net/manual/filter.html -->
+         
+  <Match>
+    <Class name="com.graphhopper.resources.IsochroneResource" />
+    <Bug pattern="RV_EXCEPTION_NOT_THROWN" />
+  </Match>
```
karussell: Somehow this is necessary as otherwise it fails and says:

```
[INFO] Total bugs: 1
[INFO] new IllegalArgumentException(String) not thrown in com.graphhopper.resources.IsochroneResource.doGet(HttpServletRequest, UriInfo, String, String, int, boolean, GHPoint, String, long, double) [com.graphhopper.resources.IsochroneResource, com.graphhopper.resources.IsochroneResource, com.graphhopper.resources.IsochroneResource, com.graphhopper.resources.IsochroneResource, com.graphhopper.resources.IsochroneResource, com.graphhopper.resources.IsochroneResource, com.graphhopper.resources.IsochroneResource, com.graphhopper.resources.IsochroneResource, com.graphhopper.resources.IsochroneResource, com.graphhopper.resources.IsochroneResource] At IsochroneResource.java:[line 62]Another occurrence at IsochroneResource.java:[line 65]Another occurrence at IsochroneResource.java:[line 70]Another occurrence at IsochroneResource.java:[line 77]Another occurrence at IsochroneResource.java:[line 92]Another occurrence at IsochroneResource.java:[line 94]Another occurrence at IsochroneResource.java:[line 101]Another occurrence at IsochroneResource.java:[line 103]Another occurrence at IsochroneResource.java:[line 110]Another occurrence at IsochroneResource.java:[line 116] RV_EXCEPTION_NOT_THROWN
```

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -977,8 +977,7 @@ public GHResponse route(GHRequest request) {
         readLock.lock();
         try {
             if (!encodingManager.supports(vehicle))
-                throw new IllegalArgumentException("Vehicle " + vehicle + " unsupported. "
-                        + "Supported are: " + getEncodingManager());
+                throw new IllegalArgumentException("Vehicle not supported: " + vehicle);
```
devemux86: I liked when router could inform easily in that message about the supported vehicles! :slightly_smiling_face:

karussell: Done

web-bundle/src/main/java/com/graphhopper/http/GHPointConverterProvider.java
```diff
@@ -0,0 +1,57 @@
...
+    public <T> ParamConverter<T> getConverter(Class<T> rawType, Type genericType, Annotation[] annotations) {
+        if (rawType.equals(GHPoint.class)) {
+            return new ParamConverter<T>() {
+                @Override
+                public T fromString(String value) {
+                    try {
+                        return (T) GHPoint.fromString(value);
+                    } catch (IllegalArgumentException ex) {
+                        throw new WebApplicationException(Response.status(Response.Status.BAD_REQUEST)
+                                .entity(new MultiException(new Throwable("Cannot parse point '" + value + "'")))
```
michaz: Maybe we can put this error message right into the `IllegalArgumentException` that is thrown in `GHPoint.fromString` -- then we can put the original `ex` that is caught here, instead of creating a new `Throwable`?

michaz: As an afterthought, this would look a lot more general then -- maybe we could then even use it for _all_ parameters instead of only for `GHPoint`.
The automagical use of a `fromString`-method by Jersey is probably implemented in a very similar way...

karussell: Thanks, have included the first suggestion.

Issue #504, opened 2015-08-24
Routing on UK bridleways


adrianQwest:
When trying to route for bicycles, GraphHopper will not follow bridleways (highway=bridleway) even if bicycles=yes.

karussell:
This is related to the issue discussed in #469 for foot, but with explicit tagging this should work

adrianhyncica:
Peter,

How long does it take for a change on openstreetmap, a change of tag for example, to propagate to GraphHopper?

Regards,

Adrian

karussell:
Currently roughly 1 day

larkang:
For me routing with bicycle along [this bridleway](https://www.openstreetmap.org/way/35318380) does not work, even though it has "bicycle=yes", while routing by foot does work.

karussell:
Fixed in #2105.

PR #1408, opened 2018-06-21
[WIP] Allow foot and bike routing over bridleways in GB

### LINKED ISSUE #469: England/Wales public bridleway implies foot=yes
Re: http://www.openstreetmap.org/directions?engine=graphhopper_foot&route=52.62530%2C1.22115%3B52.62559%2C1.21839#map=17/52.62689/1.22147&layers=N

As per http://wiki.openstreetmap.org/wiki/Tag:highway=bridleway the pedestrian routing should permit access to highway=bridleway with designation=public_bridleway.

Alternatively generally presume foot=yes like the wiki Implies section, except where the way exists inside the country of Germany, in which case presume foot=no.

boldtrn:
Fixes #469 and #504.

I added WIP since the bike routing is not yet implemented, should be quite easy though.

For now I only decided to allow routing over bridleway in GB and not in other countries, most countries like Germany are rather restrictive about bridleways, so we should probably see how this feature evolves and maybe add another country at a later point if needed.

[Here](https://graphhopper.com/maps/?point=51.724955%2C-0.981817&point=51.734125%2C-0.996237vehicle=foot&weighting=fastest) is one example in GB.
This PR:
1408_local
GH Maps:
1408_gh

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
+++ b/core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
@@ -68,7 +68,7 @@ public interface SpatialRule {
 
         @Override
         public AccessValue getAccessValue(String highwayTag, TransportationMode transportationMode, AccessValue _default) {
-            return AccessValue.ACCESSIBLE;
+            return _default;
         }
 
         @Override
--- a/core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GreatBritainSpatialRule.java
+++ b/core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GreatBritainSpatialRule.java
@@ -1,0 +1,56 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.routing.util.spatialrules.countries;
+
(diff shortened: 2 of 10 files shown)


core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
```diff
@@ -68,7 +68,7 @@ public double getMaxSpeed(String highwayTag, double _default) {
 
         @Override
         public AccessValue getAccessValue(String highwayTag, TransportationMode transportationMode, AccessValue _default) {
-            return AccessValue.ACCESSIBLE;
+            return _default;
```
boldtrn: This is a breaking change, but the behavior before didn't make sense if you want to block something by default and only allow it for certain areas.  If there was no rule for the area, it always returned accessible and not the _default.

Will add this to NOTICES

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GreatBritainSpatialRule.java
```diff
@@ -0,0 +1,56 @@
...
+        switch (highwayTag) {
+            case "motorway":
+                return 113;
+            default:
+                return super.getMaxSpeed(highwayTag, _default);
+        }
+    }
+
+    @Override
+    public AccessValue getAccessValue(String highwayTag, TransportationMode transportationMode, AccessValue _default) {
```
boldtrn: I think this method is a bit limited, we should think about passing the ReaderWay, this would allow us to make more complex decision also using other available tags, for example: `designation=public_bridleway`, which is a relevant tag for bridleways in the UK, but not for other countries, see [here](https://wiki.openstreetmap.org/wiki/Tag:highway=bridleway).

karussell:
I think we need rewrite this to the new concepts in mind (RoadClass.BRIDLEWAY). Or should we close here?

Also it might be that the issue with the center or way_start_point is no longer a problem as the SpatialRuleParser should handle that (I hope).

boldtrn:
Maybe a full rewrite would be better than reusing this PR. I think that the bridleway situation is something we should improve. 

> Also it might be that the issue with the center or way_start_point is no longer a problem as the SpatialRuleParser should handle that (I hope).

The SpatialRuleParser should happen after the acceptWay? We should know where a bridleway is in order to decide if we can accept it. Alternatively, we could accept all bridleways at first and in handleWayTags we set it as not accessible if it's in certain countries?

karussell:
> Alternatively, we could accept all bridleways at first and in handleWayTags we set it as not accessible if it's in certain countries?

Yeah, probably in general I think we should remove the acceptWay method. I'm not sure it is really faster with it.

karussell:
> Maybe a full rewrite would be better than reusing this PR

Ok

Issue #1265, opened 2018-01-09
OSRM json format


karussell:
The mapbox SDK seems to be frequently used and is open source (as far as I can see for several languages) and so we should consider implementing an alternative json format of `/route` via a different end point like `/route-osrm` or parameter like `format=osrm`

http://project-osrm.org/docs/v5.5.1/api/

Dependency: #1108

michaz:
Would suggest a new endpoint, which then emulates the OSRM API both ways.

karussell:
Ok. What do you mean with "both ways"?

michaz:
That the endpoint's requests are also like OSRMs. Not only the responses, like you were hinting at with ``format=osrm``.

karussell:
Ah, makes sense. Now the endpoint makes also more sense, yes.

boldtrn:
When we merge #1422, should we create smaller issues per feature that we want to add to this endpoint and close this issue?

karussell:
Yes, sure.

karussell:
This will go into a separate repository https://github.com/graphhopper/graphhopper-navigation

And we won't support OSRM but only mapbox (they seem to be incompatible and OSRM seems not to work with the navigation SDK)

PR #1422, opened 2018-07-11
Basic support for Mapbox / OSRM endpoint

boldtrn:
This PR approaches #1265.

So far the implementation is still quite limited, but works well in my tests with the Mapbox SDK. This PR tries not to change anything in the GraphHopper core.

Some of the current limitations are:
- Some of the response data is faked, including, driving_side, maneuver bearings, intersection data
- The available request parameters are very limited
- Voice instructions generation is pretty simple, every instruction is only spoken once
- Banner instruction generation is pretty simple

Let me know what you think about the current state :).


web-api/src/main/java/com/graphhopper/http/WebHelper.java
```diff
@@ -102,16 +102,20 @@ public static String encodePolyline(PointList poly) {
     }
 
     public static String encodePolyline(PointList poly, boolean includeElevation) {
+        return encodePolyline(poly, includeElevation, 1e5);
+    }
+
+    public static String encodePolyline(PointList poly, boolean includeElevation, double precision) {
```
boldtrn: passing the precision is required to support the request parameter `polyline6`

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
```diff
@@ -0,0 +1,324 @@
...
+import com.graphhopper.util.*;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.UUID;
+
+public class MapboxResponseConverter {
+
+    // TODO: this is already in GraphHopper, but getting it from there is not possible easily?
+    private static final TranslationMap trMap = new TranslationMap().doImport();
```
karussell: Do you mean hopper.getTranslationMap?

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
```diff
@@ -0,0 +1,324 @@
...
+
+
+    /**
+     * Converts a GHResponse into Mapbox compatible json
+     */
+    public static ObjectNode convertFromGHResponse(GHResponse ghResponse, Locale locale) {
+        ObjectNode json = JsonNodeFactory.instance.objectNode();
+
+        if (ghResponse.hasErrors())
+            throw new IllegalStateException("Cannot convert an errornous ghResponse");
```
karussell: This should be possible. I mean, what if the route is not found or you want to go to the US from Europe. We have to support error JSON too.

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
```diff
@@ -0,0 +1,324 @@
...
+        for (int i = 0; i < waypoints.size(); i++) {
+            ObjectNode waypointJson = waypointsJson.addObject();
+            // TODO get names
+            waypointJson.put("name", "");
+            putLocation(waypoints.getLat(i), waypoints.getLon(i), waypointJson);
+        }
+
+        json.put("code", "Ok");
+        // TODO: Maybe we need a different format... uuid: "cji4ja4f8004o6xrsta8w4p4h"
+        json.put("uuid", UUID.randomUUID().toString().replaceAll("-", ""));
```
karussell: What is the purpose of this uuid? Maybe we have to hash the response somehow for error reporting?

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
```diff
@@ -0,0 +1,324 @@
...
+        return instructionJson;
+    }
+
+    private static void putVoiceInstructions(InstructionList instructions, double distance, int index, Locale locale, ArrayNode voiceInstructions) {
+        /*
+            A VoiceInstruction Object looks like this
+            {
+                distanceAlongGeometry: 40.9,
+                announcement: "Exit the traffic circle",
+                ssmlAnnouncement: "<speak><amazon:effect name="drc"><prosody rate="1.08">Exit the traffic circle</prosody></amazon:effect></speak>",
```
karussell: I love this xml style inside JSON :D

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Provides an endpoint that is compatible with the Mapbox API v5
+ *
+ * See: https://www.mapbox.com/api-documentation/#directions
+ *
+ * @author Robin Boldt
+ */
+@Path("mapbox/directions/v5/mapbox")
+public class MapboxResource {
```
karussell: Is it possible to use OSRM for the Mapbox SDK? If yes, can we rename to OSRMResource? If no, this is perfect already :)

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
boldtrn: Yes, but in the web-bundle we only have a reference to the GraphHopperAPI interface. We could probably check if the object is instance of GraphHopper and then get the TranslationMap? I thought that this would be too much overhead, but we could try?

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
boldtrn: Yes, it is :). It's just a different method, see `MapboxResponseConverter#convertFromGHResponseError`. Since we have to change the http response status to 422 in `MapboxResource` (at line ~137), I thought it would be a bit cleaner to have the two methods separated. It would be easily mergeable though. WDYT?

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
boldtrn: The purpose is not documented (or I could not find it). I haven't checked how it is used in the code. I think this is used internally to match error reports with requests. I tried sending multiple requests with the same parameters, and the UUID was different every time (but only at the end, the first characters were similar, so there seems to be some logic behind this).

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
boldtrn: This is actually called [SSML](https://en.wikipedia.org/wiki/Speech_Synthesis_Markup_Language) which is used to change the way TTS speaks the instructions.

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
boldtrn: Actually I don't think that this is reasonably possible. The problem is that the OSRM endpoint does not generate voice instructions nor banner instructions which are both important for the MapboxSDK.

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
karussell: As we bind

> bind(translationMap).to(TranslationMap.class);

We could maybe inject it in the resource?

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+        } catch (Exception e) {
+            logger.error("Exception while processing: ", e);
+            throw new MultiException(e);
+        }
+    }
+
+    private void throwIllegalArgumentException(String exceptionMessage) {
+        logger.error(exceptionMessage);
```
karussell: Why do we need this method? An exception should be logged from dropwizard automatically

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
```diff
@@ -0,0 +1,185 @@
...
+                        build();
+            } else {
+                return Response.ok(MapboxResponseConverter.convertFromGHResponse(ghResponse, request.getLocale())).
+                        header("X-GH-Took", "" + Math.round(took * 1000)).
+                        build();
+            }
+
+        } catch (Exception e) {
+            logger.error("Exception while processing: ", e);
+            throw new MultiException(e);
```
karussell: Same here: I think logging + throwing an exception seems duplicated

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
```diff
@@ -0,0 +1,185 @@
...
+
+    private void throwIllegalArgumentException(String exceptionMessage) {
+        logger.error(exceptionMessage);
+        throw new IllegalArgumentException(exceptionMessage);
+    }
+
+    private List<GHPoint> getPointsFromRequest(PathSegment pathSegment) {
+
+        // TODO: Could the map at some point break the order?
+        Set<String> pointParams = pathSegment.getMatrixParameters().keySet();
```
karussell: This seems to be indeed a bit risky. What is the parameter form? Isn't it possible to use the same approach like in RouteResource? `@QueryParam("point") List<GHPoint> requestPoints`

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
karussell: I see. Then the error message needs to be changed to "should be handled before in MapboxResource"?

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    /**
+     * Puts a location array in GeoJson format into the node
+     */
+    private static ObjectNode putLocation(double lat, double lon, ObjectNode node) {
+        ArrayNode location = node.putArray("location");
+        // GeoJson lon,lat
+        location.add(Helper.round6(lon));
+        location.add(Helper.round6(lat));
```
karussell: Maybe this is also dependent on the precision asked from outside? (I think I saw something for the simplification)

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
```diff
@@ -105,7 +105,7 @@ public Response doGet(
 
             double minPathPrecision = 1;
             if (overview.equals("full"))
-                minPathPrecision = 0.1;
```
karussell: Any reason why simplification is disabled? Does it not work or do we avoid a second request for the details?

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
boldtrn: The problem is that the points in the Mapbox SDK are so called "MatrixParameters" and not as QueryParameters.

An example would look like this: `...directions/v5/mapbox/driving-traffic/9.453778,48.70906;9.450624,48.706879`

I did some research, but only found this option to read this parameter form. As noted in the TODO, I am not especially happy with this either :).

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
boldtrn: No, this is only for the polyline. The request parameter is `geometries=polyline6` and only changes the precision of the Google Polyline like encoding.

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
boldtrn: Thanks, I will try this again, but IIRC in my tests so far, errors have not been logged by dropwizard, that's why I added this :).

karussell: Sorry, IllegalArgEx are supposed to avoid extra logging. But still it should print the request and info about a 'bad request'. If not, then we need to fix it globally for all IllegalArgEx

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
boldtrn: Ah thanks, that's a good idea. I tried to inject it and it seems to work. Could you check if [this](https://github.com/graphhopper/graphhopper/pull/1422/files#diff-973e4c6fa0434d64f5e3b627775d7a19R274) is what you meant?

web-bundle/src/main/java/com/graphhopper/resources/MapboxResponseConverter.java
boldtrn: Ok, sure :)

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
boldtrn: > Any reason why simplification is disabled?

Yes, the `overview` parameter means: "Type of returned overview geometry. Can be  full (the most detailed geometry available),  simplified (a simplified version of the full geometry), or  false (no overview geometry). The default is  simplified". 

I found that the Mapbox SDK works better, when more points are in the geometry. That's why I disabled simplification. Even with disabled simplification, the Mapbox API seems to return a more detailed geometry than GH.

Regarding PathDetails: This is not yet implemented for the Mapbox endpoint. This seems to work  different to the way we do PathDetails and might require a bit of time and planning (haven't checked in detail yet).

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
karussell: The exception should not be logged separately (please just throw an IllegalArgumentException or if not a bad request but indicating an error throw RuntimeException or IllegalState or whatever). 

If there is a bad request and it is not logged at all, we have to fix it for all resources, not sure here.

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
karussell: And did you try it for multiple points and does it work?

boldtrn: I double checked. It does not work, it seems that it can change the order.

boldtrn: I opened an SO post about this here: https://stackoverflow.com/questions/51420380/read-a-list-of-parameters-in-order-12345

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
boldtrn: Ok, will simply change this :)

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
boldtrn: Ok, according to SO, this is not easily fixable with JAX-RS. I switched to parsing the URL string manually. This seems to work well enough :).

boldtrn:
I updated this PR, improved the error logging and changed the way the points are parsed.

Let mw know if we can merge this or if there is anything else?

web-bundle/src/main/java/com/graphhopper/resources/MapboxResource.java
```diff
@@ -0,0 +1,181 @@
...
+
+        String[] pointStrings = url.split(";");
+
+        List<GHPoint> points = new ArrayList<>(pointStrings.length);
+        for (int i = 0; i < pointStrings.length; i++) {
+            points.add(GHPoint.fromStringLonLat(pointStrings[i]));
+        }
+
+        return points;
+    }
```
karussell: Will the following work?

`@Path("/{profile}/{coordinatesString}")`

and then String[] pointStrings = coordinatesString.split(";"); 

This way at least the first three lines can be avoided.

boldtrn: No, this was my first idea as well :).

The problem is that the path parameter coordinate string would only contain the first coordinate up to the first `;`, everything else is parsed as matrix parameter, which is unsorted.

karussell: Oh, ugly. And this means `{coordinatesArray: .*}` is no longer in use?

boldtrn: Yes, the parameter `{coordinatesArray: .*}` is no longer used. I could rename it or see if I can remove it?

karussell: Can you try out a different regex and see if this works? See e.g. https://stackoverflow.com/q/2291428/194609

boldtrn: I tried it like this:
```
@Path("/{profile}/{coordinatesArray : .+}")
...
@PathParam("coordinatesArray") String coordinates
```

I think this is what you meant? It does not work. The variable `coordinates` only contains the first coordinate up to the first `;`, everything after that seems to be read as matrix parameter.

karussell:
This feature will be developed in its separate repository: https://github.com/graphhopper/graphhopper-navigation

PR #1424, opened 2018-07-17
Add model Dockerfile and update toronto yaml

stevededalus:
Hi Michael Z, 

Can we push modifications to the model.yamls on this branch? How do you want to manage this?


stevededalus:
(@bnaul @alexeisw)

michaz:
Ah, hi @stevededalus! Yeah, I think you're right, let's continue this on your fork. Gets rather deployment-specific.

PR #1425, opened 2018-07-20
Replace synchronized frozen with atomic

Azbesciak:
It allows high performance improvement while using multiple threads, moreover - frozen should change state only once, but it is checked much more often - then other threads are blocked on this check.

When testing with yourkit, previous implementation caused nearly 60% of time in lock (using ForkJoinPool but this makes no difference). After this change I have noticed whole this lock dissapeared (with about 60% performance improvement).
I would even think if this lock is necessary, because graph cannot be unfrozen, and even after receiving false from `isFrozen`, it gives you nothing because all after is not atomic operation. 

PS: I have also tried `ReentrantReadWriteLock` but it was even worse than `synchronized`... :disappointed:

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/BaseGraph.java
+++ b/core/src/main/java/com/graphhopper/storage/BaseGraph.java
@@ -33,6 +33,7 @@ import org.slf4j.LoggerFactory;
 import static com.graphhopper.util.Helper.nf;
 
 import java.util.Locale;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * The base graph handles nodes and edges file format. It can be used with different Directory
--- a/core/src/main/java/com/graphhopper/storage/BaseGraph.java
+++ b/core/src/main/java/com/graphhopper/storage/BaseGraph.java
@@ -85,7 +85,7 @@ class BaseGraph implements Graph {
     private GHBitSet removedNodes;
     private int edgeEntryIndex, nodeEntryIndex;
     private long maxGeoRef;
-    private boolean frozen = false;
+    private AtomicBoolean frozen = new AtomicBoolean(false);
 
     public BaseGraph(Directory dir, final EncodingManager encodingManager, boolean withElevation,
                      InternalGraphEventListener listener, GraphExtension extendedStorage) {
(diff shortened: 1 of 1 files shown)


karussell:
Cool, thanks a lot! This sounds interesting.

> with about 60% performance improvement

For which usage scenario is this number? I can imagine this is when creating the graph and shortcuts? Can you somehow create an integration test for this? 

One problem could come from the fact that the call to listener.freeze is now unsynchronized. Will have to think about it.

Azbesciak:
> One problem could come from the fact that the call to listener.freeze is now unsynchronized. Will have to think about it.

Yes, `listener.freeze` is unsynchronized, but it will be invoked just once (because `frozen.getAndSet(true)`) will(or at least should) return `false` just once. Listener would be synchronized just in case `frozen` was `false` before, so expected result is the same.

> Can you somehow create an integration test for this?

You know that testing concurrency is not the best possible job to do. Will think about it, but I just wonder is it even necessary - explanation above.

> For which usage scenario is this number? I can imagine this is when creating the graph and shortcuts?

Rather on the other side - reusing created (and frozen) graph in many threads. I think this solution is better than making shallow copy of the graph (not quite since `BaseGraph` is private final field in `QueryGraph`), so... here I am.

karussell:
> You know that testing concurrency is not the best possible job to do. Will think about it, but I just wonder is it even necessary - explanation above.

Ah, sorry. I just meant a way to reproduce the performance numbers :)

> reusing created (and frozen) graph in many threads

But why is a synchronized block reached in these cases? Isn't `isFrozen` just called for creating shortcuts & edges?

karussell:
In CHGraphImpl it is just an assert statement that should be removed for production runs:

> assert baseGraph.isFrozen() : "Traversal CHGraph is only possible if BaseGraph is frozen";

You have to enable these assertations explicitly via the JVM option -ea or also maven does it for test executions

Azbesciak:
Yeah, I didn't notice that Intellij enables it by default in run config. But does it mean that this change is redundant :)?

karussell:
> Yeah, I didn't notice that Intellij enables it by default in run config

Strange. It should not ... or ... only for tests

> But does it mean that this change is redundant :)?

I fear it is but maybe we see some use later on. Currently for writing there can be only one thread, as we did not make the graph write-thread safe yet.

PR #1431, opened 2018-08-03
Simplify encoders registration.

Azbesciak:
Not sure if JDK 8 is supported in every module (because of android). Maybe for posterity, IMO it is cleaner than previous version.


devemux86:
That would require [Android API level 24 or higher](https://developer.android.com/studio/write/java8-support#supported_features), i.e. Android 7 Nougat.

Azbesciak:
https://www.androidpolice.com/2017/12/19/play-store-require-new-updated-apps-target-recent-api-levels-distribute-native-apps-64-bit-support/ BTW

PR #1436, opened 2018-08-09
Access refactoring

karussell:
Make access handling for FlagEncoders more readable via a separate and new class (EncodingManager.Access)

Rename AccessValue to Access for SpatialRules

Merge this after #1460  so that we can use this better API in the public TagParser interface instead of "allowed". Probably then also move TagParser out of EncodingManager as it is more related to parsing not encoding the info.


core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRule.java
```diff
@@ -31,6 +31,12 @@
  */
 public interface SpatialRule {
 
+    enum Access {
+        YES,
+        CONDITIONAL,
```
boldtrn: Good choice :+1:

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -238,13 +235,70 @@ private FlagEncoder getEncoder(String name, boolean throwExc) {
...
+        boolean isWay() {
+            return this.ordinal() == WAY.ordinal();
+        }
+
+        boolean isOther() {
+            return this.ordinal() == OTHER.ordinal();
+        }
+
+        boolean canSkip() {
+            return this.ordinal() == CAN_SKIP.ordinal();
```
boldtrn: Why canSkip?

karussell: like a way for houses

boldtrn: Thanks for explaining. Maybe this is just me, but I always thought the `return 0 ` to be more a `don't accept ` than a `can skip`, that's why I asked.

karussell: "can skip" in the sense of do not use it for this flagencoder. So it could be used for another ... still we can use a different wording like "ignore"?

boldtrn: Thanks, yeah maybe can skip is not bad from this point of view. Maybe ignore would be better, but I am not sure, I think we can also leave it as it is right now.

PR #1442, opened 2018-08-21
Allow configuring the precision of the polyline encoding

boldtrn:
As discussed in https://github.com/graphhopper/graphhopper-navigation/issues/1, we should make the precision of the polyline encoding configureable. 

This PR adds the option to change the precision of the Polyline encoding to use for example a precision of `1e6` as used by OSRM/Mapbox.


karussell:
Thanks!

Issue #1443, opened 2018-08-22
Reduce dependencies for client-hc


karussell:
See the [discussion here](https://discuss.graphhopper.com/t/reduce-number-of-dependencies-for-client-hc/3389) about the used dep dropwizard-jackson.

devemux86:
Is #1444 going to be merged?

Issue #1039, opened 2017-04-25
Replace JTS with EPLicense version from LocationTech


karussell:
After gtfs addition we need to update the [NOTICE file](https://github.com/graphhopper/graphhopper/blob/master/NOTICE.md) and use the [EDL version](https://github.com/locationtech/jts) for jts once released:

```xml
<groupId>org.locationtech.jts</groupId>
<artifactId>jts</artifactId>
<version>1.15.0</version>
```

Include elevation data attribution and GTFS attribution in the NOTICE file or somewhere else.

karussell:
So for this move we need [this new dependency](https://github.com/graphhopper/graphhopper/pull/1230#issuecomment-359590077) and a [new gtfs-lib](https://github.com/conveyal/gtfs-lib/issues/97)

willcohen:
Re gtfs-lib -- since they've got some moving pieces on their own end to work with before they can release an updated version, would you prefer to wait on their release for this change or use a forked version without geotools (it's already in a PR in that repo) and an updated JTS pending that? Updating their jackson dependency to use the new JTS should also be pretty simple.

karussell:
Good question. Waiting for their changes reduces a bit our own work :)

willcohen:
Okay! If you decide you'd like to give the fork a go, let me know and I can put together PRs for those dependencies.

karussell:
@michaz It seems that the team behind gtfs-lib won't use the latest JTS soon due to their internal stuff, see [this comment](https://github.com/conveyal/gtfs-lib/issues/97#issuecomment-370706124). Should we use the forked gtfs-lib maybe with help of @willcohen or avoid future hassle and stay on JTS 1.14 for now?

willcohen:
I know Geotools is a big holdup for them to transition their other software, but 20 should be out any day now. If getting from whatever Geotools version they're currently using in other projects to 20 ends up being super complicated it may drag this out out, but if it's not too hard an upgrade then they may be able to get gtfs-lib up to speed quicker.

michaz:
No. They are apparently moving to using an actual database as a backend, and the part I am using is _specifically_ the one they are going to remove.

willcohen:
Would you all prefer to ask them to keep GTFSFeed in and migrate JTS up, or just fork the older version and stick to that?

michaz:
Right now I think I would just "inline" the bits I am using and drop the binary dependency.

michaz:
[...]

PR #1444, opened 2018-08-22
Slimmer dependencies for client. #1443

michaz:
Would this work? @devemux86 @boldtrn


karussell:
Thanks @michaz !

BTW: to use the same versions for these things like dropwizard does - maybe we can utilize their "bom" project somehow? https://github.com/dropwizard/dropwizard/blob/master/dropwizard-bom/pom.xml#L15

devemux86:
Thanks @michaz now client-hc is much lighter to use!

BTW
Why not use the `com.graphhopper.external:jackson-datatype-jts` instead of `com.bedatadriven:jackson-datatype-jts`?
With `bedatadriven` cannot avoid the older `com.vividsolutions:jts` dependency due to `JtsModule` creation in `Jackson`.

karussell:
GH core uses 1.14.0. To use the latest version 1.15.0 there was still an issue (but I think this should work now): https://github.com/graphhopper/graphhopper/issues/1039

PR #1450, opened 2018-09-03
Simplifies path unpacking code for CH.

ammagamma:
In this PR the path unpacking code in Path4CH is simplified.
The simplification makes the code easier to understand (hopefully, for me at least) and makes it easier to adjust for the turn-cost supporting version of CH.
I could not measure any significant performance difference running several comparisons on the map for Germany.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ch/Path4CH.java
+++ b/core/src/main/java/com/graphhopper/routing/ch/Path4CH.java
@@ -18,7 +18,6 @@
 package com.graphhopper.routing.ch;
 
 import com.graphhopper.routing.PathBidirRef;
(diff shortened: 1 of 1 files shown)


karussell:
Interesting. This looks indeed much simpler - thanks! 

Did you test this with a bigger area? As there were a few edge cases where I'm not sure if the current test coverage is enough.

And as this reduces the getEdgeIteratorState calls in a few cases - has it a positive influence on performance (routingCH_no_instr.mean)?

karussell:
Thanks!

> I could run the same on the Europe map or which edge cases do you have in mind?

If Germany then probably all edge cases are already covered. 

Will merge.

Issue #1346, opened 2018-04-20
Same source code formatting


karussell:
Do this for all files as otherwise we get these changes for every single commit:

 * https://github.com/graphhopper/graphhopper/pull/1341#discussion_r182501361
 * https://github.com/graphhopper/graphhopper/issues/1330#issuecomment-383080385

PR #1451, opened 2018-09-03
Fixes some code formatting.

easbar:
I just updated an older branch to master and noticed that in the current master there are quite a few classes (63 actually) where default code formatting in IntelliJ changes existing code. 

image
Auto formatting breaks some javadocs, like this one in `InstructionsOutgoingEdges.java`:

image

Anyway, I fixed some (hopefully) harmless cases, because different formatting can be quite annoying as well ;)
The `reader-gtfs` module has a few more files that aren't formatted, but I wasn't sure if those should be adjusted as well, possibly risking unnecessary conflicts with other branches @michaz?


michaz:
> The reader-gtfs module has a few more files that aren't formatted, but I wasn't sure if those should be adjusted as well, possibly risking unnecessary conflicts with other branches @michaz?

Yeah, I'd be happier if you wouldn't. Thanks. :)

karussell:
Strange, why there are again so many formatting changes, see recently #1346. Maybe we should use some maven plugin that enforces this with every commit?

ammagamma:
> Strange, why there are again so many formatting changes, see recently #1346. 

I think in #1346 we mostly fixed license headers and <> operator uses.

karussell:
Having the code formatting consistent for all modules including reader-gtfs, would help also with refactorings that go across the whole project like issue 1447. What are the reasons we cannot apply the usual formatting here @michaz?

michaz:
I don't mind any particular formatting, I just use long-running branches which are tedious to merge after global reformattings.

> would help also with refactorings that go across the whole project like issue 1447.

How?

karussell:
> I just use long-running branches which are tedious to merge after global reformattings.

:+1: 

> > would help also with refactorings that go across the whole project like issue 1447.

> How?

When I edit a file I automatically do formatting as a lazy habit to avoid manual indenting which also leads to conflicts for longer running branches. But of course this is not important as I do this very rarely but just a note that we should migrate to a standard formatting in a year or so.

karussell:
btw: in IntelliJ there is a "format on commit" option that we could use to migrate slowly (only a few edited files) to the same formatting for the whole project.

PR #1455, opened 2018-09-07
Replaces truncation with rounding when setting edge distance or CH shortcut weight.

easbar:
Truncating the shortcut weight before storing it introduces small deviations from Dijkstra, which sometimes makes it harder to check the correctness of CH. Therefore I have replaced the long truncation with a `Math.round()` call.
Performance-wise isolated `Math.round()` will certainly be slower, but the profiler shows that the time spent on weight setting during the CH preparation is negligible. `setWeight` is only used for CH preparation.


ammagamma:
tests failed with API limit reached ...

karussell:
@ammagamma Would a higher precision also help? As it is very likely that we change this behaviour (we could use the same EncodedValue approach we have for non-CH but then it would be likely without rounding) can you introduce a test that is different with the rounding?

ammagamma:
> can you introduce a test that is different with the rounding

Good point. I added a test showcasing the problem where CH yields a different route than Dijkstra. Its quite intricate and I had to carefully pick distance values to reproduce the problem. I am not even sure if the test will fail on any machine. Not sure how deterministic the code / truncation is, need to refresh my knowledge here. I think the problem is that something like 2.010 might be represented as 2.009999 and after the truncation it will be 2.009. If this happens a few times a shortcut or set of nested shortcuts appears to be shorter than the real road and CH can take another way than Dijkstra.

Also this test does not pass with my previous fix, so I reverted it for now.

@Azbesciak Thanks for posting your problem, I might have a look and see if this is related to the issue I am referring to here.

core/src/test/java/com/graphhopper/routing/ch/NodeBasedNodeContractorTest.java
```diff
@@ -288,6 +287,65 @@ public void testContractNode_directed_withWitness() {
...
+        double[] distances = {4.019, 1.006, 1.004, 1.006, 1.004};
+        graph.edge(0, 4, distances[0], false);
+        EdgeIteratorState edge1 = graph.edge(0, 1, distances[1], false);
+        EdgeIteratorState edge2 = graph.edge(1, 2, distances[2], false);
+        EdgeIteratorState edge3 = graph.edge(2, 3, distances[3], false);
+        EdgeIteratorState edge4 = graph.edge(3, 4, distances[4], false);
+        graph.freeze();
+        setMaxLevelOnAllNodes();
+
+        // make sure that distances do not get changed in storage (might get truncated)
```
ammagamma: When edges get stored the distance is also truncated, for example 1.006 can become 1.005. I tried choosing values where this does not happen because this is for another issue.

karussell:
Is this still an issue with the latest version of CH+turn costs?

easbar:
> Is this still an issue with the latest version of CH+turn costs?

Good question :) I am getting much fewer differences between CH and Dijkstra in `CHMeasurement` in the turncost_ch branch with this fix. Also when I increase `WEIGHT_FACTOR` to 10_000 I achieve the same. But then again I am not sure if I understood the issue completely and most of the differences are very small anyway (below one second of driving time, but some rare cases where the difference is a few seconds). I do not think this fix is strictly necessary, but it would be a bit cleaner this way.

easbar:
I updated this to master, but I am still unsure. I think rounding is definitely 'more correct' than truncating and e.g. when comparing CH with Dijkstra we get less deviations (but they are always quite small). For better consistency we might even apply the rounding in `EdgeAccess#setDist` (which I have done here). But I am not sure if this is worth it. Asking the other way around: What are the cons of using Math.round instead of truncation ?

karussell:
> What are the cons of using Math.round instead of truncation 

Good question. Maybe performance. But we do this now already in FactorizedDecimalEncodedValue and MappedDecimalEncodedValue too:

private int toInt(double val) { return (int) Math.round(val / factor); }

And so this should be probably not be a big issue. Especially since rounding would then just more consistent.

easbar:
>  But we do this now already in FactorizedDecimalEncodedValue and MappedDecimalEncodedValue too.

Ok in this case I think we should apply the changes of this PR.

core/src/main/java/com/graphhopper/storage/CHGraphImpl.java
```diff
@@ -853,7 +853,7 @@ int weightToWeightFlags(long edgePointer, double weight) {
             if (weight > MAX_WEIGHT)
                 weightInt = MAX_WEIGHT_31;
             else
-                weightInt = ((int) (weight * WEIGHT_FACTOR)) << 2;
+                weightInt = Math.round((float) (weight * WEIGHT_FACTOR)) << 2;
```
karussell: In terms of precision wouldn't it be better to do `(int)Math.round(x)`? (If not, then we would need to do this in the EncodedValue classes like here)

easbar: Oh, no idea tbh! Not sure if its better to cast to float and then round to int or round to long and cast to int. I changed it to the way you proposed.

core/src/main/java/com/graphhopper/storage/EdgeAccess.java
```diff
@@ -73,12 +73,14 @@ final void setDist(long edgePointer, double distance) {
      * Translates double distance to integer in order to save it in a DataAccess object
      */
     private int distToInt(double distance) {
-        int integ = (int) (distance * INT_DIST_FACTOR);
-        if (integ < 0)
+        if (distance < 0)
```
easbar: I now check if distance is negative, because when integ is is negative it could also be because it overflowed.

core/src/main/java/com/graphhopper/storage/EdgeAccess.java
```diff
@@ -73,12 +73,14 @@ final void setDist(long edgePointer, double distance) {
...
             throw new IllegalArgumentException("Distance cannot be negative: " + distance);
-        if (integ >= Integer.MAX_VALUE)
-            return Integer.MAX_VALUE;
-        // throw new IllegalArgumentException("Distance too large leading to overflowed integer (#435): " + distance + " ");
+        if (distance > MAX_DIST) {
+            distance = MAX_DIST;
+//            throw new IllegalArgumentException("Distance too large leading to overflowed integer (#435): " + distance + " ");
```
easbar: This comment can probably be removed ?

core/src/test/java/com/graphhopper/storage/AbstractGraphStorageTester.java
```diff
@@ -122,12 +122,9 @@ public void testSetTooBigDistance_435() {
         EdgeIteratorState edge1 = graph.edge(0, 1, maxDist, true);
         assertEquals(maxDist, edge1.getDistance(), 1);
 
-        // max out should NOT lead to infinity as this leads fast to NaN!
```
easbar: This test only passed previously because there was no check that an exception is thrown at all!

core/src/main/java/com/graphhopper/storage/EdgeAccess.java
```diff
@@ -73,12 +73,14 @@ final void setDist(long edgePointer, double distance) {
      * Translates double distance to integer in order to save it in a DataAccess object
      */
     private int distToInt(double distance) {
```
easbar: Can you please double check this method ? I was trying to follow the same logic as in CHGraphImpl, but for example I am not really sure why there is -1 in MAX_DIST

core/src/main/java/com/graphhopper/storage/EdgeAccess.java
karussell: The problem is that we do not handle negative integers in getDist yet:

val / INT_DIST_FACTOR

and I prefer having this check as early as possible. See also #1544

core/src/main/java/com/graphhopper/storage/EdgeAccess.java
karussell: Yes

core/src/main/java/com/graphhopper/storage/EdgeAccess.java
karussell: Ok, due to MAX_DIST an integer overflow should never occur

core/src/test/java/com/graphhopper/storage/AbstractGraphStorageTester.java
karussell: ok

core/src/main/java/com/graphhopper/storage/EdgeAccess.java
easbar: ok maybe you can just adjust this method ?

core/src/main/java/com/graphhopper/storage/EdgeAccess.java
karussell: Likely due to `>=` vs. `>`. Maybe we should somehow use the code from FactorialDecimalEncodedValues so we do not introduce inconsistency and use all 32 bits

core/src/main/java/com/graphhopper/storage/EdgeAccess.java
easbar: With this code from master: 
```
    private int distToInt(double distance) {
        int integ = (int) (distance * INT_DIST_FACTOR);
        if (integ < 0)
            throw new IllegalArgumentException("Distance cannot be negative: " + distance);
        if (integ >= Integer.MAX_VALUE)
            return Integer.MAX_VALUE;
        return integ;
    }
```
We would get the exception for too large values of distance because integ overflows and becomes negative, that's why I wanted to check if distance>MAX_DIST directly.

karussell: Ok, yes. It looks that we do not have a test for this. So when I remove "if (integ >= Integer.MAX_VALUE) return Integer.MAX_VALUE;" everything is fine, which I would fix before doing the refactoring.

easbar: ok this check makes no sense, because integ is an int and cannot have a value larger than Integer.MAX_VALUE. There is `AbstractGraphStorageTester#testSetTooBigDistance_435` which should fail if a too large distance is not reset to Integer.MAX_VALUE.

karussell: > integ is an int and cannot have a value larger than Integer.MAX_VALUE
Ups! 

> There is `AbstractGraphStorageTester#testSetTooBigDistance_435` which should fail if a too large distance is not reset to Integer.MAX_VALUE

I see. Looks like your propose change makes a lot more sense.

karussell: Ouch, the test is also buggy and it should be:

```
 @Test
    public void testSetTooBigDistance_435() {
        graph = createGHStorage();

        double maxDist = EdgeAccess.MAX_DIST;
        EdgeIteratorState edge1 = graph.edge(0, 1, maxDist, true);
        assertEquals(maxDist, edge1.getDistance(), 1);

        try {
            EdgeIteratorState e = graph.edge(0, 2, maxDist + 1, true);
            assertEquals(maxDist, e.getDistance(), .1);
        } catch (Exception ex) {
            fail();
        }
    }
```

karussell:
Will remove the "-1" for clarity

PR #1459, opened 2018-09-19
Inline Gtfslib

michaz:
Inline the bits of Gtfslib I'm using. Mostly just the model classes (deserializing GTFS CSV files), and putting them into MapDB.

Paste Conveyal's license file.


karussell:
The forbidden API checker has a problem due to the default locale:
```
[ERROR] Forbidden method invocation: java.lang.String#format(java.lang.String,java.lang.Object[]) [Uses default locale]
```

PR #1460, opened 2018-09-20
refactor the EncodingManager to use the builder pattern

karussell:
This will make the EncodingManager more flexible, e.g. it will be possible later on to do something before adding a FlagEncoder like adding a "shared" EncodedValue (like road type to be available for multiple FlagEncoders)

Should we use the same Builder patter approach like we do in AlgorithmOptions:
```java
EncodingManager.start().add(new CarFlagEncoder()).build();
```
or something different?

This PR is based on #1447


ammagamma:
I also like `.hasEncoder()` better than `.supports()`. Regarding the builder pattern I am not sure if I can follow your example. Do you want to just store the "shared" encoded value within the encoding manager or should it be added to every encoder that gets added during the building ? Do you have any specific alternative to the builder pattern in mind ?

karussell:
> Do you want to just store the "shared" encoded value within the encoding manager or should it be added to every encoder that gets added during the building ?

We already have EncodedValues in master but they live inside the FlagEncoders. With #1447 I extend the idea further and at some point in the future we could get rid of FlagEncoders and just have EncodedValues. So, yes. The shared EncodedValue is part of the EncodingManager and can be used from a Weighting but also within any FlagEncoder without duplicating it. E.g. in master the roundabout EncodedValue is duplicated for every FlagEncoder and the same ugliness would happen for a "surface" or "road class" EncodedValue but with a bigger bit range impact. So we need something like these shared EncodedValues to solve this.

> Do you have any specific alternative to the builder pattern in mind ?

 ... not really - I hoped you would have one :) !

karussell:
Please have a look into [this discussion](https://discuss.graphhopper.com/t/flexible-vehicle-profiles/3612), where I link to [this highly experimental branch](https://github.com/graphhopper/graphhopper/compare/flex_vehicleprofile), where I develop the idea further, introduce a yml/json format to define a different profile per request.

ammagamma:
Thanks for the pointers, I will have to take a closer look at EncodedValues. I think using the builder pattern like you did seems to be a reasonable way to share the encoded values between different encoders that get added to the manager.

karussell:
Somehow I cannot change the target branch, so created this new at #1546

Issue #1368, opened 2018-05-13
Find some better way to handle parameters for algorithm preparation


easbar:
Graph preparation like CH and LM typically depends on some kind of parameters like the number of lazy updates in CH or the number of landmarks in LM. As of now these parameters are set via setters or the constructor in the corresponding preparation classes and the number of parameters is relatively small. For CH with turn costs (and possibly other future algorithms) there will be a few more parameters that are also needed in different places and while there should always be some good default values there should still be a way to adjust the parameters to special situations, hardware etc. Especially when the preparation classes are split into multiple (like PrepareContractionHierarchies and NodeContractor) we need to pass the parameters down from one class to the other (for example from `CHAlgoFactoryDecorator` to `PrepareContractionHierarchies` to `NodeContractor` etc.). This is rather cumbersome and adding/changing/removing a parameter requires changing multiple classes. I think it would be cleaner to pass along a `PMap` options object (that is where the parameters originate from anyway) such that each related class can pick the options that it requires. This might sound a bit like some kind of global object and obviously should not be misused to hold some other state, but I think as long as it is solely used as a read-only parameter object this is totally fine and makes the code easier to adjust to future changes. What do you think ?

ammagamma:
> But when we put everything into a ConfigObject, we still need to change each class if we change a parameter?

This depends on the amount of code that is required to pass along the config object and the amount of code that is used to write/read from the object. As soon as the config object is passed down the chain of classes that need it adding a parameter is only a matter of adding it initally and using it wherever it is needed, while adding a single parameter requires adjusting all the setters used in the chain.

> The problem with maps is that you can change a key at one end but forget to change it at the other end. You would only find this once the code crashes.

Yes, we lose the safety provided by the type system, but since the config fields are only used to override the (hard-coded) default values nothing would crash, but it might happen that a misspelled key leads to some parameter not being overwritten (which you might argue is even worse ;). At the same time I think this is also an advantage we have less compatibility issues. I remember this was the reason why we used a simple String instead of a static constant in a similar situation here:
https://github.com/graphhopper/graphhopper/blob/06ef11047737cb86c6c81090dedf7d986055dc94/core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java#L197

[...]

PR #1476, opened 2018-10-06
Makes parameters available to CH classes via PMap.

ammagamma:
Following the discussion in #1368 I have tried to improve the parameter handling for CH and made the following changes:

1. `CHAlgoFactoryDecorator` is no longer aware of which parameters are used for `PrepareContractionHierarchies` and `NodeBasedNodeContractor`, but instead passes on the `CmdArgs` it is given in its `init()` method to `PrepareContractionHierarchies`. I think this is an improvement, because the decorator is no longer 'polluted' by CH-specific parameters. There will be more parameters when we add edge-based CH and even right now there are some parameters that are simply hard-coded and can't be changed at all (like the weightings in `NodeBasedNodeContractor#calculatePriority`). On the other hand since I removed the setters (`setPreparationContractedNodes` etc.) there is no longer a Java API to modify these parameters (not sure if that is ok regarding backwards-compatibility!). However, I think these really are expert settings and if someone really wants to make changes its still possible to add them to the `CmdArgs` object passed to the `init()` method. 

2. Since the `PMap` is simply passed along the parsing of the parameters now happens in `PrepareContractionHierarchies` and `NodeBasedNodeContractor` and can be used to overwrite the defaults that are hard-coded in these classes. This mechanism is mostly useful to overwrite the defaults for very specific use-cases or performance analysis.

3. I added/changed a few methods in `PMap` to be able to only take action / overwrite the defaults if the `PMap` does contain a valid value for a certain key.

4. I have encapsulated the parameters in `PrepareContractionHierarchies` into an inner class, for example this will be useful to be able to easily switch these parameters depending on the `TraversalMode`.

5. As application/demonstration of this method to pass parameters I have made the above mentioned node priority weights in `NodeBasedNodeContractor` configurable via the `PMap`.

I think this PR allows adding new parameters (if needed) with very little code and there is no coupling of higher-order classes (like the decorator) to details of the lower-lying classes (like the node priority weights of the node contractor).

Another philosophy would be to parse the input parameters / configurations much earlier in the application, create Java objects from them and pass these along. This would lead to more type-safety of course, but I came to the conclusion that I like using the `PMap` better for the reasons I gave. 
I'll be very happy to hear your opinion on this 'exciting' topic ;)

Minor: I changed the logMessages parameter from double to int for consistency with the other parameters.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ch/CHParameters.java
+++ b/core/src/main/java/com/graphhopper/routing/ch/CHParameters.java
@@ -1,0 +1,17 @@
+package com.graphhopper.routing.ch;
+
+import com.graphhopper.util.Parameters;
+
+public final class CHParameters {
+    static final String PERIODIC_UPDATES = Parameters.CH.PREPARE + "updates.periodic";
+    static final String LAST_LAZY_NODES_UPDATES = Parameters.CH.PREPARE + "updates.lazy";
+    static final String NEIGHBOR_UPDATES = Parameters.CH.PREPARE + "updates.neighbor";
+    static final String CONTRACTED_NODES = Parameters.CH.PREPARE + "contracted_nodes";
+    static final String LOG_MESSAGES = Parameters.CH.PREPARE + "log_messages";
+    static final String EDGE_DIFFERENCE_WEIGHT = Parameters.CH.PREPARE + "node.edge_difference_weight";
+    static final String ORIGINAL_EDGE_COUNT_WEIGHT = Parameters.CH.PREPARE + "node.original_edge_count_weight";
+    static final String CONTRACTED_NEIGHBORS_WEIGHT = Parameters.CH.PREPARE + "node.contracted_neighbors_weight";
+
+    private CHParameters() {
+    }
+}
--- a/api/src/main/java/com/graphhopper/util/PMap.java
+++ b/api/src/main/java/com/graphhopper/util/PMap.java
@@ -85,101 +85,78 @@ public class PMap {
     }
 
     public long getLong(String key, long _default) {
-        Long longOrNull = getLongOrNull(key);
-        return longOrNull != null ? longOrNull : _default;
-    }
-
-    public Long getLongOrNull(String key) {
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/graphhopper/routing/ch/CHParameters.java
```diff
@@ -0,0 +1,17 @@
+package com.graphhopper.routing.ch;
+
+import com.graphhopper.util.Parameters;
+
+public final class CHParameters {
+    static final String PERIODIC_UPDATES = Parameters.CH.PREPARE + "updates.periodic";
```
ammagamma: I kind of see the point of reusing `Parameters.CH.PREPARE` here but I would prefer inlining these constants here (e.g. `"prepare.ch.updates.periodic"` instead of `Parameters.CH.PREPARE + "updates.periodic"`, because this makes it easier to search for this setting in the code base (I imagine seeing this setting in some configuration file and the first thing I would do would be to do a full text search to find out what it does in the code). But maybe this is for another discussion and have left this as it was.

karussell: Yeah, we have this for the Parameter class too. Not sure about it, because now you can click "find usage" and no need for text search ;)

ammagamma: I was thinking about the case that you only know the name of the parameter (from a config file for example) in which case you would do a text search (you cannot search for usages at this point because you do not know that this constant exists).  I think when the parameter value is half constant and half string you might miss it when you search for the full parameter name.

ammagamma: of course doing it this way allows finding everything related to PREPARE.CH, but not sure if this is needed that often ?

karussell: Yes. Probably this would be more consistent. In earlier version I had no such constants (like there are still a few), but this is also suboptimal.

PR #1483, opened 2018-10-15
Postgis 0.11

mbasa:
Creates a PostGIS data reader to convert road networks located in PostGIS directly into GraphHopper's native network format. Please see the README.postgis for more information.

### DIFF
--- a/README-PostGIS.md
+++ b/README-PostGIS.md
@@ -1,0 +1,28 @@
+## PostGIS Reader
+To create a GraphHopper Graph from a PostGIS table, do:
+
+* Uncomment the PostGIS parameters in the config.yml and set the appropriate values. The PostGIS reader will only be used if all the PostGIS parameters are set.
+
+ 
+* Create a Table or View in the specifed database and schema that has these necessary columns (refer to OSM documentation for the proper values of these columns) : 
+
+``` 
+osm_id, maxspeed, oneway, fclass, name, geom 
+```
+
+For example:
+
+``` 
+ create view philview 
+   (osm_id,maxspeed,oneway,fclass,name,geom ) 
+   as select id,0,oneway,'tertiary'::text,name,geom from phil;
+```
+
+* Start the GraphHopper server by adding the parameter ``/<path where graph will reside>/<table or view name>``. The example below will create a graph, philview-gh, in /Users/mbasa/t for the philveiw view of PostgreSQL:
+
+``` 
+./graphhopper.sh web /Users/mbasa/t/philview
+```
+
+* If the data in PostgreSQL changes and the graph has to be updated, just delete the created graph directory and restart GraphHopper using the above method.
+
--- a/config-example.yml
+++ b/config-example.yml
@@ -3,6 +3,15 @@ graphhopper:
   # OpenStreetMap input file
   # datareader.file: some.pbf
 
+  ##### PostGIS Connection Parameters #####
+  
+  # db.host: localhost
+  # db.port: 5432
+  # db.database: eraseme
+  # db.schema: public
+  # db.user: mbasa
+  # db.passwd: xx
+
   ##### Vehicles #####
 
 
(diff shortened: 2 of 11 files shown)


karussell:
Thanks a lot, this is a valuable addition to the GraphHopper ecosystem! Probably many users have waited for it. Please do not hesitate to post it on [our forum](https://discuss.graphhopper.com/c/graphhopper) too.

Still, I think, this should be better reside in a separate repository as e.g. we won't be able to include it in web-bundle per default (too many dependencies that most of the users won't need, would be pulled in) and also maintenance needs to be guaranteed once included.

With a similar reasoning we are also planning to move reader-shp out of the repository as this is a similar module and no one from the community seems to be interested (and we core devs do not use it).

Please try to do this in a separate module like we are doing it for [map-matching](https://github.com/graphhopper/map-matching) and [our navigation module](https://github.com/graphhopper/graphhopper-navigation). E.g. if you still like to use the same graphhopper.sh script file then this should already work using the `--jar` option.

Wiring the modules together to get a running web server is trivial as [this example](https://github.com/graphhopper/map-matching/blob/0.11/matching-web/src/main/java/com/graphhopper/matching/http/MapMatchingApplication.java) in the map matching module shows.

Please let us know if you need assistance in this process and if changes to the GH core are required. Will let it open until we are sure it works for you :)

Nakaner:
To be honest, I don't understand why someone needs a PostGIS import for GraphHopper.

This pull requests adds a PostGIS reader but it can only read PostGIS tables which have been imported from Geofabrik's data free shape files. Geofabrik's free shape files are not designed to be used for routing applications. As you have observed, roads are not split at intersections.

It is faster to use the OSM raw data import of GraphHopper (avoids database related overhead). I can imagine one benefit only – a reduced memory consumption because the expensive caching of node locations happens on a Geofabrik server where the shape file is produced. However, someone who suffers this issue will not become happy if he/she tries to read a whole planet from a PostGIS database. Not to mention that Geofabrik's free shape files are not available for continents or the whole world.

I think that it is not a good idea to merge code into master which relies on a proprietary product of a company which can be replaced by a free and open source solution which is already supported by GraphHopper. I agree with @karussell that it should be in a separate project using GraphHopper.

@mbasa  Or is there any use case which cannot be solved without a PostGIS import? I am happy to give suggest alternative solutions.

TimMcCauley:
@Nakaner I see one clear benefit here: the ability to enrich/change data (osm tags) in postgis easily which can be used in the street network.

karussell:
> @Nakaner I see one clear benefit here

Good idea. But this can/should happen in a separate repository too as readers are pluggable.

TimMcCauley:
Yep, I agree @karussell - separate repository makes sense for this.

mbasa:
This came about when we had a project that needed the very detailed Digital Road Map (DRM) data of Japan for GraphHopper usage. The data was already in PostGIS and converting the ~20 million road data into OSM was very painful due to the way the OSM data is structured. We also had to use the DRM's attributes when determining the cost in the route creation.

So the benefit of a PostGIS reader for us is:
1. ability to use non-OSM data (i.e. Navteq) easily with GraphHopper
2. ability to have dynamic cost for route creation (i.e. Traffic data)

I blogged about this here: 

https://georepublic.info/en/blog/2018/graphhopper-with-postgis-data-reader/

And I agree with everybody that this can be a separate repository. This was beneficial for us and it was hoped that it can be useful for others as well.

karussell:
Thanks for the details. Please open a new issue if you stumble over problems when externalizing this.

boldtrn:
Thanks for the great work @mbasa. I am currently working on reading data from PostGIS and create a GH graph from this. I took the liberty to transform @mbasa’s fork into a GraphHopper module, update to the latest version of GraphHopper and extending some of the functions, fix some minor issues, etc.

You can find the current state here, contributions are welcome :slight_smile: :
https://github.com/boldtrn/graphhopper-postgis

Issue #1324, opened 2018-03-31
Reimplement block_area with LocationIndexTree instead of graph traversal


karussell:
There are several limitations with the current approach and so we should use a better node search

Discussed partially in https://github.com/graphhopper/graphhopper/pull/1306#discussion_r177076643

Failing examples likely caused by this bug: https://discuss.graphhopper.com/t/4747 and https://discuss.graphhopper.com/t/3415 and https://discuss.graphhopper.com/t/3680

Also what should happen if the area is outside of the graph - should we silently ignore this, because the area does just not contain any graph or is this a "bad request"?

michaz:
Subtask: ``LocationIndexTree`` needs a ``query(Envelope envelope)`` or ``query(BBox bbox)`` method to get everything within a bounding box.

Also required for getting the isochrones right.

VanDerLars:
A fast fixing would be appreciate ;)

karussell:
Have now a [working prototype](https://github.com/graphhopper/graphhopper/compare/loc_idx_changes) and will try to condense this in a clean pull request. Have still to test it world wide.

Issue #1127, opened 2017-08-04
block_area does not work if start and/or end are on the same edge


karussell:
See [this route](https://graphhopper.com/maps/?point=50.17016%2C14.63381&point=50.17286%2C14.64287&block_area=50.17147%2C14.63683%2C100&ch.disable=true) vs. [this route](https://graphhopper.com/maps/?point=50.169037%2C14.631515&point=50.17393%2C14.64735&block_area=50.17147%2C14.63683%2C100&ch.disable=true) where the block_area parameter is identical and so it is expected that the point in the middle of the road section should be avoided in both cases. Or properly document the limitation of block_area.

boldtrn:
Thanks for reporting this, will investigate what's happening and if we can fix it.

boldtrn:
Ok, so I debugged this issue a bit. Currently, we find edges in the graph using the LocationIndex. When we place a waypoint on an edge, we don't traverse on that edge, but we use a VirtualEdge, which is not included in the LocationIndex.

A possible solution would be to get all QueryResuls from the QueryGraph and check if we found a blockingEdge here, if yes, we have to find the corresponding VirtualEdges, which should be possible. Then we could check if the blocked point/circle/BBox intersects with this edge, if yes we block the edge.

I outlined this approach [here](https://github.com/graphhopper/graphhopper/compare/master...boldtrn:debug_1127?expand=1). I don't like the added overhead, but right now I don't know a better solution? Also I think that the 10 meters are a bit heuristic and can match more than one VirtualEdge, which sometimes might be not ideal.

pa5cal:
> See [this route](https://graphhopper.com/maps/?point=50.17016%2C14.63381&point=50.17286%2C14.64287&block_area=50.17147%2C14.63683%2C100&ch.disable=true) vs. [this route](https://graphhopper.com/maps/?point=50.169037%2C14.631515&point=50.17393%2C14.64735&block_area=50.17147%2C14.63683%2C100&ch.disable=true) where the block_area parameter is identical and so it is expected that the point in the middle of the road section should be avoided in both cases. 

Can't reproduce the described bug. The blocked section is avoided by both routes.

karussell:
> Can't reproduce the described bug. The blocked section is avoided by both routes.

Yes, maybe there was a new junction introduced in OSM but the bug wasn't fixed in GH core.

karussell:
I think this is a duplicate of #1324.

PR #1485, opened 2018-10-16
add LocationIndexTree.query(BBox)

karussell:
This PR adds a new query(BBox) method. Currently we can fetch only nodes from one leaf and so with this querying areas is inefficient. Related to #1127 and #1324.

Furthermore a possibility to visualize the index 

image

and some description for the index is added. The description now makes clear that we have a Quadtree implementation with some special tuning. E.g. for bigger areas one node branches not only into 4 subtrees but can branch into 16. Previously also 64 was possible, but we removed this as this has no advantages in terms of storage, query or creation speed. (Measured it for Germany and world wide)

TODOs:

 - [x] ~~to make this correct we need to add +1/2 cell on every side of the BBox. The problem is that we do not add the node to both cells due to the [bresenham line to cell limitation](http://stackoverflow.com/q/24679963/194609)~~ Does not seem to be necessary (no problems e.g. in #1572).
 - [ ] make it working with Polygon too (use JTS Polygon e.g. via embedding in ours so that we can implement `Shape`)
 - [x] use PrepareGeometry  -> #1589
 - [ ] use intersect and contains!
 - [ ] use quality vs. speed properties like UNIQUE_NODES (just make sure nodes come to the Visitor just once), FILTERED_NODES (unique nodes plus filtered again through the shape), HIGH_PRECISION (filtered node but use fetchGeometry instead of just junction nodes)


reader-osm/src/test/java/com/graphhopper/reader/osm/GraphHopperOSMTest.java
```diff
@@ -149,6 +145,67 @@ public void testLoadOSMNoCH() {
...
+            }
+
+            @Override
+            public void onNode(int nodeId) {
+                double lat = na.getLatitude(nodeId);
+                double lon = na.getLongitude(nodeId);
+                if (bbox.contains(lat, lon))
+                    indexNodeList.add(nodeId);
+
+                // TODO should be done behind the scene?
```
karussell: Will make it possible via a `SimpleVisitor` class or something. Often this iteration has to be done again on the client side e.g. to fetch the edges and so this work would be an unnecessary overhead then.

michaz:
Nice, this closes some open ends for me. Is the current state mergeable, meaning, do we think that the bounding-box query does more or less what it should? Because I'd like to.

karussell:
Well, this is not fully ready unfortunately. Currently all nodes are returned where the tile intersects with the shape. This is ok for visualisation, but for other use cases we have to filter these nodes and call "shape.contains(node)" as a post processing step. Now the problem is that this could be done cheap like this or in a more expensive manner if we need high precision via "shape.intersects(edge.fetchWayGeometry)". Currently exploring the options on how to do this so that this also works for a Polygon (which basically means using JTS Polygon behind our Polygon class). But we could also bring it into a mergable state faster and I'll create a new issue for the open tasks.

michaz:
Oh, that's exactly what I need: "May include more nodes than are contained, but never fails to return something that _is_ contained". I think that's also how other implementations work.

karussell:
> Oh, that's exactly what I need: "May include more nodes than are contained, but never fails to return return a node that is contained". I think that's also how other implementations work.

Ah, ok. I think there are indeed two use cases. This one, which you need and we need for fast visualization. But also the other to e.g. block certain edges but not too many!

Maybe I'll bring this PR in a mergable form for the first use case and explicitly mention this and work later on the second "high precision" use case?

michaz:
> Maybe I'll bring this PR in a mergable form for the first use case and explicitly mention this and work later on the second "high precision" use case?

Yes, please! For me, this doesn't even mean that this is currently "unfinished" -- _filtering_ the result to the desired level of exclusivity is a _second_ task, and can happen outside of the current method. This one is finished.

michaz:
And this method can still narrow down what it returns later on without changing its promise.

karussell:
> filtering the result to the desired level of exclusivity is a second task

This is exactly where I'm unsure because the method `index.query(shape)` could also imply that this should be an exact filter and no post-processing work needs to be done.

michaz:
Well, "query" as a method name is vague enough that I think it can't be misunderstood as promising anything specific. ;-)

Let's add a Javadoc which explains what this method does and does not (not _how_ it does that), and it'll be fine.

karussell:
Ok :D 

Will do this ... let me just do some minor stuff.

karussell:
I did the following changes (let me know if this was too much):

 * moved the Visitor interfaces to the LocationIndex interface
 * moved the query method to the interface
 * removed the extra visualize method and made it possible via the Visitor setting isTileInfo (now Visitor has to be an abtract class as default methods are not possible in 7). onTile now directly includes the depth not the too-visualization-specific `width`.
 * avoid duplicate nodes (and edges for the EdgeVisitor)
 * introduced a new "queryBBox.contains" branch to avoid bbox checks on greater depths, should make the query slightly faster
 * renamed Shape.intersect to intersects, also to make later migrations to JTS easier
 * renamed `*cell*` to `*tile*`
 * made some methods non-public

core/src/main/java/com/graphhopper/storage/index/LocationIndex.java
```diff
@@ -62,4 +67,56 @@
...
+    /**
+     * This method explores the nodes in this LocationIndex with the specified Visitor. It guarantees to visit all
+     * unique nodes included in the queryBBox but it could visit more.
+     */
+    void query(BBox queryBBox, Visitor function);
+
+    /**
+     * This interface allows to visit every node stored in the leafs of a LocationIndex.
+     */
+    abstract class Visitor {
```
michaz: Just as a small remark: This Visitor is not very Java-8-friendly, and also a bit unusual. If I understand it correctly, as a client, I have to override `isTileInfo` if I want `onTile` to be called? I think interfaces are most clear when control flows in one direction only: An interface is either for me to call, or for me to implement and be called.

michaz: > An interface is either for me to call, or for me to implement and be called.

Okay, that's actually the case here. :-) But the _information_ flows both ways.

karussell: Yes, I tried with a separate QueryOptions class but this was not much prettier. Should I revert it back to this to make it easier with lambda?

michaz: Ah no, it's okay.

michaz: Thanks! I'm doing more Isochrone experiments with this right now... ,-)

Issue #471, opened 2015-07-22
GraphHopper.setCHEnable vs graph CH state


devemux86:
After #447 the ```setCHEnable(true)``` works also with non CH graphs, that could be a problem.

We should either configure CH automatically or notify the user about the wrong setting.

See also [here](https://github.com/graphhopper/graphhopper/pull/447#issuecomment-123821614) for the conversation.

devemux86:
In my mind that's somehow related to #285 "Inspect graph properties".

I mean in the end we should somehow not rely on setter methods for loading a graph, that could be handled automatically without failures for user convenience.

karussell:
That is true and should be done :)
But automating this is hard and configuring very easy :)

devemux86:
>But automating this is hard and configuring very easy :)

I have to agree to that! :)

karussell:
Will be fixed after #464 is merged

devemux86:
Peter after testing, I confirm that the loading of various graph types is working again fine.

karussell:
Thanks!

PR #1488, opened 2018-10-17
Allows using less ch weightings than loaded, fixes #1413.

### LINKED ISSUE #1413: Exception raised when loading a map generated with more CHWeightings than needed
Hi,
The following lines:
https://github.com/graphhopper/graphhopper/blob/5c980b4627578626268873f9424de35188187441/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java#L253-L254
raise an exception in case the map being loaded has been generated with more CHWeightings than the configured ones in the application.

e.g. a map is generated with graph.flag_encoders=car & bike and prepare.ch.weightings=shortest & fastest in graphhoper/config.properties, while the software reading the map is configured only with car/fastest.

It would be better to perform a loop checking if each CHWeighting returned by getCHWeightings() is included in loadedCHWeightings string one by one.

Very good repo

ammagamma:
Fixes #1413

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
+++ b/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
@@ -25,10 +25,7 @@ import com.graphhopper.util.EdgeExplorer;
 import com.graphhopper.util.EdgeIteratorState;
 import com.graphhopper.util.shapes.BBox;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 
 /**
  * This class manages all storage related methods and delegates the calls to the associated graphs.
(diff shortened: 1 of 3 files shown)


easbar:
There is still a small inconsistency: When there are CH profiles from a previous import we can configure less profiles and its fine (this was done in this PR). However, we cannot configure no CH profile at all, so its like this:

```
2 CH profiles imported, 2 configured -> ok
2 CH profiles imported, 1 configured -> ok
2 CH profiles imported, 0 configured -> error
```

I would say it should also be possible to configure no profile at all? However it looks like in #471 this was purposefully prevented...

Btw for LM its entirely different: Modifying `config.yml` both allows loading less or none of the previously configured LM profiles and its even possible to add LM profiles in `config.yml` which are then added to the graph folder during the next import.

Maybe some of the confusion arises from CH adding a default CH profile (calling `GraphHopper.setCHEnable` adds a single CH weighting)?.

easbar:
I fixed this here: 0a43887654d1b1786fa93a111d6862942d89a176. Its now OK to configure 0 CH profiles when previously there were some. Its still not possible to add additional CH profiles.

PR #1489, opened 2018-10-21
Replace String concatenation with StringBuilder at GhRequest#toString.

Azbesciak:
Also prevents NPE while transforming points lists.


karussell:
Thanks!

Do you see measurable speed improvements somewhere? (IMO it is harder to read, see this discussion: https://stackoverflow.com/q/1532461/194609)

> Also prevents NPE while transforming points lists.

Do you have a test showing the NPE?

Azbesciak:
To be honest, I just applied Intellij hint - IMO any improvement may be noticed on very huge scale, but this `toString` is not so offen called, so it is rather "best practice" refactor.

I trully didn't found any case when this may occur at gh server (maybe possible to pass null in `RouteResource#doGet`, but then you get deserialization error). However, it is public API (`GHRequest(points)`) so it should be prevented.
At least I had such situation :).

karussell:
If there is no issue and no failing test then I prefer to avoid a change. IMO it is better to read.

>  However, it is public API (GHRequest(points)) so it should be prevented.

If points is null then the NPE will still a problem here:

> for (GHPoint point : points) {

Azbesciak:
> If points is null then the NPE will still a problem here

I don't think so. Iterator may return null.
image

> What do you mean with "At least I had such situation :)"?

I'm logging requests (among others) when exception occurs.
Occurred because of request with null in the list of points.

karussell:
> Occurred because of request with null in the list of points... And also it caused NPE because of toString

If you can implement a test reproducing the issue then this would help.

api/src/test/java/com/graphhopper/util/GHRequestTest.java
```diff
@@ -101,6 +101,11 @@ private void compareFavoredHeadings(GHRequest request, List<Double> expected) {
             assertEquals(ind + " favored Heading does not match" + expected.get(ind) + " vs ." + favoredHeading,
                     expected.get(ind), favoredHeading, 0.01);
         }
+    }
 
+    @Test
+    public void testSerializationOnNullPoints() {
+        GHRequest ghRequest = new GHRequest(Arrays.asList(null, null, null, null));
```
karussell: To me a list with a null point is a bug, or what would be a use case?

chollemans:
In my opinion the toString function should really be used here. Your now concating the GHPoint to a String which is really confusing. The `toString()` function is ment for this. In case you are calling `toString()` on a object that is NULL and your trying to fix it there IMO your not fixing the problem at it's source. If this still needs to be fixed and `points` may not be null I'd suggest adding e.g. `@NotNull` to the field.

Azbesciak:
> Your now concating the GHPoint to a String which is really confusing. The toString() function is ment for this

So, in your opinion all from Scala, Groovy, Kotlin etc were wrong doing String interpolation? Most of the world works in this way. Also, are you using `toString` each time during concatenation?
You always may replace it with null safe `String#valueOf` - fell free to make changes in this PR; I just want to make it works.

The only way to prevent nulls in this option is to validate all points during construction - however, reference to the original list should be lost - this list may be mutable.
> I'd suggest adding e.g. @NotNull to the field

IMO this annotation is ment to say that this field cannot be null, not its inner elements.
as said for [`javax.validation.constraints.NotNull`](https://docs.oracle.com/javaee/7/api/javax/validation/constraints/NotNull.html):
>The annotated element must not be null. Accepts any type.

I think that `toString` should not throw exception in any case. Moreover, it should be suitable for debugging purpose, not the reason for it.

chollemans:
> So, in your opinion all from Scala, Groovy, Kotlin etc were wrong doing String interpolation? 

That's not what I said, I think in the case of a plain java project you should follow it's language standards, which are the most readable.

> The only way to prevent nulls in this option is to validate all points during construction - however, reference to the original list should be lost - this list may be mutable.

Good thoughts but sounds indeed like more work than needed.

> IMO this annotation is ment to say that this field cannot be null, not its inner elements.
> as said for [`javax.validation.constraints.NotNull`] https://docs.oracle.com/javaee/7/api/javax/validation/constraints/NotNull.html):

I was not paying attention, your completely right!

> I think that `toString` should not throw exception in any case. Moreover, it should be suitable for debugging purpose, not the reason for it.

I'm not quite sure. If `toString` throws a NPE you can immediately see in the stracktrace where it is originated from. But because this is a dependency it might not so neat to throw a NPE in your face. Your Implementation is in my opinion (now I've heard more about it) a proper solution.

Azbesciak:
@karussell ?

Azbesciak:
Currently it will not ignore, rather will cause another problem and hide actual. But ok.

karussell:
How can something "hide" and "not ignore" at the same time?

What I meant is: better to throw a late NPE than hiding this.

Azbesciak:
> How can something "hide" and "not ignore" at the same time?

Exception may be also suppressed. That is what I mean saying *"hide actual"*, because you even won't be informed about it, when your code will throw exception, because you wanted it to do so.
```
java.lang.NullPointerException
	at com.graphhopper.GHRequest.toString(GHRequest.java:272)
...
```
This is the only message which you will receive - no real exception. Why so?
Because I have such code in exception handling (kotlin):
```kotlin
 private fun <T: Response> prepareLogMessage(request: GHRequest, response: T) =
                """Error:
                |    request: $request
                |    response: $response
                |    stacktraces:
                |    ${formatStackTrace(response)}
                """.trimMargin()
```
I'm also using [Klogging](https://github.com/shafirov/klogging) for logging. This is code which handles exceptions durring logging:
```kotlin
internal inline fun (() -> Any?).toStringSafe(): String {
    try {
        return invoke().toString()
    } catch (e: Exception) {
        return "Log message invocation failed: $e"
    }
}
```
which will result in 
```
Log message invocation failed: java.lang.NullPointerException
```
Helpful, isn't it?
Even if it would print detailed stacktrace of this NullPointer, you still need to see external code, not exact reason which would be otherwise:
```
request: null; null (Hints:{instructions=false, avoid_payments=false, alternatives=1, vehicle=car, encoded=false})
    response: RouteResponse(isEncoded=false, hasInstructions=false, routes=[], errors=[java.lang.NullPointerException])
    stacktraces: ...
```

Issue #1487, opened 2018-10-17
Make limits for isochrone configurable


karussell:
There are a few limits hardcoded that we could make configurable. See [discussion](https://discuss.graphhopper.com/t/50km-limit-in-isochrone-api-calls/3635).

karussell:
Have thought about this a bit and probably the best is to remove the limits and ensure those limits in a proxy layer if required at all. They were there to avoid many requests overloading the server and bringing it down but even with the current limits this is still possible if the request count is enough, so the limits are a bit useless.

chollemans:
Allright, I've looked a bit trough the code and could find 2 kind of limit's. I found a `maxMeter` limit and a `maxSeconds` limit. Both of them have a check which throw exceptions when violated. I'll remove the limit variables and the if statements checking if a limit is violated and make a PR for it :)

PR #1491, opened 2018-10-25
Removed redundant violation checks of maxMeter and maxSeconds in isohrone API

chollemans:
Fixes: #1487


chollemans:
Apparently tests are failing with maven but not with intellij, I'll fix those later

karussell:
Thanks!

Issue #1370, opened 2018-05-14
Make OSM import easier customizable


Nakaner:
I am writing a [railway router](http://github.com/geofabrik/railway_routing) (on the tracks, not on timetables) and talked about it on FOSSGIS 2018 ([video](https://media.ccc.de/v/2018-5257-eisenbahnrouting_mit_graphhopper#t=739) in German, I will give a [talk](https://2018.stateofthemap.org/2018/T071-Routing_on_rails_with_OpenStreetMap/) at SotM).

I am currently implementing more features to improve routing results. My router adds custom turn restriction between two edges on all nodes if the angle between them is not soft (only slight left, straight on and slight right are permitted). However, some of these nodes need a special treatment. If an OSM node is tagged with `railway=railway_crossing`, other limits for the angles should apply (only straight on permitted). I think that the `OSMReader` class is the right location to populate a set of internal node IDs of OSM nodes with `railway=railway_crossing`. The turn angle evaluation code would check if a nodes is in that set.

However, both `processNode()` and `writeOsm2Graph()` are `private`. All interesting methods of that class are either `private` or don't have a visibility modifier at all. I can't just derive my own reader class from `OSMReader` as I did with `GraphHopperOSM` class (my derivation is called [`RailwayHopper`](https://github.com/geofabrik/railway_routing/blob/master/src/main/java/de/geofabrik/railway_routing/RailwayHopper.java)). I came to the point now where my only choice seems to be to maintain my own (slightly modified) fork of GraphHopper to avoid reading the input file a third time because the `OSMReader` class is not extensible by third-party developers.

A simple solution would be to open this class and make more methods protected. This comes at the cost of a larger public API and difficulties to change things later because the API is  bound too tightly to the implementation (see [comment](https://github.com/graphhopper/graphhopper/pull/1336#discussion_r180881729) by @karussell). I also think that this would not be good design.

@karussell mentioned hooks as an alternative. They are an alternative but they come with disadvantages as well. They are less flexible and it has to be thought out where to offer which hooks/callbacks. Hooks added now might serve my needs but slightly different requirements might lead to the need of different hooks. Intensive usage of `protected` offers more freedom and does not limit users of the library to the possible use cases the developers of the library could imagine. My use cases are quite strange and the number of users of such a broad API is very small. That's why I think that we should make more methods `protected` and mark such methods as "no stability guaranteed", i.e. they might change without prior notice or deprecation.

What do the maintainers of GraphHopper prefer?

[...]

PR #1492, opened 2018-10-26
Only update the location if it was empty

efranqueiro:


karussell:
The assumption is that init overwrites all local variables. So instead of changing this, we should better document that init should be called before calling setGraphHopperLocation. Would this help in your case too?

efranqueiro:
For my specific case, I can use `setMinNetworkSize()`, but in general, a builder pattern would be much better.
Your suggestion also works for me, but it should also mention that it must be called after `setDataReaderFile()`, then (see below).
But I don't really like that. There shouldn't be a dependency in the order you call the setters and `init()` (other than perhaps `init()` being the first or last call, but not in between 2 specific setters).

This works
```
			GraphHopper graphHopper = new GraphHopperOSM()
					.setDataReaderFile(osmFile)
					.setEncodingManager(new EncodingManager(profile))
					.init(new CmdArgs().
							put("prepare.min_network_size", "200").
							put("prepare.min_one_way_network_size", "200"))
					.setGraphHopperLocation(workDir)
					.forServer();

```

These don't
```
			GraphHopper graphHopper = new GraphHopperOSM()
					.setDataReaderFile(osmFile)
					.setGraphHopperLocation(workDir)
					.setEncodingManager(new EncodingManager(profile))
					.init(new CmdArgs().
							put("prepare.min_network_size", "200").
							put("prepare.min_one_way_network_size", "200"))
					.forServer();
			GraphHopper graphHopper = new GraphHopperOSM()
					.setEncodingManager(new EncodingManager(profile))
					.init(new CmdArgs().
							put("prepare.min_network_size", "200").
							put("prepare.min_one_way_network_size", "200"))
					.setDataReaderFile(osmFile)
					.setGraphHopperLocation(workDir)
					.forServer();
			GraphHopper graphHopper = new GraphHopperOSM()
					.setGraphHopperLocation(workDir)
					.setEncodingManager(new EncodingManager(profile))
					.init(new CmdArgs().
							put("prepare.min_network_size", "200").
							put("prepare.min_one_way_network_size", "200"))
					.setDataReaderFile(osmFile)
					.forServer();
```

We could be to force `init()` to be the first by not throwing an error if there's no location data set, and leave that to the validation in `importOrLoad()`. But this wouldn't be fail-fast though. Or force it to be the last call but keeping sane values if they are present.

efranqueiro:
A config object with sane defaults when possible, and some sort of validation, seems like a sensible solution to me. Same with a builder.

Azbesciak:
Just a note - if you are going to make a Builder for graphhopper, just remember about their subclasses - maybe it is a good moment to prefer composition over inheritance (for data readers for example)? IMO this Builder should be generic and return given subclass to allow Inheritance also.

karussell:
> prefer composition over inheritance 

Yes, this would be another goal and related to #1370

> this Builder should be generic and return given subclass to allow Inheritance also

Isn't this the opposite of what you requested?

Azbesciak:
> Isn't this the opposite of what you requested?

Yeah, I am aware of it, but there is a need (as I think) for extending functionality of gh over `route` only.
So, there are two ways:
- make gh some kind of dto which will accumulate graph, algs etc, and extract `route` outside of it, to allow extensions.
- keep gh open and make it in some kind immutable from outside.

I think second option will be easier (also in use-some kind of API) and maybe cleaner, but it requires Inheritance - users will get some prepared base and add what their want. It is better than current solution, because now you can't just add your functionality ahead of graph processing (from shp/pbf etc). 

Those are just a quick thoughts so...

karussell:
Will be done via #1895.

PR #1507, opened 2018-12-19
Android sample + JTS working

devemux86:
- [VTM 0.10.0](https://github.com/mapsforge/vtm) OpenGL vector map library
- Update Android sample build system
- Downgrade JTS for work on Android


android/app/build.gradle
```diff
@@ -35,22 +36,23 @@ configurations.all {
...
+    implementation 'org.mapsforge:vtm:0.10.0'
+    implementation 'org.mapsforge:vtm-android:0.10.0'
+    implementation 'org.mapsforge:vtm-android:0.10.0:natives-armeabi-v7a'
+    implementation 'org.mapsforge:vtm-android:0.10.0:natives-arm64-v8a'
+    implementation 'org.mapsforge:vtm-android:0.10.0:natives-x86'
+    implementation 'org.mapsforge:vtm-android:0.10.0:natives-x86_64'
+    implementation 'org.mapsforge:vtm-jts:0.10.0'
+    implementation 'org.mapsforge:vtm-themes:0.10.0'
+    implementation 'com.caverock:androidsvg:1.3'
+    implementation 'org.locationtech.jts:jts-core:1.15.1'
```
karussell: Is this required if GH core contains the dependency to jts-core?

karussell: Ok, will probably just manually add a commit with this dependency change and then we can wait until first snapshot is released.
@michaz are there any objections from your side to downgrade from 1.16 to 1.15.1?

karussell: @michaz are there any objections from your side to downgrade JTS from 1.16 to 1.15.1?

michaz: @karussell Not that I know of. I think everything I use in JTS has been there forever.

karussell: @devemux86 have now made the change and the snapshot is deployed

devemux86: Then can remove the 2 lines:
`exclude group: 'org.locationtech.jts', module: 'jts-core'`
...
`implementation 'org.locationtech.jts:jts-core:1.15.1'`

devemux86: @karussell now should be ok.

PR #1516, opened 2019-01-11
Makes Measurement the default main in gh tools jar with dependencies.

easbar:
The Import class is gone since dropwizard refactoring, so 
```bash
java -jar  ...-with-dependencies.jar
```
yields `ClassNotFoundException`.
I think it is still useful to have the jar with dependencies and changed the main class to Measurement.


karussell:
Thanks!

PR #1518, opened 2019-01-14
Removes dir parameter from PrepareContractionHierarchies constructor.

easbar:
`AbstractNodeContractor` uses a `DataAccess` to keep track of originalEdges that are used for node priority calculation during CH preprocessing. originalEdges are an implementation detail of the CH preprocessing and are only needed temporarily and never get flushed to disk. Therefore I think it is better to create the `DataAccess` inside `AbstractNodeContractor` as it is done here. This allows removing the dir parameter from the `PrepareContractionHierarchies` and `NodeContractor` constructors and thus decouples the algorithm code from the storage. In a next step I think it would be quite easy to replace `GraphHopperStorage` with the `Graph` interface such that it will be possible to run CH with any graph implementation implementing the `(CH)Graph` interface.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
+++ b/core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
@@ -264,7 +264,7 @@ public class CHAlgoFactoryDecorator implements RoutingAlgorithmFactoryDecorator 
 
         for (Weighting weighting : getWeightings()) {
             PrepareContractionHierarchies tmpPrepareCH = new PrepareContractionHierarchies(
-                    new GHDirectory("", DAType.RAM_INT), ghStorage, ghStorage.getGraph(CHGraph.class, weighting), traversalMode);
+                    ghStorage, ghStorage.getGraph(CHGraph.class, weighting), traversalMode);
             tmpPrepareCH.setParams(pMap);
             addPreparation(tmpPrepareCH);
         }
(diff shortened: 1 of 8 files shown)


karussell:
> are an implementation detail of the CH preprocessing and are only needed temporarily and never get flushed to disk

If the default graph data access is MMAP, then we would now request a large portion of heap memory. I'm not sure how important this would be for existing users but IMO having a default should be considered where possible. 

E.g. we explicitly ignore the default for the elevation provider and always use MMAP if not otherwise configure. So we could make this configurable too (and use RAM as default) or should we use graphHopperStorage.getDirectory instead?

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
```diff
@@ -264,7 +264,7 @@ public void createPreparations(GraphHopperStorage ghStorage, TraversalMode trave
 
         for (Weighting weighting : getWeightings()) {
             PrepareContractionHierarchies tmpPrepareCH = new PrepareContractionHierarchies(
-                    new GHDirectory("", DAType.RAM_INT), ghStorage, ghStorage.getGraph(CHGraph.class, weighting), traversalMode);
```
easbar: so far we were using RAM_INT.

karussell: Uhm, yeah. Let's merge ;)

easbar:
Ok if this is required we can make original_edges storage configurable, but I am not sure if this really makes sense, for example we a large portion of memory in DijkstraOneToMany also (for example int[] parents).
If it should be configurable I would prefer doing this via the CHGraph interface, because I was hoping to make PrepareContractionHierarchies unaware of GraphHopperStorage.

PR #1524, opened 2019-01-17
Faster alternative routes for longer routes

baumboi:
This pull request changes the algorithm for alternative routes. It now supports CH and has the possibility to be precomputed. Since precomputing needs the graph to be split into different parts, I added a GraphPartition class which may also be useful for other purposes as well.

The whole algo is based on this paper:
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.665.4573&rep=rep1&type=pdf
This poster summarizes everything quite well and has some nice visualisations
https://algo2.iti.kit.edu/download/ls-csarr-13-poster.pdf

I did a lot of testing on this algorithm and found out, that the new algo with CH, but without preparation should take about 10x the time the classic CH needs to find a route (~15 ms vs ~1.5 ms). In 40% to 50% of all cases at least one alternative is found in Saarland (about 100K nodes, the biggest graph my computer can do), but this number should increase with the size of the graph.
With precomputing, the success rate is slightly higher and the algorithm is 2x to 3x faster than without (~6 ms). For bigger graphs the prepared algo should get even faster, relatively to the unprepared one, as the number of precomputed viaNodes should stay the same, while the number of during the request computed contact nodes increases with the size of the graph.
All known factors of the old algorithm will stay the same, except for `minPlateauFactor` which isn't needed anymore. I lowered `maxWeightFactor` from 1.4 to 1.25 and increased `maxShareFactor` from 0.6 to 0.75 in order to get better results. With the old values some alternatives did not seem very reasonable. Using theese new values, bad alternatives became very rare (maybe about 1 out of 25). I also increased `maxPaths` from 2 to 3. If you want better query times you can lower it again together with `additionalPaths`.

Precomputation itself takes about 5x to 10x the time preparing CH needs (~25 s vs ~180 s), but it heavily depends on the precision and the amount of areas the graph is split into. Precision should be fine at a value of 5 (default) or 4, meaning 5 (or 4) nodes are chosen on average per area to run alternative route requests from and to during precomputation. Since this has to be done for nearly each pair of areas, about
(precision ^ 2) * (areas ^ 2)
requests need to be done to prepare alternative routes. I can't really tell how precomputing times will scale with the size of the graph. Lowering the amount of areas will drop preparing times a lot, but will also result in bigger area sizes. And because the prepared algorithm can only be used for nodes in two different, not directly connected areas, this may result in higher query times for short routes. To get the best out of it, we should test up to which graph size the unprepared algorithm has good query times and match the amount of areas accordingly. (the unprepared algo takes 15 ms on average in Saarland, so good query times without preparing should be able to up to maybe 2 million nodes?)

I'm really excited to know how the algorithm performs on a much larger graph like Germany.

I wasn't quite sure on how to implement the alternative route preparation with CH, so I added the method `createPrepareAlternativeRoute(Graph graph, AlgorithmOptions opts)` to `PrepareContractionHierachies`. Is there a better way to do this?
Currently, to let my algorithms know if they run CH, I'm looking whether the `Graph` is a `CHGraph` and the `Weighting` is a `PreparationWeighting` like in line 64 in `PrepareAlternativeRoute`. Is there another way to do this?
Moreover, in `AlternativeRouteAlgorithm` in line 193 to 205, I don't know how to get a "normal" `Weighting` out of a `PreparationWeighting`. I'm sure there is a better solution than this^^


karussell:
@baumboi really nice - thanks a lot!

I'm curious to try this out. Can you try to make it compile? See the error messages at the moment [here](https://travis-ci.org/graphhopper/graphhopper/jobs/480644404#L750). Highly likely you started your work from an older master version and this needs an update. The following should make it work:

 * [update the master branch of your fork](https://help.github.com/articles/syncing-a-fork/)
 * go to the `alternative` branch: `git checkout alternative`
 * merge the changes of `master` into the `alternative` branch: `git merge --no-ff master`
 * now you need to resolve conflicts and run tests
 * commit and push

baumboi:
Now it should work. It was indeed the old master version, but I also forgot to change two tests in GraphhopperIT.

karussell:
Thanks, this is yet another improvement. I was not able to find the bug. Still there seems to be "obviously suboptimal" alternatives. E.g. this route: http://localhost:8989/maps/?point=52.491874%2C13.36092&point=52.548697%2C13.367443&vehicle=car&algorithm=alternative_route

image

The "left suboptimality" is one as can be shown with the following subpath: http://localhost:8989/maps/?point=52.516012%2C13.347015&point=52.530948%2C13.347015

image

The "middle suboptimality" is also one: http://localhost:8989/maps/?point=52.519805%2C13.355802&point=52.521861%2C13.360662

image

Could it be that there is still something wrong with picking the alternative node set? E.g. in the "middle suboptimality" case I would have expected that a candidate on Alice-Berend-Str is always overruled by a candidate on the parallel Lüneburge Str. 
Another problem can be if a contact node is directly chosen without looking at its weights. I think (not 100% sure) that the contact node can only be used if the sum of the from+to weights in the queue are bigger then the from+to weights of that contact node.

karussell:
Although this PR was faster it was replaced by #1722. We'll revisit this PR to e.g. improve perf and for fixing the "motorway loop"-bug.

PR #1526, opened 2019-01-19
Adds getEdges() to graph interface. No more GraphHopperStorage in PrepareCH.

easbar:
So far `PrepareContractionHierarchies` was directly coupled to `GraphHopperStorage`. To be able to use the same algorithm with another `Graph` implementation I did the following: 

-  replaced the `GraphHopperStorage` field with a reference to a `CHGraph`. I had to add `getOriginalEdges()`  and `disconnect()` to `CHGraph`.
- removed the `freeze()` call, which is already done in `GraphHopper#prepareCH`. I only had to add it in two tests.

I also added `getEdges()` to `Graph` which can be used instead of `.getAllEdges().length()`, because I thought this is more consistent with `getNodes()`.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/QueryGraph.java
+++ b/core/src/main/java/com/graphhopper/routing/QueryGraph.java
@@ -531,6 +531,11 @@ public class QueryGraph implements Graph {
     }
 
     @Override
+    public int getEdges() {
+        return virtualEdges.size() + mainEdges;
+    }
+
+    @Override
     public NodeAccess getNodeAccess() {
         return nodeAccess;
     }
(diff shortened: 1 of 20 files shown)


easbar:
For turncosts_ch we will need another method like `createOriginalEdgeExplorer()` that only traverses edges, but not shortcuts in `CHGraph`. Previously this was done using `ghStorage.createEdgeExplorer()` instead of `prepareGraph.createEdgeExplorer()`. But I think this would not be a problem ?

easbar:
> If just for CHGraph this is no problem.

yes just for chgraph.

PR #1531, opened 2019-01-29
Do not create loop-edges

### LINKED ISSUE #1525: Backward-only loops make problems with edge-based routing.
When we add a loop edge that only has the `bwd` but not the `fwd` bit set we are getting problems with edge-based routing. The loop edge will not be found by forward searches (using `DefaultEdgeFilter.outEdges`) so if there is a path that requires a loop (e.g. because of a restricted direct turn at the loop node) it will not be found or another (a sub-optimal) route will be found. With a bidirectional algorithm we would discover the loop in the backward search, but right now there will be a runtime exception in `AbstractWeighting`. Also the forward search would have to 'wait' at the loop for the backward search which can harm performance. Its probably questionable if a bwd-only loop is ever a real problem, but it is possible in principle. I think the current import does not produce it. It would be needed if we wanted to get the reverse geometry of a loop.
See tests here: https://gist.github.com/easbar/06e7a9c71888fe147887e1d83dde9fe1
Also see discussion in #1247.

karussell:
This PR is the minimal change to fix #1525 (problems with edge-based algorithms). The fix avoids loop-edges from the import side. It is still possible to create loops via the Java API e.g. to attach properties like currently done in reader-gtfs.

With this PR no change needs to be done in DefaultEdgeFilter like it was necessary for #1247.

Do we need an explicit test that the OSM import results in no loop-edges?

Please note, the changes uncovered a bug in the landmark algorithm regarding the QueryGraph, which will be fixed in a separate issue. And so the test was slightly modified so that the original situation for #1014 was kept.

The other possible fix for #1525 is quite involved. Still, if results of the current fix make too many problems elsewhere, we'll have to go the more complex way, which requires a few more things:

 * the biggest change is to store the edge key in the SPTEntry, best would be if we can remove the edgeId to avoid increased memory usage for non-CH algorithms
 * introduction of graph.init(edgeKey)
 * make edge.detach working for loops too
 * the loop-edge needs to be returned twice on `edgeIter.next()` (note that the [tried implementation in this PR](https://github.com/graphhopper/graphhopper/commit/b95c6d1bbd7c84b493e8b9f3e79f56752e5dc6a7#diff-2d1b4844ee0fa4d7280ee50d5029188cR1031) has a bug if reverse=true from a previous edge state)
 * no changes in DefaultEdgeFilter are necessary

### DIFF
--- a/client-hc/src/test/java/com/graphhopper/api/GraphHopperWebIT.java
+++ b/client-hc/src/test/java/com/graphhopper/api/GraphHopperWebIT.java
@@ -87,7 +87,7 @@ public class GraphHopperWebIT {
         path = paths.get(0);
         isBetween(20, 30, path.getPoints().size());
         isBetween(800, 900, path.getDistance());
-        assertTrue("expected: " + path.getDescription().get(0), Arrays.asList("Jacobistraße", "Ludwig-Gercke-Straße").contains(path.getDescription().get(0)));
+        assertTrue("expected: " + path.getDescription().get(0), Arrays.asList("Jacobistraße", "Ludwig-Gercke-Straße", "Eichendorffplatz").contains(path.getDescription().get(0)));
     }
 
     @Test
--- a/core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
+++ b/core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
@@ -52,6 +52,7 @@ public class DefaultEdgeFilter implements EdgeFilter {
 
     @Override
     public final boolean accept(EdgeIteratorState iter) {
+        if(iter.getBaseNode() == iter.getAdjNode()) return iter.isForward(encoder) || iter.isBackward(encoder);
         return fwd && iter.isForward(encoder) || bwd && iter.isBackward(encoder);
     }
 
--- a/core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
+++ b/core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
@@ -52,7 +52,7 @@ public class DefaultEdgeFilter implements EdgeFilter {
 
     @Override
     public final boolean accept(EdgeIteratorState iter) {
-        if(iter.getBaseNode() == iter.getAdjNode()) return iter.isForward(encoder) || iter.isBackward(encoder);
+        if(iter.getBaseNode() == iter.getAdjNode()) throw new IllegalStateException("loop cannot happen");//return iter.isForward(encoder) || iter.isBackward(encoder);
         return fwd && iter.isForward(encoder) || bwd && iter.isBackward(encoder);
     }
 
--- a/core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
+++ b/core/src/main/java/com/graphhopper/routing/util/DefaultEdgeFilter.java
@@ -52,7 +52,6 @@ public class DefaultEdgeFilter implements EdgeFilter {
 
     @Override
     public final boolean accept(EdgeIteratorState iter) {
-        if(iter.getBaseNode() == iter.getAdjNode()) throw new IllegalStateException("loop cannot happen");//return iter.isForward(encoder) || iter.isBackward(encoder);
         return fwd && iter.isForward(encoder) || bwd && iter.isBackward(encoder);
     }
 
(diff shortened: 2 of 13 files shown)


easbar:
> It is still possible to create loops via the Java API e.g. to attach properties like currently done in reader-gtfs.

Maybe we can document this somewhere, e.g. in `BaseGraph#edge(int, int)` ? Or even print a warning (probably not because loops are needed for gtfs).

> Do we need an explicit test that the OSM import results in no loop-edges?

Will do.

> The other possible fix for #1525 is quite involved. 

Good to keep this in mind, but probably shouldn't do this before #1447.

easbar:
> With this PR no change needs to be done in `DefaultEdgeFilter` like it was necessary for #1247.

I am not so sure about this. Even when we exclude loops in the OSM import there can still be loop shortcuts (for example the additional pillar node will be contracted right at the beginning, immediately creating a loop shortcut). The other thing I am a bit worried about is that I spent quite a bit of time to get CH work with loops and it might be a bad idea to remove the corresponding code and tests in case at a later time we would like to allow loops again. Is there any reason not to use the current version of `DefaultEdgeFilter` in master ? With this PR all tests in `turncost_ch` just passed on my laptop, except
`GraphHopperGeocodingIT#testForwardGeocodingNominatim` failed with java.net.SocketTimeoutException` (probably another issue?).

karussell:
Thanks!
> Maybe we can document this somewhere, e.g. in BaseGraph#edge(int, int) ? 

Good idea!

>> Do we need an explicit test that the OSM import results in no loop-edges?
> Will do.
Thanks!

> The other thing I am a bit worried about is that I spent quite a bit of time to get CH work with loops and it might be a bad idea to remove the corresponding code and tests in case at a later time we would like to allow loops again

Couldn't we somehow inject the tweaked EdgeFilter for these tests? Or how many tests will fail? (As those tests are still in git they wouldn't be 'lost'.)

The only problem I have with this code in the default filter is that it feels a bit dirty (as it is not necessary *now*)

> With this PR all tests in turncost_ch just passed on my laptop, except GraphHopperGeocodingIT#testForwardGeocodingNominatim failed with java.net.SocketTimeoutException` (also fails here, probably another issue?)

Uhm, yes. Unrelated & should be fixed.

karussell:
> (As those tests are still in git they wouldn't be 'lost'.)

To make sure they are really preserved (and in master to make them easy to find), we could merge the turn cost branch with the tweaked DefaultEdgeFilter, then remove the tweak and also all failing tests.

easbar:
> Even when we exclude loops in the OSM import there can still be loop shortcuts

This is the real issue here, see https://github.com/graphhopper/graphhopper/pull/1247/commits/be4e6ccd4ef0828bdd6e1d15fc0c6bb2295d2af0#diff-094ce488919a6ffa3ad06adbc7df7213R409

There will always be loop shortcuts and the backward search can be blocked by them so no shortest path is found. 

> To make sure they are really preserved (and in master to make them easy to find), we could merge the turn cost branch with the tweaked `DefaultEdgeFilter`, then remove the tweak and also all failing tests.

Yes, this is a good idea would be good to have these changes in one commit, but like I said the real problem is that the routing does not work without the changes in the filter at the moment even with the fixed OSM import. One possibility would be to apply the special condition in the filter only on loop-*shortcuts*, but this would probably be even dirtier ;)

easbar:
Btw the shortcut in this test is 'forward' because all shortcuts in edge-based CH are unidirectional (forward only) at the moment. Maybe (but just maybe, no idea atm) another solution would be to set all loop shortcuts fwd&bwd.

karussell:
> One possibility would be to apply the special condition in the filter only on loop-shortcuts, but this would probably be even dirtier ;)

Got it :) ... so yes, we need the special tweak in the EdgeFilter.

> Maybe (but just maybe, no idea atm) another solution would be to set all loop shortcuts fwd&bwd.

It will likely still break if we read properties in the wrong direction. Now I think this can be only fixed without the tweaked EdgeFilter with the more involved changes described above.

karussell:
@easbar Uhm, sorry, this is unsatisfying: I merged the PR via squashing to have just one comment **and** the commit contains "Signed-off-by: <name+email>" as you can see via git log but your work still does not appear via the icon. 

It looks like we have to do "Co-authored-by: <name+email>" instead, which is a bit unusual:

https://help.github.com/articles/creating-a-commit-with-multiple-authors/

easbar:
Ok no worries :)

spalsher:
it not generating 'Names' file after running the script. Please help, i have tried maps from different countries but same result.

Issue #1509, opened 2018-12-28
Make CH contraction order re-usable.


easbar:
A large part of CH pre-processing consists of finding a useful contraction order of the nodes. Therefore I think it would be useful to be able to store and re-use a contraction order (a simple mapping from node ids to node levels). This would allow to re-calculate CH when an area is blocked or some road weights are changed. Of course the node ordering might be sub-optimal then, but I imagine it will still yield reasonable results in many cases if the graph does not change too much. Changing the node ordering does not affect the correctness of the routing, but only the query times. Also for different vehicles that usually use the same roads (like cars and trucks) I expect that the same node ordering can be applied even though the edge-weights are different. On the contrary I think it would not work as well to re-use the node ordering obtained for a fastest weighting for a shortest weighting (in which case the graph exposes a different (or hardly any) hierarchy), but it would still be interesting to try.

PR #1534, opened 2019-02-04
Allows reusing CH node ordering.

easbar:
As discussed in #1509  this PR allows re-using the node ordering in between CH contractions. Since we were already storing CH levels no additional DataAccess is needed and I only added a method to retrieve the inverse of the ch levels (switch array indices and levels).  A performance test on the map of Germany showed this:

|profile|contraction time|average query time|
|-|-|-|
|car|672s|0.97ms|
|motorcycle|445s|0.80ms|
|motorcycle with ordering from car|149s|0.86ms|
|car with ordering from motorcycle|>>672s|?|

The test shows that for example reusing the car ordering for motorcycle speeds up the preparation by a factor of about 3 and does not alter the query time too much. However, reusing the motorcycle ordering for car did not work at all, i.e. the preparation took (a lot) longer than when recalculating the ordering heuristically.
So I think in some cases re-using the node ordering can be beneficial, but the pre-processing is still too slow for a flexible scenario and it is hard to know beforehand whether it leads to good results. For example I do not understand why reusing the motorcycle ordering for car is so slow.

The shown query time is an average over 50k random queries without path unpacking and location lookup. For the profiles I used `FastestWeighting` with `Car/MotorcycleFlagEncoder`.

In any case using a fixed node ordering is quite useful for testing (previously in `ManualPrepareContractionHierarchies`).
I also added a `CHGraph#isFrozen()` check to `PrepareContractionHierarchies` to be a bit more defensive there.


core/src/main/java/com/graphhopper/storage/CHGraph.java
```diff
@@ -82,4 +83,8 @@
      * @return the number of original edges in this graph (without shortcuts)
      */
     int getOriginalEdges();
+
+    NodeOrderingProvider getNodeOrderingProvider();
+
+    boolean isFrozen();
```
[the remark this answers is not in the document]
easbar: I just had to put it out of the constructor because in many cases PCH is created before the graph is 'frozen'. So I moved the check to `doSpecificWork()`. Actually `isFrozen()` is probably not a good name either, because the graph is not 'frozen', but rather nodesCH is initialized so we can start adding shortcuts and set CH levels. So I renamed it to 'isReadyForContraction' or what do you think ?

core/src/test/java/com/graphhopper/routing/ch/PrepareContractionHierarchiesTest.java
```diff
@@ -363,6 +360,57 @@ public void testUnpackingOrder_Fastest() {
         assertEquals(IntArrayList.from(10, 0, 1, 2, 3, 4, 5, 6), p.calcNodes());
     }
 
+    @Test
+    public void testDisconnects() {
```
easbar: Ah this is unrelated to this PR and commited accidentally, but does not really hurt either ?!

core/src/test/java/com/graphhopper/routing/ch/PrepareContractionHierarchiesTest.java
```diff
@@ -363,6 +360,57 @@ public void testUnpackingOrder_Fastest() {
...
+                    }
+
+                    @Override
+                    public int getNumNodes() {
+                        return g.getNodes();
+                    }
+                });
+        prepare.doWork();
+        CHEdgeExplorer explorer = lg.createEdgeExplorer();
+        // shortcuts leading to or coming from higher level nodes should be disconnected
```
easbar: btw actually I thought this was quite interesting ;)

karussell: You mean if we store the node ordering there could be some more edges than before this change?

easbar: no like I said this is rather unrelated. I just pointed out here that we only disconnect shortcuts during CH preparation (because we can/do not change the base graph), but I think this definitely slows down the queries as we iterate many edges that lead to (or come from) lower level nodes anyway.

easbar: This took me quite some time to realize.

karussell: > (because we can/do not change the base graph),

As we have a special nodeCH DataAccess we store a different list of edges per node and we should be able to remove this per CH-profile. 

So this is the reason I was thinking the same: that we do this disconnect for base graph edges too.

> Sorry for the confusion I accidentally commited this test ;)

No problem. Maybe we should add this comment in the code instead of the test?

easbar: > As we have a special nodeCH DataAccess we store a different list of edges per node and we should be able to remove this per CH-profile.

Good point, but are you sure this will work ? We only store the pointer to the first edge in nodeCH so not sure how this will work for edges that are further behind in the linked list of edges ? I was more thinking in the direction of storing the edge connections separately and reusing the rest of the data in base graph (flags etc.) by edge id, but seems to be a bigger issue.

> Maybe we should add this comment in the code instead of the test?

done

karussell: Ah, yes. The linked list pointer (nodeCH) points into the shortcuts first and then the last entry of the shortcuts points into the original edges. So yes, duplicating this information would be a bigger issue and increase RAM usage further (but maybe faster as you said).

easbar:
Ok I ran a few tests with random weight (not speed) deviations (+/- the given percentage) on the map of Germany. I also added the test to this PR, TrafficChangeWithNodeOrderingReusingTest`.

|max weight deviation (%)| contraction time (s)| querying time (ms)| contraction time, with deviations, reused node ordering (s) | querying time, with deviations, reused node ordering (ms)|
|-|-|-|-|-|
|0|599|0.98|148|1.01|
|5|629|1.02|159|1.00|
|10|610|0.98|151|1.05|
|50|598|0.93|183|1.16|

So it looks like up to 10% weight deviations the query speed is hardly affected at all and even with 50% deviation it stays close to the original query speed. I guess for a more realistic scenario it would be better to use real traffic data instead of random weights (often whole roads will be blocked by traffic not single road segments).

Maybe a better model is to only consider *increases* of the weights (so the deviations do not cancel out each other). I tested this for Bavaria:

|max weight deviation (%)| contraction time (s)| querying time (ms)| contraction time, with deviations, reused node ordering (s) | querying time, with deviations, reused node ordering (ms)|
|-|-|-|-|-|
|0|103|0.53|27|0.52|
|1|103|0.55|25|0.48|
|5|102|0.51|26|0.49|
|10|101|0.50|26|0.49|
|50|101|0.50|28|0.55|

This did not change (rather confirmed) my conclusions above.

karussell:
Thanks for the fixes and of course for the analysis, this looks promising! So I'll squash&merge (?)

easbar:
>  So I'll squash&merge (?)

Yes, from my side this is ready.

karussell:
Really cool, thanks!

PR #1539, opened 2019-02-08
Update config-example.yml

### LINKED ISSUE #1536: Add logging to config example
The current example config comes without logging config which we should avoid and add:

```yml
logging:
  level: INFO

  loggers:
    "com.graphhopper": INFO

  appenders:
    - type: file
      timeZone: UTC
      currentLogFilename: logs/graphhopper.log
      logFormat: "%d{YYYY-MM-dd HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n"
      archive: true
      archivedLogFilenamePattern: ./logs/graphhopper-%d.log.gz
      archivedFileCount: 30
    - type: console
      timeZone: UTC
      logFormat: "%d{YYYY-MM-dd HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n"
```

karussell:
Fixes #1536

### DIFF
--- a/config-example.yml
+++ b/config-example.yml
@@ -157,15 +157,15 @@ logging:
     "com.graphhopper": INFO
 
   appenders:
-    - type: file
-      timeZone: UTC
-      currentLogFilename: logs/graphhopper.log
-      logFormat: "%d{YYYY-MM-dd HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n"
-      archive: true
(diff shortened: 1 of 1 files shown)


config-example.yml
```diff
@@ -149,3 +149,23 @@ server:
   - type: http
     port: 8990
     bindHost: localhost
+
+logging:
+  level: INFO
+
+  loggers:
+    "com.graphhopper": INFO
```
boldtrn: Is this part needed?

Issue #1540, opened 2019-02-11
Change Graph - not possible to revert blocked point


boldtrn:
If we block access to an edge, the forward and backward will be set to false. If we want to set this edge again to accessible, we execute a graph query in `GraphEdgeIdFinder#findClosestEdge` with the DefaultEdgeFilter, that requires the edge to be at least forward or backward.

See the referenced PR for an executable test.

PR #1541, opened 2019-02-11
Cannot set a blocked edge to accessible again

boldtrn:
This PR highlights the issue described in #1540.


core/src/main/java/com/graphhopper/storage/change/ChangeGraphHelper.java
```diff
@@ -93,7 +93,7 @@ private long applyChange(JsonFeature jsonFeature, FlagEncoder encoder) {
         BooleanEncodedValue accessEnc = encoder.getAccessEnc();
         DecimalEncodedValue avSpeedEnc = encoder.getAverageSpeedEnc();
         long updates = 0;
-        EdgeFilter filter = DefaultEdgeFilter.allEdges(encoder);
+        EdgeFilter filter = EdgeFilter.ALL_EDGES;
```
karussell: @boldtrn do you think this is a reasonable fix?

boldtrn: I actually had the same idea to approach this issue. 

The biggest issue I see with this is that when we try to block an edge, we might snap the wrong edge. For example one wants to avoid a road and puts an avoidance on that road, we might incorrectly snap for example the cycle way next to the road, that has no relevance for the car. Other than that I think this can work :+1:.

PR #1542, opened 2019-02-13
Edge CH speedup

karussell:


reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/RealtimeFeed.java
```diff
@@ -257,6 +257,11 @@ public Graph copyTo(Graph g) {
             public GraphExtension getExtension() {
                 throw new RuntimeException();
             }
+
+            @Override
+            public int getBaseNode(int edge, int adjNode) {
+                return 0;
```
karussell: This is a bit dangerous

core/src/main/java/com/graphhopper/storage/Graph.java
```diff
@@ -117,4 +117,6 @@
      * @return the graph extension like a TurnCostExtension
      */
     GraphExtension getExtension();
+
+    int getBaseNode(int edge, int adjNode);
```
easbar: Maybe `getOtherNode` would be a better name (and rename `adjNode` parameter to `node`)?

core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
```diff
@@ -159,8 +160,8 @@ protected int getOrigEdgeId(EdgeIteratorState edge, boolean reverse) {
 
     @Override
     protected int getTraversalId(EdgeIteratorState edge, int origEdgeId, boolean reverse) {
```
karussell: The name should be also getEdgeKey and we do only need the adjNode not `edge` itself here?

reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/RealtimeFeed.java
easbar: or better throw `RuntimeException` like in `getExtension` ? Btw have you considered adding mockito to core ?

core/src/main/java/com/graphhopper/storage/CHGraphImpl.java
```diff
@@ -261,6 +261,14 @@ public boolean isReadyForContraction() {
         return isReadyForContraction;
     }
 
+    @Override
+    public int getBaseNode(int edge, int adjNode) {
+        EdgeAccess edgeAccess = isShortcut(edge) ? chEdgeAccess : baseGraph.edgeAccess;
+        long edgePointer = edgeAccess.toPointer(edge);
+        int nodeA = edgeAccess.edges.getInt(edgePointer + edgeAccess.E_NODEA);
+        return nodeA != adjNode ? nodeA : edgeAccess.edges.getInt(edgePointer + edgeAccess.E_NODEB);
```
karussell: With edgeIterableImpl.selectEdgeAccess we would not need to duplicate this here but I see no chance to make this working atm.

core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
easbar: hmm for node based traversal the traversal id is the node id so I think `getTraversalId` makes sense (more general) ? But yes, let's replace the edge parameter with the adj node id.

core/src/main/java/com/graphhopper/storage/CHGraphImpl.java
easbar: we could add `EdgeAccess#getOtherNode` ?

reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/RealtimeFeed.java
karussell: > or better throw `RuntimeException` like in `getExtension` ?

Yes

> Btw have you considered adding mockito to core ?
You mean in the sources directly? I would strongly argue against it :) (It does too much magic, hard to debug, not good for portability etc)

core/src/main/java/com/graphhopper/storage/Graph.java
karussell: Good idea

core/src/main/java/com/graphhopper/storage/CHGraphImpl.java
karussell: Yes, this is probably the way to go

reader-gtfs/src/main/java/com/graphhopper/reader/gtfs/RealtimeFeed.java
easbar: oh right this is production code. but for tests I like mockito (just not power mockito) ;)

easbar: replaced with exception.

easbar: i replaced with exception

core/src/main/java/com/graphhopper/storage/Graph.java
easbar: done

core/src/main/java/com/graphhopper/storage/CHGraphImpl.java
easbar: edgeAccess already has getothernode. done.

easbar:
there are still some `api limit reached` errors in the tests.

karussell:
> there are still some api limit reached errors in the tests.

Probably increasingly more people are testing the code base (or some misusage ;)). Will fix.

Would it be ok to merge #1447 now - i.e. before this here?

core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
easbar: we cannot change the edge parameter with adj node id, because it is used by `AbstractBiDirAlgo`.

easbar:
> Would it be ok to merge #1447 now - i.e. before this here?

Yes sure, this here can wait.

core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
karussell: Ah, ok, I see.

karussell:
Ah, cool you already updated it - thanks!

karussell:
Do you think we can merge?

easbar:
yes

PR #1548, opened 2019-02-21
Introduce RoadClass, RoadEnvironment, CarMaxSpeed, Surface, Toll and more EncodedValues

karussell:
This PR shows how shared EncodedValues look like and how we populate them (via TagParsers). The prefix "OSM" indicates that such parsers can look differently for a different data source. A TagParser has not necessarily a one-to-one relationship to an EncodedValue. With the new mechanism we can have multiple EncodedValues and just one TagParser like e.g. for height, width and weight.

We likely need RoadClass, RoadEnvironment and a few more to be added by default. (As we could really need them for improving instructions, improving ETAs, finally fixing bugs like #713 for all vehicles, ...)

image

image

The following EncodedValues are available as path detail and could be used to influence the routing for all vehicle profiles (see [removed AvoidWeighting](https://github.com/graphhopper/graphhopper/pull/1548/commits/cb2c98301a2728c42f4eef096d3502808eb2535d)):

 * roundabout
 * road_class, road_class_link
 * road_environment
 * road_access
 * max_speed (earlier versions: car_max_speed)
 * surface
 * country
 * max_height, max_width, max_weight
 * toll

Done:

 - [x] refactor DataFlagEncoder to use those new EncodedValues
 - [x] ~improve calculating instructions with e.g. roadclass~ -> separate issue
 - [x] refactor Tunnel&Bridge interpolator to use RoadEnvironment and enable it for all vehicles!
 - [x] The new path details classes show that as soon as there is an EncodedValue, we can show it along the path, e.g. replace AverageSpeedDetails (see #1551)
 - [x] added toll and car_max_speed EncodedValue and parser
 - [x] add a few of these EncodedValues per default
 - [x] introduced a simple AvoidWeighting that supports toll, motorway, ferry etc avoidance (or preference)
 - [x] remove AvoidWeighting (move to separate issue)
 - [x] implement a loading mechanism like we have for flag encoders
 - [x] make it possible to enable storing those new fields without the DataFlagEncoder (Should be possible with `graph.encoded_values=surface,road_class`!)
 - [x] new EncodedValueFactory and TagParserFactory to provide custom builds also loading from configuration
 - [x] new SpatialRuleParser to provide country-based decisions to other vehicle profiles
 - [x] It could be even good to introduce a dependency of TagParsers. E.g. the SpatialRuleParser requires the info from RoadAccessParser. We could enforce this via checking the EncodedValueLookup in createEncodedValue.  Can be done with lookup.hasEncodedValue
 - [x] maxspeed returns 0 for default value (no sign) in Java API only, breaks backward compatibility => if fixed remove TODO NOW
 - [x] DataFlagEncoder requires e.g. road_environment - throw an exception if not added before. Breaks backward compatibility.
 - [x] remove all TODO NOW in src and test
 - [x] ~~should we introduce special rounding behaviour? E.g. for max_weight or max_height it can happen that the value increases due to rounding which might be a problem.~~ For now do not do this
 - [x] generic MappedStringEncodedValue -> different issue
 - [x] a real ObjectEncodedValue that uses e.g. MapDB under the hood and just stores the integer ID in the edge -> different issue
 - [x] move back to EnumEncodedValue and remove the IndexedBase stuff, see #1570, see [this branch and the resulting changes](https://github.com/graphhopper/graphhopper/compare/shared_enc2...shared_enc2_enum?expand=1)
 - [x] ~rename max_speed to speed_limit?~ no. let's just use max and average speed...
 - [x] default value is 0, but use a different value e.g. for max_speed (infinity)
 - [x] support "unlimited" on German autobahns: maybe we just use the maximum value. E.g. 5 bits and factor==5 => 5*(2⁵-1)=155. But then custom implementations couldn't easily change the precision of maxspeed as this changes the maximum value too. Maybe let's define 30*5=150 (can be precisely returned if precision is 5, 3, 2, 1). As the 140 limit exists several times and 160 is very rare (basically [2 ways world wide](http://overpass-turbo.eu/s/Hfw)) and everything above is a tagging mistake or something like [this](https://www.openstreetmap.org/way/38108321), [2](https://www.openstreetmap.org/way/40816550), [3](https://www.openstreetmap.org/way/227906554), [4](https://www.openstreetmap.org/way/39959598). [Valhalla uses 255 (8 bits)](https://github.com/valhalla/valhalla/blob/534b400a4cba4b1b4b643b69e2e3a549075dccfd/src/mjolnir/osmway.cc#L12). If we use 150 max speed, then the max average speed is `0.9*150=135`
 - [x] `other` should not be the first (default) access
 - [x] rename file back to map.geo.json
 - [x] improve error message of MappedDecimalEncodedValue; for now remove unused MappedDecimalEncodedValue?
 - [x] should int details also get returnMinus like decimals? removed this param for now.
 - [x] try returning `destination` access for tracks
 - [x] fix toll order handling (see comment)
 - [x] there should be an exception if only one direction is supported but setReverse is used. Also make the twoDirections variable accessible from outside.
 - [x] for JSON return -1 instead of the default "infinity" string value
 - [x] Forestry is not a highway (we would tag it via track and maybe add access=forestry) but add briddleway.

later:

 - For the JSON API we should omit default entries if they are infinity and not return -1
 - important follow up: #1269 and #1551


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -841,22 +869,14 @@ public void postProcessing() {
...
-        if (ghStorage.getEncodingManager().hasEncoder("generic")) {
-            final FlagEncoder genericFlagEncoder = ghStorage.getEncodingManager()
-                    .getEncoder("generic");
-            if (!(genericFlagEncoder instanceof DataFlagEncoder)) {
-                throw new IllegalStateException("'generic' flag encoder for elevation interpolation of "
-                        + "bridges and tunnels is enabled but does not have the expected type "
-                        + DataFlagEncoder.class.getName() + ".");
-            }
-            final DataFlagEncoder dataFlagEncoder = (DataFlagEncoder) genericFlagEncoder;
+        if (ghStorage.getEncodingManager().hasEncodedValue(RoadEnvironment.KEY)) {
```
boldtrn: This became really clean and easy to understand :+1:

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
```diff
@@ -43,17 +43,19 @@ public MappedDecimalEncodedValue(String name, Collection<Double> values, double
...
 
+        List<Double> sortedValues = new ArrayList<>(values);
+        Collections.sort(sortedValues);
         int index = 0;
         IntHashSet dupCheck = new IntHashSet();
-        for (double val : values) {
+        for (double val : sortedValues) {
             int intVal = toInt(val);
             if (!dupCheck.add(intVal)) {
                 throw new IllegalArgumentException("The value " + val + " was converted to " + intVal + " but this already exists. Remove it to improve efficiency.");
```
boldtrn: Can we improve this error message? Right now this sounds more like an INFO log statement. We could also mention that one could also increase the precision to fix the error.

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
```diff
@@ -63,18 +65,53 @@ private int toInt(double val) {
         return (int) Math.round(val / precision);
     }
 
+    int findClosestIndex(int val) {
```
boldtrn: Should this method be private? The exception below sounds like that as well.

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
```diff
@@ -63,18 +65,53 @@ private int toInt(double val) {
...
+        int res = Arrays.binarySearch(toValueMapArray, val);
+        if (res >= 0)
+            throw new IllegalStateException("Cannot happen as toStorageMap.getOrDefault should have caught this case");
+
+        // TODO NOW rounding up from e.g. max_height=10 to 11 might be ugly. Not sure how or what to do about it or if we should make it even configurable.
+        res = -res;
+        double delta1 = res < 2 ? Double.POSITIVE_INFINITY : val - toValueMapArray[res - 2],
+                delta2 = res >= toValueMapArray.length ? Double.POSITIVE_INFINITY : toValueMapArray[res - 1] - val;
```
boldtrn: It might be worth to add a comment here explaining what is done. The two inline ifs are a bit hard to read :).

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
```diff
@@ -63,18 +65,53 @@ private int toInt(double val) {
         return (int) Math.round(val / precision);
     }
 
+    int findClosestIndex(int val) {
+        int res = Arrays.binarySearch(toValueMapArray, val);
+        if (res >= 0)
+            throw new IllegalStateException("Cannot happen as toStorageMap.getOrDefault should have caught this case");
+
+        // TODO NOW rounding up from e.g. max_height=10 to 11 might be ugly. Not sure how or what to do about it or if we should make it even configurable.
```
boldtrn: Maybe adding a precise mode would make sense. Every value that cannot be exactly matched will throw an error or return a fail flag in `setDecimal`. 

I actually assumed this class would work like that. If you map Strings, you would probably fail and not search for the closest String :).

core/src/main/java/com/graphhopper/routing/profiles/MaxSpeed.java
```diff
@@ -0,0 +1,40 @@
...
+     * speed value used for "none" speed limit on German Autobahn
+     */
+    public static final double UNLIMITED_SIGN_SPEED = 140;
+
+    /**
+     * speed value used for road sections without known speed limit.
+     */
+    public static final double UNSET_SPEED = 0;
+
+    public static DecimalEncodedValue create() {
```
boldtrn: Would it make sense to have a common interface for these profiles?

core/src/main/java/com/graphhopper/routing/profiles/RoadAccess.java
```diff
@@ -0,0 +1,51 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.routing.profiles;
+
+import com.graphhopper.util.Helper;
+
+public enum RoadAccess {
+    // order is important here as we assume "smaller index" means "broader access"
+    OTHER("other"), YES("yes"),
```
boldtrn: Should other really be the first/broadest access? 

There are access values like [unknown, military, residents, and many undocumented values](https://taginfo.openstreetmap.org/keys/access#values).

core/src/main/java/com/graphhopper/routing/util/parsers/OSMCarMaxSpeedParser.java
```diff
@@ -0,0 +1,86 @@
...
+        if (bwdSpeed > maxPossibleSpeed)
+            bwdSpeed = maxPossibleSpeed;
+
+        if (fwdSpeed <= 0)
+            fwdSpeed = UNSET_SPEED;
+        carMaxSpeedEnc.setDecimal(false, edgeFlags, fwdSpeed);
+
+        if (bwdSpeed <= 0)
+            bwdSpeed = UNSET_SPEED;
+        carMaxSpeedEnc.setDecimal(true, edgeFlags, bwdSpeed);
```
boldtrn: This assumes, that car max speed is stored in 2d. If one might change this in `MaxSpeed`, this might go unnoticed and the speed in the edge will be the backward speed?

This is a bit of an edge case, but I think it would be nice if there would be either a exception or a fallback that uses `Math.min(forward,backward)`. Maybe there is something like this in place already and I missed it :).

core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxHeightParser.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public void createEncodedValues(EncodedValueLookup lookup, List<EncodedValue> registerNewEncodedValue) {
+        registerNewEncodedValue.add(heightEncoder);
+    }
+
+    @Override
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, EncodingManager.Access access, long relationFlags) {
+        List<String> heightTags = Arrays.asList("maxheight", "maxheight:physical"/*, do not use "height" for way height! */);
```
boldtrn: > do not use "height" for way height!

What do you mean by that?

core/src/main/java/com/graphhopper/routing/util/parsers/OSMTollParser.java
```diff
@@ -0,0 +1,51 @@
...
+    public void createEncodedValues(EncodedValueLookup lookup, List<EncodedValue> list) {
+        list.add(tollEnc);
+    }
+
+    @Override
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, EncodingManager.Access access, long relationFlags) {
+        if (readerWay.hasTag("toll", "yes"))
+            tollEnc.setEnum(false, edgeFlags, Toll.ALL);
+        if (readerWay.hasTag("toll:hgv", "yes"))
+            tollEnc.setEnum(false, edgeFlags, Toll.HGV);
```
boldtrn: For an edge with:
```
toll=yes
toll:hgv=yes
```

This would result in `Toll.HGV` and not `Toll.ALL`

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GermanySpatialRule.java
```diff
@@ -49,10 +50,10 @@ public double getMaxSpeed(String highwayTag, double _default) {
     }
 
     @Override
-    public Access getAccess(String highwayTag, TransportationMode transportationMode, Access _default) {
+    public RoadAccess getAccess(String highwayTag, TransportationMode transportationMode, RoadAccess _default) {
         if (transportationMode == TransportationMode.MOTOR_VEHICLE) {
             if (highwayTag.equals("track"))
-                return Access.CONDITIONAL;
+                return RoadAccess.FORESTRY;
```
boldtrn: Isn't this often Agricultural?

[This](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restrictions#Germany) page proposes destination.

core/src/main/java/com/graphhopper/routing/weighting/AvoidWeighting.java
```diff
@@ -0,0 +1,114 @@
...
+            avoidPrimary = avoidStr.contains("primary");
+            avoidSecondary = avoidStr.contains("secondary");
+            avoidTertiary = avoidStr.contains("tertiary");
+            avoidTrack = avoidStr.contains("track");
+            avoidResidential = avoidStr.contains("residential");
+            roadClassEnc = encoder.getEnumEncodedValue(RoadClass.KEY, RoadClass.class);
+        }
+
+        // can be used for preferring too
+        avoidFactor = Math.min(10, Math.max(0.1, map.getDouble("avoid.factor", 10)));
```
boldtrn: Is/Should there be a safety check when using LM that the avoid.factor is at least 1?

core/src/main/java/com/graphhopper/util/details/IntDetails.java
```diff
@@ -15,27 +15,33 @@
...
-public class TunnelElevationInterpolator extends AbstractEdgeElevationInterpolator {
+public class IntDetails extends AbstractPathDetailsBuilder {
+
+    private final IntEncodedValue ev;
+    private int intVal = -1;
 
-    public TunnelElevationInterpolator(GraphHopperStorage storage,
-                                       DataFlagEncoder dataFlagEncoder) {
-        super(storage, dataFlagEncoder);
+    public IntDetails(String name, IntEncodedValue ev) {
```
boldtrn: Why is there no return minus option here?

web/src/test/java/com/graphhopper/http/GraphHopperLandmarksTest.java
```diff
@@ -32,7 +32,7 @@
 import static org.junit.Assert.assertTrue;
 
 /**
- * Tests the creation of Landmarks and parsing the map.geo.json file
```
boldtrn: The file is still called map.geo.json :)

boldtrn:
This PR looks really great. I had some minor comments (see above). I will do some more tests and see how it works in action :).

core/src/main/java/com/graphhopper/util/details/IntDetails.java
karussell: Do you mean why it is not possible to store negative values?

core/src/main/java/com/graphhopper/routing/weighting/AvoidWeighting.java
karussell: Maybe we move the AvoidWeighting stuff outside of this already big PR as this already introduces a feature that is maybe not wanted.

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GermanySpatialRule.java
karussell: I think I was trying to satisfy some unit tests but will try `destination`

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
karussell: I'm unsure if we need two modes. Actually the naming "MappedDecimal" might be wrong as it is more a compressed "fequency based" storage format, but not losless.

Do you know a use case where you would need the precise mode?

core/src/main/java/com/graphhopper/routing/profiles/MaxSpeed.java
karussell: As it is static - how could this work?

core/src/main/java/com/graphhopper/routing/profiles/RoadAccess.java
karussell: Yes, indeed.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMCarMaxSpeedParser.java
karussell: Yes, there should be probably an exception if not two directions.

karussell:
Thanks for your review. I think there are still a few conceptual problems like the default value for DecimalEV.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxHeightParser.java
karussell: the height in OSM != height of a street. Unlike width.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMCarMaxSpeedParser.java
karussell: This is tricky. Or at least we would have to make the twoDirections variable accessible from outside so that this parser can handle the two cases.

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
boldtrn: Yes probably renaming MappedDecimal would make this clearer.

Right now I don't have a good example where the precise mode would be necessary. One could always extend the list of stored values, if 1 or 2 values should be available. 

I think my issues was that I expected it to work like a mapped String.

core/src/main/java/com/graphhopper/routing/profiles/MaxSpeed.java
boldtrn: Ah you are right, sorry :(. I overlooked the static.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMCarMaxSpeedParser.java
boldtrn: We could also throw and exception in the encoded value. Something like ``!storeBothDirections && reverse`` in `SimpleIntEncodedValue`. But I am not sure if this would break something elsewhere?

core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxHeightParser.java
boldtrn: Ah, thanks for the explanation. I wasn't sure if this would be related to elevation. Could we maybe change the comment slightly, like:

` the OSM tag "height" is not used for maxheight` 

or something like this?

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GermanySpatialRule.java
boldtrn: I think it's worth discussing about this :). Essentially for most cases Agricultural or Forestry will results in no access. Whereas destination would result in access, but with a high weight or an actual destination check. I am not sure which is the better solution.

core/src/main/java/com/graphhopper/routing/weighting/AvoidWeighting.java
boldtrn: Yes, I think this is a good idea :)

core/src/main/java/com/graphhopper/util/details/IntDetails.java
boldtrn: The DecimalDetails have the option `returnMinus`. The IntDetails don't have that option. I was wondering why.

boldtrn:
> I think there are still a few conceptual problems like the default value for DecimalEV

Yes indeed. I didn't mention it, because it is still an open todo.

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
boldtrn: I just had an idea where a precise mode could make sense. Maybe this could allow us to map `Double.INFINITY` and `Double.NaN`, with NaN referring to the default value and INFINITY referring to unlimited speed :).

core/src/main/java/com/graphhopper/util/details/PathDetailsBuilderFactory.java
```diff
@@ -50,6 +54,20 @@
         if (requestedPathDetails.contains(DISTANCE))
             builders.add(new DistanceDetails());
 
+        for (String key : Arrays.asList(MaxSpeed.KEY, MaxWidth.KEY, MaxHeight.KEY, MaxWeight.KEY)) {
+            if (requestedPathDetails.contains(key) && encoder.hasEncodedValue(key))
+                builders.add(new DecimalDetails(key, encoder.getDecimalEncodedValue(key), true));
+        }
+
+        for (Map.Entry entry : Arrays.asList(new MapEntry<>(RoadClass.KEY, RoadClass.class),
```
boldtrn: Could we extract these arrays to class level?

karussell: What do you mean here?

core/src/main/java/com/graphhopper/routing/profiles/RoadClass.java
```diff
@@ -1,55 +1,55 @@
...
-            "service", "road", "track", "forestry", "steps", "cycleway", "path", "living_street");
+/**
+ * This class defines a IndexBased but is type safe.
+ */
+public enum RoadClass {
+    OTHER("other"), MOTORWAY("motorway"), MOTORROAD("motorroad"),
+    TRUNK("trunk"), PRIMARY("primary"), SECONDARY("secondary"),
+    TERTIARY("tertiary"), RESIDENTIAL("residential"), UNCLASSIFIED("unclassified"),
+    SERVICE("service"), ROAD("road"), TRACK("track"),
+    FORESTRY("forestry"), STEPS("steps"), CYCLEWAY("cycleway"),
```
boldtrn: Why do we need FORESTRY here in RoadClass?

BTW: Would it make sense to add briddleway here?

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
karussell: Yeah, exactly that problem appears when we want to introduce a default value (INFINITY). But my problem is that in reality you want it both: 
 * if the stored value is 0 (the default) then return INFINITY. 
 * For everything else find the closest entry. 

Having a simple MappedDecimal class would be much simpler than it is now, but then the "compactness" advantages over the FactoredDecimalEncodedValue would disappear IMO, especially for large areas with many distinct value.

core/src/main/java/com/graphhopper/routing/profiles/RoadClass.java
karussell: You are right. Forestry is not a highway (we would use track with access=forestry). And surely we can add briddleway.

core/src/main/java/com/graphhopper/routing/profiles/MappedDecimalEncodedValue.java
karussell: > Double.INFINITY and Double.NaN

I would use INFINITY as default value, i.e. map the 0-index to infinity (all vehicle allowed by default)

Another idea could be to collect the values "on-the-go" while import and throw an exception if the bits (array size) are too small. And prefill this array with one value: infinity.

But maybe the best solution for now would be to remove this class and use a FactoredDecimalEncodedValue? The current "mapped"-approach uses 5 bits. A "factor"-approach with factor=0.1 and max=25 means 8 bits. Do you think this is acceptable for now?

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GermanySpatialRule.java
karussell: Ok, for Germany I'll do destination only. But for Austria we should keep the access to forestry (general access is `no`). Agriculture shouldn't be used in those cases. At least they say this:

> Note a farmer's access track would be private rather than agricultural unless the track is open to any vehicle used for agricultural purposes. 

https://wiki.openstreetmap.org/wiki/Key:access

karussell: Please see this commit: https://github.com/graphhopper/graphhopper/commit/4896d85452d68a686da263b08fac9f7010f733da

core/src/test/java/com/graphhopper/routing/ch/PrepareContractionHierarchiesTest.java
```diff
@@ -473,7 +473,7 @@ public void testStallOnDemandViaVirtuaNode_issue1574() {
         // would not happen, because node 2 would not even be explored in the forward search, but because of the virtual
         // node the strict upward search is modified and goes like 0-3-1-2 (i.e. it finds node 2).
         // so no we fine tune the weight for 2-4 such that node 4 gets also stalled
-        edge24.setReverse(carEncoder.getAverageSpeedEnc(), 27.5);
+        edge24.set(carEncoder.getAverageSpeedEnc(), 27.5);
```
karussell: @easbar this was necessary to make the test pass. 

When I allow both directions via:

```
new CarFlagEncoder(new PMap("speed_two_directions=true"))
```

then I get 4 shortcuts (not 2) and a different path:

```
java.lang.AssertionError: wrong or no path found
Expected :[0, 3, 8, 1, 2, 4, 5, 6, 7]
Actual   :[0, 3,    1, 2, 4, 5, 6, 7]
```

core/src/main/java/com/graphhopper/routing/profiles/SimpleIntEncodedValue.java
```diff
@@ -107,7 +107,10 @@ public final void setInt(boolean reverse, IntsRef ref, int value) {
     }
 
     final void uncheckedSet(boolean reverse, IntsRef ref, int value) {
-        if (storeBothDirections && reverse) {
+        if (reverse && !storeBothDirections)
+            throw new IllegalArgumentException(getName() + ": value for reverse direction would overwrite forward direction. Enable 'storeBothDirections' or always use reverse==true when storing the value");
```
easbar: Nice! Will this fix #1588 ?

karussell: I think so. As it avoids the overwriting of the reverse property (setReverse throws an exception now). 

Do you think we should do the same for getReverse without bi-direction support? Currently I don't do this for convenience reasons (no need to do `if(encodedValue.isStoreBothDirections)` everytime you access a property and you want bi-directional support e.g. in a Weighting).

core/src/test/java/com/graphhopper/routing/ch/PrepareContractionHierarchiesTest.java
easbar: Ok this is strange. setting `speed_two_directions` only allows setting two different speeds right ? all the edges are bidirectional anyway. And what I really do not understand is that the calculated path seems to skip the (virtual) node 8 (probably because some shortcut is used, but this should not be). I will take a closer look to see what is going on here.

core/src/test/java/com/graphhopper/routing/ch/CHTurnCostTest.java
```diff
@@ -747,64 +747,6 @@ public void testFindPath_loopEdge() {
         compareCHQueryWithDijkstra(pch, 0, 4);
     }
 
-    @Test
```
easbar: Can you move this back below issue1593_simple ? looks like nothing really changed here ?

core/src/main/java/com/graphhopper/routing/profiles/SimpleIntEncodedValue.java
easbar: > Do you think we should do the same for getReverse

I would say no. If you call `getReverse` even though its not bidirectional that's just fine you get the same value as if you had called `get`. I do not think this would be something unexpected. You do not know the value beforehand and calling `get` or `getReverse` gives you some value you need to trust. Maybe both directions are the same or the property is not bi-directional - this should not be of your concern when just calling the getter.

In comparison, calling `setReverse`, really gives you the impression that you can set a special reverse value, but if its not bidirectional it will be lost or even overwrite the other direction, which is a nasty surprise, so here I think it is good that it throws now.

core/src/test/java/com/graphhopper/routing/ch/CHTurnCostTest.java
karussell: Ah, sure. Was just an ugly merge conflict where I had many duplicated methods and I obviously picked the wrong that I removed ;)

core/src/test/java/com/graphhopper/routing/ch/PrepareContractionHierarchiesTest.java
easbar: Ok I took a look at this test and fixed some inaccurate comments, see here: ec01e5e458bce46ffddea8a7b266cacbc49e6a1e.
I also explicitly set the speed for *both* directions in this test now, see here: 
c979cbd4511a3263bc6ec312c4a55ba6a1ec1c20

> then I get 4 shortcuts (not 2) 

Ok this is clear, because here we are only counting the number of shortcut edges (independent of their access flags). so far `set()` set the speed for both directions, but with the new code only one of the directions is set, right ? if the edge speeds are different for the two directions the previous two 'bidirectional' shortcuts will turn into four unidirectional ones.

If I comment out the assertEquals(there are two shortcuts), the next thing that fails is this:
`sptWeight at node 4 should be smaller than shortcut weight 3->4 to make sure node 4 gets stalled`
This is because you changed `setReverse` to `set` and the `getWeight` method read the wrong weight (it encounters the 'wrong' edge first and returns it), I fixed this here: a5297537fb5bd15bd9a7c512bc494d503ab9a670

>  And what I really do not understand is that the calculated path seems to skip the (virtual) node 8 

This is because when we use `setReverse` instead of `set` the edge weights are not such that the stalling happens. When the best path is found via the 3->4 shortcut the virtual node will not be included in the final path (because when the shortcut is unpacked it does not know about `QueryGraph`s virtual node). 

`setReverse` was wrong in the test so far, but so far this was no problem, because it did the same as `set` (it set *both* directions).

The best way to check everything is fine is reverting the #1574 fix (set precision back to 0 in the stalling condition in `DijkstraBidirectionCH`) and make sure no path is found. Also run `RandomCHRouting#random` many times as described in it's comment, but be aware that it currently does not use `speed_both_directions=true`.

easbar: If you update to master no changes should be required for the test to work and setting `speed_both_directions=true` should be fine.

karussell:
Currently testing this in production and the minor problem is motorroad=yes. Currently we have this in RoadClass.java but the following is missing in the OSMRoadClassParser:

```
  if (!Helper.isEmpty(highwayValue) && way.hasTag("motorroad", "yes")
                && highwayValue != "motorway" && highwayValue != "motorway_link")
    highwayValue = "motorroad";
```

(this is also done in master CarFlagEncoder.getSpeed)

Or should we better establish a separate "motorroad" encoded value to be more align with OSM tags? Will have a look how this is done for other road data.

boldtrn:
I personally would prefer to use trunk or motorway instead of motorroad. In OSM the highway is set to trunk or motorway, so returning motorroad seems a bit wrong IMHO. Maybe it would make sense to store motorrad as a Boolean?

karussell:
Yeah, I feel the same. I have removed motorroad from the RoadClass enum. And we can later support motorroad maybe with some country dependent access tags. The OSM wiki says that there are only some access differences to trunk.

I think this PR is now in a good shape and will merge it in the next days if no objections.

PR #1555, opened 2019-02-26
Improves error message in case of forgotten/wrong edge_based=true/false parameter.

easbar:
Its easy to forget setting `edge_based=true` at query time when only edge based CH is prepared.


karussell:
Thanks!

PR #1563, opened 2019-03-04
Allow to retrieve SpatialRule by Id

boldtrn:
With this PR it's possible to retrieve a SpatialRule from the Lookup using the id. This can be especially helpful when one wants to store the id in the graph and retrieve the SpatialRule from that id.

One thing to consider is that if you change the SpatialRules (add one or change the the order), this will fail. Currently, we don't save the version of the SpatialRuleLookup.


karussell:
Thanks!

Issue #1177, opened 2017-10-05
GH Maps - Make PathDetails visible


boldtrn:
Currently GH Maps uses [this elevation plugin](https://github.com/MrMufflon/Leaflet.Elevation), which seems to be not maintained anymore (last change was in 2016, 8 open PRs). 

I have just seen this [repo](https://github.com/GIScience/Leaflet.Heightgraph). Essentially, they do more or less the same as the current elevation plugin, but they also allow to visualize additional information in the height graph, like surface, as shown in the [example](https://giscience.github.io/Leaflet.Heightgraph/).

karussell:
BTW: why has this issue this name :) ?

boldtrn:
> BTW: why has this issue this name :) ?

Good question :). Because it could make the PathDetails easily visible on the frontend without requiring debug=true. For me using the updated elevation plugin has two benefits: a) switching to an actively developed elevation plugin and b) it is more powerful and could allow to show the PathDetails more easily.

Should I rename it? I only addressed one of the two issues in the title.

karussell:
Ah, you mean you would show the path details in the elevation view - interesting! How would you show multiple details like street name and max speed?

boldtrn:
> How would you show multiple details like street name and max speed?

Have a look at their [example](https://giscience.github.io/Leaflet.Heightgraph/), they included a way to switch between multiple types in the bottom-right.

karussell:
Ah, didn't see this in the first place. This is really cool work from @TimMcCauley and @Kinas and better than our current solution with hidden parameters and blue dots!

PR #1569, opened 2019-03-08
Use Leaflet.Heightgraph to show PathDetails

boldtrn:
This PR changes the elevation diagram library as proposed in #1177. If we want to merge this PR, I would like to first publish the library Leaflet.Heightgraph to NPM, this will make the PR cleaner (also see [this](https://github.com/GIScience/Leaflet.Heightgraph/issues/36) issue).

One issue with Leaflet.Heightgraph is that it does not support a miles display yet, so the units are fixed to metric units. We could add this feature to Leaflet.Heightgraph though. I haven't studied the code of Leaflet.Heightgraph in detail yet, so I don't know how much work this would require.

The results are already pretty good, so I think this would be a nice addition :). Also the diagram itself has some nice little features included, for example you can toggle it's state from shown to hidden or highlight parts of the route that are above a certain elevation.

If you don't use any PathDetails, the resulting diagram will be a plain elevation diagram, similar to the old elevation diagram. If you add path details to the request (currently you need to add them to the url, for example add `&details=average_speed&details=street_name`), you can see the results of the path details in the diagram.

screenshot from 2019-03-08 15-07-20


karussell:
Really cool, thanks a lot!

karussell:
Will try (and merge) on Monday latest.

karussell:
Just a quick question: the elevation data is always shown implicitly, right? Or could we use the height of the diagram for integer based path details like average_speed?

boldtrn:
> Or could we use the height of the diagram for integer based path details like average_speed?

This should be possible as well for number based path details like speed. We can also define the color mapping, so that higher speeds could be green and lower speeds red or anything like this. Should I add something like this?

boldtrn:
It's possible :)

screenshot from 2019-03-08 19-15-09

karussell:
This is really great! 

Potentially this is also useful for max_speed, but maybe there are problems for the unlimited cases on a German Autobahn leading to infinity peaks ;)

boldtrn:
> Potentially this is also useful for max_speed, but maybe there are problems for the unlimited cases on a German Autobahn leading to infinity peaks ;)

We can catch these ;). I just pushed the code. The current solution will require manual handling for every path detail that we want to show like this instead of the elevation.

karussell:
Cool - thanks!

Could we automatically enable it for all path details with numbers?

karussell:
Ok, after "npm update" this works for me!

And yes, those new features are really neat - wow! So this is an important thing to have.

There is some limitation when I use `details=average_speed`, then the elevation is no longer an option, I would have expected that I can see the `elevation` and `average_speed` values. It seems the values are globally changes which will be problematic when we want the height of the chart representing the values of e.g. details=max_speed and details=average_speed. Do you know what I mean?

A tiny thing: without a details parameter the elevation is shown - here we should use the old green color like before :)

Then we need to test if it is possible to show details even if the graph has no elevation data imported.

> so the units are fixed to metric units. 

This is not important to fix.

boldtrn:
> It seems the values are globally changes which will be problematic when we want the height of the chart representing the values of e.g. details=max_speed and details=average_speed. Do you know what I mean?

Thanks for finding this! I just double checked, you are right, this is rather ugly :). I was able to fix this part, but then the elevation plugin didn't work properly anymore. I reverted my latest change for now, it's not as easy as I hoped to use the height of the bars to indicate the average speed. I think this would need further changes in the plugin itself.

One issues was that with average_speed, the bars will be between 0-140. For elevation the bars might be between 300-500. The plugin did not handle this properly, so we'd need an update mechanism when switching between the different path details to also update the "height" of the graph.

> Then we need to test if it is possible to show details even if the graph has no elevation data imported.

This should be possible, once we improve the plugin :). For only a single PathDetail this should be possible already.

> It is already on graphhopper.com/maps :)

This is awesome :)

karussell:
I would also revert change regarding the "speed as height" in the chart. In a separate PR we can make the necessary changes in this matter - WDYT?

boldtrn:
> I would also revert change regarding the "speed as height" in the chart. In a separate PR we can make the necessary changes in this matter - WDYT?

Yes :).

karussell:
Can you check if my latest change regarding the color is fine?

web/src/main/resources/assets/js/map.js
```diff
@@ -285,23 +285,24 @@ module.exports.addElevation = function (geoJsonFeature, useMiles, details) {
...
+           margins: {
+               top: 10,
+               right: 30,
+               bottom: 55,
+               left: 50
+           },
+           position: "bottomright"
+        }
+        // for path details use random colors
+        if(jQuery.isEmptyObject(details))
```
karussell: Here jquery is required as details is `{}` if no details parameter was specified

web/src/main/resources/assets/js/map.js
```diff
@@ -285,23 +285,24 @@ module.exports.addElevation = function (geoJsonFeature, useMiles, details) {
...
+           options.mappings = { Elevation: {'elevation': {text: 'Elevation [m]', color: '#27ce49'}}};
```
karussell: This was a bit tricky: the first key `Elevation` must match the summary and the second `elevation` the attributeType

web/src/main/resources/assets/js/map.js
boldtrn: We could add this part [here](https://github.com/graphhopper/graphhopper/pull/1569/files#diff-e1a9691b5e85c253cc337b86010c30abR274) and avoid using jQuery.

web/src/main/resources/assets/js/map.js
boldtrn: Yes that's indeed tricky :). We could think about color mappings for the other details as well.

boldtrn:
This looks good thanks :). I slightly rectored the code, so we don't need jQuery. WDYT?

karussell:
>I slightly rectored the code, so we don't need jQuery. WDYT?
Sure - perfect - thanks!

karussell:
I had to make this PR ready for review ... is this correct that it is no longer a WIP?

boldtrn:
> ... is this correct that it is no longer a WIP?

We could merge it.

I wanted to first publish Leaflet.Heightgraph on NPM and use the dependency from NPM instead of hard copying the js file. But we can do this later as well. Also we could delete all old references from the old elevation library (like css).

karussell:
Ah, ok. This would be good to be done before the merge, but of course no hurry :) (can also remove the 0.12 tag)

karussell:
Thx!

web/package.json
```diff
@@ -33,7 +33,8 @@
     "leaflet-loading": "0.1.24",
     "moment": "2.22.1",
     "uglifyify": "5.0.0",
-    "flatpickr": "4.4.6"
+    "flatpickr": "4.4.6",
+    "leaflet.heightgraph": "0.0.1"
```
karussell: Oh, interesting. So it is possible to release stuff on npm without any namespace prefix? How does one know that it is original content?

boldtrn: > So it is possible to release stuff on npm without any namespace prefix?

Yes, in the beginning npm had no namespace prefix. There is now the option to register for a namespace/organisation. The dependency would then look like `@npmcorp/package-name`.

BTW: we could register an organisation for GH [here](https://www.npmjs.com/org/create) :)

> How does one know that it is original content?

Once you publish a package with a name, only you (or selected contributors) can upload another package with the same name. Usually you could first check the Github repo and check the package.json if the names match.

Issue #1558, opened 2019-02-28
Upgrade dropwizard to 2.0.0 when it is out to fix Exception in logs


karussell:
See [discussion](https://discuss.graphhopper.com/t/cant-load-pre-imported-graph-on-different-machine/4109/24)

Stacktrace:
```
org.glassfish.jersey.server.ServerRuntime$Responder: An I/O error has occurred while writing a response message entity to the container output stream.
! java.io.IOException: Broken pipe
! at sun.nio.ch.FileDispatcherImpl.write0(Native Method)
! at sun.nio.ch.SocketDispatcher.write(SocketDispatcher.java:47)
! at sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:93)
! at sun.nio.ch.IOUtil.write(IOUtil.java:65)
! at sun.nio.ch.SocketChannelImpl.write(SocketChannelImpl.java:471)
! at org.eclipse.jetty.io.ChannelEndPoint.flush(ChannelEndPoint.java:264)
! ... 11 common frames omitted
! Causing: org.eclipse.jetty.io.EofException: null
! at org.eclipse.jetty.io.ChannelEndPoint.flush(ChannelEndPoint.java:286)
! at org.eclipse.jetty.io.WriteFlusher.flush(WriteFlusher.java:429)
! at org.eclipse.jetty.io.WriteFlusher.completeWrite(WriteFlusher.java:384)
! at org.eclipse.jetty.io.ChannelEndPoint$3.run(ChannelEndPoint.java:133)
! at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.runTask(EatWhatYouKill.java:333)
! at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.doProduce(EatWhatYouKill.java:295)
! ... 6 common frames omitted
! Causing: org.glassfish.jersey.server.internal.process.MappableException: org.eclipse.jetty.io.EofException
...........................
```

Reason: client disconnects too early

[Fix available in dw 2.0.0](https://github.com/dropwizard/dropwizard/pull/2541)

karussell:
Already dropwizard 2.0.8 is out now, so we could consider switching to it: 
https://github.com/graphhopper/graphhopper/tree/dw20

@michaz any problems with this version?

michaz:
@karussell None that I was aware of. I just dropped it in and it worked.

karussell:
Ok, great! Will merge it then :)

PR #1572, opened 2019-03-19
Render vector tiles from graph storage

### LINKED ISSUE #1456: Make graph visible in web UI
I think for debugging and analysis of our routing algorithms it would be incredibly useful to make the graph as visible as possible. I think for some use-cases drawing the graph directly on top of the OSM layer also has several advantages over the MiniGraph/java-based/Swing approach. I have started to build a little prototype along this line and at the moment it looks like this:

image

I can draw a bounding box on the map layer and then query the server for the edges of the graph and draw them on the map. I have a few more ideas that would be useful like drawing CH shortcuts, filtering edges by node level (to see the CH contraction evolve), visualize fwd/bwd searches etc.

So for development and debugging I think this will be very useful, but at the same time this should probably not be available on graphhopper.com/maps, so there are a few options:

1.  keep this code in a separate branch (I would try to put as much code as possible in additional files so that its easy to merge master at any time)

2.  add this to master and use some isProduction() flag to disable the additional code (same goes for the REST endpoint I am using on the server)

3.  do this in a separate project (maybe without routing) (probably not a very good idea, because we need to change code on the server as well)

Does anybody else think this would be useful ? Should this be part of this repository in any way ? I guess as long as I am the only one using this I will go with the first option. It will probably take some time to figure out what is the best way to make use of this feature and in many cases there will probably be some quick experimental code to debug a specific feature.

karussell:
Fixes #1456 

We can now render the roads in a browser for debugging purposes with an acceptable performance - will be especially interesting with #1548:

image

After several trials I think I got something I find interesting:

 0. this PR is based on #1485 as it fetches all nodes&edges for a certain bounding box. Everything worked really well, did not yet stumbled over a bug although I should have found one due to the minor limitation listed there.
 1. this PR introduces a MVTResource that returns a "roads" layer as vector tiles in a binary and compact format. It uses https://github.com/wdtinc/mapbox-vector-tile-java which is nearly 10x faster compared to the [other MVT library](https://github.com/ElectronicChartCentre/java-vector-tile) but still it is the slowest part in this visualization process. It converts real world coordinates into a "local" vector tiles coordinate system using JTS and something is a bit slow. Also the library is a bit low level and unusual to use (was a bit tricky to get it done), but now we can send also edge properties to the browser and customize this with just JavaScript.
 2. These vector tiles can be consumed with any client library supporting the mapbox vector tiles spec (MVT) and I found a way to consume this even in leaflet using Leaflet.VectorGrid https://gist.github.com/karussell/0720b3d3b297621c8bb10cf8c86b2906 and the best thing is: it is not so slow like all the things I tried before (highly likely due to the canvas renderer which was pointed out from @michaz already before :))

TODO: 

 - [x] make it working as base layer for e.g. Mapbox GL JS (then it should work for deck.gl too)
 - [x] remove a few things from the isochrone module (I started my work there)
 - [x] make it working in deck.gl either with the base vector tile map or using the geojson-layer
 - [x] ~~allow to export GeoJSON alternatively (rename to /graph-explore endpoint or something)~~ no. Use GeoJSON only for responses that are not easy to serve as tiles like shortest path trees. See also #1577 
 - [x] integrate this into our JS demo using Leaflet.VectorGrid
 - [x] write a test via decoding the returned MVT stuff
 - [x] ~With maputnik this gives me Error: http status 200 returned without content. Can we do better at returning 'an empty tile' ?~ fix later


easbar:
Sounds very interesting, will have a closer look! Is this just about visualizing the route or the whole / parts of the graph ? For the purposes I was looking for in #1456 this here is my (probably still incomplete) wish list. Do you think we can do all (or most of) this using the mvt approach (<- sorry no idea atm what the possiblities/limitations are) ?

- draw bounding box and show graph (including node/edge infos like id, weight, distance, ch level etc., possibly by clicking or hovering as a popup as otherwise it becomes hard to see anything)
- show turn restrictions
- zoom/fly to coordinate
- zoom to a given node/edge id and/or draw bounding box around a certain node/edge id
- somehow visualize loops and duplicate edges (this I have found is not so easy would have to split/slightly offset edges that lie on top of each other, or use some 3D features)
- show more/less details depending on zoom level
- show/hide details via UI (enable/disable checkbox, move slider etc.)

karussell:
> Is this just about visualizing the route or the whole / parts of the graph ? 

Whole graph

> draw bounding box and show graph

no bounding box necessary as everything that is visible is shown (for every "tile" it does a request and the server response gives MVT data)

> and there is possibility for a popup showing the name of the road :)

You can actually show every data you like be it as a color or if you click on the road via a popup, yes :)

> show turn restrictions

:+1: ... this get tricky though :)

> show more/less details depending on zoom level

This can be easily done. And is done already. E.g. for the overview it just returns junction nodes and deeper zooms gives full way geometry

> somehow visualize loops and duplicate edges (this I have found is not so easy would have to split/slightly offset edges that lie on top of each other, or use some 3D features)

Could be color-encoded or something, sure.

> zoom/fly to coordinate

As we can easily integrate this into GH Maps as overlay or basemap this should be possible like with normal leaflet functionality: https://graphhopper.com/maps/?point=41.95949%2C2.790527

> show/hide details via UI (enable/disable checkbox, move slider etc.)

Definitely possible. 

We could pack all data we have into the MVT and evaluate them on the client side. The current problem is for many points it takes long to convert our data into MVT on the server side. Still it already works reasonable fast, especially if you have just a few cities or even less.

easbar:
Cool!

> The current problem is for many points it takes long to convert our data into MVT on the server side. Still it already works reasonable fast, especially if you have just a few cities or even less.

A single or even a part of a single city would be sufficient for debugging routing problems I think.

>  show turn restrictions ... this get tricky though :)

Ok but why ? Could they not be somehow attached as metadata ? At a minimum we should be able to use an extra endpoint that serves turn restriction information (for example per node id or bounding box) and use this as an extra layer (<- can we draw an extra layer on top of the vector tiles from the client side?)

karussell:
> Ok but why ? Could they not be somehow attached as metadata ? 

Sorry, my wording was bad. I just meant the graphhopper part ... currently we would have to loop through all edges to find if there are turn costs instead of saying "please give me all turn cost in this area". If we have this data we can easily return this as e.g. a multiline with turn_restriction=true or something

> can we draw an extra layer on top of the vector tiles from the client side

Yes, this should be possible. But multiple layers are not necessary as we can use a single layer and evaluate properties of it feature by feature

easbar:
>  currently we would have to loop through all edges to find if there are turn costs instead of saying 

but with `locationIndex.query(bbox)` we can get all nodes in a bbox and then we retrieve all turn restrictions for these node ids ?

>  But multiple layers are not necessary as we can use a single layer and evaluate properties of it feature by feature

Ok yes if we collect all data on the server with a single request we create one layer.

karussell:
> but with locationIndex.query(bbox) we can get all nodes in a bbox and then we retrieve all turn restrictions for these node ids ?

Sure. But still suboptimal if you just want the turn restrictions that are rather rare. But again: shouldn't be a big problem nor a bottleneck, so you can likely ignore my comment ;)

> Ok yes if we collect all data on the server with a single request we create one layer.

(in this PR the approach would be: one request per "tile")

web-bundle/src/main/java/com/graphhopper/resources/MVTResource.java
```diff
@@ -0,0 +1,151 @@
...
+
+    @Inject
+    public MVTResource(GraphHopper graphHopper, EncodingManager encodingManager, DelaunayTriangulationIsolineBuilder delaunayTriangulationIsolineBuilder) {
+        this.graphHopper = graphHopper;
+        this.encodingManager = encodingManager;
+        this.delaunayTriangulationIsolineBuilder = delaunayTriangulationIsolineBuilder;
+    }
+
+    @GET
+    @Produces({MediaType.APPLICATION_JSON})
```
boldtrn: Is this correct? Wouldn't this be `application/x-protobuf` as specified a couple lines lower?

boldtrn:
This PR looks awesome!

> It uses https://github.com/wdtinc/mapbox-vector-tile-java which is nearly 10x faster compared to the other MVT library

I made good experiences with this library as well.

Currently a request would look like `/mvt?x=X&y=Y&z=Z....`?

While there is nothing wrong with this, maybe it would make sense to implement a ZXY style endpoint like:

`/mvt/car/z/x/y.pbf` or `/mvt/z/x/y.pbf?vehicle=car`.

In my experience the ZXY scheme is the most supported scheme by different tools. Also an approach like `/mvt/car/z/x/y.pbf` could make caching easier.

easbar:
This is very nice!

I could make it somewhat run with mapbox gl js here: https://github.com/easbar/graphhopper-graphtool/commit/6779a481ccaf47034d4aa58df91c65059db95b11

Well, at least I can see the roads (need to zoom in a bit and they are all green :))

Any idea how to include text labels on the map ? Would I have to add them on the server side (just like the graph edges) and how ?

boldtrn:
> Any idea how to include text labels on the map ? Would I have to add them on the server side (just like the graph edges) and how ?

The easiest thing is to probably add them as property to the feature like: `name:main road`.

You would then need the appropriate render rule in the style.json. Also for font's you might need a font endpoint to load the appropriate fonts.

Have a look at [Maputnik](https://maputnik.github.io/editor) it makes creating a style and inspecting the underlying data quite easy. You should be able to point it to your GH endpoint from localhost and debug straight away :+1:.

Most modern tile servers generate two different layers, one layer for the roads and one for the labels, but this shouldn't be relevant for this case.

web-bundle/src/main/java/com/graphhopper/resources/MVTResource.java
karussell: Thanks - fixed!

karussell:
Thanks @easbar for trying this - will try this too :) !

> Any idea how to include text labels on the map ? Would I have to add them on the server side (just like the graph edges) and how ?

The problem might be on the client side (style.json?) as this is added on the server side here:  https://github.com/graphhopper/graphhopper/pull/1572/files#diff-09211cea445332d8952b74ca8dcd6f7eR132

easbar:
> You would then need the appropriate render rule in the style.json.

ok thanks! do you have an example for such a render rule ? 
 
> Have a look at Maputnik it makes creating a style and inspecting the underlying data quite easy. You should be able to point it to your GH endpoint from localhost and debug straight away +1.

Ok this sounds great will have a look :) Is maputnik something like mapbox studio but possible to use with custom vector tiles ?

> The problem might be on the client side (style.json?) as this is added on the server side here:

As far as I understood it so far is that on the server side we add a 'layer' (its called 'roads' in our case) and then reference this in the mapbox style like here: https://github.com/easbar/graphhopper-graphtool/blob/6779a481ccaf47034d4aa58df91c65059db95b11/deck-gl/src/map.js#L54.
So on the server side we only define the data and the style on the client side allows us to choose what is shown and how (?). Not really a problem here ?
In my example it also works to draw a deck.gl vector layer (blue/red) on top of the vector tile roads (green atm) (could be useful to draw the shortest path tree for example)
image

Also to me it sounds like it should be possible to show the osm map as second layer side by side ?

karussell:
Got this also working in your deck.gl branch with the green roads :+1: 

Now I think the task is:

 * to create glyphs (or for now use `"glyphs": "mapbox://fonts/mapbox/{fontstack}/{range}.pbf"` but this requires a mapbox key)
 * style the labels via a new layer on the client side (See the `road_major_label` in the maputnik examples)

Still I do not get it working. And no error is shown :(

> So on the server side we only define the data and the style on the client side allows us to choose what is shown and how (?)

Yes. But we can also create different layers on the server side.

> Also to me it sounds like it should be possible to show the osm map as second layer side by side ?

Sounds doable ;)

karussell:
Got the labels working :) 

https://gist.github.com/karussell/f2c7ecafd980afa764e2e0afa734314b

image

easbar:
> Sounds doable ;)

This is really easy using maputnik. You can just add a 'datasource' pointing it to gh-mvt and e.g. use satellite raster tiles (or other vector tiles whatever you want) in addition
image
This is already very helpful because it provides a complete UI to modify the 'style' (gives you the option to filter by road surface type, change colors, show labels and all this). Thanks for pointing this out @boldtrn !

web-bundle/src/main/java/com/graphhopper/resources/MVTResource.java
```diff
@@ -0,0 +1,149 @@
...
+            @Context HttpServletRequest httpReq,
+            @Context UriInfo uriInfo,
+            @QueryParam("vehicle") @DefaultValue("car") String vehicle,
+            @PathParam("z") int zInfo,
+            @PathParam("x") int xInfo,
+            @PathParam("y") int yInfo) {
+
+        if (zInfo <= 9) {
+            VectorTile.Tile.Builder mvtBuilder = VectorTile.Tile.newBuilder();
+            return Response.fromResponse(Response.ok(mvtBuilder.build().toByteArray(), PBF).build())
```
easbar: With maputnik this gives me `Error: http status 200 returned without content`. Can we do better at returning 'an empty tile' ?

boldtrn:
> Got the labels working

That's great to hear! The results look really good!

> to create glyphs (or for now use "glyphs": "mapbox://fonts/mapbox/{fontstack}/{range}.pbf" but this requires a mapbox key)

If you'd like I can also provide you with a [Mapilion](https://mapilion.com/) key and use the Mapilion glyph endpoint.

> Thanks for pointing this out @boldtrn !

No worries, Maputnik helped a couple of times as well :).

web-bundle/src/main/java/com/graphhopper/resources/MVTResource.java
boldtrn: I think it should be possible to convert an empty geometry to an MVT tile.

karussell:
> If you'd like I can also provide you with a Mapilion key and use the Mapilion glyph endpoint.

To avoid the API key hassle and service dependency I tried to use: https://github.com/mapbox/node-fontnik but I wasn't able to install node-fontnik due to compiler errors. Then I went to this repo: https://github.com/openmaptiles/fonts/ which worked easily:

```
npm install
node ./generate.js
cd _output
# see https://github.com/indexzero/http-server for installation
http-server --cors -a localhost -p 8888
```

Then replace the glyphs URL with: `http://localhost:8888/{fontstack}/{range}.pbf` (we could serve them via github pages or something)

( Now the only external dependency is the css which should be trivial to replace :) )

karussell:
Got it integrated in our maps demo

image

easbar:
For the graph debugging I think I will be using this [modified maputnik version](https://github.com/easbar/editor/tree/graphhopper_debug) and I might add some stuff like showing the route or shortest path tree. The mvt tiles are already integrated (and can also be chosen via the 'Open' button).

easbar:
I am getting this error 
``` 
ERROR [2019-03-24 15:26:15,096] org.glassfish.jersey.server.ServerRuntime$Responder: An I/O error has occurred while writing a response message entity to the container output stream.
! java.io.IOException: Broken pipe
! at sun.nio.ch.FileDispatcherImpl.writev0(Native Method)
! at sun.nio.ch.SocketDispatcher.writev(SocketDispatcher.java:51)
! at sun.nio.ch.IOUtil.write(IOUtil.java:148)
! at sun.nio.ch.SocketChannelImpl.write(SocketChannelImpl.java:504)
! at org.eclipse.jetty.io.ChannelEndPoint.flush(ChannelEndPoint.java:266)
! ... 83 common frames omitted
! Causing: org.eclipse.jetty.io.EofException: null
! at org.eclipse.jetty.io.ChannelEndPoint.flush(ChannelEndPoint.java:286)
! at org.eclipse.jetty.io.WriteFlusher.flush(WriteFlusher.java:429)
```
quite regularly. Any idea ?

easbar:
> Do you have an example on how to do this? Which json did you enter in the datasources?

It is basically this (satellite + your speed-colored road layers): https://github.com/easbar/editor/blob/graphhopper_debug/src/gh/assets/gh-mvt-style.json

karussell:
> quite regularly. Any idea ?

Highly likely those exceptions are the result of tiles that the browser initially asked for, but then due to zoom level or bbox changes they are not needed anymore and explicitly rejected from the browser (connection dropped or something). 

So it should not happen if you zoom so slow that all tiles can render. Would be nice if we can somehow properly "fix" them.

> It is basically this (satellite + your speed-colored road layers)

Ah, thanks! (For firefox I had a minor glitched due to mixed content and so I had to replace localhost with 127.0.0.1: http://127.0.0.1:8989/mvt/{z}/{x}/{y}.mvt)

boldtrn:
> but then due to zoom level or bbox changes they are not needed anymore and explicitly rejected from the browser (connection dropped or something).

Yes, the client usually cancels these requests. You can see that the easily in the network tab in the developer tools.

easbar:
> Yes, the client usually cancels these requests. You can see that the easily in the network tab in the developer tools.

Maybe there is an option to delay the client-side requests so not to many requests are sent before the zooming 'stops' at the point one really wants to zoom to ?

boldtrn:
> Maybe there is an option to delay the client-side requests so not to many requests are sent before the zooming 'stops' at the point one really wants to zoom to ?

IMHO both the client and GH work correctly. The problem seems to be that dropwizard does not catch this exception gracefully. Also see https://github.com/graphhopper/graphhopper/issues/1558

easbar:
This has probably been discussed a few times so far, but `main.js` seems to blow up the git repository size (because all its content is on the same line). This is the output I got running this nice [script](https://stackoverflow.com/a/42544963) showing git object sizes. I think the hashs are object ids not git commit hashs. Maybe it would be acceptable to remove main.js from the repository and include a manual build step before running the graphhopper server (could be included in `graphhopper.sh` as well). To be able to run without system-wide node/npm installation there is [nvm](https://github.com/creationix/nvm). Or maybe another easier way would be to not minify main.js (so not all content is on the same line and git can produce smaller diffs) for the repository and only use a minified version of main.js for production ?

```
0ffd88166777  1.3MiB core/files/N49E011.hgt.zip
53404479bc31  1.4MiB core/files/north-bayreuth.osm.gz
623adc98a9e2  1.4MiB core/files/N50E011.hgt.zip
8a76769f58ab  1.4MiB web/src/main/webapp/js/main.js
4dfa59baa221  1.4MiB reader-gtfs/files/beatty.osm
892c6aa9860e  2.0MiB reader-gtfs/files/rnv.zip
a2e129b7087a  2.3MiB web/src/main/resources/com/graphhopper/http/countries.geo.json
2c3edcd23bd2  2.3MiB core/files/spatialrules/countries.geo.json
ccfc0f941342  2.4MiB docs/slac-r-186-fixed-radius-near-neighbors.pdf
e6e61c15b914  2.8MiB web/src/main/webapp/js/main.js
2474cec2eb70  2.8MiB web/src/main/webapp/js/main.js
815c82541fc0  2.9MiB core/files/N48W125.hgt.zip
747c116e5f84  3.0MiB web/src/main/webapp/js/main.js
9b77c19142f3  3.0MiB web/src/main/webapp/js/main.js
7bea392542b3  3.0MiB web/src/main/webapp/js/main.js
491be81e73b2  3.1MiB reader-shp/src/test/resources/data/gis.osm_roads_free_1.shp
6b84456772ce  3.1MiB reader-shp/src/test/resources/data/malta-latest.osm.pbf
116ff9c6482f  3.5MiB reader-shp/src/test/resources/data/gis.osm_roads_free_1.dbf
2804fe44083f  3.6MiB pt-integration-tests/files/rnv.osm.pbf
b64626134bff  3.6MiB pt-integration-tests/files/rnv.zip
3a0a96c9703c  3.6MiB web/src/main/webapp/js/main.js
39cd164c65ff  3.7MiB web/src/main/webapp/js/main.js
c48da93cfe1a  4.0MiB web/src/main/resources/assets/js/main.js
65e62a49ea3f  4.0MiB web/src/main/resources/assets/js/main.js
```

karussell:
> Maybe it would be acceptable to remove main.js from the repository

Yes. Let's remove it. Or maybe we still point to a released main.js and download it instead of building it.

Let's move this discussion to: https://github.com/graphhopper/graphhopper/issues/1584

web-bundle/src/main/java/com/graphhopper/resources/MVTResource.java
karussell: As I cannot reproduce this problem (and also do not know how to create an empty geometry), can we delay and fix this in a later issue?

crazycapivara:
I really like the vector tiles.

Here is an example using the `mvt` endpoint with [openlayers for R](https://github.com/crazycapivara/openlayers)
```r
devtools::install_github("crazycapivara/openlayers")

library(openlayers)

map_source <- "http://localhost:8989/mvt/{z}/{x}/{y}.mvt"

# Basic example
ol() %>%
  add_vector_tiles(map_source) %>%
  set_view(9.503174, 52.583026, 12)
```

Adding some styles from [here](https://gist.github.com/crazycapivara/5efd362173604e25b985ce1cf68f2dad):

```r
fn <- "https://gist.github.com/crazycapivara/5efd362173604e25b985ce1cf68f2dad/raw/069281132f093ce11a874daaf6293e0397770408/gh-ol-mvt-style.js"
style <- read_js_function(fn)

ol() %>%
  add_vector_tiles(map_source, style = style) %>%
  set_view(9.503174, 52.583026, 12)

# Add a raster layer as well ...
ol() %>%
  add_stamen_tiles("toner", options = list(name = "stamen")) %>%
  add_vector_tiles(map_source, style = style, options = list(name = "gh")) %>%
  set_view(9.503174, 52.583026, 12) %>%
  add_layer_switcher()
```

karussell:
@crazycapivara nice, thanks!

Here it is now also possible to include more information. On the server side you need the `graph.encoded_values` property, see https://github.com/graphhopper/graphhopper/blob/master/config-example.yml#L19

And on the client side you request a sub set of them, see https://github.com/graphhopper/graphhopper/pull/1572/commits/bf3ece86bc055841f430472dfd397ec8cbef7a44#diff-fdabf1e86cc9047458301bd084fc7a8b

Additionally we now include a different level of details in different zoom levels based on the road_class (OSM highway tag):

image

PR #1577, opened 2019-03-26
Improve pointlist output for isochrone

karussell:
## Update: this is now a pure CSV endpoint:

```csv
longitude,latitude,time,distance
lon1, lat1, time1, distance1
lon2, lat2, time2, distance2
```

where you can add more columns like `prev_latitude,prev_longitude` or path details like `road_class`

## Original text:

The pointlist option was not really useful as it came with just the coordinates. ~Have now created a compact but extendable JSON format~ (read update).

```json
{  "description": ["longitude", "latitude", "time", "distance"],
   "items": [
     [lng, lat, time, distance],
     [...]
   ]
} 
```

Thx @crazycapivara for the valuable discussion! He created a [deck.gl for R](https://crazycapivara.github.io/deckgl/) tool and future versions will likely work with graphhopper. See also the work [here](https://www.graphhopper.com/blog/2018/07/04/high-precision-reachability/) with deck.gl and a forked graphhopper (this should be possible after this PR without forking). See also related #1572.

TODOs:

 - [x] use different endpoint name, e.g. /shortest-path-tree or /spt
 - [x] create `List<IsoLabelWithCoordinates>` in IsochroneResource to make Isochrone method more powerful?
 - [x] ~allow paging through the pointlist result with some caching?~ later


crazycapivara:
Here is an example on how to visualize data with _deckgl_ and _R_

```r
library(magrittr)
library(jsonlite)
library(deckgl)

lat <- 51.33399
lng <- 9.438801

# Get data
apiCall <- paste0(
  "http://localhost:8989/isochrone?point=", lat, ",", lng,
  "&time_limit=600&result=pointlist&pointlist_ext_header=prev_longitude,prev_latitude"
)

resp <- httr::GET(apiCall) %>%
  httr::content("text") %>%
  fromJSON(flatten = TRUE)

# Prepare data
tbl <- resp$items %>%
  set_colnames(resp$header) %>%
  tibble::as_tibble() %>%
  subset(time != 0)

# Plot data
properties <- list(
  getStrokeWidth = 3,
  getSourcePosition = get_position("prev_latitude", "prev_longitude"),
  getTargetPosition = get_position("latitude", "longitude"),
  getTooltip = get_property("distance")
)

# Create deckgl widget
deck <- deckgl(latitude = lat, longitude = lng, pickingRadius = 2)

deck_black <- deckgl(
  latitude = lat,
  longitude = lng,
  pickingRadius = 2,
  style = list(background = "black")
)

# Arc layer
deck %>%
  add_arc_layer(data = tbl, properties = properties)

# Create colors for line layer
pal <- leaflet::colorBin(tbl$distance)
tbl$color <- pal(tbl$distance)

# Line layer
deck_black %>%
  add_line_layer(
    data = tbl, properties = properties,
    getColor = get_color_to_rgb_array("color")
  )
```

karussell:
Thanks a lot!

> subset(time != 0)

Here it says something about the type: 

> Fehler in time != 0 :   Vergleich (2) ist nur für atomare und Listentypen möglich

karussell:
Ups, using a different starting point fixed this for me. E.g. for Berlin:
```R
lat <- 52.509535
lng <- 13.40332
```

karussell:
Nice, got some streetnetwork as output -> looks funny those curved streets. Thanks @crazycapivara!

For the leaflet part I had to do:

```
install.packages("leaflet")
library(leaflet)
```

But then I get:
>Fehler in leaflet::colorBin(tbl$distance) : 
>  Argument "domain" fehlt (ohne Standardwert)

crazycapivara:
Sorry, I posted an old version of the script, will send an working update.

crazycapivara:
Start a container with Berlin data set:

```bash
docker run -p 8989:8989 -d --name gh-berlin crazycapivara/graphhopper:ext_pointlist 
```
Run some visualizations wit _R_ and _deckgl_:

```r
library(magrittr)
library(jsonlite)
library(deckgl)

lat <- 52.517057
lng <- 13.3992

# Get data
apiCall <- paste0(
  "http://localhost:8989/isochrone?point=", lat, ",", lng,
  "&time_limit=600&result=pointlist&pointlist_ext_header=prev_longitude,prev_latitude"
)

resp <- httr::GET(apiCall) %>%
  httr::content("text") %>%
  fromJSON(flatten = TRUE)

# Prepare data
tbl <- resp$items %>%
  set_colnames(resp$header) %>%
  tibble::as_tibble() %>%
  subset(time != 0)

# Plot data
properties <- list(
  getStrokeWidth = 3,
  getSourcePosition = get_position("prev_latitude", "prev_longitude"),
  getTargetPosition = get_position("latitude", "longitude"),
  getTooltip = get_property("distance")
)

# Create deckgl widget
deck <- deckgl(latitude = lat, longitude = lng, pickingRadius = 2)

# Arc layer
deck %>%
  add_arc_layer(data = tbl, properties = properties)

# Create colors for line layer
pal <- leaflet::colorBin("RdYlBu", tbl$distance)
tbl$color <- pal(tbl$distance)

# Create a deckgl widget with black background
deck_black <- deckgl(
  latitude = lat,
  longitude = lng,
  pickingRadius = 2,
  style = list(background = "black")
)

# Line layer
deck_black %>%
  add_line_layer(
    data = tbl, properties = properties,
    getColor = get_color_to_rgb_array("color")
  )

# Arc and line layer
deck %>%
  add_data(tbl) %>%
  add_arc_layer(data = get_data(), properties = properties) %>%
  add_line_layer(
    data = get_data(), properties = properties,
    getColor = get_color_to_rgb_array("color")
  )
```

If you have valid mapbox api key you can add a basemap as well:

```r
deck %>%
   ... %>%
  add_mapbox_basemap(api_key)
```

crazycapivara:
graphhopper-deckgl-R

crazycapivara:
Maybe we could add an `offset` parameter to the query to avoid too large data sets to be transferred at once, e. g.:

```
...?time_limit=3600&offset=3000
```

would return points where `time_limit` is between 3000 and 3600.

Furthermore, at the moment, the request returns a lot of points above the given `time_limit`.

karussell:
> would return points where time_limit is between 3000 and 3600

Why would this help? To print parts of the result before everything arrived? The problem is that we would then have to remember the state somehow (cache) which I would like to avoid if possible.

> Furthermore, at the moment, the request returns a lot of points above the given time_limit.

Ah, yes. This is ugly and I can fix it here: https://github.com/graphhopper/graphhopper/blob/master/isochrone/src/main/java/com/graphhopper/isochrone/algorithm/Isochrone.java#L94

crazycapivara:
> Why would this help? To print parts of the result before everything arrived? The problem is that we would then have to remember the state somehow (cache) which I would like to avoid if possible.

Yes, this was one idea and to maybe allow larger `time_limits`  but I can unsterstand that you want to avoid caching.

karussell:
Hmmh ... if we allow caching then we could return the isochrone-shortest-path-tree via the MVT endpoint (#1572), which would save us from a non-standard format.

crazycapivara:
Ups, Nope, I would like to keep the format. It is perfect for a lot of use cases. I just connected the API to a postgis container, so that you can query the roads with `SELECT way, time, distance FROM graphhopper.way(lat, lng, time_limit)`. Furthermore, `deck.gl` supports the format as well.

crazycapivara:
see [docker-postgis-graphhopper](https://github.com/crazycapivara/docker-postgis-graphhopper)

karussell:
Ah, ok. Thanks for the feedback!

> It is perfect for a lot of use cases. I just connected the API to a postgis container, so that you can query the roads with 

This sounds great!

michaz:
> The pointlist option was not really useful as it came with just the coordinates. Have now created a compact but extendable JSON format:

@karussell Did you just invent CSV-over-JSON? ,-)

karussell:
Lol. I'm pretty sure this is already existent :D ... how do you like it ;) ?

michaz:
I like it! But maybe consider using CSV right away for this endpoint?

- It's exactly CSV. Variable number of columns, string-indexed.
- CSV is inherently streamable. You _can_ parse JSON as a stream, but a JSON output still has this "I am a document, not a stream of lines. Wait for the closing bracket before you start interpreting me" feeling to it.
- People are doing the _opposite_, JSON-over-CSV, for when you have structured data, so CSV seems to have some merit..  http://csvjson.org

@crazycapivara Any thoughts?

crazycapivara:
For now I would like to add it as a second mediatype and compare both endpoints for my use cases.

karussell:
A related info: I just stumbled over https://www.interline.io/blog/geojsonl-extracts/ that mentions GeoJSON Line and also the following streamable JSON stuff:

 * http://jsonlines.org/
 * http://ndjson.org/ 
 * https://tools.ietf.org/html/rfc8142 (from mapbox)
 * https://macwright.org/2015/03/23/geojson-second-bite.html#streaming

web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
```diff
@@ -124,19 +119,71 @@ public Response doGet(
...
+                        case "edge_id":
+                            list.add(label.edgeId);
+                            break;
+                        case "longitude":
+                            list.add(label.adjCoordinate.x);
+                            break;
+                        case "latitude":
+                            list.add(label.adjCoordinate.y);
+                            break;
+                        case "prev_longitude":
```
Komzpa: Can we please add options to export previous node's time and distance, so we can get values in both ends of segment without a lookup join?
```
                        case "prev_distance":
                            list.add(Optional.ofNullable(index.get(label.baseNodeId)).map(i -> i.distance).orElse(0D));
                            break;
                        case "prev_time":
                            list.add(Optional.ofNullable(index.get(label.baseNodeId)).map(i -> i.time).orElse(0L));
                            break;
```

Komzpa:
This thing is useful for us. Format is fine, the only thing is - can we have travel times and distances in both ends of a segment, and export as such? Snippet attached.

web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
karussell: Do you have an example why this would be useful for you?

Komzpa: Yeah. Builiding a constrained Delaunay triangulation and clipping it in PostGIS, I need to have two-point segments to feed into function. Here's more detail: https://www.patreon.com/posts/isochrones-are-20933638

karussell:
Have added it. Please note that I have changed the time to seconds to reduce bandwidth a bit and ms does not make much sense - or is there a good reason to keep milliseconds?

karussell:
> for this ~endpoint~ variant of the endpoint?

Moved this now in a separate endpoint

karussell:
Have it now further refactored and I'm fine with merging this, especially since the format is simple and the endpoint is separate from /isochrone.

@crazycapivara @Komzpa can you try once again if this is still working for you? Please note the different endpoint and different unit like mentioned in a previous comment.

@michaz let me know if you see something ugly. (the different format flavoures could be indeed implemented via different mediatypes or should we already use something != `application/json`)

michaz:
> let me know if you see something ugly

Nothing ugly seen. :-) Would only suggest removing the `result=pointlist` from the test query (this isn't a parameter on this endpoint anymore).

I'm not really in favor of supporting multiple formats here. I have a slight preference for using CSV instead of JSON, because it seems just like what CSV is often used for, but CSV-like JSON is fine with me, since "all the rest of the API is JSON" may be an argument in favor.

michaz:
Oh, and could we try and find a more descriptive name for the `pointlist_ext_header` parameter? `pointlist` is redundant, `ext` doesn't mean a lot, and it's not really about the `header` either. Maybe.... <thinks>

michaz:
How about `columns`? And then, I'd suggest to use it in a different way: The user specifies _all_ the columns they want, not only those in addition to `"longitude", "latitude", "time", "distance"`. But let `"longitude", "latitude", "time", "distance"` be the default.

Thus, all columns are treated equal. If I want `"latitude", "longitude", "distance"`, I say `columns=latitude,longitude,distance`.

michaz:
> or is there a good reason to keep milliseconds?

Consistency. Milliseconds are used in the other API responses, and in the Java API. It has some value being able to know things like "GraphHopper travel times are in milliseconds".

karussell:
Thanks - good questions and suggestions! Will work on them :)

karussell:
Before I merge I'll wait for feedback on the current status and regarding CSV vs. JSON from @crazycapivara and @Komzpa. This issue is really about making the graph from GraphHopper better accessible from "outside" and so a simpler CSV format that is also streamable would be preferred from our side. **But** if this is more complicated, then I would propose using the current JSON response and add CSV later if required at all.

Komzpa:
Hi, we're using it with other server app on local loopback, so any readable format will do. CSV will come with a benefit of nobrainer shell import to postgres tables for adhoc queries.

michaz:
> (think about a real time SPT traversal in the browser)

@karussell 
Did that once for debugging transit routing. Works and looks cute.

crazycapivara:
@karussell CSV is fine for me as well.

karussell:
Great - thanks for the feedback! Will then move this endpoint to CSV-only (which would have the advantage of streeming in later implementations)

karussell:
Ok, have now converted this endpoint into returning CSV. Additionally as dropwizard makes this very easy, the CSV is already streamed, i.e. even for large shortest path trees the rendering engine gets something to do right after the request :)

karussell:
Oh, this means we can fetch country-sized (or even continental-sized) shortest path trees without any memory problems. The client just waits a bit for the data transmission and sees the exploration in (near) real time.

karussell:
Country-sized SPTs are definitely doable with this endpoint:

image

The code snippet is [here](https://gist.github.com/karussell/a8203e66c4fea399d94189570be89451).

The data is not small (gzipped size=70MB) but the response time is ok (< 10sec) and there is a good solution for [streamed CSV parsing](https://github.com/mholt/PapaParse) for the browser.

But with leaflet (renderer:canvas) I was not able to get updates while downloading and also everything was very slow (updates takes ages for loading and zooming).

Would be curious if the [deck.gl for R](https://crazycapivara.github.io/deckgl/) work from @crazycapivara is able to handle that somehow :)

michaz:
Nice!

crazycapivara:
```r
lat <- 52.321911
lng <- 10.393066
time_limit <- 7200

api_call <- paste0(
  "http://localhost:8989/spt?point=",
  lat, ",", lng,
  "&time_limit=", time_limit,
  "&columns=prev_longitude,prev_latitude,longitude,latitude,time"
)

dt <- data.table::fread(api_call, na.strings = "null") %>%
  na.omit()

pal <- leaflet::colorBin("RdYlBu", dt$time)
dt$color <- pal(dt$time)

library(deckgl)

deckgl(
  latitude = lat,
  longitude = lng,
  style = list(background = "black")
) %>%
  add_line_layer(
    data = dt,
    getStrokeWidth = 3,
    getSourcePosition = get_position("prev_latitude", "prev_longitude"),
    getTargetPosition = get_position("latitude", "longitude"),
    getTooltip = get_property("time"),
    getColor = get_color_to_rgb_array("color")
  )
```

In this case it takes some time until the layer is rendered because first all data is downloaded and then passed to the widget (from R to json). But in the end when it shows up zooming is smooth.

I need to add a csv-reader on javascript side to speed up loading.

crazycapivara:
[
gh-spt
](url)

Added a pure _JavaScript_ snippet [here](https://gist.github.com/crazycapivara/8e4716eb9ba3945357a64276e0fe1fdc) where you can click on the map to set the _center point_ and a drop down list to select the _time limit_.

michaz:
Love it!
@karussell What do you think about this:

> The user specifies all the columns they want, not only those in addition to "longitude", "latitude", "time", "distance". Just let "longitude", "latitude", "time", "distance" be the default.

> Thus, all columns are treated equal. If I want "latitude", "longitude", "distance", I say columns=latitude,longitude,distance.

I think that would it more flexible (can also drop columns) and easier do document: "Specify the columns you want" instead of "specify the columns _in addition to_ the default columns that you want".

karussell:
> I think that would make it more flexible (can also drop columns) and easier do document

Yes, this was a good idea! And if I did not misunderstand it, then this is already done :D ?

BTW: in some days or weeks we'll have #1548 i.e. information like road_class, max_speed, tunnel, toll etc then we can make this available here too. Furthermore we can allow to specify a filter if only main roads are required (or a smaller+faster output) or visualizing tunnels and bridges is required for risk analysis etc

crazycapivara:
@karussell That sounds great, then we can also use this information for the styling.

karussell:
You can now request additional information like road_class (highway tag) and road_environment (tunnel, bridge, ...) and more. You need three things:

 * update to latest version
 * use the following param in config.yml: `graph.encoded_values: surface,road_class,road_environment,road_access,max_speed`
 * do a re-import
 * specify them as additional columns in the request

karussell:
@michaz do you want to review this again? Especially as you were asking:

> The user specifies all the columns they want, not only those in addition to "longitude", "latitude", "time", "distance".

karussell:
If no opposing opinion I'll merge this today

Komzpa:
I haven't run it but reading through description looks good.

karussell:
Ok. If there are problems we can easily create new issues :)

PR #1581, opened 2019-03-28
Fixes wrong routing occurring when stall on demand is used with virtual via-nodes.

### LINKED ISSUE #1574: Rare bug for stall on demand
Currently I'm facing an issue when stall_on_demand=true (if false it disappears). It is very rare and hard to reproduce in a unit test.

The reason is likely in DijkstraBidirectionCH.entryIsStallable:
```
adjNode.weight + weighting.calcWeight(iter, !reverse, getIncomingEdge(entry)) < entry.weight
```

It returns true although it should return false. In my case it is the 4th decimal place: `35.643+50.2975 < 50.297522`

So, when I round to the precision we currently use in CHGraphImpl:
`Math.round(entry.weight / 1000f) * 1000f`
the problem disappears. But maybe it is related to virtual nodes because the special case is a QueryGraph with 3 locations where the fastest route B->C crosses point A.

easbar:
Fixes #1574.

I added a test that creates random routing situations including ones that lead to wrong routing results (sometimes connection not found but also wrong weights). This PR also adds a failing integration test and a specific test with only 10 nodes that shows the problem in detail (also see comments in #1574).

This PR fixes all these tests and also the random test did not fail after running it for more than 15 minutes. 

The performance measurements on the Germany map showed that the query performance before and after this fix is the same.

When I was experimenting with the the random test I ran into two more (probably very rare) error situations: #1582, #1583


core/src/main/java/com/graphhopper/routing/DijkstraBidirectionCH.java
```diff
@@ -62,10 +64,16 @@ private boolean entryIsStallable(SPTEntry entry, IntObjectMap<SPTEntry> bestWeig
         // reached via a suboptimal path. We do this regardless of the CH level of the adjacent nodes.
         EdgeIterator iter = edgeExplorer.setBaseNode(entry.adjNode);
         while (iter.next()) {
+            // no need to inspect the edge we are coming from
+            if (iter.getEdge() == entry.adjNode) {
```
easbar: First I was hoping this might improve performance, but not really. In the `iter.getEdge() == entry.adjNode` case `adjNode.weight+weighting.calcWeight` always exactly equals `entry.weight` below and in this case we never want to stall.

core/src/test/java/com/graphhopper/routing/ch/PrepareContractionHierarchiesTest.java
```diff
@@ -463,18 +463,14 @@ public void testStallOnDemandViaVirtuaNode_issue1574() {
         // the weight of edge 3-1 must be such that node 2 gets stalled in the forward search via the incoming shortcut
         // at node 2 coming from 3. this happens because due to the virtual node x between 3 and 1, the spt entries
         // calculated on the query graph (using the virtual edges) use different floating point rounding / arithmetics.
-        IntsRef flags = edge31.getFlags();
-        carEncoder.getAverageSpeedEnc().setDecimal(false, flags, 22);
-        edge31.setFlags(flags);
+        edge31.set(carEncoder.getAverageSpeedEnc(), 22);
```
easbar: This is the 'new' & right way to do it, right ? very nice!

core/src/main/java/com/graphhopper/routing/VirtualEdgeIterator.java
```diff
@@ -213,7 +213,11 @@ public EdgeIteratorState setName(String name) {
 
     @Override
     public String toString() {
-        return edges.toString();
+        if (current >= 0 && current < edges.size()) {
+            return "virtual edge: " + edges.get(current) + ", all: " + edges.toString();
+        } else {
+            return "virtual edge: (invalid)" + ", all: " + edges.toString();
```
karussell: Shouldn't this better throw an IllegalStateException?

core/src/test/java/com/graphhopper/routing/ch/PrepareContractionHierarchiesTest.java
karussell: Yes :)

core/src/main/java/com/graphhopper/routing/VirtualEdgeIterator.java
easbar: No I don't think so because `current=-1` is the state the iterator is in before calling `next()` for the first time.

karussell:
Will cherry-pick for 0.12 too

core/src/main/java/com/graphhopper/routing/DijkstraBidirectionCH.java
easbar: Oh this was a stupid error, it should be `entry.edge` instead of `entry.adjNode`, fixed in ec01e5e458bce46ffddea8a7b266cacbc49e6a1e

easbar:
See also ec01e5e458bce46ffddea8a7b266cacbc49e6a1e

Issue #1585, opened 2019-03-28
Issues with time calculation.


easbar:
I think the routing algorithms do not properly calculate the times when turn costs are involved. I identified three problems:

- [x] The turn cost time is not contributing enough to the traveling time, for example for a 100m edge setting the turncosts to 2 increases the weight from 18s to 20s while the time is only increased from 18s to 18.002s. Fixed in #1586.

- [x] edge-based CH does not include the turn cost times in the calculated time at all. Everything is fine for the weight calculation and when using the fastest weighting the fastest route will be found with respect to turn costs, but `Path.getTime()` will not include the turn cost times. This has to be adjusted in `Path4CH`. Fixed in #1611.

- [x] The bidirectional algorithms (dijkstra&astar, probably also ALT) do not pay attention to turn costs at the meeting point of fwd&bwd search (the turn costs are properly included for the weight calculation, but the turn cost *time* is not added when unpacking the path. Fixed in #1586.

- [ ] `TurnWeighting` does not include default u-turn cost in `calcMillis`.

For turn *restrictions* (infinite not finite turn costs) this issue is not important, because restricted turns will not be used anyway.

See this test: 44f5c068d4e672dd905ee0b55bc5bd5095c30991

PR #1586, opened 2019-03-29
Fixes some issues with turn cost times.

easbar:
Fixes two of the three issues mentioned in #1585.

Added some tests and did some cleanup in some tests, but the actual fixes are one-liners.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/PathBidirRef.java
+++ b/core/src/main/java/com/graphhopper/routing/PathBidirRef.java
@@ -83,7 +83,7 @@ public class PathBidirRef extends Path {
         setFromNode(currEdge.adjNode);
         reverseOrder();
         currEdge = edgeTo;
-        int prevEdge = nextEdgeValid ? sptEntry.edge : EdgeIterator.NO_EDGE;
+        int prevEdge = EdgeIterator.Edge.isValid(sptEntry.edge) ? sptEntry.edge : EdgeIterator.NO_EDGE;
         int tmpEdge = currEdge.edge;
         while (EdgeIterator.Edge.isValid(tmpEdge)) {
             currEdge = currEdge.parent;
--- a/core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
+++ b/core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
@@ -89,15 +89,14 @@ public class TurnWeighting implements Weighting {
         if (prevOrNextEdgeId == EdgeIterator.NO_EDGE)
             return millis;
 
-        // TODO for now assume turn costs are returned in milliseconds?
         // should we also separate weighting vs. time for turn? E.g. a fast but dangerous turn - is this common?
-        long turnCostsInMillis;
+        long turnCostsInSeconds;
         if (reverse)
-            turnCostsInMillis = (long) calcTurnWeight(edgeState.getEdge(), edgeState.getBaseNode(), prevOrNextEdgeId);
+            turnCostsInSeconds = (long) calcTurnWeight(edgeState.getEdge(), edgeState.getBaseNode(), prevOrNextEdgeId);
         else
-            turnCostsInMillis = (long) calcTurnWeight(prevOrNextEdgeId, edgeState.getBaseNode(), edgeState.getEdge());
+            turnCostsInSeconds = (long) calcTurnWeight(prevOrNextEdgeId, edgeState.getBaseNode(), edgeState.getEdge());
 
-        return millis + turnCostsInMillis;
+        return millis + 1000 * turnCostsInSeconds;
     }
 
     /**
--- a/core/src/test/java/com/graphhopper/routing/weighting/TurnWeightingTest.java
+++ b/core/src/test/java/com/graphhopper/routing/weighting/TurnWeightingTest.java
@@ -1,0 +1,62 @@
+package com.graphhopper.routing.weighting;
+
+import com.graphhopper.routing.util.CarFlagEncoder;
+import com.graphhopper.routing.util.EncodingManager;
+import com.graphhopper.routing.util.FlagEncoder;
+import com.graphhopper.storage.Graph;
+import com.graphhopper.storage.GraphBuilder;
+import com.graphhopper.storage.TurnCostExtension;
+import com.graphhopper.util.EdgeIteratorState;
+import org.junit.Before;
+import org.junit.Test;
+
+import static com.graphhopper.util.GHUtility.getEdge;
+import static org.junit.Assert.assertEquals;
+
+public class TurnWeightingTest {
+
+    private Graph graph;
+    private FlagEncoder encoder;
+    private EncodingManager encodingManager;
+    private Weighting weighting;
+    private TurnCostExtension turnCostExt;
+
+    @Before
+    public void setup() {
+        encoder = new CarFlagEncoder(5, 5, 10);
+        encodingManager = EncodingManager.create(encoder);
+        graph = new GraphBuilder(encodingManager).create();
+        weighting = new FastestWeighting(encoder);
+        turnCostExt = (TurnCostExtension) graph.getExtension();
+    }
+
+    @Test
+    public void calcWeightAndTime_withTurnCosts() {
+        graph.edge(0, 1, 100, true);
+        EdgeIteratorState edge = graph.edge(1, 2, 100, true);
+        // turn costs are given in seconds
+        addTurnCost(0, 1, 2, 5);
+        TurnWeighting turnWeighting = new TurnWeighting(weighting, turnCostExt);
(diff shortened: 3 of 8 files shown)


core/src/main/java/com/graphhopper/routing/PathBidirRef.java
```diff
@@ -83,7 +83,7 @@ public Path extract() {
         setFromNode(currEdge.adjNode);
         reverseOrder();
         currEdge = edgeTo;
-        int prevEdge = nextEdgeValid ? sptEntry.edge : EdgeIterator.NO_EDGE;
+        int prevEdge = EdgeIterator.Edge.isValid(sptEntry.edge) ? sptEntry.edge : EdgeIterator.NO_EDGE;
```
easbar: This fixes the turn cost time evaluation at the meeting node

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
```diff
@@ -89,15 +89,14 @@ public long calcMillis(EdgeIteratorState edgeState, boolean reverse, int prevOrN
...
+        long turnCostsInSeconds;
         if (reverse)
-            turnCostsInMillis = (long) calcTurnWeight(edgeState.getEdge(), edgeState.getBaseNode(), prevOrNextEdgeId);
+            turnCostsInSeconds = (long) calcTurnWeight(edgeState.getEdge(), edgeState.getBaseNode(), prevOrNextEdgeId);
         else
-            turnCostsInMillis = (long) calcTurnWeight(prevOrNextEdgeId, edgeState.getBaseNode(), edgeState.getEdge());
+            turnCostsInSeconds = (long) calcTurnWeight(prevOrNextEdgeId, edgeState.getBaseNode(), edgeState.getEdge());
 
-        return millis + turnCostsInMillis;
+        return millis + 1000 * turnCostsInSeconds;
```
easbar: Here we have to convert the turn costs to milliseconds to make the turn cost time calculation consistent with the weight calculation.

core/src/test/java/com/graphhopper/routing/weighting/TurnWeightingTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Test
+    public void calcWeightAndTime_withTurnCosts_shortest() {
+        graph.edge(0, 1, 100, true);
+        EdgeIteratorState edge = graph.edge(1, 2, 100, true);
+        // turn costs are given in seconds
+        addTurnCost(0, 1, 2, 5);
+        TurnWeighting turnWeighting = new TurnWeighting(new ShortestWeighting(encoder), turnCostExt);
+        // todo: for the shortest weighting turn costs cannot be interpreted as seconds ? at least when they are added
+        // to the weight ? how much should they contribute ?
+//        assertEquals(105, turnWeighting.calcWeight(edge, false, 0), 1.e-6);
```
easbar: Any thoughts here ?

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
karussell: Maybe we should decide before going more public with turn costs to make a breaking change and use the same unit for turn costs? I think I stumbled over the same problem somewhere.

core/src/test/java/com/graphhopper/routing/weighting/TurnWeightingTest.java
karussell: What do you think about deprecating shortest weighting in general? It does not make sense in practise. Most people say they need "shortest" for car (hear this often for taxis as they calculate the cost based on the distance), but what they really mean is a "fastest" cost calculation with a tight distance weight.

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
easbar: You mean store the turn costs as milliseconds ? In this PR or is this just a reminder for later ? Storing turn costs as milliseconds seems a bit extreme, seconds should be enough ? Maybe use 0.1 seconds ?

core/src/test/java/com/graphhopper/routing/weighting/TurnWeightingTest.java
easbar: Ok I see. But not sure if it should be really deprecated, because someone might find it still useful. We could emphasize somewhere that the ShortestFastestWeighting is typically what most people are looking for. Anyway the question how turn costs should contribute to weightings that are not purely based on time remains also for the ShortestFastestWeighting. In ShortestFastestWeighting this is controlled by the time/distanceFactors.

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
karussell: > You mean store the turn costs as milliseconds ?

I would keep storing them as seconds (and later use the same EncodedValue classes to define a different precision), but shouldn't we use the same unit `ms` in the Java API `turnCostEncoder.getTurnCost`? Furthermore to make it more clear maybe we should rename `calcTurnWeight` into `calcTurnCost`? Or even introduce a separate method `calcTurnMillis` like we have for normal weighting ... but are there really cases where a priority to avoid a turn is different to the turn time?

> In this PR or is this just a reminder for later?

Good question - probably later :)

core/src/test/java/com/graphhopper/routing/weighting/TurnWeightingTest.java
karussell: For now IMO the turn costs for ShortestWeighting must be 0 as there are no ways to introduce a timely slow down or priority stuff.

In ShortestFastestWeighting the turn cost adds to the time (and gets multiplied by the timeFactor)

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
easbar: > but shouldn't we use the same unit ms in the Java API turnCostEncoder.getTurnCost

The way I understood it so far is that the turn costs do not necessarily have to be turn *times*, maybe someone is using an advanced weighting and drivers just do not want to take turns so each turn adds some penalty (but not necessarily time) to some advanced weight function ? 

> Furthermore to make it more clear maybe we should rename calcTurnWeight into calcTurnCost?

Hmm yeah makes sense because we call it turn cost everywhere else, then again `calcTurnWeight` fits `calcWeight` in the `TurnWeighting` class.. no idea.

> but are there really cases where a priority to avoid a turn is different to the turn time

Hmm I could imagine this, but don't know if anybody does this :)

core/src/test/java/com/graphhopper/routing/weighting/TurnWeightingTest.java
easbar: Yes, I agree as long as we do not allow giving less priority to routes with many turns etc. the shortest weighting should pay attention to turn restrictions, but the turn costs (assuming they are turn *times*) should not be included. However to make this work we would in fact have to do this:

>  Or even introduce a separate method calcTurnMillis like we have for normal weighting ... but are there really cases where a priority to avoid a turn is different to the turn time?

because right now the TurnWeighting simply adds the turn costs without knowing what it's underlying Weighting is.

easbar:
Will merge this and open separate issue for the turn cost issues discussed here.

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
karussell: > Hmm I could imagine this, but don't know if anybody does this :)

But which use case do you have in mind where "time != weight"?

The only thing that comes to my mind is an efficient storage. E.g. only two bits are stored and the 0 maps to no turn time, the 1 maps to 5 seconds, the 2 maps to 10 seconds and the 3 maps to infinity (turn restriction). But once we would introduce something like the MappedDecimalEncodedValue I do not see a reason to use the same `ms` as unit.

Maybe there are already users of this more generic turn cost feature, but if we document this change this is no problem IMO. And we should not hesitate to make this API more consistent before it is used more widespread.

BTW: As the defaultUTurnCost is in seconds - is this correct?

core/src/test/java/com/graphhopper/routing/weighting/TurnWeightingTest.java
karussell: Hmmh, indeed. This is ugly and one reason more (for me) to remove ShortestWeighting :D

easbar: see #1590

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
easbar: see #1590

PR #1594, opened 2019-04-08
Update documentation regarding usage of EncodingManager's constructor

Nakaner:
The constructor is private and you have to use the factory methods.


karussell:
Ups, thanks a lot!

Issue #1550, opened 2019-02-22
directions-api-client-hc 0.11+ requires Android min sdk version >= 26


devemux86:
directions-api-client-hc 0.11+ now has as dependency the graphhopper-web-api whose some code seems requiring Android min sdk version >= 26.

Note that directions-api-client-hc 0.10.3 works fine on Android, even in older versions.

```java
Caused by: com.android.builder.dexing.DexArchiveBuilderException: Error while dexing com/graphhopper/jackson/MultiExceptionSerializer.class
	at com.android.builder.dexing.DxDexArchiveBuilder.getExceptionToRethrow(DxDexArchiveBuilder.java:143)
	at com.android.builder.dexing.DxDexArchiveBuilder.convert(DxDexArchiveBuilder.java:89)
	at com.android.build.gradle.internal.transforms.DexArchiveBuilderTransform.launchProcessing(DexArchiveBuilderTransform.java:904)
	... 6 more
Caused by: com.android.dx.cf.code.SimException: ERROR in com.graphhopper.jackson.MultiExceptionSerializer.serialize:(Lcom/graphhopper/MultiException;Lcom/fasterxml/jackson/core/JsonGenerator;Lcom/fasterxml/jackson/databind/SerializerProvider;)V: invalid opcode ba - invokedynamic requires --min-sdk-version >= 26 (currently 16)
```

Since client-hc is used also in [graphhopper-navigation-example](https://github.com/graphhopper/graphhopper-navigation-example), such Android crashes could be catched if it [uses](https://github.com/graphhopper/graphhopper-navigation-example/blob/master/app/build.gradle) latest GraphHopper release and a Android recommended `minSdkVersion` (instead of 23).

boldtrn:
I think the navigation example uses the generated client that can be found here: https://github.com/graphhopper/directions-api-clients

So this should be unrelated to this issue.

devemux86:
Off topic: my advice would be to lower its `minSdkVersion` to current Android recommendation (15 or 16) and test it also on older Android emulator to cover all corner cases. Unless some dependency requires such high min Android version, but that leaves outside a large percentage of Android users and most developers definitely don't want that forced from their libraries. :slightly_smiling_face:

boldtrn:
> Unless some dependency requires such high min Android version, but that leaves outside a large percentage of Android users and most developers definitely don't want that forced from their libraries.

[...]

PR #1595, opened 2019-04-08
Don't interpolate travel times in GPX; fixes #1490

### LINKED ISSUE #1490: Time stamps calculated by  `createGPXList()` are sometimes inaccurate
The time stamps of a GPXList obtained from `createGPXList()` are sometimes inconsistent with the results of another GraphHopper call. The duration of a route from the origin to the the point of the entry in the GPXList can differ from the time stamp of the entry by several seconds.

This happens even when all requests are using the "fastest" weighting. See [here for a failing test of createGPXList()](https://discuss.graphhopper.com/t/time-stamps-in-gpxlist-inaccurate/3629/12).

This issue occurs with Graphhopper versions 0.10.3, 0.11.0, and the master branch as of the time of this writing.

michaz:
(See issue #1490.) The reason those timestamps are inaccurate is that they are made up. The code that generates the GPX thinks that every track point must have a timestamp, so it just makes up the timestamps for points where none is available (pillar nodes, I guess).

But that's not true, timestamps are optional. Just don't include them when we don't have one.

We _could_ get accurate travel times out of the routing engine, but that's not what the code wants to do -- structurally, it says that a GPX track is another representation of an instruction list, and the instruction list doesn't know travel times at pillar nodes.

This was all pretty hidden and non-obvious, because the code was mixing up the data model with its serialization: `InstructionList` looked like it could legitimately be converted into a `List<GPXEntry>` with mandatory timestamps, when in fact it couldn't.

So I went completely down the rabbit hole and mostly untangled it. The GPX generation is now with the other serializers, and all the tests that were for some reason testing against JSON instead of the data classes now don't.


michaz:
Note that this still doesn't do what the OP of #1490 wants (travel times at pillar nodes). It just makes it clear that the stuff about `GPXEntry` is for creating GPX-XML and nothing more -- it doesn't contain any information that `InstructionList` doesn't contain already.

karussell:
This is much cleaner. Finally you moved this xml out of the core and we have it like with the JSON (now crying for more serializers ;)).

@devemux86 do people use GPX in Android? And would this still work using jackson serializer? Or is this the same problem like in #1550?

devemux86:
> @devemux86 do people use GPX in Android?

Very much or you mean just the GPX export in GraphHopper core?

> And would this still work using jackson serializer? Or is this the same problem like in #1550?

The pull request fortunately does not introduce new dependencies in GraphHopper core or you mean something else?

karussell:
>  The pull request fortunately does not introduce new dependencies in GraphHopper core

It moves instructionList.createGPXList and createJson to the InstructionListSerializer

devemux86:
> It moves instructionList.createGPXList and createJson to the InstructionListSerializer

In Android most need to import ready GPX files, not create them from routers, so don't know how much use actually had those methods. Anyway they don't need extra dependencies, just create a plain text file so it's easily doable.

karussell:
Thanks for the clarification!

Issue #1520, opened 2019-01-15
Reduce TraversalModes to edge- and node-based.


easbar:
See discussion in #1247.

easbar:
What about `TraversalMode.EDGE_BASED_1DIR` ? Does it have any valid use-case ? It does edge-based routing, but does not differentiate between the direction an edge (identified by its ID) is encountered (so it only uses one 'state' for each edge (instead of two), thats the 1DIR part of the name). It can model turn restrictions and costs, but for example it cannot find a path with a p-turn.  The javadocs say that it is 
```
Strictly not recommended as it could lead to 'route not found' for bidirectional algorithms.
```
I think if there is a required p-turn it can also lead to route not found for standard Dijkstra ? On the other hand according to the javadocs it is 
```
As fast as node-based
```

Would be interesting to know if anybody is actively using `TraversalMode.EDGE_BASED_1DIR` and for which purpose ?

karussell:
> On the other hand according to the javadocs it is

Yes, 1DIR is really nice but only theoretically. I think in practise the only good solution for edge-based algorithm is bidirectional landmarks with 2DIR. IMO we can remove two of the four modes.

easbar:
Ok, `TraversalMode.EDGE_BASED_1DIR` was removed with #1606. Next one on the list will be `TraversalMode.EDGE_BASED_2DIR_UTURN` and then we can eventually rename `EDGE_BASED_2DIR` to `EDGE_BASED`. I think removing the u-turn mode is mostly about removing/replacing `TraversalMode#hasUTurnSupport`. Instead of checking something like

```
if (traversalMode.hasUTurnSupport() && edgeA==edgeB)
```
or
```
if (traversalMode.hasUTurnSupport() && turnCostExtension.isUTurn(edgeA, edgeB)
```
we should just 'calculate' the turn costs and by default return infinity if `edgeA==edgeB` or `turnCostExtension.isUTurn(edgeA, edgeB)`. @karussell do you remember why this was not done like this from the beginning ? I never understood why instead these explicit u-turn checks were introduced. Maybe I am missing something here ? If one wants to enable/allow u-turns (possibly only under certain conditions, like smaller roads etc.) it would still be possible to do this in the `calcTurnWeight` method (and this would probably be even more flexible than explicit u-turn checks within the algorithms).

[...]

PR #1596, opened 2019-04-10
Fixes edge-based CH problem with u-turns at virtual edges, fixes #1593

### LINKED ISSUE #1593: Edge-based CH does not handle u-turns with virtual edges correctly
This test fails with an exception: abf35b0a7443f2f131c912abd98776d370ee6881

```java
    @Test
    public void test_Issue1592() {
        //      6   5
        //   1<-x-4-x-3
        //  ||    |
        //  |x7   x8
        //  ||   /
        //   2---
        NodeAccess na = graph.getNodeAccess();
        na.setNode(0, 49.407117, 9.701306);
        na.setNode(1, 49.406914, 9.703393);
        na.setNode(2, 49.404004, 9.709110);
        na.setNode(3, 49.400160, 9.708787);
        na.setNode(4, 49.400883, 9.706347);
        EdgeIteratorState edge0 = graph.edge(4, 3, 194.063000, true);
        EdgeIteratorState edge1 = graph.edge(1, 2, 525.106000, true);
        EdgeIteratorState edge2 = graph.edge(1, 2, 525.106000, true);
        EdgeIteratorState edge3 = graph.edge(4, 1, 703.778000, false);
        EdgeIteratorState edge4 = graph.edge(2, 4, 400.509000, true);
        // cannot go 4-2-1 and 1-2-4 (at least when using edge1, there is still edge2!)
        addRestriction(edge4, edge1, 2);
        addRestriction(edge1, edge4, 2);
        // cannot go 3-4-1
        addRestriction(edge0, edge3, 4);
        graph.freeze();
        LocationIndexTree index = new LocationIndexTree(graph, new RAMDirectory());
        index.prepareIndex();
        List<GHPoint> points = Arrays.asList(
                // 8 (on edge4)
                new GHPoint(49.401669187194116, 9.706821649608745),
                // 5 (on edge0)
                new GHPoint(49.40056349818417, 9.70767186472369),
                // 7 (on edge2)
                new GHPoint(49.406580835146556, 9.704665738628218),
                // 6 (on edge3)
                new GHPoint(49.40107534698834, 9.702248694088528)
        );


        List<QueryResult> queryResults = new ArrayList<>(points.size());
        for (GHPoint point : points) {
            queryResults.add(index.findClosest(point.getLat(), point.getLon(), EdgeFilter.ALL_EDGES));
        }

        RoutingAlgorithmFactory pch = automaticPrepareCH();
        QueryGraph queryGraph = new QueryGraph(chGraph);
        queryGraph.lookup(queryResults);
        RoutingAlgorithm chAlgo = pch.createAlgo(queryGraph, AlgorithmOptions.start()
                .traversalMode(TraversalMode.EDGE_BASED_2DIR)
                .build());
        Path path = chAlgo.calcPath(5, 6);
        // there should not be a path from 5 to 6, because first we cannot go directly 5-4-6, so we need to go left
        // to 8. then at 2 we cannot go on edge 1 because of another turn restriction, but we can go on edge 2 so we
        // travel via the virtual node 7 to node 1. From there we cannot go to 6 because of the one-way so we go back
        // to node 2 (no u-turn because of the duplicate edge) on edge1. And this is were the journey ends: we cannot
        // go to 8 because of the turn restriction from edge1 to edge4 -> there should not be a path!
        assertFalse("there should not be a path, but found: " + path.calcNodes(), path.isFound());
    }
```

easbar:
Fixes #1593.

The problem is this:
image

Let's say there are edges like this: 0-1-2 with edge-ids 0:0-1 and 1:1-2 and a shortcut gets introduced from 0 to 2 (skipping 1) (with id 2). Then there is a virtual node (id 3) between 1 and 2 like this: 1-(3)-2 which will add some virtual edges 3:1-3 and 4:3-2. When we detect u-turns we compare the edge ids. So if we travel via the shortcut from 0 to 2 we arrive at 2 with the shortcut id 2, but to calculate the turn costs we use the 'origEdgeLast' (here this is the edge 1-2 with id 1). When we check if going to the virtual node 3 is a u-turn we used to check 1==4 which tells us it is no u-turn (when in reality it is). So instead of the virtual edge id 4 we have to use the original edge id 1 for the comparison.

An important case is also checking for a u-turn at node 3 when coming from node 1 and going to node 2. In this case we must not use the original edge id (1) for both virtual edges, but instead compare 3==4.

In this PR I have:

- cleaned-up and extended `RandomCHRoutingTest` for edge-based CH. In the past there were tests on random graphs, but they did not include `QueryGraph` so such issues with virtual edges did not come up (similar to #1574). 

- fixed the issue by adding a `isUTurn()` method to TurnCostExtension

Ultimately I think the u-turn detection needs some more sophisticated refactoring and we should get rid of the `EDGE_BASED_2DIR_UTURN` traversal mode (#1520), but for now I think it is most important to fix the bugs we already know.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/TurnCostExtension.java
+++ b/core/src/main/java/com/graphhopper/storage/TurnCostExtension.java
@@ -197,6 +197,10 @@ public class TurnCostExtension implements GraphExtension {
         return nextCostFlags(edgeFrom, nodeVia, edgeTo);
     }
 
+    public boolean isUTurn(int edgeFrom, int edgeTo) {
+        return edgeFrom == edgeTo;
+    }
+
     private long nextCostFlags(int edgeFrom, int nodeVia, int edgeTo) {
         int turnCostIndex = nodeAccess.getAdditionalNodeField(nodeVia);
         int i = 0;
--- a/core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
+++ b/core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
@@ -93,6 +93,7 @@ public abstract class AbstractBidirectionEdgeCHNoSOD extends AbstractBidirAlgo {
             }
         }
 
+        // todo: it would be sufficient (and maybe more efficient) to use an original edge explorer here ?
         EdgeIterator iter = reverse ?
                 innerInExplorer.setBaseNode(edgeState.getAdjNode()) :
                 innerOutExplorer.setBaseNode(edgeState.getAdjNode());
@@ -102,7 +103,11 @@ public abstract class AbstractBidirectionEdgeCHNoSOD extends AbstractBidirAlgo {
         while (iter.next()) {
             final int edgeId = getOrigEdgeId(iter, !reverse);
             final int prevOrNextOrigEdgeId = getOrigEdgeId(edgeState, reverse);
-            if (!traversalMode.hasUTurnSupport() && edgeId == prevOrNextOrigEdgeId) {
+            boolean isUTurn = edgeId == prevOrNextOrigEdgeId;
+            if (graph.getExtension() instanceof QueryGraph.QueryGraphTurnExt) {
+                isUTurn = ((QueryGraph.QueryGraphTurnExt) graph.getExtension()).isUTurn(edgeId, prevOrNextOrigEdgeId);
+            }
+            if (!traversalMode.hasUTurnSupport() && isUTurn) {
                 continue;
             }
             int key = GHUtility.getEdgeKey(graph, edgeId, iter.getBaseNode(), !reverse);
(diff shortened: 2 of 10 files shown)


karussell:
Wow the necessary detailed tests are many. Thanks for accepting this challenge :)

BTW: We could need a cleaner solution to this virtual edges problem that bites us again and again :/ ... but any solution that comes to my mind (that does not use virtual edges) becomes tricky with all the special cases like many virtual nodes on one original edge etc and then we have this in every algorithm instead of the QueryGraph.

core/src/main/java/com/graphhopper/storage/TurnCostExtension.java
```diff
@@ -197,6 +197,10 @@ public long getTurnCostFlags(int edgeFrom, int nodeVia, int edgeTo) {
         return nextCostFlags(edgeFrom, nodeVia, edgeTo);
     }
 
+    public boolean isUTurn(int edgeFrom, int edgeTo) {
+        return edgeFrom == edgeTo;
+    }
```
karussell: Wouldn't it make more sense in the TurnWeighting as you wrote in #1593?

easbar:
> Wow the necessary detailed tests are many.

I would not say they are all strictly necessary and maybe they could even be simplified / made a bit more to the point. During development I had a few more and already deleted some again. The most important is the random one, because you can just have it running for a while and it spits out some failing cases, which was really helpful to arrive at a solution (now I think it is green).

> We could need a cleaner solution to this virtual edges problem that bites us again and again :/ ... 

I think the `QueryGraph` approach is a good one and the reason #1593 and #1574 came up was mostly that we had no tests that included virtual via nodes before. Taking this into account I think we did not get bitten so often after all :)

core/src/main/java/com/graphhopper/storage/TurnCostExtension.java
easbar: Yes I was tempted to do it, especially since `AbstractBidirectionEdgeCHNoSOD` already has a reference to a `TurnWeighting`. Then again to make it really consistent we should also use `isUTurn` in `AbstractRoutingAlgorithm#accept` (but I haven't), but this one only has a reference to `Weighting` (while it could obtain a `TurnCostExtension` reference from `Graph.getExtension`).  In any case `TurnCostExtension` needs this method, so we can override it in `QueryGraphTurnExt`. The routing algorithms could call it via something like `TurnWeighting#isUTurn`. I don't know. Either way is fine for me. Maybe postpone this decision until doing a proper u-turn clean up in #1520 ?

karussell: Thanks for the explanation - makes sense

> Maybe postpone this decision until doing a proper u-turn clean up in #1520 ?

:+1:

PR #1600, opened 2019-04-18
Android: VTM 0.11.1 OpenGL vector map library

devemux86:
- [VTM 0.11.1](https://github.com/mapsforge/vtm) OpenGL vector map library
- Update Android build system


karussell:
Thanks!

PR #1603, opened 2019-04-27
Extracts new ShortcutUnpacker from Path4CH.

easbar:
Partly in preparation to include turn cost times with edge-based CH (#1585) and easier testing shortcut unpacking is moved from Path4CH to a separate class.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/ShortcutUnpacker.java
+++ b/core/src/main/java/com/graphhopper/storage/ShortcutUnpacker.java
@@ -1,0 +1,97 @@
+package com.graphhopper.storage;
+
+import com.graphhopper.routing.ch.PrepareContractionHierarchies;
+import com.graphhopper.util.CHEdgeIteratorState;
+import com.graphhopper.util.EdgeIterator;
+import com.graphhopper.util.EdgeIteratorState;
+
+import java.util.Locale;
+
+/**
+ * Recursively unpack shortcuts.
+ * <p>
+ *
+ * @author Peter Karich
+ * @author easbar
+ * @see PrepareContractionHierarchies
+ */
+public class ShortcutUnpacker {
+    private final Graph graph;
+    private final Visitor visitor;
+    private boolean reverseOrder;
+
+    public ShortcutUnpacker(Graph graph, Visitor visitor) {
+        this.graph = graph;
+        this.visitor = visitor;
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/graphhopper/storage/ShortcutUnpacker.java
```diff
@@ -0,0 +1,97 @@
...
+            CHEdgeIteratorState sk1 = getEdge(skippedEdge1, from);
+            CHEdgeIteratorState sk2 = getEdge(skippedEdge2, from);
+            if (sk1.getAdjNode() == sk1.getBaseNode() || sk2.getAdjNode() == sk2.getBaseNode()) {
+                // this is a loop where both skipped edges are loops. but this should never happen.
+                throw new IllegalStateException(String.format(Locale.ROOT,
+                        "error: detected edge where both skipped edges are loops. from: %d, to: %d, " +
+                                "skip-edge1: %d, skip-edge2: %d, reverse: %b", from, to, skippedEdge1, skippedEdge2, reverse));
+            }
+
+            if (reverseOrder == reverse) {
```
easbar: I changed the condition here, to handle cases where a loop shortcut is read backwards. This was a hidden bug that did not show up previously, because of the way paths were unpacked, but is now covered in `ShortcutUnpackerTest`.

core/src/main/java/com/graphhopper/storage/ShortcutUnpacker.java
```diff
@@ -0,0 +1,97 @@
...
+                expandEdge(sk1, !reverseOrder);
+                expandEdge(sk2, reverseOrder);
+            } else {
+                expandEdge(sk2, !reverseOrder);
+                expandEdge(sk1, reverseOrder);
+            }
+        } else {
+            // get properties like speed of the edge in the correct direction
+            if (reverseOrder != reverse) {
```
easbar: Here I changed the condition to make the code more intuitive: If reverseOrder=false and reverse=false from/to are not swapped.

easbar:
Looks like travis has a problem downloading jdk11:
```
curl: (51) SSL: no alternative certificate subject name matches target host name 'download.java.net'
```

karussell:
Thanks!

> Looks like travis has a problem downloading jdk11:

Yes, will see how fast they fix the cert issue and apply the workaround if not fast enough :) https://github.com/sormuras/bach/issues/51#issuecomment-487380743

karussell:
Thanks for this refactoring and the bug fix + tests!

Issue #1388, opened 2018-06-01
When starting web service many warnings occur with JDK 10


karussell:
See log https://travis-ci.org/graphhopper/graphhopper/jobs/386197200

```
WARN  [2018-05-31 14:16:05,672] org.glassfish.jersey.internal.Errors: The following warnings have been detected: WARNING: HK2 service reification failed for [org.glassfish.jersey.message.internal.DataSourceProvider] with an exception:
MultiException stack 1 of 2
java.lang.NoClassDefFoundError: javax/activation/DataSource
	at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
	at java.base/java.lang.Class.privateGetDeclaredConstructors(Class.java:3090)
	at java.base/java.lang.Class.getDeclaredConstructors(Class.java:2316)
	at org.jvnet.hk2.internal.Utilities$3.run(Utilities.java:1308)
	at org.jvnet.hk2.internal.Utilities$3.run(Utilities.java:1304)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
...
	at org.eclipse.jetty.server.handler.AbstractHandler.doStart(AbstractHandler.java:113)
	at org.eclipse.jetty.server.Server.doStart(Server.java:386)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
	at io.dropwizard.cli.ServerCommand.run(ServerCommand.java:53)
	at io.dropwizard.cli.EnvironmentCommand.run(EnvironmentCommand.java:44)
	at io.dropwizard.cli.ConfiguredCommand.run(ConfiguredCommand.java:87)
	at io.dropwizard.testing.DropwizardTestSupport.startIfRequired(DropwizardTestSupport.java:237)
	at io.dropwizard.testing.DropwizardTestSupport.before(DropwizardTestSupport.java:151)
	at io.dropwizard.testing.junit.DropwizardAppRule.before(DropwizardAppRule.java:159)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:46)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:367)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:274)
[...]

PR #1606, opened 2019-05-03
Fixes missing best path update for bidir algos when there is a loop at the meeting point (#1592)

### LINKED ISSUE #1592: Edge-based bidirectional algorithms do not agree with Dijkstra when there is a loop and turn restrictions.
Here is a graph involving a loop and some turn restrictions for which the edge-based bidirectional algorithms do not agree with Dijkstra / do not find the optimal path:

https://github.com/graphhopper/graphhopper/blob/fe8df1d7e56ac6052571eaf2bfc675ba1cd6415e/core/src/test/java/com/graphhopper/routing/EdgeBasedRoutingAlgorithmTest.java#L162-L185

easbar:
- Fixes #1592.
- removes no longer needed `TraversalMode.EDGE_BASED_1DIR` (see #1520)


easbar:
Hmm not sure what exactly failed the build for JDK11, but there are several warnings like this:
```
WARNING: HK2 service reification failed for [org.glassfish.jersey.message.internal.DataSourceProvider] with an exception:
MultiException stack 1 of 2
java.lang.NoClassDefFoundError: javax/activation/DataSource
``` 
?

karussell:
@michaz could it be that this commit https://github.com/graphhopper/graphhopper/commit/e8b5db919c06b0c74c9d03b27e1b1a3229c3464e#diff-4db58d2546fb73b749d75a3df93f1c10 now requires this javax dependency in the reader-gtfs module too like in #1388? Or are some of the dependencies supposed to be `test` scope?

michaz:
I liberally use dropwizard from `reader-gtfs`, yes.

michaz:
Ah, I thought we had found out that Dropwizard doesn't support newer Javas yet and would wait for them?

michaz:
But if that `activation` block is the only thing that's missing, then sure, let's put it in. Totally forgot about that.

easbar:
Sorry, which `activation` block is missing ?

michaz:
What Peter was refering to was adding

```
<dependency>
            <groupId>javax.activation</groupId>
            <artifactId>activation</artifactId>
            <version>1.1.1</version>
        </dependency>
```

to the `pom` of `reader-gtfs`.

Don't know if that helps, and also don't know why that would fail on your branch but not on `master`.

easbar:
> Don't know if that helps, and also don't know why the lack of that would fail on your branch but not on master.

Ok thanks! Looks like the warnings are gone with this, but no idea why there should be a difference to master.

Now the build passed. However I had to start it twice, the first time AppVeyor failed with this: 
```
[ERROR] Tests run: 7, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 11.249 s <<< FAILURE! - in com.graphhopper.api.GraphHopperGeocodingIT
[ERROR] testExtent(com.graphhopper.api.GraphHopperGeocodingIT)  Time elapsed: 5.013 s  <<< ERROR!
java.lang.RuntimeException: Problem performing geocoding for https://graphhopper.com/api/1/geocode?reverse=false&q=new york&limit=7&locale=en&provider=default&key=78da6e9a-273e-43d1-bdda-8f24e007a1fa: connect timed out
	at com.graphhopper.api.GraphHopperGeocodingIT.testExtent(GraphHopperGeocodingIT.java:38)
Caused by: java.net.SocketTimeoutException: connect timed out
	at com.graphhopper.api.GraphHopperGeocodingIT.testExtent(GraphHopperGeocodingIT.java:38)
```
and also this:
```
[ERROR] testPathDetails(com.graphhopper.api.GraphHopperWebIT)  Time elapsed: 5.012 s  <<< ERROR!
java.lang.RuntimeException: Problem while fetching path [49.6724,11.3494, 49.655,11.418]: connect timed out
	at com.graphhopper.api.GraphHopperWebIT.testPathDetails(GraphHopperWebIT.java:310)
Caused by: java.net.SocketTimeoutException: connect timed out
	at com.graphhopper.api.GraphHopperWebIT.testPathDetails(GraphHopperWebIT.java:310)
```

Looks like it has a 5s time limit ?

core/src/main/java/com/graphhopper/util/GHUtility.java
```diff
@@ -465,7 +465,7 @@ public static EdgeIteratorState getEdge(Graph graph, int base, int adj) {
     public static int createEdgeKey(int nodeA, int nodeB, int edgeId, boolean reverse) {
         edgeId = edgeId << 1;
         if (reverse)
-            return (nodeA > nodeB) ? edgeId : edgeId + 1;
+            return (nodeA >= nodeB) ? edgeId : edgeId + 1;
```
karussell: Ah, this is a nice fix! (is it not related to this issue?)

karussell:
> Don't know if that helps, and also don't know why the lack of that would fail on your branch but not on master.

Uhm, this is indeed strange as the travis config files are the same.

> However I had to start it twice, the first time AppVeyor failed with this

Thanks for letting us know. Will investigate if there is a hidden problem in our API service.

core/src/main/java/com/graphhopper/util/GHUtility.java
easbar: Yes this is one of the two fixes needed here. Imagine a loop edge with `nodeA==nodeB`. If the edge key is not independent of `reverse` the bwd search does not find the `SPTEntry` of the fwd search in `updateBestPath` and vice versa.

PR #1611, opened 2019-05-07
Implements correct time calculation for edge-based CH.

easbar:
This PR adjusts the path unpacking such that turn cost times are included (previously they were ignored) for edge-based CH, see #1585.

### DIFF
--- a/api/src/main/java/com/graphhopper/util/Helper.java
+++ b/api/src/main/java/com/graphhopper/util/Helper.java
@@ -32,7 +32,6 @@ import java.util.*;
 import java.util.Map.Entry;
 
 /**
- *
  * @author Peter Karich
  */
 public class Helper {
@@ -64,11 +63,11 @@ public class Helper {
         return new Locale(param.substring(0, index), param.substring(index + 1));
     }
 
-    public static String toLowerCase(String string){
+    public static String toLowerCase(String string) {
         return string.toLowerCase(Locale.ROOT);
     }
 
-    public static String toUpperCase(String string){
+    public static String toUpperCase(String string) {
         return string.toUpperCase(Locale.ROOT);
     }
 
@@ -463,10 +462,13 @@ public class Helper {
      * Equivalent to java 8 String#join
      */
     public static String join(String delimiter, List<String> strings) {
-        StringJoiner joiner = new StringJoiner(delimiter);
-        for (CharSequence cs : strings) {
-            joiner.add(cs);
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < strings.size(); i++) {
+            if (i > 0) {
+                sb.append(delimiter);
+            }
+            sb.append(strings.get(i));
         }
-        return joiner.toString();
+        return sb.toString();
     }
 }
--- a/api/src/main/java/com/graphhopper/util/Helper.java
+++ b/api/src/main/java/com/graphhopper/util/Helper.java
@@ -458,4 +458,15 @@ public class Helper {
 
         return sb.toString();
     }
+
+    /**
+     * Equivalent to java 8 String#join
+     */
+    public static String join(String delimiter, List<String> strings) {
+        StringJoiner joiner = new StringJoiner(delimiter);
+        for (CharSequence cs : strings) {
+            joiner.add(cs);
+        }
+        return joiner.toString();
+    }
 }
--- a/core/src/main/java/com/graphhopper/routing/PathBidirRef.java
+++ b/core/src/main/java/com/graphhopper/routing/PathBidirRef.java
@@ -17,10 +17,14 @@
  */
 package com.graphhopper.routing;
 
+import com.graphhopper.routing.weighting.TurnWeighting;
 import com.graphhopper.routing.weighting.Weighting;
 import com.graphhopper.storage.Graph;
 import com.graphhopper.storage.SPTEntry;
 import com.graphhopper.util.EdgeIterator;
+import com.graphhopper.util.EdgeIteratorState;
+
+import static com.graphhopper.util.EdgeIterator.NO_EDGE;
 
 /**
  * This class creates a DijkstraPath from two Edge's resulting from a BidirectionalDijkstra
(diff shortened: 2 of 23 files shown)


core/src/main/java/com/graphhopper/routing/PathBidirRef.java
```diff
@@ -69,30 +73,77 @@ public Path extract() {
...
+    }
+
+    /**
+     * Similar to {@link #processEdge(int, int, int)}, but with the situation we encounter when doing a backward
+     * search: nextEdgeId--x<--edgeId--adjNode
+     */
+    protected void processEdgeBwd(int edgeId, int adjNode, int nextEdgeId) {
+        EdgeIteratorState edge = graph.getEdgeIteratorState(edgeId, adjNode);
+        distance += edge.getDistance();
+        // special case for loop edges: since they do not have a meaningful direction we always need to read them
```
easbar: Here I wonder if we should do the loop-handling within calcMillis.

core/src/main/java/com/graphhopper/routing/PathBidirRef.java
```diff
@@ -69,30 +73,77 @@ public Path extract() {
...
+        }
+        addEdge(edgeId);
+    }
+
+    private void processTurn(int inEdge, int viaNode, int outEdge) {
+        if (!EdgeIterator.Edge.isValid(inEdge) || !EdgeIterator.Edge.isValid(outEdge)) {
+            return;
+        }
+        if (weighting instanceof TurnWeighting) {
+            time += ((TurnWeighting) weighting).calcTurnWeight(inEdge, viaNode, outEdge) * 1000;
```
easbar: Casting here is ugly, I wonder if `calcTurnWeight` should be in the `Weighting` interface.

api/src/main/java/com/graphhopper/util/Helper.java
```diff
@@ -458,4 +458,15 @@ public static final String underScoreToCamelCase(String key) {
 
         return sb.toString();
     }
+
+    /**
+     * Equivalent to java 8 String#join
+     */
+    public static String join(String delimiter, List<String> strings) {
+        StringJoiner joiner = new StringJoiner(delimiter);
```
devemux86: That would cause problem on Android compatibility, since [StringJoiner](https://developer.android.com/reference/java/util/StringJoiner) being a Java 8 class, was added in API level 24 (Android 7 Nougat).

core/src/main/java/com/graphhopper/storage/ShortcutUnpacker.java
```diff
@@ -31,60 +34,86 @@ public ShortcutUnpacker(Graph graph, Visitor visitor) {
...
+    }
 
-            if (reverseOrder == reverse) {
-                expandEdge(sk1, !reverseOrder);
-                expandEdge(sk2, reverseOrder);
-            } else {
-                expandEdge(sk2, !reverseOrder);
-                expandEdge(sk1, reverseOrder);
-            }
+    private void expandSkippedEdgesNodeBased(int skippedEdge1, int skippedEdge2, int base, int adj, boolean reverse) {
```
easbar: I changed this in the following way: Previously we got the `CHEdgeIteratorState` like this: base<-mid->adj and then passed on the reverse flags such that the base<-mid edge was read 'in reverse'/base->mid. Now I get the states like this: base->mid->adj from the beginning, which I think makes it a lot easier to understand.

core/src/main/java/com/graphhopper/storage/ShortcutUnpacker.java
```diff
@@ -31,60 +34,86 @@ public ShortcutUnpacker(Graph graph, Visitor visitor) {
...
+            expandEdge(sk2, reverse, NO_EDGE);
+        }
+    }
+
+    private int getOppositeEdge(CHEdgeIteratorState edgeState, int adjNode) {
+        assert edgeState.getBaseNode() == adjNode || edgeState.getAdjNode() == adjNode : "adjNode " + adjNode + " must be one of adj/base of edgeState: " + edgeState;
+        // since the first/last orig edge is not stateful (just like skipped1/2) we have to find out which one
+        // is attached to adjNode, similar as we do for skipped1/2.
+        // todo: it is wasteful to create a CHEdgeIteratorState object just to check which edge we have to use!
+        return graph.getEdgeIteratorState(edgeState.getOrigEdgeLast(), adjNode) == null
```
easbar: Maybe add a special method to CHGraph here to prevent creating the edge iterator state object ?

api/src/main/java/com/graphhopper/util/Helper.java
easbar: Oh right! Thanks for pointing this out, will replace.

core/src/main/java/com/graphhopper/routing/PathBidirRef.java
```diff
@@ -69,30 +73,77 @@ public Path extract() {
...
-        int prevEdge = EdgeIterator.Edge.isValid(sptEntry.edge) ? sptEntry.edge : EdgeIterator.NO_EDGE;
-        int tmpEdge = currEdge.edge;
-        while (EdgeIterator.Edge.isValid(tmpEdge)) {
-            currEdge = currEdge.parent;
-            processEdge(tmpEdge, currEdge.adjNode, prevEdge);
-            prevEdge = tmpEdge;
-            tmpEdge = currEdge.edge;
+    }
+
+    private void extractBwdPath() {
```
easbar: I changed the way the edges are traversed here to make sure we are always in the same situation as in the backward search: adj-->base-nextEdgeId. Otherwise the 'incEdge' of the prev/next SPTEntry would not be the one we need to calculate the turn cost.

core/src/main/java/com/graphhopper/storage/ShortcutUnpacker.java
easbar: done, by adding `isAdjacentNode` method to `Graph`.

core/src/main/java/com/graphhopper/routing/PathBidirRef.java
karussell: Probably yes

karussell: btw: instanceof was really expensive, so having it here would be ugly as this is kind of a loop. But in recent JVMs (at least server-side) this shouldn't be a problem

core/src/main/java/com/graphhopper/storage/ShortcutUnpacker.java
```diff
@@ -31,60 +34,85 @@ public ShortcutUnpacker(Graph graph, Visitor visitor) {
...
+            int tmp = skippedEdge1;
+            skippedEdge1 = skippedEdge2;
+            skippedEdge2 = tmp;
+        }
+        CHEdgeIteratorState sk2 = getEdge(skippedEdge2, adj);
+        assert sk2 != null : "skipped edge " + skippedEdge2 + " + is not attached to adjNode " + adj + ". this should " +
+                "never happen because edge-based CH does not use bidirectional shortcuts at the moment";
+        CHEdgeIteratorState sk1 = getEdge(skippedEdge1, sk2.getBaseNode());
+        if (base == adj && (sk1.getAdjNode() == sk1.getBaseNode() || sk2.getAdjNode() == sk2.getBaseNode())) {
+            throw new IllegalStateException(String.format(Locale.ROOT,
```
karussell: Does this never happen because we remove loops in the import? Or this is something else as we still have loop support elsewhere like in getShortcutUnpacker?

easbar: I think a skipped edge that is a loop is impossible unless there is a bug in CH preparation. I think this check has been there more or less since the very beginning and I think I never saw this error :)

easbar: I did not see this error before and after we excluded loops during import.

easbar: Regarding loop support I think our algorithms can handle them in most cases but what is not possible is to decide in which direction a loop should be read (problematic when there is geometry or different fwd/bwd speeds etc.)

karussell: Ok, thanks for the explanation!

> (problematic when there is geometry or different fwd/bwd speeds etc.)

Should we create a separate issue for this? I think there is a solution (although not without problems) https://github.com/graphhopper/graphhopper/issues/1525#issuecomment-456138104

easbar: Yes! And another one for some turn-cost cleanup stuff, like u-turn in calcMillis and calcTurnCost in Weighting. Will do.

karussell: Thanks, so from my perspective you can merge (have not fully understood everything but is also not required yet ;) )

easbar: Ok thanks for the review! The changes in shortcutunpacker and pathbidirref are really not easy to understand without thinking about it in detail, sketching some examples etc. Was not so easy to get it right to be honest. I might post a little drawing to make it clear.

PR #1612, opened 2019-05-07
Remove unnecessary check if way has at least two nodes

Nakaner:
This check exists at the top of the method already and immediately returns if it the way has less than 2 nodes.


karussell:
Thanks!

PR #1620, opened 2019-05-23
removed experimental UnsafeDataAccess

karussell:
I'm not sure why I sometimes get problems when freshly installing IntelliJ with JDK11. But this is annoying and just caused by the explicit usage of sun.misc.Unsafe in UnsafeDataAccess, which is not really used from anyone. And although one gets a ~15% speed-up I would not recommend to run this in production as it could crash the whole JVM (happened often in the early times when there were bugs in it).

This seems to make similar problems elsewhere too https://github.com/akka/akka/issues/25640

Additionally this DataAccess implementation does not grow gracefully (double memory required) i.e. it is unsuited for importing large areas as it is now.

So I suggest to remove UnsafeDataAccess.

(We still use Unsafe via reflections for cleanMappedByteBuffer but this does not cause compile time problems)


easbar:
What kind of problems do you get ? I often get an error containing `ByteBuffer` (don't remember the exact message) when I switch between different branches. In this case I do `Build->Rebuild Project` and the problem goes away.

karussell:
Indeed the error message was missing :)

> .../graphhopper/core/src/main/java/com/graphhopper/storage/UnsafeDataAccess.java
> Error:(45, 26) java: package sun.misc does not exist
> ...

I tried rebuilding the project without success. I'm using `openjdk 11.0.2 2019-01-15` and via command line it compiles fine (otherwise we would get a problem from travis I hope, as it uses the same JDK).

Also this JDK has the jdk.unsupported module:

image

And yes, I also got these ByteBuffer problems in the past but resolving them was faster ;)

Maybe the problem is that IntelliJ uses the module path and not the classpath per default for JDK11?

karussell:
@easbar as it works now after reinstalling everything: should I close this or merge :) WDYT?

easbar:
hmm not sure tbh. the issue I described does not justify deleting unsafe data access, since I was always able to fix really quickly. I also have not used unsafe data access so far.

karussell:
> the issue I described

I had this Buffer issue when I executed mvn via command line and then tried to run a test in IntelliJ. Rebuilding the module helped there, yes. The other issue still persists and I'm unsure why it happened or how to fix it. But then let's close this until someone else stumbles over this.

easbar:
This is the error message I was talking about: 

```
java.lang.NoSuchMethodError: java.nio.ByteBuffer.position(I)Ljava/nio/ByteBuffer;

	at com.graphhopper.storage.MMapDataAccess.getBytes(MMapDataAccess.java:419)
	at com.graphhopper.storage.StorableProperties.loadExisting(StorableProperties.java:56)
	at com.graphhopper.storage.GraphHopperStorage.loadExisting(GraphHopperStorage.java:248)
	at com.graphhopper.storage.GraphHopperStorageTest.testSave_and_fileFormat(GraphHopperStorageTest.java:83)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitForkedStarter.main(JUnitForkedStarter.java:41)
```

karussell:
The main problem is that we do not recommend and do not use Unsafe for production (and never heard someone is using it), so I really think we should remove Unsafe for 1.0.

Maybe later we add the possibility for a faster but still safe usage via normal Java APIs (memory access api https://openjdk.java.net/jeps/370).

PR #1626, opened 2019-05-29
Rename SimpleIntEncodedValue to UnsignedIntEncodedValue

### LINKED ISSUE #1619: SimpleIntEncodedValue can't simply encode an int
I have a variable which is an `int` and uses the entire range of that type, and I'd like to put it in a `SimpleIntEncodedValue`. So I create one with 32 bit, which is allowed. But then I can't put in an `int` because `SimpleIntEncodedValue` doesn't allow negative values. 

In fact, it wastes the 32nd bit because of that. I think that's why this caught me by surprise.

karussell:
Fixes #1619 via renaming.

Supporting negative values is doable but introduces further complexity and overhead in the EncodedValue classes which I wouldn't want to do. At least not now, as the most use cases what to store per edge should be positive.


karussell:
ups, wrong base branch

PR #1632, opened 2019-06-02
JUnit5 Migration

Anvoker:
What better way to figure out exactly how hard migrating to JUnit5 is than doing the whole migration? That's what I ended up doing.

I haven't had time to thoroughly go through every single test to _guarantee_ everything is working. But all the tests pass, I've tried my best to be careful in rewriting things. The test count ends up slightly different between JUnit4 and JUnit5 for reasons I've not fully accounted for, but I haven't seen any tests being dropped.

I have not fundamentally rewritten any tests that didn't map more or less directly to JUnit5. I've migrated the ones that had a clear mapping.

I'll explain all the pain points, incompatibilities and other relevant things. 

I don't expect this to be merged since it's such a huge PR and I can't ensure prompt availability if it subtly messes something up, but it _is_ fully functioning as far as I can tell, so it's not completely out of the question, I suppose. The purpose was more to get a good idea of the process.

## How

The migration was done by a combination of project-wide string replacing JUnit4 imports and annotations with JUnit5 imports and annotations then manually combing all the files for any needed local edits.

## General Considerations

**It's important to note that JUnit5 and JUnit4 tests are perfectly capable of co-existing within the same project.** 

The golden rule I encountered was to not have any JUnit4 tests interact with JUnit5 tests, e.g. by inheritance. JUnit5 has a Vintage runner and supplies all of the JUnit4 dependencies so the old imports still work. This means that the migration can be done incrementally.

The way it works is that any annotations sourced from JUnit4 imports will run on the Vintage runner automatically and annotations sourced from JUnit5 imports will run on the new runner.

**JUnit5 requires Java 8 at runtime but can test code compiled with Java 7 and runs perfectly fine in language level 7 modules** as long as you don't write language level 7 code. 

In practice, this is only mildly annoying due to lambdas and assertThrows, which I'll mention in detail later. Otherwise, it all seems to work perfectly fine.

### 1. Pom Files

The easiest way to introduce the new JUnit5 dependencies is to replace the JUnit4 dependencies with these ones:

```
<dependency>
    <groupId>org.junit.platform</groupId>
    <artifactId>junit-platform-launcher</artifactId>
    <version>1.4.2</version>
    <scope>test</scope>
</dependency>
<dependency>
    <groupId>org.junit.jupiter</groupId>
    <artifactId>junit-jupiter-engine</artifactId>
    <version>5.4.2</version>
    <scope>test</scope>
</dependency>
<dependency>
    <groupId>org.junit.vintage</groupId>
    <artifactId>junit-vintage-engine</artifactId>
    <version>5.4.2</version>
    <scope>test</scope>
</dependency>
<dependency>
    <groupId>org.junit.jupiter</groupId>
    <artifactId>junit-jupiter-params</artifactId>
    <version>5.4.2</version>
    <scope>test</scope>
</dependency>
```

It's possible to stop right here and the project still works and runs tests properly.

### 2. Before/After

These annotations are pretty simple to migrate from JUnit4 to JUnit5 and present no difficulties:

```
BeforeClass -> BeforeAll
AfterClass  -> AfterAll
Before      -> BeforeEach
After       -> AfterEach

org.junit.BeforeClass -> org.junit.jupiter.api.BeforeAll
org.junit.AfterClass  -> org.junit.jupiter.api.AfterAll
org.junit.Before      -> org.junit.jupiter.api.BeforeEach
org.junit.After       -> org.junit.jupiter.api.AfterEach
```

There's one thing to keep in mind: The new JUnit runner will throw an exception if methods annotated with Before or After are declared static. In every place I've encountered, the static keyword could be safely removed and things continued to work.

### 3. Assert/Assume

JUnit5 has perfectly equivalent asserts and assumptions. The import has changed but the method names have not.

```
org.junit.Assert -> org.junit.jupiter.api.Assumptions
org.junit.Assume -> org.junit.jupiter.api.Assertions
```

There's one hugely annoying issue though. This is probably the biggest time waster in the migration. JUnit4 asserts have a parameter list that looks like ``assert(condition, message)`` whereas JUnit5 has ``assert(message, condition)``. This means you can easily waste an hour or two just switching the parameters around. To my shame, that's exactly what I resorted to. Someone more capable in text munging could write some script to do it automatically in a way that doesn't hit edge cases. Besides this, there aren't any other concerns.

### 4. Theories

Theories are completely unsupported. They should be ran as JUnit4 tests.

JUnit5 has decided to not add them, so the concept is deprecated for all intents and purposes. [This comment](https://github.com/junit-team/junit5/pull/1422#issuecomment-389644868) recommends using property-based testing in a third party runner such as [jqwik](https://jqwik.net/).

My recommendation is to stop writing theory based unit tests. If datapoint combination functionality is required, then use something like [CartesianProductProvider](https://github.com/junit-team/junit5-samples/blob/master/junit5-jupiter-extensions/src/main/java/com/example/cartesian/CartesianProductProvider.java).

### 5. The New assertThrows

Both these forms of exception throwing tested are deprecated in JUnit5:

```java
@Rule
public ExpectedException thrown = ExpectedException.none();
```

```java
@Test(expected = IllegalArgumentException.class)
```

The new way to deal with this is ``assertThrows``. The only problem is that it takes a lambda as one of its parameters and a lot of the modules are still on language level 7 which doesn't have lambdas.

```java
@Test
public void testEncoderWithWrongVersionIsRejected() {
    assertThrows(IllegalArgumentException.class, 
        () -> EncodingManager.create("car|version=0"));
}
```

However, there's an easy workaround. Its only downside is that it's somewhat ugly.

```java
@Test
public void testEncoderWithWrongVersionIsRejected() {
    assertThrows(IllegalArgumentException.class, new Executable() {
        @Override
        public void execute() throws Throwable {
            EncodingManager manager = EncodingManager.create("car|version=0");
        }
    });
}
```

It's perfectly functional and offers a clear way of migrating this feature from JUnit4 to JUnit5.

### 6. Rules

Rules are entirely unsupported in JUnit5. There is some life support for a very limited set of rules using the migration module. The closest equivalent JUnit5 has is Extensions.

Conclusion:
- Stop writing new tests with Rules when possible.
- Keep existing tests on JUnit4.
- Use callbacks or Extensions to get similar functionality in new tests.

### 7. Parameterized

There is no way to properly emulate JUnit4 Parameterized in JUnit5.

JUnit4 achieves test class-wide parameterization by using a static source to get parameters and using them on the constructor of the class.
 
JUnit5 only does per-method parameterization. 

You can approximate some of what JUnit4 does with BeforeAll + JUnit5 ParameterizedTest but the rewrite ends up being pretty extensive. JUnit5 extensions might be able to fill this gap, but I haven't figured out a clear way of doing so yet.

There are [discussions](https://github.com/junit-team/junit5/issues/878) of having this feature in JUnit5. Nothing concrete yet, but it's likely that @ParameterizedTest will be changed to work on classes as well as methods when 5.5 ships.

Instead of trying to approximate per-class parameterization in JUnit5, an option is to rewrite existing tests to use per-method parameterization. This is perfectly possible but somewhat time consuming and the other downside is that it more or less forces the setup to be done for every test method in part. This is not a problem in some places, but might make tests a lot slower in others. There are laziness schemes that can allow the setup to run only once and then be 'cached' for the rest of the tests if that's too much of a problem.

Conclusion:
- Stop using Parameterized and RunsWith when possible. 
- Write new tests with per method parameterization in mind.
- Keep old Parameterized tests on JUnit4 (Vintage runner).

### 8. Ignore

Ignore has been replaced with Disabled.

```
org.junit.Ignore -> org.junit.jupiter.api.Disabled
```

### 9. Summary of the major pain points

- Class-wide parameterization cannot be done in JUnit5.
- RunsWith superseded by ExtendWith.
- Entire Rule system deprecated in favor of Extensions.
- New Assert/Assume methods reverse message and condition parameters.
- Thoeries have been completely deprecated.

## Performance

I did some very basic benchmarking and it seems that JUnit5 is slightly slower.

```
JUnit4
Total time:  53.497 s
Total time:  54.360 s
Total time:  51.140 s

JUnit5, no tests migrated to the new runner
Total time:  59.478 s
Total time:  01:04 min
Total time:  57.331 s

JUnit5, most tests migrated to the new runner
Total time:  01:05 min
Total time:  01:00 min
Total time:  59.978 s
```

mvn install was run once for each category and then mvn test was run three times. The three mvn test total times were noted for each category.

Please note that the testing conditions were far from ideal, with many things that could bias the tests, including variable background CPU load. This is just meant to give a very rough sketch.

It's worth taking into consideration that [JUnit5 has support for running the tests in parallel](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution) that was not enabled as it requires some effort to set up. But that's a future avenue for performance improvement.

## Test Count

Are all the tests there? We didn't lose any along the way, right?

```
feature/junit5-migration --> Tests ignored: 12, passed: 2104
master                   --> Tests ignored: 8, passed: 1914
```

This was startling but upon further investigation the stark difference is largely the result of JUnit5 counting repeated tests and JUnit4 not doing so.

I counted every test repetition: 100 + 10 * 11. That's 210. 2104 - 210 = 1894. That's still a difference of 20 tests. I combed manually through the test runner and couldn't see any problem. I tried to output mvn test to a file with some grepping and diff the JUnit4 and JUnit5 outputs, but I ran into problems with outputs being ordered differently.

The ignored tests are the same when looking through IntelliJ's test runner:

Screenshot_20190602_201345_intellj_junit4_ignored
JUnit4

Screenshot_20190602_201345_intellj_junit5_ignored
JUnit5

So it's most likely just counting them differently.

## Overall Recommendation

JUnit5's attractive features are:
- Per method parameterization that can make some tests look so so much better.
- Better assertion-based way of testing for thrown exceptions.
- Ability to use nested test classes without any fuss (e.g. without RunWith Enclosed).
- The Tag annotation allows an easier and more structured way to filter out lengthy tests.
- Can programmatically generated new test cases using the Dynamic Tests feature.
- Native support for repeated tests via the @RepeatedTest annotation.
- Parallel Execution is still experimental but easier to use than JUnit4's ParallelComputer.

I'm not sure what these mean to the GraphHopper team overall, if any of these are more than merely "nice to have".

Migration is pretty easy overall, it took me basically just an entire day by myself.

Going to JUnit5 is definitely more future proof. But that depends what you're proofing against. JUnit5 will continue to be worked on while JUnit4 will be on bug fix support mostly. The last release for JUnit4 was on 2018-11-25. 

One thing is for sure, continuing to write JUnit4 tests will make ripping the bandaid off at least somewhat harder later. New tests written to use Theories, Parameterized, RunWith or any Rule will make migration increasingly more difficult. 

If JUnit5 ships class-wide parameterization in 5.5 then Parameterized tests have a clear migration path. 

However as the size of the codebase grows, migration will mean touching a higher amount of tests and their code, so more places to mess up when migrating. So a very minor form of technical debt would still accrue regardless.

I would recommend at least switching to a JUnit5 dependency and writing new tests to be run by the JUnit5 runner while letting old tests run on the Vintage runner. The only cost of doing that is slightly slower execution speed, at least without parallelism.

Extensively rewriting old tests that use Parameterized or Rule is very likely not worth it at the moment.

The tests with clear migration paths have very good time investment ratios. I know, since I've basically already migrated all of those.

---

If, when and how to migrate is up to GraphHopper. I hope this is at least informative.


karussell:
Thanks a lot for this effort :)

It is interesting as it comes not only with advantages - like with so many decisions ;)

But it is likely that we need to migrate to junit5, maybe not now but as soon as possible. Will have to think about it. Maybe others can jump in with their opinion about it :)

easbar:
Thanks a lot @Anvoker, I can tell you put in a lot of work here and for sure it helps a lot making a decision about the migration! I will have to look in detail and next time I am working on some tests I can think about what would have been different with JUnit5. Being able to do a soft migration (migrate those test that are easy to migrate) first and do it step by step (maybe whenever one is working on a test) is a plus in my opinion.

karussell:
> Being able to do a soft migration (migrate those test that are easy to migrate) first and do it step by step (maybe whenever one is working on a test) is a plus in my opinion.

:+1: 

> If JUnit5 ships class-wide parameterization in 5.5 then Parameterized tests have a clear migration path.

JUnit 5.5 was released shortly after your work here. But I do not see a mentioning of class-wide parametrization. Is this now included? It would be cool to get rid of the junit4 dependency and then start migrating e.g. new or changed tests.

easbar:
> It would be cool to get rid of the junit4 dependency and then start migrating e.g. new or changed tests.

But would it not be better to do the opposite (adding junit5 and start using it where its useful and then when we feel like its an improvement commit to it and remove junit4) ? just assuming this kind of soft migration is possible, I never tried.

karussell:
What I thought could be a problem with this approach is that we could potentially use more deprecated junit4 features. But yes, re-reading the post I now understand that we would keep the junit4 references until we do no longer need the "vintage runner".

easbar:
Closing in favor of #https://github.com/graphhopper/graphhopper/pull/2324

PR #1638, opened 2019-06-18
do code formatting

karussell:
Cleanup like #1451

There are more files involved as I used "clean up" option, that removes e.g. unnecessary return false & return true statements. I find it good improvements overall.

This does not touch reader-gtfs which might be easier to merge @michaz ?


core/src/test/java/com/graphhopper/routing/util/AbstractFlagEncoderTest.java
```diff
@@ -28,21 +28,21 @@
     @Test
     public void testAcceptsCar() {
         CarFlagEncoder encoder = new CarFlagEncoder(5, 5, 0);
```
vzhikserg: It seems this object is not used and can be deleted

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
```diff
@@ -19,16 +19,13 @@
...
 import com.graphhopper.storage.index.LocationIndex;
 import com.graphhopper.storage.index.QueryResult;
 import com.graphhopper.util.BreadthFirstSearch;
 import com.graphhopper.util.EdgeIteratorState;
 import com.graphhopper.util.PointList;
-import com.graphhopper.util.shapes.BBox;
-import com.graphhopper.util.shapes.Circle;
-import com.graphhopper.util.shapes.GHPoint;
-import com.graphhopper.util.shapes.Shape;
+import com.graphhopper.util.shapes.Polygon;
```
vzhikserg: Can be removed because the next line should cover it as well. But I would try to avoid `*` and add separate import statements to use the same style in the project.

easbar:
next time we should just do this, maybe starting with the whitespace stuff + imports only

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
easbar: intellij defaults to use star imports as soon as there are 5 of the same package. this means sometimes its going back and forth between listing all imports and doing star import. to me this threshold seems a bit low and we could set it to 10 (or infinity even). but yeah not a big issue really...

karussell:
Yes, good idea. The smaller the PR the faster it is mergable without headache :)

PR #1640, opened 2019-06-22
Removes TraversalMode.EDGE_BASED_2DIR_UTURN

easbar:
This PR removes `TraversalMode.EDGE_BASED_2DIR_UTURN` and renames `TraversalMode.EDGE_BASED_2DIR` to `TraversalMode.EDGE_BASED`, see #1520.

So far we were able to set turn costs for each junction via `TurnCostExtension` and wrap a `Weighting` with `TurnWeighting`, which applies turn costs in `calcMillis/calcWeight`. To make sure the turn costs are considered when calculating routes the routing algorithms need to be passed this `TurnWeighting`. For u-turns we can choose a default cost which is defined within the `TurnWeighting` object. So far this default u-turn cost was set to `40s` and to prevent u-turns there were explicit u-turn checks within the algorithms that were or were not applied depending on the `TraversalMode` we used.

I do not see a good reason why there are these explicit u-turn checks and have thus changed the code in this PR such that: 

- default u-turn costs are set to infinity, so no u-turns will be taken unless the cost is reduced
- the algorithms no longer need to perform explicit u-turn checks. if there is a u-turn the turn costs will simply be calculated via the (turn)weighting if they are infinite the turn will not be considered.
- there is only one edge-based traversal mode (the u-turn traversal mode is no longer needed and the same behavior can be achieved by reducing the u-turn cost in `TurnWeighting`).

So for routing with turn costs but without u-turns we previously did this:

```java
TurnWeighting turnWeighting = new TurnWeighting(weighting, turnCostExtension);
Dijkstra dijkstra = new Dijkstra(graph, turnWeighting, TraversalMode.EDGE_BASED_2DIR);
dijkstra.calcPath(5, 6);
```

and now we do this:

```java
TurnWeighting turnWeighting = new TurnWeighting(weighting, turnCostExtension);
Dijkstra dijkstra = new Dijkstra(graph, turnWeighting, TraversalMode.EDGE_BASED);
dijkstra.calcPath(5, 6);
```

For routing with turn costs and including u-turns we previously did this:

```java
TurnWeighting turnWeighting = new TurnWeighting(weighting, turnCostExtension);
// optional: adjust u-turn costs, 40 is the default
turnWeighting.setDefaultUTurnCost(39);
Dijkstra dijkstra = new Dijkstra(graph, turnWeighting, TraversalMode.EDGE_BASED_2DIR_UTURN);
dijkstra.calcPath(5, 6);
```

and now we do this:

```java
TurnWeighting turnWeighting = new TurnWeighting(weighting, turnCostExtension);
// required: set u-turn costs to some finite value
turnWeighting.setDefaultUTurnCost(39);
Dijkstra dijkstra = new Dijkstra(graph, turnWeighting, TraversalMode.EDGE_BASED);
dijkstra.calcPath(5, 6);
```

Note that we now have to choose `TraversalMode.EDGE_BASED` **and** use `TurnWeighting` to enable turn costs and exclude u-turns. If we use `TraversalMode.EDGE_BASED`, with a normal `Weighting` the resulting route might include u-turns. This is the price we pay for removing the explicit u-turn checks I think.

Another issue was that `TurnWeighting#calcTurnWeight` did **not** include the default u-turn costs and returned 0 for u-turns (unless set explicitly). I have changed this so u-turn costs are included now.

The default u-turn costs overwrite any manually set u-turn cost in `TurnCostExtension` and there is no way to apply specific u-turn costs on a per-junction basis. Checking for explicit turn costs in `TurnCostExtension` before applying the default u-turn costs would be expensive. It is probably rare that one needs to set explicit u-turn costs, but if one wants to set default left-turn costs (or similar) this is a problem. To make this more clear I have renamed `defaultUTurnCost` to `uTurnCost`.

I think this PR is a step in the right direction, because 'applying turn costs' is no longer spread over different parts of the code (traversal mode/explicit u-turn checks + turn weighting), but maybe we should go further and remove `TurnWeighting` entirely. I will open another issue/PR for this.


core/src/main/java/com/graphhopper/routing/PathBidirRef.java
```diff
@@ -140,10 +126,6 @@ private void processTurn(int inEdge, int viaNode, int outEdge) {
         }
         if (weighting instanceof TurnWeighting) {
             time += ((TurnWeighting) weighting).calcTurnWeight(inEdge, viaNode, outEdge) * 1000;
-            // bugfix: should be cleaned up with u-turn refactoring #1520
```
easbar: this is no longer needed, because `calcTurnWeight` now includes u-turn costs.

core/src/main/java/com/graphhopper/routing/ch/EdgeBasedPathCH.java
```diff
@@ -30,19 +26,7 @@ protected ShortcutUnpacker getShortcutUnpacker(Graph routingGraph, final Weighti
             @Override
             public void visit(EdgeIteratorState edge, boolean reverse, int prevOrNextEdgeId) {
                 distance += edge.getDistance();
-                // a one-way loop that is not a shortcut cannot possibly be read in the 'right' direction, because
```
easbar: I have changed `calcMillis` such that loops are read in fwd direction only so this is no longer needed here. This change is not strictly related to removing the u-turn mode though.

core/src/main/java/com/graphhopper/routing/util/TraversalMode.java
```diff
@@ -37,25 +37,16 @@
...
+     * The edged-based traversal mode with turn restriction and cost support. 2 times slower than node based.
      */
-    EDGE_BASED_2DIR(true, false),
-    /**
-     * Not recommended as it leads to strange routes that outsmart the turn costs. The most feature
-     * rich edge-based traversal mode with turn restriction and cost support, including u-turns. 4
-     * times slower than node based.
-     */
-    EDGE_BASED_2DIR_UTURN(true, true);
+    EDGE_BASED(true);
```
easbar: Should we keep `EDGE_BASED_2DIR` as alias for `EDGE_BASED` for better compatibility with user code ?

core/src/test/java/com/graphhopper/routing/AlternativeRouteTest.java
```diff
@@ -44,6 +42,8 @@
 public class AlternativeRouteTest {
     private final FlagEncoder carFE = new CarFlagEncoder();
     private final EncodingManager em = EncodingManager.create(carFE);
+    private final TurnCostExtension turnCostExtension = new TurnCostExtension();
```
easbar: This test was using edge-based traversal mode but without `TurnWeighting` I had to adjust this otherwise the routes would contain u-turns. This is somewhat nasty unfortunately.

isochrone/src/main/java/com/graphhopper/isochrone/algorithm/Isochrone.java
```diff
@@ -187,10 +187,11 @@ private void searchInternal(int from) {
             int neighborNode = currEdge.adjNode;
             EdgeIterator iter = explorer.setBaseNode(neighborNode);
             while (iter.next()) {
-                if (!accept(iter, currEdge.edge)) {
+                if (!accept(iter)) {
                     continue;
                 }
                 // minor speed up
+                // todonow: really ?
```
easbar: Is this a u-turn check ?

core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
```diff
@@ -160,9 +154,12 @@ protected int getTraversalId(EdgeIteratorState edge, int origEdgeId, boolean rev
     protected boolean accept(EdgeIteratorState edge, SPTEntry currEdge, boolean reverse) {
         final int incEdge = getIncomingEdge(currEdge);
         final int prevOrNextEdgeId = getOrigEdgeId(edge, !reverse);
-        if (!traversalMode.hasUTurnSupport() && turnCostExtension.isUTurn(incEdge, prevOrNextEdgeId))
+        double turnWeight = reverse
+                ? turnWeighting.calcTurnWeight(prevOrNextEdgeId, edge.getBaseNode(), incEdge)
+                : turnWeighting.calcTurnWeight(incEdge, edge.getBaseNode(), prevOrNextEdgeId);
+        if (turnWeight == Double.POSITIVE_INFINITY) {
```
karussell: This should be Double.isInfinite(turnWeight) instead (?)

core/src/main/java/com/graphhopper/routing/util/TraversalMode.java
karussell: I'm for removing (as it is easy to do on the user code side)

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
```diff
@@ -103,19 +100,19 @@ public long calcMillis(EdgeIteratorState edgeState, boolean reverse, int prevOrN
...
     }
 
     /**
      * This method calculates the turn weight separately.
-     * Be aware that it always returns 0 turn weight unless a turn weight was explicitly set. This means it does
-     * NOT include default u-turn costs and calling methods have to check for u-turns.
-     * todo: this should be cleaned up in #1520.
      */
     public double calcTurnWeight(int edgeFrom, int nodeVia, int edgeTo) {
+        if (edgeFrom == EdgeIterator.NO_EDGE || edgeTo == EdgeIterator.NO_EDGE) {
```
karussell: Do you think the EdgeIterator.isValid is more clear here (and in the calcMillis)?

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
```diff
@@ -103,19 +100,19 @@ public long calcMillis(EdgeIteratorState edgeState, boolean reverse, int prevOrN
...
+            return 0;
+        }
+        if (turnCostExt.isUTurn(edgeFrom, edgeTo)) {
```
karussell: If there is an explicit turn cost stored in the data - then it will be overwritten from this. Should we document this?

I guess this problem was already the case before this PR but the intention was that it should be picked from the data IMO (at least the name setDefaultUTurnCost implies this ;))

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
```diff
@@ -56,16 +56,16 @@ public TurnWeighting(Weighting superWeighting, TurnCostExtension turnCostExt) {
      * Set the default cost for an u-turn in seconds. Default is 40s. Should be that high to avoid
      * 'tricking' other turn costs or restrictions.
      */
-    public TurnWeighting setDefaultUTurnCost(double costInSeconds) {
-        defaultUTurnCost = costInSeconds;
+    public TurnWeighting setUTurnCost(double costInSeconds) {
+        uTurnCost = costInSeconds;
```
karussell: Maybe we remove this setter and move it to a new constructor?

core/src/test/java/com/graphhopper/routing/EdgeBasedRoutingAlgorithmTest.java
```diff
@@ -143,11 +142,15 @@ private void initTurnRestrictions(Graph g) {
...
-        return createWeighting(carEncoder, 40);
+        return createWeighting(carEncoder, Double.POSITIVE_INFINITY);
     }
 
     private Weighting createWeighting(FlagEncoder encoder, double uTurnCosts) {
-        return new TurnWeighting(new FastestWeighting(encoder), tcs).setDefaultUTurnCost(uTurnCosts);
+        TurnWeighting result = new TurnWeighting(new FastestWeighting(encoder), tcs);
+        if (!Double.isInfinite(uTurnCosts)) {
+            result.setUTurnCost(uTurnCosts);
+        }
```
karussell: Can this be moved into the new constructor?

isochrone/src/main/java/com/graphhopper/isochrone/algorithm/Isochrone.java
karussell: Yes, I think so. It was basically copied from Dijkstra where we introduced this a long time ago:

https://github.com/graphhopper/graphhopper/commit/f40ef5abcbfd235717b187709217ae1458d00a4a#diff-fc38211c3f0fdf3ba26000cce27b9ce7

core/src/test/java/com/graphhopper/routing/AlternativeRouteTest.java
karussell: Should we write this in the changelog that using edge-based alternative routes is now only possible with enabled turn costs?

core/src/main/java/com/graphhopper/routing/util/TraversalMode.java
easbar: ok

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
easbar: yes, changed.

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
easbar: ok yes good idea, done.

core/src/test/java/com/graphhopper/routing/EdgeBasedRoutingAlgorithmTest.java
easbar: yes this was unnecessary anyway.

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
easbar: Yes this has always been like this and this is why I changed the name from `defaultUTurnCost` to `uTurnCost`. Maybe we should check if this is really important for performance. It saves us from looking up the turn cost value from turn cost extension when `edgeFrom==edgeTo`, but this is only a fraction of all turns (we still look up the value for all pairs of edges where `edgeFrom!=edgeTo`. If we want to do something like default left-turn costs or similar we would not be able to do it the same way, because in this case it would be required to check, e.g. for a turn restriction on a left turn.

I added a comment about this in the TurnWeighting constructor.

core/src/test/java/com/graphhopper/routing/AlternativeRouteTest.java
easbar: done

easbar:
Thanks for the review!

isochrone/src/main/java/com/graphhopper/isochrone/algorithm/Isochrone.java
easbar: Ok I removed my todo comment. Isochrone uses node-based traversal, so it is not really related to this PR (?).

easbar: Ah no, the same check was applied for other algorithms via `AbstractRoutingAlgorithm#accept`. First I thought this is only harmful because it prevents u-turns regardless of the u-turn costs configured in  `TurnWeighting`, but for node-based traversal this check is ok and saves us calculating some edge-weights etc.

-> I have included this check in `accept` again, but only for node-based traversal.

easbar: here: 5d616b491ebb6e21dbb0a42b8aacc2e54b12160e

karussell: Ah, ok!

Issue #1648, opened 2019-06-27
EdgeElevationInterpolator takes too long for large graphs


karussell:
There is an ugly bug here that makes the interpolation extremely slow. Good news is that it is easy to fix (times for this goes down from 40minutes to under 40 seconds ;)). Still we also need to change more to avoid this in the future.

PR #1649, opened 2019-06-27
Improve speed of elevation interpolation

karussell:
#1648

### DIFF
--- a/core/src/main/java/com/graphhopper/util/XFirstSearch.java
+++ b/core/src/main/java/com/graphhopper/util/XFirstSearch.java
@@ -18,18 +18,18 @@
 package com.graphhopper.util;
 
 import com.graphhopper.coll.GHBitSet;
-import com.graphhopper.coll.GHBitSetImpl;
 
 /**
  * This abstract class defines commonalities for BFS and DFS
- * <p>
  *
  * @author Jan Sölter
  */
 public abstract class XFirstSearch {
-    protected GHBitSet createBitSet() {
-        return new GHBitSetImpl();
-    }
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/graphhopper/util/XFirstSearch.java
```diff
@@ -18,18 +18,18 @@
...
  */
 public abstract class XFirstSearch {
-    protected GHBitSet createBitSet() {
-        return new GHBitSetImpl();
-    }
+    /**
+     * Pick the BitSet implementation wisely. Use GHBitSetImpl only if we are sure you visit a large portion of the graph.
+     * And if you choose GHTBitSet the initial capacity can be also important for performance.
+     */
+    protected abstract GHBitSet createBitSet();
```
easbar: Ah this again :) Good idea to make this abstract so one has to pick the bit set explicitly.

core/src/main/java/com/graphhopper/util/XFirstSearch.java
```diff
@@ -18,18 +18,18 @@
...
- * <p>
  *
  * @author Jan Sölter
  */
 public abstract class XFirstSearch {
-    protected GHBitSet createBitSet() {
-        return new GHBitSetImpl();
-    }
+    /**
+     * Pick the BitSet implementation wisely. Use GHBitSetImpl only if we are sure you visit a large portion of the graph.
```
easbar: Maybe use {@link GHBitSetImpl} for easier code navigation and refactoring ?

Issue #1591, opened 2019-04-01
Improve isochrone API response object 


nilsnolde:
I know the isochrone endpoint is fairly new (as FOSS anyways), still wanted to give you some feedback:

- Why not make it a proper `geojson`? Doesn't require much, just replace `polygons` with `features` and add `type=FeatureCollection`
- Please provide relevant information in the `properties` object of each feature. IMO, necessary info is 
  - `center`: right now I have no idea to which network point it snapped to
  - `value`: how are you breaking your buckets into values? E.g. 10 seconds with 3 buckets? 

Nice-to-have:
- request parameters `time_limit`/`distance_limit` should be able to take an arbitrary list of ranges, which makes it much more flexible for end users, rather than an integer with `buckets`

nilsnolde:
I only post this here, because I think the missing `properties` make it rather unusable at the moment.

karussell:
Sounds good. Any time to contribute this :) ?

nilsnolde:
Hehe, thought so;) Not in the immediate future unfortunately, but def when I'm out of open-source hassle in a few weeks.

All of it? You don't care that it'd be a breaking change in terms of response (the `geojson` part)?

karussell:
> You don't care that it'd be a breaking change in terms of response (the geojson part)?

The clients should not break, but maybe there is some trick or we can do this via `type=geojson` instead of `type=json`?

manueltimita:
> maybe there is some trick or we can do this via `type=geojson` instead of `type=json`?

I personally think this is the safest option. I began work on this feature, with a couple of tests already added to `/web/src/test/java/com/graphhopper/http/isochrone/IsochroneResourceTest.java`. Everything is going well so far, but before I proceed to modifying `jsonSuccessResponse()` in `/web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java` I have a question:

For the purpose of reading the response straight into something like [GDAL](https://gdal.org/drivers/vector/geojson.html), it would be very useful that the FeatureCollection is not just a nested property in the response body (the way `polygons` is now). As such, I need to find a place for `info`. Personally, since the GeoJSON spec only mentions the [mandatory members](https://tools.ietf.org/html/rfc7946#section-3.3) and **does not prohibit** [foreign members](https://tools.ietf.org/html/rfc7946#section-6), I'd prefer to just put this in the top level, like this:
```
{
  "type": "FeatureCollection",
  "info": {
    "copyrights": [
[...]

PR #1651, opened 2019-06-28
Response as GeoJSON

manueltimita:
As discussed in https://github.com/graphhopper/graphhopper/issues/1591, I have added the possibility to make a request to the isochrone endpoint specifying the response type as GeoJSON with `type=geojson`. Using `type=json` or no `type` parameter returns the output as usual.


karussell:
Can you update this to the latest master? (there seems to be some conflicts)

manueltimita:
> Can you update this to the latest master? (there seems to be some conflicts)

Fixed

web/src/test/java/com/graphhopper/http/isochrone/IsochroneResourceTest.java
```diff
@@ -117,7 +120,87 @@ public void requestWithShortest() throws Exception {
...
+        assertEquals(lastFeature.path("geometry").path("type").asText(), "Polygon");
+    }    
+    
+    private Response requestIsochrone(String path) {
+    	String url = "http://localhost:8080" + path;
+        return app.client().target(url).request().buildGet().invoke();
+    }
+    
+    private JsonNode parseRequestResponse (Response response) throws IOException {
+    	response.bufferEntity();
```
karussell: Why is this necessary?

web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
```diff
@@ -117,15 +122,31 @@ public Response doGet(
                 JsonFeature feature = new JsonFeature();
                 HashMap<String, Object> properties = new HashMap<>();
                 properties.put("bucket", features.size());
+                if (respType.equalsIgnoreCase("geojson")) {
+                    // If you replace GraphHopper with your own brand name, this is fine.
+                    // Still it would be highly appreciated if you mention us in your about page!
+                    String[] copyrights = {"GraphHopper", "OpenStreetMap contributors"};
+                    properties.put("copyrights", copyrights);
```
karussell: I have slightly updated master to reduce this to one line:

properties.put("copyrights", WebHelper.COPYRIGHTS);

Can you update to master and also remove the comment?

web/src/test/java/com/graphhopper/http/isochrone/IsochroneResourceTest.java
manueltimita: My bad, this is a leftover from when I was calling `readEntity` twice. I will fix and resubmit

web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
manueltimita: Done

karussell:
Nice, thanks!

@manueltimita did you test this for leaflet or some different client?

@nilsnolde would you have time to try this for your use case? Before providing a new format in the public JSON API I would like to make sure it works for at least two different use cases (i.e. clients). I mean GeoJSON is standardized and everything but just to make sure it works...

manueltimita:
> @manueltimita did you test this for leaflet or some different client?

I don't have a **graphopper** installation, but I did check the GeoJSON tests output against the standard and saw no problems. Here's an example of how to check for yourself if you need to: 

Test 1: Run https://github.com/manueltimita/graphhopper/blob/response_as_geojson/web/src/test/java/com/graphhopper/http/isochrone/IsochroneResourceTest.java#L152-L172 and copy/paste the console output (`System.out.println(json);`) into http://geojson.io or http://geojsonlint.com/. They are both using https://github.com/mapbox/mapbox.js/ but if it works there it will definitely work with [Leaflet GeoJSON](https://leafletjs.com/examples/geojson/). Here's the output using Leaflet: https://jsfiddle.net/e2zuqhtp/ 

Test 2 (multiple buckets):
https://github.com/manueltimita/graphhopper/blob/response_as_geojson/web/src/test/java/com/graphhopper/http/isochrone/IsochroneResourceTest.java#L175-L189 and here's the fiddle with the output in Leaflet 1.0.3: https://jsfiddle.net/t1bsLqu8/1/ (in this one, you can click on the buckets)

karussell:
Thanks! I think we should test it with openlayers as a second way, just to make sure. Feel free to do this or I'll try later (not sure yet when though)

manueltimita:
Here is how `requestIsochrone("/isochrone?point=42.531073,1.573792&time_limit=130&type=geojson&buckets=3")` looks like on OL v4: https://jsfiddle.net/6gp1v2cn/

nilsnolde:
Oops, sorry @karussell I just saw this.. Great stuff @manueltimita! Seems like you tested already, output looks good.

I'll try to find time to try with `routing-py`.

karussell:
Thanks a lot @manueltimita - this looks good! Will merge.

@nilsnolde would be cool, if you can test this before 0.13 release (in ~6 weeks).

nilsnolde:
Tested with [routingpy](https://github.com/gis-ops/routing-py), all good with `folium` maps, eats the GeoJSON without any parsing :+1: 

Will release soon on our side.

PR #1654, opened 2019-07-01
Don't show the heightgraph on small screens

boldtrn:
This PR hides the heightgraph on small screens.

In general there are two options for hiding:
- css
- js

The benefit of the css solution would be that on some devices the heightgraph would show up, when you turn the screen. Right now you would have to recalculate the route.

The benefit of the js solution is that we can avoid unnecessary calculations, which might speed up the overall performance, especially on potentially less powerful devices and long routes.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -27,7 +27,7 @@
     "leaflet.vectorgrid": "1.3.0",
     "browserify": "16.2.0",
     "browserify-swap": "0.2.2",
-    "d3": "5.9.1",
+    "d3": "5.9.7",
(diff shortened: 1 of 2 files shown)


karussell:
Is now on GH Maps

easbar:
Just a minor issue, but if I close the height graph (say because I want to see more of the map) and then drag one of the route markers it pops up again.

karussell:
:+1: https://github.com/graphhopper/graphhopper/issues/1658

PR #1656, opened 2019-07-01
Ignore unknown properties of spatial rules

### LINKED ISSUE #1655: Reading Spatial Rule Json fails for unknown property
When reading a spatial rule geojson, the import fails if there are unknown attributes. For example qgis will add a name to the feature collection.

`java.lang.RuntimeException: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "name" (class com.graphhopper.json.geo.JsonFeatureCollection), not marked as ignorable (2 known properties: "type", "features"])`

Maybe we could tell the `ObjectMapper` to ignore unknown properties, like:
```
ObjectMapper mapper = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
``` 

Since the geojson is valid, but GH fails, I added the bug label, I hope that's ok :).

boldtrn:
Fixes #1655

The PR fixes the issue of reading unknown properties, but this changes the ObjectMapper configuration. I don't expect any issues due to that though. WDYT?

### DIFF
--- a/web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
+++ b/web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
@@ -45,6 +45,7 @@ public class GraphHopperManaged implements Managed {
     private final GraphHopper graphHopper;
 
     public GraphHopperManaged(CmdArgs configuration, ObjectMapper objectMapper) {
+        objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
         String splitAreaLocation = configuration.get(Parameters.Landmark.PREPARE + "split_area_location", "");
         JsonFeatureCollection landmarkSplittingFeatureCollection;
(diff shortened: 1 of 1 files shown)


karussell:
Maybe apply this change to landmarkSplittingFeatureCollection too?

```java
objectMapper = objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```

boldtrn:
Sure, I moved the change further up :).

karussell:
I think, I actually misunderstood the ObjectMapper API. This means this change is done also in other places as we modify the ObjectMapper and do not create a new object?

Should we clone the ObjectMapper (just copy the deserialization config)? Or even more decoupled: remove the ObjectMapper parameter from GHManaged and start with an unconfigured ObjectMapper?

boldtrn:
> This means this change is done also in other places as we modify the ObjectMapper and do not create a new object?

Yes exactly :).

There is an option to copy the ObjectMapper and then change the config only locally. Should we test that?

boldtrn:
I updated the PR to copy the ObjectMapper

karussell:
Thanks!

PR #1660, opened 2019-07-10
use staticHashCode to fix problems on Android but could occur elsewhere

### LINKED ISSUE #1650: Cannot load graph on Android due to mismatching version for road_class EncodedValue
See the error report [here](https://discuss.graphhopper.com/t/android-v-0-13-cannot-find-encodedvalue-roundabout-in-collection-null/4483/3):

```
An error happened while creating graph:Version of EncodedValue road_class|version=1237974758|bits=5|index=0|shift=1|store_both_directions=false does not match road_class|version=-2100336611|bits=5|index=0|shift=1|store_both_directions=false. Stored 1237974758 vs. in code -2100336611
```

The likely reason is that we currently use hashcode as version identifier but it is not intended for that (can change in different JVMs or even different runs).

The question is why this worked in the past e.g. in BaseGraph:

edges.setHeader(2 * 4, encodingManager.hashCode());

Reproducable via

 * do import via config-example.yml
 * load with `new GraphHopper().load(ghLocation)` where we create the EncodingManager via `EncodingManager.create(encodedValueFactory, flagEncoderFactory, ghLocation)`

Update: Potential fix available at https://github.com/graphhopper/graphhopper/tree/issue_1650

karussell:
Fixes #1650

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/profiles/UnsignedIntEncodedValue.java
+++ b/core/src/main/java/com/graphhopper/routing/profiles/UnsignedIntEncodedValue.java
@@ -19,6 +19,7 @@ package com.graphhopper.routing.profiles;
 
 import com.graphhopper.storage.IntsRef;
 
+import java.util.Arrays;
 import java.util.Locale;
 import java.util.Objects;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/graphhopper/routing/profiles/UnsignedIntEncodedValue.java
```diff
@@ -171,12 +172,76 @@ public boolean equals(Object o) {
...
+
+    /**
+     * Produces a static hashcode for an Enum arrays that is platform independent and still compatible to the default
+     * of openjdk.
+     */
+    static int staticHashCode(Enum... vals) {
+        if (vals == null)
+            return 0;
+        int len = vals.length;
+        Enum[] locVal = vals;
```
easbar: very minor, but do we really need the extra `locVal` reference or can we not just do `vals[idx]` in the loop below ?

easbar:
Not sure if this is relevant here, but in another project I once ran into a performance problem, because `Objects.hash(Object...)` was creating lots of temporary `Object[]` instances.

easbar:
Apparently this was also a problem [under mac OS](https://discuss.graphhopper.com/t/is-it-possible-to-relaunch-graphhopper-without-deleting-the-gh-folder/4593)

karussell:
> but in another project I once ran into a performance problem, because Objects.hash(Object...) 

Yeah. It is kind of ugly that this is now the default for IntelliJ as hashing should be very fast and lightweight

core/src/main/java/com/graphhopper/routing/profiles/UnsignedIntEncodedValue.java
karussell: I read somewhere (cannot find the link now) that a local variable can have different behaviour (faster). When thinking where this can come for one explanation would make sense if you assign an instance variable to a local variable (moving it from heap to stack). But here it is basically a method parameter (stack) to stack. There must be a reason as this is done e.g. in `Arrays#hashCode`. But obviously the stack allocation has to be larger and also more bytecode lines are produced (checked via `javap -v`) and [some say they are identical](https://coderanch.com/t/556166/java/Parameter-Variable-considered-Local-Variable). Will remove it and read a bit more about the JVM internals.

easbar: > Will remove it and read a bit more about the JVM internals.

Cool looking forward to hear what you find out.

karussell: Haven't found much. Just tested it in JMH and it looks that between arrays as parameter or local variable there is no difference, but as a class variable the code is indeed ~5% slower:

https://gist.github.com/karussell/4cd7999235c651f7e714630b61297c6b

PR #1661, opened 2019-07-12
Adds new fail-fast option to hand-crafted matrix client.

easbar:
With this PR we can disable the 'fail fast' option for matrix requests which means that the matrix calculation is not cancelled as soon as some points cannot be found or some points are not connected. 

Instead, the full matrix is returned and the matrix entries that involve points that were not found and those for which no connection could not be found will be set to `Double.MAX_VALUE` for weight and distance and `Long.MAX_VALUE` for time. Additionally, `MatrixResponse` now has two arrays of the indices of the from/to points that could not be found and another list of index pairs of the points for which no connection could be found.

In code:

```java
GraphHopperMatrixWeb matrixApi = new  GraphHopperMatrixWeb(new GHMatrixSyncRequester());
GHMRequest request = new GHMRequest();
request.addPoint(new GHPoint(lat, lon));
// more points etc.

// this is new!
request.setFailFast(false);

// send matrix request
MatrixResponse matrix = matrixApi.route(request);

// will be false unless something unusual happens, because if failFast is false we do not consider invalid points and points that are not connected as errors
assertFalse(matrix.hasErrors());

// if some connections could not be found or if there are invalid points this will be true
assertTrue(matrix.hasProblems());

// will contain the weight 0->5, or Double.MAX_VALUE if the connection was not found or points 0 or 5 could not be found
matrix.getWeight(0, 5);

// just like weight
matrix.getDistance(0, 5);

// just like distance/weight, except it is a long and Long.MAX_VALUE is returned if there were problems
matrix.getTime(0, 5);

// get the disconnected point pairs
List<PointPair> disconnectedPoints = matrix.getDisconnectedPoints();

// get the points that could not be found
List<Integer> invalidFromPoints = matrix.getInvalidFromPoints();
List<Integer> invalidToPoints = matrix.getInvalidToPoints();
```


client-hc/src/main/java/com/graphhopper/api/GHMRequest.java
```diff
@@ -146,7 +149,17 @@ public boolean hasPointHints() {
                 this.toPointHints.size() == this.toPoints.size() && !toPoints.isEmpty();
     }
 
-    private int called = 0;
```
easbar: was this down here intentionally ? sorry if yes

client-hc/src/main/java/com/graphhopper/api/GHMatrixAbstractRequester.java
```diff
@@ -146,24 +153,36 @@ protected void fillResponseFromJson(MatrixResponse matrixResponse, JsonNode solu
             }
 
             JsonNode distancesFromArray = null;
-            int[] distances = null;
+            double[] distances = null;
```
easbar: this was cast to int for a moment and later its a double again so I changed this.

client-hc/src/main/java/com/graphhopper/api/MatrixResponse.java
```diff
@@ -13,8 +13,12 @@
 
     private String debugInfo = "";
     private final List<Throwable> errors = new ArrayList<>(4);
+    private final List<PointPair> disconnectedPoints = new ArrayList<>(0);
+    // todonow: should we add hppc to classpath and use IntArrayList instead ?
```
easbar: wasn't sure here

easbar:
`GraphHopperMatrixBatchIT#testPointsNotFound_doNotFailFast` still fails but should be green once the fail fast option is supported by the GH Matrix API.

client-hc/src/main/java/com/graphhopper/api/GHMRequest.java
karussell: in rare situations I need to store state but they are used in just one method, but moving this up is fine :)

client-hc/src/main/java/com/graphhopper/api/GHMatrixAbstractRequester.java
karussell: But why would we need double here (8 byte) when we store distances in meter (0 to 10 000 000)?

client-hc/src/main/java/com/graphhopper/api/MatrixResponse.java
karussell: Yeah those cases of small objects in a list are ultra memory waste but we should improve this later if this makes a problem for bigger matrices.

client-hc/src/main/java/com/graphhopper/api/MatrixResponse.java
```diff
@@ -188,6 +230,39 @@ public String toString() {
...
+        }
+        if (!invalidToPoints.isEmpty()) {
+            result += ", invalidToPoints: " + invalidToPoints.size();
+        }
+        return result;
+    }
+
+    public static class PointPair {
+        private final int sourceIndex;
+        private final int targetIndex;
```
karussell: (`public final` would be also fine for those structs allowing as to avoid the getter stuff)

client-hc/src/main/java/com/graphhopper/api/GHMatrixAbstractRequester.java
easbar: Yes, as we discussed before it would be best to store distances as ints in units of 0.1m. In this case precision should be sufficient also for indoor stuff and the maximum distance should be enough also for very long shortcut edges. Here I changed this to double because the numbers stored were retrieved as floating point numbers (from json) and also `MatrixResponse#getDistance()` returns a double. Whatever you like better I can change this back to int as well. In the long run we should switch all distances to ints in my opinion.

client-hc/src/main/java/com/graphhopper/api/MatrixResponse.java
easbar: ok changed it.

client-hc/src/main/java/com/graphhopper/api/MatrixResponse.java
easbar: we could even return an int[] ?

client-hc/src/main/java/com/graphhopper/api/GHMatrixAbstractRequester.java
karussell: Yeah, I would keep this as int[] here. 

IMO we should differentiate between storage and API. The Java API should stay as `double` and only the storage should be int (or less). For edge.getDistance it could be valuable to investigate if we can include it in IntsRef. (btw: there is an interesting difference to e.g. max_speed or road_class then: the distance needs to be splitted for the virtual edges in the QueryGraph. And that is also the reason I did not try this yet ;) )

easbar: Ok I changed back to int[]. Also opened #1663.

client-hc/src/main/java/com/graphhopper/api/MatrixResponse.java
```diff
@@ -128,7 +128,7 @@ public double getDistance(int from, int to) {
         } else if (to >= distances[from].length) {
             throw new IllegalStateException("Cannot get 'to' " + to + " from distances with size " + distances[from].length);
         }
-        return distances[from][to];
+        return distances[from][to] == Integer.MAX_VALUE ? Double.MAX_VALUE : distances[from][to];
```
karussell: oh, thx - this is a good catch :+1:

karussell: Shouldn't it better map to `infinity` instead of Double.MAX_VALUE?

karussell: Or the other way around: disconnected == i.e. "infinity" gets serialized into `null` and should be back in Java world infinity (?)

easbar: Yes previously this case never occurred because there would have been an error if points were not connected. Now we have this case if failFast==false, yes. Hm not sure about infinity vs. max. For time we also return `Long.MAX_VALUE`.

easbar: The 'impossible' weight is `Double.MAX_VALUE` on the server side as well (so far). The same for `Path#weight`:

https://github.com/graphhopper/graphhopper/blob/c43d03e9cbc2690b1c3bef319921f6f15adc5d82/core/src/main/java/com/graphhopper/routing/Path.java#L72-L74

karussell: Ah, yes. For `long` there is not infinity. Ugly, as we currently use "infinity" for "disconnected" in the Weighting and so I thought we could make it more consistent (and maybe refactor `Path#weight` too) but this would make it inconsistent with `long`.

easbar: Do you prefer one over the other, I do not have a strong opinion about this and either way is fine for me.

karussell: > because normally when hasProblems()==true we can use getDisconnectedPoints() to check which connections are missing (instead of checking getWeight(3, 5) == Double.MAX_VALUE`

This seems like the best solution to force or promote the usage of disconnected points in this cases (or have a convenience method like `isConnected(from, to)`)

easbar: Yes having `isConnected(from, to)` is a very good idea, if we have this there should not really be a reason to check something like `getWeight(3,5)==Double.MAX_VALUE` and we only do this internally making the concrete 'impossible value' even 'less important'. Will do.

karussell: Nice, thanks!

easbar:
What kind of response should we expect if `failFast=false` and *none* of the points are found ? Is it ok to simply return a matrix full of max values or should this be treated as an error (with a message field etc. ?). So far this case is not handled properly on the server side as I just noticed.

karussell:
> What kind of response should we expect if failFast=false and none of the points are found ?

Good question. On the one hand I would not expect something special but if everything fails it would be also fine (and I have a tiny preference here) if this counts as an error and we do not have to return a large matrix where all indices are included in the "failed" array. But maybe then this should be a special message and explicitly documented?

easbar:
> we do not have to return a large matrix where all indices are included in the "failed" array.

yes, but probably this case is rather rare ?

> But maybe then this should be a special message and explicitly documented?

What is a bit ugly I think is that one has to handle this special case on the client side and there is not much information that can be used to recognize this case other than the actual error message (unless we use a dedicated http status, but this also makes it more complicated). Also the reason why the connections were not found is not always clear. Something like
```
if (json.has("message") && json.get("message").contains("None of the requested connections were found") {
    // no connections found, but was it because all sources were invalid ? or all targets ? or were there just no routes found between the (otherwise valid) sources/targets ?
}
```
So we should at least keep the lists of invalid points on the response. But checking the error message is still ugly ? Maybe it is more clear  to just blindly return the matrix full of nulls and return the full list of invalid points and disconnected points (even when its the whole matrix) as usual ?

easbar:
Looks like the backend is supporting this now, the tests are green :)

PR #1662, opened 2019-07-12
Adds logging to exception mappers.

easbar:
So far for some requests the GH server does not log anything, for example if there is an `IllegalArgumentException`. Here I have added some logging to the `ExceptionMapper`s.

There is also `io.dropwizard.jersey.errors.LoggingExceptionMapper`, maybe it would make more sense to use this, we would have to be careful to make sure the server response does not change. Not sure, I am not very familiar with dropwizard.


web-bundle/src/main/java/com/graphhopper/http/IllegalArgumentExceptionMapper.java
```diff
@@ -1,14 +1,21 @@
...
 import com.graphhopper.MultiException;
+import com.graphhopper.util.Helper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
 
+@Provider
```
easbar: No idea if this is needed, but `MultiExceptionMapper` had it as well ;)

web/src/main/java/com/graphhopper/http/GHJerseyViolationExceptionMapper.java
```diff
@@ -19,17 +19,25 @@
...
+import org.slf4j.LoggerFactory;
 
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
 import java.util.stream.Collectors;
 
+@Provider
 public class GHJerseyViolationExceptionMapper implements ExceptionMapper<JerseyViolationException> {
```
easbar: btw there is also `io.dropwizard.jersey.validation.JerseyViolationExceptionMapper`, but no idea really :)

karussell: Good question - does not look too different except now the logging info vs. debug: https://github.com/dropwizard/dropwizard/blob/bb6209d053995c4b2111a0525ae231902f662edb/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/JerseyViolationExceptionMapper.java

web-bundle/src/main/java/com/graphhopper/http/IllegalArgumentExceptionMapper.java
```diff
@@ -1,14 +1,21 @@
...
 public class IllegalArgumentExceptionMapper implements ExceptionMapper<IllegalArgumentException> {
+    private static final Logger logger = LoggerFactory.getLogger(IllegalArgumentExceptionMapper.class);
 
     @Override
     public Response toResponse(IllegalArgumentException e) {
+        logger.info("bad request: " + (Helper.isEmpty(e.getMessage()) ? "unknown reason" : e.getMessage()));
```
karussell: Do you know a way to attach the request path and query parameters?

karussell:
I mean we could try to exclude all resources that provide a custom logging with this: https://github.com/dropwizard/dropwizard/pull/1435 instead of adding logging to all these exception mappers where a more detailed access log would be better. I'm unsure tbh ...

web-bundle/src/main/java/com/graphhopper/http/IllegalArgumentExceptionMapper.java
easbar: Apparently the full request url is included in the thread name, so when the logging pattern includes `%thread` the url will be included in the logs. For requestLog there is `%r`.

easbar:
> Can you try to change this to e.g. a console logger and see if it already produces some logs for you?

Ah yes, `requestLog` does the trick :) For example this:
```yaml
  requestLog:
        appenders:
          - type: console
            logFormat: "%t{YYYY-MM-dd HH:mm:ss.SSS} \"%r\" %s"
```
will log the request url (r) and the response status (s), more options are documented here: https://logback.qos.ch/manual/layouts.html#logback-access

easbar:
> I mean we could try to exclude all resources that provide a custom logging with this: dropwizard/dropwizard#1435 instead of adding logging to all these exception mappers where a more detailed access log would be better. I'm unsure tbh ...

Ok I think there is no need to add logging to the exception mappers because of `requestLog`. Do you mean it would be nice to enable `requestLog` by default, but disable it whereever we are doing a more detailed manual logging like for successful requests in `RouteResource` ?

karussell:
>  but disable it whereever we are doing a more detailed manual logging like for successful requests in RouteResource ?

Yes, exactly. Currently we disable it to avoid two logs per route request leading to no logs for bad requests.

easbar:
> Currently we disable it to avoid two logs per route request leading to no logs for bad requests.

The request log can be customized using a [logging filter](https://www.dropwizard.io/1.3.13/docs/manual/core.html#logging-filters). But as far as I understand the filter cannot be adjusted without recompiling (it can only be disabled/enabled via the config file). We could make it only log status >= 400, but sometimes we are also manually logging some errors with more details (e.g. in `RouteResource`) so these would be duplicated. Sure we could exclude or remove them as well, but to me it now seems simpler to just add the manual logging to the exception mappers ? Then we can also provide more details in the logs (I did not manage to show the full response json with requestLog) and disable/enable the logging using just the logging configuration.

karussell:
> Sure we could exclude or remove them as well, but to me it now seems simpler to just add the manual logging to the exception mappers ?

Ok. Is it possible somehow to add the original path + query parameters to the log?

easbar:
>  Is it possible somehow to add the original path + query parameters to the log?

You mean when we use manual log statements ? Yes, this can be configured in `config.yml`: If the log pattern contains `%thread` (which it already does) the url including url parameters appear in the logs (apparently they are part of the thread name.

karussell:
I meant for the new log entries like:

```java
logger.info("bad request: " + (Helper.isEmpty(e.getMessage()) ? "unknown reason" : e.getMessage()));
```

Is it possible to include host and path and query parameters?

karussell:
> Is it possible to include host and path and query parameters?

Or is this too complicated or not necessary?

easbar:
> Or is this too complicated or not necessary?

Not sure, like I said this can be achieved via the logging configuration or are you missing something that is not possible like this ?

karussell:
Hmmh, I did not understand this. With this PR the bad requests would be logged, but without any URL. I do not see how this could be configured using the config.yml.

If there is e.g. a ConnectionNotFoundException we would know that it occurred but not for which URL.

easbar:
> With this PR the bad requests would be logged, but without any URL. I do not see how this could be configured using the config.yml.

When I add `[%thread]` to the pattern in `logFormat` (as we currently do in `config-example.yml` already: https://github.com/graphhopper/graphhopper/blob/d093043e33921b0e80aba3651fab15956d964e70/config-example.yml#L173

The logging output looks like this for example:
```shell
2019-09-17 11:12:46.847 [dw-37 - GET /route?point=52.597626%2C13.311138&point=52.552559%2C13.358688&type=json&locale=en-US&vehicle=car&weighting=fastest&elevation=false&key=] INFO  c.g.h.IllegalArgumentExceptionMapper - bad request: some error whatever
```

so `%thread` shows the URL and the query parameters. The same can be done for request log using `%r`. Does this not work ?

Apart from this I think we need to make a decision here whether or not we shall add manual logging for error responses or use request log instead.

karussell:
> the same can be done for request log using %r. Does this not work ?

Ah, did not know this, nice!

> Apart from this I think we need to make a decision here whether or not we shall add manual logging for error responses

I think your argument convinces me here:

> Sure we could exclude or remove them as well, but to me it now seems simpler to just add the manual logging to the exception mappers ? Then we can also provide more details in the logs

If this also your opinion we should merge :)

web-bundle/src/main/java/com/graphhopper/http/HttpErrorsOnlyFilterFactory.java
```diff
@@ -0,0 +1,20 @@
+package com.graphhopper.http;
+
+import ch.qos.logback.access.spi.IAccessEvent;
+import ch.qos.logback.core.filter.Filter;
+import ch.qos.logback.core.spi.FilterReply;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import io.dropwizard.logging.filter.FilterFactory;
+
+@JsonTypeName("http-errors-only")
+public class HttpErrorsOnlyFilterFactory implements FilterFactory<IAccessEvent> {
```
karussell: Is this class used or was it just for demo-purposes?

easbar: No this one can be used to filter the request logs (showing only errors).

easbar:
> If this also your opinion we should merge :)

I don't think I have a preference, so lets do the manual logging for now.

Issue #1647, opened 2019-06-27
Allow restricting start/target edge for edge-based routing.


easbar:
Even for edge-based algorithms (supporting turn costs & restrictions) we are calculating routes between *nodes* (not edges) of the routing graph, i.e. `calcPath(fromNodeIndex, toNodeIndex)`, not `calcPath(fromEdgeIndex, toEdgeIndex)`. To do this we simply use all adjacent outgoing edges at the start node as well as all incoming edges at the target node. 

However, what we cannot do at the moment is restricting the route calculation such that only certain edges at the start/target nodes are considered. For example this would be needed to be able to start a route in a certain direction (say start driving in northern instead of southern direction).

The feature that comes closest to this requirement is the `heading` parameter. The way this is implemented we increase the cost of edges that are adjacent to the start/target nodes and do not face in a specified direction. Because of their higher cost it will thus be less likely they are included in the shortest path. 

Unfortunately, the `heading` parameter has several restrictions:

1) the heading penalty can only be applied to virtual edges, i.e. it does not work for tower nodes
2) since the heading penalty is given in time units it is not compatible with all weightings (#485)
3) heading is not supported for speed mode / CH (#483)
4) the heading tolerance is fixed to a hard-coded value at the moment (#1646)

On the other hand enforcing the start/target edge with edge-based routing is a very simple thing, especially for bidirectional algorithms. As I explained, when the route calculation starts we consider all edges adjacent to the start/target node. To enforce a single direction all we need to do is skipping all edges except the one(s) we want at the first step. This is how it can be done (using `DijkstraBidirectionRef`):

```java
public Path calcPathRestricted(int source, int target, int sourceOutEdge, int targetInEdge) {
    AbstractBidirAlgo algo = new DijkstraBidirectionRef(graph, weighting, TraversalMode.EDGE_BASED);
    algo.createAndInitPath();
    algo.initFrom(source, 0);
    algo.initTo(targe, 0);

    // filter out all edges except the one we want for the first step
    algo.setEdgeFilter(new SingleEdgeFilter(sourceOutEdge));
    algo.fillEdgesFrom();

    algo.setEdgeFilter(new SingleEdgeFilter(targetInEdge));
    algo.fillEdgesTo();

    // no filtering from now (unless there should be some other filter used for the whole time of course) 
    algo.setEdgeFilter(null);
    algo.runAlgo();
    return algo.extractPath();
}

private static class SingleEdgeFilter implements EdgeFilter {
[...]

Issue #1557, opened 2019-02-28
Shortest path weight calculated by LM deviates from the one calculated by Dijkstra.


easbar:
Following the discussion in the [forum](https://discuss.graphhopper.com/t/alt-stopping-criterion-incomplete/4129/4), I tried running random queries ALT vs Dijkstra and found cases where there is a difference, see here: 1f34cd6307f01b3f0f04fa458f134dd061453243

However, there do not seem to be deviations between AStarBidirection and Dijkstra, so not sure if the problems are due to the stopping critirion which is the same for ALT and AStarBidirection.

easbar:
I was wondering if the `buildRandomGraph` method has a problem, but modifying the distances should be ok, since they are only increased ? https://github.com/graphhopper/graphhopper/blob/022697da8dfc9ab6e6537fbc7b81efc55ac2e142/core/src/main/java/com/graphhopper/util/GHUtility.java#L189 Also the method uses `Helper.DIST_PLANE` to calculate the distances (should be ok because this is what the WeightApproximators use as well).

karussell:
> so not sure if the problems are due to the stopping critirion which is the same for ALT and AStarBidirection

The suggestion in the forum was that we have to use a different stopping criterion for ALT, so your observation is an indication there is a problem with it for ALT

> but modifying the distances should be ok, since they are only increased ?

Yes, increasing the weight shouldn't be a problem for A*

> Also the method uses Helper.DIST_PLANE to calculate the distances (should be ok because this is what the WeightApproximators use as well).

Yes, should also be okayish for "smaller" distances.

easbar:
> The suggestion in the forum was that we have to use a different stopping criterion for ALT, so your observation is an indication there is a problem with it for ALT

Are the active landmarks fixed per query or can they change during the query (onHeuristicChange) ? This could have implications on the stopping criterion. Then again I used `numLandmarks=numActiveLandmarks=4` in the tests, so there is nothing to change. 

The slides mentioned in the forum do not say there needs to be done anything particular for ALT, even though the stopping criterion is explained in the ALT chapter (but more generally for bidirectional AStar).

karussell:
(renamed title from ALT to LM as the similarity to alternative routes is a bit ugly ;) Please revert if bad)

> Are the active landmarks fixed per query or can they change during the query (onHeuristicChange) ? 

They are fixed.

> The slides mentioned in the forum do not say there needs to be done anything particular for ALT

[...]

PR #1665, opened 2019-07-16
Adds AbstractBidirAlgo#calcPath with restricted source/target edges.

easbar:
For #1647 I have added a new `calcPath` method to `AbstractBidirAlgo` that, additionally to the from/to node ids, also takes two edge ids that are enforced as start/target edges. This allows to strictly enforce the first and last edge of a path.

For now this is only supported on the low-level java API level and how to obtain the edge ids for the edges one wants to restrict is another question (?).


core/src/main/java/com/graphhopper/routing/AbstractBidirAlgo.java
```diff
@@ -88,6 +96,24 @@ protected void initCollections(int size) {
...
+    /**
+     * like {@link #calcPath(int, int)}, but this method also allows to strictly restrict the edge the
+     * path will begin with and the edge it will end with.
+     *
+     * @param fromOutEdge the edge id of the first edge of the path. using {@link EdgeIterator.ANY_EDGE} means
+     *                    not enforcing the first edge of the path
+     * @param toInEdge    the edge id of the last edge of the path. using {@link EdgeIterator.ANY_EDGE} means
+     *                    not enforcing the last edge of the path
+     */
+    public Path calcPath(int from, int to, int fromOutEdge, int toInEdge) {
```
easbar: In principle we could also move this to the `RoutingAlgorithm`, but for now it was easier to implement this only for bidirectional algorithms. Not sure if we really need this for the others as well and we can add this later if needed (?)

core/src/main/java/com/graphhopper/util/EdgeIterator.java
```diff
@@ -45,6 +45,11 @@
      */
     int NO_EDGE = -1;
 
+    /**
+     * integer value used to indicate that any edge is accepted
+     */
+    int ANY_EDGE = Integer.MIN_VALUE;
```
easbar: what is a bit ugly here is that `EdgeIterator.Edge.isValid` only checks if an edge id is positive or not. this does not allow us to have other special values like this one here. maybe we should make `isValid` check if `==-1` ?

core/src/test/java/com/graphhopper/routing/weighting/DirectedRoutingTest.java
```diff
@@ -0,0 +1,323 @@
...
+    public RepeatRule repeatRule = new RepeatRule();
+
+    @Parameterized.Parameters(name = "{0}, u-turn-costs: {1}")
+    public static Collection<Object[]> params() {
+        return Arrays.asList(new Object[][]{
+                {Algo.ASTAR, Double.POSITIVE_INFINITY, false, false},
+                {Algo.CH_ASTAR, Double.POSITIVE_INFINITY, true, false},
+                {Algo.CH_DIJKSTRA, Double.POSITIVE_INFINITY, true, false},
+                // todo: yields warnings and fails
+//                {Algo.LM, Double.POSITIVE_INFINITY, false, true}
```
easbar: With landmarks the test fails sometimes, maybe its the same problem as in #1557 ? Or maybe I am using bad settings (there were also some warnings about some weights being maxed out)

core/src/main/java/com/graphhopper/util/EdgeIterator.java
karussell: Yeah, this is a good idea. Additionally we should have a look if we use `==-1` instead of `isValid` call. Have created #1666

core/src/test/java/com/graphhopper/routing/weighting/DirectedRoutingTest.java
karussell: This is ugly. Wasn't #1557 only for shortest weighting and shouldn't happen here?

easbar: Hmm, I might need your help here. I guess the test for LM fails because I am getting these warnings:

```shell
com.graphhopper.routing.lm.LandmarkStorage - landmark 1 (49.40121656807996,9.705012881122817): too many delta weights were maxed out (17/82). Factor is too small 0.003551062425503072. To fix this increase maximum in config.yml: prepare.lm.weighting: fastest|maximum=363.056352875689
```

I tried setting the maximum weight with `LandmarkStorage#setMaximumWeight` and also setting it to `-1` (automatic determination of maximum weight), but the warnings remain. I do not really understand what the problem is here. The random graph is created within a rather small bounding box (just the size of a village): http://tools.geofabrik.de/calc/#type=geofabrik_standard&bbox=9.7,49.41,9.71,49.4&tab=1&proj=EPSG:4326&places=2
and I checked that the edge distances are all something like 100-1000m. The speeds used are all between 10 and 120 so there should not be any unusual values for the weights (I checked and they are all somewhere between 60 and 420 (one to seven minutes).
However the `fromWeight` here: https://github.com/graphhopper/graphhopper/blob/022697da8dfc9ab6e6537fbc7b81efc55ac2e142/core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java#L554 seems to be really large (something like 30-80k, so several hours up to a full day) and because of this the `delta` in the same method is often below `DELTA_MIN`, which eventually leads to the warnings I mentioned above.

What I do not understand is how `fromWeight` can get so large ? I would expect in such a small area no matter where the landmarks are the weights should be much smaller ?

karussell: Thanks, will investigate! This sounds like a bug in LM code.

karussell: Hmmh, there are a few confusing things in the code like 16 bits for the precision although we should have 18 for the "from" and 13 for the "delta", plus old comments making it hard to "re-understand" this. It will take me a bit more time before I can understand this problem here.

easbar: Ok, thanks for taking a look! Maybe we can leave this test commented out for this PR ? The problem seems to be unrelated to the source/target edge restrictions that are new here ?

karussell: Will let you know when I know if it is indeed unrelated. (But maybe it is related as the preparation and also parts of the algo are node based)

easbar: I tried looking into this a bit and  found failing cases also without the new feature discussed here (and also for node-based routing). See https://github.com/graphhopper/graphhopper/issues/1687

karussell: Thanks a lot!!

core/src/main/java/com/graphhopper/routing/AbstractBidirAlgo.java
```diff
@@ -147,9 +174,48 @@ protected void postInit(int from, int to) {
     }
 
     protected void postInitFrom() {
+        // using a local reference is important to prevent stack overflow after calling setEdgeFilter
```
karussell: What is the problem here exactly? Is it due to the not fully initialized AbstractBidirAlgo constructor?

karussell: Ah, no. It seems we use additionalEdgeFilter inside the method. Wouldn't it then be safer to handle this inside the method especially since it is protected?

core/src/test/java/com/graphhopper/routing/DirectedBidirectionalDijkstraTest.java
```diff
@@ -0,0 +1,536 @@
...
+        }
+    }
+
+    @Test
+    @Ignore("todo: do we require running the algo with an additional edge filter ?")
+    public void worksWithEdgeFilter() {
+        // todo:
+        // e.g. block area is implemented using infinite weighting
+        // when an edge filter is used need to think about what should happen if the filter filters out the
+        // given in out edges etc.
```
karussell: Interesting and yes, if we do a block area and restrict the edge (with infinite weight) then we should get "connection not found".

core/src/main/java/com/graphhopper/routing/AbstractBidirAlgo.java
easbar: Hmm, I would say `fillEdgesFrom/ToUsingFilter` is 'sane': It sets `additionalEdgeFilter` to some other filter, calls `fillEdges` and sets `additionalEdgeFilter` back to the value it had before calling the method. So I do not think there is anything we can handle here ? The one thing that we must not do is using the `additionalEdgeFilter` reference directly in the `edgeFilter` object we pass into the method (because it will be modified itself). I tried making this a bit clearer in the code now.

core/src/test/java/com/graphhopper/routing/DirectedBidirectionalDijkstraTest.java
easbar: Yes, this is already covered. I added a test for this and removed the todo.

core/src/main/java/com/graphhopper/routing/AbstractBidirAlgo.java
karussell: And should we throw an illegal arg exc if edgeFilter==additionalEdgeFilter?

easbar: No, passing `edgeFilter==additionalEdgeFilter` is ok (we even do this here), the only thing that is not ok is passing a new edge filter that *uses* `additionalEdgeFilter` directly.

PR #1667, opened 2019-07-17
Replaces some == NO_EDGE comparisons with isValid calls.

easbar:
Are there any situations where we use edge ids that are `<0` and are not `-1` ? I added a check in `isValid` and ran the tests and there were no such situations in the tests.


karussell:
Thanks!

> Are there any situations where we use edge ids that are <0 and are not -1 ? 

I don't think so too.

PR #1669, opened 2019-07-25
Replaces GraphHopperStorage#getGraph(Class) with getCHGraph().

### LINKED ISSUE #1668: Why do we use Class<> to retrieve graphs from GraphHopperStorage ?
What is the reasoning behind `GraphHopperStorage#getGraph(Class<T> clazz)` ? 
There is already `GraphHopperStorage#getBaseGraph()`, which is the same as `GraphHopperStorage#getGraph(Graph.class)`, but to get the/a CH graph we need to do `GraphHopperStorage#getGraph(CHGraph.class)`. 

Would it not be more obvious to have `GraphHopperStorage#getCHGraph()` and `GraphHopperStorage#getCHGraph(Weighting)` ?

easbar:
Fixes #1668.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
+++ b/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
@@ -34,12 +34,11 @@ import java.util.List;
  * This class manages all storage related methods and delegates the calls to the associated graphs.
  * The associated graphs manage their own necessary data structures and are used to provide e.g.
  * different traversal methods. By default this class implements the graph interface and results in
- * identical behavior as the Graph instance from getGraph(Graph.class)
+ * identical behavior as the Graph instance from getBaseGraph()
  * <p>
  *
  * @author Peter Karich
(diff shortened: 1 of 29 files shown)


core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
```diff
@@ -95,42 +94,37 @@ public void freeze() {
         }
     }
 
+    public CHGraph getCHGraph() {
+        Collection<CHGraphImpl> chGraphs = getAllCHGraphs();
+        if (chGraphs.isEmpty()) {
+            throw new IllegalStateException("There is no CHGraph");
+        } else if (chGraphs.size() > 1) {
+            throw new IllegalStateException("There are multiple CHGraphs, use getCHGraph(Weighting) to retrieve a specific one");
```
easbar: Previously we simply returned the first CHGraph, but I think throwing an exception here is safer so one does not accidentally take the wrong one.

Issue #1652, opened 2019-07-01
Allow finite u-turn costs for CH


easbar:
Currently we cannot use finite u-turn costs with CH, but with #1640 this should now be possible to do.

easbar:
Doing this raises several questions.

1) Should we start with one global value for u-turns (the same for all road-types, junctions, dead end streets etc.) ?
2) If yes, should it be possible to use different values for the u-turn costs for each profile (probably makes sense, e.g. bike is certainly not the same as car) ? 
3) So far the u-turn costs 'live' inside `TurnWeighting`, but in the configuration file we would probably configure it with the encoder(s) ? Something like `car|turn_costs=true|u_turn_cost=40` ? Maybe we should move the u-turn-cost into the (turn cost)encoder ? If we ever add more options like different u-turn costs for different road types etc. this configuration might become lengthy...
4) Do we need to allow different u-turn-costs for the same vehicle/weighting (now or in the long run) ? related: #493
5) Once we have this configuration we definitely need to store the u-turn cost with each CH profile, because it will be fixed after the CH preparation. I think we should store it as an int in (units of seconds ?), so we do not run into floating point comparison problems when loading a profile. We could use -1 as special value indicating the u-turn costs shall be infinite (this is the default at the moment).

karussell:
As a start a global U-turn cost per profile makes sense. (Later on it should be at least dependent on road_class and dead end. But let's focus in this issue just on making it "possible" not especially "configurable")

> If we ever add more options like different u-turn costs for different road types etc. this configuration might become lengthy...

Yeah, before we do this it will be time to push forward the development of a yaml config per profile. Like I have done in a branch. See https://github.com/graphhopper/graphhopper/issues/193#issuecomment-512753969

> Do we need to allow different u-turn-costs for the same vehicle/weighting (now or in the long run) ? related: #493

Yes, I think this would be good. Like for different trucks. But probably also related to a different issue.

easbar:
Ok, sounds like a plan!

> But let's focus in this issue just on making it "possible" not especially "configurable")

Ok, so the most 'possible and not configurable' way would be to hard-code it somewhere in the FlagEncoder then we do not even need to add it to the current config.

karussell:
> Ok, so the most 'possible and not configurable' way would be to hard-code it somewhere in the FlagEncoder then we do not even need to add it to the current config.

Sure!

easbar:
[...]

PR #1671, opened 2019-07-29
Identifies CH graphs by new CHProfile class instead of Weighting.

easbar:
Instead of a list of `Weighting`s for edge/node-based preparations `GraphHopperStorage` now takes a list of `CHProfile`:

```java
public GraphHopperStorage(List<CHProfile> chProfiles, Directory dir, EncodingManager encodingManager, boolean withElevation, GraphExtension extendedStorage)
```

To get a `CHGraph` from `GraphHopperStorage` we are using now:
```java
graph.getCHGraph(CHProfile);
```

In contrast to the previous approach where the exact `Weighting` instance was used to identify a `CHGraph` the `CHProfile` now works like a finger-print of a `CHGraph` and it does not have to be the same object anymore (at least as long as the `Weighting` derives from `AbstractWeighting` which provides its own equals method).

Also see discussion in #1652.


core/src/main/java/com/graphhopper/routing/util/BikeFlagEncoder.java
```diff
@@ -36,8 +36,8 @@ public BikeFlagEncoder(String propertiesString) {
     }
 
     public BikeFlagEncoder(PMap properties) {
-        this((int) properties.getLong("speed_bits", 4),
-                properties.getLong("speed_factor", 2),
+        this(properties.getInt("speed_bits", 4),
```
easbar: Right ?

karussell:
Thanks! 

Is it possible with this PR to make this part cleaner here? https://github.com/graphhopper/map-matching/blob/master/matching-core/src/main/java/com/graphhopper/matching/MapMatching.java#L116

easbar:
> Is it possible with this PR to make this part cleaner here?

Hmm, I am not sure (I do not understand this comment):

> I want to use my own instance of FastestWeighting because it uses its own U-Turn-penalty

How can `FastestWeighting` use a u-turn penalty ? Or should this read *heading*-penalty here ? If the heading penalty is meant this is not supported by CH and since the heading penalty is not included in `FastestWeighting#toString` it does not affect `AbstractWeighting#equals`. 

The important thing that changed in this PR is that `Weightings` are now identified by `equals` instead of `==`. All properties of a `Weighting` should be included in `toString`, such that the CH preparation matches the `Weighting` that we use. For the heading penalty its a special case, because this is not supported for CH.

easbar:
Any idea why the CI build fails for open jdk 12 ? 

> [ERROR] Failed to execute goal de.thetaphi:forbiddenapis:2.6:check (default-cli) on project graphhopper-tools: Check for forbidden API calls failed while scanning class 'com.graphhopper.ui.MiniGraphUI' (MiniGraphUI.java): java.lang.ClassNotFoundException: com.graphhopper.storage.CHProfile (while looking up details about referenced class 'com.graphhopper.storage.CHProfile') -> [Help 1]

karussell:
> Any idea why the CI build fails for open jdk 12 ?

Strange, can you try to remove the cache?

```
cache:
  directories:
  - $HOME/.m2
```

> How can FastestWeighting use a u-turn penalty ? 

@michaz do you remember what the problem was and why you had to create a separate Weighting instance?

easbar:
Hmm, I do not get it. Disabling the cache did not help and adding gh-core as explicit dependency to tools did not help either. Somehow the wrong version of the core (that does not have `CHProfile`) must be pulled in for the forbiddenapis check. 

I also tried changin all versions to `0.13-ch_profile-test`, but then the Open JDK 12 build fails with another error:

```shell
[ERROR] Failed to execute goal on project graphhopper-web-api: Could not resolve dependencies for project com.graphhopper:graphhopper-web-api:jar:0.13-ch_profile-test: Could not find artifact com.graphhopper:graphhopper-api:jar:0.13-ch_profile-test in google-maven-central (https://maven-central.storage-download.googleapis.com/repos/central/data/) -> [Help 1]
```

So it seems that it is trying to obtain graphhopper module artifacts from maven central instead of using the local ones from the reactor build ?

karussell:
Really strange.

I got something that seems to works ... don't ask why though ;)
I've just set the in-built "openjdk12" jdk for travis and it works! Although it does the same (downloads the install-jdk script and runs it).

easbar:
Thanks! I enabled the cache again and set back the artifact versions and the build passed a second time. No idea what this was...

michaz:
> How can FastestWeighting use a u-turn penalty ? Or should this read heading-penalty here ? If the heading penalty is meant this is not supported by CH anyway. Since the heading penalty is not included in FastestWeighting#toString it does not affect AbstractWeighting#equals so for CHGraph#getCHGraph with this PR we will get the same CHGraph regardless what heading penalty we are using.

Yes, it should read heading-penalty. The heading-penalty is used only in the map-matching specific entry/exit phase. As you say, not during CH routing.

The entire comment is basically a complaint that I have to use a weighting that I don't want to route with (`((PrepareContractionHierarchies) routingAlgorithmFactory).getWeighting()`) as a key, just to get the `Graph` that I want, even though I know that my own weighting will work with that `Graph`.

And if that's what your change fixes, all is fine, and the ugliness _and_ the comment can go away.

michaz:
Oh, by the way, this is the exact right moment to change this: In (one step of the) the customizable routing project, we will need to do exactly that: Have a fully dynamic weighting that we create on the fly, and use it on a `CHGraph` of which we ensure "externally" that it fits the weighting.

michaz:
(Or is that the idea?) :-)

michaz:
In the end, I think, profiles will just have names (strings):

- I will define profiles in cofiguration files, giving them a name
- Users will say which profile they want by query parameter
- ... so I should also be able to get a profile, or something pertaining to a profile, by name in Java.

easbar:
> In the end, I think, profiles will just have names (strings):

Yes I agree. It is more or less like this at the moment, because `CHProfile#equals` uses `equals` from `Weighting`, which for `AbstractWeighting` uses `toString`... So should we add `getCHGraph(String)` ? This should be easy.

For the customizable scenario I think we should keep two things in mind:

- The 'definition' of a CH profile can get arbitrarily complex: Its not just about the weighting or node/edge based, but for example when blocking (multiple) areas or even specifying exact speeds for many single edges (think traffic conditions) a lot of information has to go into the 'profile' specification. So it would probably better to use some structured format for the 'profile-string' ?

- I am not sure if `graph.getCHGraph()` will be really what we want for customizable CH, because it should also be possible to create the CH graph on the fly (basically create it, run a request and throw it away immediately). So there will be no need to store it in `GraphHopperStorage` even. At the same time there might be places where it is more like creating a profile and then use it multiple times, in this case it might be best to assign some kind of ID to the profile to retrieve it later (using the full specification of the profile would be too much if, e.g. it includes specifying all traffic conditions etc.).

> (Or is that the idea?) :-)

The idea here was mostly that I wanted to be able to specify u-turn costs (#1672), but sure I think its a step in the right direction for more customized stuff as well. The key thing here is that everything that needs to be known at CH preparation time must be included in the CH profile (used to be just the weighting, then edge/node-based and with #1672 also u-turn costs).

michaz:
> to assign some kind of ID to the profile to retrieve it later

Yes, exactly. Like a filename, almost. We already have this `Directory` abstraction -- don't know how far that carries, but basically, our "normal" profiles are currently:

/car/fastest
/car/shortest
/bike/fastest
/bike/shortest
.
.

(right?)

But maybe I want to do something outside that structure. Like, create a CH Graph where some parts of the graph are explicitly banned from contraction (so I can customize them). Maybe I want several of those. They would be created / programmed in user code, and I would give them a name in user code, maybe

/blockableRoadsExperiment/car/fastest

so I can get at them (in other parts of user code).

(Just a thought, maybe non-hierarchical names are enough in the end..)

karussell:
> to assign some kind of ID to the profile to retrieve it later

I think, this should be another issue and should not be done before the actual code is required.

The flexible work (flex_vehicleprofile and flex_vehicleprofile2) indicates that everything can be weighting based (we do not need FlagEncoders anymore). I.e. we just use a weighting name as ID "mycar/fastest" and a base name "car/fastest" from which we extend behaviour. Customization properties are sent via a json/yaml file either through http or from the file system.

easbar:
> I think, this should be another issue

Yes, is there anything that needs to be done for this PR ?

> The flexible work (flex_vehicleprofile and flex_vehicleprofile2) indicates that everything can be weighting based (we do not need FlagEncoders anymore). I.e. we just use a weighting name as ID "mycar/fastest" and a base name "car/fastest" from which we extend behaviour. 

What do you mean with 'we do not need FlagEncoders anymore' ? Will they be removed entirely ? "car/fastest" still sounds like a `FastestWeighting` with a `CarFlagEncoder` ? Is `flex_vehicleprofile2` just a more recent version of `flex_vehicleprofile` or are they something different ? Sorry I did not have a look at these yet.

core/src/main/java/com/graphhopper/storage/GraphBuilder.java
```diff
@@ -36,22 +37,28 @@
...
 
     /**
-     * This method enables creating a CHGraph with the specified weighting.
+     * This method enables creating a CHGraph with the specified CHProfile
      */
-    public GraphBuilder setCHGraph(Weighting singleCHWeighting) {
-        this.singleCHWeighting = singleCHWeighting;
+    public GraphBuilder setCHProfiles(List<CHProfile> chProfiles) {
+        if (chProfiles.size() != new HashSet<>(chProfiles).size()) {
+            throw new IllegalArgumentException("Duplicate CH profile: " + chProfiles);
```
karussell: The message is slightly confusing and should be "Contains duplicate CH profile:" (?)

karussell:
> "car/fastest" still sounds like a FastestWeighting with a CarFlagEncoder ? 

Yes. Still conceptionally the FastestWeighting does not need CarFlagEncoder. In theory it is sufficient to have e.g. roadClassEnc, maxSpeedEnc and a few other EncodedValues that "read the properties of an endge" in the Weighting class. The "CarFlagEncoder-specific stuff" that precalculates e.g. average speed and access for car seems to be required only for performance reasons. Or maybe the otherway around: should we create a separate TruckFlagEncoder to store size and width or should we include this in CarFlagEncoder? But e.g. cargo bikes also need at least the width of an edge. And why should we force a single averageSpeedEnc for car, when in fact there could be many speeds e.g. depending on the time or if it is an electric car?
I do not know yet what a good replacement for FlagEncoder should look like as it is still handy to have some precalculated average speed for foot and car, but still IMO we should try to remove its usage as it is only a specific collection of EncodedValues (?) 
(For now EncodedValues handle edges only, later they could handle also nodes and relations.)

> Is flex_vehicleprofile2 just a more recent version of flex_vehicleprofile or are they something different ? 

Yes, flex_vehicleprofile2 uses a recent master and has no hacks I think, but has also not all features.

easbar:
Ok, thanks for your explanations and I see it is difficult to find replacement for everything FlagEncoder does. To specify a CH profile I think we will eventually only need to define a weight for all edges of the base graph (in fwd&bwd direction) and give this collection of weights some name/id to retrieve the corresponding profile later. Of course for json/yaml configuration etc. specifying these weights needs to be possible on a much higher level (like 'weight=dist/speed', 'block this area', 'use this traffic information' etc.). 

Right now `Weighting` (combined with `FlagEncoder`) does all these things: It has a `calcWeight` method, there are different implementations that allow specifying how the weight is calculated from the different edge properties and it is used as identifier of the CH graph (therefore it is still part of `CHProfile`). Maybe at some point we need to separate these things from each other. For example the CH preparation should not be concerned with any details (encoded values, flag encoders, traffic information sources,...) *how* the weights are calculated (it only needs a way to retrieve the two weights per edge) and as @michaz pointed out it should be possible to choose an arbitrary name/id of a CH profile, because it will rather be impractical to include the full specification of a weighting (blocked areas etc.) in its name.

karussell:
> I see it is difficult to find replacement for everything FlagEncoder does

Yes. This is another problem: it does too many things :) 

> it only needs a way to retrieve the two weights per edge

Yeah, basically it just needs two EncodedValues. (Maybe a "profile" is like a FlagEncoder: a collection of EncodedValues...)

easbar:
> Yeah, basically it just needs two EncodedValues.

I was more thinking of two doubles ? I thought every edge has some 'flags', then there are encoded values to give these flags some meaning (like speed or road type), then there is Weighting that uses the encoded values to calculate a weight (just a number). For CH preparation we need this weight, why would we convert the weight (number) into an encoded value at this point ?

karussell:
> For CH preparation we need this weight, why would we convert the weight (number) into an encoded value at this point ?

An EncodedValue transform some bits in the edge into a value. The same can be done for shortcuts. In fact the only reason we don't use EncodedValues also for the shortcut weights yet is entirely because of performance. Maybe you remember this commit: https://github.com/graphhopper/graphhopper/commit/cb880186ddcba1bb46df94ee9f0631a658e7123f (was reverted or not included)

> But isn't it more than a collection of encoded values ?

It is more than that, yes. Like including relations and stuff. 

> It is also about how to combine the encoded values into a weight ?

But isn't this the task of the Weighting class?

easbar:
>  The same can be done for shortcuts

Ah ok yes you were thinking about the shortcuts, which need to store some weight (and like you said could be encoded values). I was thinking about the weights of the original/base graph edges that *enter* the CH preparation.

> Isn't this the task of the Weighting class?

Yes, exactly. I was probably more thinking in terms of a CH profile, which is defined by the calculated weights of the edges or at least by the 'method'/weighting how these weights are calculated. This is different to a 'profile', which rather defines the properties of an edge and is needed *before* the weights are calculated.

PR #1672, opened 2019-07-29
Allows setting finite u-turn costs.

### LINKED ISSUE #1590: Using turn costs, not just turn restrictions
OSM data only contains turn restrictions like 'right only' etc., but graphhopper routing can also handle turn *costs*, for example to prefer routes with less turns, less left-turns or similar. However, there are some open questions how turn costs should be handled, see discussion in: #1586

### LINKED ISSUE #1652: Allow finite u-turn costs for CH

easbar:
This PR is based on #1671 

With this PR it is possible to allow u-turns at some cost (when edge-based routing is enabled). On the request we have to specify, e.g., `u_turn_costs=40` which means each u-turn that the route takes adds `40s` to the overall travelling time. This only really makes sense as long as the weighting is based on time, like `FastestWeighting`. See also #1590.

Maybe this needs a bit of further thought and the u-turn costs should only be possible to apply if the weighting is time-based. Maybe we should deal with #1590 first ? Or just add some documentation and declare this parameter as experimental ?

However the largest part of this PR is about CH with finite u-turn costs. As for CH the u-turn costs must be specified in advance. This is done via the new `CHProfile` from #1671. 

In config.yml the new configuration looks like this:
```yaml
# use finite u-turn costs of 40s
prepare.ch.weightings: fastest|u_turn_costs=40
# if nothing is specified u-turns are forbidden as before
prepare.ch.weightings: fastest
# explicitly deny u-turns
prepare.ch.weightings: fastest|u_turn_costs=-1
```

Fixes #1652, also see discussion there.


easbar:
There is a problem here: When there are additional virtual nodes between the start and target (as in via-routing) there are additional possibilities to take a u-turn (at the virtual nodes). This means the shortest path can become dependent on the existence/non-existence/position of virtual nodes which is rather ugly. One possibility would be to never allow u-turns at virtual nodes ? Related #1593.

See here: bbf6337

michaz:
> This means the shortest path can become dependent on the existence/non-existence/position of virtual nodes which is rather ugly. One possibility would be to never allow u-turns at virtual nodes ?

Yeeeaaaaah, we had exactly the same conceptual difficulty once in map matching.

My suggestion was to construct the `QueryGraph` differently, so that the entry/exit is basically next to the original graph (connected with one-way edges) rather than inside it, so shortest paths in the original graph aren't altered. Don't know how hard that is, I think `QueryGraph` is pretty sophisticated. ,-)

karussell:
> My suggestion was to construct the QueryGraph differently, so that the entry/exit is basically next to the original graph

Can you elaborate your idea? Let's assume the following graph with an original edge A-B, where the entry is x:

```
 --A---x-----B---
```

And now you mean we add one-way edges `A->x` and `x->B`? Wouldn't this lead to other bugs as the original edge is basically duplicated?

If I remember correctly I had 3 options: 
1. add edges to the graph like we do, 
2. modify the algorithm and do not start with x and a weight=0 but use two starts A and B with a weight depending e.g. distance to x (seemed too complex to me as this problem will be spread over multiple areas, but now we already have special cases for virtual nodes in the CH and LM algorithms it does not seem to bad)
3. at import create only tower nodes and if distance is too large create additional tower node (this is prohibitive now in terms of memory used and also it was much slower)

michaz:
> And now you mean we add one-way edges A<-x and x->B

Yep.

> Wouldn't this lead to other bugs as the original edge is basically duplicated?

Don't think so. You can even add A->y and y<-B, then you have x for one direction and y for the other, without headings or blocking or anything..

michaz:
> have special cases for virtual nodes in the CH and LM 

Aaaahhhhh! :-)

karussell:
> You can even add x<->A and y<->B, then you have x for one direction and y for the other, without headings or blocking or anything.

Let's say you use heading and forbid going east, then, if the original edge still exists, how do you avoid such strange u-turns at B ala `y->B->A`?

michaz:
I don't. Shortest path is shortest path. Otherwise you need (U)-turn costs.

But I get it, all these details serve as heuristics for other things, right...

michaz:
Also this wasn't the full truth, there are more issues when origin/destination are on the same link etc.

michaz:
:'-(

michaz:
But do you see my point about the headings? If I talk about headings, I am implicitly talking about directed edges, and also about non-zero turn costs, because otherwise headings would make no difference.

So in a way it is cheating to just solve the U-turn problem by removing the original graph edge, because _in reality_ it could still be that _doing_ the U-turn is, _in fact_, the fastest way of getting to my destination.

easbar:
> modify the algorithm and do not start with x and a weight=0 but use two starts A and B with a weight depending

I think if I would have to do it again this is the way I would try. But unless we find a case where it is really needed I do not think this will be changed anytime soon :)

core/src/main/java/com/graphhopper/routing/ch/CHProfileSelector.java
```diff
@@ -0,0 +1,185 @@
...
+
+    private CHProfile select(HintsMap weightingMap, Boolean edgeBased, Integer uTurnCosts) throws CHProfileSelectionException {
+        this.weightingMap = weightingMap;
+        findCHProfilesMatchingWeighting();
+
+        if (!foundCHProfilesMatchingWeighting()) {
+            throw new CHProfileSelectionException("Cannot find CH profile for weighting map " + weightingMap + " in entries: " + entriesStrs + ".");
+        }
+
+        if (edgeBased != null && uTurnCosts != null) {
```
easbar: This is again quite a bit of nesting (even though its spread over different methods), but I thought this way it is easier to be sure we cover all the possible cases ? wdyt ?

easbar:
I think I am happy with forbidding u-turns at virtual nodes. I added a comment to the docs stating that the u-turn penalty is time-based and therefore only works with time-based weightings, like fastest here: https://github.com/graphhopper/graphhopper/pull/1672/commits/3fafe816f53d21b8b58e77946f0a486ef17546a8. Is this sufficient or should there be a more prominent warning (maybe at the place where the `u_turn_costs` parameter is documented ? Should we document it already ?

Should this go into `0.13` ? I think it's ready.

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
```diff
@@ -251,46 +252,22 @@ public RoutingAlgorithmFactory getDecoratedAlgorithmFactory(RoutingAlgorithmFact
...
-            }
-        } else {
-            // no edge_based parameter was set, we determine the CH preparation based on what is there (and prefer edge-based
-            // if we can choose)
-            return edgeBasedPCH != null ? edgeBasedPCH : nodeBasedPCH;
+    private CHProfile selectProfile(HintsMap map, Boolean edgeBased, Integer uTurnCosts) {
+        try {
+            return CHProfileSelector.select(chProfiles, map, edgeBased, uTurnCosts);
+        } catch (CHProfileSelectionException e) {
+            throw new IllegalArgumentException(e.getMessage());
```
karussell: Is the checked exception CHProfileSelectionException required? Couldn't we let it extend DetailedIllegalArgumentException and avoid this method here and the 'throws CHProfileSelectionException' in a few methods? Or do you want it to be explicit? (I try to avoid checked exception as they are not often very helpful IMO)

core/src/main/java/com/graphhopper/routing/ch/CHProfileSelector.java
karussell: Looks fine due to the named cases IMO

karussell:
Is this sufficient or should there be a more prominent warning (maybe at the place where the u_turn_costs parameter is documented ? Should we document it already ?

Yes. May in ch.md and in the config-example.yml?

> Should this go into 0.13 ? I think it's ready.

Sure :+1: - Thanks!

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
easbar: Ok, I changed it to a runtime exception. I do not necessarily think checked exceptions are never helpful because they give you a compile time check that helps you remembering to handle the error cases (for example here I think this is more of an advantage than it is a disadvantage that some 'intermediate' methods need to declare the exception). Anyway, either way is fine for me :+1:.

easbar:
> Yes. May in ch.md and in the config-example.yml?

Its in turn-restrictions.md and config-example.yml now

PR #1678, opened 2019-08-02
Allow disabling the clean hack garbage collection

ChristophKaser:
.. by setting the property "com.graphhopper.disableCleanhack" to "true".

(because System.gc triggers a several second pause in our process)

### DIFF
--- a/api/src/main/java/com/graphhopper/util/Helper.java
+++ b/api/src/main/java/com/graphhopper/util/Helper.java
@@ -345,6 +345,9 @@ public class Helper {
      * http://stackoverflow.com/q/2972986/194609 and use only if you know what you are doing.
      */
     public static void cleanHack() {
+        if ("true".equals(System.getProperty("com.graphhopper.disableCleanhack"))) {
+    		return;
+        }
         System.gc();
     }
 


karussell:
Thanks! What is the reason you need this? Can you describe in more detail what you tried?

karussell:
Can you try if the JVM parameter `-XX:+DisableExplicitGC` helps?

ChristophKaser:
We have a web service that sometimes (only rarely) needs to calculate a route using graphhopper. If we use a graphhopper-instance configured using .forServer() it takes about 1 second to load the map and calculate the request. Using .forMobile(), opening the map and calculating the request takes about 60 ms and does not use a lot of memory in the heap. However, without this patch, when we close the map the whole server process freezes for 5 seconds due to the forced STW garbage collection.

The JVM parameter would help, but has two disadvantages:
- There is no way to specify it from within the java process after it has launched
- It disables every call of System.gc(), and we have some debug code that calls System.gc() to evaluate memory requirements of the process

karussell:
This is an interesting use case and I see your point. Can you benchmark this and see if the memory usage increases with the amount of these opening and routing requests? To do this you could limit -Xmx to a very low value where the system still works without "out of memory" errors and then do this e.g. 100K times.

ChristophKaser:
I tried opening, routing a single request and closing the same map file for 10 000 times (100K would have taken about 3 hours) and did not find any issues - the heap usage seemed to be stable as far as I could tell.

karussell:
How large is your PBF file and which Xmx setting did you use? 

The question is: when the release of mmap buffer is not an issue why do we use the hack at all?

ChristophKaser:
I did not keep the pbf file, but the graphhopper files have a combined size of 700 MB. The process had a Xmx of 4 GB (which was primarily filled with other caches and data objects).
But if I understand correctly, the Graphhopper map is not loaded into the heap when I use .forMobile(), but is only mapped into the virtual memory of the java process. Since I use a 64-bit process, I doubt it is easy to exhaust the virtual address space, so I don't think there will be OOM-issues - it might however lead to catastrophic swapping if the OS prefers the mmap buffer to the java heap. This is especially likely if the heap is large and GCs occur only rarely, because this means it can take a long time until the MMAP buffer object is garbage collected and therefore unmapped. Maybe that is the reason for the hack?

Another reason I can think of: As long as the files are mmapped, they cannot be moved or deleted (on windows). This would mean that even though I closed a graphhopper map, I cannot remove the files on the disk.

ChristophKaser:
I just checked the graphhopper source, and there is a method "cleanMappedByteBuffer" that is used to unmap the ByteBuffer via reflection and the sun.misc.Unsafe class. Therefore, I don't really understand why the forced garbage collection is needed - maybe it was meant as a fallback?

karussell:
> Therefore, I don't really understand why the forced garbage collection is needed - maybe it was meant as a fallback?

Good question. Will investigate. Maybe we can remove "cleanHack" completely (would be my favourite :) ) and it was only for Android in old versions (#933).

You are right that the Java heap is not the limiting factor. Instead we would have to try this loop an a (virtual) machine with only e.g. 1GB RAM and see if the JVM is able to release the natively requested memory mappings (that occupy memory not just virtual space IMO). 

The System.gc call would be only required if no cleaner method was available on the system. So maybe we return true/false in cleanMappedByteBuffer or something.

karussell:
Can you try the following branch with your MMAP setting:
https://github.com/graphhopper/graphhopper/commits/remove_cleanhack 

We need to ensure that this works on desktop, server and Android. 

@devemux86 can you try this too? (have released 0.13.0-tmp-cleanhack, will be available in ~30min on maven central) 
And see if it introduces problems when the app is often reopened?

devemux86:
> @devemux86 can you try this too?

I'm still on version `0.10.3`. :slightly_smiling_face: 
Need to test with GraphHopper Android sample app.

karussell:
> I'm still on version 0.10.3

oh, ok. What was the problem again?

devemux86:
>>I'm still on version 0.10.3

>oh, ok. What was the problem again?

The `0.11` module changes leaded to new inherited dependencies that load (or could break) Android apk and can be critical in future with the known Java 7/8 compatibility issue on Android.

For start I don't use anymore the `client-hc` module (created my own light client without third party dependencies), so part of the problem was remedied.

Now I need to find some free time to study what is needed for the rest GraphHopper modules on Android + Desktop.

karussell:
Will close here as #1699 replaces this.

PR #1683, opened 2019-08-08
Accept turn relation

sguill:
To enable the _flagEncoders_ to accept or not a **turn relation**.

For example, a _busFlagEncoder_ does not take into account the following turn relation, whereas the _carFlagEncoder_ does:
type=restriction
restriction=no_left_turn
**except**=bus

And a _busFlagEncoder_ wants to respect the following turn relation dedicated to bus, whereas _carFlagEncoder_ doesn't:
type=restriction
**restriction:bus**=no_left_turn


karussell:
Thx, we'll take care of this (in a few weeks)

easbar:
Would you mind updating this to recent master ? The problem why the build failed should now be fixed with #1671.

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -181,6 +181,14 @@ public int defineRelationBits(int index, int shift) {
         return shift;
     }
 
+    /**
+     * Decide whether a restriction relation is enabled for a given mode of travel. This skips some relations before
```
easbar: No need to repeat the javadocs that are already in `FlagEncoder` here

core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
```diff
@@ -53,6 +54,14 @@
      */
     DecimalEncodedValue getAverageSpeedEnc();
 
+    /**
+     * Decide whether a restriction relation is enabled for a given mode of travel. This skips some relations before
+     * handleRelationTags is called.
+     *
+     * @return the encoded value to indicate if this encoder allows travel or not.
+     */
+    EncodingManager.RelationAcceptation getRelationAccept(ReaderRelation relation);
```
easbar: Why return `RelationAcceptation` here ? Why not just a boolean ? All this method does is deciding whether or not the turn rrestriction should be applied for this encoder, or am I missing something ?
I am also wondering why this is about 'relations' in general since we are only using it to filter turn restrictions here. Also, I'm not a native English speaker, but I do not think `accept` is a noun ? What also bothers me a bit is that the parsing of the 'dedicated'/conditional turn restrictions is kind of spread across `OSMReader#createTurnRelation` and `CarFlagEncoder#hasRestrictedRelationTag`. What if we wanted to do the same for another encoder than car ? We would have to do the same parsing again. Can we not do this in `OSMReader` already and provide the parsed information (like the `except` list to the encoder) (probably by putting it into `OSMTurnRelation`) ?

For these reasons I think the signature of this method should rather be:

```java
public abstract boolean acceptsRelation(OSMTurnRelation relation);
```

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -510,6 +510,76 @@ public boolean canSkip() {
...
+        if (!acceptRelation.isEmpty())
+            throw new IllegalArgumentException("AcceptRelation must be empty");
+
+        for (AbstractFlagEncoder encoder : edgeEncoders) {
+            acceptRelation.put(encoder.toString(), encoder.getRelationAccept(relation));
+        }
+        return acceptRelation.hasAccepted();
+    }
+
+    public static class AcceptRelation {
```
easbar: Can you explain what this class is needed for ?

core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
```diff
@@ -186,6 +186,36 @@ protected double getSpeed(ReaderWay way) {
...
+    private boolean hasRestrictedRelationTag(ReaderRelation relation) {
+        boolean hasExcept = relation.hasTag("except");
+        boolean hasRestrictionDedicated = relation.hasTagStartsWith("restriction:");
+        if (hasExcept || hasRestrictionDedicated) {
+            String except = relation.getTag("except");
+            for (String r : restrictions) {
+                if (relation.hasTag("restriction:" + r)) {
+                    return true;
+                }
+                if (except != null && !except.equals("")) {
```
easbar: Maybe use `Helper.isEmpty(except)` here

core/src/main/java/com/graphhopper/reader/ReaderElement.java
```diff
@@ -95,6 +95,15 @@ public String getTag(String name) {
         return val;
     }
 
+    public <T> T getTagStartsWith(String keyPrefix, T defaultValue) {
+        for (String key : properties.keySet()) {
+            if (key.startsWith(keyPrefix)) {
+                return getTag(key, defaultValue);
```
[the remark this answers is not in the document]
sguill: Ok, which exception should I throw? Maybe `UnsupportedOperationException`?

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
sguill: ok

core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
sguill: Ok, I think I understand how to correct it. I will try. I did that way to make similar with `Way`, but you're right, only a boolean is needed for `Relation`.

core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
sguill: ok

core/src/main/java/com/graphhopper/reader/ReaderElement.java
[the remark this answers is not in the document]
easbar: Its basically an error caused by the user using a keyPrefix than can be ambiguous (?). I think I would go with IllegalArgument or IllegalState.

core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
sguill: I don't know how to deal with the class `OSMTurnRelation` (subproject reader-osm) used into `FlagEncoder` (subproject core), the dependency is in the other direction. Do I need to create a new class (for example `ReaderTurnRelation`) which extends `ReaderRelation` to store the parsed data (like the `except` list)? Or do I need to use the `members` list of `ReaderRelation` which doesn't seem appropriate? Thanks

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
sguill: It is not needed.

core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
easbar: @karussell is there any reason why we should not move `OSMTurnRelation` into core (next to `ReaderRelation` etc.) ?

easbar: I think I would just move it into core.

sguill: `OSMTurnRelation` uses `OSMReader` in function `getRestrictionAsEntries`.

easbar: Right, but that does not make much sense to me. `OSMTurnRelation` is an (intermediate) data class here. Extracting the `TurnCostTableEntry`s is part of the (OSM-parsing) algorithm which happens in `OSMReader`, so the dependency should be turned inside out: `OSMTurnRelation` should give access to its data and the `getRestrictionAsEntries` method should go into `OSMReader`. To make sure its still testable (as in `OSMTurnRelationTest`) maybe either we can make the method static or put this part of `OSMReader` into a separate class. Does that make sense ?

karussell: This definitely needs some refactoring. The best change, likely in a separate issue, would be if it is decoupled from the FlagEncoder so that the CarFlagEncoder can use car turn restrictions but it can be also used from a BusFlagEncoder without the requirement to extend this class from the CarFlagEncoder (i.e. composition over inheritance).

And the other problem is that the parsing part is something OSM-specific that should not be in core. We have the same difference for e.g. the `max_weight`-EncodedValue vs. OSMMaxWeightParser. But even there the OSM-specific class is still in core.

easbar: @sguill do you want to do this refactoring in a separate PR ? if not I can do it as well.

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -859,7 +863,25 @@ private long createNewNodeId() {
...
     OSMTurnRelation createTurnRelation(ReaderRelation relation) {
-        OSMTurnRelation.Type type = OSMTurnRelation.Type.getRestrictionType(relation.getTag("restriction"));
+        String tagRestriction = null;
+        String vehicleTypeRestricted = null;
+        List<String> vehicleTypesExcept = null;
+        if (relation.hasTag("restriction")) {
+            tagRestriction = relation.getTag("restriction");
+        } else if (relation.hasTagStartsWith("restriction:")) {
+            vehicleTypeRestricted = relation.getKeyStartsWith("restriction:");
+            if (!Helper.isEmpty(vehicleTypeRestricted)) {
```
easbar: how can `vehicleTypeRestricted` be empty here (it is supposed to start with "restriction:" already.

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -859,7 +863,25 @@ private long createNewNodeId() {
...
+                tagRestriction = relation.getTag(vehicleTypeRestricted);
```
easbar: this one I do not understand. we know relation has a tag that startss with "restriction:", but how can we know there is a tag for the key `vehicleTypeRestricted` ? if there is not we get a null pointer exception ?

reader-osm/src/test/java/com/graphhopper/reader/osm/OSMReaderTest.java
```diff
@@ -800,6 +798,70 @@ public void testTurnFlagCombination() {
         }
     }
 
+    @Test
+    public void testConditionalTurnRestriction() {
```
easbar: nice! :+1:

core/src/main/java/com/graphhopper/reader/OSMTurnRelation.java
```diff
@@ -0,0 +1,151 @@
...
+                result = tags.get(tag);
+            }
+            return (result != null) ? result : UNSUPPORTED;
+        }
+    }
+
+    /**
+     * Helper class to processing purposes only
+     */
+    public static class TurnCostTableEntry {
```
easbar: This can probably be moved into OSMReader now ?

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -877,12 +899,71 @@ OSMTurnRelation createTurnRelation(ReaderRelation relation) {
                 }
             }
             if (fromWayID >= 0 && toWayID >= 0 && viaNodeID >= 0) {
-                return new OSMTurnRelation(fromWayID, viaNodeID, toWayID, type);
+                return new OSMTurnRelation(fromWayID, viaNodeID, toWayID, type, vehicleTypeRestricted, vehicleTypesExcept);
```
easbar: I am wondering: Do we need to do any kind of trimming for `vehicleTypeRestricted` and `vehicleTypesExcept` values (i.e. can there be whitespace before or after the tags in xml ? Also I wonder if we should keep the constructor with the first four arguments for `OSMTurnRelation` and add `setVehicleTypeRestricted`/`addVehicleTypesExcept` so we can get rid of some of the null checks ?

core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
easbar: I think this is good now. `OSMReader` does the osm parsing part and each encoder can decide what turn restrictions should apply for it (given the information about what vehicles where specified in the restriction tags (including the except list)). The flag encoders still contain a lot of stuff that is specific to OSM tags, but at least the parsing happens in OSMReader. What is still a bit ugly is that `OSMTurnRelation` contains osm way ids, while for the purpose here this is not needed. We could separate this further to make this class less osm specific, but not sure if this needs to be done here.

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -181,6 +178,8 @@ public int defineRelationBits(int index, int shift) {
         return shift;
     }
 
+    public abstract boolean acceptsRelation(OSMTurnRelation relation);
```
easbar: Maybe we should provide a default implementation (returning true) here, such that already existing flag encoders (in other projects) do not need to be updated to keep the same behavior.

core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
```diff
@@ -186,6 +187,17 @@ protected double getSpeed(ReaderWay way) {
         return speed;
     }
 
+    @Override
+    public boolean acceptsRelation(OSMTurnRelation relation) {
+        if (relation == null) {
```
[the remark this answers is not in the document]
sguill: I added this test because of `OSMReaderTest.testTurnFlagCombination` with the call `reader.analyzeTurnRelation(null)`. Now, to respect your comment and this contraint, I updated the code in `OSMReader.analyzeTurnRelation` to test the `OSMTurnRelation turnRelation` is not null.

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -859,7 +863,25 @@ private long createNewNodeId() {
      * @return the OSM turn relation, <code>null</code>, if unsupported turn relation
      */
     OSMTurnRelation createTurnRelation(ReaderRelation relation) {
-        OSMTurnRelation.Type type = OSMTurnRelation.Type.getRestrictionType(relation.getTag("restriction"));
+        String tagRestriction = null;
```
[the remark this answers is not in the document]
sguill: ok

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
sguill: It's right, I remove the test.

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
sguill: I added a function `getKeyStartsWith` in `ReaderElement`, to get the key (not the value) starting with "restriction:", for example, we will have `vehicleTypeRestricted` = "restriction:motor_vehicle". Then it is parsed (removing "restriction:") `vehicleTypeRestricted`  = 
"motor_vehicle" to be stored in OSMTurnRelation.

core/src/main/java/com/graphhopper/reader/OSMTurnRelation.java
sguill: ok

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
sguill: I added trim function and try to remove the unneeded tests.

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
sguill: ok

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -431,7 +430,7 @@ public void processRelation(ReaderRelation relation) {
         Map<Long, TurnCostTableEntry> entries = new LinkedHashMap<>();
 
         for (FlagEncoder encoder : encodingManager.fetchEdgeEncoders()) {
-            if (encoder.acceptsRelation(turnRelation)) {
+            if ((turnRelation == null) || ((turnRelation != null) && encoder.acceptsRelation(turnRelation))) {
```
easbar: We can remove this null check (we already check for null after `createTurnRelation` and the 'default' is 'nothing is null').

docs/core/turn-restrictions.md
```diff
@@ -28,4 +28,17 @@ profiles with different u-turn costs). To disable the 'speed mode' per request y
...
+> type=restriction  
+> restriction:bus=no_left_turn
+
+Another example, using the *except* tag, means only *bicycle* are allowed to turn left:
+
+> type=restriction  
+> restriction=no_left_turn  
+> except=bicycle
+
+Redefining `acceptsRelation` function in *encoders* enable to change the default turn restriction use.
```
easbar: How about:

You can overwrite `FlagEncoder#acceptsTurnRelation` to change the default handling of turn restrictions in your customized vehicle profile.

core/src/main/java/com/graphhopper/reader/ReaderElement.java
```diff
@@ -95,6 +97,32 @@ public String getTag(String name) {
         return val;
     }
 
+    public <T> T getTagStartsWith(String keyPrefix, T defaultValue) {
```
easbar: this one does not really get a tag that starts with some prefix (as the name suggests), but rather the tag associated with some key that starts with the given prefix. anyway, since this method is no longer needed, let's remove it ?

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -510,6 +511,21 @@ public boolean canSkip() {
         }
     }
 
+    /**
+     * Determine whether a relation is routable for one of the added encoders.
+     *
+     * @return if at least one encoder consumes the specified relation. Additionally the specified acceptRelation is changed
```
easbar: This javadoc is outdated I think ? Something like

```java
/**
 * @return true if any of the encoders accepts the given turn relation
 * @see FlagEncoder#acceptRelation(OSMTurnRelation)
 */
```

would be enough. Or just remove the entire method in case its not needed.

core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
```diff
@@ -53,6 +54,14 @@
      */
     DecimalEncodedValue getAverageSpeedEnc();
 
+    /**
+     * Decide whether a restriction relation is enabled for a given mode of travel. This skips some relations before
+     * handleRelationTags is called.
+     *
+     * @return true if this encoder allows travel.
+     */
+    boolean acceptsRelation(OSMTurnRelation relation);
```
easbar: Maybe now that this one is only about turn restrictions, lets call it `acceptsTurnRelation` ?

core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
```diff
@@ -53,6 +54,14 @@
      */
     DecimalEncodedValue getAverageSpeedEnc();
 
+    /**
+     * Decide whether a restriction relation is enabled for a given mode of travel. This skips some relations before
```
easbar: How about

```java
/**
 * @return whether or not this encoder should apply the given turn restriction
 * @see OSMTurnRelation
 */
```

easbar:
Did you sign the contributor agreement yet ? https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md#license-agreement

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
sguill: If I remove the test `turnRelation == null`, `OSMReaderTest#testTurnFlagCombination` fails.
There are 2 entries expected from the call:
```
Collection<OSMReader.TurnCostTableEntry> entries = reader.analyzeTurnRelation(null);
// we expect two different turnCost entries
assertEquals(2, entries.size());
```
Without the test, `entries` is empty, so the assert fails.

docs/core/turn-restrictions.md
sguill: ok

core/src/main/java/com/graphhopper/reader/ReaderElement.java
sguill: ok, I remove it.

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
sguill: I remove it too.

core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
sguill: ok

core/src/main/java/com/graphhopper/routing/util/FlagEncoder.java
sguill: ok

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: Ok, yes, but `turnRelation` should not be null here in the first place. In fact it can never be null in production (because of the null check right after `createTurnRelation`. In `OSMReaderTest#testTurnFlagCombination` we explicitly pass null, and in fact this test uses some customized `analyzeTurnRelation` method that does not even us the `turnRelation`. So I think we should remove the null check and fix the test by passing just some `OSMTurnRelation` instance in `testTurnFlagCombination` instead of null.

easbar: Yeah, thanks for your work! This looks good.

easbar:
Thanks!

PR #1691, opened 2019-08-21
Keep Expand State Elevation Diagram

### LINKED ISSUE #1658: Do not show height graph again if closed
(for desktop)

reported in #1654 from @easbar 

> Just a minor issue, but if I close the height graph (say because I want to see more of the map) and then drag one of the route markers it pops up again.

boldtrn:
Fixes #1658 

As discussed, the elevation diagram remembers it's state in a request variable and will be expandable even on mobile devices. 

I also updated some js dependencies and fixed a minor issue when showing boolean PathDetails. For details with value=false, the former solution showed "Undefined", now it correctly shows "false".


karussell:
Thanks!

> the former solution showed "Undefined", now it correctly shows "false".

:+1: 

> the elevation diagram remembers it's state in a request variable 

Wouldn't it be better to avoid yet another parameter as discussed in #1658?

> > Better just temporary in some JavaScript object to avoid complexity.

> Sure we can certainly do that as well +1

boldtrn:
> Wouldn't it be better to avoid yet another parameter as discussed in #1658?

I think I missunderstood the following, that you wanted a parameter :):

> and somehow storing the state (in the elevation parameter?)

Should I remove it again?

karussell:
> and somehow storing the state (in the elevation parameter?)

Ah, ok. Was using the initial way it was described in the issue - obviously this was confusing :)

> Should I remove it again?

Yes, I think the link should not store the device size indirectly.

boldtrn:
> Yes, I think the link should not store the device size indirectly.

There should only be a setting in the URL if the user explicitly toggles the diagram (either opens or closes it).

karussell:
Yes. But why should we put this state in the URL? I see it as a personal setting that you do not want to share (btw: use_miles and locale are maybe a similar thing but that is for another issue ;) ).

boldtrn:
If you reload the website, the setting is still active. Or if you save this as bookmark, the setting is still active is as well. But I can remove it, that's easy :). I don't have a strong opinion on that.

karussell:
> If you reload the website, the setting is still active

Yes, I think this is ok

boldtrn:
I updated the PR, the expand state is now kept as a js variable. Also I updated the version of the elevation diagram to the latest version, there were some improvements, including a slightly improved style.

karussell:
It does not seem to fix #1658 although it now just hides if screen is too small.

Additionally a black box is shown instead of the properties:

image

boldtrn:
> Additionally a black box is shown instead of the properties

That's probably a caching problem, die you fully reload the website and rebuild the JS?

> It does not seem to fix #1658 although it now just hides if screen is too small.

For me it stays closed, once I closed it, until I reopen it, then it stays open. Could you provide some more details to reproduce this?

karussell:
> That's probably a caching problem

I'm using the private mode of firefox. What else could make a problem?

> die you fully reload the website and rebuild the JS?

Yes and I also tried `npm run watch` and tried it without jetty on GraphHopper Maps (it is still there to try and see it)

web/package.json
```diff
@@ -27,15 +27,15 @@
     "leaflet.vectorgrid": "1.3.0",
     "browserify": "16.2.0",
     "browserify-swap": "0.2.2",
-    "d3": "5.9.7",
+    "d3": "5.11.0",
     "jquery": "3.4.1",
-    "leaflet": "1.3.1",
+    "leaflet": "1.5.1",
```
karussell: I think also the css and index.html needs an update if leaflet js is updated

boldtrn: Ah good point, there was indeed a minor css change, I updated this PR.

boldtrn:
> Yes and I also tried npm run watch and tried it without jetty on GraphHopper Maps (it is still there to try and see it)

Now I know what the issue could be :). You need to run `npm install`. Did you do that?

karussell:
Oh, no. This is ugly. Will create an issue for it.

Thanks for this PR - works perfectly!

PR #1696, opened 2019-08-26
Throw error if edge_based=true, but encoder does not support turn costs

### LINKED ISSUE #1695: Edge-based routing without turn cost supporting encoder
When we specify `edge_based=true` but the encoder does not have `|turn_costs=true` we get an edge-based algorithm, but the weighting will not be wrapped by `TurnWeighting`. U-turns will then not be prevented properly. 

Should we throw an error if `edge_based=true`, but `!encoder.supports(TurnWeighting.class)` ?

easbar:
Fixes #1695.


karussell:
Thx!

PR #1697, opened 2019-08-26
Add new 'curb_side' parameter, allow specifying start/target edges

### LINKED ISSUE #680: Force side of the street
Implement a new feature to support right side pick-up.

### LINKED ISSUE #1647: Allow restricting start/target edge for edge-based routing.

### LINKED ISSUE #1694: Way geometry can have duplicate (pillar) nodes
I encountered some cases where the way geometry contained duplicate points (so basically 'loops' within the geometry of an edge). I could work around this somehow, but at the same time I am wondering whether or not this should even be allowed. Maybe make sure this does not happen during import already ?

easbar:
This PR adds a new `curb_side` parameter to the `/route` endpoint. The `curb_side` specifies on which side a query point should be relative to the driver when she leaves/arrives at a start/target/via point. This feature is similar to `heading`, but does not suffer from (all) the drawbacks mentioned in #1647, importantly it works with speed-mode/CH. However, using this parameter requires using `edge_based=true` (so for example to use this with speed-mode/CH one needs to setup the graph preparation accordingly).

Take this example of a road going from West to East with some location x south of the road and another location y north of the road:

```java
    y
-----------
       x
```

The new parameter then has the following meaning:

|location|type|curb_side=right|curb_side=left|
|-|-|-|-|
|x|start|start towards E|start towards W|
|x|via|arrive from W, leave towards E|arrive from E, leave towards W|
|x|end|arrive from W|arrive from E|
|y|start|start towards W|start towards E|
|y|via|arrive from E, leave toward W|arrive from W, leave towards E|
|y|end|arrive from E|arrive from W|

You need to either specify the `curb_side` for all points of the routing query or for none of them. `curb_side=either` can be used to not specify the direction of arriving/leaving a point (in this case the routing engine will choose the shortest path considering all options).

Note that this definition of `curb_side` is independent of right/left-hand traffic in different countries, `curb_side=right` always means the location will be on the right of the driver (in driving direction), no matter whether she is sitting on the right/left of the vehicle and/or driving on the right/left lane.

There are some corner cases where the `curb_side` parameter will be ignored, for example when the query point lies on a junction, see `DirectionResolverTest.java` for more details.

Fixes #1647, fixes #680.

Current todo:

* [x] Find a solution for #1694 first (or stick with the workaround here) (the workaround is fine for now)
* [x] Fall back to `curb_side=either` in case the requested curb side is not possible (e.g. in one-way streets), or make this configurable. this is now configurable.


devemux86:
Users know only their heading in degrees (e.g. via GPS bearing) and can pass that conveniently to all routing engines in order to ask for a directed start during route calculation.

> This feature is similar to heading, but does not suffer from (all) the drawbacks mentioned in #1647, importantly it works with speed-mode/CH.

Is it an extra feature or the heading will be deprecated in the future?

easbar:
> Users know only their heading in degrees (e.g. via GPS bearing) and can pass that conveniently to all routing engines in order to ask for a directed start during route calculation.

Ok I see for a user with GPS it is convenient to pass on the heading in degrees. But for route planning (looking at a map) isn't it also convenient to be able to specify on which side of the road some location is ?

> Is it an extra feature or the heading will be deprecated in the future?

These features can co-exist. In the long run it would be nice to implement heading similarly to curb-side (specifying some heading (in degrees), extracting the relevant edge-ids and initializing the routing algorithm with these edges). For node-based routing I think the current implementation of heading would still be required.

easbar:
> But working with code, all that is available in clients are waypoints in coordinates and their headings in degrees. Low level parameters or conversions are not known in apps which are unaware of graph internals.

This depends on the kind of code/application/client as well (?). For example using `curb_side` it is possible to just set `curb_side=right` for all points yielding a route that reaches all points such that the point will be on the right side of the driver, without having to think about headings in degrees, c.f. #680.

devemux86:
> For example using curb_side it is possible to just set curb_side=right for all points yielding a route that reaches all points such that the point will be on the right side of the driver, without having to think about headings in degrees, c.f. #680.

That seems related more to via points and destination?
Route's start seems a different case, as shouldn't have to know if start on a street which is oneway or not, has multiple lanes, etc., instead just ask to start routing (if possible) towards a specific heading.

easbar:
> That seems related more to via points and destination?

Yes, you can use `curb_side` for all points (start, via, destination).

devemux86:
> Yes, you can use curb_side for all points (start, via, destination).

More meant that cannot see it as usable in place of start's heading.
Users cannot convert it to left / right without graph data knowledge.

easbar:
> More meant that cannot see it as usable in place of start's heading.
Users cannot convert it to left / right without graph data knowledge.

Ok `curb_side` was not meant to replace heading. But even for the start direction I do not understand what graph data knowledge is needed. All the user needs to know is whether the start location should be left or right of him ? Or are you asking to add the extension of this feature I mentioned above (take heading in degrees and determine edge ids) to this PR ? I would rather open another issue for this.

devemux86:
I'm more concerned if heading feature will be removed from GraphHopper in the future.
That will make it practically unusable in navigation (we use GraphHopper in [Kurviger](https://kurviger.de/)).

> All the user needs to know is whether the start location should be left or right of him ?

Vehicles on the road don't see map to know left/right, navigator does the (re)routing.
And it can only know what GPS reports, that is location's coordinates and heading.
That data is passed in routers so far and they calculate route towards that heading.

easbar:
> I'm more concerned if heading feature will be removed from GraphHopper in the future.

Ok thanks for your explanations and reminder what heading is used for. Of course it will not be removed, and if it will be re-implemented we have to make sure it can be used as before (but I do not think anyone is planning to change anything about the heading feature atm).

easbar:
Here is a little illustration of this parameter (the same route calculation for different values of `curb_side` for the start/target points).

`curb_side=left,curb_side=left`:
image

`curb_side=left,curb_side=right`:
image

`curb_side=right,curb_side=left`:
image

`curb_side=right,curb_side=right`:
image

easbar:
Ok, the only thing that I am not sure of here is what should happen when a `curb_side` is specified that can not be realized (like in a one-way) street.

1) connection not found error (this is what happens currently)
2) ignore the curb_side constraint
3) add another parameter that lets us choose between 1)&2)

For 1) and 2) it would be nice to know which curb_side constraint could not be fulfilled, maybe this information needs to be added to the response ? This is still missing.

easbar:
Ok, to make progress I decided to add a new parameter that controls the behavior: `ignore_impossible_curb_side`. So in case a curb_side constraint can not be fulfilled we get an error telling us at which point the constraint failed **unless** this parameter is set to true in which case the constraint ist ignored.

karussell:
I think option 3 is the best - the question is now, which of 1+2 should be default :) ? As we are in other matter rather strict it would be consistent to choose 1, but I have really no preference.

BTW: can or shall we rename `ignore_impossible_curb_side` to the shorter `force_curb_side` or `curb_side_force`? (which would also negate its meaning)

easbar:
Ok then I'll go with `force_curb_side`, and use `true` as default.

easbar:
image

PR #1703, opened 2019-08-28
New MaxLengthParser

otbutz:
* add support for the [maxlength](https://wiki.openstreetmap.org/wiki/Key:maxlength) tag
* moved shared value parsing logic from OSMMaxWidthParser to `com.graphhopper.routing.util.parsers.helpers.OSMDimensionExtractor`


otbutz:
I think we should combine them. How about `OSMValueExtractor` ?

karussell:
Sounds good!

karussell:
Thanks!

PR #1704, opened 2019-08-28
Add support for hgv toll sub-categories

otbutz:
This PR modifies the OSMTollParser to treat `toll:N2` and `toll:N3` as special cases of `toll:hgv`.

[Wikipedia](https://en.wikipedia.org/wiki/Large_goods_vehicle):
> A heavy goods vehicle (HGV), also large goods vehicle (LGV) or medium goods vehicle, is the European Union (EU) term for any truck with a gross combination mass (GCM) of over 3,500 kilograms (7,716 lb). Sub-category N2 is used for vehicles between 3,500 kilograms (7,716 lb) and 12,000 kilograms (26,455 lb) and N3 for all goods vehicles over 12,000 kilograms (26,455 lb) as defined in Directive 2001/116/EC.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTollParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTollParser.java
@@ -46,6 +46,10 @@ public class OSMTollParser implements TagParser {
             tollEnc.setEnum(false, edgeFlags, Toll.ALL);
         else if (readerWay.hasTag("toll:hgv", "yes"))
             tollEnc.setEnum(false, edgeFlags, Toll.HGV);
+        else if (readerWay.hasTag("toll:N2", "yes"))
+            tollEnc.setEnum(false, edgeFlags, Toll.HGV);
+        else if (readerWay.hasTag("toll:N3", "yes"))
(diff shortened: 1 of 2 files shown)


karussell:
Should we introduce those two subcategories instead of hiding them behind hgv? Or did you avoid this because they don't have a meaning outside of the EU?

otbutz:
I thought about introducing them as separate entries but i decided otherwise because

1. they're EU only (not really that problematic on its own IMHO)
2. they add more complexity
3. the `toll:N2`([taginfo](https://taginfo.openstreetmap.org/keys/toll:N2#overview): 714 ways) which describes a "smallish HGV" isn't used really often compared to `toll:N3`([taginfo](https://taginfo.openstreetmap.org/keys/toll:N3#overview): 19910 ways) which describes a regular truck with 12t or more
4. we'll never reach 100% toll model coverage because there are to many factors involved: load, axles, emission class, etc

I'd say we should keep it simple and geared towards the most common use cases. It's fine grained enough to make proper routing decisions for car/truck routing and to get an estimation for toll mileage.

karussell:
Thanks, that are good reasons :)

Issue #1702, opened 2019-08-28
Edge-based routing without turn-costs


easbar:
Discussion came up in #1698.

easbar:
I think ideally we would have:

* `traversal_mode=edge_based/node_based` (or `edge_based=true/false` ?) to determine the traversal mode, default: node. I think I would prefer `traversal_mode`, because this way we could throw an error if somebody still uses `edge_based`.
* `turn_costs=true/false` to enable/disable turn costs, default: false

Maybe we need to make the defaults for `traversal_mode` and `turn_costs` configurable via `config.yml` ?

|case|traversal_mode|turn_costs|expected behavior|
|-|-|-|-|
|1|node_based|false|node-based routing, no turn costs (and no u-turns)|
|2|node_based|true|enable `traversal_mode=edge_based` automatically|
|3|edge_based|false|edge-based routing, no turn costs (u-turns are free) **unless** CH is enabled in this case this should be an error because there will only be an edge-based CH profile **with** turn costs !? (see below)|
|4|edge_based|true|edge-based routing with turn costs and default u-turn costs (infinity/no u-turns)|

So from the user perspective the most common cases would be:

- specify nothing -> node based, no turn costs, no u-turns
- set `turn_costs=true` -> edge based, with turn costs, no u-turns

For expert users:

- set `traversal_mode=edge_based` -> edge-based routing, no turn costs, free u-turns

*After* this selection is made we check (do not use the vehicle or encoding manager to 'derive' the values for edge_based/turn_costs as is done currently):

* if `turn_costs=true`: does the selected vehicle/encoder support turn costs (i.e. is there a turn cost extension for this encoder) ? if not throw an error.
* is there a CH profile for the selected vehicle and the chosen traversal mode ? if not throw an error. however see #1637...
* if an edge-based CH profile is selected and `turn_costs=false` throw an error (or should this be possible as well (requires edge-based CH preparation without turn costs and keeping track if CH preparation included turn-costs or not (in `CHProfile`)).

Issue #1637, opened 2019-06-18
If only edge based CH is prepared no error should be thrown


karussell:
When a route is done the 'edge_based' url parameter should not be required. Currently you get:
```
Caused by: java.lang.IllegalArgumentException: Cannot find CH RoutingAlgorithmFactory for weighting map {weighting=fastest, vehicle=car} in entries: [fastest|car|edge, faste .... The 'edge_based' url parameter is missing or does not fit the weightings. Its value was: 'false'
        at com.graphhopper.routing.ch.CHAlgoFactoryDecorator.getPreparation(CHAlgoFactoryDecorator.java:266)
        at com.graphhopper.routing.ch.CHAlgoFactoryDecorator.getDecoratedAlgorithmFactory(CHAlgoFactoryDecorator.java:246)
        at com.graphhopper.GraphHopper.getAlgorithmFactory(GraphHopper.java:760)
```

> @easbar Yes, maybe we can:
>
> - always fall back to edge based if there really is only one edge-based CH and no node based CH (for any weighting)
>
> - add a config value that changes the default value of `edge_based`

easbar:
`GraphHopper` already has a `traversalMode` field that is used as default for routing requests. I think this is also what it should be called: `defaultTraversalMode`, so this is what I will call it here. It can be set via `setTraversalMode()` and is automatically set by `setEncodingManager()` if the `EncodingManager` needs turn cost support.

The default can be overwritten on a per-request basis by the `edge_based` parameter, but also by the `traversal_mode` parameter (`traversal_mode` is overwritten by `edge_based`):

https://github.com/graphhopper/graphhopper/blob/bf0c7b40fa5694672f79c835244f29c0e369f04b/core/src/main/java/com/graphhopper/GraphHopper.java#L977-L980

The `traversal_mode` parameter definitely seems redundant (at least after #1640). Can we remove it ? I did not find documentation for it either.

To me it also seems a bit strange that the default `traversalMode` is `edge-based` as soon as `encodingManager.needsTurnCostSupport() == true` (so as soon as one encoder needs turn cost support). 
Would it not be much more obvious to check if the encoder that is selected by the `vehicle` parameter supports turn costs and if yes use edge-based routing (unless `edge_based` is given explicitly with the request) ? I mean here:

https://github.com/graphhopper/graphhopper/blob/bf0c7b40fa5694672f79c835244f29c0e369f04b/core/src/main/java/com/graphhopper/GraphHopper.java#L982

[...]

PR #1705, opened 2019-08-28
Do not consider turn cost support of other encoders when selecting traversal mode

### LINKED ISSUE #1698: Route request fails if edge_based is not set to false and there is a turn cost encoder
If we use multiple encoders out of which one has `turn_costs=true`, e.g. `foot,car|turn_costs=true`,  we have `EncodingManager#needsTurnCostsSupport()===true` and the `GraphHopper#traversalMode` field will be `EDGE_BASED` (see `GraphHopper#setEncodingManager`). This `traversalMode` field is then used to determine the traversal mode in `GraphHopper#calcPaths`.

It *can* be overwritten using the `edge_based` or the `traversal_mode` parameters, but since these are optional it might remain `EDGE_BASED` even when using a vehicle without turn costs (like `foot`). So we use edge-based routing for foot, just because there is *another* encoder with `turn_costs=true`. 

This used to be odd behavior, but with #1696 this is now even more problematic because we now get an exception in this case.

I think we should remove the `GraphHopper#traversalMode` field entirely, because it seems very counter-intuitive to determine the 'default' traversal mode based on the presence of *some* encoder that supports turn costs.

Related: #1637

easbar:
This is supposed to be a minimal fix to #1698.

Instead of deriving an 'overall' default traversal mode for *all* encoders every request checks whether or not the selected encoder supports turn costs and selects the traversal mode based on this (in case no edge_based parameter is sent along with the request).

Fixes #1698, but also related to #1637 and #1702.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -96,7 +96,6 @@ public class GraphHopper implements GraphHopperAPI {
     // for routing
     private int maxRoundTripRetries = 3;
     private boolean simplifyResponse = true;
-    private TraversalMode traversalMode = TraversalMode.NODE_BASED;
     private int maxVisitedNodes = Integer.MAX_VALUE;
 
     private int nonChMaxWaypointDistance = Integer.MAX_VALUE;
@@ -166,9 +165,6 @@ public class GraphHopper implements GraphHopperAPI {
     public GraphHopper setEncodingManager(EncodingManager em) {
         ensureNotLoaded();
         this.encodingManager = em;
-        if (em.needsTurnCostsSupport())
-            traversalMode = TraversalMode.EDGE_BASED;
-
         return this;
     }
 
@@ -205,18 +201,6 @@ public class GraphHopper implements GraphHopperAPI {
      */
     public GraphHopper setWayPointMaxDistance(double wayPointMaxDistance) {
         this.dataReaderWayPointMaxDistance = wayPointMaxDistance;
-        return this;
-    }
-
-    public TraversalMode getTraversalMode() {
-        return traversalMode;
-    }
-
-    /**
-     * Sets the default traversal mode used for the algorithms and preparation.
-     */
-    public GraphHopper setTraversalMode(TraversalMode traversalMode) {
-        this.traversalMode = traversalMode;
         return this;
     }
 
@@ -972,13 +956,15 @@ public class GraphHopper implements GraphHopperAPI {
             if (!encodingManager.hasEncoder(vehicle))
                 throw new IllegalArgumentException("Vehicle not supported: " + vehicle + ". Supported are: " + encodingManager.toString());
 
+            FlagEncoder encoder = encodingManager.getEncoder(vehicle);
             HintsMap hints = request.getHints();
-            String tModeStr = hints.get("traversal_mode", traversalMode.toString());
-            TraversalMode tMode = TraversalMode.fromString(tModeStr);
+
+            TraversalMode tMode = encoder.supports(TurnWeighting.class) ? TraversalMode.EDGE_BASED : TraversalMode.NODE_BASED;
+
+            String tModeStr = hints.get("traversal_mode", tMode.toString());
+            tMode = TraversalMode.fromString(tModeStr);
             if (hints.has(Routing.EDGE_BASED))
                 tMode = hints.getBool(Routing.EDGE_BASED, false) ? TraversalMode.EDGE_BASED : TraversalMode.NODE_BASED;
-
-            FlagEncoder encoder = encodingManager.getEncoder(vehicle);
 
             if (tMode.isEdgeBased() && !encoder.supports(TurnWeighting.class)) {
                 throw new IllegalArgumentException("You need a turn cost extension to make use of edge_based=true, e.g. use car|turn_costs=true");
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -960,7 +960,7 @@ public class GraphHopper implements GraphHopperAPI {
             HintsMap hints = request.getHints();
 
             // we use edge-based routing if the encoder supports turn-costs *unless* the edge_based parameter is set
-            // explicitly
+            // explicitly.
             TraversalMode tMode = encoder.supports(TurnWeighting.class) ? TraversalMode.EDGE_BASED : TraversalMode.NODE_BASED;
             if (hints.has(Routing.EDGE_BASED))
                 tMode = hints.getBool(Routing.EDGE_BASED, false) ? TraversalMode.EDGE_BASED : TraversalMode.NODE_BASED;
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -959,10 +959,9 @@ public class GraphHopper implements GraphHopperAPI {
             FlagEncoder encoder = encodingManager.getEncoder(vehicle);
             HintsMap hints = request.getHints();
 
+            // we use edge-based routing if the encoder supports turn-costs *unless* the edge_based parameter is set
+            // explicitly
             TraversalMode tMode = encoder.supports(TurnWeighting.class) ? TraversalMode.EDGE_BASED : TraversalMode.NODE_BASED;
-
-            String tModeStr = hints.get("traversal_mode", tMode.toString());
-            tMode = TraversalMode.fromString(tModeStr);
             if (hints.has(Routing.EDGE_BASED))
                 tMode = hints.getBool(Routing.EDGE_BASED, false) ? TraversalMode.EDGE_BASED : TraversalMode.NODE_BASED;
 
--- a/core/files/changelog.txt
+++ b/core/files/changelog.txt
@@ -17,6 +17,7 @@
     CHGraph.shortcut(int, int) was removed (use .shortcut(int, int, ...) and/or .shortcutEdgeBased(int, int, ...) instead, #1693
     CH graphs are now identified using CHProfile instead of Weighting, #1670
     removed the 'traversal_mode` request parameter for /route, instead of 'traversal_mode=edge_based_2dir' use edge_based=true
+    removed GraphHopper.set/getTraversalMode() methods, #1705
 
 0.12
     renamed VirtualEdgeIteratorState.getOriginalEdgeKey to more precise getOriginalEdgeKey #1549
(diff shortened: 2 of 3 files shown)


easbar:
I think the thing that now bothers me the most is the asymmetry between CH and non-CH: For CH we always use node-based **unless** `edge-based=true` is set explicitly. For non-CH we use edge-based as soon as the encoder has `|turn_costs=true` (and we can go back to node-based by explicitly setting `edge_based=false`). This also means that with the same settings we might get turn cost support in the one case and no turn cost support in the other (when switching CH on/off).

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -972,14 +956,16 @@ public GHResponse route(GHRequest request) {
...
                 throw new IllegalArgumentException("Vehicle not supported: " + vehicle + ". Supported are: " + encodingManager.toString());
 
+            FlagEncoder encoder = encodingManager.getEncoder(vehicle);
             HintsMap hints = request.getHints();
-            String tModeStr = hints.get("traversal_mode", traversalMode.toString());
-            TraversalMode tMode = TraversalMode.fromString(tModeStr);
+
+            TraversalMode tMode = encoder.supports(TurnWeighting.class) ? TraversalMode.EDGE_BASED : TraversalMode.NODE_BASED;
+
+            String tModeStr = hints.get("traversal_mode", tMode.toString());
```
easbar: Can I remove this `traversal_mode` parameter to avoid further confusion ? Right now it almost does the same as the `edge_based` parameter and I did not find documentation for it anywhere ?

karussell: Yes, let's do this.

karussell:
Should we add this to changes file and mention that edge_base parameter is/was equivalent?

karussell:
Hmmh, this PR is actually a breaking change for all users that do not yet support edge-based routing for CH but have turn_costs enabled.

easbar:
> this PR is actually a breaking change f

is it ? in which case ?

karussell:
Maybe I have misunderstood something :D 

Assume I have one vehicle car with turn costs and I have (node-based) CH enabled. 

(1) Then request from A to B is per CH (node based) and

(2) if I disable CH it is edge based with turn costs. 

No intuitive, yes. But this was the best we could do for several years :)

Now, with this change, wouldn't the edge based algorithm be selected for CH case (1) and fail if no additional parameter edge_based=false is specified?

easbar:
> Now, with this change, wouldn't the edge based algorithm be selected for CH case (1)

No, I do not think so, but let me add a test to make sure. This of course is an important case.

easbar:
Is this the case you had in mind ? https://github.com/graphhopper/graphhopper/pull/1705/commits/c4581490d2fda0691f63d29851d5fab49199db8a

I think it is fine. Why ? Because `tMode` is determined to be `edge_based` at the beginning of `GraphHopper#calcPaths`, but the CH preparation is later selected solely by the `edge_based` parameter... (also ugly :))

easbar:
Build failed with 
```shell
A[PI limit reached. Reset in 14699 seconds]>
```
anyway, I will re-trigger it again tomorrow

easbar:
@michaz the `GraphHopper#getTraversalMode` method is used in map-matching (in `MatchingResource` and `MatchCommand`), but can be replaced with 

```java
graphHopper.getEncodingManager().needsTurnCostsSupport() ? TraversalMode.EDGE_BASED : TraversalMode.NODE_BASED
```
?!

Also added this to changelog

michaz:
Yes! Thank you! I would also be happy with just _always_ using EDGE_BASED, if it can do strictly more than NODE_BASED. The less combinatorics, the better. (That's why I was so sad at the idea that I could get an exception by choosing EDGE_BASED if some part further down thinks I don't _need_ EDGE_BASED).

easbar:
>  if it can do strictly more than NODE_BASED.

unfortunately not yet: heading (discussion in #1697), alternative routes (#1524), block edges/areas + its slower and most of all uses a lot more memory (especially for planet-wide)

> The less combinatorics, the better.

I can tell ...

> That's why I was so sad at the idea that I could get an exception by choosing EDGE_BASED if some part further down thinks I don't need EDGE_BASED

Well if you choose edge_based but the server is not prepared for it its better to get a warning than not ? But I also agree with your kind of thinking (choose behavior not algorithm) so if you choose turn_costs you should not get a warning (just a slow algorithm...)

michaz:
Sometimes I think half of the configuration confusion (not the implementation limitations) is just because a handful of things are historically inside-out or the wrong way round.

Like Alternatives: It's not only a different "algorithm", it's a different _question_ to ask. On that layer, it has as little to do with regular shortest-path routing as map-matching.

I think the implementation of Alternatives should just _use_ node-based routing if it knows it can't do it differently. And not rely on the outer layer to find out for it that node-based isn't configured when Alternatives is configured.

michaz:
EDIT: 

> "just"

easbar:
Regarding combinatorics, look at this: https://github.com/graphhopper/graphhopper/blob/88c9c7ceeb86304a2d693c2bde53058b5eedb7fa/core/src/main/java/com/graphhopper/routing/ch/PCHSelector.java#L34

This is already starting to explode (and the more flexible everything shall become the worse it is going to get). Right now the only way I see out of this mess is to:

1) assign IDs to 'prepared profiles' that have to be passed back to the engine to specify what profile should be used. if this was on the top layer thats would be the opposite extreme of what you describe with "let user pick behavior not implementation". however, I think we need put the  "select profile based on user needs" logic *in front of this*. But for this to work we need:

2) create some way to determine what user needs a profile can support (like 'does it support turn-costs' ? 'can I use this value of u-turn costs for it' ? 'does it allow heading' ? 'is it the fastest possible' ?)

So eventually we would first get a 'wish-list' from the user, then scan the available (CH) profiles and check if there is one that supports everything that was requested by the user. If there is no such profile fall back to some non-CH algorithm that supports everything. However, we still might need some low-level flags for debugging/testing etc. like ('do not fall back to dijkstra', 'enforce this algorithm' etc.)

easbar:
> Sometimes I think half of the configuration confusion (not  the implementation limitations) is just because a handful of things are  historically inside-out or the wrong way round.

I do not know the details of the alternative route implementations, but yes sometimes this might be the problem. Identifying the places where this confusion arises will sure help.

easbar:
> Maybe one day, a "Grand Redesign in the Sky" [Clean Code].

I think we should try to improve the current state rather than doing this, but for the flexible CH scenario we are still at the very beginning so its good to think about what would be 'ideal' (and then check how far this is from the current state).

michaz:
To get this to a point where we say "that's it!" would be a major project, including finding a terminology that we all agree on.

To me, "profile" means "cost function", i.e. "how much does a path cost, including everything", i.e. "what is the thing I want to minimize".

So it _defines_ things like turn-costs, U-turn restrictions, other hard constraints, everything. Together with input points, headings etc., it's the optimization problem to solve. (Not to be confused with "optimization" as in tour planning.)

That's _almost_ the same as your `CHProfile`, right? Except that when the required `CHProfile` isn't there, we can't use CH but can still route?

michaz:
And then there's an Algorithm, which is a thing that solves the optimization problem P(Profile, Input points). It must _support_ what the Profile _requires_.

easbar:
> That's almost the same as your CHProfile, right?

Yes, I think so. And the reason for this is that almost everything that defines a 'profile' needs to be known at preparation time (for CH). For me a CHProfile is the set of conditions that I fixed when I ran the preparation.

> Except that when the required CHProfile isn't there, we can't use CH but can still route?

Yes when a request comes in it defines the/your 'profile', then we have to check if there is a ch-profile that matches it, otherwise we cannot use CH

michaz:
Nice. Almost there then.

michaz:
But can't this be simplified _enormously_ if we just give every profile that a GraphHopper instance supports (not only every CHProfile; every profile) a name in the configuration file?

Then, at start time, we prepare the best thing that we can prepare for that profile, remember to use it, and that's it?

michaz:
```
carWithTurnCost: {
  linkCostSpecification: {...},
  turnCostSpecification: {...},
  ch: [default] true,
  lm: [default] false,
  extraHeadingRules: [default] false
}
```

And then at the _totally_ outermost layer, like in the web service, we pick one of these based on the query parameters. Or if we really want to over-engineer, put the rules for mapping query params to profile names in the config file. :-D

michaz:
But maybe that's almost what we are doing, except for the details, idk.

michaz:
Perhaps the secret blocker here has been that we don't have hierarchical configuration files in the core yet.

karussell:
> Is this the case you had in mind ? c458149

Yes, thanks! Perfect!

easbar:
Hmm, many questions about profile configuration here :) Closing this here, maybe we can continue in issues dedicated to the discussion about the configuration etc. ? For example I opened #1708.

easbar:
> But can't this be simplified enormously if we just give every profile that a GraphHopper instance supports (not only every CHProfile; every profile) a name in the configuration file?

@michaz The issue I am having with this is that the profiles may contain parameters. For example the `ShortFastestWeighting` has a `distanceFactor`, the `DefaultTurnCostProvider` has `uTurnCosts` (and in the future hopefully will have `leftTurnCosts`, for different roads etc.). There could be `blocked_areas` as well etc. In other words there is no complete discrete set of profiles we can specify in advance (and give a name to each). 

Even if we gave a name to every profile (which otherwise sounds very promising to me) the difficulty would still be mapping the requests onto these profiles. Sure, at first sight this seems easy: As @karussell suggested we could maintain the order of the profiles and if for example `vehicle=car` is requested we search the list of profiles from top to bottom for profiles with `vehicle=car` and use the first match. Same with `weighting` for example. But what if `weighting=short_fastest` and `distance_factor=0.3` (but `distance_factor=0.15` was used for the preparation)? Is this a match? How do we even check whether it is a match or not? Which of the request parameters are we supposed to check? Its the same with turn costs: Without CH or even with LM its very easy (in theory) to calculate a single route with different configurations of left/right/u-turn costs etc. (its easy to imagine a weighting (or turn-weighting with tons of parameters), so this is something we should (be able to) support. But as soon as we support this and have all these configuration options how are we supposed to map this onto the CH profiles? Right now we are doing this for edge-weights (#1776) and the answer seems easy in this case: Whenever a `CustomModel` is sent with the request we just do not use CH. But this is not a good solution for turn costs (or `distance_factor` with `ShortFastestWeighting`) where there might be a prepared CH profile, but its just hard to find out if the request matches it. 

> And then at the totally outermost layer, like in the web service, we pick one of these based on the query parameters

I like the idea of moving this to the outermost layer, but doing the mapping (and this seems like the real problem to me) does not become easier this way (in fact I think it might be harder to do it like that because there might be less information about the profiles at this point?).

Maybe we can even remove all the request parameters that (together) make up the routing profile (currently vehicle,weighting,turn_costs,u_turn_costs) and replace them with a single `profile` parameter instead?

PR #1706, opened 2019-08-28
Choose edge-based CH if it was prepared and the `edge_based` parameter was not set

### LINKED ISSUE #1637: If only edge based CH is prepared no error should be thrown

easbar:
This fixes #1637 and is based on #1705.

When we are determining the traversal mode and when the `edge_based` parameter is missing, we check what CH preparations are there (instead of looking at the turn cost support of the *encoder* as for non-CH).

So with this change we do no longer need to set `edge_based=true` to get edge-based CH and instead edge-based CH is picked automatically if it was prepared (if not we fall back to node-based CH). If `edge_based` is set on the request its value is respected (and there will be an error if the specified preparation is missing).

Note that with this PR we now choose edge-based CH when there are node *and* edge-based CH preparations for the chosen weighting **and** the `edge_based` parameter was not set.

### DIFF
--- a/core/files/changelog.txt
+++ b/core/files/changelog.txt
@@ -16,6 +16,7 @@
     GraphHopperStorage.getGraph(Class) was replaced by GraphHopperStorage.getBase/CHGraph(), #1669
     CHGraph.shortcut(int, int) was removed (use .shortcut(int, int, ...) and/or .shortcutEdgeBased(int, int, ...) instead, #1693
     CH graphs are now identified using CHProfile instead of Weighting, #1670
+    removed the 'traversal_mode` request parameter for /route, instead of 'traversal_mode=edge_based_2dir' use edge_based=true
 
 0.12
     renamed VirtualEdgeIteratorState.getOriginalEdgeKey to more precise getOriginalEdgeKey #1549
--- a/core/files/changelog.txt
+++ b/core/files/changelog.txt
@@ -18,7 +18,7 @@
     CH graphs are now identified using CHProfile instead of Weighting, #1670
     removed the 'traversal_mode` request parameter for /route, instead of 'traversal_mode=edge_based_2dir' use edge_based=true
     removed GraphHopper.set/getTraversalMode() methods, #1705
-    removed the 'traversal_mode` request parameter for /route, instead of 'traversal_mode=edge_based_2dir' use edge_based=true
+    edge-based CH is now chosen by default if it was prepared, #1706
 
 0.12
     renamed VirtualEdgeIteratorState.getOriginalEdgeKey to more precise getOriginalEdgeKey #1549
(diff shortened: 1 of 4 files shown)


core/files/changelog.txt
```diff
@@ -18,6 +18,7 @@
     CH graphs are now identified using CHProfile instead of Weighting, #1670
     removed the 'traversal_mode` request parameter for /route, instead of 'traversal_mode=edge_based_2dir' use edge_based=true
     removed GraphHopper.set/getTraversalMode() methods, #1705
+    removed the 'traversal_mode` request parameter for /route, instead of 'traversal_mode=edge_based_2dir' use edge_based=true
```
karussell: I think this is a merge conflict and duplicated?

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
```diff
@@ -251,23 +251,51 @@ public RoutingAlgorithmFactory getDecoratedAlgorithmFactory(RoutingAlgorithmFact
...
+                }
+            }
+        } else {
+            // no edge_based parameter was set, we determine the CH preparation based on what is there
+            if (edgeBasedPCH != null) {
+                return edgeBasedPCH;
+            } else {
+                return nodeBasedPCH;
+            }
+        }
```
karussell: Would it make sense to reduce the nesting a bit? E.g.  (or is this then too unreadable?)
```java
if (edgeBased && edgeBasedPCH != null)
    return edgeBasedPCH;

if (!edgeBased && nodeBasedPCH != null)
    return nodeBasedPCH;

String str = edgeBased? 
  "Found a node-based CH preparation for weighting map " + map + ", but requested edge-based CH :
  "Found an edge-based CH preparation for weighting map " + map + ", but requested node-based CH : 
throw new IllegalArgumentException(str + ". You either need to configure edge-based CH preparation or set the '" + Parameters.Routing.EDGE_BASED 
+ "' parameter to '"+!edgeBased+"' (was '"+edgeBased+"'). all entries: " + entriesStrs);
```

and 

```java
return edgeBasedPCH != null ? edgeBasedPCH : nodeBasedPCH;
```

core/files/changelog.txt
easbar: ah sorry yes

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
easbar: ok yes I did this. there was an error in the error message as well ('need to configure node-based' instead of 'need to configure edge-based' in the second case)

PR #1707, opened 2019-08-29
Add test cases for bike encoder on piers with access restrictions

Nakaner:
A few weeks ago I investigated a complaint about the acceptance of a pier with `bicycle=no` for bicycles. I found out that it was my fault but in the meanwhile I added this test.


karussell:
Thanks for this. I do not see a need to add tests, when there was no actual bug. We also need to keep our tests lean.

PR #1711, opened 2019-09-03
Add support for various hazmat tags

otbutz:
This PR adds support for parsing `hazmat`, `hazmat:water` ~~and EU tunnel codes (`hazmat:B`, `hazmat:C`,...)~~. This brings us one step further to deliver more realistic routing for commercial vehicles.

I decided to split them up into several enums/parsers because they serve different concepts. The more general `hazmat` tag is quite clear and prohibits any kind of hazardous materials without the need to categorize the types of goods.

The `hazmat:water` tag prohibits the transport of goods which might threaten water protection areas. This is a critical feature for trucks carrying e.g oil.

~~The last addition is the support of EU tunnel codes tagged with `hazmat:B` down to `hazmat:E`. I know that they add a bit of complexity and seem to be tagged rather sparsely but they dictate which tunnels a truck is able to pass in the EU. As tunnels provide major shortcuts around natural obstacles these tags become critical because we might broadly underestimate the route for a truck.~~

Edit: Tunnel codes will be subject of a future PR as discussed with @karussell


karussell:
Thanks! Is it possible that two of these EncodedValues are set for one edge e.g. hazmat water and hazmat tunnel? Or is there a chance to combine them e.g. Hazmat=YES,TUNNEL_A,...,WATER,...?

otbutz:
@karussell In a tunnel we should check first of all for existing tunnel tags as they are the most specific information we're going to get. As water protection *areas* are a bit vague i'd trust the tunnel tags more if they're present.

According to taginfo there are only 348 cases of ways being tagged with a combination of `hazmat` and `hazmat:water`. We could combine `Hazmat` and `HazmatWater` but we'd be mixing a general yes/no decision with a broadly used special case which some users may like to ignore.

I was thinking about several hazmat categories(water,explosives,...) but there are regulations targeting the type of hazard like eg `hazmat:1` (which is therefore not part of this PR) and the potential impact they expose like eg `hazmat:B`.

With the current approach it would be enough to know the restriction code of the vehicle and if the load is allowed to be carried through water protection areas.

Examples:
* Restriction code `A` and water friendly goods: no hazmat restrictions
* Restriction code `D` and water friendly goods: all tunnels with category `A`,`B` or `C` and every other way not tagged with `hazmat=no`
* Restriction code `D` and non water friendly goods: all tunnels with category `A`,`B` or `C` and every other way not tagged with `hazmat=no` or `hazmat:water=no`

karussell:
Thanks for the explanation. Should the encoded values that are only for the EU prefixed with `eu`, e.g. `eu_hazmat_tunnel`? Or is this unnecessary complexity as in e.g. the US the schema is different and does not overlap?

otbutz:
According to the wiki they don't overlap as US seems to use the `hazmat:1`, `hazmat:2` etc scheme.

karussell:
Not really something missing, but I'm still unsure about all the new introduced encoded values. I would for now delay the inclusion in 0.13 and merge it early for 0.14-SNAPSHOT. I hope this is ok.

otbutz:
No problem with that  :+1:

core/src/main/java/com/graphhopper/routing/profiles/HazmatTunnel.java
```diff
@@ -0,0 +1,33 @@
...
+    /** driving with any dangerous goods allowed */
+    A("A"),
+    /** no goods with very large explosion range */
+    B("B"),
+    /** no goods with large explosion or poisoning range */
+    C("C"),
+    /** no goods which threaten a large explosion, poisoning or fire */
+    D("D"),
+    /** forbids all dangerous goods except: UN 2919,3291, 3331, 3359, 3373 */
+    E("E");
```
karussell: I'm not sure if we should introduce this EncodedValue. The world wide the usage is under 500 ways:

```
290 | hazmat:E=no 
24  | hazmat:D=no 
59  | hazmat:C=no 
115 | hazmat:B=no 
15  | hazmat:A=no 
```

See https://taginfo.openstreetmap.org/keys/hazmat%3AA#values

core/src/main/java/com/graphhopper/routing/profiles/HazmatWater.java
```diff
@@ -0,0 +1,22 @@
+package com.graphhopper.routing.profiles;
+
+/**
+ * Defines general restrictions for the transport of goods through water protection areas.<br>
+ * If not tagged it will be {@link #YES}
+ */
+public enum HazmatWater {
+    YES("yes"), NO("no");
```
karussell: Should we introduce "permissive"? As this is used over 2000 times (comparable order like "no" that has ~3500 ways)

core/src/main/java/com/graphhopper/routing/util/parsers/OSMHazmatTunnelParser.java
```diff
@@ -0,0 +1,42 @@
...
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, Access access,
+                    long relationFlags) {
+        if (readerWay.hasTag("hazmat:E", "no")) {
+            hazTunnelEnc.setEnum(false, edgeFlags, HazmatTunnel.E);
+        } else if (readerWay.hasTag("hazmat:D", "no")) {
+            hazTunnelEnc.setEnum(false, edgeFlags, HazmatTunnel.D);
+        } else if (readerWay.hasTag("hazmat:C", "no")) {
+            hazTunnelEnc.setEnum(false, edgeFlags, HazmatTunnel.C);
+        } else if (readerWay.hasTag("hazmat:B", "no") || readerWay.hasTag("hazmat:A", "no")) {
+            hazTunnelEnc.setEnum(false, edgeFlags, HazmatTunnel.B);
```
karussell: This looks a bit illogical to me. That it has the value "B" although B is not allowed.

karussell:
There is not much missing to merge this PR. Can you remove HazmatTunnel from this PR and add "permissive" to HazmatWater?

core/src/main/java/com/graphhopper/routing/profiles/HazmatTunnel.java
otbutz: The numbers are low but we're talking about tunnels. They are huge shortcuts and have drastic effect on the final route. Just imagine if your vehicle isn't allowed to use the Gotthard Tunnel.

core/src/main/java/com/graphhopper/routing/profiles/HazmatWater.java
otbutz: "permissive" means that the drive *should* use a other route. Is our routing able to work with such soft restrictions?

karussell: Sure, this is possible.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMHazmatTunnelParser.java
otbutz: You're right. I misinterpreted the meaning of A because it seemed a bit like a special case to me.

core/src/main/java/com/graphhopper/routing/profiles/HazmatTunnel.java
karussell: There can be many big issues when certain road properties are not considered, this property is nothing special. And it is always possible to include it even if this is not part of GH core.

The question is more 1. if this should be included in core to be "immediately available" and 2. if yes, if the current implementation like naming is ok. And for me both is currently unclear e.g. should we really pick the same naming and accept [potential conflicts in Finland](https://wiki.openstreetmap.org/wiki/Key:hazmat#Finland)?

To merge this PR I would remove it here and if you like, you can submit another PR to clarify the open questions?

otbutz: I think it's a good idea to postpone the addition of the tunnel codes. The Finland issue needs to be sorted out first.

otbutz:
@karussell i updated my PR :)

PR #1712, opened 2019-09-03
New MaxAxleLoadParser

otbutz:
This adds another crucial tag for proper commercial vehicle routing. The `maxaxleload` tag describes how many tonnes a single axle is allowed to have in order for a truck to be able to pass an edge. Like #1711 this may greatly change the possible routes of a vehicle as it dictates if a bridge can or cannot be crossed.

I went with an UnsignedDecimalEncodedValue with 7 bits and a 0.5 factor which allows us to store values up to 64 tonnes with a 0.5 precision. This is fully covers ~97% of [all tagged values](https://taginfo.openstreetmap.org/keys/maxaxleload#values) without loosing information and is still good enough to only be off by max 200kg for cases like `3.8t`.

### DIFF
--- a/core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
@@ -1,9 +1,34 @@
 package com.graphhopper.routing.util;
 
+import static com.graphhopper.util.GHUtility.updateDistancesFor;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.util.Arrays;
+import java.util.Collections;
+
+import org.junit.Test;
+
 import com.graphhopper.reader.ReaderWay;
-import com.graphhopper.routing.profiles.*;
-import com.graphhopper.routing.util.parsers.*;
-import com.graphhopper.routing.util.parsers.helpers.OSMWeightExtractor;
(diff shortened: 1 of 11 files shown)


core/src/test/java/com/graphhopper/routing/util/DataFlagEncoderTest.java
```diff
@@ -335,23 +336,23 @@ public void stringToMeterException() {
...
         // maximum gross weight
-        assertEquals(6, OSMMaxWeightParser.stringToTons("6t mgw"), DELTA);
+        assertEquals(6, OSMWeightExtractor.stringToTons("6t mgw"), DELTA);
     }
 
     @Test(expected = NumberFormatException.class)
     public void stringToTonsException() {
         // Unexpected values
-        OSMMaxWeightParser.stringToTons("weight limit 1.5t");
+        OSMWeightExtractor.stringToTons("weight limit 1.5t");
```
karussell: As there is now a dedicated class for this, would you mind to move the related tests into a new OSMWeightExtractorTest class?

otbutz: Ah yes. Didn't really think that the values might have special cases.

btw: we should think about optional tests using the live json data provided by taginfo: https://taginfo.openstreetmap.org/api/4/key/values?key=maxweight&filter=all&lang=de&sortname=count&sortorder=desc&rp=27&qtype=value&format=json_pretty

It should be pretty easy to automate this and get an idea how big the coverage of our parsers is.

Issue #1717, opened 2019-09-09
point_hint too slow for certain cases


karussell:
The newly introduced number handling caused an issue. If I just remove

```
try {
    if (Integer.parseInt(rewrite) > 0)
        isNumber = true;
} catch (NumberFormatException ex) {
}
```

routing requests like /maps/?point=1.346528%2C103.753174&point=1.348105%2C103.747788&point_hint=&point_hint=Singapore are already much faster. Will see what else can be done.

otbutz:
I suspect that creating the `NumberFormatException` is a rather costly operation.

Ignoring any leading `-` or `+` signs:

```java
private String prepareName(String name) {
    // \s = A whitespace character: [ \t\n\x0B\f\r]
    String[] arr = name.split("\\s");
    List<String> list = new ArrayList<>(arr.length);
    for (int i = 0; i < arr.length; i++) {
        String rewrite = NON_WORD_CHAR.matcher(toLowerCase(arr[i])).replaceAll("");
        String tmp = rewriteMap.get(rewrite);
        if (tmp != null)
            rewrite = tmp;
        
        // Ignore matching short frases like de, la, ... except it is a number
        if (rewrite.length() > 2 || StringUtils.isNumeric(rewrite)) {
            list.add(rewrite);
        }
    }
    return listToString(list);
}
```

Bonus point: it should be only evaluated for the `rewrite.length() <= 2` case now

karussell:
Yeah, I was really surprised to see that this was so slow. See #1718. In general point_hint parsing is very costly, where I tried to improve some other parts.

One of the causes for that seems to be the rather expensive NON_WORD_CHAR.replaceAll. 

Or even a cache for "name->prepare_name-cache" could make sense (that we could try to reuse for different edges but also different via points somehow, although I would try to avoid this as caching is hard)

otbutz:
Maybe it's better to create a `WORD_CHAR` matcher for the whole name and iterate over the positive matches.

otbutz:
```java
Pattern WORD_CHAR = Pattern.compile("\\p{LD}+");
Matcher wordCharMatcher = WORD_CHAR.matcher(name);
List<String> list = new ArrayList<>();
while (wordCharMatcher.find()) {
    String rewrite = toLowerCase(wordCharMatcher.group());
    String tmp = rewriteMap.get(rewrite);
    if (tmp != null)
        rewrite = tmp;
    
[...]

PR #1718, opened 2019-09-09
make parsing in prepareName for point_hint faster

karussell:
Improves #1717


core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -103,33 +105,42 @@ String getNormalizedPointHint() {
...
+            if (lastMatch == i) {
+                lastMatch++;
+                continue;
             }
-            // Ignore matching short frases like de, la, ... except it is a number
-            if (rewrite.length() > 2 || isNumber) {
-                list.add(rewrite);
+            String token = name.substring(lastMatch, i);
+            lastMatch = i + 1;
+            // end string.split
```
karussell: Will eventually revert this split complexity as it seems it does not improve speed.

karussell:
Just some stats for running Measurement on "Sachsen" with default and car:

> maximum number of prepareName calls: 4400 and average calls: 190

So it makes sense that the cache helps :)

karussell:
I usually just squash and merge

PR #1722, opened 2019-09-10
Alternative routes with node-based CH

michaz:
So this is my attempt. Goal was to start with few moving parts, and finish with fewer. :-)

Note that this doesn't even use the existing `AlternativeRoutes`. This is really all the code there is.

Actually, the most hairy code in this class is stuff for walking backwards and forwards on `Path` because we don't have that as a utility.

The algorithm seems to be pretty "self-regulating" -- thanks to CH, candidates are already filtered down. So we don't have to check so many candidates. The share-factor is a very good filter criterion, so we don't even seem to really need a maximum number of paths here. 

Screen Shot 2019-09-10 at 17 46 26


core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
```diff
@@ -0,0 +1,253 @@
...
+        final TreeSet<AlternativeInfo> alternatives = new TreeSet<>(new Comparator<AlternativeInfo>() {
+            @Override
+            public int compare(AlternativeInfo o1, AlternativeInfo o2) {
+                return Double.compare(o1.path.getWeight(), o2.path.getWeight());
+            }
+        });
+        alternatives.add(new AlternativeInfo(bestPath, 0));
+
+        final double maxWeight = maxWeightFactor * bestPath.getWeight();
+        bestWeightMapFrom.forEach(new IntObjectPredicate<SPTEntry>() {
```
easbar: You basically to a normal point-to-point query and then you consider all other nodes where the fwd/bwd search trees overlap by the time the point-to-point query has finished ? And other (possible longer) routes where the search trees do not overlap already are discarded ?

easbar:
Cool! I def. agree that the `Path` class needs some thought. Especially because it is half algorithm (path extraction) and half a data/result type. I'll open an issue for this (started typing it earlier, but not finished yet). 

The method used here is one that (in literature) predates the one used in #1524 ? What are the advantages/disadvantages of the two methods ?

easbar:
For the path issues I opened https://github.com/graphhopper/graphhopper/issues/1727

core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
michaz: Nitpick: They don't overlap, they only touch. There are no plateaus in CH.

You're right, I'm sure we could find more candidates by searching a bit longer, but (as you say) what is the right bit? :) To me, it looks like we're getting just enough candidates to filter them down carefully, without missing much.

But this is totally open for experimentation.

karussell:
I was going to merge master but the newly introduced [tests fail](https://travis-ci.org/graphhopper/graphhopper/jobs/612660686#L1019):

```
[ERROR] Tests run: 3, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 0.021 s <<< FAILURE! - in com.graphhopper.routing.AlternativeRouteCHTest

[ERROR] testCalcAlternatives2(com.graphhopper.routing.AlternativeRouteCHTest)  Time elapsed: 0.011 s  <<< FAILURE!
java.lang.AssertionError: expected:<3> but was:<1>
	at com.graphhopper.routing.AlternativeRouteCHTest.testCalcAlternatives2(AlternativeRouteCHTest.java:128)

[ERROR] testCalcAlternatives(com.graphhopper.routing.AlternativeRouteCHTest)  Time elapsed: 0.005 s  <<< FAILURE!
java.lang.AssertionError: expected:<2> but was:<1>
	at com.graphhopper.routing.AlternativeRouteCHTest.testCalcAlternatives(AlternativeRouteCHTest.java:98)
```

Is this a work in progress?

michaz:
> Is this a work in progress?

Ah. Yes. I don't even know what I was trying to do there. I don't have a proper test case yet. Coming up.

michaz:
I would cautiously call this finished. Merge?

core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
```diff
@@ -0,0 +1,240 @@
...
+                SPTEntry toSPTEntry = bestWeightMapTo.get(v);
+                if (toSPTEntry == null)
+                    return true;
+
+                // Filter alternatives that are too long
+                if (fromSPTEntry.getWeightOfVisitedPath() + toSPTEntry.getWeightOfVisitedPath() > maxWeight)
+                    return true;
+
+                // This gives us a path s -> v -> t, but since we are using contraction hierarchies,
+                // s -> v and v -> t need not be shortest paths. In fact, they can sometimes be pretty strange.
```
karussell: IMO it is not something special for CH that `s -> v -> t` is not a shortest path. Can we instead write something like

> Please note that if there is a shortest path `s -> v` and a shortest path `v -> t` this does not mean that `v` is on the shortest path `s  -> t`.

michaz: No, s -> v and v -> t are not shortest paths at that point, and that's the problem.

karussell: Can you explain why this is the case? I do not understand why e.g. `s -> v` is not shortest although it was fetched from the bestWeightMapFrom.

michaz: Because bestWeightMapFrom is the open set, not the shortest path tree. v is not closed yet, from either direction.

michaz: You're right since _that_ will also be the case without CH. But without CH, this will _almost_ be a shortest path. With CH, it can be arbitrarily ugly.

michaz: Because bestWeightMapFrom is the open set, not the shortest path tree. v is not necessarily closed yet, from either direction. You're right in that's also the case without CH, but _with_ CH, those paths can get arbitrarily ugly because one single shortcut can span a really circuitous route.

easbar: Not only that. The CH fwd/bwd searches even *settle* nodes that are not reached via a shortest path. Essentially this is because the searches are restricted to only relax relax edges leading to higher (or equal) level nodes. This is also the reason why we can prune the searches with stall on demand. See e.g. ch. 3.5.2 here: https://algo2.iti.kit.edu/documents/routeplanning/geisberger_dipl.pdf

karussell: > v is not necessarily closed yet, from either direction
> ...
> The CH fwd/bwd searches even settle nodes that are not reached via a shortest path. 

Ah, ok - thanks! I need to relearn some basics of GH and CH it seems :)

> See e.g. ch. 3.5.2 here: https://algo2.iti.kit.edu/documents/routeplanning/geisberger_dipl.pdf.

Thx! BTW: for firefox this link works :)

https://algo2.iti.kit.edu/documents/routeplanning/geisberger_dipl.pdf#subsubsection.3.5.2

michaz: > BTW: for firefox this link works :)

>:-(

karussell: > `>:-(`

if this is a sad Emoji I have good news: it works for chrome too ;)

karussell: Currently the query speed is 50 times slower than the ordinary `routingCH.mean metric`, but there is room for improvement, e.g. this branch: https://github.com/graphhopper/graphhopper/compare/try_alt_ch_faster 

For this branch and on our big map the speed up is even 6x and the `routingCH_alt.alternative_rate`is above 2. So it is basically just faster without "quality reduction" (I have no idea how to measure the real quality here, so I hope the count is a good indicator for the moment to analyze speed :) )

PR #1723, opened 2019-09-10
in GraphHopper class set minOneWayNetworkSize to 200

karussell:
We currently have  
```yml
  prepare.min_network_size: 200
  prepare.min_one_way_network_size: 200
```

in config-example.yml and so we need to make it consistent in the GraphHopper class too. Unfortunately many tests need a change as the do not expect the PrepareRoutingSubnetworks to kick in. To make it a bit convenient I introduced GraphHopper.forTesting, but I'm unsure if we should introduce this there or better create a GraphHopperOSMTest class with this method instead.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -242,14 +242,6 @@ public class GraphHopper implements GraphHopperAPI {
     }
 
     /**
-     * For testing, i.e. preserve raw network topology and use in-memory.
-     */
-    public GraphHopper forTesting() {
-        return setStoreOnFlush(false).
-                setMinNetworkSize(0, 0);
-    }
-
-    /**
      * Precise location resolution index means also more space (disc/RAM) could be consumed and
      * probably slower query times, which would be e.g. not suitable for Android. The resolution
      * specifies the tile width (in meter).
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -106,7 +106,7 @@ public class GraphHopper implements GraphHopperAPI {
     private int maxRegionSearch = 4;
     // for prepare
     private int minNetworkSize = 200;
-    private int minOneWayNetworkSize = 0;
+    private int minOneWayNetworkSize = 200;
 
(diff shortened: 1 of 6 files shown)


easbar:
I think either way is somehow acceptable, but I think I have a slight preference for not putting the method into `GraphHopper`. `forTesting()` sounds like we are supposed to use it when testing something, but this is not the case (rather the other way around, it should only be used for old tests (or for tests with small graphs where we do not want subnetwork removal). Maybe create a `createHopper(EncodingManager, location, datareaderfile)` (or similar) method in `GraphHopperOSMTest` ? The hopper creation is duplicated quite often in this test, so maybe now is the time to put this into a separate method (and put minnetwork=200 there). And maybe in some other cases we can adjust the test expectations to make the tests work with the default settings ?

karussell:
> I have a slight preference for not putting the method into GraphHopper. forTesting()

Yeah, after thinking over this I agree and will create a new class with just different defaults, to avoid too much friction when creating tests.

> Maybe create a createHopper(EncodingManager, location, datareaderfile) (or similar) method in GraphHopperOSMTest ?

Ah, this is confusing I more meant a new class that just extends GraphHopperOSM and is for Test (GraphHopperOSMForTest :) ?). Will try to find something simple.

> And maybe in some other cases we can adjust the test expectations to make the tests work with the default settings ?

This will be not possible as smaller unit tests will always have too few nodes.

michaz:
Why do you even create a new class? Now it looks like there's something special about subnetwork pruning, and you should always test GraphHopper without it.

Can't you just set the parameter to 0 on the instance, with a config flag, only on the required tests?

michaz:
Or is the problem that all the example networks just completely disappear because they are small? 

Then maybe go all the way and make the change we talked about: Let the "oneWay"-pruner behave like the regular one (not delete the biggest component, ..), and maybe even remove the other one? Or is that too hairy?

karussell:
> Or is the problem that all the example networks just completely disappear because they are small?

Yes.

> Let the "oneWay"-pruner behave like the regular one (not delete the biggest component, ..), and maybe even remove the other one?

Yeah, you are right: keeping the biggest component should be done even it is is smaller than the config.

michaz:
Superseded by 2042.

PR #1724, opened 2019-09-11
WIP: Add new `turn_costs=true/false` routing parameter

easbar:
This PR adds a new parameter `turn_costs=true/false`. This does the exact same as `edge_based=true/false` at the moment but is supposed to be more user-friendly (users who want to enable turn costs (or at least OSM turn restrictions) most likely do not know (and should not need to know) about the routing algorithm being edge-based.

This is basically the minimum change that adds this parameter, but I am not really happy with it, see #1725.


easbar:
Ok :) :+1:

karussell:
Uups, I found this already very useful and did not recognize the WIP :D

PR #1726, opened 2019-09-12
Wheelchair routing profile

don-philipe:
This pull request adds a new routing profile for wheelchairs. It takes surfaces and smoothness into account. It mainly uses values and rules from the foot routing profile (as, at least in Germany, wheelchair users are equivalent to pedestrians). But in some cases it is more restrictive.

The icon is just a placeholder. I'm sure there are people who can do that better than me.


core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,209 @@
...
+
+        // for now no explicit avoiding #257
+        //avoidHighwayTags.add("cycleway"); 
+        allowedHighwayTags.addAll(safeHighwayTags);
+        allowedHighwayTags.addAll(avoidHighwayTags);
+        allowedHighwayTags.add("cycleway");
+        allowedHighwayTags.add("unclassified");
+        allowedHighwayTags.add("road");
+        // disallowed in some countries
+        //allowedHighwayTags.add("bridleway");
```
karussell: Can the commented lines be removed?

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Override
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, EncodingManager.Access access, long relationFlags) {
+        if (access.canSkip()) {
+            return edgeFlags;
+        }
+
+        if (!access.isFerry()) {
+            String sacScale = way.getTag("sac_scale");
```
karussell: Why is this required? The way should have `access.canSkip` (see getAccess method)

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,209 @@
...
+            double ferrySpeed = getFerrySpeed(way);
+            setSpeed(false, edgeFlags, ferrySpeed);
+            accessEnc.setBool(false, edgeFlags, true);
+            accessEnc.setBool(true, edgeFlags, true);
+        }
+
+        return edgeFlags;
+    }
+
+    protected int handlePriority(ReaderWay way, int priorityFromRelation) {
```
karussell: Can this method be removed? Looks identical...

web/src/main/resources/assets/img/wheelchair.svg
```diff
@@ -0,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
karussell: There is something in the middle of the wheel. Not visible with white background:

image

image


Sorry, if this is a bit nickpicking but can you remove/change this?

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,209 @@
...
+public class WheelchairFlagEncoder extends FootFlagEncoder {
+    static final int SLOW_SPEED = 2;
+    static final int MEAN_SPEED = 5;
+    static final int FERRY_SPEED = 15;
+    final Set<String> safeHighwayTags = new HashSet<>();
+    final Set<String> allowedHighwayTags = new HashSet<>();
+    final Set<String> avoidHighwayTags = new HashSet<>();
+    final Set<String> avoidSurfaces = new HashSet<>();
+    final Set<String> avoidSmoothness = new HashSet<>();
+    // convert network tag of hiking routes into a way route code
```
karussell: Can this comment be removed?

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
don-philipe: I changed the method to still use the super method, but also added wheelchair specific evaluation.

don-philipe:
Do I need to do something more, additionally to committing and pushing the requested changes?

karussell:
Thanks! Somehow I wasn't notified about this. Will review in the next hours&days.

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,206 @@
...
+
+        avoidSurfaces.add("cobblestone");
+        avoidSurfaces.add("gravel");
+        avoidSurfaces.add("sand");
+
+        avoidSmoothness.add("bad");
+        avoidSmoothness.add("very_bad");
+        avoidSmoothness.add("horrible");
+        avoidSmoothness.add("very_horrible");
+        avoidSmoothness.add("impassable");
```
karussell: This all should not be just avoided but completely excluded from the road network

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,206 @@
...
+ */
+public class WheelchairFlagEncoder extends FootFlagEncoder {
+    static final int SLOW_SPEED = 2;
+    static final int MEAN_SPEED = 5;
+    static final int FERRY_SPEED = 15;
+    final Set<String> safeHighwayTags = new HashSet<>();
+    final Set<String> allowedHighwayTags = new HashSet<>();
+    final Set<String> avoidHighwayTags = new HashSet<>();
+    final Set<String> avoidSurfaces = new HashSet<>();
+    final Set<String> avoidSmoothness = new HashSet<>();
```
karussell: The following variables are already defined in the super class: safeHighwayTags, allowedHighwayTags, ...

Also the static variables should not be overwritten and as they are identical you should be able to just remove them.

don-philipe:
> I additionally found wheelchair:no or wheelchair:limited - is this already considered? See https://wiki.openstreetmap.org/wiki/Wheelchair_routing#Requirements_for_Wheelchair_Accessible_Routes

`wheelchair:no` is already considered and the way is then skipped. I added now `wheelchair:limited` to have lower priority.

> Also it really needs two directions support to avoid steep elevation. This is now easy, see Bike2WeightFlagEncoder (new UnsignedDecimalEncodedValue("xy", bits, factor, true))

I'm not sure what you mean with that. Isn't that already be solved by skipping ways that are too steep (see below)? Or do you mean that the speed should be adjusted to the incline?

> Also it would be important to consider the slope tag.

As long as the `incline` tag has a numerical value it can now be evaluated. Ways with incline > 6% will be skipped.

karussell:
> I added now wheelchair:limited to have lower priority.

:+1: 

> Or do you mean that the speed should be adjusted to the incline?

Yes, but you are right that this is not necessary for a first version

> As long as the incline tag has a numerical value it can now be evaluated. Ways with incline > 6% will be skipped.

OSM tags are very rare for this and I would prefer skipping too steep roads via elevation. See [here](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/Bike2WeightFlagEncoder.java#L91) of how we do this for Bike2WeightFlagEncoder

don-philipe:
> OSM tags are very rare for this and I would prefer skipping too steep roads via elevation. See here of how we do this for Bike2WeightFlagEncoder

I assumed, that elevation data from `PointList.getElevation()` is in percent. Am I right?

Now the `applyWayTags()` method will slow down speed for inclines between 3 and 6 percent and set speed = 0 for inclines greater than 6 percent (same for decline).

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -111,6 +117,25 @@ public int getVersion() {
...
+    public void createEncodedValues(List<EncodedValue> registerNewEncodedValue, String prefix, int index) {
+        super.createEncodedValues(registerNewEncodedValue, prefix, index);
+        for (int i = 0; i < registerNewEncodedValue.size(); i++) {
+            if (registerNewEncodedValue.get(i).getName().endsWith("average_speed")) {
+                registerNewEncodedValue.remove(i);
+                registerNewEncodedValue.add(speedEncoder = new UnsignedDecimalEncodedValue(
+                        getKey(prefix, "average_speed"), speedBits, speedFactor, true));
+            }
+        }
+    }
```
karussell: It would be simpler to make the property configurable in FootFlagEncoder and then set storeTwoDirections=true in this encoder. See https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java#L67

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -208,6 +233,55 @@ public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, EncodingManager.A
...
+     * @param bwdSpeed speed value in backward direction
+     */
+    private void setFwdBwdSpeed(EdgeIteratorState edge, int fwdSpeed, int bwdSpeed) {
+        if (accessEnc.getBool(false, edge.getFlags())) {
+            setSpeed(false, edge.getFlags(), fwdSpeed);
+        }
+        if (accessEnc.getBool(true, edge.getFlags())) {
+            setSpeed(true, edge.getFlags(), bwdSpeed);
+        }
+    }
```
karussell: As the setSpeed method is deprecated we need a different approach here. Some more hints:

 * use edge.get(accessEnc) instead accessEnc.getBool
 * instead of setting speed to 0, set access to false

core/src/test/java/com/graphhopper/routing/util/WheelchairFlagEncoderTest.java
```diff
@@ -548,4 +546,44 @@ public void testSmoothness() {
...
+    @Test
+    public void testApplyWayTags() {
+        Graph graph = initExampleGraph();
+
+        EdgeIteratorState edge0 = GHUtility.getEdge(graph, 0, 1);
+        ReaderWay way0 = new ReaderWay(1);
+        wheelchairEncoder.applyWayTags(way0, edge0);
+
+        IntsRef flags0 = edge0.getFlags();
+        assertEquals(2, wheelchairEncoder.getSpeed(false, flags0), 1e-1);
```
karussell: Instead of the encoder use the DecimalEncodedValue:

```
edge0.get(encoder.getAverageSpeedEnc())
edge0.getReverse(encoder.getAverageSpeedEnc())
```

karussell:
> I assumed, that elevation data from PointList.getElevation() is in percent. Am I right?

It is in meter.

don-philipe:
Sorry for disturbing you again, but can you please have a look at the changes?

karussell:
> Sorry for disturbing you again

Not a problem. Please leave a comment as I do not track of the commits in the PRs, especially I do not know when something is ready etc

Will review this again later. For now can you send us the [signed CLA](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md#license-agreement) (or just your email address) and add yourself as a contributor in CONTRIBUTORS.md?

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
karussell: The second case will fail if speed_two_directions=false.
Can you please add a test for this?

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,312 @@
...
+    public WheelchairFlagEncoder(PMap properties) {
+        this((int) properties.getLong("speedBits", 4),
+                properties.getDouble("speedFactor", 1));
+        this.properties = properties;
+        this.setBlockFords(properties.getBool("block_fords", true));
+    }
+
+    public WheelchairFlagEncoder(String propertiesStr) {
+        this(new PMap(propertiesStr));
+    }
```
karussell: Can you remove the constructors that are unused?

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,312 @@
...
+    /**
+     * Should be only instantiated via EncodingManager
+     */
+    public WheelchairFlagEncoder() {
+        this(4, 1);
+    }
+
+    public WheelchairFlagEncoder(PMap properties) {
+        this((int) properties.getLong("speedBits", 4),
+                properties.getDouble("speedFactor", 1));
```
karussell: Sorry, there is a bug in Foot and HikeFlagEncoder: can you use speed_bits and speed_factor instead of camel case here?

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,206 @@
...
+
+        avoidHighwayTags.add("trunk");
+        avoidHighwayTags.add("trunk_link");
+        avoidHighwayTags.add("primary");
+        avoidHighwayTags.add("primary_link");
+        avoidHighwayTags.add("secondary");
+        avoidHighwayTags.add("secondary_link");
+        avoidHighwayTags.add("tertiary");
+        avoidHighwayTags.add("tertiary_link");
+        avoidHighwayTags.add("steps");
```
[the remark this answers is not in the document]
karussell: The meaning is different now, so it should not use the same avoidHighwayTags list.

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
karussell: Please remove this method and introduce speedTwoDirections in the super class

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
karussell: Hmmh, but when we introduce it there it should also work for FootFlagEncoder. Let me know if you need help there.

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,312 @@
...
+
+                    if (kerbHeight > maxKerbHeightCm) {
+                        return EncodingManager.Access.CAN_SKIP;
+                    }
+                } catch (NumberFormatException ex) { }
+            }
+        }
+
+        return super.getAccess(way);
+    }
```
karussell: Can you list what you changed? Maybe we can make the super method a bit more generic and remove it here.

karussell: Have fixed a few things. Still, can you list what you changed?

don-philipe: > Still, can you list what you changed?

Will do this in the next days when I have some more time for this (hopefully).

don-philipe: Regarding the `getAccess` method, what I wrote weren't changes in the sense that behavior of the super method is overwritten. My intention was to extend the restrictions. For example to skip ways with bad surface. Further restrictions for pedestrians from the super method should be applied as well as, at least in Germany, wheelchairs are legally equivalent to pedestrians and therefore must use footways. I don't know it for sure, but I can imagine, that is similar in other countries.

Overall my additions in the `getAccess` methods are:
- allow access for ways with `wheelchair=<intendedValue>`
- skip ways with `sac_scale` tag - these are hiking trails
- skip ways with `highway=<excludeHighway>` and no sidewalk tag
- skip ways with `surface=<excludeSurface>` and no sidewalk
  - same for smoothness
- skip ways with in/decline is too steep
- skip ways with high kerbs

karussell: Thanks!

karussell:
Thanks again :) !

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,286 @@
...
+        excludeSurfaces.add("sand");
+
+        excludeSmoothness.add("bad");
+        excludeSmoothness.add("very_bad");
+        excludeSmoothness.add("horrible");
+        excludeSmoothness.add("very_horrible");
+        excludeSmoothness.add("impassable");
+
+        allowedHighwayTags.addAll(safeHighwayTags);
+        allowedHighwayTags.addAll(excludeHighwayTags);
```
karussell: @don-philipe I think this is a bug, right?

cbayerlein:
I managed to enable wheelchair routing thanks to this pull! I think it’s really awesome that wheelchair routing is available. Thank you so much @don-philipe and @karussell  !

I wasn’t aware that it’s only in later releases. I used 0.13 first (as described in the quick start guide), now I’m using 1.0-pre13.

Unfortunally, I’m too much of a newbie especially in Java, so I’d need to learn a lot before I can contribute myself :-(. But I’ll definitivly look into it and will see what I can do.

For now I hope it's helpful if I do my best as a test user. I already found some issues. Maybe you can help. e. g. for the following request

http://s7.thalon.me:8989/maps/?point=Brunnenstra%C3%9Fe%2011a%20H%C3%B6hr-Grenzhausen&point=Kirchstr.%20H%C3%B6hr-Grenzhausen&locale=de-DE&vehicle=wheelchair&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale

I get:

Cannot find point 0: 50.439114,7.655696
Cannot find point 1: 50.438526,7.656077

Bike2 and foot routing works for the two points. How can I check what's going wrong and maybe solve the issue?

cbayerlein:
Another issue with this route:

http://s7.thalon.me:8989/maps/?point=L%C3%B6hrstr.%20107%20Koblenz&point=Gem%C3%BCsegasse%20Koblenz&locale=de-DE&vehicle=wheelchair&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale

Going via Bahnhofstr. to avoid Löhrstr. seems odd (I can see no reason), also this strange diversion to cross Am Plan/Altengraben... Plus, at the end(after turning right to Gebrüder-Dommermuth-Weg), there're actually (in rl) steps, as far as I remember.

cbayerlein:
And another one:

http://s7.thalon.me:8989/maps/?point=Weinbergstr.%20Koblenz&point=Pastor-Klein-Str.%20Koblenz&locale=de-DE&vehicle=wheelchair&weighting=fastest&elevation=true&use_miles=false&layer=OpenStreetMap.de

yields "Connection between locations not found" although there is one (long way, but possible).

cbayerlein:
I think the problems can be divided into two causes:

- missing data in OSM (nothing we can do here except mapping ourselves)
- too restrictive routing - maybe we should loosen up a bit.

What do you think?

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
cbayerlein: > at least in Germany, wheelchairs are legally equivalent to pedestrians and therefore must use footways. 

That's not entirely true. I have a power chair with lights and indicators and thus am allowed to go on the road. Furthermore, going on the sidewalk is sometimes simply not possible because of high kerbs or bad surfaces.

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,286 @@
...
+        excludeHighwayTags.add("primary");
+        excludeHighwayTags.add("primary_link");
+        excludeHighwayTags.add("secondary");
+        excludeHighwayTags.add("secondary_link");
+        excludeHighwayTags.add("tertiary");
+        excludeHighwayTags.add("tertiary_link");
+        excludeHighwayTags.add("steps");
+        excludeHighwayTags.add("track");
+
+        excludeSurfaces.add("cobblestone");
```
cbayerlein: I'd prefer not to exclude so strictly. Sometimes, it is unavoidable. Is t possible to give a very bad speed penalty, like 1km/h?

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,286 @@
...
+        excludeSurfaces.add("gravel");
+        excludeSurfaces.add("sand");
+
+        excludeSmoothness.add("bad");
```
cbayerlein: same as with surfaces. The worse the slower...

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,286 @@
...
+/**
+ * A flag encoder for wheelchairs.
+ *
+ * @author don-philipe
+ */
+public class WheelchairFlagEncoder extends FootFlagEncoder {
+    private final Set<String> excludeSurfaces = new HashSet<>();
+    private final Set<String> excludeSmoothness = new HashSet<>();
+    private final Set<String> excludeHighwayTags = new HashSet<>();
+    private final int maxInclinePercent = 6;
```
cbayerlein: That's iso norm for ramps, but for ways very restrictive and thus unrealistic. I would raise it to 15, giving it a speed penalty over 6.

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,286 @@
...
+        safeHighwayTags.add("service");
+        safeHighwayTags.add("platform");
+
+        excludeHighwayTags.add("trunk");
+        excludeHighwayTags.add("trunk_link");
+        excludeHighwayTags.add("primary");
+        excludeHighwayTags.add("primary_link");
+        excludeHighwayTags.add("secondary");
+        excludeHighwayTags.add("secondary_link");
+        excludeHighwayTags.add("tertiary");
```
cbayerlein: Is tertiary a Kreisstraße/Landstraße? Then please don't exclude. Sometimes, that's the only available way in rural areas - except for forrest/agricultural roads (Feldwege), wich are much worse.

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,286 @@
...
+            throw new IllegalStateException("Infinite distance should not happen due to #435. way ID=" + way.getId());
+        }
+
+        if (fullDist2D < 1) {
+            return;
+        }
+
+        double eleDelta = pl.getElevation(pl.size() - 1) - prevEle;
+        double elePercent = eleDelta / fullDist2D * 100;
+        int smallInclinePercent = 3;
```
cbayerlein: I'd raise that to 5,giving a speed penalty over 3.

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -0,0 +1,286 @@
...
+        excludeHighwayTags.add("trunk");
+        excludeHighwayTags.add("trunk_link");
+        excludeHighwayTags.add("primary");
+        excludeHighwayTags.add("primary_link");
+        excludeHighwayTags.add("secondary");
+        excludeHighwayTags.add("secondary_link");
+        excludeHighwayTags.add("tertiary");
+        excludeHighwayTags.add("tertiary_link");
+        excludeHighwayTags.add("steps");
+        excludeHighwayTags.add("track");
```
cbayerlein: tracktype=grade1 would be possible

cbayerlein:
Maybe we should add excludeHighwayTags.add("path");

cbayerlein:
In the meantime I changed a bit of code in the WheelchairFlagEncoder and now the links which I posted above produce better results as they point to my GH instance.  Basically I removed some restrictions like excluding cobblestones and others which were too strict from my point of view and raised the max slope limit to 15%. The changes are very rudimentary (commenting out lines and replacing 6 with 15 ) and rather suboptimal in a way so that there is a lot of room for improvement. For example it would be better to give a speed penalty to bad surfaces. Unfortunally I didn't have time and knowledge yet to implement that, but at least I wrote some code reviews in regards to this. Maybe we can continue working on that topic. I would be very happy about that and will look more deeply into the matter myself next year. For now Guten Rutsch!

PR #1729, opened 2019-09-17
Android: VTM 0.12.0 OpenGL vector map library

devemux86:
- [VTM 0.12.0](https://github.com/mapsforge/vtm) OpenGL vector map library
- Update Android build system

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -1,8 +1,8 @@
 apply plugin: 'com.android.application'
 
 android {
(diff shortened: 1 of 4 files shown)


devemux86:
Graph loading currently does not work with GraphHopper latest snapshot.
Probably the downloaded graphs need to be updated?

karussell:
> Probably the downloaded graphs need to be updated?

Have updated them. Should work - at least for the now stable 0.13 release.

PR #1730, opened 2019-09-17
Move path extraction logic out of `Path` class

easbar:
This PR is supposed to move the path extraction stuff out of `Path`. So far `Path` was used by the routing algorithms to store some intermediate state (`SPTEntry`, `switchToFrom`, etc.) that later was used to 'extract' the `Path`. So basically some parts of the routing algorithm used to live inside `Path` and (because the extraction depends on the type of algorithm) the subclasses (`PathNative`, `Path4CH`, `PathBidirRef`, `EdgeBasedPath4CH`). However, since `Path` is the return type of the `RoutingAlgorithm` interface it should really be a data-only class IMO.

Here I basically did the following:

- Added dedicated classes for path extraction: `PathExtractor`, `BidirPathExtractor`, `Node/EdgeBasedCHBidirPathExtractor`, which basically take some `SPTEntry` (or two) and create a `Path` from it. In turn I removed the path extraction stuff from `Path` and deleted its subclasses.

- Instead of holding a reference to an (intermediate) `Path` instance (`bestPath`) `AbstractBidirAlgo` now just has a reference to the best fwd/bwd `SPTEntry` and the `bestWeight`. We can override `extractPath` if we want something different than the default path extraction, or just change the path extraction implementation by overriding `createPathExtractor`.


core/src/main/java/com/graphhopper/routing/BidirPathExtractor.java
```diff
@@ -0,0 +1,130 @@
...
+ * @author easbar
+ */
+public class BidirPathExtractor {
+    private final Graph graph;
+    private final Weighting weighting;
+    protected final Path path;
+
+    public static Path extractPath(Graph graph, Weighting weighting, SPTEntry fwdEntry, SPTEntry bwdEntry, double weight) {
+        return new BidirPathExtractor(graph, weighting).extract(fwdEntry, bwdEntry, weight);
+    }
```
karussell: If this is the only required public method, can we make the constructor and other methods (package) protected or even private?

easbar: the idea was that one can override stuff like 'onEdge' if need be, but `protected` works

karussell:
The build error for openjdk13 drives me crazy. There was a mistake regarding the version, but now [this is fixed](https://github.com/graphhopper/graphhopper/commit/f65576ca0de54f677970cf1d3778b4a09384f0f9) and locally it builds fine with it, even if I do `rm -rf ~/.m2/repository/com/graphhopper/*`. Also I removed the caches for travis, but this does not help. Do you have an idea or can you reproduce the problem that travis has with openjdk13?

The command should be:
```
export JAVA_HOME=/path/to/jdk-13/
mvn -version # verify that jdk-13 is used
mvn clean test very checkstyle:check forbiddenapis:check
```

Or maybe we try without the caching setting again in .travis.yml ...

karussell:
> Or maybe we try without the caching setting again in .travis.yml ...

Didn't help.

The problem is that it downloads the core dependency:

```
Downloading from sonatype-snapshots: https://oss.sonatype.org/content/repositories/snapshots/com/graphhopper/graphhopper-core/0.14-SNAPSHOT/graphhopper-core-0.14-20190918.103923-1.jar
```

Although it builds this dependency before:

```
[INFO] Building jar: /home/travis/build/graphhopper/graphhopper/core/target/graphhopper-core-0.14-SNAPSHOT.jar
```

.travis.yml
```diff
@@ -42,7 +42,7 @@ before_script:
   - cd web && npm install && BROWSERIFYSWAP_ENV='development' npm run bundleProduction && cd ..
 
 script:
-  - "mvn -Dkey=$API_KEY clean test verify -B"
+  - "mvn -Dkey=$API_KEY clean test verify install -B"
```
easbar: this seems to have fixed the build ? should we keep it like this ?

karussell: Yes, the install forces maven to always fetch the dep from local repo. A bit strange as I thought that maven knows this already, but maybe also a bug of that forbiddenapis plugin. A bit strange that it passes for openjdk8...

easbar: Ok, `install` should be enough (instead of `verify install`) since install already includes verify, Ill do this in master and then merge this.

easbar: And thanks a lot for figuring this one out!

michaz:
Omg, thank you!

PR #1733, opened 2019-09-20
Fix sort corrupting graph

### LINKED ISSUE #1732: Sorting the graph breaks it
Whenever I set `graph.do_sort: true` it results in a broken graph. The location index is huge, and all routing requests fail. I traced the issue down to the sort code copying over the old base/adjacent node IDs to the new edge once it gets created since `copyProperties` is copying node IDs from old to new edge. I have a pull request ready with a test that illustrates the failure and a potential fix.

msbarry:
Fix #1732 by making `copyProperties` not copy over the node IDs from old to new edge. No tests fail if I just remove the the call to `writeEdge` entirely in that method, but I'm assuming the reverse handling needs to be in there for a reason.

### DIFF
--- a/core/src/test/java/com/graphhopper/util/GHUtilityTest.java
+++ b/core/src/test/java/com/graphhopper/util/GHUtilityTest.java
@@ -73,6 +73,9 @@ public class GHUtilityTest {
         DistanceCalc calc = new DistanceCalc2D();
         AllEdgesIterator iter = graph.getAllEdges();
         while (iter.next()) {
+            // This is meant to verify that all of the same edges (including tower nodes)
+            // are included in the copied graph. Can not use iter.getDistance() since it
+            // does not verify new geometry. See #1732
             distance += iter.fetchWayGeometry(3).calcDistance(calc);
         }
         return distance;
--- a/core/src/test/java/com/graphhopper/util/GHUtilityTest.java
+++ b/core/src/test/java/com/graphhopper/util/GHUtilityTest.java
@@ -19,6 +19,7 @@ package com.graphhopper.util;
 
 import com.graphhopper.coll.GHIntLongHashMap;
 import com.graphhopper.routing.profiles.BooleanEncodedValue;
+import com.graphhopper.routing.util.AllEdgesIterator;
 import com.graphhopper.routing.util.CarFlagEncoder;
 import com.graphhopper.routing.util.EncodingManager;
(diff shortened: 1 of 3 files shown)


karussell:
Oh, this looks indeed ugly. Thanks for the quick fix! 

Please add yourself as [a contributor](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTORS.md) and [send us the CLA](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md#license-agreement).

core/src/test/java/com/graphhopper/util/GHUtilityTest.java
```diff
@@ -67,6 +68,16 @@ Graph initUnsorted(Graph g) {
         return g;
     }
 
+    double getLengthOfAllEdges(Graph graph) {
+        double distance = 0;
+        DistanceCalc calc = new DistanceCalc2D();
+        AllEdgesIterator iter = graph.getAllEdges();
+        while (iter.next()) {
+            distance += iter.fetchWayGeometry(3).calcDistance(calc);
```
karussell: Is this also possible with iter.getDistance()?

karussell:
Oh, this looks indeed ugly. Thanks for the quick fix! 

Please add yourself as [a contributor](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTORS.md) and [send us the CLA](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md#license-agreement).

core/src/test/java/com/graphhopper/util/GHUtilityTest.java
karussell: This does not work then. Will look a bit deeper whats wrong here.

msbarry:
Thanks @karussell! Sent a request for the CLA and added myself as a contributor.  I removed the entire `writeEdge` call and all of the tests still pass, but `getDistance` doesn't seem to verify this failure in the tests so kept as-is and added a comment.

core/src/test/java/com/graphhopper/util/GHUtilityTest.java
msbarry: getDistance doesn't seem to verify this failure in the tests so kept as-is and added a comment.

karussell:
Thanks!

Issue #1735, opened 2019-09-20
Several matched keys found for keyPrefix


routexl:
Ran into an error when creating new graph:

```
## using java  from C:/Program Files/Java/jdk1.8.0_152
./graphhopper.sh: line 138: mvn: command not found
## using existing osm file netherlands-latest.osm.pbf
## building graphhopper jar: web/target/graphhopper-web-0.14-SNAPSHOT.jar
## using maven at ./maven
## now web. JAVA_OPTS=-Xmx14G -Xms2G
2019-09-20 14:47:44.878 [main] INFO  i.d.server.DefaultServerFactory - Registering jersey handler with root path prefix: /
2019-09-20 14:47:44.880 [main] INFO  i.d.server.DefaultServerFactory - Registering admin handler with root path prefix: /
2019-09-20 14:47:45.284 [main] INFO  i.d.b.assets.ConfiguredAssetsBundle - Registering ConfiguredAssetBundle with name: assets for path /maps/*
2019-09-20 14:47:45.291 [main] INFO  io.dropwizard.server.ServerFactory - Starting GraphHopperApplication
                        _     _
   __ _ _ __ __ _ _ __ | |__ | |__   ___  _ __  _ __   ___ _ __
  / _` | '__/ _` | '_ \| '_ \| '_ \ / _ \| '_ \| '_ \ / _ \ '__|
 | (_| | | | (_| | |_) | | | | | | | (_) | |_) | |_) |  __/ |
  \__, |_|  \__,_| .__/|_| |_|_| |_|\___/| .__/| .__/ \___|_|
  |___/          |_|                     |_|   |_|

2019-09-20 14:47:45.329 [main] INFO  o.e.jetty.setuid.SetUIDListener - Opened application@ea52184{HTTP/1.1,[http/1.1]}{localhost:8989}
2019-09-20 14:47:45.329 [main] INFO  o.e.jetty.setuid.SetUIDListener - Opened admin@3e7b65d7{HTTP/1.1,[http/1.1]}{localhost:8990}
2019-09-20 14:47:45.330 [main] INFO  org.eclipse.jetty.server.Server - jetty-9.4.z-SNAPSHOT; built: 2019-04-29T20:42:08.989Z; git: e1bc35120a6617ee3df052294e433f3a25ce7097; jvm 1.8.0_152-b16
[...]

PR #1736, opened 2019-09-21
Add handling of conditional turn restrictions

easbar:
Unfortunately #1683 does not work yet, because there can be cases with multiple conditional turn restrictions like so:

```
restriction:bus=...
restriction:hgv=...
```

Therefore I reverted the previously merged #1683 and reopened here.

See #1735


easbar:
@sguill Do you know how to address this problem ?

sguill:
Yes, my first idea was to return silently the first match, if you remember. But, now I think, it is better to return a list. Do you agree ?

easbar:
> Yes, my first idea was to return silently the first match, if you remember.

Yeah, but this way we would simply ignore some restrictions (and arbitrarily pick some of them) which I think is worse. 

> But, now I think, it is better to return a list. Do you agree ?

Yes apparently a relation with restriction must yield a list of multiple restrictions. However, if possible I would still try to feed information about single restrictions to the encoder ? So take the OSM relation, build a list of things that each represent a specific restriction (e.g. left only for buses) and then feed them one by one into `encoder.acceptTurnRelation`, does that make sense ?

I also thought about how the `except` list should be interpreted when there are multiple conditional restrictions and came to the conclusion that all these exceptions should be applied to *all* the restrictions within the relation (?).

sguill:
I made the changes in commit 15ba63f *multiple conditional turn restriction types* in branch *accept_turn_relation* but it is not taken into account. I don't know how to do?

easbar:
> but it is not taken into account.

Ah yes this is because you pushed to `sguill/accept_turn_relation`, but this PR uses `graphhopper/accept_turn_relation`. I probably should have created a PR using `sguill/accept_turn_relation`, because you will not be able to push to `graphhopper/accept_turn_relation` directly.

Now I think you need to checkout `graphhopper/accept_turn_relation` locally, make your recent changes, push it to your fork (under some new name) and create a new PR. You could also just create a new PR using `sguill/accept_turn_relation`, but then you need to make sure that everything is still included. I think right now it is missing most of your changes (compared to `graphhopper/accept_turn_relation`), they got lost because you merged master that included the revert commit (I already took care of this in `graphhopper/accept_turn_relation`).

easbar:
Closing in favor of #1738

PR #1738, opened 2019-09-24
Handling turn relation

sguill:
To deals with multiple conditional turn restrictions:

```
restriction:bus=...
restriction:hgv=...
```

See #1736


core/src/main/java/com/graphhopper/reader/ReaderElement.java
```diff
@@ -95,6 +94,16 @@ public String getTag(String name) {
         return val;
     }
 
+    public List<String> getKeysStartsWith(String keyPrefix) {
```
easbar: `getKeysStartWith` (because keys is plural now) or even `getKeysWithPrefix` ?

core/src/main/java/com/graphhopper/reader/ReaderElement.java
```diff
@@ -145,6 +154,15 @@ public boolean hasTag(List<String> keyList, Set<String> values) {
         return false;
     }
 
+    public boolean hasTagStartsWith(String keyPrefix) {
```
easbar: According to OSM: "A tag consists of two items, a key and a value", so its a bit unclear what it means when a tag starts with a key prefix. Maybe rename to `hasTagWithKeyPrefix` ?

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -411,14 +414,16 @@ void processWay(ReaderWay way) {
...
-            if (turnRelation != null) {
-                GraphExtension extendedStorage = graph.getExtension();
-                if (extendedStorage instanceof TurnCostExtension) {
-                    TurnCostExtension tcs = (TurnCostExtension) extendedStorage;
-                    Collection<TurnCostTableEntry> entries = analyzeTurnRelation(turnRelation);
-                    for (TurnCostTableEntry entry : entries) {
-                        tcs.addTurnInfo(entry.edgeFrom, entry.nodeVia, entry.edgeTo, entry.flags);
+            List<OSMTurnRelation> turnRelations = createTurnRelation(relation);
+            for (OSMTurnRelation turnRelation : turnRelations) {
+                if (turnRelation != null) {
```
easbar: this null check should not be needed

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -411,14 +414,16 @@ void processWay(ReaderWay way) {
...
+                    GraphExtension extendedStorage = graph.getExtension();
+                    if (extendedStorage instanceof TurnCostExtension) {
```
easbar: In case we have no turn cost extension we can return early here (we can change the order of the if statement and the for loop)

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -858,8 +865,41 @@ private long createNewNodeId() {
      *
      * @return the OSM turn relation, <code>null</code>, if unsupported turn relation
      */
-    OSMTurnRelation createTurnRelation(ReaderRelation relation) {
-        OSMTurnRelation.Type type = OSMTurnRelation.Type.getRestrictionType(relation.getTag("restriction"));
+    List<OSMTurnRelation> createTurnRelation(ReaderRelation relation) {
```
easbar: should be called `createTurnRelations` now

reader-osm/src/test/resources/com/graphhopper/reader/osm/test-conditional-turn-restrictions.xml
```diff
@@ -0,0 +1,129 @@
...
+        <nd ref="5"/>     
+        <nd ref="10"/>
+        <tag k="highway" v="cycleway" />
+    </way>
+        
+    <relation id="1001">
+        <member type="way" ref="23" role="from"/>
+        <member type="way" ref="34" role="to"/>
+        <member type="node" ref="3" role="via"/>
+        <tag k="restriction" v="only_straight_on"/>
```
easbar: We should add a test that covers the case that previously failed: A relation that has multiple conditional restrictions. Maybe also covering the case where there are multiple conditional restrictions and an except list. We could do this here, but usually I think its better to have isolated tests for the different cases (they can be as small as a single junction ?)

core/src/main/java/com/graphhopper/reader/OSMTurnRelation.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /**
+     * For a conditional turn restriction, test each vehicle type to verify if it is concerned.
+     * For a normal turn restriction (non conditional), the restriction is necessary considered.
+     * @param vehicleTypes
+     * @return
+     */
+    public boolean isVehicleTypeConcernedByTurnRestriction(List<String> vehicleTypes) {
+        if (vehicleTypeRestricted.isEmpty() && vehicleTypesExcept.isEmpty()) {
```
easbar: Should there be any preference between `restriction:x` and `except=x` ? I cannot see clearly from OSM documentation. IMO we could just do
```java
        // if the restriction explicitly does not apply for one of the vehicles we do not accept it
        if (!Collections.disjoint(vehicleTypes, vehicleTypesExcept)) {
            return false;
        }
        return vehicleTypeRestricted.isEmpty() || vehicleTypes.contains(vehicleTypeRestricted);
```
here.
```

File: core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java [hunk 8]
```diff
@@ -186,6 +187,14 @@ protected double getSpeed(ReaderWay way) {
         return speed;
     }
 
+    @Override
+    public boolean acceptsTurnRelation(OSMTurnRelation relation) {
+        if (relation.isVehicleTypeConcernedByTurnRestriction(restrictions)) {
```
easbar: just `return relation.isVehicleTypeConcernedByTurnRestriction(restrictions)` (here and in other encoders)

core/src/test/java/com/graphhopper/routing/util/CarFlagEncoderTest.java
```diff
@@ -651,4 +652,34 @@ public void testIssue_1256() {
         assertEquals(2.5, encoder.getFerrySpeed(way), .1);
         assertEquals(.5, lowFactorCar.getFerrySpeed(way), .1);
     }
+
+    @Test
+    public void testAcceptsTurnRelation() {
```
easbar: This should go into `OSMTurnRelationTest` instead. Its very simple, just replacing `acceptsTurnRelation` calls with `isVehicleTypeConcerned..`.

core/src/main/java/com/graphhopper/reader/OSMTurnRelation.java
```diff
@@ -0,0 +1,140 @@
...
+
+    public void setVehicleTypeRestricted(String vehicleTypeRestricted) {
+        this.vehicleTypeRestricted = vehicleTypeRestricted;
+    }
+
+    public List<String> getVehicleTypesExcept() {
+        return vehicleTypesExcept;
+    }
+
+    public void addVehicleTypesExcept(List<String> vehicleTypesExcept) {
```
easbar: this method does not 'add' vehicles, it replaces them. either rename or change behavior.

sguill: ok, I changed it in *setVehicleTypesExcept*.

core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
sguill: ok

core/src/test/java/com/graphhopper/routing/util/CarFlagEncoderTest.java
sguill: Ok I moved the test and changed the call.

core/src/main/java/com/graphhopper/reader/OSMTurnRelation.java
sguill: ok

easbar:
Thanks a lot!

Issue #1687, opened 2019-08-18
Routing issues with Landmarks


easbar:
As noticed in #1665 I found some cases where A-Star with LM does not find the shortest path. I tried to figure out why and here: https://github.com/graphhopper/graphhopper/tree/lm_issue I created a similar test that does not use the new feature of restricted source/target edges as in #1665 (and so far does not use edge-based routing either) and the tests still fail in some cases.

There might be different issues at play:

This test: https://github.com/graphhopper/graphhopper/blob/bfc6878df50bef8d3b0c732f0493723ac341de6d/core/src/test/java/com/graphhopper/routing/BidirectionalRoutingTest.java#L128 seems to fail, because the `LMApproximator` uses its `fallBackApproximator`, but the `toNode` of the `fallBackApproximator` is never set leading to an overestimation of the minimum distance. This might be easy to fix and we should do first.

This one: https://github.com/graphhopper/graphhopper/blob/bfc6878df50bef8d3b0c732f0493723ac341de6d/core/src/test/java/com/graphhopper/routing/BidirectionalRoutingTest.java#L181 also fails with a very large weight difference compared to Dijkstra, but I do not know why yet (have not really looked into it).

Not sure if this is related to #1557.

TODOs:

 - [ ] fix bug(s)
 - [ ] include tests for LM that were introduced in #1665

karussell:
I have fixed some issues in this branch: https://github.com/graphhopper/graphhopper/tree/lm_bug_1687

 - [x] your mentioned missing initialization of the fallbackApproximator
 - [x] separate factor for from and delta
 - [x] use double instead int for weights for cleaner API and virtual node addition (otherwise unclear which factor applies)
 - [x] skip infinite weight when selecting maximum weights of all landmarks
 - [x] set unclear subnetwork for smaller subnetworks (and handle smaller subnetworks first <-> sort list). This avoids that a landmark is created in such a smaller subnetwork leading to many maxed out weights as one direction is not reachable to or from the bigger subnetwork
 - [ ] try fallbackApproximator if weight is 0 or less
 - [ ]  we could explore the graph from the landmarks and use the maximum weight there to call setMaximumWeight, instead of the guess based on the area

[...]

PR #1745, opened 2019-10-07
Landmark bugs

michaz:
Minimal fix for the two failing tests in #1687.

The first was already discussed (missing setter for the 'fallback' approximator).

The second one was due to an incorrect treatment of "minned out" "delta" weights. See comment in code.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/lm/LMApproximator.java
+++ b/core/src/main/java/com/graphhopper/routing/lm/LMApproximator.java
@@ -194,13 +194,11 @@ public class LMApproximator implements WeightApproximator {
         return maxWeightInt;
     }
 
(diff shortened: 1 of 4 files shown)


michaz:
I understand you are already investigating creating an extra `factor` for the delta weights, doing performance tests etc., but IMO step one is maintaining correctness (in this case doing the right thing when a delta weight is minned out). Then, step two is seeing how we can make sure this happens less often.

michaz:
I think (?) one issue here was the test -- there was a "setGet"-test, but it was rather technical and white-box -- it was basically testing that the bits would be twiddled in the same way the code says they would be twiddled. :-) But the important property -- that the remaining weight must never be over-estimated -- was not tested.

PR #1747, opened 2019-10-13
Landmarks+QueryGraph bug

michaz:
I think this was all about finding the _closest_ non-virtual node to the destination node, i.e. with Dijkstra, instead of guessing. Actually, it's less code than the workaround.

Correcting for the edge length is not necessary for correctness, I think [citation needed]. And I'm sure it's not necessary for the _current_ node `v` (we can just return 0 if it's virtual*). If anything, something has to be done for the _destination_ node (`a` or `b`), but so far I don't see anything failing.

*) That shouldn't be a performance problem. It should just mean that the algorithm behaves like Dijkstra (approximation 0) until it has expanded the virtual edges around `a` and `b`, and when it finds virtual edges on the way, it should just "explode" into them before doing anything else. Seems safe.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/lm/LMApproximator.java
+++ b/core/src/main/java/com/graphhopper/routing/lm/LMApproximator.java
@@ -119,23 +119,19 @@ public class LMApproximator implements WeightApproximator {
         if (!doALMRecalc && fallback || lms.isEmpty())
             return fallBackApproximation.approximate(queryNode);
 
-        int towerNode = queryNode;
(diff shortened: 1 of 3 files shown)


michaz:
Oops, perhaps we have to tell Dijkstra not to allocate big maps for a small search (same as with MapMatching a year ago).

michaz:
Btw, this is more important than it looks: Being able to do the right thin when no landmark distance is available for a node is an ingredient of things to come, like Core-ALT.

PR #1749, opened 2019-10-13
Makes all tests pass for LM

michaz:
1. Sorry, but this reverts #1376. I just couldn't get it to work. Like others have pointed out, it _may_ be possible, but the numerics involved are totally not clear to me, and in its current state, it's demonstrably wrong.

2. I also add a treatment of the virtual edge at the target node in the LM-approximator[1]. This is different to what was there before. Before, the _current_ node got a special treatment, which is _not_ necessary.

I now also test _unidirectional_ LM, which catches (2). In bidirectional LM, apparently the stopping criterion is loose enough so (2) doesn't show up.

EDIT (added):

3. `LMApproximator` now rounds down, so it always underapproximates.

4. It also tells the bidirectional A* algorithm that it needs a certain slack for the stopping criterion. This is necessary _in addition_ to 3, because `LMApproximator`'s rounding makes it violate the triangle inequality ("consistency"), and this probably can't be fixed locally in the approximator.

5. I shifted the `BalancedWeightApproximator` (which also has a new name) by constants so that both sides are '0' at their destinations. (Recommended by one of the papers, mainly for clarity, doesn't change the algorithm's behavior if the stopping criterion (6.) is also adapted.)

6. I changed the stopping criterion in A* so it works with 5. (found the new formulation in one of the papers, and couldn't find the previous one; possibly the previous one was correct for the previous approximator, but I suspect it was slightly wrong). 

Supercedes #1747.

[1] Goldberg, A.V., Kaplan, H., Werneck, R.F.: Better Landmarks Within Reach. In: Demetrescu, C. (ed.) WEA 2007. LNCS, vol. 4525, pp. 38–51. Springer, Heidelberg (2007)


michaz:
This still doesn't mean the rounding errors discussed in #1557 are gone. Maybe, maybe not.

karussell:
I removed the todonow and replaced CarFlagEncoder with MotorcycleFlagEncoder. Then to remove several log warnings I changed the default maximum weight, but still I get exceptions ala:

```
randomGraph_withQueryGraph seed: 96076558697230
expected: [51, 44, 14, 5, 38, 22, 35, 50]
given:    [51, 44, 14, 8, 17, 32, 35, 50]

java.lang.AssertionError: wrong weight: 51->50, expected: 161.94306753707676, given: 162.04722980980404

	at org.junit.Assert.fail(Assert.java:88)
	at com.graphhopper.routing.weighting.RandomizedRoutingTest.comparePaths(RandomizedRoutingTest.java:322)
	at com.graphhopper.routing.weighting.RandomizedRoutingTest.randomGraph_withQueryGraph(RandomizedRoutingTest.java:280)
	at sun.reflect.GeneratedMethodAccessor2.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
```

Is this expected as those errors are not fixed in this PR (e.g. still rounding errors) or is this unexpected?

michaz:
> Then to remove several log warnings I changed the default maximum weight

Careful please, I would say that's not really a minor thing when the PR is about landmark precision.

I still stand by this:

> This still doesn't mean the rounding errors discussed in #1557 are gone. Maybe, maybe not. But they're not what has been failing the randomized tests.

So the answer is apparently 'not'. But perhaps now the change that is discussed in #1557 can be applied? (I wanted to remove only one rounding error source at a time..)

And I put the unidirectional LM test back in, because

> I now also test unidirectional LM, which catches (2). In bidirectional LM, apparently the stopping criterion is loose enough so (2) doesn't show up.

michaz:
Maybe we should remove the warning? A too large "max value" is much worse than a too small one, it will just quietly produce wrong routes.

michaz:
We should write an `LMApproximatorTest` that verifies that the `LMApproximator` _always underestimates_ the remaining weight, and then go through the code until it does.

karussell:
> Careful please, I would say that's not really a minor thing when the PR is about landmark precision

Isn't the underlying problem here more that the max weight guessing procedure is not good enough? I.e. it relies on some distances instead of the weights? We could try to make this better and e.g. go through the graph "from one end to the other". This could be easily done when selecting the first landmark or first 2 landmarks for example.

> encoder = new MotorcycleFlagEncoder(5, 5, 1);
> // todonow: make this work with speed_both_directions=true!
> encoder = new CarFlagEncoder(5, 5, 1);

I replaced this with the MotorcycleFlagEncoder because it uses `speed_both_directions=true` and the todonow could be removed. In one branch I made this easy for CarFlagEncoder too, so we could also wait for that change. But in several tests MotorcycleFlagEncoder is already used in this case already.

> This still doesn't mean the rounding errors discussed in #1557 are gone. Maybe, maybe not. But they're not what has been failing the randomized tests.

Probably I misunderstood this part. The randomized tests are failing occasionally now. How would applying the fix proposed in #1557 help with them, when you say they are not related to the random test failures?

I think it is a good idea to apply the fix and yes, let's do this in a separate PR.

michaz:
> The randomized tests are failing occasionally now.

Are they randomly failing on this branch in this state, without `setMaxValue`?

My current hypothesis is that the remaining random failures are due to #1557, and they don't happen with a low `maxValue` and happen more often, or always, with a high `maxValue`.

michaz:
> Isn't the underlying problem here more that the max weight guessing procedure is not good enough?

I don't know! Is the rounding supposed to be always correct as long as the `maxValue` is really the maximum value, or is it still heuristic even then?

michaz:
Even when I put back ` lm.setMaximumWeight(1000);`, which is currently in there on master (not 10000), I can run it a lot of times without failing.

So what I meant was that the troubles that we were having back then (it's already a long time ago..) with randomly failing tests on master should be fixed with this PR.

michaz:
Now the API client live test is failing. ,-)

michaz:
So this is the test that tests if `LMApproximator` is actually a valid A*-Approximator.

It systematically overestimates. The better the landmarks, the more.

michaz:
Oh. I think the case where it overestimates by a tiny bit is actually the case where it doesn't use landmarks at all, but the beeline fallback approximator.

michaz:
Done.

easbar:
Very nice! Now that we were on it I added unidir Dijkstra&AStar as well to the randomized test. There seem to be some rare cases with failing AStar/LM still, but the weight difference always seems to be very small, e.g.

```
expected: [19, 16, 35, 38]
given:    [19, 61, 81, 3, 28, 43, 79, 38]
seed: 39316239491387

java.lang.AssertionError: wrong weight: 19->38
expected: 106.36985678961038
given:    106.7326455944056
seed: 39316239491387
```

Looks like there are two paths with very similar weight. Is this expected ? Maybe we can increase the weight tolerance for LM/AStar a bit so we can differentiate between real bugs and such small errors (in case its not possible or not worth the effort to get rid of) ?

easbar:
> except for non-Euclidean edges; 

`buildRandomGraph` does this:

```java
  double fwdSpeed = 10 + random.nextDouble() * 120;
  double bwdSpeed = 10 + random.nextDouble() * 120;
```

So the speed  might turn out as 130 even though the motor cycle encoder has max speed 120 (?). Is this what you mean with non-euclidean edges ? It certainly seems to be problematic for AStar/LM.

>  it doesn't use landmarks at all, but uses the beeline fallback approximator,

Did you check / can we be sure the random tests cover cases with the LM *and* the fallback approximator ?

michaz:
Thanks!!

> Looks like there are two paths with very similar weight. Is this expected ? Maybe we can increase the weight tolerance for LM/AStar a bit so we can differentiate between real bugs and such small errors (in case its not possible or not worth the effort to get rid of them) ?

If I'm not mistaken, the `BeelineApproximator` _cannot_ be a reliable underapproximator at the moment. It doesn't know how much slack it has to subtract from its approximation to correct for rounding, because that depends on the number of edges of the path it will be compared against (because the link of every edge could be rounded down), and it doesn't know that.

I would say that's a real bug (in our engineering), and we must sort it out.

michaz:
Arrgh, those are _really_ good test cases. They keep failing even when I set the beeline approximator to 0 and let the LM approximator round down by more than the number of terms that are rounded. :-)

easbar:
I think its most likely because of the distance/speed deviations we **explicitly** add in `GHUtility#randomGraph`, don't you think ?

michaz:
Err, ah, okay, wait.

michaz:
Nope, I can edit MotorCycleFlagEncoder and change maxSpeed to 200 and it still fails.

easbar:
Ok and did you remove the code that manipulates the distance ?

michaz:
> Hmm or maybe not because it only increases distance

Exactly. Strangely, it _does_ make fail1 and fail2 pass, but it _doesn't_ make it pass the LMApproximatorTest when I put it in there.

easbar:
> Strangely, it does make fail1 and fail2 pass,

Be careful here, if you comment out the `nextDouble()` calls the same seed might produce another graph / example

core/src/main/java/com/graphhopper/routing/lm/PerfectApproximator.java
```diff
@@ -0,0 +1,62 @@
...
+        this.weighting = weighting;
+        this.traversalMode = traversalMode;
+        this.reverse = reverse;
+    }
+
+    @Override
+    public double approximate(int currentNode) {
+        Dijkstra dijkstra = new Dijkstra(graph, weighting, traversalMode);
+        Path path = reverse ? dijkstra.calcPath(to, currentNode) : dijkstra.calcPath(currentNode, to);
+        if (path.isFound()) return path.getWeight(); else return Double.POSITIVE_INFINITY;
```
karussell: really only btw :) .... wouldn't 
`return path.isFound() ? path.getWeight() : Double.POSITIVE_INFINITY;`
better as you used the ternary op in the line above.

core/src/test/java/com/graphhopper/routing/lm/LMApproximatorTest.java
```diff
@@ -0,0 +1,181 @@
...
+    @Repeat(times = 5)
+    public void randomGraph() {
+        final long seed = System.nanoTime();
+        System.out.println("random Graph seed: " + seed);
+        run(seed);
+    }
+
+    @Test
+    public void fail7() {
+        run(337586249813919L);
```
easbar: I'd rather not have these tests depend on a specific seed, because this way they are strictly coupled to the exact usage of the random number generator in `GHUtiliy#randomGraph`. The lazy method would be to duplicate the randomGraph method in this class and never change it again. The better method would be writing down the corresponding graph explicitly (there is GHUtiliy#printGraphForUnitTest to do this quickly).

core/src/test/java/com/graphhopper/routing/weighting/RandomizedRoutingTest.java
```diff
@@ -281,19 +301,52 @@ public void randomGraph_withQueryGraph() {
...
     }
 
+    @Test
+    public void fail1() {
+        run(39457616822163L);
+    }
+
+    @Test
+    public void fail2() {
+        run(39386680060285L);
```
easbar: Same as in aprroximator test, better not depend on these seeds here

core/src/test/java/com/graphhopper/routing/lm/LMApproximatorTest.java
michaz: Thanks, I agree. The even lazier and almost-as-good method would be to remove maybe all but one of them completely. They're reasonably well covered by the general random test, I think.

PR #1751, opened 2019-10-15
Extract GraphModification from QueryGraph

easbar:
Here I have separated all the data that is needed to determine how the `EdgeExplorer`s of `QueryGraph` have to be modified (compared to the underlying `Graph`) when we insert virtual nodes from `QueryGraph`. Or, to put this in other words: So far we put a `List<QueryResult>` into `QueryGraph.lookup()` and got a `QueryGraph` that acts like a `Graph` with additional virtual nodes. With this PR we put a `List<QueryResult>` into `GraphModificationBuilder`, get back a `GraphModification` and use this to realize `QueryGraph`. Why ? Because I want to use `GraphModification`, but not `QueryGraph`. 
So far all this is (package) internal (in a new `routing.querygraph` package).

Possible (future) improvements: 

- [ ] do not create `GraphModification` *inside* `QueryGraph`, but rather pass it to it, so `QueryGraph` can deal with (more general/different (?)) modifications
- [ ] generalize `QueryGraph` to something like `WrapperGraph` (maybe thats the same as above)


core/src/main/java/com/graphhopper/routing/querygraph/QueryGraph.java
```diff
@@ -592,110 +309,61 @@ public EdgeExplorer createEdgeExplorer(final EdgeFilter edgeFilter) {
         }
     }
 
-    private EdgeExplorer createUncachedEdgeExplorer(EdgeFilter edgeFilter) {
-        // Iteration over virtual nodes needs to be thread safe if done from different explorer
-        // so we need to create the mapping on EVERY call!
-        // This needs to be a HashMap (and cannot be an array) as we also need to tweak edges for some mainNodes!
-        // The more query points we have the more inefficient this map could be. Hmmh.
```
easbar: This comment is the reason why I did not merge this PR yet. I dont understand it :) First, we are creating `QueryGraph` per request and we do not use concurrent/multi-threaded edge explorers on the same `QueryGraph`. Second, I did not understand why we built a map of `VirtualEdgeIterator` (with the mutable `current` indices). Why not just a map of `VirtualEdgeIteratorState` ? (ok also not immutable because of `unfavored` flags, but almost..).

core/src/main/java/com/graphhopper/routing/querygraph/QueryGraph.java
```diff
@@ -592,110 +309,61 @@ public EdgeExplorer createEdgeExplorer(final EdgeFilter edgeFilter) {
...
-        final IntObjectMap<VirtualEdgeIterator> node2EdgeMap
-                = new GHIntObjectHashMap<>(queryResults.size() * 3);
-
+    private EdgeExplorer createUncachedEdgeExplorer(final EdgeFilter edgeFilter) {
+        // build a map of virtual edge lists for every node that is modified compared to the mainGraph. this map
+        // depends on the given edgeFilter, but we could just as well build this map independent from the edge filter
+        // and apply the filter while iterating the edges
```
easbar: I think applying the filter while iterating the edges would be cleaner, performance-wise I do not expect a difference ?

core/src/main/java/com/graphhopper/routing/querygraph/QueryGraph.java
karussell: Good questions :) I think the "thread-safety" is regarding all downstream read-operations (not used here). In this case it was the multi-threaded matrix algorithm but could be also multi-threaded bidirectional dijkstra that I wanted to support.

Regarding the VirtualEdgeIteratorState: how would you implement the custom EdgeExplorer with this?

core/src/main/java/com/graphhopper/routing/querygraph/QueryGraph.java
karussell: What would be the advantage - less memory used?

core/src/main/java/com/graphhopper/routing/querygraph/QueryGraph.java
easbar: > I think the "thread-safety" is regarding all downstream read-operations (not used here).

Ok that explains it for me / makes sense

> Regarding the VirtualEdgeIteratorState: how would you implement the custom EdgeExplorer with this?

I think the main thread-safety issue in master is that we build a map of `VirtualEdgeIterator` which is just a `List<VirtualEdgeIteratorState>` and an index (`current`). So there is lots of mutable state because of the `current` indices: When one thread calls `next()` on some iterator it will affect the behavior of the same iterator in another thread. This is problem can be easily solved: Just use a map of `List<VirtualEdgeIteratorState>` (without the `current` indices) and keep the `current` indices within the scope of the single threads (one `VirtualEdgeIterator` per thread, but they all share the `List<VirtualEdgeIteratorState>`s).  You can see this in this PR, I have already done this.
If these lists were immutable the code would be thread-safe. Unfortunately the  `VirtualEdgeIteratorState`s are not immutable, so we would either have to work towards making them immutable or use them in a thread-safe way (do not change any of the mutable state).

easbar: Btw I think the query graph cache is also not thread-safe ?

core/src/main/java/com/graphhopper/routing/querygraph/QueryGraph.java
easbar: yes but that does not matter anyway. I dont know it just seems ugly to build this map beforehand for every edge filter. And maybe we can even get rid of query graph cache ?

core/src/main/java/com/graphhopper/routing/querygraph/QueryGraph.java
karussell: > Btw I think the query graph cache is also not thread-safe ?

Oh, yes. 

> You can see this in this PR, I have already done this.

Ok, so this means QueryGraph (incl. the cache) is finally read-thread safe :) (?)

IMO everything in GraphHopper should be read-thread safe and write thread-safe is currently forced via locking: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/GraphHopper.java#L965

core/src/main/java/com/graphhopper/routing/querygraph/QueryGraph.java
karussell: Yes, definitely ugly. The advantage is a huge boost to the performance for e.g. map matching and also matrix algorithms.

easbar: > Yes, definitely ugly. The advantage is a huge boost to the performance for e.g. map matching and also matrix algorithms. (Basically every algo that requires many snaps/locations)

I dont know. The bottleneck for many snaps/locations is the LocationIndex lookup, not building the QueryGraph. The query graph cache I think is an improvement because we do not need to build the map for every edge explorer (saves memory but does it really speed up the request?). But now that we build most of the map beforehand I do not expect a performance issue if we apply the filter while iterating. But will try (better than thinking :))

core/src/main/java/com/graphhopper/routing/querygraph/QueryGraph.java
easbar: > Ok, so this means QueryGraph (incl. the cache) is finally read-thread safe :) (?)

Hmm in which sense do you think it was not read-thread safe before ?

karussell: I just misinterpreted the code and thought the iterator.reset was no longer used. So I think it is still not "read-thread-safe" in the sense that the iterator safes state (getCurrentEdge)

easbar: Yes the iterator safes state (it has to), but the `currentEdge` is no longer part of the map. Its thread-safe in the sense that every thread gets its own iterator instance (and hence its own `currentEdge` index).

core/src/main/java/com/graphhopper/routing/querygraph/QueryGraph.java
karussell: > The bottleneck for many snaps/locations is the LocationIndex lookup, not building the QueryGraph.

Ups, yes. In my head this is still tightly coupled :)

easbar: Ok I have changed it such that QueryGraph first builds a map of all virtual edge states and virtual edge iterator filters the edges while calling next. Now this is as fast (if not faster) as master, even without the query graph cache. I think we can remove the query graph cache entirely (in a separate PR/commit).

I ran a little measurement with via routing for 5000 points for Germany (with and without query graph cache). With the cache only 5 edge explorers are created, without there are 24995, but since they are rather small objects and performance is no longer improved by the cache I dont think the added complexity of query graph cache is really worth it after this PR).

| branch | time routing | time query graph explorer building |
|-|-|-|
|master|270s|250s|
|master, cached|16s|50ms|
|query_graph_refactoring (this PR)|15s|2ms|
|query_graph_refactoring (this PR), cached|15s|1ms|

karussell: Wow, this is rather unexpected and really nice of course :)

I fear I still do not understand the reason. What exactly was slow in uncached master? I always thought it was the EdgeExplorer creation :)

easbar: > I dont know. The bottleneck for many snaps/locations is the LocationIndex lookup, not building the QueryGraph. The query graph cache I think is an improvement because we do not need to build the map for every edge explorer (saves memory but does it really speed up the request?).

I think looking at my little table I have to correct this: For many locations the edge explorer creation indeed was a bottleneck (and query graph cache fixed this). The interesting question here however is what 'edge explorer creation' means. In previous (uncached) master every time we called `createEdgeExplorer` we built a mapping between node ids and `VirtualEdgeIterator` for all virtual nodes and all real nodes adjacent to real nodes. (for 5000 locations this would be something like 15000). I think the only reason we had to rebuild this mapping everytime is because `VirtualEdgeIterator` stores the (mutable) `currentIndex`. What I changed in this PR is that I just build the mapping once at the time the `QueryGraph` is built, but instead of a map of `VirtualEdgeIterator` I create a map of (more or less immutable) `List<VirtualEdgeIterator**State**>`. 

> What exactly was slow in uncached master? I always thought it was the EdgeExplorer creation :)

Building the `IntObjectMap<VirtualEdgeIterator>`.

Another improvement I did is that I no longer put all virtual nodes and adjacent real nodes into the same mapping, but only the real nodes. The virtual nodes can be retrieved from a list, because all the virtual node ids are consecutive. This is a small improvement, because the hashmap stays a bit smaller.

PR #1755, opened 2019-10-19
fix sort memory usage

### LINKED ISSUE #1754: Old graph not garbage-collected after sort
When importing with `graph.do_sort: true`, memory usage is much higher for all steps after the sort. I traced this down to the old graph not getting garbage-collected because `OSMReader` holds a reference to the old graph, which is still on the stack outside of the try/catch in `GraphHopper.process`. I'll post a PR shortly with a fix that worked for me.

msbarry:
Fixes #1754 by using `OSMReader` from a separate function so it is no longer referenced on the stack and can get garbage-collected when `postProcessing` sorts the graph. Ran this on the world and memory was ~20GB before sorting, 40GB during the sort, and back down to ~20GB shortly after the sort.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -635,15 +635,7 @@ public class GraphHopper implements GraphHopperAPI {
                     throw new RuntimeException("To avoid multiple writers we need to obtain a write lock but it failed. In " + graphHopperLocation, lock.getObtainFailedReason());
             }
 
-            try {
-                DataReader reader = importData();
-                DateFormat f = createFormatter();
-                ghStorage.getProperties().put("datareader.import.date", f.format(new Date()));
-                if (reader.getDataDate() != null)
-                    ghStorage.getProperties().put("datareader.data.date", f.format(reader.getDataDate()));
-            } catch (IOException ex) {
-                throw new RuntimeException("Cannot read file " + getDataReaderFile(), ex);
-            }
+            readData();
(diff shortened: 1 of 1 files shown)


karussell:
Thanks a lot! This is especially nice as it solves another headache that we occasionally face: we use memory mapping for elevation and these files are not freed even while CH preparation leading to crashes due to no left memory.

Would you mind to update your github nickname, as the current one does not seem valid:

> msb5014, fixes like #1733

michaz:
Woohoo, nice catch. I would have thought "only variable holding reference to X goes out of scope" == "X becomes GCable". Apparently not so.

easbar:
me too :) can you explain this ?

michaz:
No, sorry. I only found something to the effect of "this is mostly so, but not always, and in the case of a try-catch-block it is *not* so, _unless_ you actually go into the catch".

Apparently, according to the specification, variable scope and GC have nothing at all to do with each other. The GC decides when it GCs, period.

karussell:
Indeed it is a bit unintuitive as try-catch already defines a scope - maybe not as strong (for GC) like a method?

We certainly observed memory problems in CH preparation that showed mmap files for elevations not collected from GC - have not tried yet if they persist.

> Apparently, according to the specification, variable scope and GC have nothing at all to do with each other. The GC decides when it GCs, period.

Uh, yeah. GC is very lazy at times ;)

karussell:
I think the main point is that try-catch has a different scope for variables, but does not release the stack-allocated objects. This is only done per-method.

Issue #1398, opened 2018-06-16
error "The turn cost storage must be initialized only once."


develar:
Version: 0.11.0-pre1

In my config: `graph.flag_encoders: car|turn_costs=true,bike2,racingbike,mtb,hike`

```
+ ./graphhopper.sh -a import -i /Volumes/Pegasus/maps/australia-latest.osm.pbf -c /Users/develar/Documents/gh-routing-data/config.yml
## using java 1.8.0_172 (64bit) from 
## using existing osm file /Volumes/Pegasus/maps/australia-latest.osm.pbf
## existing jar found web/target/graphhopper-web-0.11-SNAPSHOT.jar
## now import. JAVA_OPTS=-Xmx15g -Xms10g -XX:+UseG1GC
INFO  [2018-06-16 06:13:58,777] com.graphhopper.reader.osm.GraphHopperOSM: version 0.11|2018-06-07T15:12:38Z (5,14,4,3,3,2)
INFO  [2018-06-16 06:13:58,782] com.graphhopper.reader.osm.GraphHopperOSM: graph CH|car,bike2,racingbike,mtb,hike|RAM_STORE|3D|turn_cost|,,,,, details:edges:0(0MB), nodes:0(0MB), name:(0MB), geo:0(0MB), bounds:1.7976931348623157E308,-1.7976931348623157E308,1.7976931348623157E308,-1.7976931348623157E308,1.7976931348623157E308,-1.7976931348623157E308, CHGraph|fastest|car, shortcuts:0, nodesCH:(0MB), CHGraph|fastest|bike2, shortcuts:0, nodesCH:(0MB), CHGraph|fastest|racingbike, shortcuts:0, nodesCH:(0MB), CHGraph|fastest|mtb, shortcuts:0, nodesCH:(0MB), CHGraph|fastest|hike, shortcuts:0, nodesCH:(0MB)
INFO  [2018-06-16 06:13:58,825] com.graphhopper.reader.osm.GraphHopperOSM: start creating graph from /Volumes/Pegasus/maps/australia-latest.osm.pbf
INFO  [2018-06-16 06:13:58,826] com.graphhopper.reader.osm.GraphHopperOSM: using CH|car,bike2,racingbike,mtb,hike|RAM_STORE|3D|turn_cost|,,,,, memory:totalMB:10240, usedMB:19
[...]

PR #1757, opened 2019-10-20
Sort edges

### LINKED ISSUE #1756: Sort graph edges
Currently `graph.do_sort: true` only sorts the nodes of a graph, but in `GHUtility.createSortedGraph` it is also possible to sort graph edges. I use `MMAP` storage so implemented this thinking that sorting edges would batch disk reads closer together and reduce page faults when checking edges in the location index, getting edge details for the route, etc.  I did get a performance boost and reduction in average page faults per request and will post a PR shortly.

msbarry:
Fixes #1756 to sort edges in addition to nodes by iterating through edges to add them to the graph in the same order that we iterated through them during DFS. This improved some performance problems I was having (particularly around closest point lookup) and I started running this in production on 10/17.

There are a number of other changes included as well (ubuntu 16->18, graphhopper 0.8->0.13) but I do see a significant reduction in average major page faults per request which was the goal of this change:
image


msbarry:
Made some changes to simplify the approach and added an integration test for sorting the graph. Not sure about the build failure - seems to be unrelated?

karussell:
Thanks a lot, this look good!

msbarry:
Thanks! One issue I saw on the measurement tests is that it causes location index lookups to be ~2x slower, shuffling ids before storing in the index makes it faster again. Any idea why this might be causing that?

msbarry:
Actually after running `graphhopper.sh -a measurement` some more, it looks like there's a lot of variability in `location_index` measurements so not a meaningful difference with and without sorting edges.

core/src/main/java/com/graphhopper/util/GHUtility.java
```diff
@@ -317,32 +326,55 @@ public static Graph shuffle(Graph g, Graph sortedGraph) {
...
-        AllEdgesIterator eIter = fromGraph.getAllEdges();
-        while (eIter.next()) {
+    static Graph createSortedGraph(Graph fromGraph, Graph toSortedGraph, final IntIndexedContainer oldToNewNodeList, final IntIndexedContainer newToOldEdgeList) {
+        int edges = fromGraph.getEdges();
+        for (int i = 0; i < edges; i++) {
+            int edgeId = newToOldEdgeList.get(i);
+            if (edgeId < 0)
+                continue;
+
+            EdgeIteratorState eIter = fromGraph.getEdgeIteratorState(edgeId, Integer.MIN_VALUE);
```
easbar: Did you try how long the sorting takes now (say for a continental sized graph) ? Usually calling `getEdgeIteratorState` many times is not very efficient, because it creates a new object on every call and thus leads to lots of GC.

This goes beyond this PR, but I think its odd that the sorting methods are not part of the `(Base)Graph` API. For example if a `Graph` has a `TurnCostExtension` we need to sort the turn cost entries also, but given a `Graph` I cannot even know what kind of `GraphExtension` it is (see #1398). Similarly, I have to remember to sort the node coordinates here. More generally, I should not need to know all these details about the graph to sort it, the sorting details should be handled within the graph implementation.

core/src/main/java/com/graphhopper/util/GHUtility.java
```diff
@@ -317,32 +326,55 @@ public static Graph shuffle(Graph g, Graph sortedGraph) {
...
             }.start(explorer, startNode);
         }
-        return createSortedGraph(g, sortedGraph, list);
+        return createSortedGraph(g, sortedGraph, nodeList, edgeList);
     }
 
-    static Graph createSortedGraph(Graph fromGraph, Graph toSortedGraph, final IntIndexedContainer oldToNewNodeList) {
-        AllEdgesIterator eIter = fromGraph.getAllEdges();
-        while (eIter.next()) {
+    static Graph createSortedGraph(Graph fromGraph, Graph toSortedGraph, final IntIndexedContainer oldToNewNodeList, final IntIndexedContainer newToOldEdgeList) {
```
easbar: Why is it `oldToNewNodeList` vs. `newToOldEdgeList` ? Should it not be `oldToNewEdgeList` ? As far as I understand this naming it means that I can use an old edge id (from `fromGraph`) as index into this list to retrieve the edge id this edge should have in the new graph (`toSortedGraph`) -> 'old to new'

msbarry: It's `newToOldEdgeList` because we can only append edges to the new graph so we need to iterate through edges in the new order. This list lets us know which old edge we need to get data from for the new edge.  See line 354 where I set the values.

It's backwards  from nodes because `NodeAccess` provides an API to set a node with a given index. As far as I can tell `EdgeAccess` does not have a similar API.

core/src/main/java/com/graphhopper/util/GHUtility.java
msbarry: No, I originally implemented this and successfully ran it on the world by adding a capability to `AllEdgesIterator` to accept an edge ordering provider (`newToOldEdgeIndex`), but that diff got a lot larger (https://github.com/graphhopper/graphhopper/pull/1757/files/f6c4354d243330b2b9126eba7ff9404a324c9669) whereas this change was more self-contained and I assumed it was equivalent.

I could go back to that more complex approach if you think it would be better? Or add an API to `AllEdgesIterator` to set the edgeId?

I agree moving this into `(Base)Graph` would be good, but would involve changing more parts of the code than I'm familiar with at this point so probably better for another PR.

core/src/main/java/com/graphhopper/util/GHUtility.java
easbar: Ah yes sorry I should have looked closer. Yes you can only append edges.

core/src/main/java/com/graphhopper/util/GHUtility.java
easbar: Ok, no I do not think the more complex approach is needed at the moment. I will probably move the sorting into the graph for #1398, so I guess we can leave this as is and optimize in another PR.

msbarry: OK, just tested this out sorting the world. There were slightly more GC activity during the sort operation compared to modifying `AllEdgesIterator`(30 GC events vs. 23 - both with 56G RAM and java 8 with G1GC) and both finished in <200 seconds which seems reasonable.

verbose GC log from modifying `AllEdgesIterator`:
```
[GC pause (G1 Evacuation Pause) (young) 26717M->25106M(57344M), 0.0923583 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 27602M->25378M(57344M), 0.0365206 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0005099 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0466707 secs]
[GC remark, 0.0493437 secs]
[GC cleanup 25418M->25418M(57344M), 0.0036449 secs]
[GC pause (G1 Evacuation Pause) (young) 27954M->25779M(57344M), 0.0380569 secs]
[GC pause (G1 Evacuation Pause) (young) 28723M->26229M(57344M), 0.0444583 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 30501M->26892M(57344M), 0.0583504 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0008904 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0477114 secs]
[GC remark, 0.0490855 secs]
[GC cleanup 26932M->26932M(57344M), 0.0034052 secs]
[GC pause (G1 Evacuation Pause) (young) 31964M->27676M(57344M), 0.0723461 secs]
[GC pause (G1 Evacuation Pause) (young) 33820M->28649M(57344M), 0.0871456 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 35721M->29789M(57344M), 0.1029310 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0020978 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0483938 secs]
[GC remark, 0.0548950 secs]
[GC cleanup 29845M->29845M(57344M), 0.0035427 secs]
[GC pause (G1 Evacuation Pause) (young) 37789M->31077M(57344M), 0.1149464 secs]
[GC pause (G1 Evacuation Pause) (young) 39861M->32503M(57344M), 0.1292224 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 42039M->33986M(57344M), 0.1374616 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0032296 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0495207 secs]
[GC remark, 0.0535295 secs]
[GC cleanup 34034M->34034M(57344M), 0.0029495 secs]
[GC pause (G1 Evacuation Pause) (young) 47064M->36126M(57344M), 0.1380340 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 44638M->37473M(57344M), 0.1547417 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0043262 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0528809 secs]
[GC remark, 0.0627294 secs]
[GC cleanup 37529M->37529M(57344M), 0.0033535 secs]
[GC pause (G1 Evacuation Pause) (young) 47537M->39121M(57344M), 0.1481099 secs]
[GC pause (G1 Evacuation Pause) (young) 47825M->40709M(57344M), 0.1504248 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 48101M->42066M(57344M), 0.1330208 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0060035 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0527521 secs]
[GC remark, 0.0726408 secs]
[GC cleanup 42114M->42114M(57344M), 0.0032893 secs]
[GC pause (G1 Evacuation Pause) (young) 48274M->43156M(57344M), 0.1108996 secs]
```

verbose GC log with `getEdgeIteratorState`:
```
[GC pause (G1 Evacuation Pause) (young) 24677M->24164M(57344M), 0.1245786 secs]
[GC pause (G1 Evacuation Pause) (young) 26660M->25057M(57344M), 0.0654925 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 27553M->25211M(57344M), 0.0306106 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0011887 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0586427 secs]
[GC remark, 0.0597587 secs]
[GC cleanup 25323M->25323M(57344M), 0.0030835 secs]
[GC pause (G1 Evacuation Pause) (young) 27915M->25500M(57344M), 0.0259677 secs]
[GC pause (G1 Evacuation Pause) (young) 27996M->25790M(57344M), 0.0356115 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 28590M->26119M(57344M), 0.0519495 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0012558 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0372878 secs]
[GC remark, 0.0478013 secs]
[GC cleanup 26192M->26192M(57344M), 0.0031108 secs]
[GC pause (G1 Evacuation Pause) (young) 29655M->26520M(57344M), 0.0723304 secs]
[GC pause (G1 Evacuation Pause) (young) 30664M->26972M(57344M), 0.0831304 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 32220M->27550M(57344M), 0.1013616 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0016264 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0378618 secs]
[GC remark, 0.0491968 secs]
[GC cleanup 27624M->27624M(57344M), 0.0028098 secs]
[GC pause (G1 Evacuation Pause) (young) 33774M->28257M(57344M), 0.1214522 secs]
[GC pause (G1 Evacuation Pause) (young) 34465M->28932M(57344M), 0.0785504 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 36532M->29837M(57344M), 0.1537344 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0017540 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0390381 secs]
[GC remark, 0.0511714 secs]
[GC cleanup 29917M->29917M(57344M), 0.0028430 secs]
[GC pause (G1 Evacuation Pause) (young) 37213M->30694M(57344M), 0.0923386 secs]
[GC pause (G1 Evacuation Pause) (young) 43895M->32168M(57344M), 0.0942218 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 40136M->33114M(57344M), 0.1283859 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0031646 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0401712 secs]
[GC remark, 0.0556655 secs]
[GC cleanup 33194M->33194M(57344M), 0.0032600 secs]
[GC pause (G1 Evacuation Pause) (young) 44682M->34665M(57344M), 0.1275467 secs]
[GC pause (G1 Evacuation Pause) (young) 43769M->35967M(57344M), 0.1432414 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 45903M->37070M(57344M), 0.1242383 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0043771 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0421382 secs]
[GC remark, 0.0631872 secs]
[GC cleanup 37158M->37158M(57344M), 0.0029976 secs]
[GC pause (G1 Evacuation Pause) (young) 47742M->38286M(57344M), 0.1206002 secs]
[GC pause (G1 Evacuation Pause) (young) 48158M->39485M(57344M), 0.1221284 secs]
[GC pause (G1 Evacuation Pause) (young) (initial-mark) 48445M->40676M(57344M), 0.1215852 secs]
[GC concurrent-root-region-scan-start]
[GC concurrent-root-region-scan-end, 0.0056251 secs]
[GC concurrent-mark-start]
[GC concurrent-mark-end, 0.0414679 secs]
[GC remark, 0.0657385 secs]
[GC cleanup 40772M->40772M(57344M), 0.0029433 secs]
[GC pause (G1 Evacuation Pause) (young) 48436M->41736M(57344M), 0.1135536 secs]
[GC pause (G1 Evacuation Pause) (young) 48568M->42550M(57344M), 0.0950881 secs]
```

easbar:
Don't we have to sort the geometry and the name index as well if we change the edge ids ? For the turn costs I am sure this needs to be sorted (but at least this is optional and throws an error if `turn_costs=true` and we use sorting at the moment).

karussell:
> I do see a significant reduction in average major page faults per request which was the goal of this change

What exactly does this improve or what problems did you have?

I (quickly) tested this on my laptop against [sachsen.pbf](https://download.geofabrik.de/europe/germany/sachsen-191026.osm.pbf) but I wasn't able to see an improvement regarding speed for prepare.lm.time, prepare.ch.time, routingLM4.mean, routingCH.mean or routing.mean.

Or do you see improvements only for larger areas like country or continental size?

msbarry:
> What exactly does this improve or what problems did you have?
> 
> I (quickly) tested this on my laptop against [sachsen.pbf](https://download.geofabrik.de/europe/germany/sachsen-191026.osm.pbf) but I wasn't able to see an improvement regarding speed for prepare.lm.time, prepare.ch.time, routingLM4.mean, routingCH.mean or routing.mean.
> 
> Or do you see improvements only for larger areas like country or continental size?

I noticed the improvements when serving using MMAP for the world.  The main thing that I saw improve was location index lookups in dense areas, I'm assuming because this needs to check a lot of edge geometries and properties which are located closer together on disk when the edges are sorted.

easbar:
Should we: 

1) merge this 
2) move the graph sorting inside the graph implementation (graph.sort())
3) make sure turn cost tables are also sorted when graph.sort() is called so we can fix #1398 
4) somehow make sure the graph cannot be sorted after certain operations (for example sorting the base graph after CH preparation will not work, because we are referencing the base edge ids there also). Also I see no use-case for re-sorting the (base)graph after the initial import.

or do the same, but in another order (start with 2)) ? I think it would be more logical to start with 2), but I also see no real reason why not to merge this first.

karussell:
Have tested this for a small area with MMAP and I saw no improvements. Still this change seems logical and small enough to me that we should apply it. 

@easbar I have no preference

msbarry:
I'd probably err on the side of changing functionality in a separate commit from refactoring. This review adds a unit test to verify edge sorting behavior and integration test to make sure sorting works end-to-end which should still pass (but may move) with the refactor in 2.

> 4. somehow make sure the graph cannot be sorted after certain operations (for example sorting the base graph after CH preparation will not work, because we are referencing the base edge ids there also).

I think the sorting a prepared CH case has an explicit check but there may be other cases to add.

https://github.com/graphhopper/graphhopper/blob/bca912afe840fac2536f4fb27b9c9dfdcfd9e6ba/core/src/main/java/com/graphhopper/GraphHopper.java#L845-L846

> Also I see no use-case for re-sorting the (base)graph after the initial import.

I have done this, but only to test changes to the sort utility :-)

easbar:
Ok lets merge this then

easbar:
Thanks!

PR #1759, opened 2019-10-21
Implement importAndClose API

### LINKED ISSUE #1758: Free resources early while importing
When importing, all resources are currently held in memory until being flushed at the very end. It is possible to flush several resources early that are not needed in subsequent steps (name index, edge geometries, finished CH graphs) to reduce memory usage. I will post a PR in a few minutes with a proposal to allow early-flushing of resources on import via an opt-in flag.

msbarry:
Fixes #1758 by flushing the name index, location index, edge geometries, and finished CH graphs to disk as soon as they are no longer needed during the import process when you set `graph.conserve_memory` flag to true. I used this while importing the world with with 3 CH weightings and post-gc memory usage peaked at 40GB during the sort, then remained <40GB for the remainder of the import.


karussell:
Are you sure that this is really worth the effort? One important aspect is also to keep things simple. 

If there is not enough memory why not use `graph.dataaccess: MMAP_STORE`?

(With https://github.com/graphhopper/graphhopper/commit/127b6fb8dc6a42b89928b79ddff5a8467f2c88ca and JDK13 an import with mmap is comparable fast)

msbarry:
@karussell I agree it may not be worth the extra complexity. There are a few caveats too like the graph won't be usable for routing until it's loaded from disk and you may break people using the flag if you add a feature to use the unloaded parts again later in the import process.

The main reason to include this is that for people who run the import on a VPS like me, each hour the import runs costs money so it can become expensive to debug OOM exceptions that occur late in the import process as all of the finished CH graphs are loaded in memory. With this change, adding 10 more CH profiles won't increase peak memory usage while importing. I had thought that MMAP would be prohibitively slow on import, I'll try it out with that improvement next time I need to regenerate.

This is one of the 3 main changes I made in my fork to get the world importing successfully (this, #1757 and #1755) - I thought I'd offer them back in case they could help anyone else. Happy to track separately if you want to discard though.

karussell:
> With this change, adding 10 more CH profiles won't increase peak memory usage while importing

How are you actually serving routes if memory is not sufficient for all CH profiles at once?

> I had thought that MMAP would be prohibitively slow on import, I'll try it out with that improvement next time I need to regenerate.

Thanks, please report your findings back. Also make sure you are using a SSD :)

> This is one of the 3 main changes I made in my fork to get the world importing successfully (this, #1757 and #1755) 

Yes, they are very much appreciated. Let's also wait for other devs regarding this PR here.

msbarry:
>
> How are you actually serving routes if memory is not sufficient for all CH profiles at once?
> 

I import using `RAM_STORE` and serve using `MMAP`

easbar:
Thanks for providing your modifications to GH! For the upstream project here I am wondering would it not be more useful to provide a specific API that is intended for *just* the import (including the optimization/memory release you are doing here) instead of using `importOrLoad` which is supposed to import *and* load everything into memory, but then provide a flag that removes it from memory again ?

msbarry:
> Thanks for providing your modifications to GH! For the upstream project here I am wondering would it not be more useful to provide a specific API that is intended for _just_ the import (including the optimization/memory release you are doing here) instead of using `importOrLoad` which is supposed to import _and_ load everything into memory, but then provide a flag that removes it from memory again ?

I think that makes sense, I could change the implementation here.  What do you think the `import` method should do if the graph already exists? Blow it away, or exit early?

easbar:
> What do you think the import method should do if the graph already exists? Blow it away, or exit early?

Hm good question. I guess for consistency with `importOrLoad` the new import method should either do the import or do nothing (in case the files exist already). Or even throw an error if the files exist already ? To overwrite the existing files we could add a `graph.clean` flag that is used by both `importOrLoad` and the new import (But not sure if that is needed and this is maybe for another issue). 

In any case we should be careful with the naming, because there will be a difference between the current 'import' (which basically reads OSM, writes the GH files and loads everything into memory) and the new 'import' (which only reads OSM and writes the GH files).

msbarry:
@easbar thanks for the suggestions, I implemented an `importAndClose` method and changed `ImportCommand` to call that - let me know what you think.

karussell:
Did you have a chance to compare the time it takes in your case with MMAP vs. this PR vs. the MMAP commit I linked?

I'm still unsure about this PR, because there are two use cases and this PR introduces a 3rd that I do not understand (yet):

 * either you need it fast and you use RAM_STORE. This requires plenty of RAM because you usually also process the CH profiles in parallel, because you want it fast! In this case free early does not make much sense and releasing the wayGeometry releases 10GB, yes, but as you will have >>120GB RAM this is not a big win for the added complexity (?)
 * or you want to just "cheaply" process it with minimal RAM. For that you have MMAP, which does all this memory management under the hood without exposing much of our internals

msbarry:
Not yet, I will test out importing a medium size extract (maybe North America?) with RAM vs MMAP and let you know.

With the latest version, it removes the third way of importing so there are only the 2 you mention. The only change is that the RAM version flushes parts to dish as soon as it’s done generating them, users don’t need to know anything about it, except maybe that the directory will have partial results if it fails partway through.

msbarry:
> Did you have a chance to compare the time it takes in your case with MMAP vs. this PR vs. the MMAP commit I linked?

I can test it on a larger extract like North America (8.7GB) which imports fine in-memory with 16GB RAM. I don't have any experience running the import with MMAP though - any idea how much memory the JVM should get vs. leaving free for disk cache for a fair comparison?

> I'm still unsure about this PR, because there are two use cases and this PR introduces a 3rd that I do not understand (yet):
> 
> * either you need it fast and you use RAM_STORE. This requires plenty of RAM because you usually also process the CH profiles in parallel, because you want it fast! In this case free early does not make much sense and releasing the wayGeometry releases 10GB, yes, but as you will have >>120GB RAM this is not a big win for the added complexity (?)
> * or you want to just "cheaply" process it with minimal RAM. For that you have MMAP, which does all this memory management under the hood without exposing much of our internals

Yes, good point. My use case is just trying to get the world to import successfully with a small number of CH profiles (3 currently) that run sequentially. I do this in memory on a machine with 64GB of RAM. If MMAP ends up working fine for this that would be great, but if it's is much slower or has some caveats then I'd fall back to the minimal RAM_STORE option. One caveat is that the `graph.do_sort` option appears to cause all operations after the sort to use RAM_STORE.

karussell:
> any idea how much memory the JVM should get vs. leaving free for disk cache for a fair comparison?

If the graph folder is of size S and you use -XmxRg then you can try with -XmxMg where M=R-S (all in GB)

> One caveat is that the graph.do_sort option appears to cause all operations after the sort to use RAM_STORE

This shouldn't be. Instead it should throw an exception that this is currently not possible: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/util/GHUtility.java#L397

We could make it possible via creating a graph in e.g. /tmp/random-string-gh and when finished remove the current location and move /tmp/random-string-gh to the current location. But this is indeed a bit tricky.

msbarry:
OK I ran a few tests importing [North America (8.7GB)](https://download.geofabrik.de/north-america.html) from the unsync_mmap branch (openjdk 13, ubuntu 19.04 digitalocean droplets, all with 2 dedicated CPUs, foot CH profile, output directory is 6GB)

| Machine size | jvm `-Xmx` | | |
| --- | --- | --- | --- |
| 4GB | 2GB | MMAP | Ran out of heap space reading OSM file after 20m |
| 8GB | 4GB | MMAP | Ran out of heap space building location index after 1h45m |
| 8GB | 6GB | MMAP | Finish successfully in 18.5h (2h import/16.5h CH) |
| 16GB | 8GB | MMAP | Finished successfully in 4.5h (90m import/3h CH) |
| 16GB | 14GB | RAM | Finished successfully in 3h (40m import/2h20m CH) |

I think the goals of the 2 modes of operation you discuss are 1) minimize latency or 2) minimize cost where cost is roughly `(machine size in MB ram) * (hours to complete) * (1 + risk of needing to retry)`. Do think these results scale linearly as input pbf grows? If so, running in MMAP mode won't help cost much since to bump down one machine size, it takes 4x the time so cost increases 2x. It seems to offer comparable benefits to the `importAndClose` approach for reducing risk of needing to retry, but with a 50% cost penalty.

I was interested to try the MMAP import option, but it's still seeming like RAM mode with minimum machine RAM to squeak by may still be your best bet even when trying to minimize cost, which dumping 10+GB of RAM early on (and CH profiles early) makes a lot more viable.

karussell:
Thanks a lot for your time and the tests to compare the scenarios.

> digitalocean droplets

Are you sure they use powerful SSDs? Can you do a [quick read+write perf test](https://www.shellhacks.com/disk-speed-test-read-write-hdd-ssd-perfomance-linux/) there?

> minimize cost where cost is roughly

This is only true for "cloud environment". If you rent or buy a dedicated server it does not change cost if it takes 5 hours or 5 days :) but your argument still holds for other reasons like freshness of OSM data obviously.

msbarry:
> Are you sure they use powerful SSDs? Can you do a [quick read+write perf test](https://www.shellhacks.com/disk-speed-test-read-write-hdd-ssd-perfomance-linux/) there?

Sure thing:
```
$ sync; dd if=/dev/zero of=tempfile bs=1M count=1024; sync
1024+0 records in
1024+0 records out
1073741824 bytes (1.1 GB, 1.0 GiB) copied, 1.31433 s, 817 MB/s
$ dd if=tempfile of=/dev/null bs=1M count=1024
1024+0 records in
1024+0 records out
1073741824 bytes (1.1 GB, 1.0 GiB) copied, 0.143731 s, 7.5 GB/s
$ hdparm -tT /dev/vda1
/dev/vda1:
 Timing cached reads:   15452 MB in  1.99 seconds = 7751.94 MB/sec
 Timing buffered disk reads: 2324 MB in  3.00 seconds = 774.54 MB/sec
```

Also just out of curiosity I ran the 4G machine with a small input file ([massachusetts - 237mb](https://download.geofabrik.de/north-america/us/massachusetts.html)). RAM_STORE took 2m11s and MMAP took 2m24 seconds, which is closer to the difference you were observing.

karussell:
Ok, thanks.

So you definitely convinced me that we can further improve the memory required for the "fast" import. Please have a look into my changes:

 * instead of introducing another term like "free" or "flushAndFree" we should release resources via two method calls "flush" and "close".
 * LocationIndex cannot be released as it can be required for landmarks 

update: will fix a bug I introduced

karussell:
The main problem is that this only works for CH preparation (LM optionally requires the LocationIndex and this requires ) and so I wrote this in the docs and swapped order of LM+CH preparation.

msbarry:
Thanks for the changes! Sorry I missed LM side of this - added some integration tests to cover more CH/LM variations that would have caught that failure and also added `closeEarly` to LM preparation.

karussell:
> added some integration tests to cover

Thanks! Will merge this soonish.

PR #1760, opened 2019-10-22
PT Client

michaz:
Nice, simple web client for PT by Janek. Sits in `/maps/pt/index.html` and obviously doesn't do anything when not in PT mode.


michaz:
TODO: Remove data classes. This is literally just a View for a JSON response. There is no model besides the JSON response.

karussell:
This is nice :) :+1: 

Can you outline the advantages of webjars over the normal JS pipeline? Is it basically building the final JS bundle with maven? If yes, how does this work? Could it be also a solution for the "graphhopper maps" problem that the main.js file needs to be created everytime you switch a branch?

michaz:
> Is it basically building the final JS bundle with maven?

Nope, it's *including* the final JS bundle of libraries, like Moment or Leaflet or React, into *our* project, with Maven, so that *I* don't need to use a JS bundler, but can use plain, uncompiled JS.

easbar:
> we can switch back to plain, uncompiled JS

here you mean we do not care about ES5 support ?

michaz:
Yes and yes.

michaz:
Or, if we must support ES5, we type ES5.

easbar:
> Or, if we must support ES5, we type ES5.

But then we still need to do the bundling (putting everything into main.js), right ? Or why do we even do this ?

michaz:
Not familiar with this, but I think including other JS files via Require.js (not via ES6 includes) works inside the browser, without bundling.

I think bundling is just what everybody does, period, because a) at some point there suddenly was node.js, so runnning JS on an out-of-browser toolchain became the default, and b) you could write ES6 but run ES5 and c) it would manage dependencies for you.

But if you run on Java, you can manage dependencies with Maven (thanks to webjars, even for JS). And having a second toolchain is more of a drawback than a plus. Which leaves b).

michaz:
(Our JS apps are _small_. I'm not claiming what I'm proposing works for bigger JS apps. Only for "some JS included with a Java server", but I think that's exactly where we are.)

michaz:
(So far only _I_ want to try it. I don't want to persuade anyone else, I have too little clue for that.)

Janekdererste:
Serving plain javascript sounds pretty neat. Although I would still keep the (not so insane) toolchain for developlment.

PR #1762, opened 2019-10-22
introduce new POST /route possibility

### LINKED ISSUE #1689: POST request for /route
The problem of GET requests are: 

 1. e.g. point_hint can occur often and also e.g. in Arabic leading to 3 times the original string size in the URL due to URL encoding (you can increase the dropwizard limit `maxRequestHeaderSize` to avoid the 414 error, but still very ugly) and 
 2. GET **requests** cannot be gzipped I think

Related to #437 and #1499

karussell:
fixes #1689

With the new "curbsides" features the problems described in #1689 will be even worse and so I think we need this new option. I was surprised that it was possible with so few code lines (the actual feature is 50 lines!). Especially nice is the possibility to annotate classes like GHRequest without adding the annotation dependency to core via so called MixIn classes.

TODOs:

 - [x] maybe rename `GHRequest.put(String fieldName, Object value)` into `putHint` and add javadoc that this is different to point_hint
 - [x] should we rename `GHRequest setCurbSides(List<String> curbSides)` into `setCurbsides`?
 - [x] see: TODO NOW was this really missing before?
 - [x] also update api-doc.md for /route endpoint regarding snap prevention, details, curbsides
 - [x] see: TODO NOW for more than 50 or 100 locations we could zip the body
 - [ ] bring to production and use setPostRequest(true) in integration test

### DIFF
--- a/api/src/main/java/com/graphhopper/GHRequest.java
+++ b/api/src/main/java/com/graphhopper/GHRequest.java
@@ -252,10 +252,12 @@ public class GHRequest {
     }
 
     /**
-     * This method sets the hints and is equivalent to getHints().put(String, String) and used for deserialization
-     * with Jackson.
+     * This method sets a key value pair in the hints and is equivalent to getHints().put(String, String) but unrelated
+     * to the setPointHints method. It is mainly used for deserialization with Jackson.
+     *
+     * @see #setPointHints(List)
      */
-    public void put(String fieldName, Object value) {
+    public void putHint(String fieldName, Object value) {
         this.hints.put(fieldName, value);
     }
 
--- a/api/src/main/java/com/graphhopper/GHRequest.java
+++ b/api/src/main/java/com/graphhopper/GHRequest.java
@@ -33,11 +33,11 @@ import java.util.Locale;
  * @author ratrun
  */
 public class GHRequest {
-    private final List<GHPoint> points;
+    private List<GHPoint> points;
     private final HintsMap hints = new HintsMap();
     // List of favored start (1st element) and arrival heading (all other).
     // Headings are north based azimuth (clockwise) in (0, 360) or NaN for equal preference
-    private final List<Double> favoredHeadings;
+    private List<Double> favoredHeadings;
     private List<String> pointHints = new ArrayList<>();
     private List<String> curbSides = new ArrayList<>();
     private List<String> snapPreventions = new ArrayList<>();
@@ -101,7 +101,6 @@ public class GHRequest {
 
     /**
      * Set routing request
-     * <p>
      *
      * @param points          List of stopover points in order: start, 1st stop, 2nd stop, ..., end
      * @param favoredHeadings List of favored headings for starting (start point) and arrival (via
@@ -121,7 +120,6 @@ public class GHRequest {
 
     /**
      * Set routing request
-     * <p>
      *
      * @param points List of stopover points in order: start, 1st stop, 2nd stop, ..., end
      */
@@ -131,7 +129,6 @@ public class GHRequest {
 
     /**
      * Add stopover point to routing request.
-     * <p>
      *
      * @param point          geographical position (see GHPoint)
      * @param favoredHeading north based azimuth (clockwise) in (0, 360) or NaN for equal preference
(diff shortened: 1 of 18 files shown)


michaz:
That's great! Do you think we could let `GraphHopperWeb` decide by itself whether it should use the GET or the POST request? *If* we are going to have a hand-coded Java client, this is exactly what it should abstract away and what the user should not need to know about.

easbar:
> decide by itself whether it should use the GET or the POST request?

In which case would should it use GET, can it not just always use POST ? that said I think having the setter is useful to test the GET/POST endpoints separately (or even run the existing tests `@Parameterized` once using GET and once using POST).

easbar:
> should we rename GHRequest setCurbSides(List<String> curbSides) into setCurbsides

Yeah I thought about this also, but then used camel case everywhere. But you are right it should be curbside. Will do.

easbar:
> into putHint

`putExtra` ?

> use setPostRequest(true) in integration test
Unless it takes way too long I would do both true/false

michaz:
> In which case would should it use GET, can it not just always use POST ?

For a web standard purist, this is (semantically) clearly a GET. Using POST is "just" a workaround for the limited URL length problem. GETs can be cached, POSTs can't.

easbar:
> For a web standard purist, this is (semantically) clearly a GET. Using POST is "just" a workaround for the limited URL length problem. GETs can be cached, POSTs can't.

Ok this is true. But from a practical viewpoint I see no advantage of a GET. TBH I think something like GET with a request body is kind of missing from the http standard (exactly for usecases like the one heer). I don't think there is anything we can/want to cache ? Also I think its more simple to always use the same verb, imagine some bug that only occurs on some requests, because GraphhopperWeb internally decides to use POST instead of GET (or the other way around).

michaz:
>  I don't think there is anything we can/want to cache ?

In theory, that's the decision of the _client_, or their proxy servers, and we should choose the verb that best supports that decision, which is GET. :-}

> Also it seems more simple to always use the same verb, imagine some bug that only occurs on some requests, because GraphhopperWeb internally decides to use POST instead of GET (or the other way around).

 I agree, but we must support GET for eternity anyway (`/route` is our most commonly used endpoint).

easbar:
Then again I also like how its done now (GET by default) because it means a softer migration (keep using GET and everybody who needs it can opt for POST), and keep this until there seem to be no problems with POST... (then switch to POST by default later)

michaz:
If we want to be holy REST saints, we could, with the response to the POST request, include a URL that encodes the request in a compressed format, from which you can then GET it again if you like.

(Only half-serious. But possible.)

otbutz:
> > In which case would should it use GET, can it not just always use POST ?
> 
> For a web standard purist, this is (semantically) clearly a GET. 

https://stackoverflow.com/a/31984477

michaz:
Yeah, thanks for the summary. It also highlights the controversy:

> closed as primarily opinion-based

and 

> A lot of people have accepted the practice that a GET with too long or too complex a query string (e.g. query strings don't handle nested data easily) can be sent as a POST instead

and I don't think I disagree with any of it.

Except that IMO, taken literally, "a POST can do whatever it wants; anything can be a POST" proves too much, because -- sure, I can define every request to be the POSTing of something to an endpoint that "processes" something, and I can define every response entity to be temporary. So we don't need GET anymore. Which is true, but not what we're trying to get at.

otbutz:
> It remains the case that _the only reason_ we're doing this is because long URLs are either ugly or impossible. The rest is rationalizing. No entity gets more temporary because of that. The question is only how bad that is. (And the answer appears to be: Not too bad.)

That's my opinion, too. I just wanted to point out that the use of POST is not per se wrong and can be used without offending the gods of REST.

karussell:
> tbh I think something like GET with a request body is kind of missing from the http standard (exactly for usecases like the one here)

Yes, this would be the nice. Especially this URL encode thing and the URL length limitation is strange.

> Then again I also like how its done now (GET by default) because it means a softer migration (keep using GET and everybody who needs it can opt for POST), and keep this until there seem to be no problems with POST... (then switch to POST by default later)

Yeah, that was the reason I introduced the setter. Maybe I already deprecate it ;) ?

easbar:
> Maybe I already deprecate it ;)

I think we should keep it for tests

PR #1767, opened 2019-10-24
Skip non-virtual node silently in unfavorVirtualEdgePair, fixes #1258

### LINKED ISSUE #1258: Routing fails with hint pass_through and via points near dead-end ways or junctions
Graphhopper Version: 0.9 stable branch
JVM: Oracle x64 1.8
OS: Windows7 x64
Reproducible: Always

How to reproduce:
- set hint pass_through
- use 3 points for routing
- place the via point behind a dead-end way so it is snapped to the end point tower node of the way (see picture 1)

=> routing fails

Additional findings:
- when **pass_through** is de-activated the exact same point works.
- when the snapped point is **on** the dead end way, but not the end point, the routing works (see picture 2)

[Picture 1](https://discuss.graphhopper.com/uploads/default/original/1X/b5918e80a0f9e2beeb18c13571e4abccb06f35b2.png): via point is snapped to end point tower node, routing fails

[Picture 2](https://discuss.graphhopper.com/uploads/default/original/1X/7e34fd995faaff9d2ff1ebf745ed86289176ac88.png): via point is snapped to a point on the way, routing works

[Associated forum discussion](https://discuss.graphhopper.com/t/questions-on-migration-from-gh-0-1-to-0-9/2695/25)

hbruch:
As proposed by @NopMap in #1258, I silently skip unfavorites non-virtual nodes.
In analogy to enforceHeading unfavorVirtualEdgePair() will return false in such cases.

I'm not sure if changing the signature of this public method is worth potential incompatibilites and if that would require further documentation. If you deem it potentially harmful, I'll revert this.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/querygraph/EdgeChangeBuilder.java
+++ b/core/src/main/java/com/graphhopper/routing/querygraph/EdgeChangeBuilder.java
@@ -1,0 +1,142 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


easbar:
Thanks for this PR. I am not sure if we want to do this ? Right now we get a (very hard to understand) error in case a via-point snaps onto a real node, which is clearly unpleasant for the user. However, if we just ignore this error there might be (unexpected, because the user set `pass_through=true`) u-turns at junctions and it looks like a bug. So ultimately this should be fixed such that `pass_through=true` is also respected at real nodes, but not sure whats better for now ?

karussell:
Thanks for your contribution @hbruch ! 

> So ultimately this should be fixed such that pass_through=true is also respected at real nodes

Yes, I also think that this is the way to go. Silently ignoring this in this case is a bit ugly but could be still a good "enough workaround" to at least return a route in these cases until we find something better?

easbar:
We already do this for some time now, see: https://github.com/graphhopper/graphhopper/issues/1258#issuecomment-1100932471

Issue #1772, opened 2019-10-29
Remove the `GraphExtension` interface


easbar:
I think this abstraction is not very helpful. So far there is only `TurnCostExtension` implementing it and the functionality is rather limited. In many (currently 6 in production and 26 in tests) cases we need to cast the extension to `TurnCostExtension`. 

I can imagine all kinds of meta/additional data one likes to store for nodes/edges, but why would we use a `GraphExtension` for this ? Maybe we prefer a simple `int[]`, use an independent `DataAccess`, fetch data from disk etc., this can easily be done by using the node/edge ids, I see no real advantage of storing the `additionalNode/Edge` field right in the graph and for example for `TurnCostExtension` we already use this field only to store a reference to another `DataAccess`. 

Maybe I am missing something, but as far as I am concerned we can just keep the `TurnCostExtension` (without the interface) in `BaseGraph` (probably make it optional/can be null instead of using `NoOpExtension`).

karussell:
Yes, we should definitely do this! Maybe before this I have to make this branch production ready that allows storing arbitrary key value pairs like OSM tags or translated road names etc https://github.com/graphhopper/graphhopper/commits/string_index

Issue #1625, opened 2019-05-29
Reduce usages of FlagEncoder


easbar:
For example `DefaultEdgeFilter` uses `FlagEncoder` even though all it needs is an encoded value for the access flag. In turn many classes that use `DefaultEdgeFilter` (routing algos especially) only have a `FlagEncoder` field because they need to pass it to `DefaultEdgeFilter` and so on. See discussion in #1624.

Started replacing some occurrences of `FlagEncoder` with `BooleanEncodedValue` here: [`default_edge_filter_no_flagencoder`](https://github.com/graphhopper/graphhopper/tree/default_edge_filter_no_flagencoder).

PR #1775, opened 2019-11-01
move relation and turn relation parsing out of FlagEncoder

karussell:
This pull request removes several methods of FlagEncoders and TurnCostEncoder.java and is one of the many tasks to remove this class like #1625. Especially in FlagEncoders that do not use turn costs like FootFlagEncoder you see the effects of this change. Also OSMReader.TurnCostTableEntry is no longer part of the public API. 

Initially I wanted to create two PRs, one for the 'pure' relation and one for the turn restriction part, but I think this belongs together. 

So this PR is based on the branch: https://github.com/graphhopper/graphhopper/commits/move_rel_out_of_flag_encoder where the first two commits move the relation handling out of FlagEncoder into a new RelationTagParser (plus some clean up commits). It also fixes a bug here that calling the super method here has no side effects: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/MountainBikeFlagEncoder.java#L150 Additionally there was code reduction possible in some FlagEncoders. 

From commit 22b1483 it cleans up the turn relation parsing and moves it into OSMTurnRelationParser. The OSMReader shows its age here and needs some later rewrite maybe using mapdb or similar so that several workarounds can be avoided (this is tricky as it needs to be very fast). At the moment there are 2 phases: preprocessing and processing that both have 3 subprocedures: 

 * 1.1. preprocess nodes, 1.2. preprocess way, 1.3. preprocess relations
 * 2.1. handle nodes, 2.2. handle way, 2.3. handle relations

The RelationTagParser fetches the ReaderRelation tags in step 1.3 and compacts them into relationFlags (IntsRef) to convert them in step 2.2 into edgeFlags: see the new EncodedValues for RouteNetwork (`bike.network` and `foot.network`). These EncodedValues can be used in other TagParsers.

The TurnCostParser is more involved and different to the RelationTagParser as it needs a traversable Graph and e.g. for `car.turn_cost` it requires a ready-to-use `car.access`. So this work is done at step 2.3. Also the method TurnCostParser.handleTurnRelationTags requires several parameters like the graph and an interface to convert between OSM and internal GraphHopper IDs.

TODOs:

 - [x] cache EdgeExplorer creation in OSMTurnRelationParser like it was before?
 - [x] Should we use one parser instead of the two OSMFootNetworkTagParser and OSMBikeNetworkTagParser that have only tiny differences? -> keep it separate
 - [x] Make turn cost handling easier via https://github.com/graphhopper/graphhopper/issues/1772#issuecomment-547626445 and remove TurnCostAccess
 - [x] rename TurnCostExtension to TurnCostStorage?
 - [x] remove TODO NOWs


easbar:
Copied from now closed #1772:

> Update: this would be also helpful for the branch move_tc_out_of_flag_encoder (#1775) and we could try if the following is possible: 

>   instead of TurnCostAccess we make usage simple via:
    edgeIteratorState.setTurnCost(carTurnCostEnc, nextEdgeId, 10.0) // base->adj->nextEdgeId
    we move tcFlags inside EdgeIteratorState (remove it e.g. in TurnWeighting and TurnCostAccess)
    we should be able to remove entire TurnCostAccess

core/src/test/java/com/graphhopper/routing/ch/CHTurnCostTest.java
```diff
@@ -181,14 +181,14 @@ public void testFindPath_multipleInOutEdges_turnReplacementDifference() {
...
         setRandomCost(6, 7, 10, rnd);
-        setRandomCost(0, 5, 3, rnd);
-        setRandomCost(1, 5, 3, rnd);
-        setRandomCost(0, 5, 6, rnd);
-        setRandomCost(1, 5, 6, rnd);
-        setRandomCost(4, 7, 8, rnd);
-        setRandomCost(4, 7, 9, rnd);
-        setRandomCost(6, 7, 8, rnd);
-        setRandomCost(6, 7, 9, rnd);
+        setRandomCostOrRestriction(0, 5, 3, rnd);
```
easbar: just using setRandomCost would not lead to strict restrictions

core/src/test/java/com/graphhopper/storage/ShortcutUnpackerTest.java
```diff
@@ -231,22 +231,21 @@ public void withTurnWeighting() {
...
+        setTurnCost(edge5.getEdge(), 6, NEXT_EDGE, 6.0);
         // turn costs <-
-        tce.setExpensive(encoder.toString(), encodingManager, NEXT_EDGE, 6, edge5.getEdge(), 2.0);
-        tce.setExpensive(encoder.toString(), encodingManager, edge5.getEdge(), edge4.getEdge(), 5, 3.0);
-        tce.setExpensive(encoder.toString(), encodingManager, edge4.getEdge(), edge3.getEdge(), 4, 2.0);
-        tce.setExpensive(encoder.toString(), encodingManager, edge3.getEdge(), edge2.getEdge(), 3, 4.0);
-        tce.setExpensive(encoder.toString(), encodingManager, edge2.getEdge(), edge1.getEdge(), 2, 1.0);
-        tce.setExpensive(encoder.toString(), encodingManager, edge1.getEdge(), edge0.getEdge(), 1, 0.0);
-        tce.setExpensive(encoder.toString(), encodingManager, edge0.getEdge(), 0, PREV_EDGE, 1.0);
+        setTurnCost(NEXT_EDGE, 6, edge5.getEdge(), 2.0);
```
easbar: here the via node and to edge were intermixed

core/src/test/java/com/graphhopper/routing/DirectedBidirectionalDijkstraTest.java
```diff
@@ -250,11 +250,11 @@ public void directedRouting() {
         int leftNorth = graph.edge(9, 0, 1, true).getEdge();
 
         // make paths fully deterministic by applying some turn costs at junction node 2
-        addTurnCost(7, 2, 3, 1);
-        addTurnCost(7, 2, 6, 3);
-        addTurnCost(1, 2, 3, 5);
-        addTurnCost(1, 2, 6, 7);
-        addTurnCost(1, 2, 7, 9);
+        setTurnCost(7, 2, 3, 1);
```
easbar: I think its better to do such renamings in a separate commit in master (same with turn cost extension/storage). I changed this in master and merged here again

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
```diff
@@ -108,22 +111,14 @@ public double calcTurnWeight(int edgeFrom, int nodeVia, int edgeTo) {
         if (!EdgeIterator.Edge.isValid(edgeFrom) || !EdgeIterator.Edge.isValid(edgeTo)) {
             return 0;
         }
+        double tCost = 0;
         if (turnCostExt.isUTurn(edgeFrom, edgeTo)) {
-            // note that the u-turn costs overwrite any turn costs set in TurnCostStorage
-            // todo:
-            // also this does not allow TurnCostEncoder to set the u-turn costs to zero explicitly, like FootFlagEncoder!
-            // this problem is a bit hidden, because if you do not apply any turn restrictions (like FootFlagEncoder)
-            // you never do a u-turn anyway.
```
easbar: the comment still applies, no ? setting u-turn costs on turnCostExt has no effect, because the uTurnCosts take a higher priority here.

core/src/test/java/com/graphhopper/routing/ch/CHTurnCostTest.java
karussell: Ups, thanks for the fix!

core/src/test/java/com/graphhopper/routing/DirectedBidirectionalDijkstraTest.java
karussell: Yes, you are right. Thanks!

core/src/main/java/com/graphhopper/routing/weighting/TurnWeighting.java
karussell: Ah, ok. This is what was meant with the comment before the todo. I'll revert the removal of that.

Regarding the TODO: the FootFlagEncoder does not have any turn restriction stuff but still overwriting/overruling is not possible. 

BTW: could we make isUTurn dependent on turnCostEnc and then remove this comment?

core/src/main/java/com/graphhopper/routing/profiles/TurnCost.java
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * You need to call EncodingManager.getKey(prefix, EV_SUFFIX) as this EncodedValue can be used for e.g. car and bike
+     */
+    public final static String EV_SUFFIX = "turn_cost";
+
+    /**
+     * This creates an EncodedValue specifically for the turn costs
+     */
+    public static DecimalEncodedValue create(String name, int maxTurnCosts) {
+        int turnBits = Helper.countBitValue(maxTurnCosts);
```
easbar: Not sure if this is relevant here, but IntelliJ complains about `return counter++` in `Helper#countBitValue` (`counter is incremented "after" it is returned)

core/src/main/java/com/graphhopper/routing/util/parsers/OSMTurnRelationParser.java
```diff
@@ -0,0 +1,194 @@
...
+            while (iter.next()) {
+                int edgeId = iter.getEdge();
+                long wayId = map.getOsmIdOfInternalEdge(edgeId);
+                if (edgeId != edgeIdFrom && osmTurnRelation.getRestriction() == OSMTurnRelation.Type.ONLY && wayId != osmTurnRelation.getOsmIdTo()
+                        || osmTurnRelation.getRestriction() == OSMTurnRelation.Type.NOT && wayId == osmTurnRelation.getOsmIdTo() && wayId >= 0) {
+                    final TCEntry entry = new TCEntry(edgeIdFrom, viaNode, iter.getEdge());
+                    IntsRef tcFlags = new IntsRef(turnCostFlags.length);
+                    getTurnCostEnc().setDecimal(false, tcFlags, Double.POSITIVE_INFINITY);
+                    entries.add(entry);
+                    tcs.setTurnCost(tcFlags, edgeIdFrom, viaNode, iter.getEdge());
```
easbar: Why can we not use tcs.set(...) here and why do we (need to) create a new IntsRef everytime we are here ?

easbar:
LGTM

core/src/main/java/com/graphhopper/routing/util/parsers/OSMTurnRelationParser.java
karussell: Ah, yes. It is indeed no longer required to create a new IntsRef now that we do no collect the TurnCostEntry. And will also try to use tcs.set...

karussell:
Thanks a lot for the review - I finally could get rid of a bit more code :)

Issue #82, opened 2013-06-27
Area routing for pedestrian/bike


karussell:
Even if we already include areas/places in the routing there is still some work to do as we need to cross the polygon and the way along the lines might be suboptimal.

http://comments.gmane.org/gmane.comp.gis.openstreetmap.region.at/5213

http://openplans.org/2012/06/b-roll-david-solves-the-plaza-problem-with-help-from-de-berg-and-matt-conway/

Think about it: Instead of pre-calculation: use spline and post-processing while or after path extraction

vincentdephily:
[This][1] example was reported on the [osm-website issues][2].

[1]:https://graphhopper.com/maps/?point=51.8627%2C0.9545&point=51.8626%2C0.9559&layer=Lyrk
[2]:https://github.com/openstreetmap/openstreetmap-website/issues/900

neiljp:
What routing over areas does GH do? The bug I filed (now closed) on the osm website repo linked to one example, here is another:
http://www.openstreetmap.org/directions?engine=graphhopper_foot&route=57.16063%2C-2.11108%3B57.16071%2C-2.11137#map=19/57.16083/-2.11150
or in GH:
https://graphhopper.com/maps/?point=57.160614%2C-2.111076&point=57.160698%2C-2.111397&vehicle=foot&locale=en-US&elevation=true&layer=Lyrk

Both involve routing over a car-park area; the first from a road to a foot way, both nodes on the area boundary, the second as the assumption would be that steps would be a preferred route down into car-park area to get to shops (simplified to simply getting 'down steps', which it doesn't do).

Happy to split this up into different issues, but I'm not sure what GH does and doesn't cope with in this regard.

sebalis:
[...]

PR #1778, opened 2019-11-06
Add snap prevention option: pedestrian_area

minisu:
In order to avoid snapping to the "wrong side" of a long pedestrian area.
The typical pedestrian area (`highway=pedestrian, area=true`) that causes
problems is a long coastal esplanade: There are several connecting edges
from the land side, but none from the water side. If we snap to the water
side of the esplanade, the route will trackaround  the edges of the area
before getting to the connecting land side edges. This results in routes
that are way too long (a route that should be a few meters can become
several kilometers).

One example of such esplanade is Norra Hammarbyhamnen, Stockholm, Sweden.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/profiles/RoadEnvironment.java
+++ b/core/src/main/java/com/graphhopper/routing/profiles/RoadEnvironment.java
@@ -25,7 +25,8 @@ import com.graphhopper.util.Helper;
  */
 public enum RoadEnvironment {
     OTHER("other"), ROAD("road"), FERRY("ferry"),
-    TUNNEL("tunnel"), BRIDGE("bridge"), FORD("ford"), SHUTTLE_TRAIN("shuttle_train");
+    TUNNEL("tunnel"), BRIDGE("bridge"), FORD("ford"), SHUTTLE_TRAIN("shuttle_train"),
+    PEDESTRIAN_AREA("pedestrian_area");
 
     public static final String KEY = "road_environment";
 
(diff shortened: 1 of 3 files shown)


karussell:
I do not understand the use case as highway=pedestrian should not accessible for motor vehicles anyway and if it is accessible why would you want to avoid snapping and why only in combination with area=true? Can you post a GraphHopper Maps link of the situation?

minisu:
Sorry, I was not being very clear. I'm talking about vehicle=foot here, not car. Given a set of objects, I want to find out which object that is closest to the user of my application. Through manual testing, I have found that the routed distance can be way too long when the user is located on these kind of coastal esplanades or quays. Some examples:

 - [Dublin example](https://graphhopper.com/maps/?point=53.347298%2C-6.232248&point=53.346885%2C-6.233496&locale=en-US&vehicle=foot&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale)
 - [Gothenburg example](https://graphhopper.com/maps/?point=57.708603%2C11.959649&point=57.708542%2C11.960725&locale=en-US&vehicle=foot&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale). 
 - [Stockholm example](https://www.openstreetmap.org/directions?engine=fossgis_osrm_foot&route=59.30749%2C18.09401%3B59.30723%2C18.09500#map=16/59.3059/18.0878&layers=H) (OSRM being used since GraphHopper hasn't yet updated the routing since the virtual paths that I tried to add were removed).

This can result in a far away object being reported as closest to the user.

By not snapping to these areas (but still allowing routing on them them), this problem goes away. The downside is of course that GraphHopper won't snap to regular plazas either, but in my experience this affects routed distances marginally compared to how badly these coastal esplanades/quays can mess the result up.

karussell:
Thanks for the examples.

This is IMO not a viable solution and will fail for many situations and has nothing to do with a wrong snapping. See #82 for a proper solution.

Issue #1121, opened 2017-07-31
Create a Common Interface for PathDetails and InstructionList


boldtrn:
As discussed in [#1091](https://github.com/graphhopper/graphhopper/pull/1091/files/c7768bfe4ece793adb5e1c7b8ea56c4189671086#r129300232) we should try to create a common interface for PathDetails and InstructionList to make it easier to handle them. Maybe in the end we do not distinguish between either anymore.

The goal should be to make 'instructions' identical to any other path details.

michaz:
If I may: PathDetails and Instructions are part of the response, right? So they are basically data transfer objects.

I would caution against doing too much object oriented design around them. They should best be simple data structures, not too far from JSON object representations. Ideally, they shouldn't require much handling at all, or having interfaces made common.

I know that is "un-OO", and there once was this pattern where you would use real, rich, OO-designed "business" objects with high-level methods, but that has some issues, see Clean Code, Chapter 6, "Data/Object Anti-Symmetry", but more importantly:

People with e.g. auto-generated clients don't know about these interfaces and methods, and have to make sense of the data themselves.

michaz:
Related to [this](https://github.com/graphhopper/graphhopper/issues/1139#issuecomment-325204261) type of problem.

PR #1782, opened 2019-11-09
Speed up path simplification with path details/instructions

### LINKED ISSUE #1764: Path simplification is very slow compared to routing
For example for 50 random locations in Germany `PathSimplification.simplify` takes about 80% of the time needed for calculating the route (`GraphHopper.route(request)` for node-based CH), something like: 600 of 720ms.
Apparently, for more locations this gets even worse, e.g. for 500 locations the path simplification takes 55s out of the 57s it takes to calculate the route (>95%). This is maybe expected, because for longer routes CH does not get so much slower, while the path contains a lot more points that need to be simplified.

I added a little measurement file in the `performance_via_routing` branch.

easbar:
Fixes #1764.

For very long routes this PR strongly reduces route query time, for example for 500 via points through germany the query time is reduced from something like a minute to about 1.5s. The effect is the stronger the longer the route is. For long routes with path details/instructions the computation time was almost exclusively spent on the path simplification step (>99%).

The first commit of this PR provides the most important fix and is only a few lines long, speeding up my test query from about 60s to something like 8s already. However with this version there is still a lot of time spent on the simplification (>80%), and it is not spent with actual `Douglas-Peucker`, but rather with updating the interval indices in path details/instructions.

I have therefore cleaned up the code and mostly rewritten the algorithm in `PathSimplification` step which yields another speed up of about 4-5x resulting in above mentioned query time of about 1.5s. I am not sure if `PathSimplification` is now easier or harder to understand than before. Tbh I do not think I fully understand what the old code did, so happy to hear opinions here as well.

The tests are green, but I am not 100% sure all this works and maybe someone with path detail experience can try this ? It seems like my code deletes a few more points than in master, but all start/end points of the path detail/instruction intervals should be preserved. Also the instructions in the web demo seem to be the same.


core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -65,77 +65,98 @@ public PointList simplify() {
...
-                        for (int i = 0; i < listsToSimplify.size(); i++) {
-                            List pathDetails = listsToSimplify.get(i);
-                            reduceLength(pathDetails, offsets[i], startIntervals[i], endIntervals[i] - removed);
-                            // This is not needed for Instructions, as they don't contain references, but PointLists
-                            if (pathDetails.get(0) instanceof PathDetail) {
-                                for (int j = offsets[i] + 1; j < pathDetails.size(); j++) {
-                                    PathDetail pd = (PathDetail) pathDetails.get(j);
-                                    reduceLength(pathDetails, j, pd.getFirst() - removed, pd.getLast() - removed);
-                                }
-                            }
```
easbar: This code took most of the time after I fixed the `compressNew` bug in the first commit

easbar:
Btw I also added a measurement to `Measurement` that does CH routing queries with path details, instructions and point hints (I called it `routingCH_full`). For random routing in Germany (without via nodes) this is already x2-x3 slower than for routing without the simplification.

boldtrn:
Thanks for the great analysis and for fixing this :tada:

easbar:
> Is it possible to avoid the new public PointList.compress method? If not easily possible then let's keep it

its easy to put it somewhere else, but where do you think should it be ? since its an operation on PointList I think this is where it belongs. What I do not like so much is that `compress`, know which values it should consider as 'holes' that need to be filled (`Double.NaN`). We could also make this a method parameter, or call the method `removeNaN` or something.

> Should we keep assertConsistencyOfPathDetails always enabled for now? Or do you think the possibility that the indices are corrupted is low now? (in the early days we had lots of problems here)

I thought the same and decided to keep it because it has no performance impact.

easbar:
@boldtrn did you try this with kurviger or something with lots of instructions ? I dont think something should be broken by this PR, but would be nice to be double check

karussell:
> but where do you think should it be ?

I just do not like the fact that this is a public method (so a tiny issue :)). I do not see a different use case (yet) and it is very specific to the simplification process IMO and should be there.

boldtrn:
Thanks for this great PR. I just had a look at the results. I think something is wrong with merging two routes. The instructions index for the first continue instruction of the nth route (n > 0) seems wrong.

I tried [this route](https://graphhopper.com/maps/?point=48.891069%2C9.33162&point=48.890991%2C9.330027&point=48.893439%2C9.328798&locale=de-DE&vehicle=car&weighting=fastest&elevation=true&details=road_class&debug=true&points_encoded=false&use_miles=false), but happens for other routes as well. If you look at the red dots (instruction dots) they are skewed when using this branch. 

If you compare the results to the current master it looks like this.

Master:
```
{
distance: 0,
sign: 5,
last_heading: 265.1591380085937,
interval: [
2,
2,
],
text: "Wegpunkt 1",
time: 0,
street_name: "",
},
{
distance: 207.923,
heading: 265.16,
sign: 0,
interval: [
2,
3,
],
text: "Dem Straßenverlauf von Schillerstraße, K 1849 folgen",
time: 10396,
street_name: "Schillerstraße, K 1849",
},
```

This branch:
```
{
distance: 0,
sign: 5,
last_heading: 265.1591380085937,
interval: [
2,
2,
],
text: "Wegpunkt 1",
time: 0,
street_name: "",
},
{
distance: 207.958,
heading: 265.16,
sign: 0,
interval: [
2,
2,
],
text: "Dem Straßenverlauf von Schillerstraße, K 1849 folgen",
time: 10694,
street_name: "Schillerstraße, K 1849",
},
``` 

I think the interval should be `[2,3]?

boldtrn:
We should definitely write a test for this to catch this in the future :).

easbar:
Thanks for testing! I'll have a look

easbar:
The problem was I did not correctly handle via points (intervals with a single point). I now added an integration test, made the code easier to test and fixed the problem. I checked instructions and path details via the UI and as far as I can tell it works correctly now, but please do not hesitate to try again.

easbar:
> I cannot find the comment in the code anymore where we wanted to "merge" PathDetails and Instructions, which is now basically done via the new Partition interface (?)

Hmm, ok, I dont think I removed such a comment ? You mean this issue: https://github.com/graphhopper/graphhopper/issues/1121 ? I now added a comment in this direction, here: https://github.com/graphhopper/graphhopper/pull/1782/commits/d82587291398390f7d5d89a2d1865bf8f1ae8820. With this PR path details and instructions only use the `Partition` interface temporarily in PathSimplification (because this makes testing much easier, we do not need a PathWrapper etc.). Here, it would help if `Instruction` just stored indices referencing the `PointList` instead of storing its own (partial copy of the full) `PointList`, just like `PathDetails` do. Also I would not make `Instructions` extend `AbstractList`, but rather use `List<Instructions>`, just like `List<PathDetail>`.

easbar:
Random routing queries in Germany with instructions+path details got about twice as fast with this PR (3.5ms instead of 7ms)

karussell:
> You mean this issue: #1121 ? I now added a comment in this direction, here: d825872.

Ah, yes. Thanks!

> Random routing queries in Germany with instructions+path details got about twice as fast with this PR (3.5ms instead of 7ms)

Wow, really nice!

PR #1783, opened 2019-11-10
Remove GraphExtension interface

### LINKED ISSUE #1772: Remove the `GraphExtension` interface

easbar:
`GraphHopperStorage` and `BaseGraph` no longer take a `GraphExtension` as constructor arguments. Instead, `TurnCostExtension` is created (or not created) in `BaseGraph` directly. The `E_ADDITIONAL` flag is gone (it was never used in core). I renamed `N_ADDITIONAL` to `N_TC`, because it is now (solely) used to store the turn cost table index for a node. `Graph#getExtension` is now called `Graph#getTurnCostExtension` and returns `null` if the storage was configured without turn costs.

I also removed `TurnCostExtension#init()` and moved everything to the constructor. This means the error we got in #1398 can no longer happen. However, we still have not implemented graph sorting in the presence of turn costs and I am now throwing a more explicit error in this case.

Overall I just moved around code here, any more sophisticated refactorings related to turn cost storage probably will go with #1775 or later PRs.

Fixes #1772


core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -368,7 +374,9 @@ void create(long initSize) {
         initSize = Math.min(initSize, 2000);
         wayGeometry.create(initSize);
         nameIndex.create(initSize);
-        extStorage.create(initSize);
+        if (supportsTurnCosts()) {
```
easbar: These checks are a bit ugly but I don't think its really worth it to have something like `NoOpTurnCostExtension extends TurnCostExtension` that overrides all the methods with no-op implementations, or have an extra interface for this purpose.

karussell: > NoOpTurnCostExtension extends TurnCostExtension that overrides all the methods with no-op implementations

Why not :D ?

easbar: Option 1: create a subclass of turn cost extension and overwrite all methods (with no-ops): I dont like this, because in future whenever turn cost extension changes we need to update this subclass as well

Option 2: create some interface: I dont like this because we need to put all methods we are using in turn cost extension into this interface, also not sure what type to return in `coypTo` for example.

The way its done now we have to check every access to turnCostExtension for null (`supportsTurnCosts`). I don't know, I did both and liked this better :D but of course we can do it another way if you have a preference

karussell: Yes, I don't like option 1 too much either, but I still have a low preference for it due to the better readability **inside** BaseGraph (the ifs attract too much attention IMO).

> turnCostExtension for null

With the NoOp we could introduce an isActive() method and make the **public** API slightly more readable?

`storage.getTurnCostExtension().isActive()`

But I have no strong opinion, so please feel free to merge it like it is :)

easbar: Yes now you got me thinking again :) I'll try option 2 again it would be nice to have `Graph#getTurnCostExtension` return an interface rather than the implementation (`TurnCostExtension`)

easbar: Now I extracted an interface for `TurnCostExtension` (see last two commits). I also tried with the no-op implementation. Still dont really like it :)

karussell: Not so sure with the TurnCostExtension interface as we only have a single (real) implementation. But yes, NoOp extending "TurnCostExtensionImpl" also smells. Uhm.

easbar: I thought its a bit of an improvement because this way `Graph#getTurnCostExtension` does not return something that has implementation specific stuff like `setSegmentSize()`. `NoOp` does not extend `TurnCostExtensionImpl`, but just implements the interface. Oh well shall I just go back to my original version for now ? I think with #1775 etc. we will do more changes anyway.

karussell: I like https://github.com/graphhopper/graphhopper/pull/1783/commits/0d487cd6dba0c27a8c29f556b4bd2d9cd529ff03 but please do not hesitate to return to your original version as you do not seem convinced :) and as you said we can later improve this.

easbar: Ok I'll go back to the original version and we'll keep this in mind in the upcoming refactorings

Issue #1709, opened 2019-08-30
Use nv-i18n to replace graphhoppers Country enum


otbutz:
The nv-i18n lib has worldwide coverage and covers more than just DE/AT:

https://github.com/TakahikoKawasaki/nv-i18n/blob/master/src/main/java/com/neovisionaries/i18n/CountryCode.java

I know that the number of constants are problematic if we would use the default EnumEncodedValue to store it. My suggestion would be to create a FilteredEnumEncodedValue class which would allow us to only store a **subset** of relevant values.

My ultimate goal would be to define relevant countries in the config file and only encode the selected ones in the graph.

What do you think?

karussell:
Why would this required? What benefits over a simple list of 2 letter isocodes do you see?

otbutz:
The overall plan would be to gather toll mileage on a per country basis for a given route.
eg: Vehicle x drives from Germany to Croatia and i want to know how many kilometers include general/hgv toll.

For this to work i'd have to store the country of every edge with the ability to do this on a global scale(195 countries). And i think that it's much easier to have a curated list which is actively developed and used across many projects.

karussell:
Why not e.g. Locale.getISOCountries() or just the countries where we have rules for like we do it for TranslationMap?

otbutz:
I was looking at the SpatialRuleParser which already does 99% of the work AFAIK. I could use the countrycodes and shapes from the spatial rules file and encode the country association as an `UnsignedIntEncodedValue` but maybe it would be better to establish the concept of "country rules" as such? eg have a typesafe enum with all countries, assign them to edges within a countries boundaries and be able write rules without resorting to String keys or index based lookup.

btw why is the SpatialRuleParser using an EnumEncodedValue only to store the index number with `setInt()`?

PR #1784, opened 2019-11-11
String index

karussell:
This PR introduces a flexible storage to store key-value String pairs for every edge and will replace the NameIndex, so that we can store translated name tags ala name_ru, ref-tags separately or even arbitrary tags per edge. (It is basically a Map ala `Map<Long, Map<String, String>>` except that you cannot define Long when you put it and get it returned when you add the `Map<String, String>`.)

This PR only introduces the index, we still need a method like `edge.putExtra("key", "value")` and we could also introduce a generic StringPathDetails and replace StreetNameDetails. Related also to #1772.

The index has the following limitations:

 * append-only storage. I.e. updating existing key-value pairs for one edge is only possible via allocating more space
 * supports up to `32768-1` unique keys (max length of each is 65k) and holds them in-memory
 * arbitrary values with a length of 255 bytes (UTF-8). Is mostly sufficient, also for names.
 * it does a simple duplicate detection to reduce storage size, but does this only for identical keys or identical values or fully equal entry maps (no partial strings)

TODOs:

 - [x] the simpleCache brings 10% smaller size on disk but makes everything a bit more complicated
 - [x] store ints even if it needs to be split over 2 segments => use get+setBytes (or use padding?)
 - [x] is storing all tags already possible? Yes! How large does the index get => 221MB for Germany i.e. increases the total size by only 15% (5310 unique keys - make sure to exclude our artificial tags like estimated_center/distance and duration. Maybe also all tags that are already handled in our parsers)
 - [x] compare speed for bigger and smaller maps - looks good
 - [x] improve storage size further for larger maps like Germany. Trying to use [lz4-java](https://github.com/graphhopper/graphhopper/compare/getall_prototype...lz4_string_index) failed and even increased the size for the compressed data. But a very interesting tool is [zstd that can be trained](https://github.com/facebook/zstd/#the-case-for-small-data-compression) to specific data which would be likely beneficial for OSM data. A similar tool is http://ed-von-schleck.github.io/shoco/ but for now we'll keep every as it is.


otbutz:
Looks good :+1: 

We could use this to store countrycodes per edge like i proposed it in #1709

karussell:
Thanks! Works indeed surprisingly well.

> We could use this to store countrycodes per edge like i proposed it in #1709

In general this is more for the path extraction phase or for the MVT endpoint. It is not really suited for routing due to its design towards more compact size e.g. loops through all values of one edge. I'm sure for smaller areas it is still possible to do that, but this shouldn't be promoted in the core and should be done via EncodedValues instead.

otbutz:
>In general this is more for the path extraction phase or for the MVT endpoint. 

That's exactly my use case as I would like to add country information to the path details. The StringIndex would allow me to do this fairly cheap and i could avoid adding ~250 countries to the `Country` enum  to get global coverage. Storing all this information in the graph would require roughly one byte per edge which seems like a waste of space considering that we currently ship just a few SpatialRule implementations.

karussell:
> That's exactly my use case as I would like to add country information to the path details. 

Have a look into this branch: https://github.com/graphhopper/graphhopper/tree/getall_prototype

More a prototype (virtualedges won't work), but the rest seems to work fine and also for bigger areas:

image

You can even misuse this to store the OSM ID ... highly inefficient but oh well...

michaz:
Thank you thank you thank you.

PR #1641, opened 2019-06-22
Avoid main.js in git repo

karussell:
Fixes #1548 

> from @michaz 
> Build and deploy it with its appropriate toolchain (npm probably?)

I think this is the easiest way to solve the problem: removing main.js from git and creating it on travis. I.e. in every released jar the main.js will be still included.

TODOs

 - [x] adopt docs for developers (check user docs too, include mention of assets overrides section - see below)
 - [x] adopt docker
 - [x] include nvm&npm install in graphhopper.sh?
 - [x] ~make sure server restart is not required for JS changes (there seems to be something broken, maybe due to the recent dropwizard upgrade?)~ I used the config-example.yml where the assets overrides section is uncommented. This is a bit ugly but as we do not want this in production we need to have it as it is.

Issue #1700, opened 2019-08-27
execute npm as part of the maven cycle


karussell:
See #1641 why we removed main.js from the repo.

Now sometimes one can forget to create the main.js and sometimes one forgets to call `npm install`. Should this be done from the maven cycle? https://github.com/eirslett/frontend-maven-plugin

michaz:
I think most Maven users rightly expect that a POM completely describes the build process and that there's nothing to forget. So I'd clearly say yes.

Issue #1584, opened 2019-03-28
Include main.js in repo?


karussell:
See https://github.com/graphhopper/graphhopper/pull/1572#issuecomment-477497841

easbar:
This has probably been discussed a few times so far, but `main.js` seems to blow up the git repository size (because all its content is on the same line). This is the output I got running this nice [script](https://stackoverflow.com/a/42544963) showing git object sizes. I think the hashs are object ids not git commit hashs. Maybe it would be acceptable to remove main.js from the repository and include a manual build step before running the graphhopper server (could be included in `graphhopper.sh` as well). To be able to run without system-wide node/npm installation there is [nvm](https://github.com/creationix/nvm). Or maybe another easier way would be to not minify main.js (so not all content is on the same line and git can produce smaller diffs) for the repository and only use a minified version of main.js for production ?

```
0ffd88166777  1.3MiB core/files/N49E011.hgt.zip
53404479bc31  1.4MiB core/files/north-bayreuth.osm.gz
623adc98a9e2  1.4MiB core/files/N50E011.hgt.zip
8a76769f58ab  1.4MiB web/src/main/webapp/js/main.js
4dfa59baa221  1.4MiB reader-gtfs/files/beatty.osm
892c6aa9860e  2.0MiB reader-gtfs/files/rnv.zip
a2e129b7087a  2.3MiB web/src/main/resources/com/graphhopper/http/countries.geo.json
2c3edcd23bd2  2.3MiB core/files/spatialrules/countries.geo.json
ccfc0f941342  2.4MiB docs/slac-r-186-fixed-radius-near-neighbors.pdf
e6e61c15b914  2.8MiB web/src/main/webapp/js/main.js
2474cec2eb70  2.8MiB web/src/main/webapp/js/main.js
815c82541fc0  2.9MiB core/files/N48W125.hgt.zip
747c116e5f84  3.0MiB web/src/main/webapp/js/main.js
9b77c19142f3  3.0MiB web/src/main/webapp/js/main.js
7bea392542b3  3.0MiB web/src/main/webapp/js/main.js
491be81e73b2  3.1MiB reader-shp/src/test/resources/data/gis.osm_roads_free_1.shp
6b84456772ce  3.1MiB reader-shp/src/test/resources/data/malta-latest.osm.pbf
116ff9c6482f  3.5MiB reader-shp/src/test/resources/data/gis.osm_roads_free_1.dbf
[...]

PR #1785, opened 2019-11-11
Mention how main.js is build without fetching it from pre-build JAR files

Nakaner:
#1641 removed the artifact called `main.js` from the Git repository which is good. However, I think that just mentioning that it can be extracted from pre-build JAR files is not optimal. If a project is open source, it should explain to how it is build and not say "download build artefact from $location".

After an upgrade of my development environment to 0.13 and building it, I was confused why the web frontend did not work any more. It took me a moment to figure out how to call npm from the Docker and Travis configuration (I basically run `git log -p -- web/src/main/resources/assets/js/main.js` to find the pull request). I don't see why building the project fully from the sources should be hidden.

### DIFF
--- a/docs/core/quickstart-from-source.md
+++ b/docs/core/quickstart-from-source.md
@@ -15,8 +15,12 @@ Then do:
 ```bash
 git clone git://github.com/graphhopper/graphhopper.git
 cd graphhopper; git checkout 0.13
-# fetches main.js, can be omitted if no UI is needed
+# Build main.js. This can be omitted if no UI is needed.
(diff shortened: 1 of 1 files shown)


karussell:
Yes, the process is currently far from optimal and we need to improve this but it is done in this way to avoid the hassle to install npm. See also discussion in #1584.

> If a project is open source, it should explain to how it is build and not say "download build artefact from $location

Not sure what you mean, but we explain this in detail here: https://github.com/graphhopper/graphhopper/blob/master/docs/core/quickstart-from-source.md#javascript

Closing in favor of #1700.

PR #1786, opened 2019-11-11
Add support for EU tunnel codes

otbutz:
This is a followup PR for the discussion in #1711 which adds the ability to honor restrictions on the transport of hazardous goods through tunnels.

Changes so far:

- added support for alternative tagging schemas `hazmat:adr_tunnel_cat` and `hazmat:tunnel_cat`
- check for the presence of `tunnel=yes` when dealing with hazmat subtags like `hazmat:B` to avoid collisions with tags in [Finland](https://wiki.openstreetmap.org/wiki/Key:hazmat#Finland).

You can run the following overpass turbo query to visualize the ways with supported tag combinations:
```
[out:json][timeout:25];
// gather results
(
  way["hazmat:tunnel_cat"];
  way["hazmat:adr_tunnel_cat"];
  way["hazmat:A"="no"]["tunnel"="yes"];
  way["hazmat:B"="no"]["tunnel"="yes"];
  way["hazmat:C"="no"]["tunnel"="yes"];
  way["hazmat:D"="no"]["tunnel"="yes"];
  way["hazmat:E"="no"]["tunnel"="yes"];
);
// print results
out body;
>;
out skel qt;
```

overpass turbo
grafik

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMHazmatTunnelParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMHazmatTunnelParser.java
@@ -1,0 +1,58 @@
+package com.graphhopper.routing.util.parsers;
+
+import java.util.List;
+
+import com.graphhopper.reader.ReaderWay;
+import com.graphhopper.routing.profiles.EncodedValue;
+import com.graphhopper.routing.profiles.EncodedValueLookup;
+import com.graphhopper.routing.profiles.EnumEncodedValue;
+import com.graphhopper.routing.profiles.HazmatTunnel;
+import com.graphhopper.routing.util.EncodingManager.Access;
+import com.graphhopper.storage.IntsRef;
+
+public class OSMHazmatTunnelParser implements TagParser {
+    
+    private static final String[] TUNNEL_CATEGORY_NAMES;
+    static {
+        HazmatTunnel[] categories = HazmatTunnel.values();
+        TUNNEL_CATEGORY_NAMES = new String[categories.length];
+        for (int i = 0; i < categories.length; i++) {
+            TUNNEL_CATEGORY_NAMES[i] = categories[i].name();
+        }
+    }
+    
+    private final EnumEncodedValue<HazmatTunnel> hazTunnelEnc;
+    
+    public OSMHazmatTunnelParser() {
+        this.hazTunnelEnc = new EnumEncodedValue<>(HazmatTunnel.KEY, HazmatTunnel.class);
+    }
+
+    @Override
+    public void createEncodedValues(EncodedValueLookup lookup,
+                    List<EncodedValue> registerNewEncodedValue) {
+        registerNewEncodedValue.add(hazTunnelEnc);
+    }
+
+    @Override
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, Access access,
+                    long relationFlags) {
+        
+        if (readerWay.hasTag("hazmat:adr_tunnel_cat", TUNNEL_CATEGORY_NAMES)) {
+            HazmatTunnel code = HazmatTunnel.valueOf(readerWay.getTag("hazmat:adr_tunnel_cat"));
+            hazTunnelEnc.setEnum(false, edgeFlags, code);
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/OSMHazmatTunnelParser.java
```diff
@@ -0,0 +1,67 @@
...
+            TUNNEL_CATEGORY_NAMES.put(code.name(), code);
+        }
+    }
+    private static final SortedMap<String,HazmatTunnel> HAZMAT_SUBTAGS;
+    static {
+        HAZMAT_SUBTAGS = new TreeMap<>(Collections.reverseOrder());
+        for (HazmatTunnel code : HazmatTunnel.values()) {
+            HAZMAT_SUBTAGS.put("hazmat:" + code.name(), code);
+        }
+    }
```
karussell: Why is this sorting code needed? Why not sort the enum as you need it?

otbutz: The tunnelcodes have a priority. eg: if a way is tagged with `tunnel=yes`,`hazmat:B=no`,`hazmat:E=no` the tunnel category would be the most restricted of the two which is `E`.

I'm going to replace the first SortedMap with a regular one. Looks like I have missed that one will refactoring.

otbutz: @karussell i've updated my PR

karussell: Ah, now I understand this better. So the basic problem is that for HAZMAT_SUBTAGS you don't want to create new strings in handleWayTags and put them in a TreeMap. Is this correct? The question is if this is really faster :) - did you had performance problems? Basically it is creating the string vs. "calculating the hashcode and the map lookup". I would still use a LinkedHashMap, so that the original order in the enum is not changed.

And the TUNNEL_CATEGORY_NAMES is only needed because we do not offer a compatible method in ReaderWay (?)

otbutz: > Ah, now I understand this better. So the basic problem is that for HAZMAT_SUBTAGS you don't want to create new strings in handleWayTags and put them in a TreeMap. Is this correct?

That's one point but i'm using the TreeMap mainly to enforce the reverse order. Would you rather prefer this approach?

```java
HazmatTunnel[] codes = HazmatTunnel.values();
for (int i = codes.length - 1; i >= 0; i--) {
    if (readerWay.hasTag("hazmat:" + codes[i].name(), "no")) {
        hazTunnelEnc.setEnum(false, edgeFlags, codes[i]);
        break;
    }
}
```

> And the TUNNEL_CATEGORY_NAMES is only needed because we do not offer a compatible method in ReaderWay (?)

I need to check a tag for multiple String values. So for the sake of readability and possibly a small performance gain, i've to keep the enum names as a Set/Array. I thought that a Map might be nice for this as it offers additional retrieval by name but `HazmatTunnel.valueOf()` should really be enough because there won't be any unexpected values at this point anymore.

Originally I wanted to use `boolean hasTag(List<String> keyList, Set<String> values)` but it's not that helpful if i still need to check for bad key=value combinations. A `getFirstPriorityTag(List<String> restrictions,Set<String> values)` would probably be helpful?

karussell: > So for the sake of readability and possibly a small performance gain

If you have not measured then I doubt that this is true ;). In fact looping over a small array can be faster than a hashcode+lookup.

> Would you rather prefer this approach?

Yes, I would have expected this as the other parsers also loop over their enums, but if you find TreeMap cleaner then this is fine as I don't think there will be big speed differences.

otbutz: @karussell i've ditched all Sets/Maps

karussell: Ok, thanks!

PR #1789, opened 2019-11-12
Use precompiled Pattern instances

otbutz:
I'm currently testing the import with all our new encoded values enabled and JFR reported that a lot of Pattern instances are being allocated.

As `String.replaceAll()` is not reusing it's `Pattern` instance, we create a new one each time it is called, which quickly becomes expensive.

Screenshot_20191112_163104


karussell:
Were you able to measure any speed ups for the import due to these changes?

otbutz:
Ran the import twice using the following setup:

* source: germany-latest.pbf 
* CH disabled
* LM enabled
* Xms23G Xmx23G
* all available encoded values
* Flightrecorder with preset `profile`
* measured from first log line until Dropwizard lists the available endpoints

## Results
* unpatched: 17m44s
* with precompiled patterns: 16m40s (:arrow_down:  ~6%)

## Memory allocation according to JFR *without* precompiled Patterns (109GiB total int[] allocation) 
Screenshot_20191114_173600

## Memory allocation according to JFR *with* precompiled Patterns (83.3GiB total int[] allocation) 
Screenshot_20191114_173619

PR #1790, opened 2019-11-12
Select appropriate CH profile when no weighting or vehicle is given

### LINKED ISSUE #1788: Request fails when not setting the weighting and/or the vehicle with CH and finite u-turn costs
If we specify something like `prepare.ch.weightings=fastest|u_turn_costs=40` in the configuration and then do not specify a weighting in `GHRequest`, GH tries to determine the weighting by looking at the `CHProfile(s)`. This leads to:

```
java.lang.IllegalArgumentException: Cannot find CH profile for weighting map {edge_based=true, weighting=fastest|u_turn_costs=40, vehicle=car} in entries: [fastest|car|edge_based=true|u_turn_costs=40.
```

easbar:
Fixes #1788 .

This PR reworks the CH profile selection logic and now also handles the cases where the weighting or the vehicle is missing. It now works like this: A request comes in with or without `vehicle,weighting,edge_based,u_turn_costs`. Missing values are treated as wildcards (match everything). This means that we do not need to specify anything on the request as long as there is only one possible match. When there are multiple matches we throw an error (with one exception, see in code). 

The CH profile selection got much simpler. The error responses are now more generic and maybe even more informative. For example if no CH profile matches the request we get:

```
Cannot find matching CH profile for your request.
requested:  fastest|*|edge_based=true|u_turn_costs=40
available: [fastest|car|edge_based=false|u_turn_costs=-1, fastest|car|edge_based=true|u_turn_costs=30, fastest|car|edge_based=true|u_turn_costs=50]
```

And when there are are too many matches we get something like:

```
There are multiple CH profiles matching your request. Use the `weighting`,`vehicle`,`edge_based` and/or `u_turn_costs` parameters to be more specific.
requested:  *|car|edge_based=true|u_turn_costs=*
matched:   [fastest|car|edge_based=true|u_turn_costs=30, fastest|car|edge_based=true|u_turn_costs=50]
available: [fastest|car|edge_based=false|u_turn_costs=-1, fastest|car|edge_based=true|u_turn_costs=30, fastest|car|edge_based=true|u_turn_costs=50]
```

Here `*` represents a wildcard, e.g. if the vehicle/weighting/edge_based/u_turn_costs are not given.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
+++ b/core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
@@ -20,14 +20,12 @@ package com.graphhopper.routing.ch;
 import com.graphhopper.routing.RoutingAlgorithmFactory;
 import com.graphhopper.routing.RoutingAlgorithmFactoryDecorator;
 import com.graphhopper.routing.util.HintsMap;
-import com.graphhopper.storage.CHGraphImpl;
 import com.graphhopper.storage.CHProfile;
 import com.graphhopper.storage.GraphHopperStorage;
 import com.graphhopper.storage.StorableProperties;
 import com.graphhopper.util.CmdArgs;
 import com.graphhopper.util.PMap;
 import com.graphhopper.util.Parameters.CH;
-import com.graphhopper.util.Parameters.Routing;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -35,7 +33,6 @@ import java.util.*;
 import java.util.concurrent.ExecutorCompletionService;
 import java.util.concurrent.ExecutorService;
 
-import static com.graphhopper.routing.weighting.TurnWeighting.INFINITE_U_TURN_COSTS;
 import static com.graphhopper.util.Helper.*;
 import static com.graphhopper.util.Parameters.CH.DISABLE;
 
@@ -63,7 +60,8 @@ public class CHAlgoFactoryDecorator implements RoutingAlgorithmFactoryDecorator 
 
     public CHAlgoFactoryDecorator() {
         setPreparationThreads(1);
-        setCHProfilesAsStrings(Arrays.asList(getDefaultProfile()));
+        // use fastest by default
+        setCHProfilesAsStrings(Collections.singletonList("fastest"));
     }
 
     @Override
@@ -229,10 +227,6 @@ public class CHAlgoFactoryDecorator implements RoutingAlgorithmFactoryDecorator 
         return this;
     }
 
-    private String getDefaultProfile() {
-        return chProfileStrings.isEmpty() ? "fastest" : chProfileStrings.iterator().next();
-    }
-
     public List<PrepareContractionHierarchies> getPreparations() {
         return preparations;
     }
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/graphhopper/routing/lm/LMAlgoFactoryDecorator.java
```diff
@@ -229,6 +229,11 @@ public RoutingAlgorithmFactory getDecoratedAlgorithmFactory(RoutingAlgorithmFact
         if (preparations.isEmpty())
             throw new IllegalStateException("No preparations added to this decorator");
 
+        // if no weighting or vehicle is specified for this request and there is only one preparation, use it
```
easbar: This was very ugly: If the weighting was missing it was added to the map in `CH(!)AlgoFactoryDecorator` and `LMAlgoFactoryDecorator` relied on the CH decorator to add it before we got here...If we did not pay attention to the missing weighting here  (now that we no longer add the weighting in the CH decorator) we would get normal AStar instead of LM, see `GraphHopperIT#testFlexMode_631`.

easbar:
Btw I first tried fixing #1788 in the same kind of style we selected the CH profile so far (see first commit), but I think the new version is much simpler (=better)

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
```diff
@@ -63,7 +60,8 @@
 
     public CHAlgoFactoryDecorator() {
         setPreparationThreads(1);
-        setCHProfilesAsStrings(Arrays.asList(getDefaultProfile()));
+        // use fastest by default
+        setCHProfilesAsStrings(Collections.singletonList("fastest"));
```
karussell: What if there is no fastest weighting like just shortest?

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -1381,4 +1381,32 @@ private GHResponse calcCurbsidePath(GraphHopper tmpHopper, GHPoint source, GHPoi
...
+                .setCHProfileStrings("fastest|u_turn_costs=40")
+                .setEdgeBasedCHMode(CHAlgoFactoryDecorator.EdgeBasedCHMode.EDGE_OR_NODE);
+        h.importOrLoad();
+
+        GHPoint p = new GHPoint(43.73397, 7.414173);
+        GHPoint q = new GHPoint(43.73222, 7.415557);
+        GHRequest req = new GHRequest(p, q);
+        // note that we do *not* set the weighting on the request, it will be determined automatically from the
+        // CH profile, see #1788
+        req.setWeighting("fastest");
```
karussell: I do not understand this comment as req.setWeighting is called. Can you explain :) ?

core/src/main/java/com/graphhopper/routing/lm/LMAlgoFactoryDecorator.java
karussell: Uh, sounds ugly.

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
easbar: Oh thanks for the review! Yes this is a mistake, I probably added this line there and back while testing, will remove it.

core/src/main/java/com/graphhopper/routing/ch/CHAlgoFactoryDecorator.java
easbar: This is not about the CH profile that is selected when a request comes in. The 'default' "fastest" here simply means that if we enable CH on `GraphHopper` it will use "fastest" unless we specify some `prepare.ch.weightings`. 

Speaking of the CH profile selection at time of a request: If only shortest is prepared and we do not specify a weighting shortest will be used. 

Does this answer your question ?

karussell: Ah, thanks!

Issue #1792, opened 2019-11-13
Move setting of segment size into `DataAccess` constructors


easbar:
Have not really looked into this, but it seems many `Storable` implementations have this method anyway (I guess because most of them use some `DataAccess` underneath), for example:

- `CHGraphImpl`
- `LocationIndex`
- `TurnCostExtension`
- `GraphStorage`
- `NameIndex`
- all `DataAccess`es

karussell:
Yes, would make sense. What is the reason you stumbled over this?

easbar:
I was checking where we are using `CHGraphImpl` instead of `CHGraph` and one of the reasons we do this e.g. in `GraphHopperStorage` was because of `setSegmentSize()`. It was similar when we discussed about using an interface for `TurnCostExtension`.

easbar:
Does it make sense to call `setSegmentSize` after a `DataAccess` has been used (written to) ? Or can this even be a wrong thing to do ? If it must never be called after writing to the `DataAccess` I wonder if we should rather remove it from the `DataAccess` interface and add a `segmentSize` parameter to the constructors of the objects that need it ?

easbar:
So if `setSegmentSize` may not be called after the first writes to the `DataAccess` its probably a bad idea to even expose it in `Storable` and I think it should rather be more hidden than exposed ?

karussell:
> Or can this even be a wrong thing to do ?

Yes, this can be even wrong (stumbled over this ugliness for the StringIndex again).

> I think it should rather be more hidden than exposed ?

Yes. The only current use case is to reduce the storage size for small data. E.g. for tests (small graph data) but also for they DataAccess `keys` in StringIndex or StorableProperties.

In early days I was experimenting with this setting and was hoping that there is a sweet spot regarding performance. But it wasn't and so I picked a size of ~1MB so that the smallest possible graph is not too big and a very big graph has not too many segments (although it still can be X-thousand)

easbar:
So something like `dir.find("nodes", daType, segmentSize)` ? This would mean that if the `segmentSize` parameter is given and the `.find()` call creates a new `DataAccess` it uses this `segmentSize` to do so. If `.find()` finds something it would just ignore the parameter. And then we have four overloads for `Directory.find` (with/without DAType and with/without segmentSize).

karussell:
[...]

PR #1794, opened 2019-11-13
Moves segment size setting of graphs into constructor

easbar:
Related to #1792. Removing the segment size setters from `DataAccess` seems to be a bigger task. For my purposes its sufficient for now to get rid of it on the graphs.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
+++ b/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
@@ -24,10 +24,7 @@ import com.graphhopper.util.EdgeExplorer;
 import com.graphhopper.util.EdgeIteratorState;
 import com.graphhopper.util.shapes.BBox;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 
 /**
  * This class manages all storage related methods and delegates the calls to the associated graphs.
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
```diff
@@ -60,6 +60,10 @@ public GraphHopperStorage(List<CHProfile> chProfiles, Directory dir, EncodingMan
     }
 
     public GraphHopperStorage(List<CHProfile> chProfiles, Directory dir, EncodingManager encodingManager, boolean withElevation, boolean withTurnCosts) {
+        this(chProfiles, dir, encodingManager, withElevation, withTurnCosts, -1);
+    }
+
+    public GraphHopperStorage(List<CHProfile> chProfiles, Directory dir, EncodingManager encodingManager, boolean withElevation, boolean withTurnCosts, int segmentSize) {
```
easbar: I think there should be a builder for this graph to get rid of these constructor variations. We have `GraphBuilder`, but honestly I almost never use it. I think it currently does 'too much', like its doing more than just being a builder, which is maybe good for a top-level user, but for internal code, tests etc. I almost always prefer using the constructor directly. WDYT shall we create another (pure) builder, or adjust the old `GraphBuilder`, or neither ?

core/src/main/java/com/graphhopper/storage/GraphStorage.java
```diff
@@ -24,8 +24,6 @@
 
     EncodingManager getEncodingManager();
 
-    void setSegmentSize(int bytes);
```
easbar: I think this whole interface is basically not needed. There is only one single usage in `OSMReader` to call `create` and `GraphStorage` even gets the `create` method from `Storable`. All these methods here are never used via the interface.

core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
easbar: Update: I found a compromise for `GraphBuilder`, its now a 'dumb' builder but it has some convenience setters like `.setMmap()`. The only thing I do not like now is that `build()` does not call `create()`. I think this should be the 'default'. I would just allow passing a negative value to `setBytes()` in which case the graph will not be created (just like `build()` does at the moment.

PR #1795, opened 2019-11-13
Remove unnecessary slight turns

### LINKED ISSUE #1441: Instructions: Remove unnecessary slight turns, if the alternative would be a sharp turn
Something that happens frequently is the creation of "slight turns" that are no real turns and users would just follow the road, but since there is another option for a turn, we create a turn instruction.

See this example: https://graphhopper.com/maps/?point=51.734514%2C9.225571&point=51.734643%2C9.22541

We create a "turn slight right" instruction. But this makes no sense. When for example driving [this route](https://graphhopper.com/maps/?point=51.733622%2C9.225914&point=51.735245%2C9.225125), a user could misinterpret the turn slight right instruction and turn right at the intersection like [this](https://graphhopper.com/maps/?point=51.733622%2C9.225914&point=51.73451%2C9.227051).

I think we could improve this situation, by checking if the orientation of the other turn. If the other turn is almost a u-turn (e.g. >140°) it might be ok to omit a turn instruction that is only a slight turn.

This is kind of related to #1007 

WDYT?

boldtrn:
Fixes #1441. We already had some logic to ignore unnecessary slight turns. I increased the angle to remove more of these examples. This changed the number of instructions in some IT. I had a look at these cases visually and the new version looks better to me. It's a bit hard to do that though, since the test data is quite old, and the map visualization moved on with the current data.


karussell:
Thanks!

> It's a bit hard to do that though, since the test data is quite old, and the map visualization moved on with the current data.

With the new MVT debug view this is easy :) ?

> This changed the number of instructions in some IT. 

Do you know why? It seems it changed only the index?

boldtrn:
> With the new MVT debug view this is easy :) ?

I thought about that as well :). I didn't use it, probably I should :+1: 

> Do you know why? It seems it changed only the index?

Yes, this is intended. Unnecessary slight turns were removed. Therefore the index of other instructions changed :).

PR #1797, opened 2019-11-16
make CHProfile selection pluggable

karussell:
minor improvement for #1790 to call Weighting.matches again. A disadvantage could be that Weighting.matches is now less strict.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ch/CHProfileSelector.java
+++ b/core/src/main/java/com/graphhopper/routing/ch/CHProfileSelector.java
@@ -32,16 +32,13 @@ import static com.graphhopper.routing.weighting.TurnWeighting.INFINITE_U_TURN_CO
  */
 public class CHProfileSelector {
     private final List<CHProfile> chProfiles;
-    // variables describing the requested CH profile
-    private final String weighting;
-    private final String vehicle;
+    private final HintsMap hintsMap;
     private final Boolean edgeBased;
     private final Integer uTurnCosts;
 
     private CHProfileSelector(List<CHProfile> chProfiles, HintsMap hintsMap) {
         this.chProfiles = chProfiles;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/graphhopper/routing/ch/CHProfileSelector.java
```diff
@@ -65,10 +62,7 @@ private CHProfile select() {
             if (uTurnCosts != null && p.getUTurnCostsInt() != uTurnCosts) {
                 continue;
             }
-            if (!weighting.isEmpty() && !getWeighting(p).equals(weighting)) {
-                continue;
-            }
-            if (!vehicle.isEmpty() && !getVehicle(p).equals(vehicle)) {
+            if (!p.getWeighting().matches(hintsMap)) {
```
easbar: If we do this we probably should use `.matches()` also a few lines below (lines 80+81)

karussell: In lines 80+81 we compare to Weightings, but here we compare the matching of a Weighting to the HintsMap. So, to use the matches method we would need `getWeighting().matches(hintsMap) && getWeighting().matches(hintsMap)` - and still both Weighting objects could be different. Or to remove the vehicle comparison we could use the Weighting.equals method instead of String.equals?

easbar: Maybe we can get rid of `Weighting#matches` altogether :)

karussell: Yeah, have created a separate issue as this seems to be quite involved: https://github.com/graphhopper/graphhopper/issues/1800

Will merge this here for now if no objections :) ?

PR #1469, opened 2018-09-28
Switch to org.locationtech's JTS version

michaz:
As discussed in #1039 and elsewhere. Seems to work.

PR #1803, opened 2019-11-20
upgrade JTS to 0.16.0

karussell:
We also need to update:

 - [x] our jackson-datatype-jts fork
 - [x] [mapbox-vector-tile-java](https://github.com/wdtinc/mapbox-vector-tile-java/pull/44) (-> #2698)

The minor advantage of 0.16.0 over 0.15.1 is that geotools is using 0.16.0 and never used 0.15.1 for its releases.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -20,7 +20,7 @@
         <dependency>
             <groupId>org.locationtech.jts</groupId>
             <artifactId>jts-core</artifactId>
-            <version>1.15.1</version>
+            <version>1.16.0</version>
         </dependency>
         <dependency>
             <groupId>junit</groupId>
--- a/web-api/pom.xml
+++ b/web-api/pom.xml
@@ -35,7 +35,7 @@
         <dependency>
             <groupId>com.graphhopper.external</groupId>
             <artifactId>jackson-datatype-jts</artifactId>
-            <version>0.12-2.5-1</version>
+            <version>1.0-2.7</version>
         </dependency>
         <dependency>
             <groupId>io.dropwizard</groupId>


devemux86:
That would cause Android compatibility problems, discussed in https://github.com/graphhopper/graphhopper/pull/1469#issuecomment-434980330.

Android JTS 1.16+ compatibility remains the same(?) forcing Java 8 and Android N (--min-api 24).

Related to https://github.com/locationtech/jts/issues/145

karussell:
Oh, wow. API 24 is so old now that I would have expected the usage numbers to be far below 10% but with this change we would actually exclude over 40%. (stark contrast to iOS where they support 2 versions and older have <10%)

The question is if we should further support Android when they have such a shitty Java support and additionally a bad update policy for their devices. Because for us productivity and fresh dependencies is really important and GH core is only still Java 7 because of Android. An example is geotools that uses either the old vividsolutions dependency or JTS 1.16.0 but not 1.15.1. Which means that using GraphHopper together with geotools is only possible at the moment with 19.x and then your jar contains two JTS libraries (as they have a different package this is nothing critical but very ugly). I don't say geotools is especially important for us, but in general these problems will increase the older our Java requirement stays and that is a problem.

So maybe we delay it for another year, but I fear we won't be able to further delay it. Or we have to rewrite GraphHopper in Kotlin then ;)

devemux86:
Android situation is certainly complicated...

For reference can see the https://developer.android.com/about/dashboards.
1% could mean thousands of devices out there that cannot be skipped easily.
Recommended min version remains for many Android 4.1 Jelly Bean (API 16).

There are 2 cases:
- Java version: many Android apps compiled with Java 8 don't run in all devices(!), so many prefer to compile still with Java 7 (and remain with Java 7 compatible libs) - until Java 8 becomes mandatory.
- Min API: in any case Android apps min API should be selected freely by developers and not mandated by third-party libraries.
See https://github.com/square/okhttp/issues/4481 which forced many from Android community to remain with previous [3.12.X branch](https://github.com/square/okhttp#requirements) due to new OkHttp Android / Java restrictions.

Question is indeed if GraphHopper wants to still support Android...

For many Android developers the solution for libraries that force Java 8 or large min SDK requirements is unfortunately to remain at old branches, fork them, even drop them.

Or don't know if GraphHopper could have more fine tuned modules organization?
On Android there is need to read graphs and perform routing, i.e. like _old_ GraphHopper core.
That could be done until now without JTS or GeoTools, which may be needed elsewhere (other modules or web platform) but Android worked fine all this time with just GraphHopper module.

devemux86:
> Or we have to rewrite GraphHopper in Kotlin then ;)

That could lead to even more [unwelcome](https://github.com/square/okhttp/issues/4723#issuecomment-511269404) Android dependencies: https://github.com/square/okhttp/issues/4723 :slightly_smiling_face:

karussell:
Ok, closing for now. Maybe we find another solution.

easbar:
> That could lead to even more unwelcome Android dependencies: square/okhttp#4723 slightly_smiling_face

@devemux86 can you explain this a bit? Maybe this is a stupid question, but if Kotlin can compile to java 6 bytecode: https://kotlinlang.org/docs/reference/faq.html#which-versions-of-jvm-does-kotlin-target why would it not be feasible to add the kotlin compiler to graphhopper (allowing to write some kotlin), compile everything to java 6 and use this bytecode on android? And if the Kotlin compiler can do this why can we not use say java 12 and compile that to java 6 and use it on android?

devemux86:
Developers mention in https://github.com/square/okhttp/issues/4723 that its new version now [pushes a new kotlin dependency](https://github.com/square/okhttp/blob/master/okhttp/build.gradle#L26) in the apps in order to use the lib and also cannot know if many libs start doing that, how those new runtime dependencies can be compatible if have different versions.

easbar:
Ah it needs runtime dependencies? Do you know why?

devemux86:
> Ah it needs runtime dependencies? Do you know why?

Probably this [discussion](https://stackoverflow.com/questions/30565036/why-kotlin-needs-to-bundle-its-runtime-after-compiled) can clarify that?

easbar:
Ah ok thanks for the pointer, this explains it I think. So we need to include some additional kotlin library. And why is this bad? Because it increases the .apk size?

devemux86:
Besides the issue that can be different runtime versions in classpath and don't know what happens then, any new dependency that adds (some may thousands) methods can lead to [problems](https://developer.android.com/studio/build/multidex).
Generally on Android we try to avoid any unneeded dependencies - if possible. :slightly_smiling_face:

easbar:
> Generally on Android we try to avoid any unneeded dependencies

Yes, and this is a wise thing to do :) I was just trying to understand what limitations there are.

otbutz:
Could we aim for JTS 1.17.0?

karussell:
The first step I would prefer to use 0.16.0 (or 0.16.1) as many dependencies use it. And if all is fine surely we can try 0.17.x.

PR #1804, opened 2019-11-22
Avoid throwing NumberFormatException in the OSMValueExtractor

otbutz:
I've modified OSMValueExtractor to return `Double.NaN` for invalid values instead of throwing exceptions. Having a control flow which doesn't rely on unchecked exceptions is one point and the overhead of creating the exceptions is the other one:

Screenshot_20191122_104055

It's on par with the memory allocation of the pbf parsing logic only because it's gathering the stacktraces which we don't even use later on.

Note that `Double.parseDouble()` might still throw a `NumberFormatException` for some cases but i'm not sure if it's worth to add an extra test method before invoking it.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxAxleLoadParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxAxleLoadParser.java
@@ -32,15 +32,13 @@ import java.util.List;
 public class OSMMaxAxleLoadParser implements TagParser {
 
     private final DecimalEncodedValue maxAxleLoadEncoder;
-    private final boolean enableLog;
 
     public OSMMaxAxleLoadParser() {
-        this(MaxAxleLoad.create(), false);
+        this(MaxAxleLoad.create());
     }
-
-    public OSMMaxAxleLoadParser(DecimalEncodedValue maxAxleLoadEncoder, boolean enableLog) {
+    
+    public OSMMaxAxleLoadParser(DecimalEncodedValue maxAxleLoadEncoder) {
         this.maxAxleLoadEncoder = maxAxleLoadEncoder;
-        this.enableLog = enableLog;
     }
 
     @Override
@@ -49,10 +47,8 @@ public class OSMMaxAxleLoadParser implements TagParser {
     }
 
     @Override
-    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, EncodingManager.Access access,
-                                 IntsRef relationFlags) {
-        OSMValueExtractor.extractTons(edgeFlags, way, maxAxleLoadEncoder,
-                Collections.singletonList("maxaxleload"), enableLog);
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, EncodingManager.Access access, IntsRef relationFlags) {
+        OSMValueExtractor.extractTons(edgeFlags, way, maxAxleLoadEncoder, Collections.singletonList("maxaxleload"));
         return edgeFlags;
     }
 }
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/helpers/OSMValueExtractor.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/helpers/OSMValueExtractor.java
@@ -28,23 +28,26 @@ public class OSMValueExtractor {
     }
 
     public static void extractTons(IntsRef edgeFlags, ReaderWay way, DecimalEncodedValue valueEncoder, List<String> keys, boolean enableLog) {
-        String value = way.getFirstPriorityTag(keys);
-        try {
-            double val = stringToTons(value);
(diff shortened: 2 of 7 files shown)


karussell:
Ah, yeah. This is ugly if used as control flow.

core/src/main/java/com/graphhopper/routing/util/parsers/helpers/OSMValueExtractor.java
```diff
@@ -79,7 +89,7 @@ public static double stringToMeter(String value) {
         value = INCH_PATTERN.matcher(value).replaceAll("in");
         value = FEET_PATTERN.matcher(value).replaceAll("ft");
         if (isInvalid(value))
-            throw new NumberFormatException("Cannot parse value for 'meter': " + value);
+            return Double.NaN;
```
karussell: As we do not log these invalid cases I think we should get rid also of the code above:

```
if (enableLog) {
  LOG.warn("Unable to extract tons from malformed road attribute '{}' for way (OSM_ID = {}).", rawValue, way.getId());
}
```

WDYT?

otbutz: I'm not a fan of useless log output and this one qualifies IMHO. Nobody is going to wade through thousand lines of messages which merely tell you that the values of certain tags are malformed.

core/src/main/java/com/graphhopper/routing/util/parsers/helpers/OSMValueExtractor.java
```diff
@@ -28,23 +24,23 @@ private OSMValueExtractor() {
     }
 
     public static void extractTons(IntsRef edgeFlags, ReaderWay way, DecimalEncodedValue valueEncoder, List<String> keys, boolean enableLog) {
```
karussell: Would you also remove this enableLog parameter :) ? (Same also for extractMeter)

otbutz: Ah, thanks. Missed it!

core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxAxleLoadParser.java
```diff
@@ -32,15 +32,9 @@
...
-    private final boolean enableLog;
 
     public OSMMaxAxleLoadParser() {
-        this(MaxAxleLoad.create(), false);
-    }
-
-    public OSMMaxAxleLoadParser(DecimalEncodedValue maxAxleLoadEncoder, boolean enableLog) {
-        this.maxAxleLoadEncoder = maxAxleLoadEncoder;
-        this.enableLog = enableLog;
+        this.maxAxleLoadEncoder = MaxAxleLoad.create();
```
karussell: I would not remove this. I think the additional constructor is a meaningful addition, so that the user could increase or decrease the precision. (same for the other parsers)

core/src/main/java/com/graphhopper/routing/util/parsers/helpers/OSMValueExtractor.java
boldtrn: > Nobody is going to wade through thousand lines of messages which merely tell you that the values of certain tags are malformed.

Just a note: I do this. I filter the logs for things I am interested in and then I fix it in OSM where there is a clear tagging error. I try to take an hour from time to time to check the logs and fix the worst things. I probably fixed several hundred tags and I think others might do the same.

karussell: But maybe this should be part of a "clean-up OSM" website, where multiple contributors could improve mistakes without running GraphHopper? There are also many other problems that I would find ugly if we print them all into the logs like e.g. #1807 or subnetworks in general.

otbutz: > But maybe this should be part of a "clean-up OSM" website

Very good idea. Projects like [StreetComplete](https://github.com/westnordost/StreetComplete#screenshots) are better suited for this because they can be used by contributors without technical background.

boldtrn: > But maybe this should be part of a "clean-up OSM" website, where multiple contributors could improve mistakes without running GraphHopper?

Sure, that would be even better :). We should think about building one or maybe contributing to an existing website to also show routing related information. 

> I would find ugly if we print them all into the logs like e.g. #1807 or subnetworks in general.

Yes, maybe we only do this for small regions or maybe configurable or something like this :)

otbutz: I already thought about using the data provided by [taginfo](https://taginfo.openstreetmap.org/keys/maxweight#values) to check the coverage of our extraction logic.

PR #1812, opened 2019-12-02
Add support for details=mtb_scale and details=sac_scale

Nakaner:
This pull request adds support to make `mtb:scale=*` and `sac_scale=*` accessible using the `details=*` parameter.

`mtb:scale` (the URL parameter uses `mtb_scale` to be in line with all the other options which are already implemented) is stored as unsigned integer because it has only unsigned integers as values. `sac_scale` uses an enum.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMSacScaleParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMSacScaleParser.java
@@ -1,0 +1,58 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.routing.util.parsers;
+
+import java.util.List;
+
+import static com.graphhopper.routing.profiles.SacScale.NONE;
+
+import com.graphhopper.reader.ReaderWay;
+import com.graphhopper.routing.profiles.EncodedValue;
+import com.graphhopper.routing.profiles.EncodedValueLookup;
+import com.graphhopper.routing.profiles.EnumEncodedValue;
+import com.graphhopper.routing.profiles.SacScale;
+import com.graphhopper.routing.util.EncodingManager;
+import com.graphhopper.storage.IntsRef;
+
+public class OSMSacScaleParser implements TagParser {
+
+    private final EnumEncodedValue<SacScale> sacScaleEnc;
+
+    public OSMSacScaleParser() {
+        this.sacScaleEnc = new EnumEncodedValue<>(SacScale.KEY, SacScale.class);
+    }
+
+    @Override
+    public void createEncodedValues(EncodedValueLookup lookup, List<EncodedValue> link) {
+        link.add(sacScaleEnc);
+    }
+
+    @Override
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, EncodingManager.Access access, IntsRef relationFlags) {
+        if (!access.isWay())
+            return edgeFlags;
+
+        String value = readerWay.getTag("sac_scale");
+        if (value == null)
+            return edgeFlags;
+        SacScale scale = SacScale.find(value);
+        if (scale != NONE)
+            sacScaleEnc.setEnum(false, edgeFlags, scale);
+        return edgeFlags;
+    }
+}
--- a/core/src/main/java/com/graphhopper/routing/profiles/MtbScale.java
+++ b/core/src/main/java/com/graphhopper/routing/profiles/MtbScale.java
@@ -26,15 +26,36 @@ package com.graphhopper.routing.profiles;
  * @author Michael Reichert
  *
  */
-public class MtbScale {
+public enum MtbScale {
+    NONE("none"),
+    MTB_0("0"),
+    MTB_1("1"),
+    MTB_2("2"),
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/OSMSacScaleParser.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, EncodingManager.Access access, IntsRef relationFlags) {
+        if (!access.isWay())
+            return edgeFlags;
+
+        String value = readerWay.getTag("sac_scale");
+        if (value == null)
+            return edgeFlags;
```
karussell: this check is not necessary

Nakaner:
@karussell wrote:
> In general I would like to be as close to OSM as possible but some tags are just wrong, too complicated or inconsistent or won't work for other data sources.
> 
> Is there a better name for mtb_scale? Or, should we use an integer value for sac_scale? This way we could also use it easier in future vehicle profiles like "forbid access if sac_scale > 3"?

Do you want me to implement sac_scale the same way as mtb_scale (storing it as unsigned integers)? I choosed integer for mtb_scale only because they are stored as numbers in OSM.

karussell:
We need to think of a solution for the inconsistent naming. So we name them differently (not both "xy_scale") or make them the same (enum or integer). Or something better that you would want or expect in the routing or path details response.

Nakaner:
I have no preference with regard to the names and whether to use integers or enums for the implementation. Tell me what you want and I will change my pull request.

karussell:
Ok, let's also use integers for `sac_scale` like 0 for default, 1 for T1, 2 for T2 etc?

boldtrn:
Default values are a good point @karussell. How should we handle mtb:scale=0? IMHO there is a different meaning if a way has mtb:scale set at all or no mtb:scale? 

While it's only a minority of the mtb:scale values, some contain: `1+` or `1-`. According to the [wiki](https://wiki.openstreetmap.org/wiki/Key:mtb:scale) these are valid values. Currently these values would result in "no value". These are probably [<3% of all values](https://taginfo.openstreetmap.org/keys/mtb:scale#values).

Nakaner:
I admit that I overlooked the + and - classes on the wiki. The more details we have the more bits we need. The 6 classes fit good into 3 bits, adding + and - would require 5 bits or a more complex encoding.

boldtrn:
> The 6 classes fit good into 3 bits, adding + and - would require 5 bits or a more complex encoding.

Sorry I wasn't very specific with my feedback :). I meant to remove the '+' and '-', for example `1+` would become `1`. That would be better than assuming it had not mtb:scale.

Nakaner:
The new commits I pushed a today store MtbScale as enums in order to get a dedicated value for "undefined" because 0 is a valid value for that value in OSM. I implemented the "rounding" of + and - suffixes and added a test for MtbScale because it is not trivial any more.

karussell:
Thanks for the improvements!

There is still the naming consistency issue I mentioned above. The best and easiest fix would be using integers from 0 to 6. Unfortunately this won't work well for the 0 of mtb:scale as it does not mean "no mtb:scale tag". What do you think about using:

`enum Scale { OTHER, SCALE_0, SCALE_1, ..., SCALE_6 }`

for both mtb_scale and sac_scale? And for sac_scale we would just never use SCALE_0? 

(BTW: this would also work for "horse scale" using SCALE_1-6)

karussell:
Two more comments to this important pull request:

 1. we need an additional MISSING value like we added to TrackType
 2. we need to think again about the enum index. E.g. it would be nice if `trackTypeEnum.ordinal() == "scale x"`. E.g. `SCALE_1.ordinal()` should be 1.

But then we would need to reorder this to `SCALE_0, SCALE_1, ..., SCALE_6, MISSING, OTHER` and then the default value 0 would mean `SCALE_0` but it should be better `MISSING`. Tricky.

joe-akeem:
@Nakaner are you still working on this? I would be interested to take over and finish this one otherwise...

karussell:
Probably the easiest thing is to avoid mtb_scale for the name. sac_scale fits easily into our schema where 0 means MISSING. mtb:scale however should probably use DecimalEncodedValue and be named like `mtb_difficulty`? 

From the wiki it is also unclear what the minus means. Does it mean that "1-" is 0.7 and "1+" the 1.3? But what is then "0-"? This does not make much sense, but we also do not need to clarify it here and can later improve when we use a decimal instead of the enum. (The default should be "no tag" and the raw number should be usable in a comparison.)

boldtrn:
mtb_scale is using http://www.singletrail-skala.de/ as reference. mtb_scale=0 means already that it's not suitable for most road bikes and inexperienced riders. IIRC 1+ means that it's a bit more challenging than the average 1. I think it would be fine for the details to ignore +/- for mtb scale?

joachimLengacher:
> mtb_scale=0 means already that it's not suitable for most road bikes and inexperienced riders

I agree! So `0` is not a good value for "missing". Maybe `-1`? `mtb:scale` goes from `0` to `6`. 

And I also agree to cut off `+` and `-` for simplicity.

joe-akeem:
> mtb_scale=0 means already that it's not suitable for most road bikes and inexperienced riders

I agree! So `0` is not a good value for "missing". Maybe `-1`? `mtb:scale` goes from `0` to `6`. 

And I also agree to cut off `+` and `-` for simplicity.

karussell:
My only objection would still be the same name. If it is called "scale" then it will be confusing that MTB_SCALE1.ordinal() != SAC_SCALE1.ordinal(). Because we would need something like MTB_SCALE_MISSING which should be the default (raw int value 0)

E.g. "horse scale" has a 'correct' name as it goes from 1 to 6 but TrackType is called "grade" as it goes from 1 to 5.

joe-akeem:
@Nakaner, I guess I would need permissions for your repository in order to push changes to this branch. Any chance to get them? Thanks!

joe-akeem:
So we don't have to wait for the permissions I created my own branch and PR (replacing this): #2158
In addition to `mtb:scale` and `sac_scale` I also added support for `mtb:scale:uphill` and `horse_scale`.

I decided to use `EnumEncodedValue` for all the tags for the following reasons:
- `mtb:scale=0` is a valid value (lowest difficulty) so it can't be an `IntEncodedValue` since for those `0` is the default value (missing)
- `mtb:scale`, `sac_scale` and `horse_scale` are all related, specifying a difficulty for different "vehicles". So IMHO they should be all encoded the same way.

The type names all end on *Scale: `MtbScale`, `MtbUphillScale`, `SacScale` and `HorseScale`.

What do you think?

Issue #30, opened 2013-04-09
Bug in Location2id index


karussell:
If you look into http://graphhopper.com/maps/?point=47.100541,17.932842&point=47.102981,17.931297 the destination is close to the way but matches the other more western node.

![location2id-index-bug](https://f.cloud.github.com/assets/129644/359061/b60f6b46-a153-11e2-8527-f791a8556038.png)

Improve new index to close #17

PR #1816, opened 2019-12-04
Enforce non-CH graphs for all LocationIndex implementations

easbar:
These calls to `CHGraph#disconnect` aren't doing anything and there aren't even any shortcut edges involved in this test. This has been like this since a long time, for example already in 2015, e.g. here 7f80425b6a0af9bdfef12c8a873676e39e0a04a6. Maybe we should rename the test or remove it even (is it still testing something useful) ?


karussell:
Yeah, very old stuff. Was introduced in #30 but at that time we had two different LocationIndex implementations. One for CH and one for none-CH graphs. So probably we do not need this anymore.

easbar:
Ok Ill remove the entire test then

easbar:
Now I am throwing `IllegalState` for all `LocationIndex` implementations and deleted the `LocationIndexCHTest`.

karussell:
Or maybe we remove all other LocationIndex implementations or move them into the tests? I think the only used one is the full edge index for correctness tests.

easbar:
> Or maybe we remove all other LocationIndex implementations or move them into the tests? I think the only used one is the full edge index for correctness tests.

Ok go ahead if you want, for me its sufficient to get rid of the .disconnect calls for now :)

PR #1823, opened 2019-12-12
Work with interfaces instead of specific RoutingAlgorithmFactory implementations

otbutz:
This PR adds support to inject custom RoutingAlgorithmFactoryDecorators without breaking LM/CH routing compatibility. The generated factories only need to implement `CHRoutingAlgorithmFactory` and `DecoratingRoutingAlgorithmFactory`.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -1062,8 +1062,8 @@ public class GraphHopper implements GraphHopperAPI {
                     }
                     // if LM is enabled we have the LMFactory with the CH algo!
                     RoutingAlgorithmFactory chAlgoFactory = tmpAlgoFactory;
-                    if (tmpAlgoFactory instanceof LMAlgoFactoryDecorator.LMRAFactory)
-                        chAlgoFactory = ((LMAlgoFactoryDecorator.LMRAFactory) tmpAlgoFactory).getDefaultAlgoFactory();
+                    if (tmpAlgoFactory instanceof DecoratingRoutingAlgorithmFactory)
+                        chAlgoFactory = ((DecoratingRoutingAlgorithmFactory) tmpAlgoFactory).getDefaultAlgoFactory();
 
                     if (chAlgoFactory instanceof CHRoutingAlgorithmFactory) {
                         CHProfile chProfile = ((CHRoutingAlgorithmFactory) chAlgoFactory).getCHProfile();
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -23,7 +23,7 @@ import com.graphhopper.reader.dem.*;
 import com.graphhopper.reader.osm.conditional.DateRangeParser;
 import com.graphhopper.routing.*;
 import com.graphhopper.routing.ch.CHAlgoFactoryDecorator;
-import com.graphhopper.routing.ch.CHRoutingAlgorithmFactory;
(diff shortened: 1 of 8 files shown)


easbar:
Can you explain a bit how this is useful ? So far I do not see how this is an improvement, but also do not think its a problem to merge this. Btw I am not a fan of the current `RoutingAlgorithmFactory` approach.

otbutz:
I'd like to inject a custom decorator in order to reuse Dijkstra OTM algorithm instances.

easbar:
Ok, honestly I am a bit undecided. This whole routing algo factory / decorator thing needs improvement or even should be removed and in any case is very likely to change, so I am not sure if we should add more classes/interfaces going this way. I do not see how `CHRoutingAlgorithmFactory` provides a useful abstraction as an interface, why should there be a `getCHProfile()` method ?

On the other hand adding two interfaces does not really hurt either and if its useful for you maybe its useful for others as well. I just like to say that this should not be considered stable or anything.

easbar:
> Work with interfaces instead of specific RoutingAlgorithmFactory implementations

Essentially we *are* using an interface (`RoutingAlgorithmFactory`) but it does not provide the right abstraction and therefore we have to case it for the different special cases, and now we need even more interfaces for the special cases. IMO this is just a hint that the current concept of a `RoutingAlgorithmFactory` is flawed.

otbutz:
> IMO this is just a hint that the current concept of a `RoutingAlgorithmFactory` is flawed.

I agree. All this `instanceof` and factory-switcheroo shouldn't be necessary:

https://github.com/graphhopper/graphhopper/blob/03dd6d4ec9320a9ec11677006777b5e3a63ae6b2/core/src/main/java/com/graphhopper/GraphHopper.java#L1064-L1091

The main problem seems to be obtaining the `CHProfile`.

Wouldn't it be easier to expose something like `CHAlgoFactoryDecorator.selectProfile(HintsMap)` ?

easbar:
Well the real problem is the `Graph` parameter in `RoutingAlgorithmFactory.createAlgo()`. `RoutingAlgorithmFactory` does not just provide a 'solution/solver' for a shortest path problem, but in the case of `CHRoutingAlgorithmFactory` we have to pass the right graph into the method (the `CHGraph`). The reason we do this is that it might be a `QueryGraph`. The only reason we have this `getCHProfile` method is that we need the profile to obtain the `CHGraph` so we can wrap a `QueryGraph` around it. `CHRoutingAlgorithmFactory` should not expose that it is even using a `CHGraph` (this is just an implementation detail of this algorithm), but still we need to pass information about the virtual nodes/edges, because this is only known at request time. Currently we pass `QueryGraph`, but this is too much information in the sense that we need to obtain the `CHGraph` to create it. I think the information that *should* be passed here is `GraphModification` (its a bad name, but it contains just the information needed here), see #1751 and the comment in `CHRoutingAlgorithmFactory#createAlgo`.

easbar:
So for me it comes down to the question where to put the abstraction: We want to provide different solvers for the shortest path problem, everything that belongs to the 'problem definition' should be created before we select the specific implementation and everything that is specific to the implementation should be encapsulated and hidden behind the abstraction. With `CHGraph` we are violating this principle (because of `QueryGraph`).

karussell:
> This whole routing algo factory / decorator thing needs improvement 

I strongly agree. I frequently stumble over bugs in this regard and a silent fallback (from CH to LM or from LM to Dijkstra) is also hard to detect/test. Have created: #1889.

Hope it is ok to close here. If not, please reopen.

PR #1824, opened 2019-12-12
Integrate PT

michaz:
Goal: _Functionally_ integrate PT with rest of GraphHopper, meaning I can put PT into a regular instance that also has other modes, CH etc.

Secondary goal: Also integrate the _code_ a bit.

I still use my own web resources. I just forward requests with `?vehicle=pt` to my own resource. (Invisibly. No HTTP redirect, just give Jersey a hint where the request must go.)

`GraphHopperGtfs` is now analogous to `GraphHopperOSM` (and inherits from it, for better or worse).

I needed two extra entry points in `GraphHopper`, and I still need to trick `GraphHopper` in two places: First, I may want to create a Graph completely without an OSM network, so I need to bypass a validation. Second, I need to reset the `LocationIndex` because I need it after importing OSM, but I can only finish it after importing GTFS.

In `/info`, I need to add PT as an extra mode if we have it, because it doesn't correspond to a `FlagEncoder`.


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -521,6 +521,7 @@ public GraphHopper init(CmdArgs args) {
         if (!flagEncodersStr.isEmpty() || !encodedValueStr.isEmpty()) {
             if (!encodedValueStr.isEmpty())
                 emBuilder.addAll(tagParserFactory, encodedValueStr);
+            registerCustomEncodedValues(emBuilder);
```
karussell: Ah, this is a bit unfortunate that this is necessary but we would need yet another factory (EncodedValueFactory) plus a method in EncodingManager to support this.

michaz: Yeah, talking about factories, extension points and such is a new issue, or epic. But I don't think it's that bad in this case. "Builder" sounds like something that should be safe to pass around and add stuff to.

Issue #1808, opened 2019-11-28
Add support for user defined SpatialRules


otbutz:
As discussed in https://github.com/graphhopper/graphhopper/pull/1799 we'd like to move from  SpatialRules defined in the code to external rule files which can be edited and extended by the user.

To sum up what we've gathered so far:
*  allow the user to define a directory with multiple geojson files describing the countries/states/areas he wants to manipulate when the graph is being built
* the accompanying rule files(e.g written in YAML) reside in a second directory
  * a rule targets one or more (multi)polygons by selecting them according to their properties
  * rules should have readable names in order to show applied rules later in the PathDetails
  * the area of rules might overlap(e.g state rules overwrite country rules) or multiple rules might share the same polygon. Therefore a rule should have some kind of priority score
  * rules can be defined sparsely; if a rule has no value for a specific situation the next rule with a lower priority will be queried with a fallback/default rule being the last one in the chain

TODO:
* [ ] load border polygons from multiple files
* [ ] define a rule format
* [ ] store the applied rules in the graph
   * StringIndex seems like a good idea but the `long` pointer is problematic. Maybe introduce a second layer of indirection? ```int id``` -> ```long pointer```
* [ ] extend the SpatialRuleLookup to handle overlapping rules
* [ ] how do we treat overlapping rules with the same priority?

Related issues:
* #1427
* #1680

otbutz:
I put a bit of thought into #1427 and came to the conclusion that rule files are simply not able cover the full span of potential use cases. They'd have to define logical operations on arbitrary tags, deal with malformed values and tag priorities.

This would lead us to one of the following scenarios:
* abandon #1427 and limit the rules to "case x: value y" logic
* hybrid approach: use external files for simple rules and allow to create more indepth ones via code
* drop the rule files idea

Otherwise we would end up reimplementing something like [JsonLogic](http://jsonlogic.com/) :face_with_head_bandage: 

@karussell @boldtrn Your opinion on this?

karussell:
I like the hybrid approach and we could start with the simple use cases that can be done in the files (like setting a certain EncodedValue like car.average_speed or car.access or max_speed to the same value for the defined area) and allow later hooking into this process.

Thanks for the JsonLogic link, interesting.

> load border polygons from multiple files

For better maintaining these rules we could have a separate folder with 2 files per SpatialRule like `de.rule.json` and `de.border.json` and then set this folder in the config.yml.

> store the applied rules in the graph 

[...]

PR #1827, opened 2019-12-16
Load spatial rule borders from a directory

otbutz:
I've modified the spatial rules handling in order to support loading multiple geojson files. This is the first step towards solving #1808

### DIFF
--- a/web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
+++ b/web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
@@ -77,10 +77,10 @@ public class GraphHopperManaged implements Managed {
     @Override
     public void start() {
         graphHopper.importOrLoad();
-        logger.info("loaded graph at:" + graphHopper.getGraphHopperLocation()
-                + ", data_reader_file:" + graphHopper.getDataReaderFile()
-                + ", encoded values:" + graphHopper.getEncodingManager().toEncodedValuesAsString()
-                + ", " + graphHopper.getGraphHopperStorage().toDetailsString());
+        logger.info("loaded graph at:{}, data_reader_file:{}, encoded values:{}, {}",
+                        graphHopper.getGraphHopperLocation(), graphHopper.getDataReaderFile(),
+                        graphHopper.getEncodingManager().toEncodedValuesAsString(),
+                        graphHopper.getGraphHopperStorage().toDetailsString());
     }
 
     public GraphHopper getGraphHopper() {
--- a/web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
+++ b/web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
@@ -33,10 +33,18 @@ import io.dropwizard.lifecycle.Managed;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.BufferedReader;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
 
 import static com.graphhopper.util.Helper.UTF_CS;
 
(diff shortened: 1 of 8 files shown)


web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
```diff
@@ -61,26 +69,33 @@ public GraphHopperManaged(CmdArgs configuration, ObjectMapper objectMapper) {
...
-        if (!spatialRuleLocation.isEmpty()) {
+        String spatialRuleBordersDirLocation = configuration.get("spatial_rules.borders_directory", "");
+        if (!spatialRuleBordersDirLocation.isEmpty()) {
             final BBox maxBounds = BBox.parseBBoxString(configuration.get("spatial_rules.max_bbox", "-180, 180, -90, 90"));
-            try (final InputStreamReader reader = new InputStreamReader(new FileInputStream(spatialRuleLocation), UTF_CS)) {
-                JsonFeatureCollection jsonFeatureCollection = localObjectMapper.readValue(reader, JsonFeatureCollection.class);
-                SpatialRuleLookupHelper.buildAndInjectSpatialRuleIntoGH(graphHopper, maxBounds, jsonFeatureCollection);
+            final Path bordersDirectory = Paths.get(spatialRuleBordersDirLocation);
+            List<JsonFeatureCollection> jsonFeatureCollections = new ArrayList<>();
+            try (DirectoryStream<Path> stream = Files.newDirectoryStream(bordersDirectory, "*.{geojson,json}")) {
```
karussell: This is nice! I can use this für #1776 too.

web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
```diff
@@ -61,26 +69,33 @@ public GraphHopperManaged(CmdArgs configuration, ObjectMapper objectMapper) {
         } else {
             graphHopper = new GraphHopperOSM(landmarkSplittingFeatureCollection).forServer();
         }
-        String spatialRuleLocation = configuration.get("spatial_rules.location", "");
-        if (!spatialRuleLocation.isEmpty()) {
+        String spatialRuleBordersDirLocation = configuration.get("spatial_rules.borders_directory", "");
```
karussell: Can we throw an exception if property `spatial_rules.location` is still used and inform about the new property?

web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
devemux86: Note that [DirectoryStream](https://developer.android.com/reference/java/nio/file/DirectoryStream) was added on Android in API level 26 (Android 8.0 Oreo). :slightly_smiling_face:

karussell: It is not in the `core` module, so this should not be a problem.

karussell: See https://github.com/graphhopper/graphhopper/pull/1841/files#diff-130ed9726000f7f80461f0c819f1e5c4R82

web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
otbutz: Sure.

web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
karussell: But it seems there is no simple way to get the file name without file extension and additionally one has to create a Buffer from the Path, so I think I'll keep using the newly introduced Helper.listFiles method https://github.com/graphhopper/graphhopper/pull/1841/files#diff-4e518ca1e483cdb8f16fea150c2fb150R171 (this would also make this code here a few lines shorter due to the catch stuff)

otbutz: The correct handling of paths makes the `Path` API superior IMHO:
```java
Path folder = Paths.get("a", "b", "c");
Path file = folder.resolve("d");
```
No need to worry about OS specific stuff or if there's a missing `\` or `/`.

otbutz:
Closing in favor of #1829

Issue #1680, opened 2019-08-05
SpatialRuleLookup issue with Polygons with holes


boldtrn:
When creating Polygons from geojson in the `SpatialRuleLookupBuilder`, we call the method `Polygon#create(org.locationtech.jts.geom.Polygon polygon)`.

The create method uses the method `org.locationtech.jts.geom.Polygon#getNumPoints`, which returns all points of the Polygon, including the points for the holes. Therefore, the holes and the connection to the whole become part of the Polygon, which can lead false positive contain results, even for points outside of the polygon.

A quick-fix could be to only use the shell:
```
Coordinate[] shell = polygon.getExteriorRing().getCoordinates();

        double[] lats = new double[shell.length];
        double[] lons = new double[shell.length];

        for(int i = 0; i < shell.length; ++i) {
            lats[i] = polygon.getCoordinates()[i].y;
            lons[i] = polygon.getCoordinates()[i].x;
        }

        return new Polygon(lats, lons);
``` 

The proper solution would be to correctly read holes as holes and use them in the contains lookup.

karussell:
Was fixed with #1680.

PR #1829, opened 2019-12-18
Use JTS to determine applicable spatial rules

otbutz:
Builds on top of #1827 and uses JTS classes to handle border geometries and SpatialLookup.

This should also solve #1680 as JTS polygons are able to deal with holes.


karussell:
Cool, thanks! Probably after this we should get rid of our own Polygon class and also use the JTS version elsewhere.

otbutz:
> Cool, thanks! Probably after this we should get rid of our own Polygon class and also use the JTS version elsewhere.

`BBox` would also be a good candidate for removal.

boldtrn:
Really nice, thanks for the great work. Did you have a chance to compare the performance of the JTS lookup vs. the array lookup?

otbutz:
> Really nice, thanks for the great work. Did you have a chance to compare the performance of the JTS lookup vs. the array lookup?

Just did a small benchmark with a dummy rule for each country in our geojson file on my machine with an i7-4790 and OpenJDK 11. Looking up the rules for 1 million random points in central europe takes about 1.5 seconds.

If i change the underlying index from STRtree to QuadTree the execution time goes up to about 2 seconds.
Screenshot_2020-01-03 Bounding Box Tool Metadata Enrichment for Catalogue Records by Visually Selecting Geographic Coordina


```java
SpatialRuleLookup spatialRuleLookup = SpatialRuleLookupBuilder.buildIndex(Collections.singletonList(Jackson.newObjectMapper().readValue(reader, JsonFeatureCollection.class)), "ISO_A3", rulePerCountryFactory);

List<GHPoint> randomPoints = new ArrayList<>();
for (int i = 0; i < 1_000_000; i++) {
    double lat = 46d + Math.random() * 7d;
    double lon = 6d + Math.random() * 21d;
    randomPoints.add(new GHPoint(lat, lon));
}

long start = System.nanoTime();
for (GHPoint point : randomPoints) {
    spatialRuleLookup.lookupRule(point);
}
long duration = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
System.out.println("Took: " + duration + "ms");
```

I ran the same test against the array lookup but had to limit the number of countries in order for them to fit into the index. With the first 100 countries the same benchmark takes about 0.5 seconds but i noticed that building the index is rather slow with 50 seconds. The JTS approach builds the complete world index with 255 countries in about 30ms.

I than reran the benchmark using the same 100 countries and the JTS index finished in around 0.9 seconds with an index build time of around 20ms.

Considering the startup costs the array lookup would need to process about 80 million points for the startup delay to amortize.

otbutz:
> SpatialRuleLookupBuilderTest.testSpeed:246 » OutOfMemory Java heap space

What is the limit for the test builds?

karussell:
We limit it explicitly to 100MB to detect memory issues very early: https://github.com/graphhopper/graphhopper/blob/master/pom.xml#L118. How much more would be required?

otbutz:
@karussell i've lowered the number of points for the test from 5 to 1 million.

boldtrn:
> With the first 100 countries the same benchmark takes about 0.5 seconds but i noticed that building the index is rather slow with 50 seconds.

Yes, that's somewhat expected. The lookup is currently optimized for look up performance. For me building the index was not that time critical. This is only done during the import, where time (adding a few seconds to minutes) plays only a minor role, but we might need to lookup > 200.000.000 edges for world wide imports. Also if we ever want to do the lookup during the route calculation.

Thanks for providing some benchmark data and this nice PR :+1:

otbutz:
> but we might need to lookup > 200.000.000 edges for world wide imports. 

The JTS approach should be fast enough to be used during the import. The benchmark is a bit pessimistic as i measured it with one rule for every country on the planet which results in an index with ~2.6k entries.

> Also if we ever want to do the lookup during the route calculation.

There is still room for improvement. The hit-rate of the spatial index seems to be a good candidate for optimization. Just take a look at the bounding box of Austria which covers half of Bavaria:

Screenshot_2020-01-08 Bounding Box Tool Metadata Enrichment for Catalogue Records by Visually Selecting Geographic Coordina

Each point in the southern region of Bavaria needs to be tested against both polygons as the fast index/envelope shortcut doesn't work here.

We could split bigger polygons in order to keep their Envelope small and therefor closer to the actual borders.

otbutz:
Did a bit of profiling using VisualVM while recording roughly 2 minutes of execution time:

## Whole world index (2.6k polygons)
Screenshot_20200108_172850

## Central europe index (54 polygons)
Screenshot_20200108_182351

## Benchmark results

|   | Whole world  | Europe |
| ------------- | ------------- | ------------- |
| 1 million points  | 1.5s  | 1s  |
| 5 million points  | 6.2s  | 4.5s  |

Seems like the size of the index isn't that crucial which means that we would need to lower the amount unnecessary PIP checks in order to speed things up. Note that the number of invocations of `SpatialRuleContainer.containsProperly(Point)` is roughly twice the amount of rule lookups.

otbutz:
With the last change the lookup of 5 million points is down to 3.5s

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleContainer.java
```diff
@@ -19,57 +19,99 @@
...
 /**
- * This class contains a collection of SpatialRule and is used for the implementation SpatialRuleLookupArray.
+ * This class contains a collection of SpatialRule which are valid for a certain Polygon.
  *
  * @author Robin Boldt
+ * @author Thomas Butz
  */
 class SpatialRuleContainer {
+    private static final PreparedGeometryFactory PREP_GEOM_FACTORY = new PreparedGeometryFactory();
+    private static final GeometryFactory FAC = new GeometryFactory();
```
karussell: I'm currently investigating this to fix the block_area feature and maybe join or use some functionality from here. 

Some questions:

Are GeometryFactory and PreparedGeometryFactory thread safe?

I found out about PackedCoordinateSequence which could lead to less memory usage for shapes with many points: 
`new GeometryFactory().createPolygon(new PackedCoordinateSequence.Double(coordinates, 2))`
Or is this not under our control as this is done from [our jackson-datatype-jts fork](https://github.com/graphhopper/jackson-datatype-jts/blob/master/src/main/java/com/bedatadriven/jackson/datatype/jts/parsers/PolygonParser.java)? Maybe recreating makes sense?

otbutz: > Are GeometryFactory and PreparedGeometryFactory thread safe?

The GeometryFactory should be thread-safe as both of its fields don't have any mutability. (check [CoordinateArraySequenceFactory](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequenceFactory.java) and [PrecisionModel](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/PrecisionModel.java))
But i'm not 100% sure that this is the case for different sequence factories.

Javadoc of `PreparedGeometry`:
>  Subclasses are intended to be thread-safe, to allow PreparedGeometry to be used in a multi-threaded context  (which allows extracting maximum benefit from the prepared state).

> I found out about PackedCoordinateSequence which could lead to less memory usage for shapes with many points

Interesting. The PreparedGeometry keeps the original Geometry so this could lead to less memory consumption.

> Or is this not under our control as this is done from our jackson-datatype-jts fork? Maybe recreating makes sense?

Correct. We'd have to recreate them but that should pay off quickly.

otbutz: Maybe my test is a bit flawed or it really starts to be more efficient for a high number of points but i wasn't able to see any benefit when using for our current country borders. 

```java
System.gc();
System.gc();
System.gc();
        
System.out.println("Before: " + (Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory()));

SpatialRuleLookup spatialRuleLookup = SpatialRuleLookupBuilder.buildIndex(Collections.singletonList(Jackson.newObjectMapper().readValue(reader, JsonFeatureCollection.class)), "ISO_A3", rulePerCountryFactory);

System.gc();
System.gc();
System.gc();

System.out.println("After: " + (Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory()));
```
The memory usage went up by about 1MB from 22Mb to 23MB.

karussell: > The GeometryFactory should be thread-safe as both of its fields don't have any mutability.  

As I didn't find any official statement about this I created https://github.com/locationtech/jts/issues/512

> i wasn't able to see any benefit when using it for our current country borders.

Oh, interesting. And yeah, probably because too few points (as we use the simplified shape)

otbutz: I can switch to the packed version at it seems to have no effect on the execution time and the difference in memory usage can safely be ignored.

karussell: Ok, thanks!

BTW: I created a new branch: https://github.com/graphhopper/graphhopper/compare/replace_polygon that shows how we could have a better JTS Polygon integration (i.e. our Polygon is implemented via JTS). Replacing all shapes is too hairy for now, but still the goal...

otbutz:
@karussell anything missing?

karussell:
As this is based on #1827 should we close #1827 when merging here? Or should we merge #1827 first?

otbutz:
We can close #1827 when we merge this branch. I just thought it would be better to break my changes into smaller units.

boldtrn:
Since we change the config value to: `spatial_rules.borders_directory`, maybe we should add this to the changelog.txt? Otherwise people might wonder if SpatialRules are suddenly not loaded anymore?

otbutz:
> Since we change the config value to: `spatial_rules.borders_directory`, maybe we should add this to the changelog.txt? Otherwise people might wonder if SpatialRules are suddenly not loaded anymore?

Thanks! Completely forgot to cherry-pick this commit: https://github.com/graphhopper/graphhopper/pull/1827/commits/9708e45edf1fc6779744ace5cc006e38dcac163d

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
```diff
@@ -60,11 +72,14 @@ public void testSmallScenario() {
     @Test
     public void testExact() {
```
boldtrn: With the SpatialRuleLookupArray there was a parameter to switch to the exact lookup. The SpatialRuleLookupJTS is exact by default. Maybe it would make sense to rename the tests?

otbutz: Maybe something like `testPrecision` ?

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
```diff
@@ -130,20 +147,45 @@ public void testExactAdjacentBorder() {
...
-        spatialRules.add(getSpatialRule(new Polygon(new double[]{-100, -100, 100, 100}, new double[]{-100, 100, 100, -100}), "big"));
-        SpatialRuleLookup spatialRuleLookup = new SpatialRuleLookupArray(spatialRules, 1, true, new BBox(1, 2, 1, 2));
+        Polygon big = FAC.createPolygon(new Coordinate[] { new Coordinate(-100, -100), new Coordinate(100, -100), new Coordinate(100, 100), new Coordinate(-100, 100), new Coordinate(-100, -100)});
+        spatialRules.add(getSpatialRule(big, "big"));
+        SpatialRuleLookup spatialRuleLookup = new SpatialRuleLookupJTS(spatialRules, new Envelope(1, 2, 1, 2));
         assertEquals("big", spatialRuleLookup.lookupRule(1.5, 1.5).getId());
     }
+    
+    @Test
+    public void testHole() {
```
boldtrn: It's great that you added this test!

I was just wondering, why is the hole the same as the shell?

Wouldn't it make more sense to have an actual polygon with a hole and a border like:

```
LinearRing shell = FAC.createLinearRing(new Coordinate[] { new Coordinate(1, 1), new Coordinate(7, 1), new Coordinate(7, 7), new Coordinate(1, 7), new Coordinate(1, 1)});
LinearRing hole = FAC.createLinearRing(new Coordinate[] { new Coordinate(2, 2), new Coordinate(6, 2), new Coordinate(6, 6), new Coordinate(2, 6), new Coordinate(2, 2)});
```

And then we could test like:

```
assertEquals("SpatialRule.EMPTY", spatialRuleLookup.lookupRule(5, 3).getId());
assertEquals("1", spatialRuleLookup.lookupRule(1.5, 1.5).getId());
```

Then I think we could drop the p2 part of the test?

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
boldtrn: Yes sure, or maybe something like testCloseToBorder etc.

I am not sure we need to change it at all, but just thought it might be worth to consider.

Also the test was designed to catch some edge cases of the exact parameter, I am not sure we need it at all for the JTS implementation? Also there is [a comment](https://github.com/graphhopper/graphhopper/pull/1829/files/ebbb38bfe281a13e2a961eddb0ea6fb1438afb38#diff-f751d631a68af77814ca82af419acd9bL74) about the border tile that existed in the array lookup, but probably not with JTS.

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
otbutz: This was the sole reason that the test passed. After i switched it to the intended hole structure the test failed because lookupRule() wants lat/lon while all JTS code is lon/lat....

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
otbutz: I think one test should be enough to enforce precision.

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
boldtrn: Sorry, I am not sure if I understand your reasoning here, could you try to explain this a bit more? My idea was that it's probably rather uncommon to have a polygon that is a complete hole?

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
boldtrn: I don't mean to add more tests. I was thinking of either a) removing this test, as it was designed to catch the edge cases involved with a tile based lookup with a precision of 1 or b) rewriting it a bit for example to test coordinates that are closer inside/outside of the polygon, so it's a precision test.

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
otbutz: The hole was not intended to cover the whole polygon. I think that i messed up a rebase at some point. After fixing it i discovered that the test doesn't pass anymore because i mixed up lat/lon in the lookup. Long story short: two bugs squashed :smile:

boldtrn: Ah I see, thanks for the explanation :)

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTS.java
```diff
@@ -0,0 +1,140 @@
...
+                applicableRules.addAll(container.getRules());
+            }
+        }
+        
+        if (applicableRules.isEmpty()) {
+            return SpatialRule.EMPTY;
+        }
+
+        //TODO support multiple rules
+        return applicableRules.get(0);
```
boldtrn: Since this would work out of the box now and since we completely change the implementation, wouldn't it make sense to break the interface? Or maybe deprecate the lookupRule method and create a new lookupRules method. lookupRule could fall back to returning only the first result.

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
otbutz: Sidenote: We should probably settle on one representation for coordinate values because GH code seems to prefer lat/lon while JTS is using lon/lat.

otbutz:
@boldtrn i've addressed the issues you mentioned and rebased on the current master

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
boldtrn: Probably it would be better to stay with lat,lon in order to keep the existing methods/interfaces working as they are? Since this would only switch two doubles, it would be hard for external parties to see the change when merging.

boldtrn:
> i've addressed the issues you mentioned

Thanks a lot. I added a lost minute note, sorry :).

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTS.java
otbutz: The interfaces still need a bit of love but that will be the focus of the following PRs. The next step is going to be adding support for multiple overlapping rules with priorities.

#1808 describes the roadmap

core/src/test/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTSTest.java
otbutz: Yeah, you're right. I think it's fine as long as we keep the contact area between these two coordinate representations low.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTS.java
boldtrn: Ok sure, that makes sense :)

boldtrn:
Thanks for the nice improvement @otbutz. I just merged this PR.

karussell:
Thanks @otbutz! And @boldtrn for the review.

karussell:
On [appveyor we get](https://ci.appveyor.com/project/graphhopper/graphhopper/builds/30434413): 

SpatialRuleLookupBuilderTest.testSpeed:256 Benchmark must be finished in less than 10000ms

Can we roughly half the complexity of that speed test to make it pass also on slightly weaker servers?

otbutz:
> On [appveyor we get](https://ci.appveyor.com/project/graphhopper/graphhopper/builds/30434413):
> 
> SpatialRuleLookupBuilderTest.testSpeed:256 Benchmark must be finished in less than 10000ms
> 
> Can we roughly half the complexity of that speed test to make it pass also on slightly weaker servers?

I can also remove the assert and keep it only for debugging

karussell:
But then probably removing the entire test would make more sense. We could move it into the Measurement class instead?

otbutz:
I wasn't aware of that class. Seems like a good fit.

michaz:
Excellent, thanks! I was wondering, @otbutz, how much performance do we gain by the tiling optimization in ``SpatialRuleContainer``? Is there a tradeoff, do you think there could be Polygons where performance gets worse?

otbutz:
> I was wondering, @otbutz, how much performance do we gain by the tiling optimization in SpatialRuleContainer?

About 10-15% as far as i can remember but don't pin me down on this.

> Is there a tradeoff, do you think there could be Polygons where performance gets worse?

I tested a few optimizations to see if i can shove off a bit of execution time by using simpler checks to avoid the whole point-in-polygon logic wherever possible. Ideally you would want to find the biggest envelope which can fit entirely into the polygon without intersecting/covering any holes.

As the country polygons are rather complex, finding this ideal inner envelope is quite hard(NP-complete?) and would require some kind of optimization algorithm. Breaking the polygon up into tiles is rather quick and servers as a good approximation.

A small grid won't yield any benefits as most tiles will intersect with the border. If the grid is too big, the costs of the tile checks outweigh the regular point-in-polygon costs.

I don't think that the tiling approach will cause problems for certain polygons. Either we have a high hitrate which allows us to dodge most of the pip logic or we don't have a lot/any tiles which eliminates the envelope checks.

PR #1833, opened 2019-12-24
Add Motorroad RoadClass

otbutz:
Uses the same logic as the CarFlagEncoder to determine if a highway should be treated as a motorroad and uses a dedicated RoadClass constant to encode it.

### DIFF
--- a/config-example.yml
+++ b/config-example.yml
@@ -15,7 +15,7 @@ graphhopper:
 
   # Add additional information to every edge. Used for path details.
   # If road_environment is added and elevation is enabled then also a tunnel and bridge interpolation is done, see #798.
-  # More options are: surface,max_width,max_height,max_weight,max_axle_load,max_length,hazmat,hazmat_tunnel,hazmat_water,toll,track_type
+  # More options are: surface,max_width,max_height,max_weight,max_axle_load,max_length,hazmat,hazmat_tunnel,hazmat_water,toll,track_type,motorroad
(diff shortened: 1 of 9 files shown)


karussell:
We'll likely not include this as motorroad is just a `trunk` with different access values. We had this discussion in the initial PR: https://github.com/graphhopper/graphhopper/pull/1548#issuecomment-499165188

boldtrn:
How about we add motorroad as its own boolean encoded value? There are some primaries which are motorroads and internationally there are a lot of trunks that are no motorroad. That way everyone who needs this value can set it up individually.

This also makes sense when looking at OSM. In OSM motorrad is a different tag than highway and I think we should not mix the two :).

karussell:
I see no real benefit of this tag from the routing perspective if we can use `trunk` and some additional access restrictions. If this cannot be (easily) replicated we can think of including it via a boolean, sure.

otbutz:
At least the CarFlagEncoder seems to use it to enforce a different default speed.

boldtrn:
I don't know if we need it for routing, but I could think of examples where this can be relevant. If you think about flexible routing. Some vehicles for example are not allowed on motorroads, but are allowed on trunk roads, that are no motorroad.

Here is an example:
https://www.openstreetmap.org/way/202170525#map=17/43.47096/-5.14263
https://goo.gl/maps/ij1t2yi2Zzf8x8X99

In OSM, this is a trunk. This is also tagged correctly. But if look at it on Google Streetview, it is quite clear that it's not what I would consider a trunk.

Issue #1835, opened 2019-12-29
Move access flag checks into Weighting


easbar:
As part of #1776 we need to stop using the access encoded values check if a road is accessible for a given vehicle and instead purely rely on the edge weight calculated by the `Weighting`. Otherwise returning a finite weight in `Weighting` might have no effect in case the access flag of the corresponding road/direction is set to false. 
This is not a trivial task especially with performance in mind.

Just to make sure: Do we really need to do this for #1776 already ? What if we had something like a flexible encoder that simply uses fwd/bwd=true for all roads, while some of the 'classic' vehicle profiles like car or bike continue to use the access flags (but cannot be used with flexible/configurable weighting). In this case using fwd/bwd=false and at the same time returning a finite weight would be considered a bug, and to offer the configurable weighting option the encoder must set all access flags to true.

karussell:
> This is not a trivial task especially with performance in mind.

Why do you think this is complicated because of performance?

> What if we had something like a flexible encoder that simply uses fwd/bwd=true for all roads

That would work, but would be ugly as we would check the access in the EdgeFilter and then again in the configurable Weighting.

> while some of the 'classic' vehicle profiles like car or bike continue to use the access flags

Yes, we can use them as "base" and then, once a road is inaccessible for these "base vehicles", you cannot enable these roads via the Weighting. Actually I did this in the various branches when I implemented a preliminary version of it and the experience was okay.

The remaining problem is still issue #729 where we need to replace the EdgeFilter with the Weighting, otherwise it can happen that the LocationIndex returns edges that are blocked due to the configurable Weighting.

BTW: should we rename this issue to "make access flags optional"? As this is not about removing the access flags and more about moving the access flags check into the Weighting. I.e. we can still implement access restrictions via pre-calculated bits in the edge flags, but also calculate it dynamically via some EncodedValues.

easbar:
> Why do you think this is complicated because of performance?

Ok, as long as you have some access flags so you can check them within weighting quickly the performance should in principle stay the same. But do you not have to set (at least most of) the access flags to true to be able to have a real 'configurable' weighting ? Edges with access=false cannot be unblocked by a configurable weighting in my current understanding (or you ignore them but then there is no quick exit and hence performance will decrease).

What I think is 'not trivial' is figuring out what the problems are and finding a good solution for all things we would like to achieve (performance & flexibility mostly). For example it took me quite some time to find a solution to the access flags issue just for the CH preparation, #1837.

[...]

Issue #1708, opened 2019-08-30
Select CH profiles by name ?


easbar:
Continuing some of the discussion that started in #1705.

> But can't this be simplified enormously if we just give every profile that a GraphHopper instance supports (not only every CHProfile; every profile) a name in the configuration file?

But isn't this the case already ? For example we have `graph.flag_encoder=foot,car|turn_costs=true`. Not thinking about CH for a moment this means we can do routing for foot and car and for car we also get a turn cost extension (i.e. we can route with turn restrictions from OSM). Also we can use the different query parameters, like heading, alternative etc. And you mean all the different combinations should be listed upfront ? 

For CH we have to list this upfront because the preparation needs to be done prior to the routing. One thing that def. can be improved is how these preparations are specified, #493. The other question is then how the request parameters are mapped to the preparations afterwards. 

Say somebody sets up a GH server for a new app that has two CH profiles for two different user groups: one group moves around a city using bicycles preferring roads with bicycle lanes and another uses trucks and can only drive to certain parts of the city. I imagine the best to do would be to have a nice way to specify all the parameters that go into the CH preparation like the road weights, the blocked areas etc. (making this easy to configure is #493), then give this profile some (arbitrary) name, and afterwards select the profile using this name (maybe the app that is used knows the user group via login). So here it would be useful to give a profile a name/id and use this on the request.

Or thinking about the GH directions API: It is running for some fixed set of profiles. It would be easy to document what these profiles are and give them an id/name and then choose the profile by name on the request. This seems a lot more intuitive than having to specify the weighting, encoder, u-turn-costs and all this just so the request 'matches' one of the profiles (and in case it does not getting an error / or a slow algorithm if falling back to non-CH). Or even better add an endpoint to the server that lists all the available profiles.

Even for flexible CH it would be similar I think. Say there is the possibility to build a CH preparation 'on the fly'/while the server is running (which I very much hope it will be at some point). This will certainly take a lot of memory and the preparation needs to be stored or kept in memory etc. To use it later it has to have some name and to me its also the most obvious way to use nothing but this name to refer to this preparation.

Right now I cannot come up with a case where it would not be feasible to 1) lookup the documentation/special endpoint etc. of whatever GH server is running, 2) check what profiles are supported and find the name of the one that should be used, 3) run a route request with the name of the desired profile attached ?

easbar:
What kind of request parameter should be used to select the profile: From #1838: 

[...]

PR #1838, opened 2019-12-30
location lookup based on Weighting additionally to access

### LINKED ISSUE #729: Improve edge filtering for LocationIndex

karussell:
fixes #729 before #1835

TODOs:

 - [x]  throw error if area contains one of the query points


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1042,27 +1042,29 @@ public GHResponse route(GHRequest request) {
             // For example see #734
             checkIfPointsAreInBounds(points);
 
+            // TODO NOW for CH it will be overwritten by chProfile.getWeighting but now we need it for locationIndex -> #729
+            Weighting weighting = createWeighting(hints, encoder);
+
```
easbar: I think its fine to create the `Weighting` early here, but should we not better take it from the `CHProfile`, something like

```
- request comes in
- if LM/CH: select preparation using request parameters and take Weighting from the preparation/profile
- else: create Weighting from request parameters
```

instead of

```
- request comes in
- create Weighting from request parameters (hopefully we managed to re-create the same Weighting that was used for the preparation)
- later: use request parameters to select CHProfile to select CHGraph etc. and overwrite Weighting (?)
```

Two questions here:

1) Should we support using another CH/LM Weighting for the request than it was used for the preparation ? Normally this is an error, but their might be some (rather special) use-cases for this (@michaz ?). Otherwise I think the first method is better because it automatically uses the right Weighting (or at least the one that is kept by the preparation).

2) How are we supposed to select the CHProfile/Weighting given the request ? I think as soon as we want to use 'real' configurable weightings for CH we need to select the profiles by name (#1708) and if we do we need to first select the CHProfile by name and then take the weighting from the profile (first method). Otherwise we would have to send the whole configuration with every request (makes sense for non-speedup algos, because thats the whole point of configurable weightings, but not so much for LM/CH where the idea is rather to create a (custom/configurable) preparation once and then use it for many or at least multiple requests.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1042,8 +1042,19 @@ public GHResponse route(GHRequest request) {
             // For example see #734
             checkIfPointsAreInBounds(points);
 
-            // TODO NOW for CH it will be overwritten by chProfile.getWeighting but now we need it for locationIndex -> #729
-            Weighting weighting = createWeighting(hints, encoder);
+            RoutingAlgorithmFactory algoFactory = getAlgorithmFactory(hints);
+            Weighting weighting;
+            CHProfile chProfile = null;
+            if (chFactoryDecorator.isEnabled() && !disableCH) {
+                if (algoFactory instanceof CHRoutingAlgorithmFactory) {
```
karussell: I removed the feature that LM can be combined with CH as we currently do not use it (and it is not faster for node-based). Or should I include it as it is a bit worth for edge-based?

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: Yes, thanks! Can you have a look into the new commit if this is what you meant?

> Should we support using another CH/LM Weighting for the request than it was used for the preparation ? 

For LM this works under special conditions (that increase the edge weight). For CH this won't work at the moment but it could be required for "Core CH"?

> I think as soon as we want to use 'real' configurable weightings for CH we need to select the profiles by name (#1708)

I think this should be part of a different issue #1776 and the configuration (with which the CH preparation is done) could be picked up via the `vehicle` and/or `weighting` parameters.

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: Have you tried it with edge-based ? Currently we are using a straight-line distance heuristic for edge-based CH AStar, not sure if its an improvement to use LM in this case. But if we never used it so far, sure why not remove it. I like this because it reduces some of the factory/decorator complexity.

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: > Yes, thanks! Can you have a look into the new commit if this is what you meant?

Yes this is perfect!

> For LM this works under special conditions (that increase the edge weight). For CH this won't work at the moment but it could be required for "Core CH"?

Yes but do we have to support this via the GraphHopper/high-level API here ? Do you (still) need this for the block-core thing @michaz ? Or maybe we can keep going here and add this functionality later (so far this is not really supported either) ?

> I think this should be part of a different issue #1776 

Do you mean it should be part of #1776 or it should not ? Sure, it should not be part of *this* (#1838) issue.

> and the configuration (with which the CH preparation is done) could be picked up via the vehicle and/or weighting parameters (?)

This is a good/tough question, I'll move this question to #1708.

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: Hmmh. Actually the decorator stuff was only introduced due to this use case (LM+CH), so will probably not remove it here and a separate refactoring is better.

karussell: Hmmh, yes. Actually the decorator stuff was only introduced due to this use case (LM+CH) if I correctly remember, so will probably not remove it here and a separate refactoring is better.

core/src/main/java/com/graphhopper/GraphHopper.java
michaz: > Should we support using another CH/LM Weighting for the request than it was used for the preparation

Depends on what 'support' means. I need to be able to _write code_ where I pick the Graph I want to use and the Weighting I want to use by myself. So there shouldn't be machinery which _actively prevents_ me from doing so. For example, nothing should try to information-hide a CH-Graph away from me until I can show it the proper weighting instance.

michaz: But I don't need to be able to have this as, for example, a config file option.

easbar: Ok I'll try to keep in mind that it should be easy enough to use another weighting together with a partially contracted graph.

easbar:
Looks good to me, except the remaining todo:

>  throw error if area contains one of the query points

Currently `createWeighting` is used for two different purposes: to create a weighting for preparation and to create a weighting at request time. For `initCH/LMAlgoFactoryDecorator()` the block area coordinates are not an issue, because there are no query points, but we need to run this check in `GraphHopper#calcPaths` and for the SPT&Isochrone endpoints. The easiest seems to be doing this check before calling `createWeighting` (wherever we have access to the requested points) with the only downside that one has to remember doing so. Basically it is about request validation, which already is partially duplicated between /route, /spt and /isochrone, so maybe its worth creating an extra method `GraphHopper#validateRequest` or similar that does the block area check and can be called from /route, /spt and /isochrone ?

karussell:
> so maybe its worth creating an extra method GraphHopper#validateRequest or similar that does the block area check and can be called from /route, /spt and /isochrone ?

This is a good idea. Especially since this is already wanted from @boldtrn (see TODO comment before checkIfPointsAreInBounds call).

Should we do this in another issue (for 1.0) and merge here for now?

easbar:
Ok. The only problem is that if we specify some query points with in the block area the virtual edges will not be blocked ? But is this even such a big deal ? In most cases they won't be reachable anyway, because all the edges leading to the point lying in the blocked area will be blocked as well (?)

karussell:
> The only problem is that if we specify some query points within the block area the virtual edges will not be blocked ?

Yes, but with a validation that ensures that query points cannot be inside the blocked areas this won't happen IMO. Which also makes sense as you cannot calculate a route where you block one of the waypoints.

Or ... if you mean this shouldn't go into a separate issue, I'll include the check here (introduce your proposed new method).

karussell:
> cannot be inside the blocked areas this won't happen IMO

Ok, there are cases where waypoints are outside of the block area but some involved virtual edges can still be inside. But when I went through the possibilities I did not find a case where the involved and blocked tower node did not manage the blocking.

Ie. and edge `A-x-B` with query on x where `A-x` is blocked is already blocked because of non-virtual edges for A are blocked. 

Okay there are rare cases where the virtual edge is inside the blocked area but not A and also not x ... hmmh

easbar:
Ah I meant its not a big problem and we can merge and do the validation later. Yes once we do the validation the problem I mentioned does not occur anymore. Ok, like you said the query point might be outside of the blocked area and the validation would not throw an error. But is this really a problem ? It just means the edge won't be blocked, but who wants to block edges that are leading / coming from the query points (the places you want to go / start from) anyway ?

easbar:
So I think we can:

- merge now
- add the validation method later where we only do a 'rough' consistency check to warn the user that the query point lies within the blocked area, which is a contradiction

easbar:
Ah I did not think of the case where a single **point** is blocked (how does that work ?). So far I was thinking of blocked **areas**.

karussell:
> but who wants to block edges that are leading / coming from the query points (the places you want to go / start from) anyway ?

The use case could be to block only parts of the road and still make the rest of the road accessible and force a u-turn in those cases. But those issues should be probably fixed within #1324 as currently it does not work anyway. BTW: See [this discussion](https://discuss.graphhopper.com/t/3680).

> Ah I did not think of the case where a single point is blocked (how does that work ?). So far I was thinking of blocked areas.

It is a LocationIndex lookup, i.e. not really a singular point: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java#L61

karussell:
I will do the left TODO, but afterwards have to fix #1127 and #1324 via a cleaner block_area implementation e.g. replace our Polygon with the JTS Polygon implementation etc and also look into the work from @otbutz to maybe use the polygon work also on a per-request basis not just for import.

otbutz:
JTS doesn't really support circle as a geometry AFAIK. So we would have to remove this type of blocking functionality or use a polygon as an approximation:

https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/util/GeometricShapeFactory.java#L199-L234

easbar:
This looks good! And even better now that block area is no longer in `createWeighting`.

PR #1839, opened 2019-12-30
Filter outside

karussell:


easbar:
Did you want to merge this or just create the PR as kind of a bookmark/reminder ?

karussell:
I wanted to comment something and to have something to discuss about, but then understood the questionable part. But we cannot delete PRs or issues I think ...

PR #1842, opened 2019-12-30
Work towards removing TurnWeighting

### LINKED ISSUE #1820: Remove TurnWeighting
This has already been discussed in https://github.com/graphhopper/graphhopper/issues/1520, and is part of #1776.

easbar:
This is some preliminary work for #1820, mainly to reduce the diff when we actually fix #1820.

1) (trivial) move the `INFINITE_U_TURN_COSTS` constant into `Weighting`. 
2) add `Weighting#calcEdgeWeight/Millis`. For now this is a pure convenience method that just calls `calcWeight` with `prevOrNext=-1`. For #1820 this will be the other way around: In `AbstractWeighting` we will have `calcWeight` which uses `calcEdgeWeight` and adds `calcTurnWeight`. Subclasses will then only override `calcEdgeWeight/Millis` and leave the turn cost calculation to `AbstractWeighting`. The turn cost calculation can then be adjusted by changing the `TurnCostProvider` implementation which is passed to `AbstractWeighting` via its constructor (at least this is what I currently think we should do).
3) replace `FlagEncoder#supports(TurnWeighting.class)` with `FlagEncoder#supportsTurnCosts()`. Note that this it what it was like before 0.4 according to the changelog

The actual work for #1820 is happening here: https://github.com/graphhopper/graphhopper/tree/remove_turn_weighting, but its not in (near) a mergeable state yet.

Any thoughts ?

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/weighting/Weighting.java
+++ b/core/src/main/java/com/graphhopper/routing/weighting/Weighting.java
@@ -39,30 +39,40 @@ public interface Weighting {
      */
     double getMinWeight(double distance);
 
-    double calcEdgeWeight(EdgeIteratorState edgeState, boolean reverse);
-
     /**
-     * This method calculates the weighting a certain edgeState should be associated. E.g. a high
-     * value indicates that the edge should be avoided. Make sure that this method is very fast and
-     * optimized as this is called potentially millions of times for one route or a lot more for
-     * nearly any preprocessing phase.
+     * This method calculates the weight of a given {@link EdgeIteratorState}. E.g. a high value indicates that the edge
+     * should be avoided during shortest path search. Make sure that this method is very fast and optimized as this is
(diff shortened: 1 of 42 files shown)


easbar:
Can we merge this ?

core/src/main/java/com/graphhopper/routing/weighting/Weighting.java
```diff
@@ -37,6 +39,8 @@
      */
     double getMinWeight(double distance);
 
+    double calcEdgeWeight(EdgeIteratorState edgeState, boolean reverse);
```
karussell: Can you add javadocs? And highlight the difference to calcWeight?

easbar: Sure, good point!

Issue #1843, opened 2019-12-31
Issue RoadClassEncoder via EncodedValueLookup when using mtb and motorway


boldtrn:
When I set multiple vehicles in the config, including mtb, for example: 
```
graph.flag_encoders: car, mtb
```

My config looks like this:
```
graph.encoded_values: road_class,road_class_link,road_environment,max_speed,road_access
graph.bytes_for_flags: 8
```

If I lookup an encoded value in `InstructionsFromEdges` via `evLookup.getEnumEncodedValue(RoadClass.KEY, RoadClass.class);` I get an EncodedValue as expected.

For motorways this seems returns "other" instead of "motorway". Other road classes seem to be fine. If I only use car it works fine. 

I cross tested with `graph.flag_encoders: car, bike` and this particular issue did not exist. I don't think it's necessarily related to the EncodedValueLookup, but that's how I discovered this.

An example route where this happens is this (I think it happens for all motorways): http://localhost:8989/maps/?point=49.303926%2C8.627148&point=49.302688%2C8.626859

Maybe something is wrong with the number of bits reserved by the mtb encoder?

karussell:
Uh, this looks ugly. Thanks for the report.

btw: this is no longer necessary: `graph.bytes_for_flags: 8`

boldtrn:
Thanks for the fast fix :tada:. 

> btw: this is no longer necessary: graph.bytes_for_flags: 8

Thanks, I missed that change!

Issue #1047, opened 2017-04-29
Instructions: Remove Unnecessary Continues


boldtrn:
Currently we have unnecessary continues in the instructions, here are some samples:

https://graphhopper.com/maps/?point=48.004367%2C8.137822&point=48.003878%2C8.137844
https://graphhopper.com/maps/?point=48.177999%2C8.572013&point=48.17788%2C8.573037

Currently, we are creating an instruction if we are not sure if an instruction is needed at this point. With the current data it's very complicated to improve the current situation. With highway information we should be able to improve it.

If we are continuing on the same highway class but the outgoing edge is a different highway class, then it is very likely that the turn instruction is not necessary.

Related to #990

boldtrn:
Should be fixed with #1844

karussell:
#1844 has improved the situation a lot but unfortunately the two examples above are not yet fixed. Also [I found an instruction](https://graphhopper.com/maps/?point=51.367105%2C14.491246&point=51.369048%2C14.483092) which is now removed but important because "follow the road" is a "right turn".

boldtrn:
Thanks for the pointer. I will have a look and see if we can improve the situation :+1:

boldtrn:
> the two examples above are not yet fixed

If you disable turn_costs they are fixed, I just created an issue for this #1883

karussell:
Ah, ups. This is indeed also a deployment problem (see #1883). Sorry for the bubble.

Issue #990, opened 2017-03-18
Use highway tag and actual edge speed for Instructions


boldtrn:
The DataFlagEncoder stores the highway type. The highway type allows us to create fancy instructions like: "leave the highway" or "enter the highway". Since the DFE works a bit different considering storing the speed of edges, we should find a way to calculate the actual speed of the edges for the Instructions calculation.

Also see #955 for some more discussion.

Updated TODOs:

 - [ ] InstructionsFromEdges => should rely on max_speed instead of accessEnc
 - [ ] remove FlagEncoder from InstructionsOutgoingEdges
 - [x] any encoder could be used (we even could remove DataFlagEncoder & GenericWeighting)
 - [x] how to replace encoder.getAnnotation(flags, tr) !?

karussell:
The solution is to use the weighting and not the encoder for calculating the time for an edge IMO (?)

boldtrn:
> The solution is to use the weighting and not the encoder for calculating the time for an edge IMO (?)

For the time, yes. The idea here was to check the speed of the edge. It might be that we don't need this check if we have highway tags.

Consider you are driving on primary and you are in a left curve. Now there is a small road leaving straight (this happens very frequent). The heuristic is, that if the leaving road is 2*slower, and if the left bending road has the same speed as the current edge, then we are probably following the street and are not leaving the road, therefore we don't need an instruction.

We might be able to do the same via highway tags, so maybe speed is not needed here.

devemux86:
> We might be able to do the same via highway tags, so maybe speed is not needed here.

Optimal instructions calculation is usually performed via road types and names, so speed indeed seems superfluous here.

boldtrn:
> so speed indeed seems superfluous here.

Yes, you are right. For encoders other than the DataFlagEncoder we don't have the road type. That's why I came up with the speed heuristic.

boldtrn:
Should be fixed with #1844 :).

PR #1844, opened 2019-12-31
Use road class to determine if a fork should have an instruction

boldtrn:
This PR solves #990 and #1047. 

Essentially, this will remove many unnecessary instructions on main roads, including the annoying continue instructions on motorways. For minor roads the difference should be rather small.

Something I am not sure about is how we should handle the tagged maxspeed. Often no maxspeed is tagged. Currently I decided to fallback to the average speed.

Let me know how this works out in your tests and if there are any cases we need to improve.

Here is a brief comparison of [GH maps route](https://graphhopper.com/maps/?point=49.498458%2C8.464966&point=47.657988%2C9.508667&locale=de-DE&vehicle=car&weighting=fastest&elevation=true&debug=true&use_miles=false&layer=OpenStreetMap) vs. the improved version, I think the difference is quite visible :)):
Screenshot from 2019-12-31 16-14-30
Screenshot from 2019-12-31 16-14-36

BTW: I stumbled over #1843 while creating this PR

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
+++ b/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
@@ -50,11 +50,13 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
     private final BooleanEncodedValue roundaboutEnc;
     private final BooleanEncodedValue accessEnc;
     private final BooleanEncodedValue getOffBikeEnc;
+    private final BooleanEncodedValue roadClassLinkEnc;
     private final EnumEncodedValue<RouteNetwork> bikeRouteEnc;
     private final EnumEncodedValue<RoadClass> roadClassEnc;
     private final EnumEncodedValue<RoadEnvironment> roadEnvEnc;
     private final EnumEncodedValue<RoadAccess> roadAccessEnc;
     private final EnumEncodedValue<Toll> tollEnc;
+    private final DecimalEncodedValue maxSpeedEnc;
 
     /*
      * We need three points to make directions
@@ -106,6 +108,8 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
         this.tollEnc = evLookup.hasEncodedValue(Toll.KEY) ? evLookup.getEnumEncodedValue(Toll.KEY, Toll.class) : null;
 
         this.roadClassEnc = evLookup.getEnumEncodedValue(RoadClass.KEY, RoadClass.class);
+        this.roadClassLinkEnc = evLookup.getBooleanEncodedValue(RoadClassLink.KEY);
+        this.maxSpeedEnc = evLookup.getDecimalEncodedValue(MaxSpeed.KEY);
         this.roadEnvEnc = evLookup.getEnumEncodedValue(RoadEnvironment.KEY, RoadEnvironment.class);
         this.roadAccessEnc = evLookup.getEnumEncodedValue(RoadAccess.KEY, RoadAccess.class);
         this.nodeAccess = graph.getNodeAccess();
@@ -390,7 +394,7 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
             forceInstruction = true;
         }
 
-        InstructionsOutgoingEdges outgoingEdges = new InstructionsOutgoingEdges(prevEdge, edge, encoder, crossingExplorer, nodeAccess, prevNode, baseNode, adjNode);
+        InstructionsOutgoingEdges outgoingEdges = new InstructionsOutgoingEdges(prevEdge, edge, encoder, maxSpeedEnc, crossingExplorer, nodeAccess, prevNode, baseNode, adjNode);
         int nrOfPossibleTurns = outgoingEdges.nrOfAllowedOutgoingEdges();
 
         // there is no other turn possible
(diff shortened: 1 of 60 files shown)


boldtrn:
Due to the recent fix in #1845 this is working fine now. WDYT? Do we need to add anything here?

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
```diff
@@ -446,12 +450,26 @@ private int getTurn(EdgeIteratorState edge, int baseNode, int prevNode, int adjN
...
         if (otherContinue != null) {
-            //We are at a fork
+            // We are at a fork
             if (!InstructionsHelper.isNameSimilar(name, prevName)
                     || InstructionsHelper.isNameSimilar(otherContinue.getName(), prevName)
-                    || !prevFlag.equals(flag)
-                    || prevFlag.equals(otherContinue.getFlags())
                     || !outgoingEdgesAreSlower) {
+
+                final RoadClass roadClass = roadClassEnc.getEnum(false, edge.getFlags());
```
karussell: It is better to use the shorter notation: `edge.get(roadClassEnc)` for these type of requests. edge.getFlags is kind of deprecated. Similar in the getSpeed method below.

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
```diff
@@ -390,7 +394,7 @@ private int getTurn(EdgeIteratorState edge, int baseNode, int prevNode, int adjN
             forceInstruction = true;
         }
 
-        InstructionsOutgoingEdges outgoingEdges = new InstructionsOutgoingEdges(prevEdge, edge, encoder, crossingExplorer, nodeAccess, prevNode, baseNode, adjNode);
+        InstructionsOutgoingEdges outgoingEdges = new InstructionsOutgoingEdges(prevEdge, edge, encoder, maxSpeedEnc, crossingExplorer, nodeAccess, prevNode, baseNode, adjNode);
```
karussell: It is not really related to this issue, but maybe you can have a look if we can solve it here:

the crossingExplorer might have a problem. It is created like:
```
crossingExplorer = graph.createEdgeExplorer(DefaultEdgeFilter.allEdges(encoder));
```

But in the InstructionsOutgoingEdges constructor we do:

allOutgoingEdges.add(tmpEdge);

Couldn't it be that some of these edges are also incoming edges and so at least the name allOutgoingEdges is not really correct?

karussell: But maybe this should go into another issue ...

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
boldtrn: Thanks, this method was used several times in the Instruction generation. I changed these parts.

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
boldtrn: The whole code looks like:

```java
            if (edgeIter.getAdjNode() != prevNode && edgeIter.getAdjNode() != adjNode) {
                tmpEdge = edgeIter.detach(false);
                allOutgoingEdges.add(tmpEdge);
                if (tmpEdge.get(accessEnc)) {
                    allowedOutgoingEdges.add(tmpEdge);
                }
            }
```

Shouldn't the if filter out the edges that are on the path, so it's only the outgoing edges, that are not on the path?

Yes, the naming could definitely be improved, as it's not strictly clear what edges are in this collection.

What name should we choose?
allUnusedOutgoingEdges
allOutgoingEdgesThatAreNotOnThePath (a bit too long :D)

It's not, possible turns, as it also contains roads you are not allowed to turn on. Possible turns are stored in `allowedOutgoingEdges`.

karussell: What I meant was more: could it be that `allOutgoingEdges` also includes the incoming edges (that are not on the path)?

As the explorer is created via `createEdgeExplorer(DefaultEdgeFilter.allEdges(encoder));` and not via `DefaultEdgeFilter.outEdges`

boldtrn: Yes, IIRC it was designed like this. For this we have the allowedOutgoingEdges, these should only contain the edges that we could legally turn onto (ignoring turn restrictions - I think).

We use this for example [here](https://github.com/graphhopper/graphhopper/blob/58cdbdfcfd9862b354e9066c0ff0b67ab3179888/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java#L398). Even though, we are only allowed to turn on one alternative, there is another edge that could potentially be mistaken for a turn, so we create a turn instruction to be sure that this is a turn.

boldtrn: Also see the [todo](https://github.com/graphhopper/graphhopper/blob/58cdbdfcfd9862b354e9066c0ff0b67ab3179888/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java#L402) a bit further down :). Maybe we should do the oneway check additionally?

karussell: Probably just the name of `allOutgoingEdges` is what confuses me. Should we rename it into `edgeList` and the variable `allowedOutgoingEdges` into `outgoingEdgeList`?

boldtrn: It's not the edgeList either. It's the edgeList of a junction without the streets that are used by the route. Maybe remainingEdgeList?

karussell: This is a bit confusing ... or maybe offPathEdges and outgoingOffPathEdges? Not sure. But this also does not reflect the problem that only vehicle-specific off-the-path edges are included...

karussell: Let's solve this is a different issue.

karussell:
In the following situation the second continue instruction is removed (see MapMatchingTest.testDoWork):

http://localhost:8989/maps/?point=51.331131%2C12.379956&point=51.330293%2C12.38149

So, before there were 2 instructions: "continue onto Windmühlenstraße" and "continue onto Bayrischer Platz" and now there is just the first instruction. This is an improvement IMO as the road name changes on a non-junction node and so the instruction is unnecessary. 

But could there be instructions removed due to this PR although the road name changes directly on a junctions? Personally I would prefer to keep these instructions to make navigation clear in a dense area.

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
boldtrn: > But this also does not reflect the problem that only vehicle-specific off-the-path edges are included...

Yes indeed, should we create a new issue for this?

karussell:
> In my opinion, we should only provide an instruction if the user has to change something, like turning, switching lanes,

Yes, in theory I agree. The problem in practice is that often junctions are not perfectly mapped and so omitting continue instructions for the inner-city case could be problematic if the road name changes. But we'll see what feedback we'll get. Just wanted to understand the now failing map matching test and will now just adapt to the new instructions.

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
karussell: Or should I just rename it to a neutral filteredEdges and filteredOutgoingEdges?

boldtrn:
Yes that makes perfect sense :). Let's see what people think. Unfortunately in the real world there are so many different scenarios and you cannot easily test them without knowing the actual situation.

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
boldtrn: Personally, I am a fan of speaking names :). So I'd prefer a name that explains the meaning of these edges. So I would ask myself what kind of edges these are and would read the comments :). But if you think, the current name leads to potential misunderstandings of the code, we should rename it and then better give it a more generic name than a name that could be misunderstood :).

We should still look into the issues of non vehicle accessible edges to be considered for the instruction generation. And maybe also oneways, turn restrictions, etc.

But maybe we do this, when we find a good test scenario where the instructions are not correct?

karussell: > So I'd prefer a name that explains the meaning of these edges.

But sometimes it is not really possible to put all the meaning in it ... i.e. it will get pretty nasty here like outgoingVehicleSpecificOffPathEdges :D and keeping the name is confusing - at least for me :)

> We should still look into the issues of non vehicle accessible edges to be considered for the instruction generation. 

Ok, yes. We can create such an issue. For car it shouldn't be a problem as ignoring bike-only paths is intended, but for bike ignoring car-only roads might be confusing, but I don't know.

boldtrn: > and keeping the name is confusing - at least for me :)

Yes, in this case, let's rename it :)

> but for bike ignoring car-only roads might be confusing, but I don't know.

Yes indeed, or maybe for some special vehicles like scooters, wheelchairs, trucks, etc.

PR #1845, opened 2019-12-31
avoid skipping ways if mixing encoders in OSMRoadClassParser

### LINKED ISSUE #1843: Issue RoadClassEncoder via EncodedValueLookup when using mtb and motorway

karussell:
fixes #1843

The reason why there is a difference for bike vs. mtb: in AcceptWay we use HashMap and getAccess uses the first entry which changes for different String keys. 

 - [x] remove missleading AcceptWay.getAccess


core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadClassParser.java
```diff
@@ -43,8 +42,8 @@ public void createEncodedValues(EncodedValueLookup lookup, List<EncodedValue> li
     }
 
     @Override
-    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, EncodingManager.Access access, IntsRef relationFlags) {
-        if (!access.isWay())
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, boolean ferry, IntsRef relationFlags) {
+        if (ferry)
```
karussell: This is the actual fix.

boldtrn: Ah I see, that explains the issue :+1:.

boldtrn: BTW: do we need to pass the ferry boolean? We pass the that should contain the ferry information? Or is this for some ferry relations or something like this?

karussell:
> We already pass the ReaderWay which should contain this information anyway

Yes, true. I would do this later when we remove the FlagEncoders completely. E.g. I'm unsure if we need the separate `encodingManager.acceptWay` method. Like you said: it might be slightly faster with it, and if it is not e.g. at least 10% faster it is certainly not necessary. And when we remove that method we could remove the ferry parameter too.

Another solution would be to create artificial tags like we already do with `estimated_distance` or set a ferry EncodedValue and ensure that it is available in every TagParser.

PR #1846, opened 2020-01-08
Make alternatives for CH faster

karussell:
This makes the CH route requests with alternatives much faster. It is still 3 to 8 times slower compared to raw "routingCH" speed (slower for longer routes).

The only required change is to add a parameter `SPTEntry entry` to BidirPathExtractor.onEdge. Not sure if this is nice @easbar?

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
+++ b/core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
@@ -75,9 +75,10 @@ public class AlternativeRouteCH extends DijkstraBidirectionCHNoSOD {
                 if (fromSPTEntry.getWeightOfVisitedPath() + toSPTEntry.getWeightOfVisitedPath() > maxWeight)
                     return true;
 
-                // The path s -> v -> t is not the shortest path. Only s -> v and v -> t are.
-                // We still use this preliminary path to filter for shared path length with other alternatives,
-                // so we don't have to work so much.
+                // This gives us a path s -> v -> t, but since we are using contraction hierarchies,
+                // s -> v and v -> t need not be shortest paths (see ).
+                // In fact, they can sometimes be pretty strange. We still use this preliminary path to filter
+                // for shared path length with other alternatives, so we don't have to work so much.
 //                Path preliminaryRoute = createPathExtractor(graph, weighting).extract(fromSPTEntry, toSPTEntry, fromSPTEntry.getWeightOfVisitedPath() + toSPTEntry.getWeightOfVisitedPath());
 //                double preliminaryShare = calculateShare(preliminaryRoute);
 //                if (preliminaryShare > maxShareFactor) {
--- a/core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
+++ b/core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
@@ -79,11 +79,6 @@ public class AlternativeRouteCH extends DijkstraBidirectionCHNoSOD {
                 // s -> v and v -> t need not be shortest paths (see ).
                 // In fact, they can sometimes be pretty strange. We still use this preliminary path to filter
                 // for shared path length with other alternatives, so we don't have to work so much.
-//                Path preliminaryRoute = createPathExtractor(graph, weighting).extract(fromSPTEntry, toSPTEntry, fromSPTEntry.getWeightOfVisitedPath() + toSPTEntry.getWeightOfVisitedPath());
-//                double preliminaryShare = calculateShare(preliminaryRoute);
-//                if (preliminaryShare > maxShareFactor) {
-//                    return true;
-//                }
                 // this reduces alternative rate from 2.0 to 1.6 but makes algorithm roughly 3x faster
                 final List<SPTEntry> sptEntries = calcSPEList(fromSPTEntry, toSPTEntry);
                 double preliminaryShare = calculateShare(sptEntries) / (fromSPTEntry.getWeightOfVisitedPath() + toSPTEntry.getWeightOfVisitedPath());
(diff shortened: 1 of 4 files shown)


easbar:
Nice! What's the (main) reason it is faster now ? To me calling `calcEdges` still sounds slow. Btw we really should add a RandomEdgeExplorer so we can efficiently (without object creation) get an EdgeIteratorState from an edge. I expect this to lead to a solid speed-up in multiple places and doing so should be preferred to working around calling `getEdgeIteratorState(edge)`. I just opened #1847 for this.

Does this PR alter the *quality* of the alternative routes ?

easbar:
> The only required change is to add a parameter SPTEntry entry to BidirPathExtractor.onEdge

Ok this is because you need the shortest path tree entries of the CH tree, right ? I think the `SPTEntry` parameter is fine because `BidirPathExtractor` is using this class extensively anyway. But maybe we can remove the edge/adjNode parameters in `onEdge` then ?

core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
```diff
@@ -73,11 +75,11 @@ public boolean apply(final int v, final SPTEntry fromSPTEntry) {
                     return true;
 
                 // This gives us a path s -> v -> t, but since we are using contraction hierarchies,
-                // s -> v and v -> t need not be shortest paths. In fact, they can sometimes be pretty strange.
-                // We still use this preliminary path to filter for shared path length with other alternatives,
-                // so we don't have to work so much.
-                Path preliminaryRoute = createPathExtractor(graph, weighting).extract(fromSPTEntry, toSPTEntry, fromSPTEntry.getWeightOfVisitedPath() + toSPTEntry.getWeightOfVisitedPath());
-                double preliminaryShare = calculateShare(preliminaryRoute);
+                // s -> v and v -> t need not be shortest paths (see ).
```
easbar: Looks like you wanted to add a reference here (see.. ?)

core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
```diff
@@ -128,15 +132,15 @@ Path extract() {
...
 
-            private double calculateShare(final Path path) {
-                double sharedDistance = 0.0;
-                List<EdgeIteratorState> edges = path.calcEdges();
-                for (EdgeIteratorState edge : edges) {
-                    if (nodesInCurrentAlternativeSetContains(edge.getBaseNode()) && nodesInCurrentAlternativeSetContains(edge.getAdjNode())) {
-                        sharedDistance += edge.getDistance();
+            private double calculateShare(final List<SPTEntry> sptEntries) {
+                double sharedWeight = 0.0;
+                for (int i = 0; i < sptEntries.size(); i++) {
```
easbar: Maybe for-each loop here or is it really faster that way ?

core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
```diff
@@ -191,6 +195,17 @@ public int compare(AlternativeInfo o1, AlternativeInfo o2) {
         return alternatives;
     }
 
+    private List<SPTEntry> calcSPEList(final SPTEntry fromSPTEntry, final SPTEntry toSPTEntry) {
+        final List<SPTEntry> list = new ArrayList<>();
+        new NodeBasedCHBidirPathExtractor(graph, graph.getBaseGraph(), weighting) {
```
easbar: Since we are not even unpacking any shortcuts here using BidirPathExtractor (without CH) should be enough ? To further clean this up we could create something that is *only* concerned with following the parent pointers (and provides some hooks like onEdge etc.) (and not with `Path` at all) and *use* this within the path extractor. But not sure if this is worth the effort atm.

michaz:
Guys, _careful_ here, there's a reason I wrote this pretty high-level. It's a high-level algorithm which uses multiple whole routers as subroutines. Let's not micro-optimize in there, otherwise we'll a) not understand what it's doing and b) break things.

Instead, if we find that _something_ (like path extraction, iterating over things without object creation) is a hot spot, then let's just plan to write a faster implementation of that something.

core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
karussell: After some micro benchmarks a few years ago I avoid this for-each loop. E.g. in EncodingManager.reverseFlag there was a change that improved routing speed when replacing the for-each loop once.

karussell:
@easbar the main problem is that in extract EdgeIteratorStates are calculated and in path.calcEdges again and that although we do not need them.

> Let's not micro-optimize in there

@michaz a speed up of at least 6x is not a micro optimization IMO. 

Are the changes making things unreadable for you? The biggest boost was when I replaced the usage of the `List<EdgeIteratorState>` for the preliminaryShare with `List<SPTEntry>` (see first commit). Everything else is neglectable and could be reverted if you think that this is not easy to read.

michaz:
> Are the changes making things unreadable for you? 

At least it's not doing the same thing as before, and I don't immediately see what it is that it's doing. Before, I did.

> The biggest boost was when I replaced

Yes, but perhaps because it is doing something different know.

easbar:
I think I agree with both of you. For one we should not do optimizations that reduce the quality of the alternative routes right now (or even risk doing so in case we cannot precisely evaluate the quality at the moment / yet). Especially not before fixing #1847, because this will indeed fix a hot spot without changing any behavior. On the other hand I would neither say the code gets too low level or is hard to understand with this PR nor we should under and circumstances leave it as is.

Maybe the reason this PR does not produce the same results as master is that the share factor was changed from 0.7 to 0.6 ?

michaz:
No, the reason is that it _calculates something completely different_.

karussell:
> Does this PR alter the quality of the alternative routes ?

Yes, likely as the `routingCH_alt.alternative_rate` is reduced from e.g. 3.9 to 2.7 for German map, but hard to measure if this is a real loss.

> No, the reason is that it calculates something completely different.

Can you revisit the first commit? Instead of calculating the preliminary path explicitly it does it now implicitly. So it should not be far from the original code. If there is no bug (will revisit this again), then the reason is that for the preliminaryShare it switches from the "distance-based" share calculation to a "weight-based" one. (The share factor does not have a big influence, neither speed, nor average count)

karussell:
Oh, no. Thanks @michaz. There are two flaws I need to fix: I'm adding the weight of the SPTEntry objects and I do not expand them. 

Probably it is best to continue with #1847 and see if this speeds this up here too.

Issue #1847, opened 2020-01-08
Speed up `Graph#getEdgeIteratorState`


easbar:
Obtaining an `EdgeIteratorState` given an integer edge id is unnecessarily expensive, because `Graph#getEdgeIteratorState` creates an object instead of simply updating the edge id being in use as we do with the other `EdgeExplorer`s. We need something like `RandomAccessEdgeExplorer#setEdgeId` in analogy to `EdgeExplorer.setBaseNode`.

 Probably we should do this after #1780.

michaz:
Can I ask something outrageous? Do you thing we could -- in the long run -- a) rename these things to *Cursor, and b) refactor them so that we can make them Iterable, like in the primitive-collections library we are using?

easbar:
Doesn't sound so outrageous :) I don't really care about Iterator vs. Cursor what do you think makes the difference ? What do you mean making them iterable ? You mean having something like prev() in addition to next() ?

michaz:
Yes. java.lang.iterable. The things in carrotsearch-collections: You can foreach them, even though they are cursors.

michaz:
(I can‘t imagine how it could be slower than manually writing a hasNext()-next()-loop. In fact I think it‘s literally compiled to that.)

PR #1850, opened 2020-01-10
Use RoutingCHGraph adapter for CH query algorithms

easbar:
This is more or less a follow up of #1818 and part of #1780. There is a new interface `RoutingCHGraph` (with corresponding implementation) that is used by all CH algorithms. Internally it still uses `BaseGraph/CHGraphImpl`, but this is no longer (as) exposed to the algorithms, which can now retrieve the edge weights via `edge.getWeight()`. It also helps for #1835/#1776, because all access flag checks are run immediately before `calcWeight` already and only need to be moved into `Weighting` in #1835. 

So far the query speed seems to decrease by about 10%, I could not find out why exactly this is, but I am assuming its because we are wrapping the routing ch graph iterators around the base/ch graph iterators. Not sure, but since this will later be removed anyway I'd say this is ok for now ?


core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
```diff
@@ -76,22 +76,21 @@ public boolean apply(final int v, final SPTEntry fromSPTEntry) {
...
-                svRouter.setEdgeFilter(additionalEdgeFilter);
+                DijkstraBidirectionCHNoSOD svRouter = new DijkstraBidirectionCHNoSOD(graph);
                 final Path svPath = svRouter.calcPath(s, v);
                 final IntIndexedContainer svNodes = svPath.calcNodes();
-                DijkstraBidirectionCHNoSOD vtRouter = new DijkstraBidirectionCHNoSOD(graph, weighting);
-                vtRouter.setEdgeFilter(additionalEdgeFilter);
+                DijkstraBidirectionCHNoSOD vtRouter = new DijkstraBidirectionCHNoSOD(graph);
                 final Path vtPath = vtRouter.calcPath(v, t);
                 final IntIndexedContainer vtNodes = vtPath.calcNodes();
+                // todonow: which graph do we want here ?
```
easbar: @michaz ? Do we wan the BaseGraph, the QueryGraph (of the CHGraph) or the CHGraph ?

core/src/main/java/com/graphhopper/routing/AbstractRoutingAlgorithm.java
```diff
@@ -56,8 +57,9 @@ public AbstractRoutingAlgorithm(Graph graph, Weighting weighting, TraversalMode
         this.traversalMode = traversalMode;
         this.graph = graph;
         this.nodeAccess = graph.getNodeAccess();
-        outEdgeExplorer = graph.createEdgeExplorer(DefaultEdgeFilter.outEdges(flagEncoder));
-        inEdgeExplorer = graph.createEdgeExplorer(DefaultEdgeFilter.inEdges(flagEncoder));
+        inEdgeFilter = DefaultEdgeFilter.inEdges(flagEncoder.getAccessEnc());
+        outEdgeFilter = DefaultEdgeFilter.outEdges(flagEncoder.getAccessEnc());
+        edgeExplorer = graph.createEdgeExplorer();
     }
```
easbar: I already removed the in/outEdgeExplorer for #1835 here. I left a few todos where we can simply move the access flags into the weighting for #1835.

core/src/main/java/com/graphhopper/routing/Dijkstra.java
```diff
@@ -68,19 +67,19 @@ public Path calcPath(int from, int to) {
...
             int currNode = currEdge.adjNode;
-            EdgeIterator iter = explorer.setBaseNode(currNode);
+            EdgeIterator iter = edgeExplorer.setBaseNode(currNode);
             while (iter.next()) {
                 if (!accept(iter, currEdge.edge))
                     continue;
 
-                double tmpWeight = weighting.calcWeight(iter, false, currEdge.edge) + currEdge.weight;
+                // todo: for #1776/#1835 move the access check into weighting
+                double tmpWeight = !outEdgeFilter.accept(iter) ? Double.POSITIVE_INFINITY : weighting.calcWeight(iter, false, currEdge.edge) + currEdge.weight;
```
easbar: In #18350 we can remove outEdgeFilter here and strictly rely on the returned weight from `Weighting`

core/src/main/java/com/graphhopper/routing/AlternativeRouteCH.java
michaz: Well, the Path will be a path through the .. well, the graph. :-) Doesn't use shortcuts, but _does_ use the query-graph nodes.

Technically, the QueryGraph of the BaseGraph, I guess. But the QueryGraph of the CHGraph works as well, I think.

easbar: So you need virtual nodes + edges but you do *not* want shortcuts ?

michaz: Correct, but I think I don't mind _if there are_ shortcuts, I just don't use them.

The graph at this point is just used because the constructor of Path wants it, and the Path will not have shortcuts. Because I concatenate it from two "expanded" Paths from the CH-router.

michaz: If (as you say) a CHGraph is ultimately something _different_ than a Graph, then this class must conceptually probably get the CHGraph _and_ the Graph, because it calls sub-routers (with the CHGraph), but it also wants to reason about the (pure) Graph.

easbar: It looks like the overriden `.extract()` method does not use the `graph` at all ? Do you really have to use the Path this way ? Can we not just get a list of edgeiteratorstates + total distance or something ?

easbar: Ah never mind the path is passed to another method that calls calcEdges (which uses the graph) ...

michaz: Yeah, the overridden extract is just a relic from something that was there before. If I don't need the Graph at all here, throw it out, no problem.

But I do want a Path. A Path through a Graph is something we should have as an element, without having to chop it up into bits.

easbar: Ok I changed to `graph.getGraph().getBaseGraph()` (sigh). This will be the query graph of the base graph, but we need to get rid of this construct. 

> A Path through a Graph is something we should have as an element, without having to chop it up into bits

? Can you explain what you mean here ? Where is the path chopped up into bits ?

easbar: There should also be some tests for alternative route CH, maybe checking that all returned paths start and end in the right places and the weight is less than twice as big as the best one or something like this ?

core/src/main/java/com/graphhopper/routing/AbstractBidirCHAlgo.java
```diff
@@ -0,0 +1,308 @@
...
+                prioQueue.remove(entry);
+                updateEntry(entry, iter, origEdgeId, weight, currEdge, reverse);
+                prioQueue.add(entry);
+            } else
+                continue;
+
+            if (updateBestPath) {
+                // only needed for edge-based -> skip the calculation and use dummy value otherwise
+                double edgeWeight = traversalMode.isEdgeBased() ? calcWeight(iter, reverse, NO_EDGE) : Double.POSITIVE_INFINITY;
+                updateBestPath(edgeWeight, entry, origEdgeId, traversalId, reverse);
```
karussell: Here the calcWeight operation is done although it could happen that the backward search has no entry and the weight is not used. Might be related to (not very likely):

> So far the query speed seems to decrease by about 10%, I could not find out why exactly this is, 

BTW: does this slowdown occur for edge- and node-based queries?

core/src/main/java/com/graphhopper/routing/AStar.java
```diff
@@ -83,19 +86,19 @@ public Path calcPath(int from, int to) {
...
             int currNode = currEdge.adjNode;
-            EdgeIterator iter = explorer.setBaseNode(currNode);
+            EdgeIterator iter = edgeExplorer.setBaseNode(currNode);
             while (iter.next()) {
                 if (!accept(iter, currEdge.edge))
                     continue;
 
-                double tmpWeight = weighting.calcWeight(iter, false, currEdge.edge) + currEdge.weightOfVisitedPath;
+                // todo: for #1776/#1835 move the access check into weighting
+                double tmpWeight = !outEdgeFilter.accept(iter) ? Double.POSITIVE_INFINITY : weighting.calcWeight(iter, false, currEdge.edge) + currEdge.weightOfVisitedPath;
```
karussell: For clarity it should be better:

```java
!outEdgeFilter.accept(iter) ? Double.POSITIVE_INFINITY : (weighting.calcWeight(iter, false, currEdge.edge) + currEdge.weightOfVisitedPath);
```

core/src/main/java/com/graphhopper/routing/AbstractNonCHBidirAlgo.java
```diff
@@ -0,0 +1,271 @@
...
+    /**
+     * @see #fillEdgesFromUsingFilter(EdgeFilter)
+     */
+    protected void fillEdgesToUsingFilter(EdgeFilter edgeFilter) {
+        // we temporarily ignore the additionalEdgeFilter
+        EdgeFilter tmpFilter = additionalEdgeFilter;
+        additionalEdgeFilter = edgeFilter;
+        finishedTo = !fillEdgesTo();
+        additionalEdgeFilter = tmpFilter;
+    }
```
karussell: What are the use cases for the EdgeFilter for non-CH algos? (For CH there is the CHLevelEdgeFilter, but here?)

Is it just regarding the vehicle edge and we can remove this code with #1835 ?

core/src/main/java/com/graphhopper/routing/AbstractBidirCHAlgo.java
easbar: oh good point, its even worse: edge-based CH overrides the updateBestPath method and does not need edgeWeight in any case. will try if this makes the difference but I would guess so.

> BTW: does this slowdown occur for edge- and node-based queries?

I think it is more pronounced for edge-based, but even after running many measurements I was not entirely sure

core/src/main/java/com/graphhopper/routing/AbstractNonCHBidirAlgo.java
easbar: In most cases one can achieve the same effect as using additionalEdgeFilter by returning infinite weight for the corresponding edges in `Weighting` (or at least this it what it should be like with #1835). The only thing that comes to my mind is the initial routing step for routing with start/target edges where we use the filter to select the start/target edge. But yes we might be able to get rid of the filter entirely for #1835.

core/src/main/java/com/graphhopper/routing/AbstractBidirCHAlgo.java
easbar: No, unfortunately this did not fix it. My next best guess is that `RoutingCHGraph#getEdgeIteratorState` creates two (vs. one) objects, which is relevant for path unpacking. So this could be fixed with #1847

core/src/main/java/com/graphhopper/routing/AbstractNonCHBidirAlgo.java
easbar: Yes I just checked with this PR we can get rid of additional edge filter. Shall I do it here or in a follow up PR ?

PR #1851, opened 2020-01-12
Android: VTM 0.13.0 OpenGL vector map library

devemux86:
[VTM 0.13.0](https://github.com/mapsforge/vtm) OpenGL vector map library.


devemux86:
Loading graphs does not work with GraphHopper snapshot.
Probably the downloaded graphs need to be updated?

karussell:
Thanks. The graph loading for snapshot should be fixed now.

PR #1853, opened 2020-01-17
Replace our polygon code with the JTS implementation

karussell:
I would like to have this before #1838. 

This PR does the following:

 * avoids our custom Polygon code and uses the JTS implementation which should make the implementation more robust and make it working for more generic polygons like with inner rings
 * it additionally improves the precision of the block_area feature through a new `Shape.intersects(PointList)` method. Now not just a single point of an edge is checked for intersection but the whole point list. Still there is more work (later) #1324.
 * I discovered PackedCoordinateSequence that allows us to be memory efficient with JTS too
 * I've introduced QuerySettings. Not sure if this should be done here, but the parameters for the printTimeOfRouteQuery method got more and more confusing. E.g. recent changes like https://github.com/graphhopper/graphhopper/commit/f69f127434e2bc18a273284d4c9b8985de88353c#diff-cd749cc545f8b84137c2b75b2d174b72 are much easier to understand.


api/src/main/java/com/graphhopper/util/PointList.java
```diff
@@ -399,7 +404,10 @@ public static PointList fromLineString(LineString lineString) {
     }
 
     public LineString toLineString(boolean includeElevation) {
-        GeometryFactory gf = new GeometryFactory();
+        return toLineString(new GeometryFactory(), includeElevation);
```
otbutz: Wouldn't it be better to create a factory for 2D and 3D each and keep them?

api/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -17,117 +17,72 @@
...
-
-        boolean inside = false;
-        int len = this.lat.length;
-        for (int i = 0; i < len; i++) {
-            if (edgesAreIntersecting(rayStartLon, rayStartLat, lon, lat, this.lon[i], this.lat[i], this.lon[(i + 1) % len], this.lat[(i + 1) % len]))
-                inside = !inside;
-        }
-        return inside;
-
+        return polygon.contains(factory.createPoint(new Coordinate(lon, lat)));
```
otbutz: We could check the envelope first.

api/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -137,115 +92,55 @@ public boolean contains(Shape s) {
...
 
     public double getMaxLon() {
-        return maxLon;
+        return envelope.getMaxX();
     }
 
     @Override
     public String toString() {
-        return "polygon (" + lat.length + " points)";
+        return "polygon (" + polygon.getGeometry().getNumPoints() + " points)";
```
otbutz: The number of holes would probably be interesting, too

api/src/main/java/com/graphhopper/util/PointList.java
karussell: Have done this (one static factory). Or why would we want 2?

api/src/main/java/com/graphhopper/util/shapes/Polygon.java
karussell: This should be done from JTS.

karussell: This should be done from JTS.

api/src/main/java/com/graphhopper/util/PointList.java
otbutz:
```java
GeometryFactory FAC_2D = new GeometryFactory(new PackedCoordinateSequenceFactory(DOUBLE, 2));
GeometryFactory FAC_3D = new GeometryFactory(new PackedCoordinateSequenceFactory(DOUBLE, 3));
```
`includeElevation` would just toggle which one of the two is being passed to the function. No need to create new objects here.

api/src/main/java/com/graphhopper/util/shapes/Polygon.java
otbutz: Ah, i missed that `polygon` is already a `PreparedGeometry`. I would recommend to rename it accordingly e.g `prepPolygon`

api/src/main/java/com/graphhopper/util/shapes/BBox.java
```diff
@@ -161,6 +164,36 @@ public boolean contains(Shape s) {
...
+        if (len == 0)
+            throw new IllegalArgumentException("PointList must not be empty");
+
+        double lat = pointList.getLatitude(0);
+        double lon = pointList.getLongitude(0);
+        if (len == 1)
+            return contains(lat, lon);
+
+        Coordinate coords = new Coordinate(lon, lat);
+        RectangleLineIntersector intersector = getCachedRectangleLineIntersector();
```
michaz: I think this will fail when this method is called after this object was updated with update(). BBox is mutable, so I think we can't cache the RectangleLineIntersector.

api/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -137,115 +92,55 @@ public boolean contains(Shape s) {
...
-
-        // If we get here, only two possibilities are left. Either the two
-        // vectors intersect in exactly one point or they are collinear, which
-        // means they intersect in any number of points from zero to infinite.
-        if ((a1 * b2) - (a2 * b1) == 0) return false;
-
-        // If they are not collinear, they must intersect in exactly one point.
-        return true;
+        // for estimation use bounding box as reference:
+        return getBounds().calculateArea() * envelope.getArea() / polygon.getGeometry().getArea();
```
michaz: Do you think we could just remove this method, and its definition from `Shape`? It isn't used generically anywhere in the code, so we could easily avoid giving an "approximate" or fudged implementation here.

karussell: Unfortunately it is used in GraphEdgeIdFinder.parseBlockArea

api/src/main/java/com/graphhopper/util/shapes/BBox.java
karussell: Yeah, will see if this is critical for speed and if there is a different possibility.

karussell:
> Ultimately (long term, new code), I would recommend we stop wrapping the JTS classes, and migrate to them.

:+1: 
Yes, I tried that here, but this is a major undertaking as we would not only remove Circle (not available in JTS) but migrate BBox and GHPoint etc. The big problems were for PointList, as it is highly tuned for our needs especially regarding memory usage when simplifying.

api/src/main/java/com/graphhopper/util/PointList.java
karussell: Have adapted the current code (via a single factory).

karussell: Or will there be something different with your proposed changed?

karussell: Hmmh, maybe you mean something else? There is just one call and it is:
```
return factory.createLineString(new PackedCoordinateSequence.Double(coordinates, includeElevation ? 3 : 2));
```

api/src/main/java/com/graphhopper/util/shapes/Polygon.java
karussell: Ok, wasn't used before (seems like a bug), but is now :)
I.e. for smaller polygons edgeIds should be used and for bigger polygons we expect that the edgeId lookup is slower than calling the polygon.intersect for every edge of the traversed graph.

api/src/main/java/com/graphhopper/util/shapes/BBox.java
karussell: Have moved the cached RectangleLineIntersector upwards to GraphEdgeIdFinder.parseBlockArea

karussell: Have moved the cached RectangleLineIntersector upwards to GraphEdgeIdFinder.parseBlockArea

api/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -137,115 +92,55 @@ public boolean contains(Shape s) {
...
+        Point centroid = prepPolygon.getGeometry().getCentroid();
+        return new GHPoint(centroid.getY(), centroid.getX());
     }
 
     @Override
     public double calculateArea() {
-        throw new UnsupportedOperationException("Not supported yet.");
-    }
-
-    private boolean edgesAreIntersecting(
```
boldtrn: I think it's great that we can remove this method! This was always a bit of a headache factor for me!

karussell:
@easbar can you have a short look into my change in Measurement (regarding QuerySetting). Do you agree roughly :) ? Should I do this in a separate PR?

easbar:
> @easbar can you have a short look into my change in Measurement (regarding QuerySetting). Do you agree roughly :) ? Should I do this in a separate PR?

Yes putting all these booleans in a config object is a good idea imo. Right now the biggest advantage is that we only have to specify the non-default values (all the booleans that should be true instead of false)?

karussell:
> Right now the biggest advantage is that we only have to specify the non-default values 

Yes (and you can better see what changes to the "default")

PR #1863, opened 2020-01-23
Remove TurnWeighting and handle turn weights within Weighting instead

### LINKED ISSUE #1820: Remove TurnWeighting

easbar:
Fixes #1820 

This is (more or less) what happens here:

* `Weighting` has two new methods `calcTurnWeight/Millis(inEdge, viaNode, outEdge)` so besides edge-weights it now also calculates turn weights

* The turn weight is calculated by implementations of a new interface `TurnCostProvider`, there is `NoTurnCostProvider` in case the turn costs should be zero. `DefaultTurnCostProvider` handles u-turn costs as we did so far and applies turn restrictions (same as `TurnWeighting` did before)

* `AbstractWeighting` has a `TurnCostProvider`, which is passed via the constructor -> It must be known at the time the `Weighting` is created (in contrast to the old `TurnWeighting` which wrapped an existing `Weighting` after it was created). 

* Since `AbstractWeighting` provides default implementations for `calcTurnWeight/Millis`, the sub-classes usually do not need to be concerned with turn weightings. They only need to override `calc*Edge*Weight/Millis`. If you want to change the turn weights in a subclass you can just pass another `TurnCostProvider to the `AbstractWeighting` constructor.

* Some `Weighting`s do not inherit from `AbstractWeighting`, but rather wrap another weighting (e.g. avoid edges and block area). For this reason its ugly to have `Weighting#calcWeight(edgeState, reverse, prevNext)` which is now just a combination of edge and turn weights anyway. For example the `AvoidEdgesWeighting` adjusts the edge weights, but is unaware of the turn cost implementation of the wrapped weighting, so its not clear how to implement `calcWeight` based on the wrapped weighting. Therefore I removed `Weighting#calcWeight`. For the (quite common) case where we combine edge- with turn-weights there is a static helper method (currently in `AbstractWeighting`).

* There is one more (bigger) complication: Turn costs at virtual nodes and edges. So far we had to use the (special) `TurnCostStorage` from `QueryGraph` and pass it into `TurnWeighting`. But now that `TurnWeighting` is gone we have to wrap the `Weighting` into `QueryGraphWeighting` when we use it with `QueryGraph` instead (at least to calculate turn costs). This is a bit ugly to do, so I added `Graph#wrapWeighting` (which is only really implemented by `QueryGraph` for now).


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -847,7 +847,8 @@ private void initLMAlgoFactoryDecorator() {
 
         for (FlagEncoder encoder : encodingManager.fetchEdgeEncoders()) {
             for (String lmWeightingStr : lmFactoryDecorator.getWeightingsAsStrings()) {
-                Weighting weighting = createWeighting(new HintsMap(lmWeightingStr), encoder, null);
+                // note that we do not consider turn costs during LM preparation?
+                Weighting weighting = createWeighting(new HintsMap(lmWeightingStr), encoder, null, new NoTurnCostProvider());
```
karussell: Yes, this is intended. In a separate issue we could try if this would have any negative consequences.

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
```diff
@@ -508,7 +508,7 @@ private void updatePointsAndInstruction(EdgeIteratorState edge, PointList pl) {
         }
         double newDist = edge.getDistance();
         prevInstruction.setDistance(newDist + prevInstruction.getDistance());
-        // todonow: why do we not account for turn times here ?
+        // todo: why do we not account for turn times here ?
```
karussell: Should we create an issue as reminder for this?

core/src/main/java/com/graphhopper/routing/weighting/DefaultTurnCostProvider.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return toString();
+    }
+
+    @Override
+    public String toString() {
+        return "" + uTurnCostsInt;
```
karussell: Shouldn't it include the FlagEncoder too? Because without it, then different providers will have the same name.

(and in weighting.matches we could adapt the equals)

core/src/main/java/com/graphhopper/routing/weighting/NoTurnCostProvider.java
```diff
@@ -0,0 +1,36 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.graphhopper.routing.weighting;
+
+public class NoTurnCostProvider implements TurnCostProvider {
```
karussell: Maybe we should avoid instantiating this provider and create a singleton from it? Or even remove this class and create a variable "NO_TURN_COST" in TurnCostProvider?

core/src/main/java/com/graphhopper/util/Constants.java
```diff
@@ -66,6 +66,7 @@
     private static final int JVM_MAJOR_VERSION;
     private static final int JVM_MINOR_VERSION;
 
+    // todonow: make sure this version is compatible also for older imports!
```
karussell: Do you think this is required? (similar question applies to the method where you kept the CH file names compatible)

We never kept backward compatibility for our files as they are recreated due to every OSM data update. I do not see value until we officially support offline routing support (maybe something for 3.0 ;) )

easbar:
Thanks a lot for taking a look!

> The only bigger problem I have is with the static AbstractWeighting.calc* methods - it just looks ugly :) ... should we make them non-static and include them in the Weighting interface? 

Yes this is not ideal, but having them non-static isn't better IMO. The reason I did not include it in `Weighting` is that its not clear how to implement when a weighting wraps another, for example `BlockAreaWeighting` (like I said in initial comment). 

> If you prefer to keep this interface lean then should we move the static methods to the interface? IMO it is applicable to all Weighting implementations not only the ones that extend AbstractWeighting (?)

Its certainly misleading that the static method is in `AbstractWeighting`. I would certainly like to move it into `Weighting`, but unfortunately that is not possible in Java 7. We could move it into `GHUtilities`? Or if you like it better we can of course also put it into the interface and somehow document the issue with wrapping `Weighting`s.

> Is it still a WIP?

I don't think there is much missing, should be possible to merge soon.

core/src/main/java/com/graphhopper/util/Constants.java
easbar: Its probably not required because (?) I kept the old file names. I thought it would be ugly to change the filenames now and then change them again soon in #1708. Do you want to remove this complexity and just change the filenames now also? Otherwise we probably do not need to change the data versions, but I left this comment to double check.

core/src/main/java/com/graphhopper/routing/weighting/NoTurnCostProvider.java
easbar: Ok, will do.

core/src/main/java/com/graphhopper/routing/weighting/DefaultTurnCostProvider.java
easbar: Hmm, the `FlagEncoder` belongs to the `Weighting` and the `DefaultTurnCostProvider` in a way.. Right now its included in the `Weighting` name only.

devemux86:
> but unfortunately that is not possible in Java 7

Indeed that would break Java 7 compatibility, as seems needing Java 8.

core/src/main/java/com/graphhopper/storage/CHProfile.java
```diff
@@ -20,14 +20,18 @@
 public class CHProfile {
     private final Weighting weighting;
     private final boolean edgeBased;
-    private final int uTurnCosts;
 
     public static CHProfile nodeBased(Weighting weighting) {
-        return new CHProfile(weighting, TraversalMode.NODE_BASED, INFINITE_U_TURN_COSTS);
+        // todonow: should we do such a check (check if the weighting has turn costs ?)
```
easbar: Any opinion here? Its no longer easy to do this check, we would need something like `weighting.hasTurnCosts()` (and maybe check if `TurnCostProvider` is a `NoTurnCostProvider`. But that doesn't seem very clean. Also using a `TurnCostProvider` and then choosing `nodeBased` is not necessarily an error (but might be useful to receive a warning...).

core/src/main/java/com/graphhopper/storage/CHProfile.java
```diff
@@ -20,14 +20,18 @@
...
+//        if (!edgeBased && uTurnCosts != INFINITE_U_TURN_COSTS) {
+//            throw new IllegalArgumentException("Finite u-turn costs are only allowed for edge-based CH");
+//        }
+        return new CHProfile(weighting, TraversalMode.NODE_BASED);
     }
 
-    public static CHProfile edgeBased(Weighting weighting, int uTurnCosts) {
-        return new CHProfile(weighting, TraversalMode.EDGE_BASED, uTurnCosts);
+    public static CHProfile edgeBased(Weighting weighting) {
+        // todonow: should we check/assure that the weighting must have some valid turn costs ?
```
easbar: Similar question here. But maybe we even want edge-based without turn costs for some(?) reason. #1702

core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
easbar: #1867

core/src/main/java/com/graphhopper/util/Constants.java
easbar: With my most recent commits this PR should be compatible with current master. On the other hand there is string manipulation logic to achieve this. I think its ok like this and this will be changed with #1708 again. If you prefer removing the string manipulation and not being compatible this is also totally fine for me.

core/src/main/java/com/graphhopper/storage/CHProfile.java
easbar: I removed the check, it would be nice to have, but not really important either imo.

core/src/main/java/com/graphhopper/storage/CHProfile.java
easbar: I removed the check, here I actually think its bad to do it because of #1702.

core/src/main/java/com/graphhopper/routing/weighting/DefaultTurnCostProvider.java
karussell: I see, this would be a bit of a circular dependency then. Still it feels strange that there will be different turn cost providers with the same name as soon as multiple vehicles are imported.

core/src/main/java/com/graphhopper/util/Constants.java
karussell: :+1: I have no strong opinion here :)

karussell:
> I would certainly like to move it into Weighting, but unfortunately that is not possible in Java 7

Ah, this is ugly. Thought this was only regarding the default methods.

core/src/main/java/com/graphhopper/storage/CHProfile.java
karussell: Ok

core/src/main/java/com/graphhopper/routing/weighting/DefaultTurnCostProvider.java
easbar: Hmm yes this is not ideal. `DefaultTurnCostProvider` and `AbstractWeighting` are both using a `FlagEncoder` (and it most likely always has to be the same), so maybe the most correct way to assign the names would be using the `FlagEncoder` name for both? But then the weighting name would be something like `car|fastest|car|40`, which seems bad as well. 

> here will be different turn cost providers with the same name 

Maybe the question here is a turn cost provider is something that can exist without a weighting?

Issue #775, opened 2016-08-23
avoid map zoom out every time a new point is added


karussell:
It would also be nice if the map did not zoom out every time a new point is added.

fbonzon:
> It would also be nice if the map did not zoom out every time a new point is added.

This feature, I have actually recently implemented on my GraphHopper fork. I will isolate needed changes and send a PR !

I did it with an additional `allow_zoom_out` parameter in GHRequest, initially `true`, then set to `false` if the request is triggered with a right-click on the map.

devemux86:
>It would also be nice if the map did not zoom out every time a new point is added.

Doesn't the zoom adjustment happens after each route geometry calculation?

fbonzon:
Yes. Currently, zoom adjustment happens after every route calculation, so that the entire route fits the map. When you were zoomed in even closer before the route calculation, staying at that zoom level is often desirable, instead of being zoomed out to see the entire route.

devemux86:
Then an external setting wouldn't be more appropriate, as the auto-zoom is app's responsibility getting the result, instead of passing options to GH core?

fbonzon:
Correct. That's the idea. Only the webapp would be changed. There is already a `do_zoom` parameter actually in the JavaScript code, for doing any zoom (in or out): https://github.com/graphhopper/graphhopper/blob/1e30238e76bacbe506593c6277e768d8f810547a/web/src/main/webapp/js/graphhopper/GHRequest.js#L40
I'm adding a new similar parameter for zoom out only.

karussell:
> Yes. Currently, zoom adjustment happens after every route calculation

It does not happen if you just drag the markers. Couldn't we use the same mechanism when adding new points? I think I introduced the zoom parameter for this pupose

fbonzon:
Correct, we could do the same. Is actually what I did, but I added a new parameter specific for zoom out. do_zoom forbids zooms in both directions.

In case adding new points through mouse click would result in a zoom in, I still allow the zoom in in my implementation, because it is not harmful. Thus is a different parameter, to disallow an automatic zoom out only. Do you agree?

ammagamma:
> intermediate points should be added just between the two closest points to the newly-added one

[...]

PR #1868, opened 2020-01-26
Disable rezooming on intermediate point addition

twitwi:
This PR prevents the automatic-rezoom features that takes place when the user adds an intermediate destination (which can be very annoying when fine tuning very-very long tracks).

It is a simplified version of the commit removed from https://github.com/graphhopper/graphhopper/pull/1856


twitwi:
There seem to be an issue about that (partly) https://github.com/graphhopper/graphhopper/issues/775

karussell:
Thanks! Will it still zoom to the route for the initial map load?

twitwi:
Yes. The added line is only in "function setIntermediateCoord(e) {" which is called only when the user adds a step.

PR #1871, opened 2020-01-27
AverageSpeedDetails: avoid exception for short edges

### LINKED ISSUE #1848: average_speed was infinite error in PathDetail calculation
I am running into this error:

https://github.com/graphhopper/graphhopper/blob/e72c67d7fdad39c5cd1b3ecaeb91ec65ae5d9eb2/core/src/main/java/com/graphhopper/util/details/AverageSpeedDetails.java#L37-L41

I think the error is thrown whenever the edge distance is below a certain threshold (because then `calcMillis` will return zero). What mechanism is supposed to prevent this error ? We'd have to exclude edges below a certain distance or something (which as far as I know we do not) ?.

And another question: Why are we calculating speed=distance/time here ? Can we not just use the `averageSpeedEnc` to obtain the speed directly from the edge ?

karussell:
fixes #1848

More or less copied your fix @easbar and added a test for it :)


reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -698,11 +698,11 @@ EdgeIteratorState addEdge(int fromIndex, int toIndex, PointList pointList, IntsR
                     pillarNodes.add(lat, lon);
             }
         }
-        if (towerNodeDistance < 0.0001) {
+        if (towerNodeDistance < 0.001) {
             // As investigation shows often two paths should have crossed via one identical point 
             // but end up in two very close points.
             zeroCounter++;
-            towerNodeDistance = 0.0001;
+            towerNodeDistance = 0.001;
```
karussell: ... and avoid 0 distance

otbutz: Would it make sense to define a constant for the minimum storeable distance? These kind of magic numbers always come back to hunt you.

easbar:
Nice thanks! I can confirm this fixes my original issues as well (infinite speed errors when running `Measurement` on `bayern-140101.osm.pbf`).

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
karussell: Yeah, I don't like it too :) . Still "providing an API" for this limitation is uglier IMO at least when this is only needed in one place. Will add a comment to the BaseGraph to remind us of this here when refactoring.

otbutz: An API would be nice but i think that a simple `MIN_STOREABLE_DISTANCE` constant with a comment would be sufficient for readability.

PR #1873, opened 2020-01-28
Make test more stable by using JSONAssert equals

contextshuffling:
Test in `web-api/src/test/java/com/graphhopper/util/InstructionListRepresentationTest.java` uses jackson to serialize objects into json and assert the serialized results with several hard-coded strings. However, the order of serialized json is not guaranteed so tests may fail if the order is different. So tests may fail or pass without any changes made to the source code and cannot serve as good regression tests.

This PR proposes to use JSONAssert and modify the corresponding json test assertions

### DIFF
--- a/web-api/pom.xml
+++ b/web-api/pom.xml
@@ -43,6 +43,12 @@
             <version>${dropwizard.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


karussell:
We never had a problem with this test and so I would not want to change it.

PR #1874, opened 2020-01-28
Fix step numbering in non-firefox browsers

twitwi:
Fixes the issue (firefox only feature) raised in the discussions about the PR that added step numbers in the web UI : https://github.com/graphhopper/graphhopper/pull/1856


karussell:
Thanks - works and is live now!

PR #1876, opened 2020-01-28
Improve picking up default graph

### LINKED ISSUE #1862: Initialization with example config should work like the default Java settings
I am author of https://graphhopper.develar.org/ and contributor of [locus-addon-graphhopper](https://github.com/asamm/locus-addon-graphhopper).

I working on updating to `1.0-pre18`, as latest release 0.13 version is not usable (see https://github.com/asamm/locus-addon-graphhopper/pull/4#issuecomment-577033241)

I regenerated all maps, but now I get error  ""Encoded values do not match:"" 
image

problem is that last shift `24` (config) is not equal to `27` (27).

I do not understand the error and what I should do to fix it. Looks like that check is too strict.

Android version: 8.1.0
Java version to generate routing data: 13.0.1+9-jre
gh config to build routing data: https://github.com/develar/gh-routing-data/blob/c0fe3f5a396c5a03a47bd8268c2691460291f56c/internal/builder/builder.go#L158

Android java code to create is very simple  (and it worked for 0.11):

```
GraphHopper gh = new GraphHopper().forMobile();
gh.setAllowWrites(false);
gh.setCHEnabled(true);
gh.setElevation(true);
boolean load = gh.load(routingItem.getAbsolutePath());
```

karussell:
fixes #1862 

Done:

 - RouteNetwork.EV_SUFFIX is replaced via the simpler RouteNetwork.key() method that enforces the parameter. Same for TurnCost.EV_SUFFIX.
 - Made bike_network a shared EncodedValue (before it was in `bike` namespace and associated to a specific bike encoder which does not make much sense as it is valid for all BikeFlagEncoders)
 - force separating shared encoded values (come first) and encoded values of FlagEncoders (later)
 - default config.yml is now loadable via `new GraphHopper().load(graph-cache)`
 - bug fix in PathTest that used the wrong encoder
 - bug fix in EnumEncodedValue that calculated the size wrong and wasted a bit space


core/src/test/java/com/graphhopper/storage/GraphHopperStorageCHTest.java
```diff
@@ -61,8 +61,10 @@ private GraphHopperStorage newGHStorage(boolean is3D, boolean forEdgeBasedTraver
     private GraphHopperStorage newGHStorage(Directory dir, boolean is3D, boolean forEdgeBasedTraversal, int segmentSize) {
         GraphHopperStorage graph = new GraphBuilder(encodingManager)
                 .setDir(dir).set3D(is3D).withTurnCosts(true).setSegmentSize(segmentSize).build();
-        FastestWeighting weighting = new FastestWeighting(carEncoder, new DefaultTurnCostProvider(carEncoder, graph.getTurnCostStorage()));
-        graph.addCHGraph(new CHProfile(weighting, forEdgeBasedTraversal));
+        for (FlagEncoder encoder : encodingManager.fetchEdgeEncoders()) {
+            FastestWeighting weighting = new FastestWeighting(encoder, new DefaultTurnCostProvider(encoder, graph.getTurnCostStorage()));
+            graph.addCHGraph(new CHProfile(weighting, forEdgeBasedTraversal));
+        }
```
karussell: @easbar can you have a look into this test. I'm getting:

```
java.lang.IllegalStateException: Configured CH profile: fastest|car|edge_based=false is not contained in loaded weightings for CH[fastest|car|edge_based=false|u_turn_costs=-1, fastest|bike|edge_based=false|u_turn_costs=-1].
You configured: [fastest|car|edge_based=false, fastest|bike|edge_based=false]
```

Is this due to the compatibility layer? Should we remove this :) ?

And I also noticed that Objects.hash is used in CHProfile which is not available in Java 7 but replacing this should be simple.

karussell: Ah, thanks a lot! This way I learn the new stuff pretty fast :)

easbar: > And cannot make it working. IMO both profiles are identical.

I think I fixed this with the last commit. The problem was just an inconsistent way of how the weightings for CH profiles were created for the graphs. 

> Is this due to the compatibility layer? Should we remove this :) ?

No this was not because of the compatibility layer. The problem was rather that in one case the weighting had a no turn cost provider and in the other a default turn cost provider.

easbar: This error would have been caught by this check: https://github.com/graphhopper/graphhopper/pull/1863#discussion_r371029426 (had we not removed it...) 

But if we really want to improve something here I think we should refactor these graph storage tests, I am often spending time figuring out which of the createGHStorage methods is actually being called, because they are overriding and calling each other, which can be very confusing.

karussell: > But if we really want to improve something here 

Yeah, I noticed this too. The number of methods is just too much there.

> This error would have been caught by this check: #1863 (comment) (had we not removed it...)

Haha :) ... maybe we reintroduce it when time shows this is ugly.

easbar:
> default config.yml is now loadable via new GraphHopper().load(loc)

What does that mean? I can do `new GraphHopper().load("/path/to/config.yml")`?

karussell:
It means that when you do an import with the default config.yml (or with CH disabled) the Java code `new GraphHopper().load("/path/to/graph-cache")` loads without problems, see #1862. I.e. our Android code still works.

There are two 'fighting' things:
 * we automatically add e.g. the encoded value `get_off_bike` if there is any BikeFlagEncoder
 * we store shared encoded values like `max_speed` first, then flag encoders with their namespaced encoded values like `bike.access`

The error happened only for car+bike as encoders: before this PR the "automatic addition" was done in the while loop of the flag encoders i.e. on import `get_off_bike` was added after `car.access` and `car.average_speed` but on `load` we added it before it leading to a different configuration.

Of course it is possible to load the encoded values 'as it', i.e. with all the known shift and bits values, and then we could load arbitrary configured GraphHopper installations, but this is for another issue and not that easy. (And the plan is to get rid of the FlagEncoders before)

develar:
Just want to say thanks again for a quick fix — as part of https://github.com/asamm/locus-addon-graphhopper/pull/6 I removed bunch of code in application and it works.

PR #1877, opened 2020-01-28
Add a button to zoomToFit the path (as sometimes the zoom is disabled)

twitwi:
This commit adds a button to the web UI, to fit the zoom to the track. This may become useful given that now adding an intermediate point does not automatically rezoom.

This uses the suggestion from https://github.com/graphhopper/graphhopper/issues/775

I was not sure about the icon so I put the same as the fullscreen but with a magnifying glass (character) as overlay.

firefox // chromium
image // image

### DIFF
--- a/web/src/main/resources/assets/css/style.css
+++ b/web/src/main/resources/assets/css/style.css
@@ -462,10 +462,12 @@ td img.pic {
     cursor: pointer;
 }
 
-.fullscreen-reverse-btn, .fullscreen-btn {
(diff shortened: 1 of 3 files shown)


karussell:
Thanks for this. This is certainly useful, but the aim for GraphHopper Maps is really to provide a simple & clear UI. For that zoom-fit feature I think you can always quickly some out or in. 

So I would like to avoid any further space used until we introduce some "expert mode" (that could be interesting for many other features we support like path details, round trips, isochrone, map matching etc). Another possibility is a complete rewrite but this is for another (bigger) issue ;)

PR #1879, opened 2020-01-28
Change GraphHopper.init(CmdArgs)` to `init(GraphHopperConfig)`

easbar:
The goal of this PR is to change `config.yml` from

```yaml
graphhopper:
  graph.flag_encoders: car
  prepare.ch.weightings: fastest
  prepare.lm.weightings: shortest
```

to

```yaml
graphhopper:
  graph.flag_encoders: car
  prepare.ch.profiles:
    - vehicle: car
      weighting: fastest
  prepare.ch.profiles:
    - vehicle: car
      weighting: shortest
```

To understand why this is useful, see this issue: #493

What makes this complicated is that so far we deserialized `config.yml` into a `Map<String,String>` (a `CmdArgs` object to be more specific). However, here the config options are no longer key-value pairs, but we have two lists of objects (one for CH and one for LM profiles). Therefore we need to deserialize `config.yml` into a custom object (I have called this `GraphHopperConfig` here). Why is this a problem? Because so far we also allowed configuring `GraphHopper` using command line arguments (system properties), which are just key-value pairs. We even merged `config.yml` with the command line arguments (easy to do because both were a list of key-value pairs).

Here I tried to keep the changes to a minimum: I am extracting the two profile lists from the config file and put everything else into a `CmdArgs` object (just as we did before). This way we can no longer use the command line arguments to overwrite the CH/LM profiles in `config.yml`, but all other properties can still be set this way.

One special case where we do the entire configuration via the command line is in `Measurement`, e.g. here:

https://github.com/graphhopper/graphhopper/blob/2e55c309e3f661c2f1306469bc571382ce561ae5/benchmark/benchmark.sh#L38-L55

but we could either use a `config.yml` file here as well and/or somehow implement some command line arguments that allow adding CH/LM profiles via the command line for the most common cases.

**Update**: For now this PR only changes the `CmdArgs` parameter of the `init` method into a `GraphHopperConfig` parameter which simply wraps `CmdArgs`


easbar:
Any (better) ideas how to deal with this? @karussell, @michaz? Whats of course a bit ugly here is that we are handling different config options in different ways (ch/lm profiles vs. the others). On the other hand so far I could not come up with a better way to specify the profiles in the key-value format and turning the whole config into a dedicated Java object (instead of using the generic `Map<String,String>`) has its own problems: Merging with command line arguments would be harder (but doable) and the configuration would be less flexible/hard-coded. Maybe this would even be an advantage (easier for validation & we could put all the default values into the same place), but there would have to be some way to extend the (standard) config options (at least as long as `GraphHopper` is meant to be extended)

Anyway this probably leads to far for now, I am just wondering is it acceptable to extract the two profile lists for now (needed for #1776)?

karussell:
Just an idea: to avoid the extra classes around GraphHopperConfig for now is the other (strange) format we already use in other places:

```yml
  prepare.ch.profiles: vehicle=car|weighting=fastest,vehicle=car|weighting=shortest
```

> The goal of this PR is to change config.yml from

(here you mean "to" instead of "from"?)

easbar:
> (here you mean "to" instead of "from"?)

Yes -> fixed

> prepare.ch.profiles: vehicle=car|weighting=fastest,vehicle=car|weighting=shortest

Yes we *could* do this as well. It would be a bit longer with edge_based=true/false and u_turn_costs, but it could be done (even though not very pretty). The point where this would not be possible any longer is when we want to have this overwriting mechanism for (parts of) the custom models, as discussed here #493. Then again so far we have not decided if we really want this or not...
With your suggestion (compared to the current master) we could at least do:

```
prepare.ch.profiles: vehicle=my_car|weighting=custom,vehicle=car|weighting=fastest
```

Then again avoiding some pojo classes does not seem to be a very good argument to use an ugly string format like this to me (but at least we could stay with the key-value format in the config for now).

michaz:
We absolutely need Dropwizard-style hierarchical configuration, sooner or later. I totally support starting the migration work for this case (lists of profiles).

easbar:
Ok, then we'll do this. Some questions related to this:

* where should we put the config objects? currently we put jackson stuff into `graphhopper-api` (e.g. `CmdArgsDeserializer`). So we would have to put the configuration classes also into `graphhopper-api` (like `GraphHopperConfig` and `CHProfileConfig` here). In this case we do not have access to core, for example we cannot use constants like `Weighting.INFINITE_U_TURN_COSTS` in `GraphHopperConfig`. So far I do not think this is too much of an issue and can be worked around, but at the same time I do not fully understand why we are trying to keep the jackson dependency out of the core?

* What about default values? I am wondering should we keep default values for everything within the config objects such that we could do `GraphHopper.init(new GraphHopperConfig())` to start with the default configuration? The defaults would probably be 'duplicated' in `config-example.yml`, but how else would it be possible to use defaults when using the `config.yml` configuration *and* the Java API. Putting default values into the config objects is better than putting them inside `GraphHopper` and related classes. Here I am thinking about the direction we are going not really what to do right now, unless we decide to do this switch all at once instead of starting step by step like in this PR. In this PR for example the question would be if `GraphHopperConfig#chProfiles` should be filled with a single default profile by default.

karussell:
> I do not fully understand why we are trying to keep the jackson dependency out of the core?

Because of Android we want to keep the dependencies down to a minimum (apps should be small) and also often there they use GSON for deserialization. On the other hand I'm not sure if "core" really should do (de)serialization or if it should only operate on POJOs or models. This requirement has led to some ugly workarounds but keeping only the necessary inside core seems like a good constraint (no server, no json).

> The defaults would probably be 'duplicated' in config-example.yml
> ..
> GraphHopperConfig#chProfiles should be filled with a single default profile by default.

Yes, either duplication or can think about changing the default to "no CH profiles" on the Java side and provide an example config with a preconfigured car profile. The only major work could be to change all the unit/integration tests then

devemux86:
> Because of Android we want to keep the dependencies down to a minimum (apps should be small) and also often there they use GSON for deserialization.

Indeed and even JTS was not needed for the regular routing until recently(?). :slightly_smiling_face:

easbar:
> Because of Android we want to keep the dependencies down to a minimum
Ok

> Yes, either duplication or can think about changing the default to "no CH profiles" on the Java side and provide an example config with a preconfigured car profile. 

Yes better no CH profiles by default, because `GraphHopper` can also be configured without calling `init` and then its ugly because there is no `EncodingManager` initially (we discussed this).
Still I think when have more fields in `GraphHopperConfig` this would be a good place to put default values (for other fields that actually have a more interesting default value than 'empty list')

michaz:
> Indeed and even JTS was not needed for the regular routing until recently(?).

Yeah. On the other hand, not using libraries has costs, too. Even modularizing has costs. If we really want to keep JTS out of core, but _also_ don't want to own custom implementations of geometry operations, we would need to provide a core that doesn't do geometry operations. I.e. factor spatial rules and such out of the core.

It's all possible, but costly.

It's a trade-off.

michaz:
Acutally, I like that idea. Does the graph import happen on a desktop machine, @devemux86 ? Could it?

devemux86:
> Does the graph import happen on a desktop machine, @devemux86?

On desktop (e.g. via desktop [Cruiser](https://wiki.openstreetmap.org/wiki/Cruiser), as users prefer UI than scripts :slightly_smiling_face:).

Though in GH forum there were cases for imports on mobile phones too?

devemux86:
> Even modularizing has costs.

I understand that, as I maintain already multiple modules in [VTM](https://github.com/mapsforge/vtm) to avoid loading core map module with many dependencies. :)

> we could move everything import-related, and probably the geometry processing, into `web`

It would be interesting if import could exist separately and then readers (android, desktop, web, ...) use the graphs on each platform.

michaz:
> On desktop

Good to know, thanks! 🤔

devemux86:
> Does the graph import happen on a desktop machine, @devemux86?

To clarify, I mean the graph creation happens on desktop, because of the high cpu / ram specifications.

The graph reading can happen anywhere, even on mobile phones with the much limited heap memory.

michaz:
Yep, I see!

michaz:
Oops, hijacked the ticket.

michaz:
I think my problem so far (you remember I struggled with finding the right split of modules previously..) was that ``core`` wanted to be "everything except the web server and the data readers".

If it could be different, and ``core`` would be "the minimal set of things you need, as a Java programmer, to solve routing problems on a pre-generated graph", I would know much better how to move...

easbar:
I removed the LM/CH profile lists and for now this PR only wraps the `CmdArgs` we pass into `GraphHopper.init()` with a new `GraphHopperConfig` class object. So nothing changes really, but we will be able to use a hierarchical GH config file this way, by adjusting the `GraphHopperConfigDeserializer`.

easbar:
@michaz can you have a quick look regarding the Dropwizard and Gtfs stuff especially?

api/src/main/java/com/graphhopper/util/CmdArgs.java
```diff
@@ -1,89 +0,0 @@
...
-        }
-        return cmdArgs;
-    }
-
-    @Override
-    public CmdArgs put(String key, Object str) {
-        super.put(key, str);
-        return this;
-    }
-}
```
karussell: Ah, nice!

api/src/main/java/com/graphhopper/GraphHopperConfig.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("properties:\n");
+        for (Map.Entry<String, String> entry : map.toMap().entrySet()) {
+            sb.append(entry.getKey()).append(": ").append(entry.getValue());
+            sb.append("\n");
+        }
+        return sb.toString();
```
karussell: This is now a bit too ugly as without indent and looks like it should be debugging info IMO:

```
2020-02-14 19:47:56.845 [main] INFO  c.g... properties:
graph.dataaccess: RAM_STORE
ch.enabled: false
graph.flag_encoders: car
prepare.ch.weightings: fastest
prepare.min_one_way_network_size: 0
graph.location: graph-cache
prepare.min_network_size: 0
routing.ch.disabling_allowed: true
routing.max_visited_nodes: 1500000
prepare.lm.weightings: fastest
routing.round_trip.max_retries: 2
```

easbar: Yes `toString` is for debugging or what do you mean? The output should be indented another way so it resembles the yaml file?

karussell: Currently we print it when it starts here: https://github.com/graphhopper/matrix/blob/master/matrix/src/main/java/com/graphhopper/matrix/http/GraphHopperEEManaged.java#L95

Ah, maybe I just change it to:

`logger.info(ghConfig.asPMap().toString());`

?

easbar: We can adjust `toString()` to whatever you/we think is useful. What do you not like about the current format? Do you think its too verbose (because of the new lines), or would you prefer `key=value` over `key: value`? Note that as soon as we add profiles to the config (or other typed parameters `asPMap().toString()` will no longer contain the whole information.

karussell: Yes, the new line is good for debugging but a bit too much for normal startup info.

> Note that as soon as we add profiles to the config (or other typed parameters asPMap().toString() will no longer contain the whole information.

Yes, for now I just need to make sure certain info is correctly picked up

easbar: Ah is it? I thought this is something quite useful at startup. But yeah feel free to change it to whatever you prefer, I do not care that much.

PR #1882, opened 2020-01-29
Improve turn instructions when leaving the roads

### LINKED ISSUE #1047: Instructions: Remove Unnecessary Continues

boldtrn:
Fixes #1047.

This fixes the two examples given in #1047 and also the new example provided by @karussell.

Let me know what you think about the results? I haven't done too much live testing yet. So feedback would be very welcome.

I also renamed the two edge variable in InstructionsOutgoingEdges again :). I hope the naming is OK? I really struggled with the filteredEdges naming.


core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
```diff
@@ -391,12 +391,12 @@ private int getTurn(EdgeIteratorState edge, int baseNode, int prevNode, int adjN
             forceInstruction = true;
         }
 
-        InstructionsOutgoingEdges outgoingEdges = new InstructionsOutgoingEdges(prevEdge, edge, encoder, maxSpeedEnc, crossingExplorer, nodeAccess, prevNode, baseNode, adjNode);
-        int nrOfPossibleTurns = outgoingEdges.nrOfAllowedOutgoingEdges();
+        InstructionsOutgoingEdges outgoingEdges = new InstructionsOutgoingEdges(prevEdge, edge, encoder, maxSpeedEnc, roadClassEnc, roadClassLinkEnc, crossingExplorer, nodeAccess, prevNode, baseNode, adjNode);
+        int nrOfPossibleTurns = outgoingEdges.nrOfAllowedTurns();
```
karussell: Can you refactor this to the normal Java style `getAllowedTurns`? (e.g. see GraphHopper.getWorkerThreads)

boldtrn: Sure, I just pushed this change.

boldtrn:
> Nice! This solves also another problem I recently stumble over when navigating with GraphHopper.

Perfect! Thanks for testing this.

karussell:
Thanks!

PR #1885, opened 2020-01-30
Gtfs ten graphml exporter

MathStPierre:
Simple graphml logger that outputs the GTFS time expanded network (e.g.: https://github.com/graphhopper/graphhopper/blob/master/reader-gtfs/pt-model.png) to a graphml file

To use just define the environement variable GH_GTFS_GRAPH_LOGGER_OUTPUT_GENERATION_PATH and set it to the path of the .graphml file to create.

### DIFF
--- a/reader-gtfs/pom.xml
+++ b/reader-gtfs/pom.xml
@@ -142,6 +142,11 @@
             <version>2.23.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.osgeo</groupId>
+            <artifactId>proj4j</artifactId>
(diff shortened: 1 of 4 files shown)


MathieuStPierre:
> @michaz is this useful for debugging purposes? Does it still work for bigger real world cases too?

@karussell this PR was created as part of my work in Rakuten Ready to understand and debug TEN graph network created by GraphHopper as I needed to optimize short path calculation latency. Since its just an export to an GraphML format I guess you could use it to export very big TEN graph network but the GraphML reader might likely have problem rendering a bit GraphML file.

PR #1888, opened 2020-02-05
BikeProfil handling bicycle=yes

rtrier:
Bike Profil - highway=pathPath and bicycle=yes should not get speed for pushing section 
see: [forum](https://discuss.graphhopper.com/t/bike-profil-highway-pathpath-and-bicycle-yes-should-not-get-speed-for-pushing-section/5148)

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -335,7 +335,7 @@ abstract public class BikeCommonFlagEncoder extends AbstractFlagEncoder {
         // Under certain conditions we need to increase the speed of pushing sections to the speed of a "highway=cycleway"
         if (way.hasTag("highway", pushingSectionsHighways)
                 && ((way.hasTag("foot", "yes") && way.hasTag("segregated", "yes"))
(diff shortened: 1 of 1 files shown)


karussell:
Thanks! Can you add a unit test that fails before the fix and then passes?

Also can you send us [your email or the signed CLA](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md#license-agreement)?

rtrier:
The unit tests Bike2WeightFlagEncoderTest and BikeFlagEncoderTest have to be adjusted. I'am not sure this is desired.

The tests fail if: highway=footpath and highway = path if bicycle=yes is set.

Currently, highway = footpath | path and bicycle = designated | official use the speed on bike paths.
IHMO this can be done by bicycle=yes as well.

When looking through the method, I also noticed that under certain circumstances the speed resulting from the surface is overwritten by the speed on bike paths.

Unfortunately at the moment I have not the time to go deeper in this

I would include in unit tests the case
    highway = footpath | path
    bicycle = designated yes |
    surface = *

I am also considering smoothness. It would be nice if the visibility of getSpeed​​() were protected.

Thank you very much.

karussell:
@ratrun has also posted a PR #1892. Should we close here?

PR #1891, opened 2020-02-06
Remove routing algorithm factory decorator abstraction

### LINKED ISSUE #1889: remove or improve RoutingAlgorithmFactoryDecorator
See #1823 and #1708. 

Additionally there is a bug if only LM preparation is done, then lm.disable=true/false does not work (was introduced [here](https://github.com/graphhopper/graphhopper/pull/1085/files)).

Also see big usage problems described in #1401.

easbar:
Fixes #1889. 

* The decorator abstraction is gone, `GraphHopper` is now fully aware of CH & LM (well it was before, but it no longer allows adding other routing decorators it knows nothing more than the interface about). 
* Its also no longer possible to combine different decorators (CH+LM, which was implemented, but already disabled before this PR).
* I renamed the previous `CH/LMRoutingAlgoFactoryDecorator`s to `CH/LMPreparationHandler`. 
* Disabling LM when there is only LM (no CH) preparation is working now (see corresponding test in `GraphHopperIT`) and c.f. #1889.


core/src/main/java/com/graphhopper/routing/lm/LMPreparationHandler.java
```diff
@@ -218,37 +217,38 @@ public int size() {
...
+        List<Weighting> lmWeightings = new ArrayList<>(preparations.size());
         for (final PrepareLandmarks p : preparations) {
+            lmWeightings.add(p.getWeighting());
             if (p.getWeighting().matches(map))
-                return new LMRAFactory(p, defaultAlgoFactory);
+                return new LMRAFactory(p, new RoutingAlgorithmFactorySimple());
         }
 
+        // todonow: what does that mean?
         // if the initial encoder&weighting has certain properties we could cross query it but for now avoid this
```
easbar: This comment can probably be removed, but I was not sure what 'cross query it' means here?

karussell: It means if we have a LM preparation for e.g. "fastest+car" and it should be okay to do a request also with a changed or even different weighting (for optimality the edge weights must not decrease and so not all changes are acceptable).

PR #1892, opened 2020-02-06
highway=path and bicycle=yes should not get speed for pushing section

ratrun:
See https://discuss.graphhopper.com/t/bike-profil-highway-pathpath-and-bicycle-yes-should-not-get-speed-for-pushing-section/5148


ratrun:
There is a difference to bicycle=designated, here we expect 18km/h, which is the cycleway speed. This seems correct to me. If you feel 14km/h is too much for bicycle=yes, what would you suggest?

karussell:
> If you feel 14km/h is too much for bicycle=yes, what would you suggest?

For footway and bicycle=yes this is too much IMO. Maybe 10? Or maybe I have the wrong situation in mind. Isn't it like for that sign?

image

ratrun:
Yes, the sign is correct. I reduced the speed for this classification to 10 km/h as suggested.

core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
```diff
@@ -375,15 +376,22 @@ int getSpeed(ReaderWay way) {
...
-            // Increase speed in case of segregated
-            if (speed <= PUSHING_SECTION_SPEED && way.hasTag("segregated", "yes"))
-                speed = PUSHING_SECTION_SPEED * 2;
+                    if (way.hasTag("segregated", "yes"))
+                        speed = highwaySpeeds.get("cycleway");
+                    else
+                        if  (way.hasTag("bicycle", "yes"))
+                            speed = 10;
+                        else
+                            speed = highwaySpeeds.get("cycleway");
```
karussell: For nested if-else blocks can you add the brackets?

ratrun:
I added brackets for the nested ifs and improved the identations.

PR #1894, opened 2020-02-10
Move SpatialRuleLookup benchmark into Measurement class

otbutz:
This should fix the sporadically failing unittests.


karussell:
Thanks!

PR #1900, opened 2020-02-12
Allow using LM routing algo factory standalone

easbar:
Creating the LM routing algo factory should require neither `LMPreparationHandler` nor `PrepareLandmarks`. This PR tries to decouple the corresponding code.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
+++ b/core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
@@ -30,12 +30,12 @@ public class LMRoutingAlgorithmFactory implements RoutingAlgorithmFactory {
     private final int defaultActiveLandmarks;
     private final int numBaseNodes;
 
-    public LMRoutingAlgorithmFactory(LandmarkStorage lms, int defaultActiveLandmarks, int numBaseNodes) {
+    public LMRoutingAlgorithmFactory(LandmarkStorage lms, int defaultActiveLandmarks) {
         this.defaultAlgoFactory = new RoutingAlgorithmFactorySimple();
         this.lms = lms;
         this.defaultActiveLandmarks = defaultActiveLandmarks;
         this.prepareWeighting = lms.getWeighting();
-        this.numBaseNodes = numBaseNodes;
+        this.numBaseNodes = lms.getBaseNodes();
     }
 
     @Override
--- a/core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
+++ b/core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
@@ -1,0 +1,81 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.graphhopper.routing.lm;
+
+import com.graphhopper.routing.*;
+import com.graphhopper.routing.weighting.Weighting;
+import com.graphhopper.storage.Graph;
+import com.graphhopper.util.Parameters;
+
+public class LMRoutingAlgorithmFactory implements RoutingAlgorithmFactory {
+    private final RoutingAlgorithmFactory defaultAlgoFactory;
+    private final LandmarkStorage lms;
+    private final Weighting prepareWeighting;
+    private final int defaultActiveLandmarks;
+    private final int numBaseNodes;
+
+    public LMRoutingAlgorithmFactory(LandmarkStorage lms, int defaultActiveLandmarks, int numBaseNodes) {
+        this.defaultAlgoFactory = new RoutingAlgorithmFactorySimple();
+        this.lms = lms;
+        this.defaultActiveLandmarks = defaultActiveLandmarks;
+        this.prepareWeighting = lms.getWeighting();
+        this.numBaseNodes = numBaseNodes;
+    }
+
+    @Override
+    public RoutingAlgorithm createAlgo(Graph g, AlgorithmOptions opts) {
+        RoutingAlgorithm algo = defaultAlgoFactory.createAlgo(g, opts);
+        return getPreparedRoutingAlgorithm(g, algo, opts);
+    }
+
+    private RoutingAlgorithm getPreparedRoutingAlgorithm(Graph qGraph, RoutingAlgorithm algo, AlgorithmOptions opts) {
+        int activeLM = Math.max(1, opts.getHints().getInt(Parameters.Landmark.ACTIVE_COUNT, defaultActiveLandmarks));
+        if (algo instanceof AStar) {
+            if (!lms.isInitialized())
+                throw new IllegalStateException("Initialize landmark storage before creating algorithms");
+
+            double epsilon = opts.getHints().getDouble(Parameters.Algorithms.AStar.EPSILON, 1);
+            AStar astar = (AStar) algo;
+            astar.setApproximation(new LMApproximator(qGraph, prepareWeighting, numBaseNodes, lms, activeLM, lms.getFactor(), false).
+                    setEpsilon(epsilon));
+            return algo;
+        } else if (algo instanceof AStarBidirection) {
+            if (!lms.isInitialized())
+                throw new IllegalStateException("Initialize landmark storage before creating algorithms");
+
+            double epsilon = opts.getHints().getDouble(Parameters.Algorithms.AStarBi.EPSILON, 1);
+            AStarBidirection astarbi = (AStarBidirection) algo;
+            astarbi.setApproximation(new LMApproximator(qGraph, prepareWeighting, numBaseNodes, lms, activeLM, lms.getFactor(), false).
+                    setEpsilon(epsilon));
+            return algo;
+        } else if (algo instanceof AlternativeRoute) {
+            if (!lms.isInitialized())
+                throw new IllegalStateException("Initialize landmark storage before creating algorithms");
+
+            double epsilon = opts.getHints().getDouble(Parameters.Algorithms.AStarBi.EPSILON, 1);
+            AlternativeRoute altRoute = (AlternativeRoute) algo;
+            altRoute.setApproximation(new LMApproximator(qGraph, prepareWeighting, numBaseNodes, lms, activeLM, lms.getFactor(), false).
+                    setEpsilon(epsilon));
+            // landmark algorithm follows good compromise between fast response and exploring 'interesting' paths so we
+            // can decrease this exploration factor further (1->dijkstra, 0.8->bidir. A*)
+            altRoute.setMaxExplorationFactor(0.6);
+        }
+        return algo;
+    }
+}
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
```diff
@@ -0,0 +1,81 @@
...
+import com.graphhopper.util.Parameters;
+
+public class LMRoutingAlgorithmFactory implements RoutingAlgorithmFactory {
+    private final RoutingAlgorithmFactory defaultAlgoFactory;
+    private final LandmarkStorage lms;
+    private final Weighting prepareWeighting;
+    private final int defaultActiveLandmarks;
+    private final int numBaseNodes;
+
+    public LMRoutingAlgorithmFactory(LandmarkStorage lms, int defaultActiveLandmarks, int numBaseNodes) {
```
easbar: The minimal requirement to create an LM routing factory should be a `LandmarkStorage`, similar as a CH routing factory needs a `CHGraph` (the pre-processed data). `defaultActiveLandmarks` is also needed to provide the default number of active landmarks which is used when `AlgoOpts` does not specify a value for this in `createAlgo()`. The `numBaseNodes` paramater is quite ugly however, maybe we can somehow store/retrieve this information from `LandmarkStorage` as well?

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
```diff
@@ -0,0 +1,81 @@
...
+        this.defaultAlgoFactory = new RoutingAlgorithmFactorySimple();
+        this.lms = lms;
+        this.defaultActiveLandmarks = defaultActiveLandmarks;
+        this.prepareWeighting = lms.getWeighting();
+        this.numBaseNodes = numBaseNodes;
+    }
+
+    @Override
+    public RoutingAlgorithm createAlgo(Graph g, AlgorithmOptions opts) {
+        RoutingAlgorithm algo = defaultAlgoFactory.createAlgo(g, opts);
```
easbar: Currently we create the 'default' algo with the `RoutingAlgorithmFactorySimple`, and afterwards we use `instanceof` to figure which algorithm we got and set the approximator. This is rather ugly, maybe we can simply create the algorithm in a single method (without the defaultAlgoFactory?)

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
```diff
@@ -0,0 +1,81 @@
...
+
+            double epsilon = opts.getHints().getDouble(Parameters.Algorithms.AStarBi.EPSILON, 1);
+            AlternativeRoute altRoute = (AlternativeRoute) algo;
+            altRoute.setApproximation(new LMApproximator(qGraph, prepareWeighting, numBaseNodes, lms, activeLM, lms.getFactor(), false).
+                    setEpsilon(epsilon));
+            // landmark algorithm follows good compromise between fast response and exploring 'interesting' paths so we
+            // can decrease this exploration factor further (1->dijkstra, 0.8->bidir. A*)
+            altRoute.setMaxExplorationFactor(0.6);
+        }
+        return algo;
```
easbar: Are there any cases besides `AStar`, `AStarBidirection` and `AlternativeRoute` (especially since we removed the CH+LM combination for now?). If not this code could be simplified.

core/src/main/java/com/graphhopper/routing/lm/PrepareLandmarks.java
```diff
@@ -45,17 +43,10 @@
     private final Graph graph;
     private final LandmarkStorage lms;
     private final LMProfile lmProfile;
-    private int defaultActiveLandmarks;
```
easbar: `defaultActiveLandmarks` has nothing to do with LM preparation, but its only needed when creating the algorithm, so I removed this here.

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
otbutz: And we should avoid successive parameters of the same type if possible.

core/src/main/java/com/graphhopper/routing/lm/PrepareLandmarks.java
```diff
@@ -135,40 +125,11 @@ public void doSpecificWork() {
...
-            // landmark algorithm follows good compromise between fast response and exploring 'interesting' paths so we
-            // can decrease this exploration factor further (1->dijkstra, 0.8->bidir. A*)
-            altRoute.setMaxExplorationFactor(0.6);
-        }
-
-        return algo;
+    /**
+     * Convenience method to obtain a routing algo factory from the preparation.
+     */
+    public RoutingAlgorithmFactory getRoutingAlgorithmFactory(int defaultActiveLandmarks) {
```
easbar: Having the defaultActiveLandmarks parameter here is a bit ugly. Maybe we can derive some sane default from landmark storage instead (and provide an optional setter on the routing factory or something?)

core/src/test/java/com/graphhopper/routing/lm/PrepareLandmarksTest.java
```diff
@@ -133,28 +139,28 @@ public void testLandmarkStorageAndRouting() {
...
         prepare.doWork();
 
         AStar expectedAlgo = new AStar(graph, weighting, tm);
         Path expectedPath = expectedAlgo.calcPath(41, 183);
 
+        final int activeLandmarks = 2;
+
         // landmarks with A*
-        RoutingAlgorithm oneDirAlgoWithLandmarks = prepare.getPreparedRoutingAlgorithm(graph, new AStar(graph, weighting, tm), opts);
+        RoutingAlgorithm oneDirAlgoWithLandmarks = prepare.getRoutingAlgorithmFactory(activeLandmarks).createAlgo(graph,
```
easbar: Should we keep some kind of option to create the prepared LM algorithm given an algorithm (like we had for AStar here)?

core/src/test/java/com/graphhopper/routing/lm/PrepareLandmarksTest.java
```diff
@@ -133,28 +139,28 @@ public void testLandmarkStorageAndRouting() {
...
         assertEquals(expectedPath.getWeight(), path.getWeight(), .1);
         assertEquals(expectedPath.calcNodes(), path.calcNodes());
         assertEquals(expectedAlgo.getVisitedNodes() - 133, oneDirAlgoWithLandmarks.getVisitedNodes());
 
         // landmarks with bidir A*
-        opts.getHints().put("lm.recalc_count", 50);
-        RoutingAlgorithm biDirAlgoWithLandmarks = prepare.getPreparedRoutingAlgorithm(graph,
-                new AStarBidirection(graph, weighting, tm), opts);
+        RoutingAlgorithm biDirAlgoWithLandmarks = prepare.getRoutingAlgorithmFactory(activeLandmarks).createAlgo(graph,
+                AlgorithmOptions.start().algorithm(ASTAR_BI).weighting(weighting).traversalMode(tm).hints(new PMap("lm.recalc_count=50")).build());
```
easbar: What is this `lm.recalc_count` parameter, does it still exist? I could not find it anywhere.

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private RoutingAlgorithm getPreparedRoutingAlgorithm(Graph qGraph, RoutingAlgorithm algo, AlgorithmOptions opts) {
+        int activeLM = Math.max(1, opts.getHints().getInt(Parameters.Landmark.ACTIVE_COUNT, defaultActiveLandmarks));
+        if (algo instanceof AStar) {
+            if (!lms.isInitialized())
+                throw new IllegalStateException("Initialize landmark storage before creating algorithms");
+
+            double epsilon = opts.getHints().getDouble(Parameters.Algorithms.AStar.EPSILON, 1);
+            AStar astar = (AStar) algo;
+            astar.setApproximation(new LMApproximator(qGraph, prepareWeighting, numBaseNodes, lms, activeLM, lms.getFactor(), false).
```
easbar: Do we really want the `prepareWeighting` (the one from `LandmarkStorage`) here or should it rather be `opts.getWeighting()`?

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
easbar: Ok, if the `numBaseNodes` parameter stays (which I do not hope) we can at least change the order of parameters

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
karussell: I don't think there are more possibilities.

core/src/main/java/com/graphhopper/routing/lm/PrepareLandmarks.java
karussell: Yeah, the defaultActiveLandmarks should probably guessed from the landmark storage (or be half of all landmarks but at least 12 or something)

core/src/test/java/com/graphhopper/routing/lm/PrepareLandmarksTest.java
karussell: The recalc_count was once there and with it, it was possible to reset the underlying datastructure in a special way to reuse it partly for a different target. Unfortunately I never got it properly working, so I removed it and obviously not fully. I.e. it can be removed.

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
karussell: At the moment we do not allow "cross weighting" query, i.e. we use the weighting that was used from preparation i.e. prepareWeighting is correct (should be identical to CH)

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
karussell: What do you mean here? If we have a single method like this - wouldn't we repeat the code from RoutingAlgorithmFactorySimple here again? (not sure if that is bad :) )

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
karussell: > The numBaseNodes paramater is quite ugly however, maybe we can somehow store/retrieve this information from LandmarkStorage as well?

Yes, this is easy as the GraphHopperStorage is available in LandmarkStorage

easbar: Oh right I must have missed this somehow -> removed the parameter.

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
easbar: Yes we repeat *some* of the code in `RoutingAlgorithmFactorySimple`, but really not that much and some of the code in the simple factory is 'overwritten' (for example because another approximator ist set). In my last commit I rewrote the createAlgo method without the simple factory and I like it better now. Note that its also more strict now: An error is thrown when an algorithm other than astar,astarbi,alternative_route is requested (because LM does not work with dijkstra for example). For CH its similar, it throws an error if we request non-bidir Dijkstra with CH. What do you think, does that make sense?

core/src/main/java/com/graphhopper/routing/lm/PrepareLandmarks.java
easbar: Ok I think as long as we keep this parameter it should be a property of the LM routing algorithm factory, but I removed it from the constructor parameter list and set it to a default as you suggested. It can be changed using a setter.

core/src/test/java/com/graphhopper/routing/lm/PrepareLandmarksTest.java
easbar: Ok I removed it

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
easbar: Ok I see and yes its the same for CH. Its just a bit misleading because `RoutingAlgorithmFactory#createAlgo(Graph, AlgoOpts)` allows passing in a weighting and then its ignored.

core/src/main/java/com/graphhopper/routing/lm/LMRoutingAlgorithmFactory.java
karussell: Ok, yes.

Issue #1895, opened 2020-02-10
Refactor GraphHopper class


karussell:
Requirements:

 * avoid inheritance
 * better extension possibilities (no inheritance needed, see proposal)
 * separate reader and writer to easier support two scenarios of 1. create and query and 2. load and query
 * allow to add CH/LM profiles even after import (no need )
 * simpler LM/CHPreparationHandler
 * view storage not as cache but more as database and also be able to load config from this database?

First proposal: https://github.com/graphhopper/graphhopper/compare/experiment_gh_class

easbar:
> separate reader and writer to easier support two scenarios of 1. create and query and 2. load and query

I am not sure if this is the optimal separation. I think what should be separated from each other is creating/loading the graph&preparations and using the graph (routing). For example `LM/CHPreparationHandler` mixes both: Its partly about creating the preparations (reading `prepare.ch/lm.weightings` configuration, creating `CH/LMProfile`, creating the `PrepareLandmark/ContractionHierarchies` classes and even doing the actual preparation (using thread executioner etc.). And then they are used also to create the algorithms which is something different imo. In many cases you probably do the import/preparation once and from this point you are just loading the graph and use it to do the routing. In a Desktop/mobile setup you might even do this on two different devices (import server and mobile app). So in my opinion this is where we should start splitting the code into two different parts. Starting with `PrepareLandmark/PrepareContractionHierarchies` and `LM/CHPreparationHandler` and then extracting a class from `GraphHopper` that receives the loaded graph+preparations and just does the routing. I started doing this here #1901

PR #1901, opened 2020-02-12
Extract GraphHopperRouter from GraphHopper

easbar:
Work towards #1895. 

The goal here is extracting everything that is related to processing routing requests from `GraphHopper` so it can be used (and tested) standalone. The idea is that we first need to load the GH data (`GraphHopperStorage` mostly, but also `LocationIndex` and `LandmarkStorage` which are currently not managed by `GraphHopperStorage`). But this loading has nothing to do with the actual routing. The new `GraphHopperRouter` here instead *uses* the loaded data and only deals with performing the routing. There are a few todos left in the code, but I think you already get the idea.


core/src/main/java/com/graphhopper/routing/GraphHopperRouter.java
```diff
@@ -0,0 +1,322 @@
...
+        chEnabled = chPreparationHandler.isEnabled();
+        lmEnabled = lmPreparationHandler.isEnabled();
+        // todonow: maybe these could go into routing config -> no need for lm/ch prep handlers here anymore
+        chDisablingAllowed = chPreparationHandler.isDisablingAllowed();
+        lmDisablingAllowed = lmPreparationHandler.isDisablingAllowed();
+    }
+
+    // todonow: these setters are not thread-safe, do we care? maybe use builder pattern instead? or just volatile?
+    // at least document thread safety of this class, also there is lock mechanism in GraphHopper but what if this class
+    // is used standalone?
```
karussell: Maybe we get rid of the change graph endpoint and its lock mechnism for now to reduce complexity?

My plan was also for #1280 to reduce the number of possibilities to modify the base graph and it seems that spatial rules can replace change graph endpoint in combination with CustomWeighting.

easbar: Ser getting rid of this lock mechanism would certainly make things a bit simpler...

karussell: Yes, the change graph endpoint was supposed to do only smaller changes and increases the weight only so that it would work with LM at least, but yeah, it is not really battle tested and probably also not used.

easbar: So we remove it for 1.0?

karussell: Ok, #1931

easbar:
I updated this to current master. I also extracted a `WeightingFactory` here: b11f5efbc2d2223d6b5893d351580ba93a16e1ef, because we need a thing that creates weightings not only when handling routing requests, but also during preparation (which is still handled in `GraphHopper`) (and map-matching and isochrones as well...).

Extracting the `GraphHopperRouter` is certainly not the end of this 'refactoring' story, but its a good first step (depending on the direction we want to be going). Shall we already merge this?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1036,211 +1001,20 @@ public GHResponse route(GHRequest request) {
      * This method calculates the alternative path list using the low level Path objects.
      */
     public List<Path> calcPaths(GHRequest request, GHResponse ghRsp) {
+        return createGraphHopperRouter().route(request, ghRsp);
```
easbar, line 1004: I basically took this code and put it into a new class...

core/src/main/java/com/graphhopper/routing/GraphHopperRouter.java
```diff
@@ -0,0 +1,312 @@
...
+    private final GraphHopperStorage ghStorage;
+    private final EncodingManager encodingManager;
+    private final LocationIndex locationIndex;
+    private final Map<String, Profile> profilesByName;
+    private final PathDetailsBuilderFactory pathDetailsBuilderFactory;
+    private final TranslationMap translationMap;
+    private final RoutingConfig routingConfig;
+    private final WeightingFactory weightingFactory;
+    private final CHPreparationHandler chPreparationHandler;
+    private final LMPreparationHandler lmPreparationHandler;
```
easbar, line 70: Similar to `GraphHopper` these `PreparationHandlers` are concerned with both the 'import' and the 'algorithm creation' (when routing), but they are mostly about preparation and I definitely want to get rid of them here.

core/src/main/resources/com/graphhopper/util/zh_TW.txt
```diff
@@ -1,82 +1,82 @@
...
+paved=路面有鋪設
+unpaved=路面無鋪設
+stopover=中途點 %1$s
+roundabout_enter=進入圓環
+roundabout_exit=於 %1$s 個出口離開圓環
+roundabout_exit_onto=於 %1$s 個出口離開圓環，進入 %2$s
+total_ascend=總共上昇 %1$s
+total_descend=總共下降 %1$s
+way_contains_ford=路徑中含有淺灘
+way_contains_ferry=搭乘渡輪
```
easbar, line 60: Ugh looks like this file had different line endings...

karussell: I think the problem is in our core/files/update-translations.sh file that grabs the file from google. And likely because zh_TW.txt is the last language the line endings of the original file are included.

easbar: Ok I reverted the file its not related here

core/src/main/java/com/graphhopper/routing/RoutingConfig.java
```diff
@@ -0,0 +1,109 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.graphhopper.routing;
+
+/**
+ * This class contains various parameters that control the behavior of {@link GraphHopperRouter}.
+ */
+public class RoutingConfig {
```
easbar, line 24: Should we rename this to `RouterConfig`, because it configures the behavior of `GraphHopperRouter`?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1036,211 +974,33 @@ public GHResponse route(GHRequest request) {
...
-    private List<String> getTurnCostProfiles() {
-        List<String> turnCostProfiles = new ArrayList<>();
-        for (Profile p : profilesByName.values()) {
-            if (p.isTurnCosts()) {
-                turnCostProfiles.add(p.getName());
-            }
+        Map<String, CHGraph> chGraphs = new LinkedHashMap<>();
+        for (CHProfile chProfile : chPreparationHandler.getCHProfiles()) {
+            String chGraphName = chPreparationHandler.getPreparation(chProfile.getProfile()).getCHConfig().getName();
+            chGraphs.put(chProfile.getProfile(), ghStorage.getCHGraph(chGraphName));
```
easbar, line 991: See my comment about the `GraphHopperRouter` constructor.

core/src/main/java/com/graphhopper/routing/GraphHopperRouter.java
```diff
@@ -0,0 +1,320 @@
...
+    private final LocationIndex locationIndex;
+    private final Map<String, Profile> profilesByName;
+    private final PathDetailsBuilderFactory pathDetailsBuilderFactory;
+    private final TranslationMap translationMap;
+    private final RoutingConfig routingConfig;
+    private final WeightingFactory weightingFactory;
+    // todo: these should not be necessary anymore as soon as GraphHopperStorage (or something that replaces) it acts
+    // like a 'graph database'
+    private final Map<String, CHGraph> chGraphs;
+    private final Map<String, LandmarkStorage> landmarks;
```
easbar, line 68: `GraphHopperRouter` needs a way to obtain the CH/LM helper structures for a given profile. Once we have it, some kind of 'graph-database' could replace this, but for now I am simply using these maps.

core/src/main/java/com/graphhopper/routing/lm/LMPreparationHandler.java
```diff
@@ -93,15 +85,6 @@ public int getLandmarks() {
...
     }
 
-    public LMPreparationHandler setDisablingAllowed(boolean disablingAllowed) {
-        this.disablingAllowed = disablingAllowed;
-        return this;
-    }
-
-    public final boolean isDisablingAllowed() {
-        return disablingAllowed || !isEnabled();
-    }
```
easbar, line 103: This was a bit weird. The default for LM disabling allowed was false, but still this returned true whenever there were some LM profiles. -> I changed the default for `lmDisablingAllowed` to true in `RouterConfig` now. Is this ok?

tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -242,7 +242,7 @@ protected DataReader importData() throws IOException {
                 System.gc();
                 if (!hopper.getCHPreparationHandler().getNodeBasedCHConfigs().isEmpty()) {
                     CHConfig chConfig = hopper.getCHPreparationHandler().getNodeBasedCHConfigs().get(0);
-                    CHGraph lg = g.getCHGraph(chConfig);
+                    CHGraph lg = g.getCHGraph(chConfig.getName());
```
easbar, line 245: I changed these, because getting a CH graph should be done by passing its name, not the entire `CHConfig` (which I want to get rid of anyway).

core/src/main/java/com/graphhopper/util/EngineWarmUp.java
```diff
@@ -59,8 +59,8 @@ private static void warmUpCHSubNetwork(GraphHopper graphHopper, int iterations)
         Random rand = new Random(0);
 
         for (int i = 0; i < iterations; i++) {
-            int startNode = rand.nextInt(graphHopper.getMaxVisitedNodes() + 1);
-            int endNode = rand.nextInt(graphHopper.getMaxVisitedNodes() + 1);
+            int startNode = rand.nextInt(graphHopper.getRoutingConfig().getMaxVisitedNodes() + 1);
```
easbar, line 62: Is this file even needed still?

karussell: See https://github.com/graphhopper/graphhopper/issues/194

It is currently not used, but we or a Java user could benefit of it when using MMAP. For now I would indeed remove it and instead suggest that the user warms up JVM and caches via some random routes.

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
```diff
@@ -79,7 +79,9 @@ public static PrepareContractionHierarchies fromGraphHopperStorage(GraphHopperSt
     }
 
     private PrepareContractionHierarchies(GraphHopperStorage ghStorage, CHConfig chConfig) {
-        this.chGraph = ghStorage.getCHGraph(chConfig);
+        chGraph = ghStorage.getCHGraph(chConfig.getName());
+        if (chGraph == null)
+            throw new IllegalArgumentException("There is no CH graph '" + chConfig.getName() + "', existing: " + ghStorage.getCHGraphNames());
```
karussell, line 84: Can we throw this inside ghStorage.getCHGraph?

easbar: Good question :) Previously there was an `IllegalStateException` inside `getCHGraph()`, but what about the case where we get *some* profile via the web UI and use this to get the CH graph? In this case we do not want a 500 error, so we would have to catch the `IllegalStateException` and convert it into a (web api friendly(ier)) `IllegalArgumentException`, but this very ugly. So we could change this into an `IllegalArgumentException` instead, but this only means that the concerns of the web layer further propagate into the core (especially if the error message contains web layer stuff). I thought returning `null` is a cleaner (layer independent) way to communicate that the corresponding `CHGraph` does not exist. I would even use a (checked) `CHGraphDoesNotExistException` to make this very explicit, but I know you don't like it and we had this discussion before :)

To summarize:

- The `IllegalStateException` used before seems very wrong to me, because it does not even allow asking `GraphHopperStorage` for a wrong `CHGraph` name.
- Using `IllegalArgumentException` is slightly better, especially because we use this quite frequently so far. But what about the error message? Its hard to make this independent from the layer the method is used in. And catching and converting an `IllegalArgumentException` also seems wrong. In this case it would be better to use a dedicated exception.
- Using `null` leaves it up to the calling code how to handle the exceptional case (this is what I decided to do here)

core/src/main/java/com/graphhopper/routing/lm/LMPreparationHandler.java
easbar: Or should we set it to `false` now? Seems a bit more simple to me to have both set to false :)

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
karussell: Ah. Thanks for these insights. So we fetch getCHGraph to know if a ch graph exists? Couldn't we do this with the profile map or when is it required?

> but I know you don't like it and we had this discussion before :)

Haha. Yeah. But every rule can have its exceptions ;)

easbar: > So we fetch getCHGraph to know if a ch graph exists?

No, we need the actual graph (here and in other places). 

> Haha. Yeah. But every rule can have its exceptions ;)

No I think there is no need for this. Its a matter of style and since we aren't using checked exceptions there is no real need to start doing it. Sorry mentioning checked/unchecked was a bit distracting here. This is more about error handling between different parts of the code base. And I guess the question was whether we should use return values like `null`, or standard exceptions (like `IllegalArgument`) or dedicated exceptions (like `CHGraphDoesNotExist`) to indicate such exceptional conditions.

core/src/main/java/com/graphhopper/routing/lm/LMPreparationHandler.java
easbar: Hm I'll leave it as it is for now we can change the default still later if we think its better.

karussell: Yes, this could work too. The (hopefully ;)) only requirement for production is that the behaviour of ch.disable=true does not depend on if there are LM profiles or not (as we have clusters with CH and LM but also with CH and without LM). An additional assumption there is that no one uses lm.disable=true and only ch.disable=true/false.

easbar: Ok I think ch.disable=true does not depend on the existence of LM profiles, but anyway for now it seems easier to not change it :)

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
karussell: > Using IllegalArgumentException is slightly better, especially because we use this quite frequently so far. But what about the error message? Its hard to make this independent from the layer the method is used in.

Are there currently different exception handling cases for `ghStorage.getCHGraph==null` where we have different messages?

(Because if there is a bug then the now unchecked cases like in createRouter will throw a NPE)

easbar: > Are there currently different exception handling cases for ghStorage.getCHGraph==null where we have different messages?

We don't know what `GraphHopperStorage` might be used for so we cannot know what a good error message would be. Actually I think we do not want to know this because `GraphHopperStorage` is not meant to be used for one particular use-case. This is kind of my point here, it should be up to the caller of the method what to do if the CH graph does not exist. In case the CH graph does not exist and we use the result we get an NPE yes, but before we got an `IllegalStateException` (I do not think either one is better). Using `null` forces the caller to handle it unless she is sure this case cannot happen (compared to a checked exceptions where we also have to handle it (and the compiler tells us) if we are sure we do not need to. Ok, the NPE will not produce a useful error message, so this would be an advantage of using a dedicated exception.

easbar: > like in createRouter

In createRouter there cannot be an NPE, because we use a profile name that we got from GraphHopperStorage (but yes this is hard to see and not so easy to reason about).

easbar: Ah but maybe we are talking about two different things... Have a look at `Router#getAlgorithmFactory`

```java
   CHGraph chGraph = chGraphs.get(profile);
            if (chGraph == null)
                throw new IllegalArgumentException("Cannot find CH preparation for the requested profile: '" + profile + "'" +
                        "\nYou can try disabling CH using " + Parameters.CH.DISABLE + "=true" +
                        "\navailable CH profiles: " + chGraphs.keySet());
```

Here we get a profile name from the user and throw an error if this profile name does not work with CH. The same error message would not make sense in `GraphHopperStorage#getCHGraph`. But to make this even more complicated: This code no longer even takes the CH graph from `GraphHopperStorage` (but it did before), because now I have this `Map<String, CHGraph>` (see the comment in the `Router` constructor if you wonder why).

easbar: But I can see why you want to prevent NPE, so my next suggestion would be an (unchecked) `CHGraphDoesNotExistException`.

karussell: > So I could also revert the change in getCHGraph and throw an exception there again

ok

> But I can see why you want to prevent NPE, so my next suggestion would be an (unchecked) CHGraphDoesNotExistException

Let's keep it like it is :) (returning null)

PR #1902, opened 2020-02-12
Trying to refactor GraphHopper class

easbar:
See #1895. 

I opened this draft PR so its easier to comment the code.

### DIFF
--- a/core/src/main/java/com/graphhopper/CHPreparationHandler.java
+++ b/core/src/main/java/com/graphhopper/CHPreparationHandler.java
@@ -1,22 +1,77 @@
 package com.graphhopper;
 
 import com.graphhopper.routing.RoutingAlgorithmFactory;
+import com.graphhopper.routing.ch.CHProfileSelector;
+import com.graphhopper.routing.ch.CHRoutingAlgorithmFactory;
+import com.graphhopper.routing.ch.PrepareContractionHierarchies;
 import com.graphhopper.routing.util.HintsMap;
+import com.graphhopper.storage.CHGraphImpl;
 import com.graphhopper.storage.CHProfile;
+import com.graphhopper.storage.GraphHopperStorage;
+
+import java.util.ArrayList;
+import java.util.List;
 
 public class CHPreparationHandler {
-    // see #1891
-    public void addPreparation(CHProfile profile, CHProfileConfig config) {
+    private final GraphHopperStorage graphHopperStorage;
+    private final List<CHGraphImpl> chGraphs = new ArrayList<>();
+    private final List<CHProfile> chProfiles = new ArrayList<>();
+    private final List<PrepareContractionHierarchies> preparations = new ArrayList<>();
+    private boolean enabled = true;
+
+    public CHPreparationHandler(GraphHopperStorage graphHopperStorage) {
+        this.graphHopperStorage = graphHopperStorage;
     }
 
-    public void prepare() {
+    public CHPreparationHandler(GraphHopperStorage graphHopperStorage, List<CHProfile> chProfiles) {
+        this.graphHopperStorage = graphHopperStorage;
+        this.chProfiles.addAll(chProfiles);
+        this.graphHopperStorage.addCHGraphs(chProfiles);
+    }
+
+    public void doPreparation(CHProfile chProfile, CHProfileConfig chConfig) {
+        graphHopperStorage.addCHGraph(chProfile);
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/graphhopper/CHPreparationHandler.java
```diff
@@ -0,0 +1,76 @@
...
+    public CHPreparationHandler setEnabled(boolean enabled) {
+        this.enabled = enabled;
+        return this;
+    }
+
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    public RoutingAlgorithmFactory getDecoratedAlgorithmFactory(RoutingAlgorithmFactory routingAlgorithmFactory, HintsMap map) {
```
easbar: This class is concerned with CH preparation and creating the routing factory, but this should be separated IMO, see #1901

core/src/main/java/com/graphhopper/GraphConfig.java
```diff
@@ -0,0 +1,43 @@
+package com.graphhopper;
+
+import com.graphhopper.storage.DAType;
+
+public final class GraphConfig {
```
easbar: I like the idea of grouping together related configuration parameters. I did the same with `RoutingConfig` recently, maybe we can already do this in master.

core/src/main/java/com/graphhopper/GraphHopperReader.java
```diff
@@ -0,0 +1,138 @@
...
+    // TODO NOW other customization mechanism than inheritance needed
+    public RoutingAlgorithmFactory getAlgorithmFactory(HintsMap map) {
+        RoutingAlgorithmFactory routingAlgorithmFactory = new RoutingAlgorithmFactorySimple();
+        if (chHandler.isEnabled())
+            return chHandler.getDecoratedAlgorithmFactory(routingAlgorithmFactory, map);
+
+        return routingAlgorithmFactory;
+    }
+
+    // TODO NOW other customization mechanism than inheritance needed
```
easbar: Should relatively easy using a `WeightingFactory` interface?

core/src/main/java/com/graphhopper/GraphHopperReader.java
```diff
@@ -0,0 +1,138 @@
...
+        } else if ("fastest".equalsIgnoreCase(weightingStr) || weightingStr.isEmpty()) {
+            if (encoder.supports(PriorityWeighting.class))
+                return new PriorityWeighting(encoder, hintsMap, turnCostProvider);
+
+            return new FastestWeighting(encoder, hintsMap, turnCostProvider);
+        }
+        throw new IllegalArgumentException("weighting " + weightingStr + " not supported");
+    }
+
+    public GHResponse route(GHRequest request) {
```
easbar: Same here: I think routing and loading should not be in the same class (or if its in the same class it should be a very thin class that only delegates to other components that are either concerned with just the loading or just the routing)

core/src/main/java/com/graphhopper/GraphHopperReader.java
```diff
@@ -0,0 +1,138 @@
...
+
+public final class GraphHopperReader {
+
+    private final TranslationMap trMap = new TranslationMap().doImport();
+    private final EncodingManager encodingManager;
+    private final GraphHopperStorage graphHopperStorage;
+    private final CHPreparationHandler chHandler;
+    private final LocationIndex locationIndex;
+
+    public GraphHopperReader(GraphHopperWriter ghWriter) {
```
easbar: A reader takes a writer :) hmm ;), but this seems to be a convenience constructor only?

reader-osm/src/main/java/com/graphhopper/reader/osm/GraphHopperWriterOSM.java
```diff
@@ -0,0 +1,16 @@
...
+import com.graphhopper.routing.util.EncodingManager;
+
+public class GraphHopperWriterOSM {
+    private GraphHopperWriterOSM() {
+    }
+
+    public static GraphHopperWriter create(EncodingManager encodingManager, GraphConfig graphConfig) {
+        GraphHopperWriter writer = new GraphHopperWriter(encodingManager, graphConfig);
+        OSMReader reader = new OSMReader(writer.getGraphHopperStorage());
+        return writer.setDataReader(reader);
```
easbar: I really like the idea we no longer need to use an extra `GraphHopperOSM` if we want to do an OSM import. Maybe instead of setting the data reader we could even have something like `hopper.read(osmReader)`?

PR #1903, opened 2020-02-12
Allow overlapping SpatialRules with priorities

otbutz:
This PR builds on top of https://github.com/graphhopper/graphhopper/pull/1880 and restructures the `SpatialRuleLookup` interface to return `SpatialRuleSets` which cover multiple rules.


otbutz:
@karussell any feedback so far?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupJTS.java
```diff
@@ -17,110 +17,113 @@
...
+                registeredRules.add(rule);
             }
         }
 
         index.build();
 
-        this.rules = new ArrayList<>();
-        rules.add(SpatialRule.EMPTY);
-        rules.addAll(spatialRules);
+        Collections.sort(registeredRules, RULE_COMP); // keep the spatial id stable
```
karussell: Why is sorting necessary? 

See https://github.com/graphhopper/graphhopper/pull/1911/files#diff-e313449003aba3db12b92952cd486e32L85-L87 where I had to remove it to avoid problems.

otbutz: I think we can drop it here as we're already passing in a sorted collection.

otbutz:
> Thanks! Did you do some performance tests how this affects speed if spatial rules are present? 

I'll test it with an import but the `SpatialRuleSet.EMPTY` should already avoid a lot of unnecessary object creation. The sorting overhead in `lookupRules()` shouldn't be that bad considering that the number of overlapping rules for a certain point is going to be very low.

The next step after this PR will be to store the active rule IDs in the graph properties in order to drop the Country enum. That should remove the last obstacle for external rule files :)

karussell:
> I'll test it with an import 

Thanks!

> The next step after this PR will be to store the active rule IDs in the graph properties in order to drop the Country enum. That should remove the last obstacle for external rule files :)

But why a special handling for the country? E.g. for road_class we also use the enum for storage. This sounds a bit strange but as we also want enums for a nicer/simpler Java API there is IMO no better solution.

otbutz:
> but as we also want enums for a nicer/simpler Java API there is IMO no better solution.

The user will be able to pass arbitrary rules with custom IDs later. I cannot use enums for this.

karussell:
Ok. I think we have two requirements here: one is a set of countries (Country enum) and the other are custom areas where we could use integer IDs or even the StringIndex. So even if we make custom areas possible without code we cannot remove the Country enum.

otbutz:
Is there a reason why we should treat country rules differently? We could also move them to external files later and use the same mechanism.

karussell:
Have not a strong opinion about this but "country" is a distinct set of values similar to what we have for road_class.

otbutz:
I've tested the import with a small dataset(Bavaria) and didn't really notice a difference. But i'm going to run the import using the Europe pbf just to be sure.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleSet.java
```diff
@@ -0,0 +1,101 @@
...
+        return rules;
+    }
+
+    /**
+     * @return the index of the rule with the highest priority or
+     *         <i>-1</i> if the set doesn't contain any rules
+     */
+    public int getSpatialId() {
+        return spatialId;
+    }
```
karussell: Why do we need to add 1 when storing it into the edge? Also we should state it returns "EMPTY.getSpatialId" if it doesn't contain any rules.

otbutz: The spatial ID is stored as an IntEncodedValue and therefor needs to be unsigned. As -1 is used to indicate "no rules applied" i'm adding 1 to avoid signed values. But i'm open for suggestions :wink:

IMHO the Javadoc of SpatialRuleSet.getSpatialId() is pretty clear about this:
> the index of the rule with the highest priority or -1 if the set doesn't contain any rules

karussell: Could we also use 0 for this "on rule" indicator? Like we currently seem to do:

```
spatialRuleEnc.setInt(false, edgeFlags, spatialRuleLookup.getSpatialId(rule));
```

otbutz: I think it's a matter of personal preference. -1 is a widely acknowledged default to indicate a missing element (e.g Javas List.indexOf()). The limitation to use non-negative numbers is enforced by the graph storage and should therefore be handled there.

karussell: It is certainly a limitation of the storage and also the normal return value for a search is -1, but 0 is also the "Java default" for all numbers and "no rule" can be seen as the default value. Also I'm unsure why we should introduce this change when "empty" worked as 0 before.

otbutz:
@karussell anything missing?

karussell:
Sorry for the delay and yes, time is missing ;)

Have already asked @boldtrn to try this out in real world, so we'll have some progress here within the next 3 weeks or so :)

boldtrn:
> Have not a strong opinion about this but "country" is a distinct set of values similar to what we have for road_class.

I think the biggest advantage of using ids is that it's easy to extend and customize. For example one might use countries most of the time, but in the US, we might use states, because rules can differ a lot between states (for example - I have not idea if this is true). Enums are a bit unflexible and hard to change. But with IDs I don't care, I can simply add them.

boldtrn:
I had a look at the code and I missed a test case for the overlapping, so I wrote one for `SpatialRuleLookupJTSTest` to also better understand the internals, maybe we could add it (or anything similar) to the PR?

```java
    @Test
    public void testOverlapWitPriority() {
        // Let the two polygons overlap
        Polygon deBorder = FAC.createPolygon(new Coordinate[] { new Coordinate(1, 1), new Coordinate(2, 1), new Coordinate(2, 2), new Coordinate(1, 2), new Coordinate(1, 1) });
        List<SpatialRule> spatialRules = new ArrayList<>();
        SpatialRule germany = new AbstractSpatialRule(deBorder) {
            @Override
            public String getId() {
                return "DEU";
            }
        };
        spatialRules.add(germany);

        Polygon atBorder = FAC.createPolygon(new Coordinate[] { new Coordinate(0.5, 1), new Coordinate(1.5, 1), new Coordinate(1.5, 2), new Coordinate(0.5, 2), new Coordinate(0.5, 1) });
        SpatialRule austria = new AbstractSpatialRule(atBorder) {
            @Override
            public String getId() {
                return "AUT";
            }

            @Override
            public int getPriority(){
                return 500;
            }
        };
        spatialRules.add(austria);

        // create lookup with bbox just for DEU (for space reduction)
        SpatialRuleLookupJTS lookup = new SpatialRuleLookupJTS(spatialRules, new Envelope(0, 3, 0, 3));
        SpatialRuleSet set = lookup.lookupRules(1.5, 1.25);
        assertEquals(2, set.getRules().size());
        // Since Austria is priority 500, Germany should be the first, Austria the second
        assertEquals(germany.getId(), set.getRules().get(0).getId());
        assertEquals(austria.getId(), set.getRules().get(1).getId());
        assertEquals(1, lookup.lookupRules(1.5, 0.99).getRules().size());
        assertEquals(2, lookup.lookupRules(1.5, 1.00).getRules().size());
        assertEquals(2, lookup.lookupRules(1.5, 1.50).getRules().size());
        assertEquals(1, lookup.lookupRules(1.5, 1.51).getRules().size());
    }
```

otbutz:
@boldtrn Thank you! I have added your test case and listed you as commit author.

boldtrn:
Thanks! The author wouldn't have been necessary :)

boldtrn:
I also did a manual test by changing the polygons of Germany and Austria, so that they massively overlap and checked if the priorities and overlapping is resolved correctly. The results look good to me.

Screenshot from 2020-04-14 16-57-26
Screenshot from 2020-04-14 16-57-09

otbutz:
> A minor thing I missed was to specify the priority for example through the geojson, but I think this could be easily added in the future, if that's a desired change in the future.

This is going to be one of the next steps:

* storing IDs as part of the graph properties
* defining an external rule file

boldtrn:
> This is going to be one of the next steps

I thought so :). So I think this should be good to merge.

karussell:
Thanks @otbutz and @boldtrn !

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -464,10 +464,10 @@ protected IntHashSet findBorderEdgeIds(SpatialRuleLookup ruleLookup) {
         IntHashSet inaccessible = new IntHashSet();
         while (allEdgesIterator.next()) {
             int adjNode = allEdgesIterator.getAdjNode();
-            SpatialRule ruleAdj = ruleLookup.lookupRule(nodeAccess.getLatitude(adjNode), nodeAccess.getLongitude(adjNode));
+            SpatialRule ruleAdj = ruleLookup.lookupRules(nodeAccess.getLatitude(adjNode), nodeAccess.getLongitude(adjNode)).getRules().get(0);
 
             int baseNode = allEdgesIterator.getBaseNode();
-            SpatialRule ruleBase = ruleLookup.lookupRule(nodeAccess.getLatitude(baseNode), nodeAccess.getLongitude(baseNode));
```
karussell, line 470: @otbutz if the first SpatialRule is always available should we provide a convenient method for this? Or will it be rather rare use case?

otbutz: No, it is not always available per se. The landmark use case is a bit special in that respect.

karussell: Ok, so if no rule is available the list could be empty (unlike before where you got SpatialRule.EMPTY?)

otbutz: Correct. If no rules apply, an empty List is returned.

karussell: There seems to be a problem with this code: https://github.com/graphhopper/graphhopper/issues/2020

PR #1907, opened 2020-02-13
Extend RoadClass

otbutz:
* added support for footway, pedestrian, [platform](https://wiki.openstreetmap.org/wiki/Tag:highway=platform) and [corridor](https://wiki.openstreetmap.org/wiki/Tag:highway=corridor)*
* avoid Exceptions in `RoadClass.find(String name)`

\* we could add support for other indoor routing attributes in the future: https://wiki.openstreetmap.org/wiki/Key:indoor


karussell:
Thanks!

Issue #1906, opened 2020-02-13
Declare dependency versions in the parent POM


otbutz:
I would like to add a `dependencyManagement` section to our parent pom and declare our dependency version there instead of using a pattern like this:

```xml
        <dependency>
            <groupId>log4j</groupId>
            <artifactId>log4j</artifactId>
            <version>${log4j.version}</version>
            <scope>test</scope>
        </dependency>
```

We could do the same for our plugins by using `pluginManagement`.

PR #1909, opened 2020-02-13
Start using JUnit 5

easbar:
I find it difficult to add the new `profiles` configurations (see #493) and currently the biggest issue is that some of our tests are quite a mess. For example `GraphHopperIT` is a huge collection of (not necessarily related) tests, some of which are sharing some (test-)class variables while others are self-contained. This makes it hard to adjust them to any kind of api changes in `GraphHopper`. So after struggling with this for a while I decided its time to clean up these tests and wanted to split `GraphHopperIT` into several test classes to be able to properly group together tests that are actually related. There is also `GraphHopperOSMTest` which in many cases is not really different to `GraphHopperIT`? 

Anyway, I started with the first test:

https://github.com/graphhopper/graphhopper/blob/e98d02ccbd67430a588e1f68e05021085ceba276/reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java#L91-L93

It is *supposed* to test something about instructions. But why is it using a non-CH algo (from the shared hopper instance?). Why is it using (only) AStar? If we want to test something about instructions why would we not stick to what is the default in production (CH without AStar)? Or wouldn't it be even better to run the test with different algorithms (the outcome/instructions should be the same anyway)? I think the answer for the first two questions is that this test is 7 years old, but just changing it to something more recent (like CH with Dijkstra) also seems a little sketchy (maybe its testing something specific about AStar (hard to know after all these years:)).

So I thought maybe this is a good use-case for junit 5's `@ParameterizedTest`, because it allows running the same test for mulitple algorithms with a very short notation. I briefly tried this and right now I cannot see a reason not to start using this feature, especially because we can just use it where we need it. There is no need to migrate the whole code base all at once. However, every test class should use junit 4 or 5 (not both). Then again if test classes are so big and inhomogeneous that this is a problem its time to split them into smaller, more cohesive ones anyway.

One has to pay attention not to confuse the junit 4 with the junit 5 imports in the same class, but I think this we can deal with. Its also a bit annoying that changed `assertTrue(String, boolean)` to `assertTrue(boolean, String)`, but to me that does not seem to be a good reason not to use it either.

Thanks again to @Anvoker for doing a comprehensive (a lot more than I know so far) analyis about junit 5 for GraphHopper here #1632.

### DIFF
--- a/reader-osm/pom.xml
+++ b/reader-osm/pom.xml
@@ -30,41 +30,17 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
-            <version>${slf4j.version}</version>
-        </dependency>   
+        </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
-            <version>${slf4j.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>log4j</groupId>
             <artifactId>log4j</artifactId>
-            <version>${log4j.version}</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.junit.platform</groupId>
-            <artifactId>junit-platform-launcher</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-engine</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.junit.vintage</groupId>
-            <artifactId>junit-vintage-engine</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-params</artifactId>
-            <scope>test</scope>
-        </dependency>
-
         <dependency>
             <groupId>com.graphhopper</groupId>
             <artifactId>graphhopper-core</artifactId>
@@ -73,16 +49,4 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-
-    <dependencyManagement>
-        <dependencies>
-            <dependency>
-                <groupId>org.junit</groupId>
-                <artifactId>junit-bom</artifactId>
-                <version>5.6.0</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
-        </dependencies>
-    </dependencyManagement>
 </project>
--- a/reader-osm/pom.xml
+++ b/reader-osm/pom.xml
@@ -45,17 +45,44 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <groupId>org.junit.platform</groupId>
+            <artifactId>junit-platform-launcher</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
+
         <dependency>
             <groupId>com.graphhopper</groupId>
             <artifactId>graphhopper-core</artifactId>
             <version>${project.parent.version}</version>
             <type>test-jar</type>
             <scope>test</scope>
-        </dependency>        
+        </dependency>
     </dependencies>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>5.6.0</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 </project>
--- a/reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
+++ b/reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
@@ -191,7 +191,7 @@ public class GraphHopperIT {
         if (lm) {
             tmpHopper.getLMPreparationHandler().
                     setEnabled(true).
-                    setWeightingsAsStrings(Collections.singletonList(weightCalcStr)).
+                    setLMProfileStrings(Collections.singletonList(weightCalcStr)).
                     setDisablingAllowed(true);
         }
         tmpHopper.importAndClose();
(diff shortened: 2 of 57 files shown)


reader-osm/pom.xml
```diff
@@ -45,17 +45,44 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <groupId>org.junit.platform</groupId>
```
easbar: Maybe we could put these dependencies into the parent pom?

reader-osm/pom.xml
```diff
@@ -45,17 +45,44 @@
...
             <groupId>com.graphhopper</groupId>
             <artifactId>graphhopper-core</artifactId>
             <version>${project.parent.version}</version>
             <type>test-jar</type>
             <scope>test</scope>
-        </dependency>        
+        </dependency>
     </dependencies>
+
+    <dependencyManagement>
```
easbar: Maybe this should also go into the parent pom

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -34,7 +34,9 @@
 import com.graphhopper.util.exceptions.PointDistanceExceededException;
 import com.graphhopper.util.shapes.GHPoint;
 import com.graphhopper.util.shapes.GHPoint3D;
-import org.junit.*;
+import org.junit.jupiter.api.*;
```
easbar: `jupiter` tells us we are dealing with the junit 5 dependencies here

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -59,7 +62,7 @@
     private static GraphHopper hopper;
     private final String tmpGraphFile = "target/graphhopperIT-tmp";
 
-    @BeforeClass
+    @BeforeAll
```
easbar: This has been renamed, but its easy to change :)

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -72,28 +75,29 @@ public static void beforeClass() {
...
     }
 
-    @Before
+    @BeforeEach
     public void setUp() {
         Helper.removeDir(new File(tmpGraphFile));
     }
 
-    @After
+    @AfterEach
```
easbar: Same here, just a rename

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -72,28 +75,29 @@ public static void beforeClass() {
...
     public void tearDown() {
         Helper.removeDir(new File(tmpGraphFile));
     }
 
-    @Test
-    public void testMonacoWithInstructions() {
+    @ParameterizedTest
+    @CsvSource({DIJKSTRA + ",1110", DIJKSTRA_BI + ",892", ASTAR + ",699", ASTAR_BI + ",396"})
```
easbar: This is quite nice, especially for testing with different api parameters. Each element of the array is split into separate parameters to the test method (`algo` and `expectedVisitedNodes`) and there are default converters (e.g. expectedVisitedNodes is parsed to an int here. So with these two annotations I was able to run the test I was talking about in the initial comment for different algorithms.

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -1370,7 +1374,7 @@ public void testEncoderWithTurnCostSupport_stillAllows_nodeBasedRouting() {
         GHRequest req = new GHRequest(p, q);
         req.setVehicle("foot");
         GHResponse rsp = tmpHopper.route(req);
-        assertEquals("there should not be an error, but was: " + rsp.getErrors(), 0, rsp.getErrors().size());
+        assertEquals(0, rsp.getErrors().size(), "there should not be an error, but was: " + rsp.getErrors());
```
easbar: The message shown when an assertion fails is now the last parrameter in the `assertXYZ` methods...

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -1587,6 +1591,22 @@ private void assertInstruction(Instruction instruction, String expectedName, Str
         assertEquals(expectedPoints, instruction.getPoints().size());
     }
 
+    private void assertTrue(boolean condition) {
```
easbar: I did not feel like changing all the method calls to the new format, but after adding these methods this could be done with a single click by using the inline refactoring in IntelliJ

easbar:
Regarding `GraphHopperIT` and related tests I still feel like cleaning this up and it will be a good exercise to figure out whether or not JUnit 5 works and is an improvement.

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
easbar: Its also nice to see the visited nodes counts for the different algos here :)

reader-osm/pom.xml
karussell: Ok

reader-osm/pom.xml
karussell: Yes, see also: #1906?

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
karussell: Interesting! Looks good.

karussell:
> I think the answer for the first two questions is that this test is 7 years old, but just changing it to something more recent (like CH with Dijkstra) also seems a little sketchy (maybe its testing something specific about AStar (hard to know after all these years:)).

Yeah :) 

> So I thought maybe this is a good use-case for junit 5's @ParameterizedTest, because it allows running the same test for mulitple algorithms with a very short notation. I briefly tried this and right now I cannot see a reason not to start using this feature, especially because we can just use it where we need it.

Ok

> One has to pay attention not to confuse the junit 4 with the junit 5 imports in the same class, but I think this we can deal with. Its also a bit annoying that changed assertTrue(String, boolean) to assertTrue(boolean, String), but to me that does not seem to be a good reason not to use it either.

Yes, seems a bit unnecessary.

> Thanks again to @Anvoker for doing a comprehensive (a lot more than I know so far) analyis about junit 5 for GraphHopper here #1632.

:+1:

easbar:
> Yes, seems a bit unnecessary.

Yes, but tbh after using it a few times it really feels a lot more natural :)

easbar:
> Yes, although the GraphHopperOSMTest uses artificial manually created xml and GraphHopperIT uses real (downloaded) data.

Ah yes, but are these two cases really important to separate from each other (routing on an xml file that was downloaded vs routing on an xml file we wrote ourselves)? I think we should rather focus on the different functionalities of GraphHopper, like import, location index, routing etc. For routing there even might be some sub-'genres' like instructions, alternative routes etc. In any case I think the tests are most readable when all tests in one test class are about doing something similar.

> If we decide to start junit5 we should probably start a workshop to migrate other classes/modules step by step. I would like to have a goal that we at some point get rid of junit4 or at least have the plan now ;)

Yes if we start this journey we should have the intention of replacing junit4 with junit5. Maybe start doing this for every test we touch (migrate+make sure its working) and when this becomes too repetitive may do some mass migrations (using search+replace etc.)?

easbar:
I started using the dependency management section in the parent pom in #1912, which should go before this PR

karussell:
> I think we should rather focus on the different functionalities of GraphHopper, like import, location index, routing etc. For routing there even might be some sub-'genres' like instructions, alternative routes etc. In any case I think the tests are most readable (and the easiest to maintain) when all tests in one test class are about doing something similar.

Yes, you are right. 

> and when this becomes too repetitive may do some mass migrations (using search+replace etc.) and maybe we can still pick some changes from #1632 ?

:+1:

PR #1911, opened 2020-02-14
fix country rule ordering

### LINKED ISSUE #1875: PathDetails Country, DEU / AUT switched
I just tried the new SpatilRule PR #1853

I am not yet sure if it's related to the PR or not.

If I set the config to:

```
graph.encoded_values: country
```

And request a route, the resulting PathDetails are switched. Germany is AUT and Austria is DEU.

Screenshot from 2020-01-28 14-57-19
Screenshot from 2020-01-28 14-57-14

karussell:
fixes #1875

As we currently already sort the strings https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupBuilder.java#L86

the only required change for #1875 would be to reorder the Country enum (AUT should come before DEU).

But as this is too error-prone I decided for a better & special handling when using the Country enum in combination with the SpatialRuleFactory. This is currently done in SpatialRuleLookupHelper.buildAndInjectSpatialRuleIntoGH. I have renamed this to a more Country-specific name as it is nothing generic. ~~Also `spatial_rules.borders_directory` is renamed to `country.borders_directory` to make usage more clear~~. To avoid mixing this in the future I have removed the SpatialRuleParser constructor the with enum-Country default.

### DIFF
--- a/web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
+++ b/web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
@@ -67,11 +67,11 @@ public class GraphHopperManaged implements Managed {
             graphHopper = new GraphHopperOSM(landmarkSplittingFeatureCollection).forServer();
         }
         if (!configuration.get("spatial_rules.location", "").isEmpty()) {
-            throw new RuntimeException("spatial_rules.location has been deprecated. Please use countries.borders_directory instead.");
+            throw new RuntimeException("spatial_rules.location has been deprecated. Please use spatial_rules.borders_directory instead.");
         }
-        String spatialRuleBordersDirLocation = configuration.get("countries.borders_directory", "");
+        String spatialRuleBordersDirLocation = configuration.get("spatial_rules.borders_directory", "");
         if (!spatialRuleBordersDirLocation.isEmpty()) {
-            final BBox maxBounds = BBox.parseBBoxString(configuration.get("countries.max_bbox", "-180, 180, -90, 90"));
+            final BBox maxBounds = BBox.parseBBoxString(configuration.get("spatial_rules.max_bbox", "-180, 180, -90, 90"));
             final Path bordersDirectory = Paths.get(spatialRuleBordersDirLocation);
             List<JsonFeatureCollection> jsonFeatureCollections = new ArrayList<>();
             try (DirectoryStream<Path> stream = Files.newDirectoryStream(bordersDirectory, "*.{geojson,json}")) {
--- a/web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
+++ b/web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
@@ -67,11 +67,11 @@ public class GraphHopperManaged implements Managed {
             graphHopper = new GraphHopperOSM(landmarkSplittingFeatureCollection).forServer();
         }
         if (!configuration.get("spatial_rules.location", "").isEmpty()) {
-            throw new RuntimeException("spatial_rules.location has been deprecated. Please use spatial_rules.borders_directory instead.");
+            throw new RuntimeException("spatial_rules.location has been deprecated. Please use countries.borders_directory instead.");
         }
-        String spatialRuleBordersDirLocation = configuration.get("spatial_rules.borders_directory", "");
+        String spatialRuleBordersDirLocation = configuration.get("countries.borders_directory", "");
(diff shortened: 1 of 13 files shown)


web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
```diff
@@ -67,11 +67,11 @@ public GraphHopperManaged(GraphHopperConfig configuration, ObjectMapper objectMa
...
         }
         if (!configuration.get("spatial_rules.location", "").isEmpty()) {
-            throw new RuntimeException("spatial_rules.location has been deprecated. Please use spatial_rules.borders_directory instead.");
+            throw new RuntimeException("spatial_rules.location has been deprecated. Please use countries.borders_directory instead.");
         }
-        String spatialRuleBordersDirLocation = configuration.get("spatial_rules.borders_directory", "");
+        String spatialRuleBordersDirLocation = configuration.get("countries.borders_directory", "");
         if (!spatialRuleBordersDirLocation.isEmpty()) {
-            final BBox maxBounds = BBox.parseBBoxString(configuration.get("spatial_rules.max_bbox", "-180, 180, -90, 90"));
+            final BBox maxBounds = BBox.parseBBoxString(configuration.get("countries.max_bbox", "-180, 180, -90, 90"));
```
karussell: Or better singular `country`?

otbutz:
> fixes #1875
> 
> As we currently already sort the strings https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupBuilder.java#L86
> 
> the only required change for #1875 is to reorder the Country enum (AUT should come before DEU).

Looks like a good workaround until we finish #1808

> I have renamed this to a more Country-specific name as it is nothing generic. Also `spatial_rules.borders_directory` is renamed to `country.borders_directory` to make usage more clear. 

I'd like to stick with the more generic term as this directory might contain arbitrary borders in the very near future.

karussell:
> Looks like a good workaround until we finish #1808

Sorry, I meant we could just reorder Country, but I did not do this workaround.

> I'd like to stick with the more generic term as this directory might contain arbitrary borders in the very near future

Ok. Still for a more generic usage we would have to implement a more generic approach. Currently I do not see how we can provider arbitrary rules in Java, i.e. I would stick to the "country directory" and people with custom rules can have their custom directories?

otbutz:
> Ok. Still for a more generic usage we would have to implement a more generic approach. Currently I do not see how we can provider arbitrary rules in Java, i.e. I would stick to the "country directory" and people with custom rules can have their custom directories?

My plan would be to ship our current rules as files. The user may provide further border and rule definitions.

e.g: `spatial_rules.borders_directory` + `spatial_rules.rules_directory`

karussell:
But how would the user provide Java code to implement these custom SpatialRules? Ah, you mean they specify with some properties what should be done with the borders similar to what we currently do with the ChangeGraphResource file format ("overlay.json")?

otbutz:
> Ah, you mean they specify with some properties what should be done with the borders similar to what we currently do with the ChangeGraphResource file format ("overlay.json")?

Yes. Not sure about the format but i'm leaning towards a similar solution like you're using for the custom weighting support.

karussell:
Ah, IMO with CustomWeighting this would work the opposite: we define the custom or country borders as separate GeoJSON (or externally) and then use them to avoid the area or cap max_speed: 

```yml
priority:
 # avoid a certain area
  area_custom1: 0.5

areas:
  custom1:
    type: "Feature"
    geometry: { type: "Polygon", coordinates: [[[13.722, 51.053], [...
```

otbutz:
> Ah, IMO with CustomWeighting this would work the opposite: we define the custom or country borders as separate GeoJSON (or externally)

That would be the same for the external rules concept.

> and then use them to avoid the area or cap max_speed:

External rules would either provide fallback values if the OSM material lacks sufficient tagging or could be used to overwrite existing values.

IMHO the custom weighting and the spatial rules are quite similar in that they allow to influence routing decisions for a given area. The main difference is that the user can define a lot of SpatialRules with detailed geometries without sacrificing runtime performance e.g for countries, states or environmental protection zones

karussell:
@otbutz do you think if I revert the renaming of spatial_rules.borders_directory to country.borders_directory we merge this here?

otbutz:
I'm just against the name change. The other changes look good :+1:

PR #1912, opened 2020-02-14
Add dependency management and junit4 dependency to parent pom

### LINKED ISSUE #1906: Declare dependency versions in the parent POM

easbar:
Fixes #1906. 

I moved most dependency versions into parent pom's dependency management section and move the junit 4 dependency there as well. For dropwizard I used their bom file.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,15 +15,6 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <slf4j.version>1.7.26</slf4j.version>
-        <log4j.version>1.2.17</log4j.version>
-        <commons-compress.version>1.18</commons-compress.version>
-
-        <!-- for the correct jackson and guava versions see https://github.com/dropwizard/dropwizard/blob/release/1.3.x/dropwizard-bom/pom.xml -->
-        <!-- make also sure that the dropwizard version is working with 1.3.5 of dropwizard-configurable-assets-bundle used in web -->
-        <dropwizard.version>1.3.12</dropwizard.version>
-        <jackson.version>2.9.9</jackson.version>
-        <guava.version>24.1.1-jre</guava.version>
 
         <maven.compiler.target>1.8</maven.compiler.target>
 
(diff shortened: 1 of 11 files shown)


pom.xml
```diff
@@ -15,15 +15,6 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <slf4j.version>1.7.26</slf4j.version>
-        <log4j.version>1.2.17</log4j.version>
-        <commons-compress.version>1.18</commons-compress.version>
-
-        <!-- for the correct jackson and guava versions see https://github.com/dropwizard/dropwizard/blob/release/1.3.x/dropwizard-bom/pom.xml -->
```
easbar: Is there a reason you added this link to the bom file, but did not include the bom file here? It provides the correct version for jackson, guava and many more (maybe too many and it would be more readable to specify these versions explicitly in our dependency management section instead?)

pom.xml
```diff
@@ -89,6 +80,68 @@
...
+            </dependency>
+            <dependency>
+                <groupId>log4j</groupId>
+                <artifactId>log4j</artifactId>
+                <version>1.2.17</version>
+            </dependency>
+            <dependency>
+                <groupId>org.slf4j</groupId>
+                <artifactId>slf4j-log4j12</artifactId>
+                <version>1.7.26</version>
```
easbar: unfortunately this logging bridge is not declared in the dropwizard bom, while for example slf4j-api is declared there (not sure why)

pom.xml
```diff
@@ -89,6 +80,68 @@
...
+                <groupId>junit</groupId>
+                <artifactId>junit</artifactId>
+                <version>4.12</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>junit</groupId>
```
easbar: I added this here with the downside that all child projects have this dependency automatically now. Then again junit is used by all children and its only test scope too. With junit 5 this will be multiple dependencies, which otherwise we would have to repeat in every child pom.

michaz: Interesting, I did the exact opposite a while ago. :-) I don't oppose at the moment, but let's remember that this is *only* for convenience: There ist absolutely zero dependency between the test frameworks each submodule uses (each one can use its own), so in a way this is an artificial restriction.

easbar: Yes its code re-use via inheritance in a way, use with care... If you want we can of course go back to declaring the dependency in the submodules for the reason you mention.

otbutz: >  in a way this is an artificial restriction 

I don't want to be pedantic but a submodule doesn't not inherit dependencies from a parents dependencyManagement section and is still free to override the dependency versions in its own pom.

easbar: Yes, but the junit dependency is an actual dependency (not dependencyManagement)

otbutz: > Yes, but the junit dependency is an actual dependency 

Sorry, my bad. 

> I think we all agree that using the dependencyManagement in the parent pom makes sense

Absolutely

PR #1914, opened 2020-02-16
No longer enable CH by default in GraphHopper class

easbar:
So far when creating a new `GraphHopper` instance the `CHPreparationHandler` was set to be enabled and `"fastest"` was used as default weighting, resulting in a `"fastest"` CH preparation for every vehicle. In this PR I have removed this default so there no longer are any CH preparations by default. I also removed the `LMPreparationHandler#setEnabled` methods, because there are only two cases: Either some CH/LM profiles are configured (then CH/LM is enabled) or not (then CH/LM is disabled), which means `setEnabled(false)` is equivalent to setting up no `CH/LMProfile`s and `setEnabled(true)` requires setting up the profiles and is thus unnecessary. 

The usage goes like this now:

```config.yml
graphhopper:
   # nothing changed here: adding some weightings means they will be prepared using CH
   prepare.ch.weightings=fastest
   # using an empty string or setting to 'no' means the list of weightings is empty, just as before this PR
   prepare.lm.weightings=no
```

```java
// create a new GraphHopper object. note that compared to before this PR there is no CH profile
GraphHopper hopper = new GraphHopper();
// these methods are all gone
// hopper.setCHEnabled(true);
// hopper.getCHPreparationHandler().setEnabled(true);
// hopper.getLMPreparationHandler().setEnabled(false);

// to enable CH/LM we need to add the profiles
// this resembles the previous behavior (CH will be 'enabled' by adding the profile)
hopper.getCHPreparationHandler().setCHProfileStrings("fastest");
```

### DIFF
--- a/reader-gtfs/README.md
+++ b/reader-gtfs/README.md
@@ -17,11 +17,10 @@ wget http://download.geofabrik.de/europe/germany/brandenburg-latest.osm.pbf
 # The following process will take roughly 5 minutes on a modern laptop when it is executed for the first time.
 # It imports the previously downloaded OSM data of the Brandenburg area as well as the GTFS.
 java -Xmx8g -Xms8g \
-# todonow: this is a) outdated, and b) why no CH (when in tests its enabled?)
-  -Dgraphhopper.datareader.file=brandenburg-latest.osm.pbf \
-  -Dgraphhopper.gtfs.file=gtfs-vbb.zip \
-  -Dgraphhopper.graph.flag_encoders=pt \
-  -Dgraphhopper.graph.location=./graph-cache \
+  -Ddw.graphhopper.datareader.file=brandenburg-latest.osm.pbf \
+  -Ddw.graphhopper.gtfs.file=gtfs-vbb.zip \
+  -Ddw.graphhopper.graph.flag_encoders=pt \
+  -Ddw.graphhopper.graph.location=./graph-cache \
   -jar web/target/graphhopper-web-*.jar server config.yml
 # view the web UI e.g. via:
 firefox http://localhost:8989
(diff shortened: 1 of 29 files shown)


reader-gtfs/src/test/java/com/graphhopper/ExtendedRouteTypeIT.java
```diff
@@ -48,6 +48,8 @@ public static void init() {
         ghConfig.put("graph.flag_encoders", "car,foot");
         ghConfig.put("graph.location", GRAPH_LOC);
         ghConfig.put("gtfs.file", "files/another-sample-feed-extended-route-type.zip");
+        // todonow: do we want the (previously by default) CH profiles here? (same question for all GraphHopperGtfs basically
+        ghConfig.put("prepare.ch.weightings", "fastest");
```
easbar: @michaz ? I wasn't sure if CH was used for this test only by accident (because of the default) or if its actually intended.

reader-gtfs/README.md
```diff
@@ -17,10 +17,10 @@ wget http://download.geofabrik.de/europe/germany/brandenburg-latest.osm.pbf
 # The following process will take roughly 5 minutes on a modern laptop when it is executed for the first time.
 # It imports the previously downloaded OSM data of the Brandenburg area as well as the GTFS.
 java -Xmx8g -Xms8g \
+# todonow: this is a) outdated, and b) why no CH (when in tests its enabled?)
```
easbar: @michaz ? We probably want to update the command line parameters to `-Ddw.graphhopper.xyz` (since `-Dgraphhopper.xyz` is no longer working after #1879), and I was wondering why we are setting `ch.weightings=no` here (while for some gtfs tests we were using CH) (see other comment)

reader-gtfs/src/test/java/com/graphhopper/ExtendedRouteTypeIT.java
easbar: Seems like an accident, pt does not use CH -> removed this

reader-gtfs/README.md
easbar: Updadet the readme and removed `ch.weightings=no` as this is now the default

PR #1841, opened 2019-12-30
Weighting support via yaml configuration

karussell:
fixes #1776 
fixes #223 

## Note

**This is a preliminary feature introduced in 1.0. The endpoint, the formula and the format might change for future versions.**

### Usage

First you need to define a custom profile:

```
profiles:
  - name: car
    vehicle: car
    weighting: custom
```

Then you can use this profile in a custom request. You query a json or yml towards the new endpoint `/route-custom`, e.g. send the following content:

```yml
# query parameters:
points: [[10.373926, 52.042989], [10.384043, 52.042289]]
details: [road_class, street_name]

# pick the previously defined custom profile
profile: car

# The formula is defined in CustomWeighting, where the speed_factor can be used to decrease speed.
# The minimum value is 0 and the maximum is 1. If two conditions are satisfied the values are
# multiplied e.g. if road_class==motorway and road_environment==tunnel, then the resulting speed is
# average_speed*0.85*0.9.
speed_factor:
  road_environment:
    tunnel: 0.85
  # trucks should be a bit slower on certain road classes compared to the 'base' car
  # a more compact JSON-way to list the properties, fully YAML-compatible:
  road_class: { motorway: 0.85, primary: 0.9 }

# You can lower the average speed for certain conditions via this max_speed entry. The speed_factor is applied before
# this operation. See the following example that sets the maximum speed to 95km/h for motorways.
max_speed:
  road_class:
    motorway: 95
    residential: 30
  road_environment:
    bridge: 85

# Trucks are slower so limit all speed to this value. In km/h
# If none of the conditions in the map above apply use the fallback, if specified:
max_speed_fallback: 100

# This term changes the influence of the distance. I.e. longer roads get a higher cost.
# The distance_influence is independent of the edge properties and does not influence the ETA. The default is 70 (seconds/1km).
# Let's assume a route that takes 1000sec and is 10km long, then a value of 30 means that I would like to drive maximum 11km
# to reduce the travel time to 970sec or 12km to reduce it to 940sec.
distance_influence: 90

# Now we want to set a preference without changing the taken time. The default priority is 1 and does not change something.
# Higher than 1 is not possible directly, only via the catch-all key "*", see road_class where motorways are preferred. Lower than 1 will avoid it.
# If two conditions are met the values will be multiplied. The minimum value is 0 (block access) the maximum value is 100.
priority:
  road_class:
    motorway: 1.0
    residential: 0.4
    "*": 0.9
    # and if you do not want that tracks go over tracks just exclude them via
[...]

PR #1917, opened 2020-02-17
Fix problems with block_area

### LINKED ISSUE #1324: Reimplement block_area with LocationIndexTree instead of graph traversal

### LINKED ISSUE #1905: Put back block_area for isochrone & spt
This feature got lost when the block area handling was moved out of createWeighting.

karussell:
This PR fixes #1324.

There were a couple of bugs with the block_area feature that are fixed within this PR:

 * virtual edges are handled now, see #1127
 * bread first search is replaced with locationIndex lookup, which avoids several problems documented in #1324
 * edgeId cache is properly used (return early!), which should lead to a speed up
 * multiple block areas are properly used
 * fixes #1905

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -1073,6 +1073,9 @@ public class GraphHopper implements GraphHopperAPI {
                 return Collections.emptyList();
 
             QueryGraph queryGraph = QueryGraph.lookup(graph, qResults);
+            if (weighting instanceof BlockAreaWeighting)
+                ((BlockAreaWeighting) weighting).setQueryGraph(queryGraph);
+
             int maxVisitedNodesForRequest = hints.getInt(Routing.MAX_VISITED_NODES, routingConfig.getMaxVisitedNodes());
             if (maxVisitedNodesForRequest > routingConfig.getMaxVisitedNodes())
                 throw new IllegalArgumentException("The max_visited_nodes parameter has to be below or equal to:" + routingConfig.getMaxVisitedNodes());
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1073,6 +1073,9 @@ else if (ALT_ROUTE.equalsIgnoreCase(algoStr))
                 return Collections.emptyList();
 
             QueryGraph queryGraph = QueryGraph.lookup(graph, qResults);
+            if (weighting instanceof BlockAreaWeighting)
+                ((BlockAreaWeighting) weighting).setQueryGraph(queryGraph);
```
karussell: Here we could think about a more generic solution as this is also required for the CustomWeighting in #1841.

easbar:
Nice :) Maybe while you are at it: #1905

karussell:
@easbar have done this now. Do you think it makes sense to repeat it or should I move it back into the createWeighting method?

easbar:
> Do you think it makes sense to repeat it or should I move it back into the createWeighting method?

I don't have a very strong opinion here, maybe leave it as is for now?

PR #1918, opened 2020-02-17
no need to explicitely list the snake case methods

karussell:
of GHRequest in GHRequestMixIn.

Unfortunately there seems to be no lowerCamelCase naming strategy. With such a strategy we could enable it globally and annotate only GraphHopperConfigMixIn (or should we create something like this?)

### DIFF
--- a/web-api/src/main/java/com/graphhopper/jackson/Jackson.java
+++ b/web-api/src/main/java/com/graphhopper/jackson/Jackson.java
@@ -20,11 +20,16 @@ package com.graphhopper.jackson;
 import com.bedatadriven.jackson.datatype.jts.JtsModule;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 
 public class Jackson {
 
     public static ObjectMapper newObjectMapper() {
         return initObjectMapper(new ObjectMapper());
+    }
+
+    public static ObjectMapper initObjectMapperWithSnake(ObjectMapper objectMapper) {
+        return initObjectMapper(objectMapper.setPropertyNamingStrategy(PropertyNamingStrategy.SNAKE_CASE));
     }
 
     public static ObjectMapper initObjectMapper(ObjectMapper objectMapper) {
--- a/web-api/src/main/java/com/graphhopper/jackson/Jackson.java
+++ b/web-api/src/main/java/com/graphhopper/jackson/Jackson.java
@@ -28,13 +28,10 @@ public class Jackson {
         return initObjectMapper(new ObjectMapper());
     }
 
-    public static ObjectMapper initObjectMapperWithSnake(ObjectMapper objectMapper) {
-        return initObjectMapper(objectMapper.setPropertyNamingStrategy(PropertyNamingStrategy.SNAKE_CASE));
-    }
-
     public static ObjectMapper initObjectMapper(ObjectMapper objectMapper) {
         objectMapper.registerModule(new GraphHopperModule());
         objectMapper.registerModule(new JtsModule());
+        objectMapper.setPropertyNamingStrategy(PropertyNamingStrategy.SNAKE_CASE);
         objectMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
         return objectMapper;
     }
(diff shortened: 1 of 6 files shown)


web-api/src/main/java/com/graphhopper/jackson/Jackson.java
```diff
@@ -20,13 +20,18 @@
...
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 
 public class Jackson {
 
     public static ObjectMapper newObjectMapper() {
         return initObjectMapper(new ObjectMapper());
     }
 
+    public static ObjectMapper initObjectMapperWithSnake(ObjectMapper objectMapper) {
```
easbar: Do we ever want the mapper without the snake strategy? If not maybe we could simply put this in the existing `initObjectMapper`?

karussell: If we define the snake case globally, which I would really prefer, then we would need to rewrite all our config.yml (should we avoid this?). See https://github.com/graphhopper/graphhopper/pull/1841/commits/2f2567bb5dacb31973321ef2b5f0437845af32e9

easbar: Wait what do you prefer? You want to use snake case everywhere? But isn't this what we do (the current `config.yml` uses snake case). And in https://github.com/graphhopper/graphhopper/commit/2f2567bb5dacb31973321ef2b5f0437845af32e9 you converted everything  to camel case?

karussell: I would prefer a global snake case, yes. But this would break existing config.yml files as it includes e.g. `applicationConnectors` and it would be necessary to change this (and several other things) to snake_case.

I'm not sure if we should do this.

> But isn't this what we do (the current config.yml uses snake case)

No, unfortunately not. See all the camelCase config properties from dropwizard. Also when switching to global snake case still the `requestLog` property made problems and wasn't available as snake_case, maybe something related to https://github.com/dropwizard/dropwizard/issues/1558

easbar: Ah right dropwizard uses camel case. But if thats the default why do we not just use camel case everywhere? Ok I would agree that snake case is prettier / more readable and more consistent with url parameters...

karussell: > Ok I would agree that snake case is prettier / more readable and more consistent with url parameters...

Yes, exactly. I don't want to advertise camelCase outside of Java code (also as you say for consistency with our API).

But maybe we could break configuration once again before 1.0? It is not really much but it is also a bit ugly to rewrite them. Especially the error message is not really helpful but we could create a ticket for that.

easbar: we are breaking the configuration in many ways I think (also this for example #1922) and overall it might need some cleanup. For example if we make `GraphHopperConfig` more typesafe I am not sure if the current grouping of parameters (using dots) is really what we want.

karussell: > we are breaking the configuration in many ways I think

Ah, indeed. Then I'll rework this PR to use the snake_case globally and document it in the changelog.

Also regarding the "dot": could we convert the current state: `datareader.file: somefile.pbf` into an object structure like

```
datareader:
 file: somefile.pbf
```

easbar: >  could we convert the current state: datareader.file: somefile.pbf into an object structure like

I think thats the point. When the property name is `datareader.file` (in the command line argument at least) dropwizard interprets this as the `file` property of the `datareader` object. The `datareader` object is missing in the configuration and thus the override fails. Otherwise yes we could convert it into an object structure, but this must be reflected on the java side (`GraphHopperConfig.java`). If we start grouping parameters together we should try to find out which ones are really related such that the sub-objects of `GraphHopperConfig` represent a meaningful decomposition and we can pass around these sub-objects to the different components (which ideally do not need to know anything about the other sub-objects). For example I did this with `RoutingConfig` in the `GraphHopper` class even though this is so far only done internally (not in `config.yml` yet).

easbar: But maybe this is something for GH 2.0 as its also related to how we will deal with the graphhopper folder cache/database in the future and what kind of refactorings will be done to the GraphHopper class, what should belong to the 'core' etc. :)

karussell: Ok. So we keep the dots in the pararameters for 1.0 but move everything to snake_case?

easbar: Sounds like a plan

karussell:
To make the transition a bit easier I have created https://github.com/karussell/snake_case which converts all the camelCase keys into snake_case (the only problem is that it removes the comments)

Issue #1859, opened 2020-01-22
Add new 'profile' parameter to select routing profile 


easbar:
In #493 its discussed how to specify CH profiles and part of this question is how they can be selected for incoming routing requests (especially when adding custom weightings in #1776). There are multiple cases that have to be differentiated. These are the current options that can be used on each routing request and affect the profile that will be used:

- the vehicle / flag encoder
- the weighting: either one of the 'classic' ones like "fastest" or a custom one
- turn costs yes/no
- u-turn-costs
- CH,LM or no speed up?
- blocked areas

The speed-ups can only be used if a matching profile exists on the server side, but how should this profile be selected? So far we are doing it like this:

- if CH/LM are not enabled use the vehicle/weighting/turn cost parameters to select the `FlagEncoder` and create the corresponding `Weighting`
- if CH/LM are enabled use the same information to find a matching CH/LM profile and use it 

In the future this will be more complicated, because the weighting will be configurable: There will be `CustomWeighting` in #1776, but also we might add the option to use the `ShortFastestWeighting` with an additional distance factor, a more advanced turn cost configuration or even blocked areas. In any case we can no longer require specifying all the properties of these weightings to select the profile on the server side (or at least its very unclear to me how this is supposed to work. How can we check if a requested and an available profile are 'equal' and would it not be very cumbersome to send the entire profile configuration with every request?). Instead, we will select these profiles by a unique name. To do this we will add a `profile` parameter to the routing endpoint. The question that arises is how this relates to the other parameters, because they will then be redundant and possibly conflicting. At the same time we do not want to break backwards compatibility, so for the current use-cases it should still be possible to choose the vehicle&weighting without setting the CH profile name for example.

These are the different cases:

#### No `CustomWeighting`

- if LM/CH are disabled the `vehicle/weighting/etc` parameters are used to create the weighting, just like we did in the past. the `profile` parameter cannot be used in this case (see below)

- if LM/CH are enabled and no `profile` is given the `vehicle/weighting/turn_costs/u_turn_costs` parameters are used to create a profile name following a certain naming convention, something like 
`gh_car_fastest_edge_30`. The profile is then selected using this profile name. This is equal to the current behavior except the profile name handling on the server side

[...]

PR #1922, opened 2020-02-19
Use new profiles section to configure LM&CH preparations

### LINKED ISSUE #493: Make creating CH profile more configurable

easbar:
This PR replaces the following configuration options:

```yaml
prepare.ch.weightings: # e.g. fastest,...
prepare.lm.weightings: # e.g. shortest,...
prepare.ch.edge_based: # e.g. edge_and_node
```

with three new sections:

```yaml
profiles: # used to define routing profiles
profiles_ch: # used to define which of the routing profiles shall be prepared for CH
profiles_lm: # used to define which of the routing profiles shall be prepared for LM
```

More details can be found in `config-example.yml` documentation in this PR, for completeness I am copying this here:

```yaml
##### Routing Profiles ####

  # Routing can be done for the following list of profiles. The fields of each profile are as follows:
  # - name (required): a unique string identifier for the profile
  # - vehicle (required): refers to the `graph.flag_encoder` used for this profile
  # - weighting (required): the weighting used for this profile, e.g. fastest,shortest or short_fastest
  # - turn_costs (true/false, default: false): whether or not turn restrictions should be applied for this profile.
  #   this will only work if the `graph.flag_encoder` for the given `vehicle` is configured with `|turn_costs=true`.
  #
  # Depending on the above fields there are other properties that can be used, e.g.
  # - distance_factor: 0.1 (can be used to fine tune the time/distance trade-off of short_fastest weighting)
  # - u_turn_costs: 60 (time-penalty for doing a u-turn in seconds (only possible when `turn_costs: true`)).
  #   Note that since the u-turn costs are given in seconds the weighting you use should also calculate the weight
  #   in seconds, so for example it does not work with shortest weighting.
  #
  # To prevent long running routing queries you should usually enable either speed or hybrid mode for all the given
  # profiles (see below). Otherwise you should at least limit the number of `routing.max_visited_nodes`.
  profiles:
    - name: car
      vehicle: car
      weighting: fastest

  #  - name: car_with_turn_costs
  #    vehicle: car
  #    weighting: short_fastest
  #    distance_factor: 0.1
  #    turn_costs: true
  #    u_turn_costs: 60

  # Speed mode:
  # Its possible to speed up routing by doing a special graph preparation (Contraction Hierarchies, CH). This requires
  # more RAM/disk space for holding the prepared graph but also means less memory usage per request. Using the following
  # list you can define for which of the above routing profiles such preparation shall be performed. Note that to support
  # profiles with `turn_costs: true` a more elaborate preparation is required (longer preparation time and more memory
  # usage) and the routing will also be slower than with `turn_costs: false`.
  profiles_ch:
    - profile: car
  #   - profile: car_with_turn_costs

  # Hybrid mode:
  # Similar to speed mode, the hybrid mode (Landmarks, LM) also speeds up routing by doing calculating auxiliary data
  # in advance. Its not as fast as speed mode, but more flexible.
  profiles_lm: []
```

With the new configuration its now possible to e.g. use CH for only one of multiple flag encoders (previously there was a preparation for *every* vehicle for every given weighting in `prepare.ch/lm.weightings`), i.e. this PR fixes #493.

Importantly the profiles have a (required) name field, which soon will be used to pick the profile during a routing request (see #1859). However, this PR still uses the old method (the profile is identified by comparing the weighting/vehicle/turn_costs parameters)!

This PR also does not yet allow the new custom weighting from #1776, but the profiles section in the configuration together with the (to be added) `profile` request parameter should make it easy to integrate this.

**Important Note:** This PR does not only remove the above mentioned configuration options, but there will be an error in case they are still being used, so it forces an upgrade of the config files.


easbar:
For completeness, here is an example for an LM profile (it includes the `maximum_lm_weight` property):
```yaml
profiles_lm: 
  - profile: car
    maximum_lm_weight: 10000
```

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -297,6 +298,16 @@ private GraphHopper setUnsafeMemory() {
         return this;
     }
 
+    public GraphHopper setProfiles(ProfileConfig... profiles) {
+        return setProfiles(Arrays.asList(profiles));
+    }
+
+    public GraphHopper setProfiles(List<ProfileConfig> profiles) {
```
karussell: Some javadocs would be good here, also how to set the LM/CHProfileConfig. Or link to this PR.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -772,6 +788,50 @@ public boolean load(String graphHopperFolder) {
         }
     }
 
+    private void checkProfilesConsistency() {
+        Set<String> profileSet = new LinkedHashSet<>(profiles.size());
+        for (ProfileConfig profile : profiles) {
+            boolean added = profileSet.add(profile.getName());
```
karussell: you can add the profile directly to the set as the hashcode & equals is already implemented for that (?)

Or maybe shouldn't we do the check when calling setProfileConfigs or create a 
`LinkedHashMap<String, ProfileConfig> profiles` in the GraphHopper class on setProfileConfigs? This would also avoid the method getProfilesByName.

api/src/main/java/com/graphhopper/config/ProfileConfig.java
```diff
@@ -0,0 +1,108 @@
...
+    public boolean isTurnCosts() {
+        return turnCosts;
+    }
+
+    public ProfileConfig setTurnCosts(boolean turnCosts) {
+        this.turnCosts = turnCosts;
+        return this;
+    }
+
+    public PMap hints() {
```
karussell: Why is it not named `getHints`?

api/src/main/java/com/graphhopper/config/ProfileConfig.java
```diff
@@ -0,0 +1,108 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.graphhopper.config;
+
+import com.graphhopper.util.PMap;
+
+public class ProfileConfig {
```
karussell: Here some JavaDoc would be good with a link to LM/CHProfileConfig

core/files/changelog.txt
```diff
@@ -1,4 +1,5 @@
 1.0
+    add profiles parameter and replace prepare.ch/lm.weightings and prepare.ch.edge_based with profiles_ch/lm config parameters
```
karussell: Can you add `, see #1922`?

api/src/main/java/com/graphhopper/config/ProfileConfig.java
```diff
@@ -0,0 +1,108 @@
...
+    private String name = "car";
+    private String vehicle = "car";
+    private String weighting = "fastest";
+    private boolean turnCosts = false;
+    private PMap hints = new PMap();
+
+    public static void validateProfileName(String profileName) {
+        if (!profileName.matches("^[a-z0-9_\\-]*$")) {
+            throw new IllegalArgumentException("Profile names may only contain lower case letters, numbers, underscores and dashs, given: " + profileName);
```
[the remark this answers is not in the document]
easbar: The dash might also be a minus and we have it currently for example as `utc-1` in the default naming scheme for `CHProfile`. There would probably be a way to change this and no longer allow dash/minus, but then again this character makes no problems when used in url parameters or filenames, so I thought its ok to allow it.

api/src/main/java/com/graphhopper/config/CHProfileConfig.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package com.graphhopper.config;
+
+import java.util.Objects;
+
+import static com.graphhopper.config.ProfileConfig.validateProfileName;
+
+public class CHProfileConfig {
+    private String profile = "unset_ch_profile";
```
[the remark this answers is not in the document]
easbar: Its certainly not necessary, but I thought its a bit more explicit like that. I do not like using null, because then it could actually be null after deserialization (not sure if this is good or bad, but for example if we always make sure everything is initialized it would be sufficient to check for empty string (if even) and we do not need null checks. -> I changed it to be initialized with an empty string now.

config-example.yml
```diff
@@ -20,6 +20,50 @@ graphhopper:
...
   # graph.encoded_values: surface,toll,track_type
 
+  ##### Routing Profiles ####
+
+  # Routing can be done for the following list of profiles. The fields of each profile are as follows:
+  # - name (required): a unique string identifier for the profile
+  # - vehicle (required): refers to the `graph.flag_encoder` used for this profile
+  # - weighting (required): the weighting used for this profile, e.g. fastest,shortest or short_fastest
+  # - turn_costs (true/false, default: false): whether or not turn restrictions should be applied for this profile.
+  #   this will only work if the `graph.flag_encoder` for the given `vehicle` is configured with `|turn_costs=true`.
```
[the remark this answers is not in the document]
easbar: Yes it might be possible, but lets say this would be another/next step (which is not strictly required here). The problem I am having with this is that the flag encoders are not simply something like `car`, but rather `car|turn_costs=true` or `motorcycle|your_property=here`. So right now if we would do any kind of such configurations of the flag encoders like this we would have to repeat it in each profile? Or somehow merge them? Or choose one by some (possibley not so obvious) rule? And even if we think a bit about future development when the flag encoders become collections of encoded values, then we will probably want to configure these collections of encoded values, give them a name and refer to this name in the profile (just like we are doing here).

> btw minor typo of missing 's' in both cases

fixed, thanks

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -797,31 +857,22 @@ public final CHPreparationHandler getCHPreparationHandler() {
...
-                    }
-                }
+        if (chPreparationHandler.hasCHProfiles()) {
+            return;
+        }
+
+        Map<String, ProfileConfig> profilesByName = getProfilesByName();
+        for (CHProfileConfig chConfig : chPreparationHandler.getCHProfileConfigs()) {
+            ProfileConfig profile = profilesByName.get(chConfig.getProfile());
+            assert profile != null : "there is no profile " + chConfig.getProfile();
```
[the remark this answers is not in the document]
easbar: The reason I am using assert here is that this is a private method and the existence of the profile is already checked in `checkProfilesConsistency` before. So at this point its clear that the profile exists (unless there is some programming error and my assumption that it was checked before is false). So we could either skip this check here entirely, or repeat the check again, or use an assert as a bit of a reminder that (if nothing is broken) we can be sure the profile exists. In general of course you are right and the program logic should not depend on an assert. Here it does not because the same check is done before already. If you think this is confusing somehow I think we should just remove this assert here (and use the null pointer exception to figure out what is going on if this ever pops up). I have to admit the assert message here is not chosen very well it would better be: "There is no profile ... even though this should have been checked before"

core/src/main/java/com/graphhopper/routing/ch/CHPreparationHandler.java
```diff
@@ -60,36 +60,21 @@ public CHPreparationHandler() {
...
         // throw explicit error for deprecated configs
-        if (!ghConfig.get("prepare.threads", "").isEmpty())
+        if (ghConfig.has("prepare.threads"))
             throw new IllegalStateException("Use " + CH.PREPARE + "threads instead of prepare.threads");
-        if (!ghConfig.get("prepare.chWeighting", "").isEmpty() || !ghConfig.get("prepare.chWeightings", "").isEmpty())
-            throw new IllegalStateException("Use " + CH.PREPARE + "weightings and a comma separated list instead of prepare.chWeighting or prepare.chWeightings");
+        if (ghConfig.has("prepare.chWeighting") || ghConfig.has("prepare.chWeightings") || ghConfig.has("prepare.ch.weightings"))
+            throw new IllegalStateException("Use profiles_ch instead of prepare.chWeighting, prepare.chWeightings or prepare.ch.weightings");
+        if (ghConfig.has("prepare.ch.edge_based"))
+            throw new IllegalStateException("Use profiles_ch instead of prepare.ch.edge_based");
```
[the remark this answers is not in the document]
easbar: Good idea, done

api/src/main/java/com/graphhopper/config/CHProfileConfig.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public CHProfileConfig(String profile) {
+        setProfile(profile);
+    }
+
+    public String getProfile() {
+        return profile;
+    }
+
+    public CHProfileConfig setProfile(String profile) {
```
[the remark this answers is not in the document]
easbar: Ok

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: Yes will do as soon as we have agreed on a final design

api/src/main/java/com/graphhopper/config/ProfileConfig.java
easbar: Because of the jackson conventions. If there was getHints() Jackson would allow a hints sub-node, but we want all the other fields on the top level (see the `@JsonAnySetter` in the corresponding mixin)

api/src/main/java/com/graphhopper/config/ProfileConfig.java
easbar: Yes will do as soon as we have agreed on a final design

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: Yes using a map is better, also because later when resolving the profile during a request we have to obtain the profile by name again. I used a map before and then changed it back to list, but now I do not remember why anymore -> its a map now :+1:

easbar:
> When using the Java API for the profiles it feels a bit too lengthy to create a profile via the setters, but I do not know how to make it shorter :D 

I tried reducing it and at least we have the varargs parameter so we do not have to create a list ourselves...

> except a constructor approach for most common combinations as the builder pattern we use e.g. for EncodingManager is also not shorter.

You mean we should use a builder pattern or we should not? We already have kind of a builder pattern using the fluent setters: `new ProfileConfig("profile").setVehicle("car").setWeighting("fastest");` A real builder makes no sense I think because the created object is not supposed to be immutable anyway (and @michaz does not like it for some other reason I do not remember). Using multiple constructors for the "most common" patterns is also not as nice I think, because to its not always clear what the most common patterns are and it won't be long until the common patterns do not work very well and someone adds another constructor... Ok here its probably very common to set profile,vehicle and weighting, even then the above approach seems more readable (you do not have to remember the order of the arguments).  Btw I like kotlin's named parameters for this kind of problem, its similar to the approach above.

> When adding a CH profile, why do we need an object and cannot just use a string? (btw: for LM I can imagine that we get rid of the maximum weight parameter via creating a better estimation procedure or put this into the hints of the ProfileConfig?). Because if it is just a string we could see LM or CH preparation as a property of ProfileConfig and avoid parts of checkProfilesConsistency?

Of course you are right and we do not strictly need an object (since its just containing a single string). Initially I had the `edge_based` field so I needed an object, but this is now gone because we derive it from the `turn_costs` property of the profile (so far I think this is a good approach). Like you say for LM there is also the maximum weight parameter (I think we discussed whether or not we need to keep it before?). I can imagine putting similar properties for CH as well (like the neighbor updates parameter etc.). What I do not like much about this is that these parameters are only needed for the import and afterwards have no effect. We can put the ch/lmConfig inside the profile anyway (no matter whether its a string or an object), but (why) would this be better? Yes we would not need to check if the profile exists or not anymore, but is this really so problematic? Same with the the object/string decision, is `setCHProfile(new CHProfileConfg("my_profile")` really that much worse than `setCHProfile("my_profile")`? Or in yaml:
```yaml
profiles_lm:
  - profile: "my_profile"
#vs
profiles_lm:
  - my_profile
```
? In yaml I think its even better with `profile: ` because its more clear that this references the profile? Hmm or maybe not because its a list of profiles anyway... To me these changes seem so minor that I think having the lm max weight is reason enough to use an object and I would certainly not try/force a (really?) better estimation procedure just to get rid of this object (especially because maybe later there might be more parameters that fall into this category?).

config-example.yml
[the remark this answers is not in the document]
karussell: Ok, yeah. Probably for another issue

> The problem I am having with this is that the flag encoders are not simply something like car, but rather car|turn_costs=true or motorcycle|your_property=here

Could this be done via

```
profile:
 - name: car
 - vehicle: car
 - your_property: your_value
```
?

core/src/main/java/com/graphhopper/GraphHopper.java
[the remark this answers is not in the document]
karussell: An assert statement always distracts a bit and should have some reason it is no if+throw because even if there is a bug we'll never know with assert, because in production it is disabled and in tests it does not happen. (except we create some randomization tests that randomize the initialization too)

So for higher level code where speed does not matter I would always use an if statement + throw an IllegalStateException OR decide it is not worth the clutter and remove it.

Interestingly even for code where speed matters I observed that "if+throw" can often be removed via JIT and does not hurt. E.g. recently I introduced this in RAMDataAccess.setInt and I should probably replace the remaining assert too or remove it.

config-example.yml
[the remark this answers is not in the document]
easbar: If we assume there is one encoder per profile, thats the problem. Otherwise it would be unclear whether the (one) car encoder would have turn instructions or not (and probably similar problems for other encoder properties), there would no longer be a single place to define the encoder.

api/src/main/java/com/graphhopper/config/ProfileConfig.java
karussell: What do you mean with "would allow a hints sub-node" - when would it allow this? As there is no setHints method.

core/src/main/java/com/graphhopper/GraphHopper.java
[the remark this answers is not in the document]
easbar: > An assert statement always distracts a bit

Thats a bit subjective :) I think asserts can do a great job in communicating intend (even though the present case is not a particularly good example for this).

>  it is no if+throw because even if there is a bug we'll never know with assert, because in production it is disabled and in tests it does not happen

Yes and like I said I did not add the assert because its supposed to throw anything (we check wether or not the profile exists in `checkConsistency()` already.

> So for higher level code where speed does not matter I would always use an if statement

I did not decide to use assert for performance reasons here. Performing a single if is probably very rarely a performance issue (and especially not here).

What do you prefer?
```java
// option 1:
Profile profile = profiles.get(profileName);
// profile cannot be null because we checked already
profile.doSth();

// option2
Profile profile = profiles.get(profile);
profile.doSth();

// option3
Profile profile = profiles.get(profile);
assert profile != null : "profile does not exist even though this should have been checked already";
profile.doSth();
```

Here, I am happy with any of the options, but I would not agree if we tried to never use assert whatsoever. Like I said this is not a particular good/important use-case here.

api/src/main/java/com/graphhopper/config/ProfileConfig.java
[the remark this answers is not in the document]
karussell: > utc-1

Ah, ok. Ugly. Would be nice to change this (likely in another issue & later)

karussell:
> A real builder makes no sense I think because the created object is not supposed to be immutable anyway

Ok

> but is this really so problematic?

Ok, the difference between query and import parameter is convincing :) and an object is of course not really problematic, would have been just a little bit easier.

> Or in yaml

In yaml we could avoid a separate `profiles_ch` node and just include `ch: true` for the profile.

api/src/main/java/com/graphhopper/config/ProfileConfig.java
easbar: The getter makes the hints serializable and deserializable: https://www.baeldung.com/jackson-field-serializable-deserializable-or-not#a-getter-makes-a-non-public-field-serializable-and-deserializable. So if we put `getHints` by convention the deserialized config would look like:
```yaml
profile:
  name: my_name
  hints: 
      foo: bar
      baz: foo
```

but we want:

```yaml
profile:
  name: my_name
  foo: bar
  baz: foo
```

easbar:
> In yaml we could avoid a separate profiles_ch node and just include ch: true for the profile.

Yes, but then we could/should(?) do the same in the java classes? I would rather keep this symmetric?

api/src/main/java/com/graphhopper/config/ProfileConfig.java
karussell: And can we use JsonIgnore?

core/src/main/java/com/graphhopper/GraphHopper.java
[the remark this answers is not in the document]
karussell: If I'm confident that it is really not null I would probably use option 2.

If not, I would pick option 4: throw an IllegalStateException ;)
I still do not see the benefit of an assert that is never thrown in production. Why should I continue in production when we are in a wrong situation? It feels worse than catching an exception and then just logging it :)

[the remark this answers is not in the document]
easbar: > If I'm confident that it is really not null I would probably use option 2.

Then the next person might ask the same question and maybe he/she is not as confident :)

> If not, I would pick option 4: throw an IllegalStateException ;)

Ok that also works for me.

> I still do not see the benefit of an assert that is never thrown in production. Why should I continue in production when we are in a wrong situation? It feels worse than catching an exception and then just logging it :)

The idea is that this code never goes to production, because some tests will fail. I think assertions can:

- help finding the source of an error (you know something goes wrong at some place, but assertions might tell you that the real reason something went wrong was that something else went wrong before this)

- help communicate under which assumptions some code was written

You are right because both things can be achieved equally well by throwing IllegalStateException (or some shorter helper method that does this) as long as performance is not an issue. My idea here was that the assert tells me that we are checking something that has already been checked anyway, but I see this is not easy to follow for others.

If however assertions are used more extensively (and possibly using some assertions that are more expensive than a simple boolean check) they can certainly be useful because they can be disabled for production. The present case is not a good example for that, but generally I think using more asserts is a good thing to do (especially in code where calculating some precondition only to find out whether or not we are in a sane state is expensive). For example RoutingKit does this a lot and its almost like you do not need tests anymore (or only a few), because when running the code with enabled assertions there is a very high chance some assertions fail when you are making a programming error and it improves readability when you are told what the expectations for the current method call were when the method was written. You are also able to easily test internal things that would never be easy to cover using tests. Ok, this is a whole other use-case than what we are talking about here, but I would just not go as far as saying that asserts are always bad or distracting.

-> I'll remove the assert :)

easbar:
I'll try to remove `LM/CHProfileConfig` and instead add some ch/lm flags to profile.

karussell:
> I'll try to remove LM/CHProfileConfig and instead add some ch/lm flags to profile.

Ah, didn't mean that. Your argument sounds good to keep them.

easbar:
> Ah, didn't mean that. Your argument sounds good to keep them.

Are you sure? You got me thinking again... Lets say we do something like this:

```yaml
profiles:
   - name: my_car
     vehicle: car
     weighting: fastest
ch_profiles:
   - profile: my_car
     update_neighbors: 84
```

At first sight its nice to not clutter the profile with things like `update_neighbors`, i.e. 

```yaml
profiles:
  - name: my_car
    vehicle: car
    weighting: fastest
    prepare_ch: true
    prepare_lm: false
    update_neighbors: 84
```

but after the import `update_neighbors` becomes useless and adding/not adding it makes no difference anymore (is it the same for `maximum_lm_weight`?). Therefore its ugly to have it in the profile in the first place. But this problem also exists in the first version (just in `ch_profiles` instead of `profiles`). So if we want to separate this we maybe need to really separate the 'how will the import be performed?' configuration from the 'what should be loaded?' configuration? Just using `profiles_ch` is not 'enough'/does not really help with this kind of separation.

easbar:
Regarding the 'what should be loaded?' configuration I think we agreed that in the future the answer should be 'everything from the -gh folder', but maybe we want to be able to only load parts of it? Then we need some kind of include/exclude list for the preparations, but this should really only be a list of profile names (for ch and lm), because this way we cannot do anything besides including/excluding the preparations (no new configuration). So for this use case the `ch/lm_profile`s in the current state also 'do too much' (especially if we'd add things like `update_neighbors: 84`?

karussell:
> At first sight its nice to not clutter the profile with things like update_neighbors, i.e.

Yes, I thought this was a good argument :) as having a CH property in a non-CH profile feels incorrect.

> Just using profiles_ch is not 'enough'/does not really help with this kind of separation.

Ok.

> configuration I think we agreed that in the future the answer should be 'everything from the -gh folder', 

Yes.

> So for this use case the ch/lm_profiles in the current state also 'do too much' (especially if we'd add things like update_neighbors: 84)?

Ah, ok. I see. Makes sense.

easbar:
> Yes, I thought this was a good argument :) as having a CH property in a non-CH profile feels incorrect.

Ok that is a good reason to split it into two/three sections. So I'll leave as is for now, but we might need to change it again when doing further refactorings (graph database etc.). I guess thats just the price we pay for adding the custom weightings before doing the graph storing refactorings. Btw another ugliness is that if we specify the profiles, do CH import and then change some profile parameter (especially with custom) *without* changing the profile name its inconsistent. This will also be resolved when we only use the profile configuration to do the import and afterwards consider the graph folder as single source of truth...

api/src/main/java/com/graphhopper/config/ProfileConfig.java
easbar: Yes with `@JsonIgnore` we can have `getHints` and Jackson will still not consider `hints` to be a property. -> I renamed to `getHints()`.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -772,6 +822,45 @@ public boolean load(String graphHopperFolder) {
...
+                throw new IllegalArgumentException("Unknown vehicle '" + profile.getVehicle() + "' in profile: " + profile + ". Make sure to add all vehicle used in profiles to 'graph.flag_encoders'");
+            }
+            FlagEncoder encoder = encodingManager.getEncoder(profile.getVehicle());
+            if (profile.isTurnCosts() && !encoder.supportsTurnCosts()) {
+                throw new IllegalArgumentException("The profile '" + profile.getName() + "' was configured with " +
+                        "'turn_costs=true', but the corresponding vehicle '" + profile.getVehicle() + "' does not support turn costs." +
+                        "\nYou need to add `|turn_costs=true` to the vehicle in `graph.flag_encoders`");
+            }
+            try {
+                createWeighting(new HintsMap(profile.getWeighting()), encoder, NO_TURN_COST_PROVIDER);
```
boldtrn, line 837: Sorry for being a bit lat to look at the changes. I was wondering why are we creating an instance of the weighting here? Wouldn't it be enough to check if a weighting with a certain name is "registered"/"available"? Or is this the easiest option to check if the weighting is available?

easbar: Well yes the weighting is only created to check whether or not this is possible. Simply creating it seems like the easiest (and certainly most complete) way to do this. For example with custom weightings in #1776 we also get an error if something is wrong about the custom model YAML configuration. Then again like you observed there is no real *need* to create the weighting. Do you see a problem with this?

boldtrn: > For example with custom weightings in #1776 we also get an error if something is wrong about the custom model YAML configuration this way.

Ah I see, that's a strong point to do this and makes sense.

> Then again, like you observed, there is no real need to create the weighting. Do you see a problem with this?

This created an exception for me. I use a weighting that requires certain parameters in the HintsMap, otherwise an exception is thrown. It is easy enough to fix, so that's no problem.

Issue #1869, opened 2020-01-27
Encapsulate mapping between routing request parameters and core properties


easbar:
So far we are doing:

```java
GHResponse GraphHopper#calcPaths(GHRequest ghRequest)
```

which strongly couples the `GHRequest/Response` objects to the GH core code. The logic of interpreting a `GHRequest` and figuring out what to do is distributed over several classes: `RoutingTemplate`, `GraphHopper`, `RoutingAlgorithmFactory`, `CHProfileSelector` etc. This makes it hard to understand and even more hard to test what kind of requests are mapped to which behavior of the routing engine. For testing basically the only option is to test `GraphHopper` as a whole and look at the returned `GHResponse`. At the same time its hard to *enforce* a certain behavior. All we can do is set some options in `GHRequest` and 'hope' `GraphHopper` does what we are intending to do. 

I think this can be improved by first mapping a `GHRequest` to one of multiple `RoutingRequest` subclasses, which hold the information needed for the different possible branches: For example a `RoundTripRequest` has different fields than an `AlternativeRoutingRequest`. A `CHRoutingRequest` has a `CHProfile` field, but cannot specify a `heading` (so it has no corresponding field) etc. This new mapping: 
``java
RoutingRequest interpretRequest(GHRequest)
```
would be very helpful for testing and at the same time one does not have to use it: One could also just create the `RoutingRequest` manually or interpret `GHRequest`s another way etc.

In `GraphHopper#calcPaths(RoutingRequest)` we could just do one big

```java
if (routingRequest instanceof RoundTripRequest) {
    calculateRoundTrip((RoundTripRequest)routingRequest);
} else if (routingRequest instanceof ...) {
}
```

Given that the mapping between request parameters and selected profiles / behavior gets more complicated because of #1776 (also see #1859) I think this would be useful and maybe should be done even before #1708 and #1776? WDYT?

One thing that is not so clear to me is the response object. What shall be returned from `GraphHopper#calcPaths(RoutingRequest)`? We could use something generic that works for all kind of routing requests or again allow returning a `RoutingResponse` that can be different for the different cases. Eventually it always has to be mapped to `GHResponse` however...

otbutz:
IMHO dedicated request/response classes are a lot more cleaner.

easbar:
> IMHO dedicated request/response classes are a lot more cleaner.

[...]

PR #1923, opened 2020-02-19
create routing template in a factory function for easy override

### LINKED ISSUE #1713: Make it possible to use a custom SnapPrevention
The snap prevention feature is great. I just wanted to modify the snap prevention in a project that uses GraphHopper as a library. Unfortunately, it's not possible to override the snap prevention easily. It would be nice to have an easy way to do this.

One thing that could already help would be to move the RoutingTemplate creation in the GraphHopper class into an overridable method:
```
// Move this part into a method
            RoutingTemplate routingTemplate;
            if (ROUND_TRIP.equalsIgnoreCase(algoStr))
                routingTemplate = new RoundTripRoutingTemplate(request, ghRsp, locationIndex, encodingManager, maxRoundTripRetries);
            else if (ALT_ROUTE.equalsIgnoreCase(algoStr))
                routingTemplate = new AlternativeRoutingTemplate(request, ghRsp, locationIndex, encodingManager);
            else
                routingTemplate = new ViaRoutingTemplate(request, ghRsp, locationIndex, encodingManager);
```

Then we could allow passing different EdgeFilters in the constructor of the RoutingTemplate or make the ViaRoutingTemplate extendable to use different EdgeFilters.

WDYT?

naser13:
Fixes #1713.


easbar:
The concept of routing templates is likely to be removed soon. Can you explain what kind of use case you have in mind? See also this related issue: https://github.com/graphhopper/graphhopper/issues/1869

naser13:
thank you @easbar 
but the issue #1869 will not satisfy my use case.
i have this situation: i have started a new project and added graphhopper as a dependency, per requirements i need to change lookup behaviour of the routing template. currently there is no clean way of doing that from outside graphhopper, but if routing templates had a factory i could have written my own routing template and use it.
also same thing for pathMerger behaviour

easbar:
Ok I understand. The thing is that in the future `GraphHopper` shall no longer be extensible by overriding some parts of it. It should rather be a convenience type of thing to cover some standard features (or maybe it will be split into smaller classes altogether). More advanced use-cases (like the custom lookup behavior you mentioned (you mean the location index lookup bwt?)) should be doable by just combining some library components in different ways. Unfortunately at the moment this is not so easy to do, because for example `GraphHopper#calcPaths` does way too many things that cannot be reused so easily. Afaik we can merge this change for now, but please do not consider this to be stable of any sort. If you explain in more detail what custom behavior you are trying to achieve it might help with some future design decisions, and maybe its even something that would be useful for the upstream project.

naser13:
thanks again for your answer.
i understand that changes are on the way that will make this pr obsolete and change the way things are done. but for now it would be appreciated if this gets merged so i can continue further developments.
and yes by lookup i meant location index lookup; i wanted to consider bearing when finding closest edge and filter by edges that have similar bearing so for example if i am at a junction it doesn't match me with the wrong edge.
as for the bigger changes you mentioned, is there a thread about them somewhere?

easbar:
>  it would be appreciated if this gets merged so i can continue further developments.
Can you send your license agreement (see contribution guidelines)?

>  i wanted to consider bearing when finding closest edge and filter by edges that have similar bearing so for example if i am at a junction it doesn't match me with the wrong edge.

Feel free to open a separate issue for this and maybe your changes can be used for master.

> as for the bigger changes you mentioned, is there a thread about them somewhere?

See this issue: #1895

naser13:
> Can you please send your license agreement (see contribution guidelines)?

done

> See this issue: #1895

looks promising. graphhopper is improving very fast.

boldtrn:
I have faced a similar issue in the past, #1713

easbar:
Thanks!

easbar:
Ugh for some reason the commits werent squashed and not attributed to you. If you want we can try again (add add a revert commit and apply the same changes again)

PR #1924, opened 2020-02-20
Uturn restriction

naser13:
isUturnAllowed always returned true, now it checks tcs for restrictions.


naser13:
i close this due to conflicts with master and the fact that this case does not happen often in real osm data.

PR #1928, opened 2020-02-25
GraphHopperServerConfiguration auto-detects available ports

taulinger:
Removed fixation to port 8080. Each test class detects an available port itself now.


karussell:
What is the use case here? Is it to have parallel tests?

easbar:
I sometimes ran into this issue when something else on my system is running on port 8080 (this port is quite commonly used by all kinds of software)

taulinger:
> I sometimes run into this issue when something else on my system is running on port 8080 (this port is quite commonly used by all kinds of software)

That's the point. In our case, we build GraphHopper on our internal CI-System, where port 8080 is also already in use.

karussell:
Ok. Makes sense. Can we make creating this URL somehow easier? E.g. with a helper method?

```java
Builder clientRequest(App app, String path) {
   if(!path.startsWith("/")) path = "/" + path;
   return app.client().target("http://localhost:"+ app.getLocalPort() + path).request();
}
```

And instead of changing the production GraphHopperServerConfiguration can we do this only in the tests? E.g. extend it for a test config?

karussell:
@taulinger thanks for the changes! Can you add yourself as a contributor (see CONTRIBUTORS.md) and [send us your Email address for the CLA](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md#license-agreement)?

karussell:
Thanks!

PR #1929, opened 2020-02-25
Mapzen (skadi) elevation provider

### LINKED ISSUE #384: Use high precision elevation data

msbarry:
Fixes #384 by adding an option to use Mapzen's skadi elevation data by setting `graph.elevation.provider: skadi` in config.yml.


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -625,6 +625,8 @@ private static ElevationProvider createElevationProvider(GraphHopperConfig ghCon
             elevationProvider = new SRTMGL1Provider(cacheDirStr);
         } else if (eleProviderStr.equalsIgnoreCase("multi")) {
             elevationProvider = new MultiSourceElevationProvider(cacheDirStr);
+        } else if (eleProviderStr.equalsIgnoreCase("skadi")) {
+            elevationProvider = new SkadiProvider(cacheDirStr);
```
karussell: Can you add it also to this table? https://github.com/graphhopper/graphhopper/blob/master/NOTICE.md#data

core/src/main/java/com/graphhopper/reader/dem/SkadiProvider.java
```diff
@@ -0,0 +1,106 @@
...
+    String getDownloadURL(double lat, double lon) {
+        String latStr = getLatString(lat);
+        String lonStr = getLonString(lon);
+
+        return latStr + "/" + latStr + lonStr + ".hgt.gz";
+    }
+
+    @Override
+    public String toString() {
+        return "skadi(" + baseUrl + ")";
```
karussell: Due to consistency I would just return "skadi" here

core/src/main/java/com/graphhopper/reader/dem/SkadiProvider.java
```diff
@@ -0,0 +1,106 @@
+package com.graphhopper.reader.dem;
```
karussell: Are you ok with adding the usual License header here?

PR #1934, opened 2020-02-27
Use new profile parameter to select routing profile

### LINKED ISSUE #1859: Add new 'profile' parameter to select routing profile

easbar:
This PR fixes #1859 and #1708 (without custom weighting still). Its now based on #1958. 

Current todo:

* [x] ~~update web UI: use profile parameter and profiles for 'vehicle icons'~~ do later -> #1993


reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -1385,17 +1416,26 @@ public void testDisablingLM() {
...
-        assertMoscowNodeBased(hopper, "false", false);
-        // edge_based=true -> use edge-based
-        assertMoscowEdgeBased(hopper, "true", false);
+        GHRequest req = new GHRequest(55.813357, 37.5958585, 55.811042, 37.594689);
+        // without turn costs
+        req.setProfile("profile1");
+        assertEquals(400, hopper.route(req).getBest().getDistance(), 1);
+        // with turn costs
+        req.setProfile("profile2");
+        assertEquals(1044, hopper.route(req).getBest().getDistance(), 1);
```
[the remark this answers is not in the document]
easbar: > To improve readability also in tests we could remove the comments and reuse the previously defined variable carWithoutTurnCosts="profile1" instead of just "profile1"?

Yes. Good point.

> And why are the assertMoscowEdge/NodeBased calls removed? Because it is expected to work once the converter is tested?

They weren't removed, but only inlined, because it no longer made sense to re-use the code in this method

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
```diff
@@ -236,6 +236,11 @@ private Request createPostRequest(GHRequest ghRequest) {
         String url = tmpServiceURL + "?";
         if (!Helper.isEmpty(key))
             url += "key=" + key;
+        if (!Helper.isEmpty(ghRequest.getProfile())) {
+            if (!url.endsWith("?"))
+                url += "&";
+            url += "profile=" + ghRequest.getProfile();
+        }
```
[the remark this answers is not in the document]
easbar: Ok at some point I thought it would be useful to have `profile` as url parameter, so one can quickly decide (without looking at the request body) what kind of profile was requested (e.g. when filtering requests).

easbar:
This is still very much WIP, but with #1949 this should become much easier to do. #1949 introduces a new `ProfileResolver` class and I think in this PR we will use this and put it, well, where should we put it?

* Leave it in `GraphHopper`: This is my least preferred option, because I would like to get rid of `vehicle/weighting` etc. on this level.

* Put it in the different `Route/Isochrone/EtcResources`. This seems reasonable to me, but we'd have to inject `GraphHopper` instead of only `GraphHopperAPI` interface, so we can create the profile resolver there, or Inject the `ProfileResolver` directly (maybe thats the best?)

* Use some kind of http filter that is applied even before the `Route/Isochrone/Etc/Resources` (not sure if that is better than the second option).

karussell:
> This seems reasonable to me, but we'd have to inject GraphHopper instead of only GraphHopperAPI interface,

Yeah, probably we can even remove the interface.

> well, where should we put it?

I face a similar question for e.g. tweaking the BlockAreaWeighting and duplicating this (minor) code into XyResource seems a bit ugly as the GraphHopper class should be somehow "self-contained". Or we gave up this intention but still keep the moving parts somehow down to a minimum (maybe a different issue :) #1895 )

easbar:
> I face a similar question for e.g. tweaking the BlockAreaWeighting and duplicating this (minor) code into XyResource seems a bit ugly as the GraphHopper class should be somehow "self-contained". 

Do you mean in Isochrone/SptResource? To me it looks like the problem with these is that they do not use the `GraphHopper` class in the first place. The code in these resources should be extracted to some classes that `GraphHopper` uses I guess. Similar to the `GraphHopperRouter` in #1901.

Regarding the `ProfileResolver` I think this is in fact something specific to `xyResource`, because its only used for legacy and `GraphHopper` will be independent from it.

api/src/main/java/com/graphhopper/GHRequest.java
```diff
@@ -243,6 +243,8 @@ public String getWeighting() {
      * By default it supports fastest and shortest. Or specify empty to use default.
      */
     public GHRequest setWeighting(String w) {
+        // todonow: setting weighting/vehicle only works on the client side, but no longer with GHRequest in GraphHopper
+        // should we remove it?
```
easbar, line 247: I wonder if we should remove `get/SetVehicle/Weighting` from `HintsMap` now? The problem is that when using GraphHopper as a Java library we pass `GHRequest` including a `HintsMap` to `GraphHopper#route`, but this method no longer accepts weighting/vehicle but needs a profile instead. On the other hand when we use `GHRequest` on the client side via `GraphHopperWeb` one can still use vehicle/weighting due to the `ProfileResolver` in the `RouteResource` (even though also here profile is the recommended parameter now). We could also  copy&paste `GHRequest` and `HintsMap` on the client side and adjust both versions accordingly, but I do not see a real reason to keep `vehicle/weighting`? When someone is updating the client version there will be a compile error and he/she will be forced to replace `setVehicle/Weighting` with `setProfile`, but is this a problem? Using older clients still works of course (because of `ProfileResolver`).

client-hc/src/test/java/com/graphhopper/api/GraphHopperWebIT.java
```diff
@@ -248,6 +248,8 @@ public void doNotReadFinishInstruction() {
 
     @Test
     public void testSimpleExport() {
+        // todonow: is it ok/meant to be that export() always uses get (regardless of the setting), looks like our
+        // POST endpoint does not support gpx yet?
```
easbar, line 252: I guess this is so because the POST endpoint does not support gpx and we can leave as is here?

web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
```diff
@@ -112,6 +107,7 @@ public Response doGet(
                     + "or equal to the number of points (" + requestPoints.size() + ")");
 
         // TODO these checks should be only necessary once in the core, e.g. pointHints problems are currently ignored for POST requests
+        // todonow: maybe its time to fix this?
```
easbar, line 110: Do you remember the reason this is done here instead of in `GraphHopper`? Can we move it over now?

web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
```diff
@@ -151,10 +148,11 @@ public Response doGet(
 
         GHResponse ghResponse = graphHopper.route(request);
 
-        float took = sw.stop().getSeconds();
+        long took = sw.stop().getNanos() / 1000;
         String infoStr = httpReq.getRemoteAddr() + " " + httpReq.getLocale() + " " + httpReq.getHeader("User-Agent");
+        // todonow: is there any reason we should keep logging vehicle/weighting? maybe to keep track of the number of 'legacy' requests or something?
```
easbar, line 153: Probably not?

api/src/main/java/com/graphhopper/GHRequest.java
karussell: > When someone is updating the client version there will be a compile error and he/she will be forced to replace setVehicle/Weighting with setProfile, but is this a problem?

In general this would be nice to have. Could we even remove HintsMap completely (replace with PMap)?

And I see `hints().putObject("vehicle", "bike")` is still possible so removing this should not be a problem. (E.g. we might not be fast enough to accept "profile" in the public API)

client-hc/src/test/java/com/graphhopper/api/GraphHopperWebIT.java
karussell: Ah, good catch. I think we should always prefer POST as the request is optionally compressed and has no other limitations (?)

> looks like our POST endpoint does not support gpx yet?

doPost should support it or why do you think it does not?

web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
karussell: Good question. I do not remember a reason why this is here.

web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
karussell: I would indeed probably keep it until 2.0?

api/src/main/java/com/graphhopper/GHRequest.java
easbar: > In general this would be nice to have. Could we even remove HintsMap completely (replace with PMap)?

Absolutely I already did this here: https://github.com/graphhopper/graphhopper/commit/d2cc0978c3dd284007816c7de5a1107d90284d7a

easbar:
> Thanks a lot! Look good and can't wait to have it merged :)

Haha me too :)

client-hc/src/test/java/com/graphhopper/api/GraphHopperWebIT.java
easbar: > doPost should support it or why do you think it does not?

Hm, yes you are right it should already support it. Ok then we should make `GraphHopperWeb#export` respect the `postRequest` flag. If I do this the test fails, but maybe thats for another issue.

web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
easbar: For GET its already contained in httpReq.getQueryString(). Same for `algoStr`, but I added it explicitly now.

web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
easbar: Maybe for backward compatibility? Is it ok to add these checks for POST as well? If yes we can move them into `GraphHopper`.

easbar: But there is more: Also the enableElevation, pointsEmpty checks and the "start heading" logic is missing for POST? All this should probably go into `GraphHopper#calcPaths` as well? Maybe this is also for another issue?

karussell: Should be good for POST too, yes. Good idea with the new issue.

karussell:
Nice, thanks :+1: ! Really cool that HintsMap is gone too :) - I think we can close https://github.com/graphhopper/graphhopper/issues/940 too now?

easbar:
>  I think we can close #940 too now?

Ok, I had never even seen that one :) `AlgorithmOptions` isn't very useful either IMO but thats another topic probably related to `RoutingAlgorithmFactory`.

web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
```diff
@@ -151,10 +148,10 @@ public Response doGet(
 
         GHResponse ghResponse = graphHopper.route(request);
 
-        float took = sw.stop().getSeconds();
+        long took = sw.stop().getNanos() / 1000;
         String infoStr = httpReq.getRemoteAddr() + " " + httpReq.getLocale() + " " + httpReq.getHeader("User-Agent");
         String logStr = httpReq.getQueryString() + " " + infoStr + " " + requestPoints + ", took:"
-                + took + ", " + algoStr + ", " + weighting + ", " + vehicleStr;
+                + took + "micros, algo: " + algoStr + ", profile: " + profileName + ", " + weightingVehicleLogStr;
```
karussell, line 154: @easbar for better readability IMO we should use milliseconds: `+ took/1000f + "ms...`?

easbar: Ok micros might still be a bit too optimistic ;) I changed to millisecond with one decimal place

karussell: :+1:

PR #1942, opened 2020-02-29
Bilinear interpolation for height tiles

msbarry:
Replace calcMean with bilinear interpolation for height tiles.  You can enable this by setting `graph.elevation.interpolate` to true (or if you had been using `graph.elevation.calc_mean` previously). See discussion [here](https://discuss.graphhopper.com/t/improving-elevation-profiles-for-short-routes/5244).

This is a stead climb and descent I used to bike on in central Pennsylvania, using SRTM elevation data without bilinear interpolation looks like this:

image

And after bilinear interpolation looks like this:

Screen Shot 2020-02-28 at 9 47 37 PM

And with the mapzen elevation provider from #1929 it looks like this, which is closest to what I expect:

Screen Shot 2020-02-28 at 9 58 16 PM


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -601,6 +601,10 @@ private static ElevationProvider createElevationProvider(GraphHopperConfig ghCon
                 ? ghConfig.getBool("graph.elevation.calcmean", false)
                 : ghConfig.getBool("graph.elevation.calc_mean", false);
 
+        boolean interpolate = ghConfig.has("graph.elevation.interpolate")
+                ? ghConfig.getBool("graph.elevation.interpolate", false)
+                : eleCalcMean;
```
karussell: Let's remove the fall back for graph.elevation.calcmean and throw an error if it is still specified.

core/src/main/java/com/graphhopper/reader/dem/HeightTile.java
```diff
@@ -77,6 +78,15 @@ void setHeights(DataAccess da) {
         this.heights = da;
     }
 
+    private short getHeightSample(int x, int y) {
+        // always keep in mind factor 2 because of short value
+        return heights.getShort(2 * (y * width + x));
+    }
+
+    private double linearInterpolate(double a, double b, double f) {
+        return a < MIN_ELEVATION ? b : b < MIN_ELEVATION ? a : (a + (b - a) * f);
```
karussell: Why is this handling with MIN_ELEVATION necessary? What does this mean and what is the unit?

And IMO the method is not guarded if a and b are smaller than MIN_ELEVATION.

core/src/main/java/com/graphhopper/reader/dem/HeightTile.java
```diff
@@ -85,38 +95,37 @@ public double getHeight(double lat, double lon) {
...
+                x = width - 1;
+            int y = height - 1 - (int) (height / verticalDegree * deltaLat);
+            if (y < 0)
+                y = 0;
+
+            elevation = getHeightSample(x, y);
         }
-
-        return (double) value / counter.get();
+        return elevation < MIN_ELEVATION ? Double.NaN : elevation;
```
karussell: Instead of returning NaN should we better throw an exception? NaN will cause trouble somewhere else.

core/src/main/java/com/graphhopper/reader/dem/HeightTile.java
msbarry: My understanding is that voids in elevation data (in SRTM at least) show up as Short.MIN_VALUE, so the MIN_ELEVATION check is meant to catch that since otherwise we'd be comparing a double to a short.

My thought here was that we make ourselves a bit more robust to voids if when any of the 4 surrounding points that we are averaging are missing, we average the other points.

core/src/main/java/com/graphhopper/reader/dem/HeightTile.java
msbarry: This is the behavior from line 102-103 of the previous version of this code - NaN was the value returned for voids. I checked out PointList.calcDistance when elevation is NaN and it can't handle it. Is there code that I'm missing elsewhere that handles these voids?

msbarry: also FWIW valhalla does a similar range check when using mapzen/skadi terrain tiles https://github.com/valhalla/valhalla/blob/9dd8b88aa93b91888c4a6da8b09454a1ec9a0c2b/src/skadi/sample.cc#L224 and has some special case handling to skip over these when computing grade/total elevation gain/3d profiles with missing elevation data. Maybe this (or a separate) review should tighten up elevation void handling?

core/src/main/java/com/graphhopper/reader/dem/HeightTile.java
karussell: Oh, indeed. You are right. This is quite surprising as we have no proper NaN handling downstream for these values. Probably we never stumbled over this as we use CGIAR that has filled the void. So yes, this is probably for another issue and good idea about the surrounding points.

core/src/main/java/com/graphhopper/reader/dem/HeightTile.java
msbarry: I started working on a PR to handle voids downstream, and it started looking rather large so I opened up #1946 so we can discuss the best path forward.

msbarry:
Thinking a bit more about this, it would probably be better to have the config option be `graph.elevation.interpolate: bilinear` so we can support other interpolation methods in the future (possibly including ones that fill voids automatically per #1946) - changed the boolean to an enum.

core/src/main/java/com/graphhopper/reader/dem/ElevationInterpolation.java
```diff
@@ -0,0 +1,49 @@
...
+            return NONE;
+
+        for (ElevationInterpolation interpolation : values()) {
+            if (interpolation.name().equalsIgnoreCase(name)) {
+                return interpolation;
+            }
+        }
+
+        return NONE;
+    }
```
karussell: I would keep it simple for now and only use the two strings :) ? Or why did you choose to introduce this enum?

msbarry: Good call, that's over-engineering for now. Since we only have one interpolation method so far, just a boolean is good.  I think we should keep the config as `graph.elevation.interpolation: bilinear` so it's clear what it's doing and so configs don't have to change if we add other interpolation methods later.

karussell:
Thanks!

BTW: I think we never experienced an issue with voids in production because we use the void-filled CGIAR data.

PR #1943, opened 2020-03-03
Simplify entire edges on import instead of just pillar nodes

### LINKED ISSUE #1353: should the DouglasPeucker be used for the whole geometry instead of only pillar nodes?
I found in the OSMReader, 
simplifyAlgo.simplify(pillarNodes);

I feel it's better to apply DouglasPeucker to the whole edges.
Guess it could reduce more unnecessary points.

msbarry:
Fix #1353 by simplifying whole edge geometries instead of just the pillar nodes. Since only edges with more than one pillar node remaining are stored in edge geometries, this ends up reducing edge geometry file size by 15-20%

This did result in some churn in the integration tests, please take a close look at those and let me know if they look OK. It also removed a case to not calculate edge distances on edges that cross the international date line - this didn't seem necessary since that code uses DistanceCalcEarth which handles these crossings correctly but let me know if I'm missing something.


reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -968,15 +968,15 @@ public void testKremsCyclewayInstructionsWithWayTypeInfo() {
 
         assertEquals("turn right onto Pfarrplatz", il.get(2).getTurnDescription(tr));
         assertEquals("turn right onto Margarethenstraße", il.get(3).getTurnDescription(tr));
-        assertEquals("keep left onto Hoher Markt", il.get(5).getTurnDescription(tr));
-        assertEquals("turn right onto Wegscheid", il.get(7).getTurnDescription(tr));
-        assertEquals("turn right onto Ringstraße, L73", il.get(9).getTurnDescription(tr));
-        assertEquals("keep left onto Eyblparkstraße", il.get(10).getTurnDescription(tr));
-        assertEquals("keep left onto Austraße", il.get(11).getTurnDescription(tr));
-        assertEquals("keep left onto Rechte Kremszeile", il.get(12).getTurnDescription(tr));
+        assertEquals("keep left onto Hoher Markt", il.get(4).getTurnDescription(tr));
```
msbarry: instruction 4 that went away was `"continue onto Margarethenstraße"`

reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -575,12 +575,12 @@ public void testMonacoVia() {
...
         PathWrapper arsp = rsp.getBest();
         assertEquals(6875.2, arsp.getDistance(), .1);
-        assertEquals(171, arsp.getPoints().getSize());
+        assertEquals(170, arsp.getPoints().getSize());
 
         InstructionList il = arsp.getInstructions();
         assertEquals(30, il.size());
         assertEquals("continue onto Avenue des Guelfes", il.get(0).getTurnDescription(tr));
-        assertEquals("continue onto Avenue des Papalins", il.get(1).getTurnDescription(tr));
+        assertEquals("turn slight left onto Avenue des Papalins", il.get(1).getTurnDescription(tr));
```
msbarry: these changes are probably because by forcing simplification to always keep the closest pillar node to each tower node, we get a bit more resolution on the angles into and out of each intersection.

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -720,12 +694,15 @@ EdgeIteratorState addEdge(int fromIndex, int toIndex, PointList pointList, IntsR
...
-        if (nodes > 2) {
+        if (pointList.size() > 2) {
             if (doSimplify)
-                simplifyAlgo.simplify(pillarNodes);
+                simplifyAlgo.simplifyExcludeFirstLast(pointList);
 
-            iter.setWayGeometry(pillarNodes);
+            // If the entire way is just the first and last point, do not waste space storing an empty way geometry
+            if (!pointList.isEmpty())
+                iter.setWayGeometry(pointList);
```
msbarry: at first I was creating a shallow copy excluding first and last points, but then I tried this approach of removing the first and last point as part of simplification and it was consistently faster than the shallow copy approach on a medium size ~1GB extract (US northeast)

msbarry: alright, looks like that wasn't safe because it needs to remove first and last point whether you simplify or not.

karussell:
Thanks again!

PR #1944, opened 2020-03-03
Refactor distance calculation utilities to be elevation-aware

msbarry:
In preparation for making Douglas Peucker polyline simplification elevation-aware ([discussion](https://discuss.graphhopper.com/t/improving-elevation-profiles-for-short-routes/5244), [larger diff I'm breaking into separate reviews](https://github.com/graphhopper/graphhopper/compare/master...msbarry:3d-polyline-simplification?expand=1)), this review implements `calcNormalizedEdgeDistance3D` on each `DistanceCalc` implementation so 3D Douglas Peucker can call it from [here](https://github.com/graphhopper/graphhopper/compare/master...msbarry:3d-polyline-simplification?expand=1#diff-5b705c3852c005af6602a800351dfb28R147) and refactors a bit:

- Now that each implementation is 3D-aware, remove `DistanceCalc3D` and add a 3D `calcDist` method onto each implementation
- Rename `DistanceCalc2D` to `DistanceCalcEuclidean` since it also handles elevation


api/src/main/java/com/graphhopper/util/DistanceCalcEarth.java
```diff
@@ -254,6 +305,10 @@ public boolean isCrossBoundary(double lon1, double lon2) {
         return abs(lon1 - lon2) > 300;
     }
 
+    protected boolean hasElevationDiff(double a, double b) {
+        return a != b && !Double.isNaN(a) && !Double.isNaN(b);
+    }
```
karussell: Why do we need this method? If calcDist3D is called then the elevation values should be valid and not NaN.

msbarry: I thought that any voids in elevation data show up as NaN?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -121,6 +121,7 @@
     // for data reader
     private String dataReaderFile;
     private double dataReaderWayPointMaxDistance = 1;
+    private double dataReaderElevationMaxDistance = 0;
```
karussell: This is currently unused.

karussell:
Thanks also here! Really valuable work @msbarry!

PR #1945, opened 2020-03-04
Clean up TurnCostStorage

michaz:
TurnCostStorage had several layers of read/write methods. To me, it looked like the class couldn't make up its mind if its IntRefs/Encoders should work the same way as for edges, or not. And then got stuck in the middle.

I just removed everything except *one* way of accessing it, namely the one that is used in production code. And adapted all the tests.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTurnRelationParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTurnRelationParser.java
@@ -21,7 +21,6 @@ import com.graphhopper.reader.OSMTurnRelation;
 import com.graphhopper.routing.profiles.*;
 import com.graphhopper.routing.util.DefaultEdgeFilter;
 import com.graphhopper.storage.Graph;
-import com.graphhopper.storage.IntsRef;
 import com.graphhopper.storage.TurnCostStorage;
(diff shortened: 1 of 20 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/OSMTurnRelationParser.java
```diff
@@ -103,7 +103,7 @@ private EdgeExplorer getOutExplorer(Graph graph) {
     /**
      * Add the specified relation to the TurnCostStorage
      */
-    void addRelationToTCStorage(OSMTurnRelation osmTurnRelation, IntsRef turnCostFlags,
+    void addRelationToTCStorage(OSMTurnRelation osmTurnRelation,
```
Azbesciak: Interface is generic, but `OSMTurnRelation` is used... Maybe redesign?
Also, turn relations can have multiple edges (which is already wrong in GH [should be](https://www.openstreetmap.org/relation/6655738) but [is](https://graphhopper.com/maps/?point=52.402065%2C16.921917&point=52.402046%2C16.922046&locale=en-US&vehicle=car&weighting=fastest&turn_costs=true&use_miles=false&layer=Omniscale)), but there... abstraction leaks even more.

PR #1949, opened 2020-03-08
Extract profile selection

### LINKED ISSUE #1800: Replace Weighting.matches(HintsMap) with a different hook
See #1797. We need one place where we:

 1. convert request parameters into a Weighting (createWeighting) and
 2. pick the correct CH/LMProfile 

Is a new hook in GraphHopper required, so that it can be used in GH core, map matching & isochrone?

easbar:
This PR aims to extract all the code that is related to selecting the routing profile using the different request parameters like `vehicle` and `weighting`. The idea is that here we put all this code into a single place (`ProfileResolver`) and as a follow-up we can then put this `ProfileResolver` *in front of* `GraphHopper#calcPaths` etc. (c.f. #1859). 

Fixes #1800.


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -913,9 +922,11 @@ private void initLMPreparationHandler() {
...
         for (LMProfileConfig lmConfig : lmPreparationHandler.getLMProfileConfigs()) {
             ProfileConfig profile = profilesByName.get(lmConfig.getProfile());
-            FlagEncoder encoder = encodingManager.getEncoder(profile.getVehicle());
-            // note that we do not consider turn costs during LM preparation?
-            Weighting weighting = createWeighting(new HintsMap(profile.getWeighting()), encoder, NO_TURN_COST_PROVIDER);
+            // note that we do not consider turn costs during LM preparation. this is important if we want
+            // to allow e.g. changing the u_turn_costs per request (we have to use the minimum weight settings (
+            // = no turn costs) for the preparation)
+            // todonow: is it too ugly this way?
+            Weighting weighting = createWeighting(profile, new PMap().put("__disable_turn_costs", true));
```
easbar: Any opinion here? `createWeighting` normally uses a `DefaultTurnCostProvider` if `profile.isTurnCosts()==true`. But for LM we always want to enforce not using turn costs for the preparation, but to me it seemed like this does not justify adding an extra parameter, so I am using the pmap workaround here...

web-bundle/src/main/java/com/graphhopper/resources/SPTResource.java
```diff
@@ -62,11 +60,15 @@ public Response doGet(
...
 
         StopWatch sw = new StopWatch().start();
-
-        if (!encodingManager.hasEncoder(vehicle))
-            throw new IllegalArgumentException("vehicle not supported:" + vehicle);
-
-        FlagEncoder encoder = encodingManager.getEncoder(vehicle);
+        HintsMap hintsMap = new HintsMap();
+        RouteResource.initHints(hintsMap, uriInfo.getQueryParameters());
+        hintsMap.put(Parameters.CH.DISABLE, true);
```
easbar: Is it fine to overwrite ch/lm.disable here or should we throw an error?

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
```diff
@@ -0,0 +1,195 @@
...
+                    "\nYou can also try disabling LM altogether using " + Parameters.CH.DISABLE + "=true" +
+                    "\nrequested:  " + getLMRequestAsString(hintsMap) + "\nmatched:   " + lmProfilesAsStrings(matchingProfiles) + "\navailable: " + lmProfilesAsStrings(lmProfiles));
+        }
+    }
+
+    protected boolean chProfileMatchesHints(CHProfile p, HintsMap hintsMap) {
+        Boolean edgeBased = getEdgeBased(hintsMap);
+        Integer uTurnCosts = getUTurnCosts(hintsMap);
+        return (edgeBased == null || p.isEdgeBased() == edgeBased) &&
+                (uTurnCosts == null || p.getWeighting().getTurnCostProvider().getName().equals(uTurnCosts.toString())) &&
```
easbar: I do not like how we rely on the turn cost provider name to contain the u-turn costs here, but so far I do not have a better idea to do this. Also this is only needed to cover a (probably exotic) case: Multiple edge-based profiles with different u-turn costs. Once we have the profile request parameter we 'only' need to support this u-turn costs comparison here for legacy reasons, so this makes me even more want to get rid of it.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -833,8 +840,9 @@ private void checkProfilesConsistency() {
                         "\nYou need to add `|turn_costs=true` to the vehicle in `graph.flag_encoders`");
             }
             try {
-                createWeighting(new HintsMap(profile.getWeighting()), encoder, NO_TURN_COST_PROVIDER);
+                createWeighting(profile, new PMap());
             } catch (IllegalArgumentException e) {
+                // todonow: update, what are we really checking here?
                 throw new IllegalArgumentException("The profile '" + profile.getName() + "' was configured with an unknown weighting '" + profile.getWeighting() + "'");
```
karussell: In #1776 I stumbled over this check as I wanted to inject the graph into the Weighting but at this time graph is null. 

But with #1950 I found a good solution so that I kept this method here :)

easbar: Ah ok. The graph being null problem when running this check should also be already fixed with this: 639bd9ab69ae63cd0ea580e9076b9c1679e855b5

karussell: Aha, where did you stumble over this?

easbar: Well its required for this branch because the turn cost provider is created in `createWeighting` now (and thus there is a call to `ghStorage.getTurnCostStorage()`).

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: I would not do this here and just force the usage of no turn cost provider in LM. IMO this is something implementation specific?

easbar: What do you mean? You mean we should always set `turn_costs: false` for all profiles with LM? But then we cannot use turn costs with LM...

web-bundle/src/main/java/com/graphhopper/resources/SPTResource.java
karussell: Isn't the none-CH case the default now? But yeah, I think I had to do this for the /custom endpoint too.

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: Ah you mean somehow disabling the turn costs in `PrepareLandmarks`? Yes that would be the right thing to do, but how? `PrepareLandmarks` just gets a `Weighting`, but how would it disable the turn costs then?

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
karussell: Yes, probably we have to accept that ProfileResolver is mainly for backward compatibility? The bigger problem is the new getTurnCostProvider method, which I would prefer to be private.

web-bundle/src/main/java/com/graphhopper/resources/SPTResource.java
easbar: Hm, no `graphHopper.resolveProfile(hintsMap)` will give us a CH profile or error if it does not exist, so we have to explicitly disable CH/LM to obtain the profile no matter whether a CH/LM profile for these hints exists or not.

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
easbar: Yes, I would like to remove/not add `getTurnCostProvider` as well. But how else can we check what u-turn costs a weighting is using? Maybe we can skip this selection rule and throw an error if there are multiple edge-based weightings for the given weighting and vehicle. In this case one would *have* to use the profile parameter when using multiple edge-based CH preparations with different u-turn costs.

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: > Ah you mean somehow disabling the turn costs in PrepareLandmarks?

Yes. Isn't it already ignored when we do `traversalMode = TraversalMode.NODE_BASED` in LandmarkStorage?

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
easbar: > Yes, probably we have to accept that ProfileResolver is mainly for backward compatibility?

Yes absolutely, maybe not yet in this PR but `ProfileResolver` *only* exists for backward-compatibility.

karussell: > In this case one would have to use the profile parameter when using multiple edge-based CH preparations with different u-turn costs.

This sounds not too bad.

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: Ah yes thats a good point. Yes that should fix it :+1:

web-bundle/src/main/java/com/graphhopper/resources/SPTResource.java
karussell: So, if there is a profile without CH, then this will work without the parameter, but if the profile is listed in `profiles_ch` then we need the disabling parameter?

The problem is, what if we add a new algorithm xy? Then in theory I would expect that the profile selection remains unchanged and we only specify `profiles_xy: [car_xy]`, i.e. the profile `car` without the algorithm xy can still exist. Not sure if this is good as it increases the profile count further.

easbar: > So, if there is a profile without CH, then this will work without the parameter, but if the profile is listed in profiles_ch then we need the disabling parameter?

Yes, or to put it differently if we put ch.disable=true as we do here we get the profile matching our hints no matter what LM/CH preparations there are (and for isochrones we are not even using them). But you are right the profile selection depends on the available LM/CH preparations (it always did).

> The problem is, what if we add a new algorithm xy? Then in theory I would expect that the profile selection remains unchanged 

But why? If there is a new algorithm we likely want to use it? If the selection remained unchanged what would the new algorithm be good for? In this sense I think its better not to expose the algorithm selection such that it can always be changed internally. As a user of the API I am generally not interested which algorithm is used to solve my problem. What I think is problematic though, that sometimes it can be hard to figure out which algorithm needs to be used, because for certain parameters only certain algorithms work (for example blocked areas etc.).

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
easbar: Done

web-bundle/src/main/java/com/graphhopper/resources/SPTResource.java
karussell: Ok, regarding the parameter setting and I think I better understand this now: we have to do it here to avoid a failure if LM or CH is configured.

karussell:
@easbar I just noticed when I merged this PR into another PR: why has the hopper.createWeighting(profileConfig, hints) method a separate `hints` parameter when the profileConfig also has a `hints`? Is this identical and we can remove `hints` from createWeighting?

easbar:
No its not identical. The idea is that hints are the hints we get from a request (in addition to the selected profile). For example in case we select a profile with short_fastest weighting the `hints` parameter might contain/overwrite the distance factor from this profile. However, we need to restrict this to certain cases (for example the distance factor cannot be overwritten for CH). See also the javadocs of `GraphHopper#createWeighting` (but maybe they need to be a bit more clear)

karussell:
Ok, I see. So all hints unrelated to the profile like which algo (e.g. ch/lm.disable) go into that extra hint.

easbar:
Yes, exactly. But we might have to add some 'merging' rules (like take the distance factor from profile.hints, but if hints also has it overwrite it). This will be similar to using server-custom-models and overwriting with request-custom-models?

karussell:
> But we might have to add some 'merging' rules (like take the distance factor from profile.hints, but if hints also has it overwrite it).

Hmmh, this is a bit ugly. I think we should 1. make the difference really strict or 2. use only one HintsMap. Because otherwise we end up doing looking into two maps for internal usage.

For CustomModel this is a bit different as we can do the merging part without internal knowledge because the "CustomRequest" is automatically serialized into a CustomModel via jackson as it has only up-front-known parameters. For example:

```
hints: {ch.disable: true, distance_factor=0}
profile.hints: {}
=> merged_hints: how can we do this without internal knowledge about a specific profile?

custom request: { vehicle_weight: 50, ch.disable=true }
=> request custom_model: : { vehicle_weight: 50 }
import custom_model: { vehicle_weight: 40, max_speed: {...} }
merged custom_model: { vehicle_weight: 50, max_speed: {...} }
```

easbar:
> either make the difference really strict

What do you mean?

> or use only one HintsMap

This is not possible imo, because the two maps are two different things: One contains the parameters that were configured on the server side and the other contains the parameters passed in with the request by the user.

> Because otherwise we end up looking into two maps for all internal usage.

Yes, and I think this is exactly what we want: There are two different sets of parameters and it should be up to 'us' or whoever implements the method what to do (how to create a weighting) with these parameters.

> how can we do this without internal knowledge about a specific profile?

Why are we supposed to do this without internal knowledge about the profile? I understand `createWeighting(profile, hints)` like this: "Create a weighting for this profile and these request hints (and whoever implements the method can decide how these parameters should be interpreted)".

karussell:
Ah, ok. So the "general query hints" are the only thing we need and use inside the Weighting but still the profile.hints could modify this "general query hints" if necessary (?) And the current location where this "overwrite" happens is in GraphHopper.createWeighting.

I see ... I had planned (for the CustomModel) the "overwrite" to happen before that method to keep the params of that method low.

easbar:
Yes there are query hints and properties given in the profile. Ok, we could do the overwrite before `createWeighting`, but I decided otherwise for the following two reasons:

* I still wanted to allow making the weighting creation customizable. Therefore I thought the most general thing we can do is to give everyone who implements this method access to both the selected profile and the hints coming in via the request. If we did some of the 'merging' before the method this might be too opinionated for some use cases.

* If we only pass `ProfileConfig` to `createWeighting` I am not sure where to put the request hints (we should not modify the profile). If we only pass the hints to `createWeighting` we no longer have access to the profile: Sure we could put the profile fields (weighting, vehicle etc.) into the hints (merge it), but to me this felt weird, because then these fields lose their 'special' meaning (and they will just be some more 'hints').

karussell:
> I still wanted to allow making the weighting creation customizable

Yes, maybe this is a good idea. The problem at the moment is that hints only allows "flat" `String,String` key-value pairs. I.e. for CustomModel I have to introduce a CustomProfileConfig (with a customModel property) and additionally CustomModel as parameter in createWeighting. And so I though I do the overwrite before createWeighting so that I can at least avoid one of these changes.

But maybe a simpler solution could be to allow `String,Object` key-value pairs?

easbar:
Sure yes we can change this to `Map<String, Object>`. That should not be a problem at all. Its already a `PMap` (a `Map<String, String>` with some camel case/underscore conversion) and I do not think we need or even want the camel case/underscroe conversion here

PR #1950, opened 2020-03-10
Avoid graph dependency

karussell:
I think I found a good solution to the graph-dependency problem with always have in the Weighting: now we can fetch the tower nodes only (should be similar fast) without a NodeAccess only via the EdgeIteratorState without introducing a new method.

And additionally I found a minor speed improvement (moved BBox creation out of the loop).

This also helps making stuff in #1776 easier.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -1122,8 +1122,6 @@ public class GraphHopper implements GraphHopperAPI {
                 return Collections.emptyList();
 
             QueryGraph queryGraph = QueryGraph.lookup(graph, qResults);
-            if (weighting instanceof BlockAreaWeighting)
-                ((BlockAreaWeighting) weighting).setQueryGraph(queryGraph);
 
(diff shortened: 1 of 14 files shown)


easbar:
Nice! Maybe we should add some constants or an enum for mode=0,1,2,3,4?

karussell:
enums would be clearer but they make the call so long :) ?

fetchWayGeometry(ONLY_PILLAR_NODES)
fetchWayGeometry(TOWER_NODES_ONLY)
fetchWayGeometry(BASE_AND_PILLAR_NODES)

or we remove NODES from them ...

easbar:
Ok but whats the problem with the long names? So far every time I see `fetchWayGeometry` I find myself rereading the method to figure out what e.g. mode=2 means... Yes we could go without the `NODES` in the names: `PILLAR_ONLY, PILLAR_AND_BASE, PILLAR_AND_ADJ, ALL, TOWER_ONLY`.

karussell:
I find the numbers kind of logical, but obviously it is not ;) ... so, will introduce the enum!

karussell:
> PILLAR_AND_BASE

If you have no objections I will use BASE_AND_PILLAR to indicate the order is "base+pillar nodes" and not "pillar nodes + base".

easbar:
Ah you mean the order in the returned point list? Yes that makes sense.

Issue #1952, opened 2020-03-12
Maps: GUI state for turn_costs=true doesn't carry over


michaz:
The web GUI can suddenly decide that we don't care about turn restrictions any more.

Steps to reproduce:

- Go here:

https://graphhopper.com/maps/?point=38.540026%2C-121.735146&point=38.535508%2C-121.741292

- Click "bike"
- Click "car"

--> turn_costs=false appears in the query string, and turn restriction isn't respected.

easbar:
This will be fixed with #1859.

easbar:
Should be fixed with #2023, but let's keep this open until its deployed and check again.

PR #1953, opened 2020-03-13
Improve resolution of elevation profiles with 3D Douglas Peucker and long edge sampling

msbarry:
Improve the resolution of elevation profiles by:

1. sampling points along long edges so we don't miss elevation changes over that segment
2. making Douglas Peucker polyline simplification elevation-aware so it keeps points that only impact the elevation profile and not the 2D shape

These options have the potential to increase geometry size, so I added 2 configuration options to control them:

1. `graph.elevation.long_edge_sampling_distance` which defines the distance in meters that it will sample points along long edges, for example you might set this to somewhere around 30m if your elevation provider has 30m resolution. It defaults to 0 which disables long edge sampling.
2. `graph.elevation.way_point_max_distance` which defines the elevation tolerance during polyline simplification, for example with a 1m lat/lon and elevation tolerance removes points that result in a <1m shift of the 3D polyline. It defaults to Double.MAX_VALUE which disables elevation-influenced simplification.

**Edge Geometry Size Increase**

I ran this code on `north-america_us_pennsylvania.pbf` with CGIAR and bilinear interpolation with different values of those parameters and saw these edge geometry size increases over what's on master now:

sampling distance ➡️<br> elevation max distance ⬇️  | 15m | 30m | 45m | 60m | 100m | 500m | off (∞)
-- | -- | -- | -- | -- | -- | -- | --
1m | 19% | 17% | 17% | 15% | 11% | 6% | 6%
1.5m | 11% | 11% | 9% | 9% | 6% | 2% | 2%
2m | 6% | 6% | 6% | 6% | 4% | 2% | 2%
3m | 2% | 2% | 2% | 2% | 2% | 0% | 0%
4m | 2% | 2% | 2% | 2% | 0% | 0% | 0%
5m | 0% | 0% | 0% | 0% | 0% | 0% | 0%
10m | 0% | 0% | 0% | 0% | 0% | 0% | 0%
50m | 0% | 0% | 0% | 0% | 0% | 0% | 0%
off (∞) | 0% | 0% | 0% | 0% | 0% | 0% | 0%

I ran the same test with mapzen/skadi provider from #1929. Results were similar, but the increase was slightly larger (26% instead of 19% for 15m/1m) probably because of the higher resolution of the underlying data.

**Import time performance**

I observed the "pass2" time during import. There was no change for different values of the elevation factor, but shorter long edge sampling distances did increase that time by 25% for 15m, 9% for 30m, 8% for 45m, and <2% for longer distances. Last time I imported the world pass2 took 110 minutes, so even a 25% increase is only about a 28 minute increase which seems negligible to me.

**Examples**

This [2km straight segment](https://graphhopper.com/maps/?point=41.7373489604297%2C-79.785794871728&point=41.7511768657977%2C-79.8039368496225&locale=en&vehicle=bike&weighting=fastest&elevation=true&turn_costs=false&use_miles=false&layer=TF%20Cycle) loses all elevation details on master
image

but with 30m edge sampling and 1m elevation resolution we get that elevation detail back:
image

Let me know if there are other examples or different values of the 2 parameters you want to see.

See further discussion [here](https://discuss.graphhopper.com/t/improving-elevation-profiles-for-short-routes/5244)


core/src/main/java/com/graphhopper/reader/dem/EdgeSampling.java
```diff
@@ -0,0 +1,72 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Ensures that elevation is sampled along a point list with no more than maxDistance between samples. Works by adding
+ * points along long edges and fetching elevation at each inserted point.
+ *
+ * This uses a simple linear approximation to interpolate between points which should be fine for short distances, but
+ * if support for longer distances is needed or to handle segments crossing the international date line, should use a
+ * more robust algorithm that interpolates along the great circle path between points.
```
msbarry: Do we think that this assumption and approximation are fine? Or should we use a more exact calculation to interpolate between points.

core/src/main/java/com/graphhopper/util/DouglasPeucker.java
```diff
@@ -53,6 +56,17 @@ public DouglasPeucker setMaxDistance(double dist) {
         return this;
     }
 
+    /**
+     * Weight to give elevation changes during simplification:
+     * 0 means elevation is ignored during simplification
+     * 0.2 means that 5 meters of elevation change is treated like
+     * 1m of lat/lon change
+     */
+    public DouglasPeucker setElevationFactor(double factor) {
```
msbarry: Do we want to stay with this unit and name for the elevation factor param? I had thought about making it either elevation max dist (with unit meters) or just inverse of what this is now.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -542,6 +560,8 @@ public GraphHopper init(GraphHopperConfig ghConfig) {
 
         // elevation
         this.smoothElevation = ghConfig.getBool("graph.elevation.smoothing", false);
+        this.longEdgeSamplingDistance = ghConfig.getDouble("graph.elevation.long_edge_sampling_distance", 0);
+        this.elevationSimplifyFactor = ghConfig.getDouble("graph.elevation.simplify_factor", 0);
```
msbarry: Without bilinear interpolation, the output can look pretty bad because it enhances the resolution on blockiness. Do we want to warn when long edge sampling is used without bilinear interpolation?

For example with SRTM and no bilinear interpolation:
image

msbarry: Also, should these be enabled by default? If so, what values should we use?

core/src/main/java/com/graphhopper/util/DouglasPeucker.java
karussell: The absolute value (elevation max dist in meters) makes sense and would be preferred (I initially did not understand it ... see above ;) ). But I guess there is a reason why you chose it as a percentage?

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: Yes, for now I would keep the current settings as default (i.e. disabling 'long edge sampling'). But if we learn that it is worth the extra costs we can change it to like 60m (and 1m for max ele simplify)?

> graph.elevation.long_edge_sampling_distance

Shouldn't the default for this be Double.MAX_VALUE?

core/src/main/java/com/graphhopper/util/DouglasPeucker.java
msbarry: That makes sense, I switched everything over to "elevation max dist" - elevation factor was an attempt to pass through directly what the calculation needed, but max distance makes more sense for users.

core/src/main/java/com/graphhopper/GraphHopper.java
msbarry: OK sounds good, and I changed both of them to default to Double.MAX_VALUE.

Do you think that I should add sections to config-examply.yml explaining how to use these 2 new parameters (and bilinear interpolation)?

karussell: Ah, yes. Adding this to config-example.yml is a good idea

core/src/main/java/com/graphhopper/reader/dem/EdgeSampling.java
```diff
@@ -0,0 +1,72 @@
...
+        double lastLat = input.getLat(0), lastLon = input.getLon(0), thisLat, thisLon, thisEle;
+        for (int i = 0; i < nodes; i++) {
+            thisLat = input.getLat(i);
+            thisLon = input.getLon(i);
+            thisEle = input.getEle(i);
+            if (i > 0 && !distCalc.isCrossBoundary(lastLon, thisLon)) {
+                double segmentLength = distCalc.calcDist(lastLat, lastLon, thisLat, thisLon);
+                if (segmentLength > WARN_ON_SEGMENT_LENGTH) {
+                    logger.warn("Edge from " + lastLat + "," + lastLon + " to " + thisLat + "," + thisLon + " is " +
+                            (int)segmentLength + "m long, might cause issues with edge sampling.");
```
boldtrn, line 53: I am wondering if we should make this warning a bit more explicit / self-explaining or maybe add a note what the user can do about it?

Crossing the date line should not happen, since we check isCrossBoundary a few lines above.

What are other cases where this could happen? For ferries, this might be an issue. For regular roads, not having a node every few kms could indicate a mapping error. For OSM this could mean to double check the roads at the given coordinates and maybe align the roads. 

I haven't tested the PR yet, are there many of these logs? I could imagine that for example in Canada or Australia, this could be an issue.

core/src/main/java/com/graphhopper/reader/dem/EdgeSampling.java
boldtrn: Yes I think this is a good decision. I did something like this using the great circle. This is quite computation heavy and probably not necessary in most cases.

core/src/main/java/com/graphhopper/reader/dem/EdgeSampling.java
```diff
@@ -0,0 +1,72 @@
...
+        PointList output = new PointList(input.getSize() * 2, input.is3D());
+        if (input.isEmpty()) return output;
+        int nodes = input.getSize();
+        double lastLat = input.getLat(0), lastLon = input.getLon(0), thisLat, thisLon, thisEle;
+        for (int i = 0; i < nodes; i++) {
+            thisLat = input.getLat(i);
+            thisLon = input.getLon(i);
+            thisEle = input.getEle(i);
+            if (i > 0 && !distCalc.isCrossBoundary(lastLon, thisLon)) {
+                double segmentLength = distCalc.calcDist(lastLat, lastLon, thisLat, thisLon);
```
boldtrn, line 50: Shouldn't this use the 3D distance?

boldtrn:
I gave this PR a try. I compared the results with an instance that doesn't use the resolution improvement, I used the following settings:
```
  graph.elevation.provider: skadi
  graph.elevation.interpolate: bilinear
  graph.elevation.long_edge_sampling_distance: 30
  graph.elevation.way_point_max_distance: 1
## -------------------------------------------------------------------
  graph.elevation.provider: skadi
  graph.elevation.interpolate: bilinear
  # Disable both, I debugged the creation in GraphHopper#init, Double.MAX_VALUE gets picked up correctly
  graph.elevation.long_edge_sampling_distance: Double.MAX_VALUE
  graph.elevation.way_point_max_distance: Double.MAX_VALUE

```

A major issue I saw is that while the elevation graph contains more details, the overall ascend and descend increases dramatically as well. 

For example in Baden-Würtemberg (Germany)

http://localhost:8989/maps/?point=48.037463%2C8.622894&point=49.135901%2C9.30542&locale=de-DE&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale

↗1514m ↘2013m => ↗2101m ↘2601m (150km)

Screenshot from 2020-05-05 16-24-36
Screenshot from 2020-05-05 16-24-24

http://localhost:8989/maps/?point=48.019324%2C7.756348&point=48.341646%2C9.898682&locale=de-DE&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale

↗3135m ↘2834m => ↗4828m ↘4523m (300km)

http://localhost:8989/maps/?point=47.702368%2C9.541626&point=49.088258%2C8.640747&locale=de-DE&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale

↗3351m ↘3611m => ↗4322m ↘4583m (251km)

The reason for the difference is most probably, that I compared car routes that contain a lot of straight sections, before we might have removed many points along the polyline. Now we add extra points, which will also increase the elevation overall.

For a rather curvy mountainous segment the difference is a lot smaller, but the elevation increased:

http://localhost:8989/maps/?point=47.97803%2C7.847929&point=47.836263%2C7.937236&locale=de-DE&vehicle=car&weighting=fastest&elevation=true&use_miles=false&layer=Omniscale

↗1104m ↘682m => ↗1150m ↘727m (27km)

When you compare this segment with:
- Google Maps, shows ↗886m ↘468m
- Here WeGo, shows ↗928m ↘514m
- Komoot, shows ↗920m ↘510m
- GraphHopper Maps using CGIAR data ↗1063m ↘637m

BTW: I am not saying that the other routers are correct, but in my experience GraphHopper tends to overestimate the elevation compared to other routers, so I am a bit concerned when the elevation is increased even more. 

Not sure if you did some more research on this?

core/src/main/java/com/graphhopper/reader/dem/EdgeSampling.java
msbarry: I haven't run this on the entire world so I don't know if it actually comes up. I did not see it running in the US or Germany.  A few options I could see are:
- handle this more gracefully by falling back to great circle interpolation in this case (or when crossing the international date line)
- not interpolating at all if the edge is too long
- or telling the user to add intermediate points to fix data in OpenStreetMap

If we can find an example, it would be good to see what it looks like on a few map providers - routes returned by GraphHopper should align with maps they might be displayed on.

boldtrn: Yes I agree, having an example or two would be great. Ok so I think we should keep this in mind and see if any issues pop up when importing a planet file. I will run a test on Australia to see if there are any such warnings popping up.

boldtrn:
I tried to import Australia (same config as above, latest downloaded just now from Geofabrik) with this PR and received an Exception:

```
2020-05-06 09:01:30.647 [main] ERROR com.graphhopper.reader.osm.OSMReader - Couldn't properly add edge with osm ids:[-9223372036854775231, 253003953, 253003952, 253003951, 253003950, 253003949, 253003948, 253003947, 253003921]
java.lang.IllegalStateException: maximum not found in [0,9]
	at com.graphhopper.util.DouglasPeucker.subSimplify(DouglasPeucker.java:151)
	at com.graphhopper.util.DouglasPeucker.simplify(DouglasPeucker.java:104)
	at com.graphhopper.util.DouglasPeucker.simplify(DouglasPeucker.java:81)
	at com.graphhopper.util.DouglasPeucker.simplify(DouglasPeucker.java:77)
	at com.graphhopper.reader.osm.OSMReader.addEdge(OSMReader.java:703)
	at com.graphhopper.reader.osm.OSMReader.addOSMWay(OSMReader.java:648)
	at com.graphhopper.reader.osm.OSMReader.processWay(OSMReader.java:404)
	at com.graphhopper.reader.osm.OSMReader.writeOsmToGraph(OSMReader.java:274)
	at com.graphhopper.reader.osm.OSMReader.readGraph(OSMReader.java:150)
```

This seems to related to this [way](https://www.openstreetmap.org/way/23366648). Not sure what happened with the first OSMID `-9223372036854775231`. I didn't dig deeper so far, but will try to find out more specifics.

msbarry:
@boldtrn thanks for running those tests!

> A major issue I saw is that while the elevation graph contains more details, the overall ascend and descend increases dramatically as well.

That is exactly the problem I'm trying to solve in my app with this change. My app currently samples up to 512 points along each route and I need to show this warning to users for overall ascent/descent over long routes since users have reported ascent/descent reported by GPS watches is substantially larger. 

image

The expectation with this change is that graphhopper will report larger elevation gains overall since it takes into accounts all of the bumps along the way. That will also amplify any noise/errors but from what I've seen though the skadi data has been fairly smooth.

I tried running a similar test on a shorter (4km) route to analyze the differences between providers more closely http://localhost:9898/maps/?point=48.221124%2C8.247481&point=48.237374%2C8.264079&locale=de-DE&vehicle=car&weighting=fastest&elevation=true&turn_costs=false&use_miles=false&layer=Omniscale

- local graphhopper with this change ↗435m ↘51m
- graphhopper.com without this change ↗430m ↘63m, some sudden ascents/descents in the profile, probably from crossing into a neighboring pixel without bilinear interpolation
- google maps ↗376m ↘0m *but* I have an old version of my site running that samples the same route using google maps elevation service and it reports ↗421m ↘47m so I suspect that the elevation profiels they report for routes are simplified quite a bit
- komoot ↗370m ↘10m looks similar overall, but digging into some of the local details that skadi/google elevation show and komoot misses I see a section where google, skadi, and topographic lines on komoot show elevation dropping from ~780 to ~760 (https://onthegomap.com/s/5bi93tvc) but komoot shows elevation steady at 760 +/-2m.

> BTW: I am not saying that the other routers are correct, but in my experience GraphHopper tends to overestimate the elevation compared to other routers, so I am a bit concerned when the elevation is increased even more.
> 
> Not sure if you did some more research on this?

I tested this out on bike rides and runs that I've done in northeastern US. The elevation profiles matched my expectation better, but I didn't compare exact ascent/descent amounts.  I suspect all of the major route providers are underestimating elevation change, possibly in a tradeoff to show less noise at the expense of losing real details? One source of graphhopper overestimating in the past could be that without bilinear interpolation some sampling noise was mixing in?

Another thought I had to reduce noise would be additional interpolation modes that take into account more of the local neighborhood than bilinear interpolation does to get an extra level of smoothness, but that could be decoupled from this change.

msbarry:
> I tried to import Australia (same config as above, latest downloaded just now from Geofabrik) with this PR and received an Exception:
> 
> ```
> 2020-05-06 09:01:30.647 [main] ERROR com.graphhopper.reader.osm.OSMReader - Couldn't properly add edge with osm ids:[-9223372036854775231, 253003953, 253003952, 253003951, 253003950, 253003949, 253003948, 253003947, 253003921]
> java.lang.IllegalStateException: maximum not found in [0,9]
> 	at com.graphhopper.util.DouglasPeucker.subSimplify(DouglasPeucker.java:151)
> 	at com.graphhopper.util.DouglasPeucker.simplify(DouglasPeucker.java:104)
> 	at com.graphhopper.util.DouglasPeucker.simplify(DouglasPeucker.java:81)
> 	at com.graphhopper.util.DouglasPeucker.simplify(DouglasPeucker.java:77)
> 	at com.graphhopper.reader.osm.OSMReader.addEdge(OSMReader.java:703)
> 	at com.graphhopper.reader.osm.OSMReader.addOSMWay(OSMReader.java:648)
> 	at com.graphhopper.reader.osm.OSMReader.processWay(OSMReader.java:404)
> 	at com.graphhopper.reader.osm.OSMReader.writeOsmToGraph(OSMReader.java:274)
> 	at com.graphhopper.reader.osm.OSMReader.readGraph(OSMReader.java:150)
> ```
> 
> This seems to related to this [way](https://www.openstreetmap.org/way/23366648). Not sure what happened with the first OSMID `-9223372036854775231`. I didn't dig deeper so far, but will try to find out more specifics.

Good catch - looks like 3d douglas peucker doesn't handle start/end points being the same. I'll add some tests for this case and fix.

api/src/main/java/com/graphhopper/util/DistanceCalcEarth.java
```diff
@@ -149,6 +149,8 @@ public double calcNormalizedEdgeDistance3D(double r_lat_deg, double r_lon_deg, d
 
         double norm = delta_lon * delta_lon + delta_lat * delta_lat + delta_ele * delta_ele;
         double factor = ((r_lon - a_lon) * delta_lon + (r_lat - a_lat) * delta_lat + (r_ele - a_ele) * delta_ele) / norm;
+        if (factor < 0 || Double.isNaN(factor)) factor = 0;
```
msbarry, line 152: @boldtrn this fixed the Australia issue and fixed some unit tests I added, I saw that other code was not trimming segments (calcNormalizedEdgeDistance sets reduceToSegment to false) - is there a reason to retain that behavior here and only handle the NaN case?

core/src/main/java/com/graphhopper/reader/dem/EdgeSampling.java
boldtrn: Ok, I cound find some instances of this warning in Australia :).

```
2020-05-06 13:58:47.635 [main] WARN  c.g.reader.dem.EdgeSampling - Edge from -32.68276041782112,120.69303365110682 to -32.87841862913412,120.4256310036547 is 33140m long, might cause issues with edge sampling.
2020-05-06 14:00:00.221 [main] WARN  c.g.reader.dem.EdgeSampling - Edge from -32.99502934504366,120.25418606223582 to -32.99459069210121,119.94914326703123 is 28448m long, might cause issues with edge sampling.
2020-05-06 14:01:07.068 [main] WARN  c.g.reader.dem.EdgeSampling - Edge from -33.58228318949677,139.9640097833576 to -33.297097123349396,140.12294594473272 is 34972m long, might cause issues with edge sampling.
2020-05-06 14:01:21.906 [main] WARN  c.g.reader.dem.EdgeSampling - Edge from -28.882821363571765,130.60277563935762 to -28.853342581987587,130.90639108955244 is 29746m long, might cause issues with edge sampling.
2020-05-06 14:01:27.928 [main] WARN  c.g.reader.dem.EdgeSampling - Edge from -27.22052787737238,153.12535546255162 to -27.079826267357756,153.36294665998847 is 28238m long, might cause issues with edge sampling.
2020-05-06 14:01:55.286 [main] WARN  c.g.reader.dem.EdgeSampling - Edge from -14.098557027397089,136.0286709523649 to -13.853749011168235,136.18335264585954 is 31930m long, might cause issues with edge sampling.
``` 

Here are some of the ways where this applies:
- https://www.openstreetmap.org/way/29477902
- https://www.openstreetmap.org/way/731710835
- https://www.openstreetmap.org/way/544311447

My feeling would be, since there are not many of these edges it would be valid to simply ignore these cases and not interpolating there. I could no add a point for these ways and slightly improve the following of the original track, using satellite imagery, where I can see it, but it's probably not worth it. WDYT?

boldtrn:
Awesome thanks, I can confirm, Australia is working fine now.

api/src/main/java/com/graphhopper/util/DistanceCalcEarth.java
boldtrn: Thanks for finding this. I am not sure if we still need this. If I understand this right, the option was used to remain backward compatibility. WDYT @karussell?

boldtrn:
Thanks for the detailed feedback.

> The expectation with this change is that graphhopper will report larger elevation gains overall since it takes into accounts all of the bumps along the way. That will also amplify any noise/errors but from what I've seen though the skadi data has been fairly smooth.

One thing I have seen some times in the past is, if a way leads along a ridge / cliff, the elevation would be picked sometimes from on top of the ridge and sometimes from the lower area, which adds a lot of incorrect noise. I am not sure how much this changed with the bilinear smoothing.

A good example where you can see the issue quite well is the Danube valley. The road more or less follows the river, so it's pretty much flat. There is no serious elevation along this road. 

http://localhost:8989/maps/?point=48.079369%2C9.191995&point=48.065634%2C9.001451&locale=de-DE&elevation=true&vehicle=car&weighting=fastest&turn_costs=false&use_miles=false&layer=TF%20Outdoors

This PR reports ↗589m ↘555m, Komoot reports ↗170m ↘150m. Komoot is definitely closer to the truth at this point. We do have a video of the Danube valley on the Kurviger Youtube channel, there you can see that it's pretty flat (the last sequence is up a hill that is not part of the above route). In the video it's also easy to see that big parts of the route lead along very steep cliffs (e.g. at 3:00).

https://www.youtube.com/watch?v=IqjvdjtQclc

That said, with the bilinear smoothing this issue has been massively improved. This was a lot worse in the past.

> Have reported ascent/descent reported by GPS watches is substantially larger.

I have no GPS watch, so I don't have any experience here. But could it be that the GPS watches are summing up errors as well? For example if we assume that the GPS has an average error of 8m, this could sum up quite fast. If they also use barometric pressure this might introduce some issues as well.

I remember a blog post from Strava, but cannot find it right now I could only find some info that they will smooth the input data to reduce noise. But I found this post which looks interesting as well, showing how errors used to sum up from GPS sensors: http://regex.info/blog/2015-05-09/2568

In any case, I think we need two things:
- Filtering noise from the elevation calculation
- Improving the resolution

I think both are important and somewhat contrary, I think we need a bit of both :).

boldtrn:
Maybe we should move the issue on how the elevation is summed up into a different issue?

By default, we could set the sampling to (just an example):
``` 
-Ddw.graphhopper.graph.elevation.long_edge_sampling_distance=100
-Ddw.graphhopper.graph.elevation.way_point_max_distance=5
``` 

Using that setting, I cannot see big differences in the overall ascend/descend, but for big gaps (>100m) in ways we would fill these gaps to not miss any serious elevation changes. Everyone who needs a higher resolution, could set it up that way.

I think we should run some more tests to find a good default value.

WDYT does that make sense?

api/src/main/java/com/graphhopper/util/DistanceCalcEarth.java
karussell: IMO we can remove the reduceToSegment parameter as it seems that only reduceToSegment==false is used. 

The option was introduced several years ago in this commit: https://github.com/graphhopper/graphhopper/commit/98e22f449f17a06db20296659b64011dbca9118e#diff-381813cce27dac5fde2afa44aec5bb13L164
but wasn't used it seems. Ugly.

msbarry:
> Maybe we should move the issue on how the elevation is summed up into a different issue?
> 
> By default, we could set the sampling to (just an example):
> 
> ```
> -Ddw.graphhopper.graph.elevation.long_edge_sampling_distance=100
> -Ddw.graphhopper.graph.elevation.way_point_max_distance=5
> ```
> 
> Using that setting, I cannot see big differences in the overall ascend/descend, but for big gaps (>100m) in ways we would fill these gaps to not miss any serious elevation changes. Everyone who needs a higher resolution, could set it up that way.
> 
> I think we should run some more tests to find a good default value.
> 
> WDYT does that make sense?

I think that makes sense, `graph.elevation.way_point_max_distance` is the knob we have tune down sensitivity and filter-out noise, given the points you raise 5m seems like a reasonable default value, and also keeps edge geometry size increase negligible. For the Danube case, it looks like FWIW, Google maps elevation profile simplifies it down to just 7 elevation points over the 20km route.

I could see how pure bilinear interpolation might make it think that points near a slope like in the Danube case are actually on the slope, I'd be curious to experiment with other interpolation methods like bicubic that might be able to infer more accurately where a slope starts between elevation samples

image

@boldtrn do you want me to set the default values to sample distance=100/elevation max dist=5 in this review and open up a separate issue for smoothing elevation noise?

core/src/main/java/com/graphhopper/reader/dem/EdgeSampling.java
msbarry: Looking into those, the difference between the road that graphhopper shows with edge sampling the difference appears to be negligible.  I'll bump the threshold out to 1 degree instead of 1/4 degree, improve the error message,  and fall back to not interpolating in this case

boldtrn: Ok, sounds good to me, thanks!

> Looking into those, the difference between the road that graphhopper shows with edge sampling the difference appears to be negligible.

How did you compare those? Visually using Leaflet? Leaflet by default uses a linear approach as well. There is a plugin for Leaflet: https://github.com/henrythasler/Leaflet.Geodesic

Or you could also use the GPX files and import them on Google MyMaps, they also use the great circle to connect two points.

msbarry: I switched to satellite view on local graphhopper web, the path was off by a couple of road widths in the middle.  I dug into the great circle approach, it wasn't too complex so I just implemented that for segments longer than 1/4 degree. Let me know what you think, it is based on the same source as `projectCoordinate` and seems to hold up pretty well in unit tests.

msbarry:
@boldtrn so to sum up here it sounds like the path forward is to:

- try different edge sampling settings to find one that gives overall elevation comparable to other providers and enable that by default
- confirm that great circle interpolation on long points is doing what we want on those australia edges
- open a new issue for smoothing noise in elevation profiles/ascent/descent calculations

?

Also, the elevation profile seems to have disappeared from my local graphhopper web UI - is there something special you need to always have in config.yml to make that show up?

karussell:
> Also, the elevation profile seems to have disappeared from my local graphhopper web UI - is there something special you need to always have in config.yml to make that show up?

Unfortunately this is a bug, likely related to #1952, but will have a closer look as it should be involved with the profiles changes.

boldtrn:
> FWIW, Google maps simplified the elevation for the Danube profile to just 7 elevation points over the 20km route.

Are these the dots shown in the image below? How do you find this out?

> I'd be curious to experiment with other interpolation methods like bicubic that might be able to infer more accurately where a slope starts between elevation samples

Yes indeed, that could help in this situation.

> so to sum up here it sounds like the path forward is to

Yes that sounds good to me.

> enable that by default

I am not 100% sure about enabling it by default or just setting these values as default, when it's enabled, like through config-example. If we enable it by default, we should probably also enable bilinear smoothing by default? WDYT? @karussell?

msbarry:
> Are these the dots shown in the image below? How do you find this out?

Those are the elevation sample locations, I used https://turfjs.org/docs/#pointGrid to overlay where they would be based on 3600 samples per degree.

msbarry:
OK a couple of updates here:
- I got the UI missing elevation issue fixed by building assets locally and linking them in, thanks @karussell !
- great circle interpolation appears to work on those Australia segments, although paths do get off from the satellite view of the roads in the middle and likely always will unless someone drops a few intermediate points
- I tried bicubic interpolation, it reports larger total ascent/descent over bilinear, probably because the splines overshoot elevation samples - so I don't think that's a solution to help out here.
- I tried a bunch of variations on sample distance and elevation max distance. For a default value I'd probably stick to <100m sampling so we are getting more data and rely more on the elevation tolerance (3-10m) to smooth out noise.

image
(green/red are difference relative to baseline CGIAR without interpolation)

Also, digging into some of the provider discrepancies issues I'm seeing cases where every contour map shows the road going up a hill, but it appears the road is actually flat there (and komoot/google don't show elevation change). 

Screen Shot 2020-05-07 at 4 28 20 PM

I think the only ways to correct for this would either be to post-process the way geometries to smooth out elevation or if you had enough user-provided GPS data snapped to OSM ways, you could use that to fix issues in elevation like in the article @boldtrn shared.  Only the former would be practical for the open source project. It looks like the `graph.elevation.smoothing` option @boldtrn added in #1220 reduces elevation changes that are reported, but it introduces some sharp ascent/descent artifacts to the profile that would be not work at least for my use-case. Do you think the new issue should be around improving that smoothing code? or a completely different approach?

otbutz:
Are there any side effects for the existing elevation interpolation used for bridges and tunnels?

otbutz:
Maybe we should avoid smoothing the elevation for all points in GraphElevationSmoothing and only use this approach if a certain slope threshold is exceeded?

karussell:
Good catch with the GraphElevationSmoothing (@boldtrn will know more) but EdgeElevationInterpolator (bridges/tunnels) is unrelated here. For this we only want to "get rid" of the elevation data.

msbarry:
I submitted #2028 to track improved elevation profile smoothing.  Let me know what else you'd like me to change on this review before merging.

boldtrn:
> great circle interpolation appears to work on those Australia segments, although paths do get off from the satellite view of the roads in the middle and likely always will unless someone drops a few intermediate points. Let me know what you think about the implementation.

Yes, I think this is the expected behavior. So no worries here :).

> I tried bicubic interpolation, it reports larger total ascent/descent over bilinear, probably because the splines overshoot elevation samples - so I don't think that's a solution to help out here.

That's a pity, but thanks for trying. I was hoping this might help in this situation.

> Also, digging into some of the provider discrepancies issues I'm seeing cases where every contour map shows the road going up a hill, but it appears the road is actually flat there (and komoot/google don't show elevation change).

That is actually quite an interesting find. Do you know why this could happen? Maybe the contour lines are approximated already? Therefore, it might not be possible to fully trust them?

Thanks for creating the new issue :).

otbutz:
> That is actually quite an interesting find. Do you know why this could happen? Maybe the contour lines are approximated already? Therefore, it might not be possible to fully trust them?

Might be caused by the interpolation in the elevation provider. @msbarry you could try to disable it for your tests

msbarry:
> How about we go with:
> 
> ```
>   graph.elevation.long_edge_sampling_distance: 60
>   graph.elevation.way_point_max_distance: 10
> ```
> 
> These would be configurable anyway, but results for these look good straight away if anyone wants to activate this feature. WDYT?

Sounds good to me, I set those default values in the comments in `config-example.yml`.

> > That is actually quite an interesting find. Do you know why this could happen? Maybe the contour lines are approximated already? Therefore, it might not be possible to fully trust them?
> 
> Might be caused by the interpolation in the elevation provider. @msbarry you could try to disable it for your tests

I don't think it's an issue sampling the actual DEM file since this based on where the samples are located, there should be enough resolution to separate the river from the hill around that curve
image

It's probably an irregularity with the DEM data provider. According to https://github.com/tilezen/joerd/blob/master/docs/data-sources.md, skadi uses [EU-DEM](https://www.eea.europa.eu/data-and-maps/data/copernicus-land-monitoring-service-eu-dem) for most of Europe which is "a hybrid product based on SRTM and ASTER GDEM data fused by a weighted averaging approach." So it could be artifacts from either in the SRTM/ASTER -> EU-DEM conversion or EU-DEM to skadi conversion.

msbarry:
Thanks for the review @boldtrn! @karussell any more changes you want me to make here?

msbarry:
Congrats on the 1.0 release! Think this might be safe to merge soon @karussell ?

karussell:
Thanks a lot @msbarry for your work here! Also thanks @boldtrn for the review!

As I understand it correctly an existing application is not affected or will it be affected if it already uses interpolation?

karussell:
@otbutz 

> Are there any side effects for the existing elevation interpolation used for bridges and tunnels?

I don't think so. Basically what we do for bridges and tunnels is to get rid of elevation.

> Edit: if i'm not mistaken, we interpolate elevation at three different locations now?

Yes, but at least the bridge/tunnel part is a different purpose.

web/src/main/resources/assets/js/main.js
```diff
@@ -0,0 +1,284 @@
+(function(){function r(e,n,t){function o(i,f){if(!n[i]){if(!e[i]){var c="function"==typeof require&&require;if(!f&&c)return c(i,!0);if(u)return u(i,!0);var a=new Error("Cannot find module '"+i+"'");throw a.code="MODULE_NOT_FOUND",a}var p=n[i]={exports:{}};e[i][0].call(p.exports,function(r){var n=e[i][1][r];return o(n||r)},p,p.exports,r,e,n,t)}return n[i].exports}for(var u="function"==typeof require&&require,i=0;i<t.length;i++)o(t[i]);return o}return r})()({1:[function(require,module,exports){
```
karussell, line 1: This file needs to be removed as it is now auto-created.

msbarry:
> As I understand it correctly an existing application is not affected or will it be affected if it already uses interpolation?

Yes, existing applications should not be affected whether they use bilinear interpolation or not since the 2 new config parameters default to Double.MAX_VALUE which retains existing behavior.

msbarry:
> Now I also better understand your images in [#1953 (comment)](https://github.com/graphhopper/graphhopper/pull/1953#issuecomment-625774030), i.e. red is the only area where the incline and decline increase ("got worse"). So this PR with the proposed settings in config-example.yml is always an improvement and should decrease incline&decline? (over CGIAR and interpolation only)

Those charts show increase/decrease relative to CGIAR without interpolation. 10m/60m plus bilinear interpolation proposed in config-example.yml should report an "improvement" with less total elevation gain/loss than people are used to with CGIAR and no interpolation today.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -711,7 +709,7 @@ protected DataReader initDataReader(DataReader reader) {
                 setElevationProvider(eleProvider).
                 setWorkerThreads(dataReaderWorkerThreads).
                 setWayPointMaxDistance(dataReaderWayPointMaxDistance).
-                setWayPointElevationMaxDistance(elevationWayPointMaxDistance).
```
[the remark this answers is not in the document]
msbarry: We could also make elevation way point max distance controllable through a request parameter for routing?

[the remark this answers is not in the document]
easbar: Thanks a lot for this very nice PR @msbarry! 

As far as I can tell the `elevationWayPointMaxDistance` is used to control how many extra pillar nodes get created (to preserve elevation changes of the geometries) and it is also used during path simplification when running routing requests (to not throw away way points needed to reflect elevation changes). These two usages are different in that the first only can have an effect during import (after the first import it becomes meaningless unless we rebuild the graph) while the other could (at least in theory) be changed for every single request. Is that correct? For this reason I think it would be cleaner to use two parameters, e.g. one would go into the (not yet existing) `ImporterConfig` and the other would go to `RouterConfig` and the latter could also be changed via the routing hints in `Router#createPathMerger`. I just wonder if this is overkill and I do not think its urgent to separate this into two parameters already. In any case the parameter used for the import is a good default value for the one used during routing?

> Should we still keep the two variables?
Which two variables do you mean? I thought so far there is only one (only one that has to do with import and routing).

> We could also make elevation way point max distance controllable through a request parameter for routing?

Yes, exactly, in `Router#createPathMerger`.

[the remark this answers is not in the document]
msbarry: Thanks @easbar, longEdgeSamplingDistance controls how many extra pillar nodes are added and elevationWayPointMaxDistance prevents them from getting removed during simplification (on import, and during routing).  I added a new elevation_way_point_max_distance request parameter in https://github.com/graphhopper/graphhopper/pull/1953/commits/9edbfbf2002b3309ef816ed7ac6777aeaa5d5028 to let you change elevationWayPointMaxDistance on each request or omit to default to what was used during import.  I agree it's probably overkill to split out separate config parameters for data reader import and router default. Let me know if that change looks sufficient.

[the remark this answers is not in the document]
easbar: Yes thanks I like it.

karussell:
Thanks!

PR #1954, opened 2020-03-13
use enum instead of int

karussell:
@easbar as discussed in #1950 :)


core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -968,14 +968,14 @@ private PointList fetchWayGeometry_(long edgePointer, boolean reverse, int mode,
...
 
-        PointList pillarNodes = new PointList(count + mode, nodeAccess.is3D());
+        PointList pillarNodes = new PointList(FetchWayGeometry.count(count, mode), nodeAccess.is3D());
         if (reverse) {
-            if ((mode & 2) != 0)
+            if (mode == FetchWayGeometry.ALL || mode == FetchWayGeometry.PILLAR_AND_ADJ)
                 pillarNodes.add(nodeAccess, adjNode);
-        } else if ((mode & 1) != 0)
+        } else if (mode == FetchWayGeometry.ALL || mode == FetchWayGeometry.BASE_AND_PILLAR)
             pillarNodes.add(nodeAccess, baseNode);
```
karussell: This is a bit more lengthy now ;)

core/src/main/java/com/graphhopper/util/FetchWayGeometry.java
```diff
@@ -0,0 +1,20 @@
+package com.graphhopper.util;
+
+public enum FetchWayGeometry {
```
easbar: I do not really like the enum name `FetchWayGeometry` too much. Its not a noun and it does not really tell us anything. Maybe we can use this name to communicate that this enum is used to specify how many nodes we fetch? For example `GeometryNodes.ALL`? `FetchMode.ALL`? `WayGeometryFetchMode.ALL`?

core/src/main/java/com/graphhopper/storage/BaseGraph.java
easbar: IMO its more lengthy and a lot more readable. With the previous code I did not understand what happened here without looking up what e.g. mode=2 means.

core/src/main/java/com/graphhopper/util/FetchWayGeometry.java
```diff
@@ -0,0 +1,20 @@
...
+    TOWER_ONLY, PILLAR_ONLY, BASE_AND_PILLAR, PILLAR_AND_ADJ, ALL;
+
+    public static int count(int count, FetchWayGeometry mode) {
```
easbar: The names here could be more expressive. `int count(int count` does not tell me anything about what this method does. Apparently the method returns the total number of nodes that will be returned by `fetchWayGeometry` given the number of pillar nodes? So maybe we can at least rename the `count` parameter to `pillarNodes`? Maybe rename the method to `getFetchedNodeCount` or `countFetchedNodes`?

core/src/main/java/com/graphhopper/util/FetchWayGeometry.java
karussell: Yes, FetchMode sounds already good.

core/src/main/java/com/graphhopper/util/FetchWayGeometry.java
karussell: I'll probably move the method into the only class where it is being used then we do not need to care much about the name for now :) (?)

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -993,16 +993,31 @@ private PointList fetchWayGeometry_(long edgePointer, boolean reverse, int mode,
...
 
             pillarNodes.reverse();
-        } else if ((mode & 2) != 0)
+        } else if (mode == FetchMode.ALL || mode == FetchMode.PILLAR_AND_ADJ)
             pillarNodes.add(nodeAccess, adjNode);
 
         return pillarNodes;
     }
 
+    static int getPointListLength(int count, FetchMode mode) {
```
easbar: I'd still change `count` to `pillarNodes` (if thats what it is?)

PR #1956, opened 2020-03-14
refactor PMap to use string,object instead of string,string

karussell:
see discussion https://github.com/graphhopper/graphhopper/pull/1949#issuecomment-598701749 with @easbar .

This is cleaner than #1955:

 * For `put(String, String)` the user gets a deprecation notice
 * For `put(String, Object)` the user gets a compilation error and is forced to use putObject
 * get("key") is removed
 * ~get("key", "default") breaks backward compatibility if the value for "key" is a number. This is acceptable IMO as this usage should be rare except maybe for debugging.~ no longer the case as renamed to getString.

parsing parameters from URL should still work due to Helper.toObject(String)

### DIFF
--- a/api/src/main/java/com/graphhopper/util/PMap.java
+++ b/api/src/main/java/com/graphhopper/util/PMap.java
@@ -18,10 +18,7 @@
 package com.graphhopper.util;
 
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.Map;
-
-import static com.graphhopper.util.Helper.toLowerCase;
 
 /**
  * A properties map (String to String) with convenient accessors
@@ -66,7 +63,7 @@ public class PMap {
      * Reads a PMap from a string array consisting of key=value pairs
      */
     public static PMap read(String[] args) {
-        Map<String, Object> map = new LinkedHashMap<>();
+        PMap map = new PMap();
         for (String arg : args) {
             int index = arg.indexOf("=");
             if (index <= 0) {
@@ -83,12 +80,12 @@ public class PMap {
             }
 
             String value = arg.substring(index + 1);
-            Object old = map.put(toLowerCase(key), value);
+            Object old = map.map.put(Helper.camelCaseToUnderScore(key), Helper.toObject(value));
             if (old != null)
-                throw new IllegalArgumentException("Pair '" + toLowerCase(key) + "'='" + value + "' not possible to " +
+                throw new IllegalArgumentException("Pair '" + Helper.camelCaseToUnderScore(key) + "'='" + value + "' not possible to " +
                         "add to the PMap-object as the key already exists with '" + old + "'");
         }
-        return new PMap(map);
+        return map;
     }
 
     public PMap put(PMap map) {
--- a/api/src/main/java/com/graphhopper/util/PMap.java
+++ b/api/src/main/java/com/graphhopper/util/PMap.java
@@ -30,7 +30,7 @@ import static com.graphhopper.util.Helper.toLowerCase;
  * @author Peter Karich
  */
 public class PMap {
-    private final Map<String, String> map;
+    private final Map<String, Object> map;
 
     public PMap() {
         this(5);
@@ -40,7 +40,7 @@ public class PMap {
         this(new HashMap<>(capacity));
     }
 
-    public PMap(Map<String, String> map) {
+    public PMap(Map<String, Object> map) {
         this.map = new HashMap<>(map);
     }
 
@@ -66,7 +66,7 @@ public class PMap {
      * Reads a PMap from a string array consisting of key=value pairs
      */
     public static PMap read(String[] args) {
-        Map<String, String> map = new LinkedHashMap<>();
+        Map<String, Object> map = new LinkedHashMap<>();
         for (String arg : args) {
             int index = arg.indexOf("=");
             if (index <= 0) {
@@ -83,7 +83,7 @@ public class PMap {
             }
 
             String value = arg.substring(index + 1);
-            String old = map.put(toLowerCase(key), value);
+            Object old = map.put(toLowerCase(key), value);
             if (old != null)
                 throw new IllegalArgumentException("Pair '" + toLowerCase(key) + "'='" + value + "' not possible to " +
                         "add to the PMap-object as the key already exists with '" + old + "'");
@@ -96,124 +96,72 @@ public class PMap {
         return this;
     }
 
-    public PMap put(String key, Object str) {
+    /**
+     * @deprecated use {@link #putObject(String, Object)} instead
+     */
+    public PMap put(String key, String str) {
         if (str == null)
             throw new NullPointerException("Value cannot be null. Use remove instead.");
-
-        // store in under_score
-        map.put(Helper.camelCaseToUnderScore(key), str.toString());
+        map.put(Helper.camelCaseToUnderScore(key), Helper.toObject(str));
         return this;
     }
 
     public PMap remove(String key) {
         // query accepts camelCase and under_score
-        map.remove(Helper.camelCaseToUnderScore(key));
+        map.remove(key);
         return this;
     }
 
     public boolean has(String key) {
         // query accepts camelCase and under_score
-        return map.containsKey(Helper.camelCaseToUnderScore(key));
-    }
-
-    public long getLong(String key, long _default) {
-        String str = get(key);
-        if (!Helper.isEmpty(str)) {
-            try {
-                return Long.parseLong(str);
-            } catch (Exception ex) {
-            }
-        }
-        return _default;
-    }
-
-    public int getInt(String key, int _default) {
-        String str = get(key);
-        if (!Helper.isEmpty(str)) {
-            try {
-                return Integer.parseInt(str);
-            } catch (Exception ex) {
-            }
-        }
-        return _default;
+        return map.containsKey(key);
     }
 
     public boolean getBool(String key, boolean _default) {
-        String str = get(key);
-        if (!Helper.isEmpty(str)) {
-            try {
-                return Boolean.parseBoolean(str);
-            } catch (Exception ex) {
-            }
-        }
-        return _default;
+        Object object = map.get(key);
+        return object instanceof Boolean ? (Boolean) object : _default;
     }
 
-    public double getDouble(String key, double _default) {
-        String str = get(key);
-        if (!Helper.isEmpty(str)) {
-            try {
-                return Double.parseDouble(str);
-            } catch (Exception ex) {
-            }
-        }
-        return _default;
+    public int getInt(String key, int _default) {
+        Object object = map.get(key);
+        return object instanceof Number ? ((Number) object).intValue() : _default;
+    }
+
+    public long getLong(String key, long _default) {
+        Object object = map.get(key);
+        return object instanceof Number ? ((Number) object).longValue() : _default;
     }
 
     public float getFloat(String key, float _default) {
-        String str = get(key);
-        if (!Helper.isEmpty(str)) {
-            try {
-                return Float.parseFloat(str);
-            } catch (Exception ex) {
-            }
-        }
-        return _default;
+        Object object = map.get(key);
+        return object instanceof Number ? ((Number) object).floatValue() : _default;
+    }
+
+    public double getDouble(String key, double _default) {
+        Object object = map.get(key);
+        return object instanceof Number ? ((Number) object).doubleValue() : _default;
     }
 
     public String get(String key, String _default) {
-        String str = get(key);
-        if (Helper.isEmpty(str))
-            return _default;
-
-        return str;
+        Object object = map.get(key);
+        return object instanceof String ? (String) object : _default;
     }
 
-    String get(String key) {
-        if (Helper.isEmpty(key))
-            return "";
+    public Object getObject(String key, Object _default) {
+        Object object = map.get(key);
+        return object == null ? _default : object;
+    }
 
-        // query accepts camelCase and under_score
-        String val = map.get(Helper.camelCaseToUnderScore(key));
-        if (val == null)
-            return "";
-
-        return val;
+    public PMap putObject(String key, Object object) {
+        map.put(key, object);
+        return this;
     }
 
     /**
      * This method copies the underlying structure into a new Map object
      */
-    public Map<String, String> toMap() {
+    public Map<String, Object> toMap() {
         return new HashMap<>(map);
-    }
-
-    private Map<String, String> getMap() {
-        return map;
-    }
-
-    public PMap merge(PMap read) {
-        return merge(read.getMap());
-    }
-
-    PMap merge(Map<String, String> map) {
-        for (Map.Entry<String, String> e : map.entrySet()) {
-            if (Helper.isEmpty(e.getKey()))
-                continue;
-
-            put(e.getKey(), e.getValue());
-        }
-        return this;
     }
 
     public boolean isEmpty() {
@@ -222,6 +170,6 @@ public class PMap {
 
     @Override
     public String toString() {
-        return getMap().toString();
+        return map.toString();
     }
 }
--- a/api/src/main/java/com/graphhopper/util/PMap.java
+++ b/api/src/main/java/com/graphhopper/util/PMap.java
@@ -137,7 +137,7 @@ public class PMap {
         return object instanceof Number ? ((Number) object).doubleValue() : _default;
     }
 
-    public String get(String key, String _default) {
+    public String getString(String key, String _default) {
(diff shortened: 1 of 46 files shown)


client-hc/src/main/java/com/graphhopper/api/GHMatrixAbstractRequester.java
```diff
@@ -106,13 +106,13 @@ protected JsonNode createPostRequest(GHMRequest ghRequest, Collection<String> ou
...
-        Map<String, String> hintsMap = ghRequest.getHints().toMap();
+        Map<String, Object> hintsMap = ghRequest.getHints().toMap();
         for (String hintKey : hintsMap.keySet()) {
             if (ignoreSet.contains(hintKey))
                 continue;
 
-            String hint = hintsMap.get(hintKey);
-            requestJson.put(hintKey, hint);
+            Object hint = hintsMap.get(hintKey);
+            requestJson.put(hintKey, hint.toString());
```
easbar: Why do we have to turn this into a string here? Can we not just keep the object? Aren't there possible cases where we want to add the actual object to the json tree instead of converting it into a string? Hm, but this requires object-to-json serialization to create a (sub)JsonNode?

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
```diff
@@ -258,13 +258,26 @@ private Request createPostRequest(GHRequest ghRequest) {
...
+            else if (hint instanceof Integer)
+                requestJson.put(hintKey, (Integer) hint);
+            else if (hint instanceof Long)
+                requestJson.put(hintKey, (Long) hint);
+            else if (hint instanceof Float)
+                requestJson.put(hintKey, (Float) hint);
+            else if (hint instanceof Double)
+                requestJson.put(hintKey, (Double) hint);
+            else
+                requestJson.put(hintKey, hint.toString());
```
easbar: Similar to AbstractMatrixRequester: Can we keep the actual object instead of turning into a string here? But we can probably add this when we really need it. I just wonder in which case the `toString()` implementation will be of actual use (if it cannot be parsed later its useless or might even break the json format (?)).

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
```diff
@@ -341,9 +354,9 @@ private Request createGetRequest(GHRequest ghRequest) {
             url += "&key=" + WebHelper.encodeURL(key);
         }
 
-        for (Map.Entry<String, String> entry : ghRequest.getHints().toMap().entrySet()) {
+        for (Map.Entry<String, Object> entry : ghRequest.getHints().toMap().entrySet()) {
             String urlKey = entry.getKey();
-            String urlValue = entry.getValue();
+            String urlValue = entry.getValue().toString();
```
easbar: Here I also wonder in which cases the `toString()` method really produces something useful. (In this case it must be valid to use as url parameter).

api/src/main/java/com/graphhopper/util/PMap.java
```diff
@@ -83,145 +80,93 @@ public static PMap read(String[] args) {
...
-        String val = map.get(Helper.camelCaseToUnderScore(key));
-        if (val == null)
-            return "";
+    public Object getObject(String key, Object _default) {
+        Object object = map.get(key);
+        return object == null ? _default : object;
+    }
 
-        return val;
+    public PMap putObject(String key, Object object) {
```
easbar: Why not just name this put as well (overload the existing put method)? Because of the camelcase/underscore conversion?

karussell:
I changed `put(String, Object)` into `put(String, String)` to force user moving off this method and use the new `putObject(String, Object)` method instead. As the new method has no camelCase conversion and no "automated deserialization from string".

And I decided against `put(String, Object)` because then we cannot later remove the camelCase conversion and also not the "string to object deserialization" (see new Helper.toObject). I.e. the intend is to move these additional features out of the PMap and force the user to use external methods, if this is necessary at all. The only place where it is necessary is converting the URL parameters from string to object.

The only problem is the `String get(String key, String _default)` that now breaks backward compatibility. E.g. also we used this for EDGE_BASED etc ... maybe I should rename this into getString to force a safe upgrade path?

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
karussell: Ah, you mean with putPOJO ... will try.

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
karussell: As we use WebHelper.encodeURL this should be fine (?)

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
easbar: Yes, maybe, does this use Jackson internally (when it is finally serialized)?

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
easbar: Hm, yeah encoding the url probably replaces invalid characters, but the real question is will this ever be useful on the server side? Sure `toString()` *might* produce something that can be parsed later, but that entirely depends on the object we put in (which seems a bit fragile and forces one to implement `toString()` in a certain way (which I do not like so much)). Then again I do not have a much better idea :) But maybe we can throw an error if the values is not a number and not a string (are there any actual use cases where we want to `toString()` an object into a url parameter?)

easbar:
> to force user moving off this method 

Ah I see this is a bit more critical because `HintsMap` is also used on the client side. Would it make sense to use an extra class for the client side (with camel case conversion etc.) and refactor more aggressively on the server side? There is a similar problem with `setVehicle/Weighting` vs `setProfile` in #1859. 

> The only problem is the String get(String key, String _default) that now breaks backward compatibility.

What is the problem here?

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
```diff
@@ -200,10 +198,10 @@ private String getCHRequestAsString(HintsMap hintsMap, Boolean edgeBased, Intege
     }
 
     private Boolean getEdgeBased(HintsMap hintsMap) {
-        return hintsMap.has(Parameters.Routing.EDGE_BASED) ? hintsMap.getBool(Parameters.Routing.EDGE_BASED, false) : null;
+        return (Boolean) hintsMap.getObject(Parameters.Routing.EDGE_BASED, null);
     }
 
     private Integer getUTurnCosts(HintsMap hintsMap) {
-        return hintsMap.has(Parameters.Routing.U_TURN_COSTS) ? hintsMap.getInt(Parameters.Routing.U_TURN_COSTS, INFINITE_U_TURN_COSTS) : null;
+        return (Integer) hintsMap.getObject(Parameters.Routing.U_TURN_COSTS, null);
```
easbar: Does this not give a NPI in case there is no u_turn_costs entry?

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
karussell: As we use by default a POST request this should not make a problem IMO. Throwing an error is a bit clumsy as we need to check for boolean, short, byte, int, double, ...

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
karussell: If there is no u_turn_cost entry it will return null and casting this to Integer is possible without NPE. So it should return `null` like before.

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
easbar: Hm, but this whole method is about creating the url parameters for a GET request. If we do not use it and only use POST we should rather remove the entire method. Checking for 'number, boolean, string or something else' might take a few lines to do but that should not be a problem. Then again if you think using `toString()` is fine we can just as well keep it :) (I just do not really know in which case it would be useful and do not want to encourage people to adjust their `toString()` implementation to make it work with this url creation code).

karussell:
> Ah I see this is a bit more critical because HintsMap is also used on the client side. Would it make sense to use an extra class for the client side (with camel case conversion etc.) and refactor more aggressively on the server side? 

We can also remove `put(String, String)` now but I thought to keep it for 1 version.

> There is a similar problem with setVehicle/Weighting vs setProfile in #1859

Yes, for client side we might keep vehicle&weighting for a while ... maybe replace "HintsMap" with "PMap" on the server side and keep using HintsMap for a while for client-hc only?

> > The only problem is the String get(String key, String _default) that now breaks backward compatibility.
> What is the problem here?

E.g. `get(EDGE_BASED, "xy")` (boolean value) returned before the actual value as string, but will now return the default `"xy"` because the value is no String instance.

karussell:
> E.g. get(EDGE_BASED, "xy") (boolean value) returned before the actual value as string, but will now return the default "xy" because the value is no String instance.

I will rename this to be on the safe side and if this is really needed one use `getObject(EDGE_BASED, "xy").toString()`

easbar:
> Yes, for client side we might keep vehicle&weighting for a while ... maybe replace "HintsMap" with "PMap" on the server side and keep using HintsMap for a while for client-hc only?

Ok yes that makes sense, HintsMap will then be the 'client-side' PMap? Maybe we can then also use `put(String, String)` only in `HintsMap` (and maybe even rename `putObject` back to `put` on the server side ;)) If its easier it should be no problem to no longer let `HintsMap` extend `PMap`? 

> E.g. get(EDGE_BASED, "xy") (boolean value) returned before the actual value as string,

Ah I see, but why would we use `get(EDGE_BASED, "xy")` instead of `getBool(EDGE_BASED, false)`? So far I think I always tried to use `getBool` if I was expecting a bool and only used `get` to retrieve strings?

karussell:
> Ah I see, but why would we use get(EDGE_BASED, "xy") instead of getBool(EDGE_BASED, false)?

Old code ... see e.g. GraphHopper.java in #1958

easbar:
> Old code ... see e.g. GraphHopper.java in #1958 

Thats not old code, its not even in master yet. And it can/should be replaced with `has(EDGE_BASED)` anyway :)

karussell:
> and maybe even rename putObject back to put on the server side ;)

Ok, but we should delay this for 1 or 2 versions to avoid silent migration issues as e.g. `put("key", "true")` had a different meaning before this PR.

easbar:
> Ok, but we should delay this for 1 or 2 versions to avoid silent migration issues as e.g. put("key", "true") had a different meaning before this PR.

Ok, another problem might be that we are using `GHRequest` on client+server side and it has a `HintsMap`.

karussell:
> And it can/should be replaced with has(EDGE_BASED) anyway :)

What I meant is that there can be code that worked before and will now (silently) break, which is ugly. So, I renamed `get` into `getString` ... which also makes sense when we later rename `putObject` into `put`, then we could rename `getObject` into `get` (?)

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
easbar: I thought casting null gives NPE, but you are right.

easbar:
> What I meant is that there can be code that worked before and will now (silently) break, which is ugly.

Yes you are right that might happen in client code. 

>  So, I renamed `get` into `getString` ... which also makes sense when we later rename `putObject` into `put`, then we could rename `getObject` into `get` (?)

:+1:

easbar:
> Yes you are right that might happen in client code.

This somewhat tells me that it is not a good idea to use 'trivial' classes like `HintsMap` and `GHRequest` both internally and in `client-hc`. It might be better to duplicate these classes to draw a clear boundary between code that is subject to change at any time and code that needs to be kept stable so we do not break external client code.

easbar:
> This is cleaner than #1955 but it slightly breaks backward compatibility:

Maybe update the initial post so it can be referenced later?

api/src/main/java/com/graphhopper/util/PMap.java
```diff
@@ -83,145 +80,91 @@ public static PMap read(String[] args) {
...
     }
 
     public PMap put(PMap map) {
         this.map.putAll(map.map);
         return this;
     }
 
-    public PMap put(String key, Object str) {
+    /**
+     * @deprecated use {@link #putObject(String, Object)} instead
```
easbar: Maybe link to this PR here?

karussell:
> It might be better to duplicate these classes to draw a clear boundary between code that is subject to change at any time and code that needs to be kept stable so we do not break external client code.

The advantage of the client code is that it only breaks existing code if the user upgrades the client dependency. And for a switch from 0.x to 1.0 I would kind of expect that I need to change my usage code. Also now it should be a "safe" upgrade (no silent issues IMO) and the compiler should tell all the problems.

> Maybe update the initial post so it can be referenced later?

Done :+1:

easbar:
Two more things: 

1) We should probably add `@Deprecated` annotation to `PMap#put(String, String)` instead of only `@deprecated` in java docs. Also the deprecation is not inherited by `HintsMap`, so we should put `@Deprecated` there as well?

2) There are still quite a few uses of the deprecated `put` method, shouldn't we get rid of it?

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
easbar: There is another problem with this approach: If the value in `hintsMap` is a String (like `"true"`) this will lead to a `ClassCastException`.

easbar:
And another: We use `putObject` in `GraphHopperConfig#put`. This means something like ` put("routing.ch.disabling_allowed", "true")` is no longer working because it will add `"true"` as String and when we do something like `config.getBool("routing.ch.disabling_allowed", false)` we will always get `false` (because the associated value is not a boolean). Maybe we need to deprecate/remove `config.put` as well and add `config.putObject`?

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
easbar: ... and the reason I ran into this is that `RouteResource#translateTurnCostsParamToEdgeBased` is not using `Helper.toObject`... This of course can be fixed, but I still think the old approach is better (because it avoids the `ClassCastException`). wdyt?

easbar:
And maybe we should rename `put(PMap)` to `putAll(PMap)`?

karussell:
> We should probably add @Deprecated annotation to PMap#put(String, String) instead of only @deprecated in java docs. Also the deprecation is not inherited by HintsMap, so we should put @Deprecated there as well?

Ok

> There are still quite a few uses of the deprecated put method, shouldn't we get rid of it?

Yes, will do.

> Maybe we need to deprecate/remove config.put as well and add config.putObject?

Ah, yes. I forgot to remove this. As this was only recently introduced I think there is no need to deprecate it first.

> And maybe we should rename put(PMap) to putAll(PMap)?

Yes

I will also have to make the getObject method type safe.

karussell:
Interesting. Did not know this:

> The @Deprecated annotation is not annotated with @Inherited. Therefore the annotation will not automatically apply to the inherited methods.

core/src/main/java/com/graphhopper/routing/ProfileResolver.java
easbar: For example I could do a query with `edge_based=chocolate` and this will yield a `ClassCastException` (a 500 error)

karussell: Oh, this is very ugly. And this is a problem we now have for all methods. E.g. `short_fastest.distance_factor=blup` will cause the same ClassCastException.

Should we modify the getter methods and if not the expected class => return the default?

easbar: Are you sure? `hints.getFloat("short_fastest.distance_factor", 3.6)` should just work. The problem here is only that we use `getObject` and then cast to Boolean?

easbar: > Should we modify the getter methods and if not the expected class we return the default?

Isn't this exactly what we do?

karussell: Oh, sorry. Again a stale-state in my mind. I had a cast approach for PMap before I realized that instanceof is better...

PR #1959, opened 2020-03-18
Added more compass directions

samruston:
As with N, E, S, W the other directions are important for the road name


karussell:
Thanks!

PR #1964, opened 2020-03-19
Disable turn costs for LM preparation

### LINKED ISSUE #1960: LM preparation might run with turn costs even though its node-based
Here: https://github.com/graphhopper/graphhopper/pull/1949#discussion_r390257047 we decided to run the LM preparation with node-based graph traversal to make sure there are no turn costs. However, this is not sufficient to disable turn costs. Since the LM preparation runs node-based we cannot include turn costs, but therefore we also have to make sure that the weighting does not return any turn weights (or we have to make the `calcWeight` method in the algorithm(s)(at least the one for LM preparation) aware of the traversal mode (do not include turn costs for node-based)).

easbar:
Fixes #1960

### DIFF
--- a/reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
+++ b/reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
@@ -36,6 +36,7 @@ import com.graphhopper.util.Parameters.Landmark;
 import com.graphhopper.util.Parameters.Routing;
 import com.graphhopper.util.details.PathDetail;
 import com.graphhopper.util.exceptions.PointDistanceExceededException;
+import com.graphhopper.util.shapes.BBox;
 import com.graphhopper.util.shapes.GHPoint;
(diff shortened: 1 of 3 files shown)


reader-osm/src/test/java/com/graphhopper/GraphHopperIT.java
```diff
@@ -1408,6 +1409,83 @@ public void testDisablingLM() {
         assertTrue(res.getHints().getInt("visited_nodes.sum", 0) > 200);
     }
 
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    public void testCompareAlgos(boolean turnCosts) {
```
karussell: This is nice. I really need to invest some time into JUnit again :)

easbar: Same here this is about everything I know so far :)

PR #1974, opened 2020-03-29
Make sure node-based algos are not running with turn costs

### LINKED ISSUE #1965: add hasTurnCosts to Weighting
We need to fix #1960 in a different way to avoid issues with e.g. customized GraphHopper classes that override createWeighting.

easbar:
Fixes #1965

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -904,9 +904,9 @@ public class GraphHopper implements GraphHopperAPI {
             // Note that we have to make sure the weighting used for LM preparation does not include turn costs, because
             // the LM preparation is running node-based and the landmark weights will be wrong if there are non-zero
             // turn costs, see discussion in #1960
-            // Running the preparation without turn costs can be useful to allow e.g. changing the u_turn_costs per
+            // Running the preparation without turn costs can also be useful to allow e.g. changing the u_turn_costs per
             // request (we have to use the minimum weight settings (= no turn costs) for the preparation)
-            Weighting weighting = createWeighting(profile, new PMap().putObject("__disable_turn_costs_for_lm_preparation", true));
+            Weighting weighting = createWeighting(profile, new PMap(), true);
             lmPreparationHandler.addLMProfile(new LMProfile(profile.getName(), weighting));
         }
     }
@@ -982,12 +982,19 @@ public class GraphHopper implements GraphHopperAPI {
         }
     }
 
+    final public Weighting createWeighting(ProfileConfig profileConfig, PMap hints) {
+        return createWeighting(profileConfig, hints, false);
+    }
+
     /**
-     * @param profileConfig The profile for which the weighting shall be created
-     * @param hints         Additional hints that can be used to further specify the weighting that shall be created
+     * @param profileConfig    The profile for which the weighting shall be created
+     * @param hints            Additional hints that can be used to further specify the weighting that shall be created
+     * @param disableTurnCosts Can be used to explicitly create the weighting without turn costs. This is sometimes needed when the
+     *                         weighting shall be used by some algorithm that can only be run with node-based graph traversal, like
+     *                         LM preparation or Isochrones
      */
-    public Weighting createWeighting(ProfileConfig profileConfig, PMap hints) {
-        return new DefaultWeightingFactory(encodingManager, ghStorage).createWeighting(profileConfig, hints);
+    public Weighting createWeighting(ProfileConfig profileConfig, PMap hints, boolean disableTurnCosts) {
+        return new DefaultWeightingFactory(encodingManager, ghStorage).createWeighting(profileConfig, hints, disableTurnCosts);
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -982,12 +982,19 @@ void interpolateBridgesAndOrTunnels() {
...
-     * @param hints         Additional hints that can be used to further specify the weighting that shall be created
+     * @param profileConfig    The profile for which the weighting shall be created
+     * @param hints            Additional hints that can be used to further specify the weighting that shall be created
+     * @param disableTurnCosts Can be used to explicitly create the weighting without turn costs. This is sometimes needed when the
+     *                         weighting shall be used by some algorithm that can only be run with node-based graph traversal, like
+     *                         LM preparation or Isochrones
      */
-    public Weighting createWeighting(ProfileConfig profileConfig, PMap hints) {
-        return new DefaultWeightingFactory(encodingManager, ghStorage).createWeighting(profileConfig, hints);
+    public Weighting createWeighting(ProfileConfig profileConfig, PMap hints, boolean disableTurnCosts) {
```
easbar: Since disabling the turn costs is actually something we need to do sometimes I made this an actual parameter

devemux86:
Does that mean that flexible mode now cannot use turn restrictions?

easbar:
> Does that mean that flexible mode now cannot use turn restrictions?

No, it just means that when e.g. using Dijkstra the traversal mode has to be edge-based to use turn restrictions. But that has not changed, its been the same before.

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: Couldn't we use `turn_costs=false` in hints? I would like to avoid further parameters here.

easbar: But why not? We are creating weightings here and sometimes we need them without turn costs.. But if you really want to avoid the extra parameter, we can use a hint as before, I would just not like to call it turn_costs, because that had a different meaning before. So I would call it `__disable_turn_costs`, like it was called at the very beginning :)

karussell: > But why not?

I see createWeighting as a more generic method and it is unclear why disabling turn costs should be so special. 

Shouldn't the caller of createWeighting be able to create a Weighting and only use edge weights?

> I would just not like to call it turn_costs, because that had a different meaning before

Isn't the meaning very similar in the sense that we ask the profile to ignore turn costs? Like a none-CH profile with turn_costs=true configured but turn_costs=false in the request hints?

karussell: Ok, one argument for the additional boolean is a minor higher type safety for users overriding the createWeighting method. The would not need to explicitly figure out what is wrong in the source code.

easbar: > I see createWeighting as a more generic method and it is unclear why disabling turn costs should be so special.

Basically there are two ways to disable turn costs: The first one is via the profile, there are profiles with and without turn costs and depending on this the weighting will include turn costs or not. But then there are cases where we have to disable the turn costs **although** they are enabled in the profile, for example for Isochrones (because they are only implemented for node-based graph traversal so far) and for LM preparation (also not implemented edge-based and the preparation is even more useful when turn costs are excluded (can be used for queries with and without turn costs)). So we need a second flag to disable turn costs also for profiles that include turn costs. This second flag is what makes disabling the turn costs special. In principle it could also go into the hints parameter, but it is  slightly different than the other hints in that it is only set internally vs. something the user can change via the request.

> Shouldn't the caller of createWeighting be able to create a Weighting and only use edge weights?

Not sure if I understand the question, but yes thats totally possible with the disableTurnCosts flag.

> Isn't the meaning very similar in the sense that we ask the profile to ignore turn costs? Like a none-CH profile with turn_costs=true configured but turn_costs=false in the request hints?

Yes its similar to that, but `turn_costs` existed as a request parameter before and now its an internal thing, so it seems a bit confusing to use the same name.

karussell: > > Shouldn't the caller of createWeighting be able to create a Weighting and only use edge weights?

> Not sure if I understand the question, but yes thats totally possible with the disableTurnCosts flag.

I meant: createWeighting could be made independent of the use case. And then in the algorithm the node based traversal leads to skip calling calcTurnWeight without the need for this disableTurnCosts flag. I.e. turn cost supports needs then two things for the algorithm: a "proper created" Weighting and edge-based traversal.

karussell: > I.e. turn cost supports needs then two things for the algorithm: a "proper created" Weighting and edge-based traversal.

I do not say this is wrong :) ... just that it might be slightly cleaner but likely also more complicated to do it at the algorithm level

easbar: No its not complicated at the algorithm level and I suggested this a few times ([here](https://github.com/graphhopper/graphhopper/issues/1965#issuecomment-605502471) and [here](https://github.com/graphhopper/graphhopper/issues/1960#issuecomment-601143072)) :) . The reason why I eventually decided against this was that this way when we pass a weighting that includes turn costs into an algorithm with node-based graph traversal the turn costs will be silently ignored. The way we are doing this now is more strict in this sense: If we choose node-based we also have to make sure the weighting does not include turn costs, otherwise we will be notified that what we are trying to do is not possible.

> createWeighting could be made independent of the use case. 

The createWeighting method does not really know what it is used for, it just has a switch that allows creating a weighting without turn costs even for profiles that have turn costs enabled

karussell: > I eventually decided against this was that this way when we pass a weighting that includes turn costs into an algorithm with node-based graph traversal the turn costs will be silently ignored. The way we are doing this now is more strict in this sense:

Ah, thanks for the explanation.

Issue #1967, opened 2020-03-26
turn_costs handling in /isochrone


easbar:
Recently we fixed the handling of turn_costs for the /isochrone endpoint on the 1.0-pre32-fixes, but these changes also need to be ported to current master. 

* [x] proper handling of turn costs parameter for /isochrone and /spt 
* [x] The other thing we are missing is the turn_costs<->edge_based conversion we are already doing for /route (at least for GET requests, see #1966). The same holds for /spt as well! -> done here: 8a24e31eb772e5989c4e3eb1fc09068986a608a5

easbar:
We should probably do this parameter handling in `ProfileResolver`, then there should no longer be need to do this in every `Resource`.

easbar:
So far we only support node-based for isochrones. So we either have to make sure there is a profile without turn costs or we make sure the isochrone algorithm (which is node-based) ignores turn costs or we use the `__disable_turn_costs_for_lm_preparation` in `GraphHopper#createWeighting` (and probably rename back it to `__disable_turn_costs`). See also #1965.

easbar:
Assuming isochrone cannot be run edge-based (as is currently the case), whats better (or less ugly):

* it does not matter whether or not a profile has turn costs and the turn costs are ignored for /isochrone even when `turn_costs: true` is set in the profile

* /isochrone does not allow profiles with `turn_costs: true`, because its not implemented. So to make use of /isochrone we always have to add a profile with `turn_costs: false` and potentially we have to add a copy with `turn_costs: false` for every profile we want to use to make it also work for /isochrone.

PR #1975, opened 2020-03-29
Allow running isochrones with profiles supporting turn costs (without turn costs)

easbar:
This is based on #1974.
Fixes one of the issues of #1967. With this fix is no longer problematic to run isochrone requests when the profile in use supports turn costs and the turn costs are simply ignored (they cannot be dealt with because we are missing an edge-based isochrone implementation). The alternative would be forcing another profile with `turn_costs: false`, but this seems ugly and at some point we probably will have an edge-based isochrone algorithm.

### DIFF
--- a/web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
+++ b/web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
@@ -36,6 +36,9 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 import java.util.*;
 
+import static com.graphhopper.util.Parameters.Routing.EDGE_BASED;
+import static com.graphhopper.util.Parameters.Routing.TURN_COSTS;
+
 @Path("isochrone")
 public class IsochroneResource {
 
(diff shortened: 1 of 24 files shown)


web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
```diff
@@ -77,13 +77,19 @@ public Response doGet(
...
         HintsMap hintsMap = new HintsMap();
         RouteResource.initHints(hintsMap, uriInfo.getQueryParameters());
+        if (!hintsMap.getBool(Parameters.CH.DISABLE, true))
+            throw new IllegalArgumentException("Currently you cannot use speed mode for /isochrone, Do not use `ch.disable=true`");
+        if (!hintsMap.getBool(Parameters.Landmark.DISABLE, true))
+            throw new IllegalArgumentException("Currently you cannot use hybrid mode for /isochrone, Do not use `ch.disable=true`");
+        if (hintsMap.getBool(Parameters.Routing.EDGE_BASED, false))
+            throw new IllegalArgumentException("Currently you cannot use edge-based for /isochrone. Do not use `edge_based=true`");
+        if (hintsMap.getBool(Parameters.Routing.TURN_COSTS, false))
+            throw new IllegalArgumentException("Currently you cannot use turn costs for /isochrone, Do not use `turn_costs=true`");
```
karussell: There is a small mistake for Landmark.DISABLE -> ch -> lm

PR #1989, opened 2020-04-16
iOS port: remove Java Runtime specifics

oflebbe:
This PR is opened in order to create a branch the work on graphhopper-ios https://github.com/graphhopper/graphhopper-ios/pull/47  can depend on for the time being. Please do not close the PR right now.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -600,7 +600,7 @@ public class GraphHopper implements GraphHopperAPI {
 
         DAType elevationDAType = DAType.fromString(ghConfig.getString("graph.elevation.dataaccess", "MMAP"));
         ElevationProvider elevationProvider = ElevationProvider.NOOP;
-        if (eleProviderStr.equalsIgnoreCase("srtm")) {
+      /*  if (eleProviderStr.equalsIgnoreCase("srtm")) {
             elevationProvider = new SRTMProvider(cacheDirStr);
         } else if (eleProviderStr.equalsIgnoreCase("cgiar")) {
             elevationProvider = new CGIARProvider(cacheDirStr);
@@ -618,7 +618,7 @@ public class GraphHopper implements GraphHopperAPI {
         elevationProvider.setInterpolate(interpolate);
         if (!baseURL.isEmpty())
             elevationProvider.setBaseURL(baseURL);
-        elevationProvider.setDAType(elevationDAType);
+        elevationProvider.setDAType(elevationDAType); */
         return elevationProvider;
     }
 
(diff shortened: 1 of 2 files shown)


oflebbe:
My intention was that this Pull request will create a branch on graphhopper this way.... At least I didn't found it ...

karussell:
I'll create a new "ios_compatibility" branch (or something) and integrate your changes.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -620,7 +620,7 @@ private static ElevationProvider createElevationProvider(GraphHopperConfig ghCon
             elevationProvider.setBaseURL(baseURL);
         elevationProvider.setDAType(elevationDAType);
         return elevationProvider;
-    }
+    } */
```
karussell, line 623: Can you explain why we need to remove ElevationProvider here? It is just an interface and should not make a problem or does it? Couldn't you set ElevationProvider.NOOP?

The advantage would be that the CI pipeline would not fail.

oflebbe: I only had to make sure that HeightTile.java is not needed, since it contains references to java.awt.*, which are not supported. Let me try to rewrite it for this usecase.

oflebbe: Ok, but now the elevation specific tests fail. For a reason . How to proceed from here ?
Should we comment the tests as well ?
Since we do not need Elevations at all on mobile ... We will never generate the graph on mobile.

karussell: Hmmh, tricky. But we could make it working when we move the whole init method out of the GraphHopper class. Probalby this is a good idea in general, but to avoid a delay here we could for now exclude building the web module in this branch. Will try.

BTW: our usual workflow is to keep all the commits of a pull request (instead of overwriting old commits) and then squash merge them as one commit into master. So the master branch stays relatively "clean" and the pull request branch still has everything in it and we can see how it evolved.

karussell: Can you try the following change:

```patch
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_script:
   - cd web && npm install && BROWSERIFYSWAP_ENV='development' npm run bundleProduction && cd ..
 
 script:
-  - "mvn -Dkey=$API_KEY clean install -B"
+  - "mvn -Dkey=$API_KEY clean install -B -pl \!web"
```

With that the branch is red only if something is broken.

karussell:
Ok, once we merge this the test should be green.

If you like we can additionally add you as a contributor. We would do this on master branch and for that we would also need our CLA signed (just to confirm the changes are under Apache License): see [this document](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md#license-agreement). Let me know your email (you can contact us [here](https://www.graphhopper.com/contact-form/))

Issue #1985, opened 2020-04-13
deploy to docker hub or state there that version is outdated


karussell:
The current version is outdated: https://hub.docker.com/r/graphhopper/graphhopper

For every tag (release) we should push to docker hub too or add a description there that it is outdated or even remove it.

PR #1990, opened 2020-04-16
Improve docker image size

HarelM:
* I've changed the docker image creation to reduce its size.
* I've added some safe checks to config copying when using graphopper.sh and config.yml does not exists
* I've moved ensure maven to maven commands only
* I've added artifacts to appveyor and updated the build image to latest image.

The last item is not mandatory for docker.
This is related to #1985


Dockerfile
```diff
@@ -1,23 +1,36 @@
...
+FROM openjdk:11.0.5-jre
+
+ENV JAVA_OPTS "-server -Xconcurrentio -Xmx1g -Xms1g -XX:+UseG1GC -Ddw.server.application_connectors[0].bind_host=0.0.0.0 -Ddw.server.application_connectors[0].port=8989"
+
+RUN mkdir -p /data
+
+WORKDIR /graphhopper
+
+COPY --from=build /graphhopper/web/target/*.jar ./web/target/
+COPY ./graphhopper.sh ./pom.xml ./config-example.yml ./
```
karussell, line 32: Is the pom.xml file really necessary here? As we use a jre (second FROM) compiling is not possible anymore.

Dockerfile
```diff
@@ -1,23 +1,36 @@
...
+
+RUN npm install && npm run bundleProduction
+
+WORKDIR /graphhopper
 
 RUN ./graphhopper.sh build
 
+FROM openjdk:11.0.5-jre
+
+ENV JAVA_OPTS "-server -Xconcurrentio -Xmx1g -Xms1g -XX:+UseG1GC -Ddw.server.application_connectors[0].bind_host=0.0.0.0 -Ddw.server.application_connectors[0].port=8989"
```
karussell, line 25: Can you remove the options -server -Xconcurrentio and -XX:+UseG1GC? We should use the JVM defaults.

Dockerfile
HarelM: It is used to get the version in the bash script. Changing the bash script will allow removing this file from the docker file if the version will be fetched from other location or getting the jar file in another method.
My initial attempt where without it, but this is the simplest solution for the docker.
I think the version in the bash file to find the right jar file should use the jar file name instead, but it is fragile and I think out of this scope..

Dockerfile
HarelM: Sure, I have no clue what all these variables mean so I decided to leave them as they were...

HarelM: Done.

Dockerfile
karussell: Ah, thanks for finding this out. Can you add a comment and point to this issue on why these files need to be in the Docker image?

HarelM: Done.

Dockerfile
```diff
@@ -1,23 +1,37 @@
...
-RUN cd web && npm install && npm run bundleProduction && cd ..
+WORKDIR /graphhopper/web
+
+RUN npm install && npm run bundleProduction
+
+WORKDIR /graphhopper
 
 RUN ./graphhopper.sh build
 
+FROM openjdk:11.0.5-jre
```
otbutz, line 23: Do we need to hardcode minor/patch versions?

Dockerfile
```diff
@@ -1,23 +1,37 @@
...
+FROM maven:3.6.3-jdk-8 as build
 
 ENV JAVA_OPTS "-server -Xconcurrentio -Xmx1g -Xms1g -XX:+UseG1GC -Ddw.server.application_connectors[0].bind_host=0.0.0.0 -Ddw.server.application_connectors[0].port=8989"
 
-RUN mkdir -p /data && mkdir -p /graphhopper
-
 # install node - only required for JS UI
 RUN apt-get install -y wget \
        && curl -sL https://deb.nodesource.com/setup_13.x | bash - \
        && apt-get install -y nodejs
```
otbutz, line 8: Dropping the package cache might further reduce image size:

```sh
RUN apt-get install -y wget \
       && curl -sL https://deb.nodesource.com/setup_13.x | bash - \
       && apt-get install -y nodejs \
       && rm -rf /var/lib/apt/lists/*
```

Dockerfile
HarelM: Thanks for the tip. I have removed the .5, using 11.0 now.

otbutz: i think `openjdk:11-jre` should work too

Dockerfile
HarelM: This is not part of the deployed image but it is part of the build image, which is not reflected in the total size.
The current image size is about 130 Mb, there's a way to reduce it if we use jre-slim but then we'll need to add wget to the image, which I'm not sure is so simple.
Also I guess that not all jar files are needed but probably just one which will reduce the image size significantly. But I'm not sure this is entirely related to this issue as the main purpose of this PR is to allow continuous images build.
I can create another PR and open anther issue if there's a need to reduce image size more.

Dockerfile
HarelM: Minor version changes can still break some code, I prefer to be on the safe side. Feel free to change it though. :-)

Dockerfile
HarelM: Sorry, I now see the name I gave to this pull request... :-/ In any case, the current changes are good enough on their own I believe. further improvement can be made later on assuming this will be tested to work as expected not just by me :-)

Dockerfile
karussell: I prefer stable and specific versions too. So I would keep it as it is.

Dockerfile
karussell: > which I'm not sure is so simple.

See https://stackoverflow.com/a/28885237/194609

Dockerfile
karussell: Somehow there are 0 commits in this PR now. Can you have a look?

Dockerfile
HarelM: My concern is not how simple it is to add wget to the image but rather what else is missing...
Since I don't know I prefer to start of like that and maybe later on reduce the size more when the whole process is up and running.
As I said, there are further improvements that can be made later on.

Dockerfile
HarelM: I'm not sure what happened, some automatic merge of some sort. Should be OK now...

HarelM:
Yay! Thanks!

Issue #1991, opened 2020-04-16
/isochrone and /spt return 404 when point is missing


mc51:
**Describe the bug**
The http://localhost:8989/isochrone endpoint is missing in my local server. I get a 404 response. The /route endpoint works flawlessly.

**To Reproduce**
I followed this guide:
https://github.com/graphhopper/graphhopper/blob/0.13/docs/web/quickstart.md
I use the exact same config as here: https://github.com/graphhopper/graphhopper/blob/0.13/config-example.yml

**Expected behavior**
I'd expect the endpoint to be available for requests and not return 404. 

**System Information**

I used this jar version:
https://graphhopper.com/public/releases/graphhopper-web-0.13.0.jar
I'm on Windows 10 and Java is this:

```
java version "1.8.0_251"
Java(TM) SE Runtime Environment (build 1.8.0_251-b08)
Java HotSpot(TM) 64-Bit Server VM (build 25.251-b08, mixed mode)
```
**Screenshots & Logs**

There are no logs from the server.

karussell:
Indeed we return a different 400 message for missing parameters. This could be more consistent.

But you only need to specify at least the point parameter: http://localhost:8989/isochrone?point=43.733647%2C7.415686 and see that it will work.

easbar:
I just had the same problem with `/spt`. Returning 404 seems very wrong as it makes it look the whole endpoint is missing and we are not shown the intended error message (no point specified). There is also another problem for /spt: Since it returns `text/csv` the serialization of exceptions to json does not work and we are getting 500 errors instead of the error messages in json format.

easbar:
@karussell don't you think this should be fixed? If not feel free to close again.

easbar:
With 7df3192d137d025e3cd32a5c9eb99379207ee6b6 we now get 400 when the point is missing. Also I made sure that for /spt (which normally returns csv) we get a json error message. Not sure if we could achieve the same with framework features, so anybody who can do better please go ahead (@michaz?).

@mc51 for 0.13 you still get 404 if you do not specify the point, but obviously you need to specify a point anyway

karussell:
Thanks!

Issue #1986, opened 2020-04-14
Remove/replace tests against graphhopper.com


easbar:
`GraphHopperMatrixIT` and `GraphHopperWebIT` are running against the public GH service, but I think this not ideal. If they fail its often because of some data changes and nothing can be done here in the open source project except fixing the expected values. 

What should be tested (instead):

* test that `GraphHopper(Matrix)Web` is **sending** the right requests (no need to actually send the requests to graphhopper.com for this). Either just unit test the methods that create the request from `GH(Matrix)Request` or do as its done right now but send the requests to some test server on localhost.

* test the way the (hosted) GH server answers certain requests (but that should not happen here in the open source project).

easbar:
I just fixed the expected distance here again:

And it seems odd that only the distance (but not the weight) changed. Here is the route: 

https://graphhopper.com/maps/?point=51.467697%2C-0.090637&point=51.521241%2C-0.171833&locale=en-us&vehicle=car&weighting=fastest&turn_costs=true&use_miles=false&layer=Omniscale

![image](https://user-images.githubusercontent.com/17603532/79213889-3e87fa80-7e4a-11ea-8b50-c631e21390e5.png)

Note that the route looks a bit weird in a few places, for example just before the end:

![image](https://user-images.githubusercontent.com/17603532/79214065-78f19780-7e4a-11ea-8c86-6af00d2155fa.png)

easbar:
Seems to be a mapping error, take a look at the different layers:

![image](https://user-images.githubusercontent.com/17603532/79214199-b1917100-7e4a-11ea-93c1-02f94503165e.png)

![image](https://user-images.githubusercontent.com/17603532/79214226-beae6000-7e4a-11ea-924b-7c55e595c5aa.png)

![image](https://user-images.githubusercontent.com/17603532/79214266-d4238a00-7e4a-11ea-9dad-085eea38073c.png)

karussell:
I removed the matrix tests that goes to graphhopper.com: https://github.com/graphhopper/graphhopper/commit/a5ca5ff4f43012f19db32285b733d2f02838c37d

The other tests I would not want to remove but to replace them with local queries so that we know that the client is "really working".

easbar:
> but to replace them with local queries 

Isn't this the same as we currently do with `RouteResourceTest` etc. but instead of using `DropwizardRule#app.client()` we use `GraphHopperWeb` to send the request?

karussell:
[...]

PR #2000, opened 2020-04-18
Move gpx test from GraphHopperWebIT to RouteResourceTest

### LINKED ISSUE #1994: GraphHopperWeb#export does not use POST /route endpoint
Here: https://github.com/graphhopper/graphhopper/blob/a5b48ceadc6b3973659075233687dd8991748d87/client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java#L368

it should be:

```java
  Request okRequest = postRequest ? createPostRequest(ghRequest) : createGetRequest(ghRequest);
```

but if I do this `GraphHopperWebIT#testExportWithoutTrack` fails.

easbar:
Fixes #1994 

Two bugs here:

Bug 1: Despite setting `usePostRequest` `GraphHopperWeb#export` was using GET
Bug 2: When using POST the `gpx.millis` parameter was an empty String (`timeString`) and when trying to parse it there was a `NumberFormatException` (leading to a misleading error message about an empty string)

See comments inline


client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
```diff
@@ -64,8 +64,8 @@
     private boolean calcPoints = true;
     private boolean elevation = false;
     private String optimize = "false";
-    boolean postRequest = true;
-    int maxUnzippedLength = 1000;
+    private boolean postRequest = true;
```
easbar, line 67: This wasn't private before but should be now problem to do so as there is a setter anyway

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
```diff
@@ -64,8 +64,8 @@
...
+    private int maxUnzippedLength = 1000;
```
easbar, line 68: Had to add a setter here, since we are now no longer testing this in the same package, do you think thats ok?

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
```diff
@@ -138,6 +138,14 @@ public GraphHopperWeb setKey(String key) {
...
     }
 
+    /**
+     * Only use this if you know what you are doing
+     */
+    public GraphHopperWeb _setMaxUnzippedLength(int maxUnzippedLength) {
+        this.maxUnzippedLength = maxUnzippedLength;
+        return this;
+    }
+
```
easbar, line 148: WDYT?

client-hc/src/test/java/com/graphhopper/api/GraphHopperWebIT.java
```diff
@@ -245,39 +246,6 @@ public void doNotReadFinishInstruction() {
         assertEquals("", finishInstructionName);
     }
 
-    @Test
```
easbar, line 248: I'm happy about every test we can remove here, see #1986.

web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
```diff
@@ -273,7 +273,14 @@ private void removeLegacyParameters(GHRequest request) {
             throw new IllegalArgumentException("Alternatives are currently not yet supported for GPX");
         }
 
-        long time = timeString != null ? Long.parseLong(timeString) : System.currentTimeMillis();
```
easbar, line 276: This was the bug: For GET a missing `gpx.millis` would give null, but for POST it would be an empty String.

web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
```diff
@@ -273,7 +273,14 @@ private void removeLegacyParameters(GHRequest request) {
...
+        long time = System.currentTimeMillis();
+        if (!Helper.isEmpty(timeString)) {
+            try {
+                time = Long.parseLong(timeString);
+            } catch (NumberFormatException e) {
+                throw new IllegalArgumentException("Invalid value for 'gpx.millis': " + timeString);
```
easbar, line 281: Not handling `NumberFormatException` gave a misleading error message.

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
```diff
@@ -368,7 +376,7 @@ private Request createGetRequest(GHRequest ghRequest) {
     public String export(GHRequest ghRequest) {
         String str = "Creating request failed";
         try {
-            Request okRequest = createGetRequest(ghRequest);
+            Request okRequest = postRequest ? createPostRequest(ghRequest) : createGetRequest(ghRequest);
```
easbar, line 379: This was 'bug' Bug1

web/src/test/java/com/graphhopper/http/resources/RouteResourceTest.java
```diff
@@ -502,6 +504,54 @@ public void testGPXWithError() {
...
+    public void testGPXExport(boolean usePost, int maxUnzippedLength) {
+        GHRequest req = new GHRequest(42.554851, 1.536198, 42.510071, 1.548128);
+        req.putHint("elevation", false);
+        req.putHint("instructions", true);
+        req.putHint("calc_points", true);
+        req.putHint("gpx.millis", "300000000");
+        req.putHint("type", "gpx");
+        GraphHopperWeb gh = new GraphHopperWeb(clientUrl(app, "/route"))
+                .setPostRequest(usePost)
+                ._setMaxUnzippedLength(maxUnzippedLength);
```
easbar, line 522: We could also remove maxunzipped length here since we are not testing it anyway, but we might need it for some other tests in `GraphHopperWebIT` (but then we can still add it -> removed it here :))

michaz:
> In `RouteResource#doGet` we have `@QueryParam("gpx.millis") String timeString)` and I was wondering first of all why is there no default value? We use `System.currMillis()` as a default later anyway. Second, why is the type String when the value is expected to be a number and we even parse it as such later? Then I tried changing to long, but in case I sent a GET request with a non-number (like `xyz`) this gives a 404 error. I assume Dropwizard interprets a query parameter that has the wrong type as an 'unknown url'. We had the same problem here: #1991, but only now I understood why we got 404. This is kind of an annoying behavior. I would prefer a 400 error stating that the parameter had the wrong type, not sure if (something in) Dropwizard has an option to change this. @michaz maybe?

When I Google 'jersey wrong query parameter type 404' I find that
a) this is by design, and
b) you're not the only one who doesn't like it

<shrug>

karussell:
@michaz correctly wrote that this GPX stuff is rather costly. Should we avoid including it for POST? Before releasing 1.0 would be the right time. See #2006

easbar:
Yeah you seem to agree that it should be removed, I do not even really know what it does :) Like I said the only thing I do not fully understand is why (if we remove it for POST) we keep it for GET, because to me GET seems to offer no advantages over POST except that it is handy to copy&paste urls of routing requests like in GH maps (for this reason I think we should keep GET, but I do not see why we need gpx for GET if it is something we can get rid of).

karussell:
People and customers are still using this feature and removing it from GET would mean a breaking change. At least we should give them time to migrate and do this conversion on the client.

Not introducing it in POST is inconsistent, yes, but on the other hand we do not have to add it for other endpoints that support POST?

easbar:
Since we removed (the anyway broken) GPX support for POST there is no longer need to fix it :) I reverted the changes here. I would still move the gpx tests to RouteResource instead of running agains graphhopper.com as I did here.

client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
```diff
@@ -376,7 +368,8 @@ private Request createGetRequest(GHRequest ghRequest) {
     public String export(GHRequest ghRequest) {
         String str = "Creating request failed";
         try {
-            Request okRequest = postRequest ? createPostRequest(ghRequest) : createGetRequest(ghRequest);
+            // todonow: should there be an error if usePost=true (because it is not supported)
```
karussell, line 371: Yes, this would be probably a good idea.

Issue #1996, opened 2020-04-17
GET /route endpoint does some additional checks compared to POST


easbar:
This stuff:

https://github.com/graphhopper/graphhopper/blob/7bee0e42ac6b13fe707699964120ed081c6f3ff7/web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java#L106-L132

should probably also be done for POST requests, so we should rather move it into `GraphHopper#calcPaths`.

c.f. https://github.com/graphhopper/graphhopper/pull/1934#discussion_r410027245

Issue #1995, opened 2020-04-17
Internal Server Error when sending POST request to /route without specifying points


easbar:
```shell
curl -X POST -H "Content-type: application/json" -d "{}" http://localhost:8989/route

Error handling a request: 17a5bee0823f74d0
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
```

michaz:
Ah, this has nothing to do with custom exception handling for once, but is a simple error.

In GH core, there's a check that checks that enough points are there, and it returns a GHResponse with an error (as is the convention there..) And it works when there are no CHs. (Like in the tests.)

But when there are CHs, we get an ArrayIndexOutOfBoundsException from somewhere where CH preconditions are checked.

michaz:
I think we agree now that all validation should be in the web service, and if the library must do it again, then it can, but.. the web service should only pass valid requests to the library.

Maybe we should start @Valid ating the GHRequest.

michaz:
... with the slight complication that validation annotations don't work with Mixins of course, because Mixins are Jackson and this is Hibernate >:-( but this would only be marginally less ugly with a subclass, ValidatableGHRequest or something, that's local to the resource.

michaz:
Or maybe (for RouteResource) just remove all the exceptions in the resource, and make sure all the exceptions in the core are thrown from the right place and as IllegalArgumentException.

No idea.

easbar:
I would check something like "point list may not be empty" in `GraphHopper#route`, because they need to be checked also when using `GraphHopper` without the resource. For GET we already do this in the resource and I would just move it into the GraphHopper class, see #1996.

easbar:
> In GH core, there's a check that checks that enough points are there, and it returns a GHResponse with an error (as is the convention there..) And it works when there are no CHs. (Like in the tests.)

Ugh its even more complicated than that: CH has to to be disabled and then there is this `checkNonChMaxWaypointDistance` check which causes the error, but this check only really runs if the max non ch distance is not set to default (`Integer.MAX`)...

easbar:
I fixed this in #2010

PR #2003, opened 2020-04-18
test client-hc routing without internet requirement

karussell:
For #1986

@easbar Do you know how we can make the parametrized testing working with the dropwizard extension? Maybe some junit5 trick :) ?

One important difference to note is that for graphhopper.com we have to stay backward compatible and so it is not required for the client to always include the vehicle (or profile).


easbar:
Cool! Thanks :)

> Do you know how we can make the parametrized testing working with the dropwizard extension? Maybe some junit5 trick :)

You mean parameterized at the class level? No, unfortunately not. It looks like this is still an open issue: https://github.com/junit-team/junit5/issues/878. Doing it at the test level is easy though, like I did here for the 'export' test: #2000. And if you put the parameter array in a class-wide constant it will only take one more line per test:

```java
private static final String[] PARAMS = {"true,0","true,1000","false,-1"};
// ...
@ParameterizedTest // instead of @Test
@CsvSource(PARAMS)
public void myTest(boolean usePost, int maxUnzippedLength) { // have to add the parameters here
```

I would argue that this might be a bit more code, but at the same time it makes the tests more decoupled so we can adjust the paramaters easily for each test if needed. For example the gpx export test does not make use of max unzipped length (that's why I think we can move it to `GraphHopperIT` like I did in #2000).
Btw in some cases I would have liked to do the server configuration on the test level as well, I am still looking for a way to do this without creating a new test class.

karussell:
> You mean we cannot rely on the vehicle/profile always being set and instead should keep some tests where this is not the case?

I meant for 2 profiles (like here) the routing engine fails if no vehicle is specified which is not the case for graphhopper.com where we always assume car as default.

> You mean parameterized at the class level? No, unfortunately not. It looks like this is still an open issue: junit-team/junit5#878.

Thanks, will try that!

karussell:
Unfortunately we have to add more boilerplate as CsvSource requires a ["String constant"](https://stackoverflow.com/a/28031809/194609) and so PARAMS (String array that could be changed) is not working. Also the GraphHopperWeb instance has to be created in every test:

```java
@ParameterizedTest(name = DISPLAY_NAME)
@CsvSource({"true,0", "true,1000", "false,-1"})
public void myTest(boolean usePost, int maxUnzippedLength) {
  GraphHopperWeb gh = new GraphHopperWeb(TestUtils.clientUrl(app, "/route")).
   setPostRequest(usePost).setMaxUnzippedLength(maxUnzippedLength);
```

easbar:
> Unfortunately we have to add more boilerplate as CsvSource requires a "String constant" and so PARAMS (String array that could be changed) is not working.

Oh you are right. We could still do `@CsvSource(value ={VAL1, VAL2, VAL3})`. Or we could use `@EnumSource`, which is actually even better:

```java
    private enum TestParam {
        GET(true, 12),
        POST_MAX_UNZIPPED_0(false, 5),
        POST_MAX_UNZIPPED_1000(false, 5);
        
        public boolean usePost;
        public int maxUnzippedLength;
        TestParam(boolean usePost, int maxUnzippedLength) {
            this.usePost = usePost;
            this.maxUnzippedLength = maxUnzippedLength;
        }
    }
    @ParameterizedTest
    @EnumSource(value = TestParam.class)
```

easbar:
>  Also the GraphHopperWeb instance has to be created in every test:

We can probably make this a class field and put the creation into a `@BeforeEach` (or `@BeforeAll` method?

> Should profile=xy already work for POST requests?

Yes

> And should we include the profile name in the exception?

Yes I only left this out because I was concerned about a multi-server setup, but now we should include this.

web/src/test/java/com/graphhopper/http/resources/RouteResourceClientHCTest.java
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    @ParameterizedTest(name = DISPLAY_NAME)
+    @CsvSource({"true,0", "true,1000", "false,-1"})
+    public void testPathDetails(boolean usePost, int maxUnzippedLength) {
+        GraphHopperWeb gh = createGH(usePost, maxUnzippedLength);
+        GHRequest req = new GHRequest().
+                addPoint(new GHPoint(42.507065, 1.529846)).
+                addPoint(new GHPoint(42.510383, 1.533392)).
+                putHint("profile", "my_car");
```
easbar, line 342: You need to use `setProfile()` here. For POST `putHint` also seems to work (because of json serialization/deserialization probably). Ugh this is a bit ugly.

easbar:
> And should we include the profile name in the exception?

The error message above is from profile resolver, which lists the requested and available profile **properties** (since no profile was specified). I think in all these error messages we should recommend the user to use the profile parameter directly and list the available profiles.

web/src/test/java/com/graphhopper/http/resources/RouteResourceClientHCTest.java
karussell: Ah, yeah.

karussell: We could fix this when we internally use the hints Map for "profile". (btw: we would have the same problem with "algorithm" and probably all the other properties)

easbar: Why is there a problem with "algorithm" and the others? The problem here seems that we are calling `putHint` instead of `setProfile`? Or do you think its better to be able to use both `putHint` and `setProfile`? Or do you want to remove `setProfile` altogether? I have no strong opinion so far, but since `profile` is even more or less required it seems to deserve its own setter? And if it has a setter it should also have its own field (because Jackson expects this by convention).

karussell:
btw: it seems this has improved the build times from ~4.5 to ~4 minutes. But unfortunately the memory issues can now be seen also on master: https://travis-ci.org/github/graphhopper/graphhopper/jobs/677183186

easbar:
> But unfortunately the memory issues can now be seen also on master

I also ran into this but not consistently. But increasing the limit by some 50mb should also be no problem?

karussell:
> But increasing the limit by some 50mb should also be no problem?

I do not like it much as in the past it uncovered some memory related problems (usually in GraphHopper itself ;)). Still as even increasing to 110MB helped I'll do it and report this problem to dropwizard.

easbar:
Ok I already set to 150mb on master, but of course feel free to set it 110mb.

web/src/test/java/com/graphhopper/http/resources/RouteResourceClientHCTest.java
karussell: > The problem here seems that we are calling putHint instead of setProfile?

But for GET this approach works and so people might do it? (not sure)

> Or do you want to remove setProfile altogether?

No, just put the profile string into hints under the hood like
```java
... setProfile(String profile) {
hints.put("profile", profile);
return this;
}
```

Not sure if this is nice ...

easbar: > But for GET this approach works and so people might do it? (not sure)

I think it only works for POST, but you are right it should better work the same way for both. If this is our only concern we could also adjust `GraphHopperWeb#putHint` to set the profile if the given key is `profile`? Btw what do you think of removing GET from client-hc altogether? I do not see how it can be useful (except for writing tests that cover both GET and POST maybe but that is not really the purpose of client-hc :)). 

> Not sure if this is nice ...

Yes me neither. I think when there is `setProfile` Jackson assumes there is a profile field, but not sure if we need to care about this here...

karussell: > Btw what do you think of removing GET from client-hc altogether?

Hmmh, sounds good. Initially I tought we need a certain transition phase to make the POST request more stable, but now everything seems to work already.

karussell: Let's do this in another issue :) ?

easbar: > but now everything seems to work already

We should also fix #1996 first then (for example we still have #1995). The only thing I am wondering is if we give POST the higher priority why would you remove gpx export (#2006)? Is this feature somehow not useful anymore or replaced by something else or only supported for GH maps?

easbar: While the purpose of client-hc is certainly not testing GET&POST /route we still need some way to do this efficiently, it is quite annoying to test both endpoints separately and having bugs in one but not the other endpoint and so on. Or we manage to do very little in the Resource such that there is not much room for errors...

easbar:
I changed to `@EnumSource` it seems nice to me. Still calling `createGH` for each test, so we can pass it the test parameters (but I think thats fine).

Open points (to be fixed somewhere else)

* [ ] memory limit for tests increased
* [ ] list available profiles when profile not found 
* [ ] (maybe) remove GET /route from client-hc entirely

I think we can merge here.

karussell:
Thanks! Will merge.

> (maybe) remove GET /route from client-hc entirely

The problem is then the GPX export again as we removed support for POST :/

web/src/test/java/com/graphhopper/http/resources/RouteResourceClientHCTest.java
```diff
@@ -0,0 +1,383 @@
...
+                putObject("graph.encoded_values", "road_class,surface,road_environment,max_speed").
+                putObject("graph.location", DIR)
+                .setProfiles(Arrays.asList(new ProfileConfig("my_car").setVehicle("car").setWeighting("fastest"),
+                        new ProfileConfig("my_bike").setVehicle("bike").setWeighting("fastest")))
+                .setCHProfiles(Arrays.asList(new CHProfileConfig("my_car"), new CHProfileConfig("my_bike")));
+        return config;
+    }
+
+    // dropwizard extension does not work with @RunWith(Parameterized.class), but we can use an @EnumSource or similar
+    // and on each test method. see https://github.com/graphhopper/graphhopper/pull/2003
```
karussell, line 88: Nice!

PR #2010, opened 2020-04-21
Consolidate validation for GET&POST /route, clean up GHRequest#heading

### LINKED ISSUE #1995: Internal Server Error when sending POST request to /route without specifying points

### LINKED ISSUE #1996: GET /route endpoint does some additional checks compared to POST

easbar:
Fixes #1995 and Fixes #1996.

I moved the validation checks we previously only did for GET into `GraphHopper` and created a new `validateRequest` method there, which did/does all kind of checks, some of which also check relations between multiple parameters. One blocker was the treatment of `headings` in `GHRequest`. This class was trying very hard to maintain that the number of points equals the number of headings, but IMO this is the wrong place to do this. `GHRequest` is just a data object that represents a routing request. Whether or not this request is 'valid' or not should be up to the entity that handles the request (`GraphHopper#route` in this case). The constraint is not even as simple as "number of headings must equal number of points", but in some cases its also allowed to only specify one heading etc. So I removed all this from `GHRequest` let me know if you think otherwise.


web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
```diff
@@ -75,7 +73,7 @@ public Response doGet(
             @Context HttpServletRequest httpReq,
             @Context UriInfo uriInfo,
             @QueryParam(WAY_POINT_MAX_DISTANCE) @DefaultValue("1") double minPathPrecision,
-            @QueryParam("point") List<GHPointParam> pointParams,
+            @QueryParam("point") @NotNull List<GHPointParam> pointParams,
```
easbar, line 76: @michaz for a moment I was tempted to set this to `@NotEmpty` with a custom error message, but then I thought its more consistent to do this later so its the same for GET&POST. But `@NotNull` cannot hurt, because by default I like to assume that nothing is `null` (for example when I later use the `GHRequest`).

michaz:
I like it, it's not too different from DropWizards own @ValidationMethod (or something) extension of Hibernate Validator.

Issue #1979, opened 2020-03-31
Isochrone issue with distance_limit compared to actual route when not using ShortestWeighting


boldtrn:
I just saw an issue with the Isochrone calculation. When I pass a time_limit, the resulting isoline surrounds more or less exactly the area that I can reach within the given time. When I pass a distance_limit, the resulting isoline seems to differ significantly from the give distance.

Here is one example with 10km, but this seems to be the case for every example I have checked so far:
`/isochrone?point=48.692958,9.461932&distance_limit=10000&vehicle=car`

The resulting polygon looks like this:
![Screenshot from 2020-03-31 17-14-22](https://user-images.githubusercontent.com/1553525/78043144-425c4d00-7373-11ea-9d69-d866208de0c3.png)

If I pick any point at random at the border of the polygon, for example [this one](https://graphhopper.com/maps/?point=48.692958%2C9.461932&point=48.670219%2C9.529438&locale=de-DE&vehicle=car&weighting=shortest&turn_costs=true&use_miles=false&layer=Omniscale). The resulting route is < 7km. I would have expected the resulting route to be close to the requested 10km.

boldtrn:
Ah, I found the issue, sorry for the trouble. The issue was in this case that I used the incorrect weighting. If you use the FastestWeight the above result is generated. If you use the ShortestWeight, the results seem correct.

In this case the polygon looks like this:
![Screenshot from 2020-03-31 18-01-22](https://user-images.githubusercontent.com/1553525/78048229-b568c200-7379-11ea-888a-5146f62d2a23.png)

boldtrn:
Ok, so I think I found the underlying issue here as well. In `Isochrone#searchInternal()` we use the check `nEdge.weight > tmpWeight`, shouldn't we use the limit criteria instead?

So for distance, we should probably check: `nEdge.distance > tmpDistance`, I tried this locally, and the results for weighting=fastest, look a lot more reasonable (still a bit different than using weighting=shortest).

When I also change the compareTo method of the IsoLabel ot use distance, the results look pretty identical.

boldtrn:
So WDYT, should we change this? I could provide a PR to add the above mentioned changes in a more flexible way based on the finish criteria.

karussell:
[...]

Issue #1607, opened 2019-05-03
distance_limit without time_limit isn't that intuitive for the Isochrone API


karussell:
Currently we allow a distance_limit for the Isochrone API. This was good for use cases like "calculate a delivery area but roughly base it on distance not on the time". But explaining this property, i.e. what this it means in the time-based Isochrone API is difficult.

If you see distance_limit as an additional constraint as suggested from @oblonski (additionally to the time_limit) it makes more sense. With time_limit=very_big_number converts into the current situation.

michaz:
Careful here: Those aren't _constraints_ when talking about isolines. They're only something like constraints when you calculate the shortest path tree, and at some point you stop calculating. But with isolines, you calculate _the isoline for 6 minutes_, or _the isoline for 5 km_. (That's why `_limit` isn't a very accurate name.)

I think it makes most sense the way it is (apart from the name): You specify _either_ a distance (or several) _or_ a time (or several).

michaz:
See also #1979. The a) objective function, b) the search limit, _and_ c) the function we want the isoline for, are (potentially) _three_ different things.

a) directs the search. We get _different values_ for b) and c) depending on how a) is set.
b) needs to be high enough so a value for c) is available at every point
c) determines where the polygon(s) go through the point set

karussell:
Thanks, for fixing this and with the clear explanation the "strange" use case (weighting=fastest plus distance_limit) makes sense to me now too.

PR #2011, opened 2020-04-22
Isochrones just by weight (rather than time or distance)

boldtrn:
This PR adds the ability to limit isochrones by weight as well.

Since weight can be any number, I also made the "offset" configurable as well. Probably offset is not the correct word. Maybe search limit? 

Related to #1979 and #1607

### DIFF
--- a/isochrone/src/main/java/com/graphhopper/isochrone/algorithm/ShortestPathTree.java
+++ b/isochrone/src/main/java/com/graphhopper/isochrone/algorithm/ShortestPathTree.java
@@ -31,8 +31,7 @@ import com.graphhopper.util.GHUtility;
 import java.util.PriorityQueue;
 import java.util.function.Consumer;
 
-import static com.graphhopper.isochrone.algorithm.ShortestPathTree.ExploreType.DISTANCE;
-import static com.graphhopper.isochrone.algorithm.ShortestPathTree.ExploreType.TIME;
+import static com.graphhopper.isochrone.algorithm.ShortestPathTree.ExploreType.*;
 import static java.util.Comparator.comparingDouble;
 import static java.util.Comparator.comparingLong;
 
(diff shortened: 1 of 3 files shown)


michaz:
Thanks, that makes a lot of sense.

Do you think you could (for now) find a made-up number that works for you and hard-code it there, instead of introducing a new parameter?

These numbers are fudge-factors anyway, _they don't mean anything_, they are just there because they (at some point) fixed something for someone. It'll be replaced by something more clever anyway, soon I hope. ,-)

michaz:
Btw. I would be super-grateful for test-cases which *work* when those numbers are there, and *break* when they don't, i.e. where the isoline is really wrong in a sense, without those numbers. Do you have one with you, by chance? :-)

michaz:
Also, what *is* your use case, if I may ask?

Iso-motorcycle-attractiveness-lines?

boldtrn:
> Do you think you could (for now) find a made-up number that works for your use case (0 perhaps?) and hard-code it there, instead of introducing the new offset parameter?

Yes sure, I could also live with a hardcoded factor of 1.14 for now and completely remove the offset at this point.

To be honest, the main reason I added this was to be compatible with the other two options and to get the same results in the test. So if you remove the `weight_limit_offset` in `IsochroneResourceTest`, the result is not equal anymore.

My goal was to show that the results with the weight limit are equal to the other limits, if you make the input equal, if that's the case that's proof it's working.

> Do you have one with you, by chance? :-)

If you look at the current `IsochroneResourceTest#requestByWeightLimit`, it uses a limit of 3000m, internally we add an offset of 2000m for distance_limit, that's why I added the same offset of the weight_limit as well.

If you compare the results for the distance_limit with and without offset, using this request: `http://localhost:8989/isochrone?point=42.531073,1.573792&profile=car_shortest&type=geojson&distance_limit=3000` (same as the test). The results look like this, red is with offset, green without offset.

Screenshot from 2020-04-23 09-12-52

Now if you compare the actual distance, you see that the one with the offset is closer to the asked 3km.

Screenshot from 2020-04-23 09-13-01
Screenshot from 2020-04-23 09-17-36
Screenshot from 2020-04-23 09-08-16

boldtrn:
Thanks for merging this :tada:

PR #2014, opened 2020-04-23
timeout should propagate to OkHttp

### LINKED ISSUE #2013: testTimeout (RouteResourceClientHCTest) occasionally fails on windows/appveyor
The test should just make sure we consider the timeout setting in the client and we expect the request to fail as 1ms for a connection is usually too low.

But windows either rejects so small timeouts or it is very fast with connecting :)

karussell:
fixes #2013 

Had to move the timeout test into the same package to avoid making the getClientForRequest method public. Now RouteResourceClientHCTest.java has only integration tests going towards localhost.

### DIFF
--- a/client-hc/src/test/java/com/graphhopper/api/GraphHopperWebTest.java
+++ b/client-hc/src/test/java/com/graphhopper/api/GraphHopperWebTest.java
@@ -8,8 +8,6 @@ import com.graphhopper.PathWrapper;
 import com.graphhopper.jackson.Jackson;
 import com.graphhopper.jackson.PathWrapperDeserializer;
 import com.graphhopper.util.shapes.GHPoint;
-import org.junit.Assume;
-import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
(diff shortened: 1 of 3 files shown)


client-hc/src/test/java/com/graphhopper/api/GraphHopperWebTest.java
```diff
@@ -0,0 +1,75 @@
...
+     * Runs the same test with each of the supported traversal modes
+     */
+    @Parameterized.Parameters(name = "POST: {0}, maxUnzippedLength: {1}")
+    public static Collection<Object[]> configs() {
+        return Arrays.asList(new Object[][]{
+                {false, -1},
+                {true, 0},
+                {true, 1000}
+        });
+    }
```
easbar, line 42: This does not run agains graphhopper.com? do we need the parameterization here?

karussell: Ok, we would need it for GET and POST but the maxUnzippedLength is indeed misleading and not required.

karussell: (to test that it is applied to OkHttp for POST and GET)

easbar: One way to do it with Junit5 would be

```java
@ParameterizedTest(name = "POST={0}")
@BooleanSource(values = {true, false})
public void myTest(boolean usePost) {
}
```

karussell: Thanks, will use this as this is perfect here as I have to create it only once :)

karussell: Btw: there is no BooleanSource ... it is done via `@ValueSource(booleans = {true, false})` it seems.

easbar: oh sorry my bad

Issue #1978, opened 2020-03-31
Isochrones with turn costs


easbar:
The /isochrone and /spt endpoints currently do not support turn costs. We need to run the algorithm edge-based to support this.

To make things a bit less complicated these endpoints currently *can* be used also for profiles with `turn_costs: false` and the turn costs are simply disabled silently, c.f. this discussion: #1967

michaz:
I already have this on a branch. Coming up.

michaz:
- [ ] /isochrone

- [ ] /spt

michaz:
Finished, up to the fact that for /spt, edge based traversal really is something different than just respecting turn restrictions. You get a different kind of tree, where you get every node several times.
How to make it clear that this happens iff your profile has turn restrictions?
Maybe I can also want this if it has _no_ turn restrictions?

easbar:
Cool thanks!

> You get a different kind of tree, where you get every node several times.

Hm this seems to depend on the requested columns? For example if all I want are the coordinates, times and distances (the default) we should probably use the minima for every node? But if also the (previous) edge_ids etc. are requested we could just return all the entries of the shortest path tree? Or maybe that is what we do already?

> How to make it clear that this happens iff your profile has turn restrictions?
Maybe I can also want this if it has no turn restrictions?

This is indeed ugly and here (and maybe only here?) it would make sense to differentiate between `edge_based` and `turn_costs` (#1702)? In one case I might be interested in the edge-based tree and in another I might be interested in the node-based tree. But in which case would I be interested in the edge-based tree without turn costs? The calculation is probably not faster when we skip the turn costs, but the routing becomes less accurate (if we neglect OSM turn restrictions this way)?

A few more minor things:

* Don't you think there should be an error when `ch/lm.disable=false` since the speed modes are not supported for /isochrone and /spt?

* We now get an internal server error for something like this: `/spt?point=xyz`. Might be the same as in #1995. Taking into account that we also get 404 when parameters have the wrong type I wonder if we should put all query parameters as strings and do the parsing manually (maybe not so elegant but at least then we can control the error codes). Or maybe this is just a minor bug that can be fixed still using the framework mechanisms.

* Should it be allowed to use dashes in profile names, like `car-with-turncosts`? Right now its only allowed for some legacy reasons where something like `-1` (minus one) had to be included, but this will no longer be the case after fixing #1999. See here: https://github.com/graphhopper/graphhopper/pull/1922#discussion_r383033522

[...]

PR #2018, opened 2020-04-26
remove legacy turn_costs for isochrone too

karussell:
@easbar is there a reason turn_costs=false/true is not removed for isochrone too?


easbar:
Not that I know, but its also not used anyway? Ah but the problem is there is an error if its used? I think the argument was that it never has been documented so its not expected to be used? Better ask @michaz and maybe have a look into this dicsussion: https://github.com/graphhopper/graphhopper/issues/1978.

easbar:
But I think this PR is fine, in case its not supposed to be used its also ok to delete it from the hints after profile selection as you do here.

easbar:
Maybe add a test that uses turn_costs and/or edge_based for profile selection, similar to the one that was removed here: 01e56c0e29704cc7bab0da73a88bf5f5b56825f6

karussell:
Thanks! And have added a test.

PR #2021, opened 2020-04-28
Change custom weighting formula

easbar:
Here I changed the formula for custom weighting from 

```
w = d/(v*p) + d*f
```

to

```
w = d/v + d*s*(1-p)/p + d*f
```

where `w=weight,d=distance,v=speed,p=priority,f=distance_influence,s=priority_sensitivity`.
`p=1` means that there are no priority costs and `p=0` means that the weight is infinite (just like with the previous formula). For `0<p<1` we get additional priority costs that are proportional to the distance and `s` is just a factor that determines how 'sensitive' these costs are to changes of `p`. For example if `s` is large, something like `p=0.9` already means large priority costs and if `s` is small  `p=0.9` does not change the weight much. `s` should therefore be roughly equal to some common speed so the priority costs are of the same order as the speed costs. 

The key difference here is that `s` is a constant and does not depend on the speed of the edge. In fact if we set `s=1/v` (instead of it being a constant) the two formulas are identical. And this brings me to the key question here: Should the priority costs depend on the speed of an edge or not? Let's look at an example of an edge that is `1000m` long and the speed is `50km/h`. The speed costs on this edge are: `72s`. If we set `p=0.5` this value doubles with the old formula (`144s`). With the new formula this depends on our choice of `s`, but let's say we set `s=1/(50/3.6)`, then we also get `144s`.
But now let's assume the speed is not `50km/h`, but `25km/h`. In this case the old formula yields `288s` while the new formula yields: `144s+72s=218s`, because setting `p=0.5` still means the priority costs are `72s` not `144s`. So for the old formula changing `p` means that we 'add' priority costs that are proportional to the speed costs while with the new formula choosing a value of `p` means we get certain priority costs regardless of the speed.

WDYT? Honestly right now I am not sure what is better here. Generally I think the stress costs should be independent from the speed costs, but here we are using the `priority` factor to 'modify' an existing weighting and its not entirely clear what the meaning of `priority` is supposed to be.


core/src/test/java/com/graphhopper/routing/weighting/CustomWeightingTest.java
```diff
@@ -41,97 +40,142 @@
...
+    public void speedOnly() {
+        // 50km/h -> 72s per km, 100km/h -> 36s per km
+        EdgeIteratorState edge = graphHopperStorage.edge(0, 1, 1000, true)
+                .set(avSpeedEnc, 50).setReverse(avSpeedEnc, 100);
+        assertEquals(72, createWeighting(new CustomModel().setDistanceInfluence(0)).calcEdgeWeight(edge, false), 1.e-6);
+        assertEquals(36, createWeighting(new CustomModel().setDistanceInfluence(0)).calcEdgeWeight(edge, true), 1.e-6);
+    }
+
+    @Test
+    public void withPriority() {
```
easbar, line 73: This test shows how the weight changes when changing the priority for a few edges with different speeds

web/src/test/java/com/graphhopper/http/resources/CustomWeightingRouteResourceTest.java
```diff
@@ -85,18 +85,22 @@ public void testAvoidArea() {
...
         JsonNode path = yamlNode.get("paths").get(0);
+        // todonow: why use assertBetweeen?
         assertBetween("distance wasn't correct", path.get("distance").asDouble(), 500, 900);
+        //assertEquals(path.get("distance").asDouble(), 661, 10);
 
         yamlQuery += "priority:\n" +
-                "  area_custom1: 0.5\n" +
+                // todonow: the polygon used here is a bit weird and we really have to use a low priority to force the
+                // large detour
+                "  area_custom1: 0.05\n" +
```
easbar, line 95: Here it looks like you were trying to enforce the green route instead of the red one using the area shown as grey box:

image

But this seems a bit strange to me because both routes are crossing the area. With the new formula I was not able to enforce this route and instead it yields a route that really avoids the grey box (goes around it and therefore the distance is much longer)

easbar: Maybe this is exactly because the priority contribution depends on the speed: The green route is probably faster and therefore decreasing the priority does not increase the weighting as much as on the red route.

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeighting.java
```diff
@@ -102,33 +106,37 @@ public CustomWeighting(FlagEncoder baseFlagEncoder, EncodedValueLookup lookup,
...
 
     @Override
     public double getMinWeight(double distance) {
-        return distance / maxSpeed + distance * distanceInfluence;
+        return distance / maxSpeed + calcDistanceCosts(distance);
     }
 
     @Override
     public double calcEdgeWeight(EdgeIteratorState edgeState, boolean reverse) {
-        double distance = edgeState.getDistance();
```
karussell, line 121: BTW: the reason I "cached" the distance here is that it gave a minor speed up. (not sure if this is still true)

web/src/test/java/com/graphhopper/http/resources/CustomWeightingRouteResourceTest.java
karussell: Hmmh, ugly. I'll have a look why I chose this area and then accepted that the path was still going thought (might be the intention but I'm no longer 100% sure)

karussell:
Thanks a lot for this! The new formula has a nice property it seems. Also thanks for making it so clear and easy as I would have a hard time in understanding the effects of it just from the formula.

What I also like is that there was no need to change the yaml it seems (except for the area thing which could also be a faulty test).

> So for the old formula changing p means that we 'add' priority costs that are proportional to the speed costs while with the new formula choosing a value of p means we get certain priority costs regardless of the speed

"regardless of the speed", but still "dependent on the distance"? I.e. has meaning of `priority` switched from "proportional to speed" to "proportional to distance"?

> Generally I think the stress costs should be independent from the speed costs

Can you expand on this a bit? I think, this was also the motivation for you for this new formula (?)

> Honestly right now I am not sure what is better here.

Ok. Tough problem :)

I'm _currently_ slightly preferring the old formula because of its simple explanation: use `speed_factor` to avoid certain values, use `priority` to do the same, but without changing the ETA and use the `distance_influence` to e.g. avoid long-distance routes.

But I'm likely just more used to the old formula and I might as well oversimplify the situation, which you probably have better studied.

We could even name these two different priorities in a different way and expose them in the yaml, but especially at the beginning of this CustomWeighting adventure I would probably try to avoid this and pick one meaning (name it "priority") and in the worst case switch to the other meaning/formula. But which one :D ?

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeighting.java
easbar: ok I added a todo and lets check this when we have decided on the formula again

web/src/test/java/com/graphhopper/http/resources/CustomWeightingRouteResourceTest.java
easbar: This seems like a key example to me: We set the priority to some low value in this area and with the old formula this means we add a bigger penalty to the slower road and a smaller penalty to the faster road. Therefore we have this situation where one of the roads crossing the area is surpressed, but the other is not. With the new formula we apply the *same* penalty to both roads (per meter) and therefore they are both neglected. What is better? I don't know :) I guess that depends on what you want to achieve.

easbar:
> What I also like is that there was no need to change the yaml it seems (except for the area thing which could also be a faulty test).

Yes there is no difference. The priority is a number between 0 and 1 and the only question is how exactly this should enter the formula.

> "regardless of the speed", but still "dependent on the distance"? I.e. has meaning of priority switched from "proportional to speed" to "proportional to distance"?

For both formulas the 'priority costs' are proportional to the distance. Nothing has changed about this. The difference is just that in one case the penalty is proportional to the speed and in the other it is not. 

> Can you expand on this a bit? I think, this was also the motivation for you for this new formula (?)

Yes this was my original motivation, but I think when I started I did not fully realize that with custom weighting we are not setting any kind of costs explicitly, but rather apply some factors to change the default behavior. The question is simply what is meant when somebody sets something like `road_surface: {sand: 0.5}`. It certainly means that there should be a penalty for sandy roads. But should there be a larger penalty for sandy roads with a low speed? A slower road already means a higher weight (because of the 'time costs'), but with the old formula the penalty for the road having a sandy surface is also higher for slower roads. In contrast the new formula applies the same penalty (per meter) to the slow and the fast sandy roads. I guess this was part of my original motivation: A slow road should of course have a higher weight, but this is already achieved by the 'time costs' and there is no need to also add a larger penalty to the slower road again when adding the 'priority costs'.

> I'm currently slightly preferring the old formula because of its simple explanation: use speed_factor to avoid certain values, use priority to do the same, but without changing the ETA and use the distance_influence to e.g. avoid long-distance routes.

Ok but this has not changed with the new formula. Setting the priority to something < 1 still means avoiding certain kinds of roads and it does not change the ETA. This simple message still applies. What is slightly more simple with the old formula is the 'size' of the priority costs. 

> We could even name these two different priorities in a different way and expose them in the yaml, but especially at the beginning of this CustomWeighting adventure I would probably try to avoid this and pick one meaning (name it "priority") and in the worst case switch to the other meaning/formula. 

Yes for sure I would not introduce another parameter in the configuration either and this is also nice, it should be easy to change the formula any time if there is need to do this.

> But which one :D ?

I am still not sure :D.

easbar:
Do you have a common use case for priority? `p=1` obviously is the default, `p=0` means blocking a road (or area), but when do we even use `0<p<1`?

easbar:
Here I plotted the two formulas for `f=0.05 s/m, s=0.1s/m` and some speeds ranging from `v=5m/s to 30m/s` 

image

We can see that with the old formula the priority affects the weight of slower roads more than faster ones and around `p=0.3` or so there is a strong difference between the weights of slow and fast roads. With the new formula both fast and slow roads are affected  by the priority in a similar way (say for `p>0.6`), but for low priorities the weights are rather similar (because the priority costs simply outweigh the 'time' costs). 

But honestly this kind of qualitative analysis has not helped me much in deciding which one we should prefer. We probably need some real-life examples to make a better decision and as far as I am concerned we can keep the old formula until we feel like we need to change something.

karussell:
Interesting. And the avoid area example could be indeed like a key example and it is indeed unexpected that roads inside the area get a different penalty.

> But honestly this kind of qualitative analysis has not helped me much in deciding which one we should prefer

:)

> Do you have a common use case for priority? and what do we expect to happen

There are many examples like avoiding motorways for car, or avoid sand for bike, or avoiding ferries, or tolls or smaller roads for trucks, or avoiding destination-only roads ... or the opposite like preferring official bike routes means avoiding everything else.

In general we expect that the weight increases and do not care about how exactly. But it sounds bad to put a bigger penalty on the roads with slower speed. I can imagine that for big speed differences that occur e.g. for walking (5kmh) when they use a ferry (30kmh) this somehow could become more obvious?

> but this has not changed with the new formula. 

:+1: 

> > But which one :D ?
> I am still not sure :D.

At least I'm now also undecided and not in favor of the old anymore ;)

web/src/test/java/com/graphhopper/http/resources/CustomWeightingRouteResourceTest.java
karussell: The old formula also needs a 0.05 priority to force the long detour, but yes, for all values until 0.05 the old formula prefers the longer way where the new prefers the shorter. So here the new formula makes more sense as inside the avoided area it influences the original path the least.

And yes, we should remove assertBetween and replace with assertEquals and a bigger delta.

easbar:
Right now we do not seem to come to a clear decision which formula is the best, so let's stick with the old one for now.

Just for future reference: Another possibility would be:

```
w = d/v + d*f/(2*p) + d*f/2
```

or simply

```
w = d/v + d*f/p
```

These formulas allow scaling the priority casts using the distance_influence factor, which also makes sense, because applying priorities < 1 always induces distance costs. Setting f=0 does not allow any priority effects, but this just expresses that when we include priorities we can no longer have a pure fastest weighting.

PR #2023, opened 2020-04-29
Use profiles in web UI / graphhopper maps

### LINKED ISSUE #1993: Update web demo to use profiles
After #1934 we should probably replace vehicle, weighting and turn_costs with profiles in the web UI.

easbar:
Fixes #1993.


web/src/main/resources/assets/js/graphhopper/GHRequest.js
```diff
@@ -118,34 +119,60 @@ GHRequest.prototype.getEarliestDepartureTime = function () {
...
+    return this.api_params.profile;
 };
 
-GHRequest.prototype.hasTCSupport = function() {
-   if(this.api_params.turn_costs !== false) {
-      var featureSet = this.features[this.api_params.vehicle];
-      this.api_params.turn_costs = featureSet && featureSet.turn_costs;
-   }
+GHRequest.prototype.setElevation = function (elevation) {
+    this.api_params.elevation = elevation;
```
easbar, line 131: elevation is now treated like a global parameter (instead of being different for every vehicle). for one server this is always the case, but in case there were multiple servers some of which do not support elevation this could be a problem?

web/src/main/resources/assets/js/graphhopper/GHRequest.js
```diff
@@ -118,34 +119,60 @@ GHRequest.prototype.getEarliestDepartureTime = function () {
...
+            this.api_params["weighting"] ||
+            this.api_params["turn_costs"] ||
+            this.api_params["edge_based"]
+        ) {
+            delete this.api_params.profile;
+        }
+     // todonow: pasting a legacy url already works the only thing that is ugly is that when we do a normal request
+     // (including profile) then add something like vehicle=car the the profile parameter appears still in the url field
+     // (even though it will not be sent to the server). when we drag a marker afterwards it finally disappears, how can
+     // we do this already here?
```
easbar, line 166: @karussell would you mind taking a look at this? I did not manage to fix this and not even sure if this needs to be fixed.

web/src/main/resources/assets/js/graphhopper/GHRequest.js
```diff
@@ -232,6 +259,10 @@ GHRequest.prototype.doRequest = function (url, callback) {
     $.ajax({
         timeout: 30000,
         url: url,
+        beforeSend: function(request) {
+            // todonow: maybe rename
+            request.setRequestHeader("gh-client", "web-ui")
```
easbar, line 264: this can be used to recognize that a request was sent using this client

web/src/main/resources/assets/js/main-template.js
```diff
@@ -34,11 +34,6 @@ if (ghenv.environment === 'development') {
     var autocomplete = AutoComplete.prototype.createStub();
 } else {
     var autocomplete = new AutoComplete(ghenv.geocoding.host, ghenv.geocoding.api_key);
-    // overwrite default for production
-    GHRequest.prototype.hasTCSupport = function() {
-       if(this.api_params.turn_costs !== false)
-          this.api_params.turn_costs = new Set(["car", "truck", "small_truck", "scooter"]).has(this.api_params.vehicle);
-    };
```
easbar, line 41: this can probably be done server-side (using the header telling us the request comes from the client)

web/src/main/resources/assets/js/main-template.js
```diff
@@ -239,6 +252,17 @@ $(document).ready(function (e) {
     checkInput();
 });
 
+function profileDisplayName(profileName) {
+   return translate.tr(profileName) + ' (' + profileName + ')'
```
easbar, line 256: Any better idea? For profile names like `my_special_super_car` we will get something like "web.my_special_super_car (my_special_super_car)"...

web-bundle/src/main/java/com/graphhopper/http/PtRedirectFilter.java
```diff
@@ -40,6 +40,13 @@ public void filter(ContainerRequestContext requestContext) {
...
         String maybeVehicle = requestContext.getUriInfo().getQueryParameters().getFirst("vehicle");
-        return maybeVehicle != null && maybeVehicle.equals("pt");
+        if (maybeVehicle != null && maybeVehicle.equals("pt")) {
+            return true;
+        }
+        String maybeProfile = requestContext.getUriInfo().getQueryParameters().getFirst("profile");
+        if (maybeProfile != null && maybeProfile.equals("pt")) {
+            return true;
+        }
+        return false;
```
karussell, line 50: this could be a bit shorter ;)

`return "pt".equals(maybeProfile)`

And should we also remove the profile param in the filter method above? (`replaceQueryParam("profile")`)

web/src/main/resources/assets/js/graphhopper/GHRequest.js
easbar: the problem is this:

1) use the UI (profile will be in the address line)
2) manually add something like `vehicle=car` and press enter
3) the profile will not be send to the server (only the vehicle), but the address line is not updated accordingly
4) move a marker. same as 3) but this time the profile is removed from the address line

> And shouldn't we remove the vehicle parameter from the URL if vehicle and profile parameter exist in the URL?

Yes that sounds right. So we only support 'real' legacy requests (without profile) and the (now) standard (including (only) a profile). I probably tried this but ran into similar problems as above?

web/src/main/resources/assets/js/graphhopper/GHRequest.js
easbar: > Is there a way we could include some version?

Sure we could add another header with a version. Would be nice to get this via maven or even git commit hash or tag during from the build process? Or simply hard-code it ;)

> And isn't it usually called "User-Agent"

Yeah I think user-agent is used for browser version, using a custom / non-standard header for what we want to achieve here seems reasonable to me

web/src/main/resources/assets/js/main-template.js
easbar: Initially I was concerned that some non-standard custom names cannot be translated, but you are right probably just one of the two is enough. If we remove the info in brackets we would get something like `web.my_special_car` for non-standard names, and something like `Auto` for standard-names like `car`. This could be slightly improved by removing the `web.` for non-standard names. Is this what you mean?

easbar:
> Have done this using the parameter from the URL (?)

Yes perfect that makes sense, right?

web-bundle/src/main/java/com/graphhopper/http/PtRedirectFilter.java
easbar: > this could be a bit shorter ;) return "pt".equals(maybeProfile)

Ah yes I was trying to keep the symmetry :)

> And should we also remove the profile param in the filter method above? 

Good idea, yes, I think so.

web/src/main/resources/assets/js/graphhopper/GHRequest.js
karussell: > Would be nice to get this via maven or even git commit hash or tag during from the build process?

Ah, yes. This sounds doable but for a later issue ;)

web/src/main/resources/assets/js/main-template.js
karussell: Yes

web/src/main/resources/assets/js/graphhopper/GHRequest.js
boldtrn: I would not recommend overwriting the User-Agent.

IIRC a lot of browsers don't allow that, also some CDNs/Firewalls might check that string and if it looks tempered they might assume something funny is going on and potentially block the request.

easbar: > Ah, yes. This sounds doable but for a later issue ;)

We could still hard-code it for now, that would be more information than not sending any version (even though it might be outdated sometimes) :)

otbutz: You could take a look at the buildnumber maven plugin: https://www.mojohaus.org/buildnumber-maven-plugin/usage.html

web/src/main/resources/assets/js/main-template.js
easbar: Cool yes this is better, done.

web/src/main/resources/assets/js/graphhopper/GHRequest.js
karussell: > manually add something like vehicle=car and press enter

Ah, ok. Then this is not a legacy use case and either the profile parameter needs to be removed or instead of the vehicle the profile needs to be changed.

> So we only support 'real' legacy requests (without profile) and the (now) standard (including (only) a profile)

Yes

> I probably tried this but ran into similar problems as above?

Unfortunately I was not able to get this mixture (yet) using a legacy URL.

easbar: > Ah, ok. Then this is not a legacy use case and either the profile parameter needs to be removed or instead of the vehicle the profile needs to be changed.

Yeah its really a very minor issue / very exotic '(mis)-use-case'. I was just annoyed by the fact that moving the marker does something that pressing enter does not do. Even when doing this (add vehicle to a profile url manually) the UI keeps working (using the vehicle), it just does not remove the profile from the url (and yes maybe it would better remove the vehicle in this case but this is probably the same problem).

easbar: Shall we do anything about this?

karussell: I tried to deploy this code to production and had a problem that although only profile=car was specified in the URL still the weighting was added in the request (resulting in an error). It could be that this is some internal problem as the /info endpoint is not yet updated. Will need to investigate this further but we can also merge here and fix later.

PR #2026, opened 2020-05-06
check profile if used for LM or CH preparation

### LINKED ISSUE #2025: Add a hash value for profiles
Right now its very easy to:

1) setup some profiles and do the import/preparation of the graph data
2) shut down the server
3) change the profiles in config.yml again
4) restart the server but 'forget' deleting the graph-cache folder
5) get bad routing results because the profiles no longer 'match' the preparations

One easy way to detect this error would be calculating and storing some kind of hash code for each profile.

See this discussion: https://github.com/graphhopper/graphhopper/issues/1999#issuecomment-618910167

karussell:
This fixes #2025

@easbar the CH and LM stuff is handled differently e.g. CH is coupled to the graph and so the LMConfig is only known at GraphHopper class level. And LM preparation is either loaded or created so this leads to the asymmetric new code (when comparing CH and LM) inside the GraphHopper class. 

It would be probably better to have this code in Landmark.createLandmarks & loadExisting but for this not only the LMConfig is necessary but also profilesByName. Same is for CH: this should be better encapsulated in PrepareContractionHierarchies.doWork and GraphHopperStorage.loadExisting (hmmh, ok this is also not correct and all the shortcuts etc. should be in CHGraphImpl or something or probably in a separate per-CHProfile-object)

### DIFF
--- a/api/src/main/java/com/graphhopper/config/Profile.java
+++ b/api/src/main/java/com/graphhopper/config/Profile.java
@@ -98,7 +98,7 @@ public class Profile {
 
     @Override
     public String toString() {
-        return "name=" + name + "|vehicle=" + vehicle + "|weighting=" + weighting + "|turnCosts=" + turnCosts + "|hints=" + hints;
+        return createContentString();
     }
 
     @Override
@@ -109,12 +109,17 @@ public class Profile {
         return name.equals(profile.name);
     }
 
+    private String createContentString() {
+        // used to check against stored custom models, see #2026
+        return "name=" + name + "|vehicle=" + vehicle + "|weighting=" + weighting + "|turnCosts=" + turnCosts + "|hints=" + hints;
+    }
+
     @Override
     public int hashCode() {
         return name.hashCode();
     }
 
     public int getVersion() {
-        return Helper.staticHashCode(toString());
+        return Helper.staticHashCode(createContentString());
     }
 }
--- a/api/src/main/java/com/graphhopper/config/Profile.java
+++ b/api/src/main/java/com/graphhopper/config/Profile.java
@@ -18,6 +18,7 @@
 
 package com.graphhopper.config;
 
+import com.graphhopper.util.Helper;
 import com.graphhopper.util.PMap;
 
 /**
@@ -111,5 +112,9 @@ public class Profile {
     @Override
     public int hashCode() {
         return name.hashCode();
+    }
+
+    public int getVersion() {
+        return Helper.staticHashCode(toString());
     }
 }
--- a/api/src/main/java/com/graphhopper/routing/util/CustomModel.java
+++ b/api/src/main/java/com/graphhopper/routing/util/CustomModel.java
@@ -132,7 +132,7 @@ public class CustomModel {
                 ", maxSpeed=" + maxSpeedMap +
                 ", maxSpeedFallback=" + maxSpeedFallback +
                 ", priorityMap=" + priorityMap +
-                ", #areas=" + areas.size() +
+                ", areas=" + areas +
                 '}';
     }
 
--- a/api/src/main/java/com/graphhopper/routing/util/CustomModel.java
+++ b/api/src/main/java/com/graphhopper/routing/util/CustomModel.java
@@ -126,14 +126,13 @@ public class CustomModel {
 
     @Override
     public String toString() {
-        return "CustomModel{" +
-                "distanceInfluence=" + distanceInfluence +
-                ", speedFactor=" + speedFactorMap +
-                ", maxSpeed=" + maxSpeedMap +
-                ", maxSpeedFallback=" + maxSpeedFallback +
-                ", priorityMap=" + priorityMap +
-                ", areas=" + areas +
-                '}';
+        return createContentString();
+    }
+
+    private String createContentString() {
+        // used to check against stored custom models, see #2026
+        return "distanceInfluence=" + distanceInfluence + ", speedFactor=" + speedFactorMap + ", maxSpeed=" + maxSpeedMap +
+                ", maxSpeedFallback=" + maxSpeedFallback + ", priorityMap=" + priorityMap + ", areas=" + areas;
     }
 
     /**
--- a/api/src/main/java/com/graphhopper/routing/util/CustomModel.java
+++ b/api/src/main/java/com/graphhopper/routing/util/CustomModel.java
@@ -131,8 +131,8 @@ public class CustomModel {
 
     private String createContentString() {
         // used to check against stored custom models, see #2026
-        return "distanceInfluence=" + distanceInfluence + ", speedFactor=" + speedFactorMap + ", maxSpeed=" + maxSpeedMap +
-                ", maxSpeedFallback=" + maxSpeedFallback + ", priorityMap=" + priorityMap + ", areas=" + areas;
+        return "distanceInfluence=" + distanceInfluence + "|speedFactor=" + speedFactorMap + "|maxSpeed=" + maxSpeedMap +
+                "|maxSpeedFallback=" + maxSpeedFallback + "|priorityMap=" + priorityMap + "|areas=" + areas;
     }
 
     /**
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -1052,31 +1052,17 @@ public class GraphHopper implements GraphHopperAPI {
 
         try {
             validateRequest(request);
-            PMap hints = request.getHints();
             final boolean disableCH = getDisableCH(request.getHints());
             final boolean disableLM = getDisableLM(request.getHints());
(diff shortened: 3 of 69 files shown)


reader-osm/src/test/java/com/graphhopper/reader/osm/GraphHopperOSMTest.java
```diff
@@ -1117,4 +1117,56 @@ private GraphHopper createGraphHopper(String vehicles) {
                 setEncodingManager(em).
                 setProfiles(profiles);
     }
+
+    @Test
+    public void testLoadingWrongCHProfiles() {
```
karussell, line 1122: This is probably better in GraphHopperTest but mocking DataReader is a bit lengthy/ugly.

core/src/main/java/com/graphhopper/storage/CHConfig.java
```diff
@@ -70,4 +71,13 @@ public int hashCode() {
...
 
+    public int getVersion() {
+        int hash = Helper.staticHashCode(chGraphName);
+        hash = 31 * hash + Helper.staticHashCode(weighting.getName());
+        // if flag encoder is removed we could use weighting.toString
+        hash = 31 * hash + Helper.staticHashCode(weighting.getFlagEncoder().toString());
+        hash *= weighting.hasTurnCosts() ? 31 : 62;
+        hash *= edgeBased ? 31 : 62;
+        return hash;
+    }
```
easbar, line 82: Looks like this is (no longer(?)) needed?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -979,8 +979,16 @@ protected void postProcessing(boolean closeEarly) {
 
         if (chPreparationHandler.isEnabled())
             chPreparationHandler.createPreparations(ghStorage);
-        if (!isCHPrepared())
+        if (isCHPrepared()) {
+            // check loaded profiles
+            for (CHProfile profile : chPreparationHandler.getCHProfiles()) {
+                if (!getProfileHash(profile.getProfile()).isEmpty()
```
easbar, line 985: Why is `isEmpty()` treated special here? Why is it possible for the hash to be empty?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1264,16 +1272,24 @@ protected void initLocationIndex() {
...
+        return "true".equals(ghStorage.getProperties().get(CH.PREPARE + "done"));
     }
 
-    private boolean isLMPrepared() {
-        return "true".equals(ghStorage.getProperties().get(Landmark.PREPARE + "done"));
+    private String getProfileHash(String profile) {
+        return ghStorage.getProperties().get("graph.profiles." + profile + ".hash");
+    }
+
+    private void setProfileHash(String profile, int version) {
```
easbar, line 1282: Can we either call it hash or version (but not both :)) ?

reader-osm/src/test/java/com/graphhopper/reader/osm/GraphHopperOSMTest.java
easbar: Ok ;)

api/src/main/java/com/graphhopper/config/Profile.java
```diff
@@ -112,4 +113,8 @@ public boolean equals(Object o) {
     public int hashCode() {
         return name.hashCode();
     }
+
+    public int getVersion() {
+        return Helper.staticHashCode(toString());
```
easbar, line 118: Hmm toString includes the hints (which is good/required for custom weighting), but unfortunately `PMap` uses an (unordered) `HashMap`, so this might go wrong? Shall we use `LinkedHashMap` in `PMap`? Or implement some method that returns the string in a deterministic way?

easbar: Oh but I just realized that also `hints` does not even include the custom model but only the reference to the custom model file... Shouldn't our check also detect changes in the custom model file?

easbar:
What do you mean? Isn't it symmetric in `GraphHopper`: `prepareCH` and `loadOrPrepareLM`?

easbar:
In the mid/long run we should store the configurations/profiles *with* the prepared data and read it from the same location instead of storing it separately and requiring to provide the same configuration that was used for the preparation again.. -> #1895

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: Ah, I just blindly assumed the symmetry between CH and LM. So, yes. It is not necessary for CH as we want to load it here.

core/src/main/java/com/graphhopper/GraphHopper.java
karussell: Haha, sure. I even had a third option in the code before ;)

api/src/main/java/com/graphhopper/config/Profile.java
karussell: > but unfortunately PMap uses an (unordered) HashMap, so this might go wrong

Ah, good catch!

> Oh but I just realized that also hints does not even include the custom model but only the reference to the custom model file... Shouldn't our check also detect changes in the custom model file?

Will write a test for this. I think we load the custom profile as object into this, but it might happen after the check.

karussell:
> What do you mean? Isn't it symmetric in GraphHopper: prepareCH and loadOrPrepareLM?

Yes, but also that CH stuff is coupled to the graph and LM isn't and so things like this can/must be handled differently.

> But for now I think its an improvement to check this hash somehow :)

Ok

api/src/main/java/com/graphhopper/config/Profile.java
karussell: The CustomModel is stored in the hints already and produces a different hash code for mostly all situation, but relies on toString. So my test case passes. But only "mostly" because for areas we currently include only the count and so e.g. for changes of an area there would be no error. I will include the full areas and if this is too much at some point we can revert this to the count.

But the question is if we should rely on toString which should have only a debugging purpose? I think, yes, for now as the hashcode is not stable. And later work towards a config that is stored in the graph storage (and is loaded without providing it).

> Shall we use LinkedHashMap in PMap?

Yes, I'll do that (the other option would be TreeMap or putting toMap into a TreeMap but I think it cannot hurt to give PMap this ordering guarantuee)

easbar: Cool thanks. We could have some method createContentString and call it from createHashCode **and** toString. This way its obvious that it has some meaning beyond debugging, we do not have to write it 'twice' and maybe we could even do the sorting there (in case we do not want to adjust PMap).

karussell: Hmmh, sounds good but in PMap this still looks suboptimal:

```java
public String createContentString() {
    StringBuilder sb = new StringBuilder();
    for(Map.Entry<String, Object> entry : map.entrySet()) {
        sb.append(entry.getKey());
        sb.append('=');
        if(entry.getValue() instanceof ContentStringCreator)
            sb.append(((ContentStringCreator)entry.getValue()).createContentString()); // let e.g. CustomModel implement this interface
        else
            sb.append(entry.getValue().toString());
        sb.append(',');
    }
    return sb.toString();
}
```
WDYT?

karussell: Ah, or do you mean to implement createContentString only in CustomModel as private method and use it in toString?

easbar: > Ah, or do you mean to implement createContentString only in CustomModel as private method and use it in CustomModel.toString?

Yes

karussell:
Can I merge here?

api/src/main/java/com/graphhopper/routing/util/CustomModel.java
```diff
@@ -126,14 +126,13 @@ public double getHeadingPenalty() {
...
-                ", speedFactor=" + speedFactorMap +
-                ", maxSpeed=" + maxSpeedMap +
-                ", maxSpeedFallback=" + maxSpeedFallback +
-                ", priorityMap=" + priorityMap +
-                ", #areas=" + areas.size() +
-                '}';
+        return createContentString();
+    }
+
+    private String createContentString() {
```
easbar, line 132: Should we not use createContentString also in getVersion() (instead of toString()) now that we have it?

easbar: Ah sorry its in another file. But maybe we can do the same in `Profile.java` (add `createContentString` and use it from `Profile#getVersion()` and `Profile#getVersion()`)?

karussell: Yes!

PR #2027, opened 2020-05-08
better estimate factor to store landmark weights more precise

karussell:
In production this improves visited nodes by at least 10% for a few examples. In our benchmarks this is ~3% faster, but could be also be noise. At least it should improve the warning logs for smaller graphs.


karussell:
Ok, this works good now and improves the query speed for global coverage (and smaller). 

In a next step we should re-use the landmarks found in estimateMaxWeight later and do this for all landmarks (and slightly reduce the current 1.008 factor to 1.002 which could further improve query speed a bit). But for now this works good enough and doesn't slow down the prepare.lm.time too much.

PR #2029, opened 2020-05-11
Update Isochrone documentation

NoamDev:
I realized it would be better to send a basic pr than create an issue.
I'm not an experienced documentor, so please review thoroughly.

### DIFF
--- a/docs/isochrone/java.md
+++ b/docs/isochrone/java.md
@@ -14,22 +14,24 @@ Next, compute the isochrone itself.
 EncodingManager encodingManager = hopper.getEncodingManager();
 FlagEncoder encoder = encodingManager.getEncoder("car");
 
+EdgeFilter edgeFilter = DefaultEdgeFilter.allEdges(encoder);
+LocationIndex locationIndex = graphHopper.getLocationIndex();
+
 // pick the closest point on the graph to the query point and generate a query graph
 QueryResult qr = hopper.getLocationIndex().findClosest(lat, lon, DefaultEdgeFilter.allEdges(encoder));
 
 Graph graph = hopper.getGraphHopperStorage();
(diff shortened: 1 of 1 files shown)


NoamDev:
Someone was faster than me:)

easbar:
> Someone was faster than me:)

Oh sorry I did not know you are working on this as well, is there still something you would like to add? Thanks for your effort in any case.

NoamDev:
Hmm, I think it should be noted that labels might update.
I mean, If the tree got to the same traversal id again, it'd send an updated label.
Maybe the example should insert all the labels to a map, so that updated labels would overwrite the old ones.

NoamDev:
You can look on my commit for example.

To be honest, I think it should actually be a method.
It could be very confusing.

easbar:
Ok sure sounds good feel free to update/re-open your PR.

NoamDev:
Sorry, could you please say that again?
seems you wrote fast and I'm unable to understand it. Probably my fault as I;m not a native english speaker..

easbar:
I had a look at `ShortestPathTree` again and I do not think that the labels are updated multiple times as `consumer.accept(label)` is called just after polling from the queue (just like I would have expected it to be). 

So I think:

>  If the tree got to the same traversal id again, it'd send an updated label.

does never happen. Do you have an example that shows labels that get called multiple times? Maybe you were using `TraversalMode.EDGE_BASED`? So currently I think there is no need to mention this in the docs and we do not need to build some `HashMap` for the example either.

NoamDev:
You seem to be right!
I couldn't create a test case, and after some debugging, I realized it cannot happen.
Because it always pulls the lightest route, and a route which started heavy can never become lighter.
Routing algorithms are such an amazing thing!
Is that a _greedy_ algorithm?

Issue #2036, opened 2020-05-14
Enforce allowWrites for GTFS files


gcheval:
**Describe the bug**

With GraphHopper.allowWrites to **false**, GTFSFeed will still attempt to write in the graph.location folder (even if the gtfs file exists).

The exception returned is: `java.io.IOError: java.io.IOException: Parent folder is not writable: graph-cache/gtfs_0`.

I was able to circumvent that issue by applying readOnly() to the DBMaker in GTFSFeed.constructDB if the file exists. 

**To Reproduce**

Steps to reproduce the behavior. For examples:

1. checkout **master** branch of GraphHopper.
2. Setup the project following the [public transit quickstart](https://github.com/graphhopper/graphhopper/blob/master/reader-gtfs/README.md#quick-start).
3. Remove the write permission on the **./graph-cache** folder (`chmod -w ./graph-cache`).
4. Set allowWrites to **true** for GraphHopperGTFS in `com.graphhopper.http.GraphHopperManaged#graphHopper`.
5. Start the application.
4. See exception in the console.

**Expected behavior**

When GraphHopper has setAllowWrites to **true**, DBMaker in GTFSFeed should attempt to load GTFS DB files as readOnly.

**System Information**

Operating system: macOS Mojave 10.14.6 (18G4032)
GraphHopper: **1.0-pre38** and **master**
JVM:
```
openjdk version "11.0.5" 2019-10-15
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.5+10)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.5+10, mixed mode)
```

**Screenshots & Logs**

- Complete stacktrace:
```
java.io.IOError: java.io.IOException: Parent folder is not writable: graph-cache/gtfs_0
	at org.mapdb.Volume$MappedFileVol.<init>(Volume.java:543)
	at org.mapdb.Volume.volumeForFile(Volume.java:202)
	at org.mapdb.Volume$1.createIndexVolume(Volume.java:276)
	at org.mapdb.StoreDirect.<init>(StoreDirect.java:202)
	at org.mapdb.DBMaker.extendStoreDirect(DBMaker.java:971)
	at org.mapdb.DBMaker.makeEngine(DBMaker.java:758)
	at org.mapdb.DBMaker.make(DBMaker.java:701)
	at com.conveyal.gtfs.GTFSFeed.constructDB(GTFSFeed.java:368)
[...]

PR #2038, opened 2020-05-14
Enforce allowWrites for GTFS files #2036:

gcheval:
- Attempt to create GTFS storage ONLY if allowWrites == true
- Load GTFS feeds as readOnly if dbFile already exists
- Validate dbFile exists when allowWrites is false


michaz:
I even think loading a previously imported DB should never require write access.
Should we try to 
- do without the new parameter on loadExisting()
- always load previously-imported DBs in read-only mode
- .. and only check the allowWrites-flag once, namely before trying to import a DB that does not exist? Compare ensureWriteAccess() in GraphHopper.importData

michaz:
(I have a vague memory that this may not work due to something I did for the GtfsRealtime-part. _Maybe_ I modify the DB there, for technical reasons.... :-/)

michaz:
#2036

gcheval:
Hi Michaz,

Applied your comments on the PR. Thanks for your assistance!

- When we find a transit_schedule file, we will attempt to load mapped files and throw a RuntimeException if the files do not reflect the transit_schedule mapping.
- If there's no transit_schedule file, we now validate the write access before attempting to create.

gcheval:
Hey @michaz, What's the remaining actions before getting that PR into a release? Thanks!

michaz:
None. Thank you, will get to this shortly.

michaz:
Merged with db29f6da3bf15d93de933b23dc870ab93f573cda.

https://github.community/t5/GitHub-API-Development-and/Feature-Request-mark-pull-request-as-merged/m-p/40295/highlight/true#M3645

PR #2041, opened 2020-05-15
include npm as part of the maven cycle

### LINKED ISSUE #1700: execute npm as part of the maven cycle

karussell:
This fixes #1700.

Setting up npm is not required for users that only need the API (no web UI), Android or Desktop. But we should still include it in the maven cycle. So I decided to give it a try via a maven profile and the plugin seems to work properly.

The npm_modules are duplicated in web/node/npm_modules in parallel to the "normal" web/npm_modules. Is this bad? (IMO this is fine as we have the maven node&npm stuff and the other is the real locally installed node&npm. But it could also be confusing?)

TODO:

 - [x] adapt javascript docs
 - [x] remove main.js for `mvn clean`

You might want to have a look into this change @easbar @michaz @Nakaner

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ debug.sh
 /web/src/main/resources/assets/js/main.js
 /web/src/main/resources/assets/js/config/options_prod.js
 /web/dependency-reduced-pom.xml
+/web/node
 *.pbf
 !/reader-osm/src/test/resources/com/graphhopper/reader/osm/*.pbf
 *.dem
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,10 +35,9 @@ before_script:
   - _JAVA_OPTIONS=
   - "mvn --version"
   - "if [ -z \"$API_KEY\" ]; then API_KEY=78da6e9a-273e-43d1-bdda-8f24e007a1fa; fi" # change in GraphHopperWebIT too
-  - cd web && npm install && BROWSERIFYSWAP_ENV='development' npm run bundleProduction && cd ..
 
 script:
-  - "mvn -Dkey=$API_KEY clean install -B"
+  - "mvn -Pinclude-web-ui -Dkey=$API_KEY clean install -B"
 
 after_success:
   # often spotbugs etc take long to be compatible with a future JDK version so skip them
--- a/config-example.yml
+++ b/config-example.yml
@@ -176,7 +176,7 @@ graphhopper:
 #
 # assets:
 #  overrides:
-#    /maps: web/src/main/resources/assets/
+#    /maps: web/target/classes/assets/
 
 # Dropwizard server configuration
 server:
--- a/core/files/changelog.txt
+++ b/core/files/changelog.txt
@@ -1,15 +1,19 @@
 1.0
+    changes to config.yml:
+      properties have to follow the snake_case, #1918, easy convert via https://github.com/karussell/snake_case
+      all routing profiles have to be configured when setting up GraphHopper, #1958, #1922
+      for developing with JavaScript, npm and the web UI it is important to change your assets overrides to web/target/classes/assets/, see #2041
+    moved SPTEntry and ShortcutUnpacker to com.graphhopper.routing(.ch)
+    renamed PathWrapper to ResponsePath
     moved package from core/src/test/java/com/graphhopper/routing/profiles to ev
     removed HintsMap
     removed /change endpoint
     removed vehicle,weighting and edge_based from GraphHopper class, replaced with new profile parameter, #1958
-    all routing profiles have to be configured when setting up GraphHopper, #1958
     there no longer is a default vehicle, not setting the vehicle now only works if there is exactly one profile that matches the other parameters. anyway you should use the new profile rather than the vehicle parameter.
     removed IPFilter. Use a firewall instead.
     PMap refactored. It is recommended to use putObject(String, Object) instead of put, #1956
     removed UnsafeDataAccess as not maintained, see #1620
     add profiles parameter and replace prepare.ch/lm.weightings and prepare.ch.edge_based with profiles_ch/lm config parameters, #1922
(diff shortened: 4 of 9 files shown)


otbutz:
I think that npm allows to define a prefix which would solve the duplicate node_modules location

michaz:
Yes please, it's great that you found a way to build this with Maven alone.

I'm not sure about the bit with the profile. What's the problem with just including it in the build, period? Combinatorial builds are a problem source.

Why would anyone want to avoid a build step that takes like 10 seconds, including downloading node? Maven always downloads and builds a hell of a lot, so I don't think this is a special issue.

michaz:
> main.js is not removed for mvn clean. Is this ok?
> main.js is overwritten even if it exists. Is this ok?

Nope, not okay. The gernerated main.js should go directly into the build target, not into the source directory. Boom, problem disappears.

otbutz:
I agree with @michaz that any generated artifacts should be put in the target folder. Creating files in the source folder is messy.

michaz:
Moreover, the _source_ .js-files _are_, in fact, source, not assets, and should probably go into something like `src/main/js`? (Not sure if the plugin has recommendations for that?)

karussell:
> I'm not sure about the bit with the profile. Why not just include it in the build, period? Combinatorial builds are a problem source. Why would anyone want to avoid a build step that takes like 10 seconds, including downloading node? Maven always downloads and builds a hell of a lot, so I don't think this is a special issue.

Ok :) 

> Moreover, the source .js-files are, in fact, source, not assets, and should probably go into something like src/main/js? (Not sure if the plugin has recommendations for that?)

There are no recommendations to my knowledge. 

> Nope, not okay. The generated main.js should go directly into the build target, not into the source directory. (It's not source.) 

Would be nice. Will try.

michaz:
(Conversely, in the pt GUI, all the JS files _are_ really assets, because they go into the web server just as they are.)

michaz:
Which reminds me, if this works really well, maybe I can use it for the pt GUI, too. Then I could put those React-tags back into the JS, which would make Janek happy I think. :-)

karussell:
> I think that npm allows to define a prefix which would solve the duplicate node_modules location

The question is if this would be good. As the local npm and the maven-npm could be different and break things. So I currently prefer the situation as it is.

karussell:
> Nope, not okay. The generated main.js should go directly into the build target, not into the source directory. (It's not source.)

Ok, this seems to work now. At least for my small tests and for maven.

IntelliJ seems to ignore the plugin settings, but once the main.js is in the target folder IntelliJ does not remove it and picks it up properly (IF you have the correct assets overrides specified in the config.yml which requires a change for your old config.yml :) ). Is this expected or should IntelliJ execute the plugin when bound to `generate-resources`?

btw: it is very important to always open a private window in your browser to avoid caching ugliness and false positives :)

karussell:
> Moreover, the source .js-files are, in fact, source, not assets, and should probably go into something like src/main/js?

Ah, sure. As I break already lots of things here I'll try to move assets to src/main/js

michaz:
I don't expect IntelliJ to execute the plugin, though it would be nice.

But when you execute the maven build once, and then start the server, _without_ changing any paths or configs, the server must pick up the JS and work. Even when started from IntelliJ. Otherwise it's going to the wrong place.

michaz:
If the solution relies on this "changeable asset path" thing to work, something is wrong.

karussell:
> But when you execute the maven build once, and then start the server, without changing any paths or configs, the server must pick up the JS and work. Even when started from IntelliJ. Otherwise it's going to the wrong place.

Yes, yes. That will work :)
The change to the config is only necessary to see edits to the JS files immediately using `npm run watch`.

michaz:
Ah, right, okay.

karussell:
(As this hot-reload is important for developing the UI I already feared this is broken now ... until I figured out that the config in use did not yet point to the target folder ...)

Nakaner:
I haven't looked in detail into the pull request yet but the following came to my mind:

* You say, it downloads Node. What about systems where Node is already installed? If I have installed Node from package manager of my Linux distribution, I don't want to have a second copy. Node's packaging generates too much copies already. :-)
* There can be cases where one wants to set up the routing API but no web frontend. It would be nice to be able to build GraphHopper without having Node as a dependency. This does not have to be the default. Documenting it somewhere or add an extra option to `graphhopper.sh` would be awesome.

michaz:
> You say, it downloads Node. What about systems where Node is already installed? If I have installed Node from package manager of my Linux distribution, I don't want to have a second copy. Node's packaging generates too much copies already. :-)

I see what you mean, but Maven builds are usually expected to be self-contained, except for Maven and the JVM. Node is basically a guest in a Maven environment here.

From the perspective of Maven, builds and artifacts, the current situation is just wrong: I say `mvn package`, but what I get is a .jar where a _required_ piece is missing, namely `main.js`. When I point my browser to the advertised address, I get an error message that `main.js` wasn't found.

Explaining this by "the web client is optional" is wrong IMHO. It isn't optional, it's expected to be there by the rest of the system, it just isn't there.

otbutz:
> I don't want to have a second copy.

You don't until your systems Node version can't be used for the build.

easbar:
The maven frontend plugin also has a few options to skip `npm` execution etc. 
https://github.com/eirslett/frontend-maven-plugin#skipping-execution

In our case: `-Dskip.installnodenpm -Dskip.npm`

karussell:
When would we want to skip installing npm?

easbar:
> When would we want to skip installing npm?

I don't know I just saw this and @Nakaner had asked about it. Actually I use it in #2049 where we only want he GH jars as dependencies (and not the frontend stuff). Hm, or maybe we want the frontend stuff there as well at some point :)

Issue #761, opened 2016-08-03
Wrong subnetwork detections&removal leads to a more disconnected graph


karussell:
There is a problem (for motor vehicles only) [here](https://graphhopper.com/maps/?point=-22.230613%2C166.531019&point=-22.229183%2C166.533455) and [here](https://graphhopper.com/maps/?point=-22.229123%2C166.533959&point=-22.229704%2C166.536518), screenshots:

![routing-bug](https://cloud.githubusercontent.com/assets/129644/17360994/71ec9a48-596e-11e6-8df9-3f2a1b92f15d.png)

No connection found although no access restriction nor barrier node was found which could causing this. I can also not reproduce this when downloading the whole island (works as expected) so it could be some subnetwork-removal bug or a data issue which is now already cleaned up. Does someone know how to get the OSM of a certain time? For now closing here, as [the specific road](http://www.openstreetmap.org/way/414570614) also contained a oneway bug which is [now removed](http://www.openstreetmap.org/changeset/41209955) and will very likely fix the problem

karussell:
I can reproduce it now via localhost when I enable elevation. Wow, how strange.

oblonski:
Do you know the reason for this already?

karussell:
It incorrectly detects the part of the road (and connected roads) as oneway-subnetwork and removes it from the graph. But what exactly goes wrong I do not yet know... still debugging ...

karussell:
Ok, can now reproduce it in a unit test and understood what is going on. The long oneway 'Route du Sud' causes several legs on right and left being treated as subnetworks, they would be removed as they are small and all would be fine i.e. the target island which is only reachable through this road would be still reachable. But [those two nodes](https://graphhopper.com/maps/?point=-22.229228%2C166.534747&point=-22.229103%2C166.533932) connect the two right&left subnetworks into one and remove the two edges of the 'Route du Sud' in-between (which leads to surrounding edges being disconnected too and removed until the bridge and towards the 'Rue Algaoué')

michaz:
Do we still think this is a bug? Because I don't.

The two components found by Tarjan's algorithm seem correct to me. It doesn't know what a main road is -- the biggest component will survive, and removing the others shouldn't change the connectedness of the biggest one.

easbar:
Yes I agree with @michaz here. Here is the graph from above unit test:

```java
        //     11-10-9
        //     |     |
        // 0-1-2->3->4->5
        //        |     |
        //        6     12
[...]

PR #2042, opened 2020-05-17
Remove min_one_way_network parameter, clean up subnetwork removal

easbar:
So far we were doing the subnetwork removal 'twice': First we were searching for strongly connected components (but excluding single-node networks) and then we were doing a BFS (breadth-first search) that finds connected components on an undirected graph and also included single node networks. The reason we did not include single-node networks in the first step was performance, but this has been fixed in: e6783acb1760c5ffff07c569244fefc97946f5e1 and is now no longer a problem. The BFS search is not the right thing to do on a directed graph and for example can depend on the order in which we traverse the nodes (see #761). 

So here I have removed the BFS search and the single-node components are now found with Tarjan's algorithm already. This way we also no longer need two different parameters to control the subnetwork search and I have removed `min_one_way_subnetwork_size`.

The subnetwork removal step is now faster and the (by far) most expensive part of it is not finding the subnetworks and setting the appropriate access flags, but the call to `graph.optimize()` afterwards.

With this we also might no longer need #1723 (I had to adjust some tests already)?


michaz:
+1

PR #2044, opened 2020-05-21
Use edge-based subnetwork removal if turn costs are enabled

### LINKED ISSUE #1807: Consider "edge based"-only subnetworks
It is possible to create "edge based"-only subnetworks so we should likely make the subnetwork removal edge-based or rework the mechanism. E.g. [this turn relation](https://www.openstreetmap.org/relation/5882938) disallows the access of cars into the subnetwork from North and [this oneway](https://www.openstreetmap.org/way/31814405) blocks access from West (which is probably a mapping error).

Leading to "Connection between locations not found" - [link](https://graphhopper.com/maps/?point=51.433411%2C7.00939&point=51.433036%2C7.010147&ch.disable=true).

Here is a screenshot of the situation: There is a 'straight-only' turn relation at the junction between Franziskastraße and Gudulastraße which restricts access to Gudulastraße, but at the same time its not accessible from the West/South, because it is a one-way:

image

easbar:
This fixes #1807. The example from #1807 now works as the subnetwork gets removed and the target point snaps to another location:

image


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1334,15 +1335,28 @@ protected void loadOrPrepareLM(boolean closeEarly) {
...
-                + " less nodes");
+        logger.info("nodes: " + Helper.nf(ghStorage.getNodes()) + ", edges: " + Helper.nf(ghStorage.getEdges()));
+    }
+
+    private List<PrepareJob> buildSubnetworkRemovalJobs() {
+        List<FlagEncoder> encoders = encodingManager.fetchEdgeEncoders();
+        List<PrepareJob> jobs = new ArrayList<>();
+        for (FlagEncoder encoder : encoders) {
+            // for encoders with turn costs we do an edge-based subnetwork removal, because they *might* be used with
+            // a profile with turn_costs=true
```
easbar, line 1349: I made this depend on the encoder, so it might happen that we have multiple profiles (some with and some without turn costs) that use the same encoder (the same access flags) and so we also use the edge-based subnetwork removal for the profiles without turn costs. This means we might remove some subnetworks that are not accessible if we include turn costs also for profiles that ignore turn costs anyway. But this does not seem to be a big problem to me. The (more general) alternative would be running the subnetwork removal *per profile*, but then we also need access flags *per profile*. Or maybe no access flags at all but this is another story so I went with the simple solution here. The other extreme would be *always* using the edge-based subnetwork removal (to make things more simple).

core/src/main/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworks.java
```diff
@@ -38,29 +36,31 @@
...
-            accessEncList.add(flagEncoder.getAccessEnc());
-        }
+        this.prepareJobs = prepareJobs;
     }
 
+    /**
+     * The subnetwork removal removes components with less than {@link #minNetworkSize} nodes from the graph if it is
+     * run node-based. For edge-based subnetwork removal it removes components with less than 2*{@link #minNetworkSize}
+     * (directed) edges.
+     */
```
easbar, line 63: We could make this more consistent and also count the number of nodes for edge-based subnetwork removal, but I am not sure if its worth it. For edge-based subnetwork search we only get the edge keys of the components and would have to collect the set of nodes adjacent to these edges to find out how many nodes a component has.

easbar:
Not sure if this should go into `1.0` or not btw

easbar:
It could be useful to report the 125 disconnected components (and also the ones we find for other countries) we find when turn restrictions are included to OSM? Does anybody know the best way to do this? Maybe we should check a few more of them manually.

karussell:
Yeah, this could be useful and e.g. tell the OSM way ID or relation ID that makes a problem? (or if the OSM IDs are not available we could print the edge or PointList as GeoJSON)

In general there are plenty of tagging issues and if we log all of them this would not be nice, but here it looks like we could report them at least to the community and hope they'll improve this.

easbar:
Yes maybe just a list of bounding boxes and maybe sort them by country so people can check if something is wrong in their area or something...

PR #2048, opened 2020-05-28
Add maven profiles to skip tools and shaded web jars

easbar:
Added two maven profiles that make the build a bit more configurable:

```
# build everything (the default, as before)
mvn install

# skip the tools jar with dependencies (which is only needed for `MiniGraphUI` and `Measurement`)
mvn install -Pskip-tools-jar

# skip the shaded all-in-one web jar (which is only needed to run the GH server)
mvn install -Pskip-shaded-web-jar
```

Of course both profiles can be activated at the same time (to skip both jars). In this case the total file size of all jars is reduced from 41MB to 4MB, which should be useful when using GraphHopper as a library.


easbar:
The biggest jar when skipping the two big ones is still the core-tests jar, which is also not needed to use GH as a library (1.4MB). But unfortunately maven is a bit odd here: We can use `-Dmaven.test.skip` to skip building the test jars, but unfortunately it does not understand that in this case we can skip the dependency (with `<scope>test</scope>` *on* these jars) and we get an error. Apparently the only solution would be to create another profile that makes the test jar dependency optional, but I was not sure if this is worth it. Or maybe we can get rid of the test jar altogether :)

PR #2049, opened 2020-05-28
Setup deployment on GitHub packages

easbar:
Every commit is built and the Maven artifacts are deployed to [GitHub packages](https://github.com/graphhopper/graphhopper/packages) (See the 'packages' icon in the 'Code' tab on GitHub) where the version is set to the git commit hash. The 'Actions' tab on GitHub shows the build jobs. The builds do not contain the ready-to-use fat jars for the GraphHopper server and are only meant to be used for development purposes.


.github/workflows/push-packages.yml
```diff
@@ -0,0 +1,24 @@
...
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Cache Maven artifacts
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
```
easbar, line 17: This takes a hash coder from the pom.xml files and uses this has caching key to cache the .m2 folder. The cached files still have to be downloaded on every build so its probably not much faster, but at least this way the logs are not as messy :)

pom.xml
```diff
@@ -264,10 +264,11 @@
     </build>
 
     <distributionManagement>
-        <snapshotRepository>
-            <id>ossrh</id>
-            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
-        </snapshotRepository>
```
easbar, line 270: I do not think we need this as we are not releasing any 'snapshot' artifacts?

.github/workflows/push-packages.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Build and publish package
         run: |
           mvn -B versions:set -DnewVersion=$GITHUB_SHA -DgenerateBackupPoms=false
-          mvn -B -DskipTests -Dskip.npm -Pskip-shaded-web-jar -Pskip-tools-jar deploy
+          mvn -B -DskipTests -Dskip.installnodenpm -Dskip.npm -Pskip-shaded-web-jar -Pskip-tools-jar deploy
```
karussell, line 21: Ah, here we can use it this.

easbar:
Another gotcha: While the packages can be downloaded from github even without signing up using them via maven requires authentication with *any* github account. So you might have to create a personal access token and modify your local `~/.m2/settings.xml` file as described here: https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-apache-maven-for-use-with-github-packages#authenticating-to-github-packages

PR #2059, opened 2020-06-06
Make sure jersey errors are english

### LINKED ISSUE #2058: IsochroneResourceTest fails for non-english locales
**Describe the bug**
`mvn test`
fails on german Linux and MacOS systems.

**To Reproduce**
On Linux 
make sure you have a non-english environment available
```
>locale
LANG="de_DE.UTF-8"
LC_COLLATE="de_DE.UTF-8"
LC_CTYPE="de_DE.UTF-8"
LC_MESSAGES="de_DE.UTF-8"
LC_MONETARY="de_DE.UTF-8"
LC_NUMERIC="de_DE.UTF-8"
LC_TIME="de_DE.UTF-8"
LC_ALL=
```
run
`>mvn test` 

Error message:
```
....
[ERROR] Failures: 
[ERROR]   IsochroneResourceTest.missingPoint:222 {"message":"query param point darf nicht null sein","hints":[{"message":"query param point darf nicht null sein"}]} ==> expected: <true> but was: <false>
[ERROR]   PtRouteResourceTest.testOnePoint:123 expected: <query param point size must be between 2 and 2> but was: <query param point Gr??e muss zwischen 2 und 2 sein>
[ERROR]   SPTResourceTest.missingPoint:128 {"message":"query param point darf nicht null sein","hints":[{"message":"query param point darf nicht null sein"}]} ==> expected: <true> but was: <false>
[INFO] 
[ERROR] Tests run: 143, Failures: 3, Errors: 0, Skipped: 0
```

You can see localized error messages  on any system by faking a "spanish" server:

`java -Duser.language=es -Ddw.graphhopper.datareader.file=andorra.osm.pbf -jar web/target/graphhopper-web-2.0-SNAPSHOT.jar server config-example.yml`

and using this curl
`>curl  localhost:8989/isochrone`
you get this Spanish error message
`{"message":"query param point no debe ser nulo","hints":[{"message":"query param point no debe ser nulo"}]}`

**Expected behavior**
Error massages should be in english by default, mvn test should not fail on non-english systems.

**System Information**
MacOS, AdoptOpenJDK 11 & 8
Debian Testing, openjdk 11

** Fix ** 
Since this error message is only localized because it is produced by jersey, we can set the locale while setting up jersey. See PR

oflebbe:
Fixes #2058

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,7 +184,7 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>2.22.2</version>
                     <configuration>
-                        <argLine>-Xmx130m -Xms130m</argLine>
+                        <argLine>-Xmx130m -Xms130m -Duser.language=en</argLine>
                     </configuration>
                 </plugin>
 
--- a/web-bundle/src/main/java/com/graphhopper/http/GraphHopperBundle.java
+++ b/web-bundle/src/main/java/com/graphhopper/http/GraphHopperBundle.java
@@ -50,6 +50,7 @@ import org.glassfish.hk2.utilities.binding.AbstractBinder;
 
 import javax.inject.Inject;
 import java.util.List;
+import java.util.Locale;
 import java.util.stream.Collectors;
 
 public class GraphHopperBundle implements ConfiguredBundle<GraphHopperBundleConfiguration> {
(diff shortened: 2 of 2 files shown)


karussell:
I would not want to change the Locale globally and instead only change the test setup like it was done here: https://github.com/dropwizard/dropwizard/pull/1056 (?)

oflebbe:
@karussell : Nice Idea to fix the testing this way. However, I am a bit concerned that running graphhopper instance on a (english/spain/german/..) machine will respond with the error messages in the respective language for this particular error (violating a Nonnull constraint ) , while everything else is english. In order to mitigate that we could add -Duser.language=en to the grasshopper start scripts as well. You may have noticed that I added the Locale setting to the Web part only, so every localization in embedded is not affected. What do you think ?

karussell:
For now I would indeed go with the minimal change (that you already applied) as GraphHopperBundle could also be used in another dropwizard application.

PR #2060, opened 2020-06-07
improve correctness for logging config

oflebbe:
* While digging into the logging configuration I found out that log4j jar is only needed for testing: (Dropwizard is using Logback , Tools are already defining it for runtime) 

* Dropwizard 2.0.8 is using slf4j 1.7.30 , therefor sl4j-log4j12 should be 1.7.30 as well (as the comment notes) 

However, no functionality change is detected:
log4j is still included in the shaded web jar (but I do not understand why)
slf4j works for testing, unchanged, as far as I can see.

I proposing to apply the change regardless for the sake of minimal surprise.


easbar:
Thanks! The build failed for a reason not related to this PR. Can you send the license agreement? https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md#license-agreement

oflebbe:
License Agreement already sent to support@ yesterday.

pom.xml
```diff
@@ -120,12 +120,13 @@
                 <groupId>log4j</groupId>
                 <artifactId>log4j</artifactId>
                 <version>1.2.17</version>
+                <scope>test</scope>
```
easbar, line 123: Note that this is the dependency*management* section. We are already declaring the scope at the places we actually use this dependency, so I think we should leave out the scope here. Actually for sf4j-log4j12 we should do the same?

pom.xml
```diff
@@ -120,12 +120,13 @@
...
             </dependency>
             <dependency>
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-log4j12</artifactId>
                 <!-- make sure this matches the slf4j version in the dropwizard bom -->
-                <version>1.7.26</version>
+                <version>1.7.30</version>
```
easbar, line 129: Its a bit unfortunate the dropwizard BOM does not declare this dependency and there is no way we can access the `slf4j.version` property from there right?

michaz:
Why are we using log4j at all in the tests? Can't we just switch them all over to Dropwizard's implementation?

easbar:
Yes using Logback in the tests as well sounds like a good idea. I guess it just was never adjusted after the dw migration?

pom.xml
easbar: ok i did this

PR #2062, opened 2020-06-09
Remove memory mx bean

oflebbe:
While looking into porting to iOS I discovered that the problematic method `getUsedMBAfterGC` isn't used anywhere. Proposing to remove it.

### DIFF
--- a/api/src/main/java/com/graphhopper/util/Helper.java
+++ b/api/src/main/java/com/graphhopper/util/Helper.java
@@ -174,13 +174,6 @@ public class Helper {
         return "totalMB:" + getTotalMB() + ", usedMB:" + getUsedMB();
     }
 
-    public static int getUsedMBAfterGC() {
-        gcAndWait();
-        long result = (ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getUsed() +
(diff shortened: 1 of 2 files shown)


easbar:
Ok its only used in downstream dependencies, but its easy to copy and paste wherever its needed and if its making problems here sure lets remove it.

pom.xml
```diff
@@ -184,7 +184,7 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>2.22.2</version>
                     <configuration>
-                        <argLine>-Xmx130m -Xms130m</argLine>
+                        <argLine>-Xmx130m -Xms130m -Duser.language=en</argLine>
```
easbar, line 187: Sorry I did not follow the language discussion fully, is this related here?

oflebbe: Sorry, forgot to clean history here before submitting. Will rebase the patch.

easbar: Ok but if you are rebasing this we can just as well just merge #2063 :) ?

oflebbe: Right, since it is included

oflebbe:
Please consider  #2063 instead

PR #2063, opened 2020-06-09
Move mx beans2 measurement

oflebbe:
Last one for now :)

Move two problematic functions for porting to iOS to the separate tool where it is used, only.

### DIFF
--- a/api/src/main/java/com/graphhopper/util/Helper.java
+++ b/api/src/main/java/com/graphhopper/util/Helper.java
@@ -174,13 +174,6 @@ public class Helper {
         return "totalMB:" + getTotalMB() + ", usedMB:" + getUsedMB();
     }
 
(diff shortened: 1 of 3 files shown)


easbar:
Ok, this makes #2062 obsolete then?

oflebbe:
Not exactly. It builds up on top of #2062. Since it will git conflicts if I would place independent PR. 
Was testing if #2062 has a chance to get accepted first .

easbar:
> Not exactly. It builds up on top of #2062. Since it will git conflicts if I would place independent PR.
Was testing if #2062 has a chance to get accepted first .

Oh ok but this would only work if there were no other commits anyway, because we are usually squashing all PRs into a single commit to keep a linear history.

easbar:
Thanks!

PR #2065, opened 2020-06-09
Make virtual edge IDs consecutive

easbar:
When we insert a virtual node between two nodes, like nodes 0 and 1 here: 

```
0-x-1
```

four `VirtualEdgeIteratorState`s get created. Two are visible for the edge explorer at the virtual node and one is visible from each of the real nodes (0 and 1). However, these edge states only represent the two directions of the two edges 0-x and x-1 and in fact `edgeState.getEdge()` returns the same number for each pair. Also the virtual edge IDs were not consecutive before this PR. So if 0-1 was the only edge with ID 0 we would get two virtual edge states with ID 1 and two with ID 3. IDs 2 and 4 could still be used to *obtain* the virtual edge states using `graph.getEdgeIteratorState(2,x)` but the returned edge state would still not have edge ID 2 (in this case it would be `edgeState.getEdge()==1` instead).

With this PR the virtual edge IDs will just be 1 and 2 (the same for both directions of the edges just as we do for non-virtual edges as well) and `getEdgeIteratorState(3, x)` will an error.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/querygraph/EdgeChangeBuilder.java
+++ b/core/src/main/java/com/graphhopper/routing/querygraph/EdgeChangeBuilder.java
@@ -66,7 +66,7 @@ class EdgeChangeBuilder {
         //    these adjacent real nodes so we can use them in the next step
         for (int i = 0; i < getNumVirtualNodes(); i++) {
             // base node
-            EdgeIteratorState baseRevEdge = getVirtualEdge(i * 4 + VE_BASE_REV);
+            EdgeIteratorState baseRevEdge = getVirtualEdge(i * 4 + SNAP_BASE);
             int towerNode = baseRevEdge.getAdjNode();
             if (!isVirtualNode(towerNode)) {
                 towerNodesToChange.add(towerNode);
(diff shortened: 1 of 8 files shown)


easbar:
> I feared a problem there will be gaps (but shortcuts come before virtual edges) and also that there can be a problem in QueryOverlay like "colliding" edge IDs, but it seems we never need to store the "raw" virtual edge ID in a map.

What do you mean? So far it was like this: 

```
0,1,2,....|8,9,10,11,12,...,|25,27,29,31
base     |shortcuts           |virtual
```

and now its like this:

```
0,1,2,....|8,9,10,11,12,...,|25,26,27,28
base     |shortcuts           |virtual
```
so if anything we 'closed' the gaps that existed between the virtual edge IDs?

karussell:
> Sorry I am not sure if I can follow. What do you mean? So far it was like this:

(I just had to think twice. I thought it was like `base | virtual |shortcuts` and then if we somewhere use the internal information that the difference has to be "virtual edges"*2 then there would have been a gap. So please just ignore my comment)

> You mean these internal ids of the directed edges?

Yes

easbar:
Ah ok I see :) Thanks for reviewing this!

PR #2069, opened 2020-06-23
Add new query graph for CH routing

### LINKED ISSUE #1780: Decouple CHGraph from (Base)Graph functionality

easbar:
So far we built a `QueryGraph` on top of `CHGraph`, which in turn had to implement the `Graph` interface. In this PR I removed the `Graph` interface from `CHGraph` and added a new `QueryRoutingCHGraph` which serves the same purpose for a `RoutingCHGraph` as `QueryGraph` does for the base graph. 

With this change it should now be possible to improve/optimize the way `RoutingCHGraph` stores (CH) edges independently from base graph. For example we can now 'disconnect' also base graph edges which should speed up both CH queries and preparation. Also we no longer need to 'know' that we have to wrap a `CHGraph` into a `QueryGraph` before passing it to the routing algorithm factory.

Fixes #1780 

Some potential follow-ups:

* [ ] rename CHGraph -> Old/Internal/CHGraph, rename RoutingCHGraph -> CHGraph, rename QueryRoutingCHGraph -> QueryCHGraph (or CHQueryGraph :))
* [ ] more and more remove CHGraph, eventually RoutingCHGraph should be what we store on disk
* [ ] fully decouple (base/virtual)edge states from CH, i.e. remove CHEdgeIteratorState interface from base EdgeIteratorState etc.


easbar:
There seems to be a bit of a slow down (about 10-20%) for CH alternatives, will try to figure out what this is.

PR #2071, opened 2020-06-23
Merge of navigation repository

karussell:
Due to the maintenance and development overhead here is the PR to reduce this via merging the navigation repository. Ping @boldtrn @easbar.

Related to #1422.

It seems I was able to successfully merge the master branch of https://github.com/graphhopper/graphhopper-navigation into this repository :). At least here and on command line it looks fine:

image

Before I did this I moved everything into the subfolder "navigation" to avoid conflicts when merging. I did not use any special tricks or subtrees, see [this SO answer](https://stackoverflow.com/a/10548919/194609).

In more details I did the following in the graphhopper-navigation repo:

```bash
git checkout -b main
mkdir navigation
mv * navigation/
git commit -a -S -m "move into navigation subfolder"
rm -rf .github/ .gitignore .travis.yml
git commit -a -S -m "remove duplicate files"
git push --set-upstream origin main
```

In the graphhopper repo I did:

```bash
git checkout master -b gh_navigation_merge
# only fetch the remote "main" branch and no tags
git remote add --no-tags -f -t main gh_navigation git@github.com:graphhopper/graphhopper-navigation.git
git merge gh_navigation/main
git merge --allow-unrelated-histories gh_navigation/main
```

From the Java side nothing is merged yet. Probably we should do this in a second PR to squash commits for this but keep the commits here? (Or we can also do it here and squash + push force)

We could remove navigation/files/andorra.osm.gz before to avoid more overhead for the main branch. This is probably only possible if we would alter history.

Work to be done:

 * remove files like navigation/files/, the license, pom.xml etc
 * merge translation keys and avoid the second TranslationMap
 * keep the navigation feature separated and also enable this only via a config switch so that everyone can at any remove those dependencies


boldtrn:
Thanks a lot for creating this PR @karussell, I think it's a great idea to combine the two repositories as this reduces maintenance overhead dramatically.

Do you think we have to keep the history of the graphhopper-navigation repository? Wouldn't it be easier to just copy the .java files and add a note in the README, that the history before merging can be looked up in the old repo and we archive the old repo? I think would make it a lot easier to avoid blowing up the repo with the andorra.pbf.

easbar:
Cool thanks! I cannot judge whether or not the history is useful (personally the only time I might use it is to build older commits, but this can be done as long as we keep the other repository, we do not need all the commits in the core repository for this). Keeping the core repository size small wherever possible is a good idea in my opinion.

> From the Java side nothing is merged yet. Probably we should do this in a second PR to squash commits for this but keep the commits here? (Or we can also do it here and squash + push force)

Either way is fine for me.

> keep the navigation feature separated and also enable this only via a config switch so that everyone can at any remove those dependencies

So we should not add the navigation resource to the default web bundle. People who use it would have to modify the bundle setup, but this was required so far anyway?.

easbar:
Even if we do not merge the history here it's still good to know how this can be done for the map-matching repository :)

boldtrn:
> So we should not add the navigation resource to the default web bundle. People who use it would have to modify the bundle setup, but this was required so far anyway?.

The way I understand it is that we add a key to the config, to enable/disable it.

easbar:
> The way I understand it is that we add a key to the config, to enable/disable it.

Ok, but then it still has to be available at compile time doesn't it? Would this be a problem? I thought if we kept it out of the web-bundle the navigation stuff could stay in an entirely separate module (within the same repository). But no real opinion here tbh...

michaz:
Okay, my opinion is always that there _should not be combinatorial builds_, i.e. `mvn package` makes the repository. If there's going to be a Maven profile aka feature switch for navigation, then ... I don't know. I mean, why? If it belongs to GraphHopper, it should be built with GraphHopper, and if it doesn't, then it should be in its own repository.

Why can't we just compile everything?

michaz:
More generally: We don't have a full plug-in architecture like OSGi and probably don't want one. So let's try not to emulate one.

Let's just put it into the web server like all the other resources. And do it the other way round: If someone, for some reason, really does _not_ want it in their web server, they should edit the code / roll their own.

For me, a dropwizard program is the outermost layer. If you want to reconfigure it, you must edit it. The one we ship is an _example_ that people can edit.

(We do it the same way, we also have our own dropwizard program for production.)

michaz:
Also, what dependencies? Navigation doesn't seem to have any dependencies.

karussell:
> Let's just put it into the web server like all the other resources so people can test it.

Ok, convinced will do without config just like we have isochrone and for people to try.

Only unsure about whether to keep history or increase repo size :) ... probably here indeed removing the PBF and squashing it make sense, but then unfortunately all the good work (and history) of @boldtrn gets lost :/ - would this be okay @boldtrn? 

> Also, what dependencies? Navigation doesn't seem to have any dependencies.

Ups, indeed.

boldtrn:
I would be fine with loosing the history (it' not lost, just in a different repo :)). I can clearly see the benefit of merging the two repos and to reduce the repo size.

michaz:
Okay cool. But just to pre-register: For map-matching, I want the history. :)

karussell:
Ok, will now squash and merge. And in a second PR integrate the navigate endpoint and adapt the remaining stuff so that the build includes navigation module and the web service the endpoint etc.

> For map-matching, I want the history. :)

Ok, sure :). This was still a good test here. I think for map matching it would work identical except that we won't squash.

karussell:
The merge initially failed with a cryptic message I did not see before and offered a "try again" button. I clicked that and now it is merged but it seems to have merged it without squashing before. Not sure what to do now :(

easbar:
> Not sure if we can fix something now as I would find a git rebasing on master wrong. Really ugly.

In this particular case it might be worth considering a hard reset?

karussell:
Would you do this? The last time I did this I messed up something and swore not to do this again ;) ... I think we also need to lift the master branch protection (?)

easbar:
> Would you do this? The last time I did this I messed up something, but not sure how and what exactly I only know that it was not worth it ... I think we also need to lift the master branch protection (?)

Hm we just need to reset the master branch to the commit before the merge and then force push it to upstream. And yes to force push we need to lift the protection on github. But if you do not think its worth it we do not have to do it. As far as I can tell it would be ugly for everyone using the updated master already (merged it to some feature branch, use it in some fork etc.), but it should be harmless for everyone who knows we did this and/or did not even fetch since you did the merge. No idea how many people would be affected tbh. 

What is our concern here? I think having the non-linear history is quite ugly and probably this also increased the repository size (but don't know by how much).

karussell:
Ok. Have temporarily allowed force pushes.

easbar:
> Ok. Have temporarily allowed force pushes.

Ok I did it :see_no_evil: I kept the previous branch as `master_old`. For example the `nav_integration` branch is based off of this one. But once we squash and merge `nav_integration` into master we should get rid of all the nav commits.

Just make sure to update your master properly now.

karussell:
Thanks! Updated my master already.

PR #2073, opened 2020-06-28
Provide edge key PathDetails

michaz:
Because edge key is at least as interesting as edge id.


michaz:
> the only downside here but also for edge_id is that we still have no ID stability

True! But my customer has control over the import, and deliberately creates edges in a certain order. So we know which edge is which.

Issue #1940, opened 2020-02-28
Future of 'Offline' Routing


karussell:
The current state of routing on mobile devices (iOS & Android) without internet access (aka "offline routing") is suboptimal:

* The Android support is "okayish", but for a meaningful usage we could need [graph merging](https://github.com/graphhopper/graphhopper/issues/293) or a tiled approach to allow cross-graph routes and a fix for [a strange, per-process virtual memory limit](https://discuss.graphhopper.com/t/spatial-limitation-in-android-offline-on-device/298/13?u=karussell) to allow loading big graphs too e.g. a native DataAccess or file cache approach. There is also still no CI integration for Android, see #803.
* Since several releases there is no update of our [iOS port](https://github.com/graphhopper/graphhopper-ios/). Here all the limitations of the [j2objc library](https://github.com/google/j2objc) apply, but it worked surprisingly well and even supports several Java 11 features and without any artificial virtual memory limitation. 
  **Update**: see this new issue: https://github.com/graphhopper/graphhopper-ios/issues/46
* For both platforms we do not have the public transit module as only the core is limited to Java 7.

I am really happy that despite these limitations I know there are apps in the app stores that use GraphHopper for offline routing (Android and iOS).

And still, we are struggling to continue supporting this as we have to limit ourselves and avoid updating dependencies (see e.g. #1803), stick to Java 7 in the core and also many dependency issues when it comes to Jackson and its annotations and many more design decisions that would be less complicates without "Android Java" in mind. Also there will be [inline types in Java](http://openjdk.java.net/jeps/169), which are already available as experimental feature in JDK 14, which it will be hard to avoid in the core. Previously we have accepted this and I've done lots of testing for Android myself and invested time or provided the map data and apks, but this does not make much sense as I have no real Android- or "app store"-experience. Additionally it is effort that I'm no longer willing to invest.

I see the following options:

 1. we abandon offline routing and embrace the latest server side Java version in GraphHopper
 2. we make it worth to limit ourselves in the core (i.e. implement the missing "production" features and update the iOS port and have real community involvement for this like we already have for other things)
[...]

PR #2077, opened 2020-07-06
Update core,client-hc and reader-osm to Java 8

easbar:
Now that it has been decided that GraphHopper only fully supports Android up to GraphHopper version 1.0 we can update all modules to Java 8. Related: #1940 (I.e. GraphHopper 2.0 drops support for very old Android versions for offline routing)

### DIFF
--- a/client-hc/pom.xml
+++ b/client-hc/pom.xml
@@ -54,19 +54,4 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <configuration>
-                    <compilerArgument>-XDignore.symbol.file</compilerArgument>
-                    <source>1.7</source>
-                    <target>1.7</target>
-                </configuration>
-            </plugin>
-            
-        </plugins>        
-    </build>
-        
 </project>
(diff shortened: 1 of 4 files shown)


client-hc/pom.xml
```diff
@@ -54,19 +54,4 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <configuration>
-                    <compilerArgument>-XDignore.symbol.file</compilerArgument>
```
easbar, line 63: Does anyone know what this `ignore.symbol.file` was meant to do and if it is still needed?

easbar:
Travis build (but not the one for jdk 8) fails with 

```
$ wget https://raw.githubusercontent.com/sormuras/bach/master/install-jdk.sh
$ . ./install-jdk.sh -F 14 -C
install-jdk.sh 2020-06-02
curl: (52) Empty reply from server
```

client-hc/pom.xml
otbutz: Looks like the intention was access to internal JRE classes.

otbutz:
A bit offtopic but we could switch to OpenJDK and retrieve the JDKs from https://adoptopenjdk.net/

karussell:
> A bit offtopic but we could switch to OpenJDK and retrieve the JDKs from https://adoptopenjdk.net/

The install script already fetches the files for OpenJDK:

> install-jdk.sh main purpose is to install the latest-and-greatest available OpenJDK release from jdk.java.net.

karussell:
@otbutz have changed the download servers to adoptopenjdk as java.net has some problems it seems.

PR #2082, opened 2020-07-12
withInstructions should be without path details

karussell:
@easbar do you know why we include the `average_speed` path details for instructions in Measurement?

### DIFF
--- a/tools/src/main/java/com/graphhopper/tools/Measurement.java
+++ b/tools/src/main/java/com/graphhopper/tools/Measurement.java
@@ -260,12 +260,12 @@ public class Measurement {
                     printTimeOfRouteQuery(hopper, new QuerySettings("routingCH_no_instr", count, isCH, isLM).
                             sod());
                     printTimeOfRouteQuery(hopper, new QuerySettings("routingCH_full", count, isCH, isLM).
-                            withInstructions().withPointHints().sod().simplify());
+                            withInstructions().withPointHints().sod().simplifyAndPD());
                     // for some strange (jvm optimizations) reason adding these measurements reduced the measured time for routingCH_full... see #2056
                     printTimeOfRouteQuery(hopper, new QuerySettings("routingCH_via_100", count / 100, isCH, isLM).
                             withPoints(100).sod());
                     printTimeOfRouteQuery(hopper, new QuerySettings("routingCH_via_100_full", count / 100, isCH, isLM).
-                            withPoints(100).sod().withInstructions().simplify());
+                            withPoints(100).sod().withInstructions().simplifyAndPD());
                 }
                 if (!hopper.getCHPreparationHandler().getEdgeBasedCHConfigs().isEmpty()) {
                     printTimeOfRouteQuery(hopper, new QuerySettings("routingCH_edge", count, isCH, isLM).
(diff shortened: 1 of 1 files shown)


easbar:
Not sure, but one thing I remember is that for #1782 some path details had to be included to reproduce the performance bug.

karussell:
Yes. For pathdetails&simplify this is fine. But I would have at least one "clean" variant that only adds instructions (and so I removed the path details for instructions=true)

easbar:
Ok sure.

tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -781,10 +784,7 @@ public int doCalc(boolean warmup, int run) {
                 if (querySettings.alternative)
                     req.setAlgorithm(ALT_ROUTE);
 
-                if (querySettings.withInstructions)
-                    req.setPathDetails(Arrays.asList(Parameters.Details.AVERAGE_SPEED));
```
easbar, line 785: This indeed looks strange. We are enabling instructions by enabling path details?

karussell:
Let's try if I break something ;)

tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -781,10 +784,7 @@ public int doCalc(boolean warmup, int run) {
...
-
-                if (querySettings.simplify) {
+                if (querySettings.simplifyAndPD) {
                     req.setPathDetails(Arrays.asList(Parameters.Details.AVERAGE_SPEED, Parameters.Details.EDGE_ID, Parameters.Details.STREET_NAME));
```
easbar, line 788: Here we enable simplify by adding path details? Do we still have a measurement with instructions and path details?

tools/src/main/java/com/graphhopper/tools/Measurement.java
karussell: We enabled two things, so we could have also renamed withInstructions into withInstructionsAndPD (?)

tools/src/main/java/com/graphhopper/tools/Measurement.java
karussell: I think "simplification" is always done for "forServer" (used here in Measurement, see line 182), except if `!simplify`, then WAY_POINT_MAX_DISTANCE=0

easbar: Aha :) It seems confusing that `setPathDetails` has anything to do with instructions?

easbar: Do we really need `forDesktop/Server/Mobile`? It happened to me at least twice that I wanted to disable/enable path simplification and then I found myself calling `forServer` or `forDesktop` to achieve this 'indirectly'...

karussell: > Aha :) It seems confusing that setPathDetails has anything to do with instructions?

Will introduce a separate pd=true/false variable

> Do we really need forDesktop/Server/Mobile? It happened to me at least twice that I wanted to disable/enable path simplification and then I found myself calling forServer or forDesktop to achieve this 'indirectly'...

You mean exposing setSimplifyResponse and removing forXY?

karussell: > It happened to me at least twice that I wanted to disable/enable path simplification and then I found myself calling forServer or forDesktop to achieve this 'indirectly'...

Hmmh, looking into this ... and "forServer" (i.e. with simplification) should be the default. Also you could call `hopper.getRouterConfig().setSimplifyResponse` to change this.

easbar: > You mean exposing setSimplifyResponse and removing forXY?

Yes if thats possible? Or was it meant similar to `ch.disablingAllowed` (make sure users cannot disable path simplification)?

> Hmmh, looking into this ... and "forServer" (i.e. with simplification) should be the default. Also you could call hopper.getRouterConfig().setSimplifyResponse to change this.

Ah ok if we can change simplify via the router config thats good (I thought the last time I wanted to do this it was not possible).

karussell: > Yes if thats possible? Or was it meant similar to ch.disablingAllowed (make sure users cannot disable path simplification)?

It is meant as the default and for desktop it doesn't make much sense as we can transmit the full and original path, but client bandwidth or client power can be limited and it is usually faster to compress the response even if it takes time. E.g. komoot returns a custom format probably unsimplified resulting in ~5MB (just picked a 200km route) where our route returns quickly and only needs <100KB :) (but probably this is more due to our compact path details representation and not so much due to the simplification. maybe interesting improvement @jtheuer ;) )

PR #2087, opened 2020-07-20
Clean-up edge iterator classes

easbar:


core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -112,23 +112,6 @@ public BaseGraph(Directory dir, final EncodingManager encodingManager, boolean w
         this.edges = dir.find("edges", DAType.getPreferredInt(dir.getDefaultType()));
         this.listener = listener;
         this.edgeAccess = new EdgeAccess(edges) {
-            @Override
-            EdgeIteratorState getEdgeProps(int edgeId, int adjNode, EdgeFilter edgeFilter) {
```
easbar, line 116: This method was not really related to EdgeAccess.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -1046,12 +1033,12 @@ private long nextGeoRef(int arrayLength) {
         return tmp;
     }
 
-    protected static class EdgeIterable extends CommonEdgeIterator implements EdgeExplorer, EdgeIterator {
+    protected static class EdgeIteratorImpl extends EdgeIteratorStateImpl implements EdgeExplorer, EdgeIterator {
         final EdgeFilter filter;
         int nextEdgeId;
 
-        public EdgeIterable(BaseGraph baseGraph, EdgeAccess edgeAccess, EdgeFilter filter) {
-            super(-1, edgeAccess, baseGraph);
```
easbar, line 1054: This `-1` appeared frequently, and was the only value used for this parameter

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -1046,12 +1033,12 @@ private long nextGeoRef(int arrayLength) {
         return tmp;
     }
 
-    protected static class EdgeIterable extends CommonEdgeIterator implements EdgeExplorer, EdgeIterator {
+    protected static class EdgeIteratorImpl extends EdgeIteratorStateImpl implements EdgeExplorer, EdgeIterator {
```
easbar, line 1036: I renamed this for consistency with the other iterator class names.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -1062,33 +1049,6 @@ final void setEdgeId(int edgeId) {
             this.nextEdgeId = this.edgeId = edgeId;
         }
 
-        /**
-         * @return false if the edge has not a node equal to expectedAdjNode
-         */
-        final boolean init(int tmpEdgeId, int expectedAdjNode) {
```
easbar, line 1068: Moved this one level up, its hardly related to the iterator but its more like the initialization code of an edge state

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -1201,12 +1154,9 @@ public final EdgeIteratorState detach(boolean reverseArg) {
         }
     }
 
-    /**
-     * Common private super class for AllEdgesIteratorImpl and EdgeIterable
-     */
-    static abstract class CommonEdgeIterator implements EdgeIteratorState {
+    static class EdgeIteratorStateImpl implements EdgeIteratorState {
```
easbar, line 1157: Renamed this because its more like an 'edge' than an iterator. Its used as base class for the iterators yes, but there wasn't much that was 'iterator-like' about this class.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -1215,15 +1165,39 @@ public final EdgeIteratorState detach(boolean reverseArg) {
         boolean freshFlags;
         int edgeId = -1;
         private final IntsRef edgeFlags;
-        int chFlags;
```
easbar, line 1218: The chFlags shouldn't belong to this (basegraph) iterator, I guess this was forgotten in some previous refactoring

core/src/main/java/com/graphhopper/storage/BaseGraph.java
karussell: Thanks! I think it is probably time to try again if we can separate the EdgeIterator interface from EdgeIteratorState (the last time I tried this it was still to heavily coupled in CH but maybe now this is possible?)

core/src/main/java/com/graphhopper/storage/BaseGraph.java
karussell: Nice!

core/src/main/java/com/graphhopper/storage/BaseGraph.java
easbar: Ok, but why would this be better? An EdgeIterator is just an EdgeIteratorState with a `next()` method. I don't think this ever bothered me.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -1130,29 +1090,22 @@ void goToNext() {
...
-
-            EdgeIteratorState iter = edgeAccess.getEdgeProps(edgeId, reverseArg ? baseNode : adjNode, filter);
-            assert iter != null;
-            if (reverseArg) {
-                // for #162
-                ((EdgeIterable) iter).reverse = !reverse;
-            }
-            return iter;
+            if (edgeId == nextEdgeId)
+                throw new IllegalStateException("call next before detaching (edgeId:" + edgeId + " vs. next " + nextEdgeId + ")");
```
karussell, line 1094: BTW: I'm not sure if it is a good thing to do these logic checks via `if` instead of `assert' as they could have a potential slow down, but I wasn't able to measure this difference and so I preferred long term safety (i.e. show these problems even in production)

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -1164,10 +1117,10 @@ public int length() {
         public boolean next() {
             while (true) {
                 edgeId++;
-                edgePointer = (long) edgeId * edgeAccess.getEntryBytes();
                 if (edgeId >= baseGraph.edgeCount)
                     return false;
 
+                edgePointer = edgeAccess.toPointer(edgeId);
```
karussell, line 1123: Should we not update also the edgePointer to the new value before returning?

easbar: Hm, if `false` is returned we are not using the iterator anymore? But maybe you are right and its more consistent the way you proposed. Should not matter performance-wise, because usually it has to be updated anyway (unless we are at the end of the iteration)

easbar: But then we need to update adjNode and baseNode etc. as well?

core/src/main/java/com/graphhopper/storage/BaseGraph.java
karussell: Hmmh, my argument would be that they should be separated like Object is separate from Iterator, but yes, here we change the properties of "Object" with every `iterator.next()` call and so `EdgeIterator` is not really an Iterator it is more an EdgeIteratorState with a `setPointer/setCursor` method (very similar to the [ResultSet](https://docs.oracle.com/en/java/javase/11/docs/api/java.sql/java/sql/ResultSet.html) from jdbc API)

core/src/main/java/com/graphhopper/storage/BaseGraph.java
karussell: Ah, indeed, you are right.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
easbar: Ok for me an `EdgeIteratorState` is a proxy to a row of data, or in some cases its an actual row of data (after calling `detach()`). An `EdgeIterator` is the same thing except that it can be moved to another edge. I think @michaz suggested calling it a 'cursor' because it changes state after calling `next()`, compared to the `java.util.Iterator` which *returns* the next element when calling `next()`. But honestly, I am so used to it I do not really care whether its called iterator or cursor. But I do not see real value in separating `EdgeIteratorState` from `EdgeIterator`, because like I said in this case the latter is really the same thing as the first except that it allows to be shifted (using the `next()` method).

Now we are at it how would we call an `EdgeIteratorState` that can be set to an arbitrary edge (not just move to the next edge at the same base node). Would it be a `SingleEdgeExplorer` or a `SingleEdgeIterator` or a `SingleEdgeCursor`? :smiley:

core/src/main/java/com/graphhopper/storage/BaseGraph.java
easbar: Hm yes good question. An integer comparison `==` should hardly ever matter? Except the method gets too long and is no longer inlined or something? Very hard or impossible to say without measuring.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
karussell: Probably the EdgeIterator is not a "cursor" because it has no direct setEdgeId method, only an implicit "next()" and so "iterator" is more appropriate (?)

> Now we are at it how would we call an EdgeIteratorState that can be set to an arbitrary edge (not just move to the next edge at the same base node).

I like SingleEdgeCursor (also it would have an explicit setEdgeId :) ? )

easbar: Ok yes `SingleEdgeCursor` is good and avoids the confusion of calling it "iterator" or "explorer" :) Yes its main method would be `setEdgeId` (or `setEdge` or `moveToEdge(Id)` or `setToEdge(Id)?).

easbar: > Probably the EdgeIterator is not a "cursor" because it has no direct setEdgeId method, only an implicit "next()" and so "iterator" is more appropriate (?)

Hppc does this:

```java
      for (IntCursor cursor : IntArrayList.from(3, 6, 5)) {
            System.out.println(cursor.value);
      }
```

and here we get the `IntCursor` when calling next on the `Iterator<IntCursor>`, so the `IntCursor` here is the equivalent to our `EdgeIteratorState` (the proxy object that gives access to the index and value of an int array for example).

michaz: Ohhhh, yes yes yes yes yes, standard iteration syntax, please. :-)

PR #2089, opened 2020-07-21
Added the possibility to remove missing StopTimes from GTFSFeed

gcheval:


michaz:
I... don't think this is right? I don't know what happens with stop times that are missing (I think they are interpolated). But if they aren't -- just removing the stop from the line run doesn't seem right.

Also I'd prefer not to have pre-processing in here that could just as well be done externally -- I take the GTFS verbatim, and leave data cleaning / normalization / adaptation to missing features to outside code.

gcheval:
@michaz The motivation behind this is that we have seen multiple operators remove **arrival_time** and **departure_time** when a stop is supposedly not in use. Would you see a way to handle those situations in the GH importing code to avoid a RuntimeException?

michaz:
Hmm, do you have a real-life example? In a public GTFS file?

gcheval:
Sure. Here are some I found:

http://data.trilliumtransit.com/gtfs/addisoncounty-vt-us/addisoncounty-vt-us.zip
http://data.trilliumtransit.com/gtfs/ccta-vt-us/ccta-vt-us.zip
http://whereismybus.advancetransit.com/GTFS/feed.zip

It would be great to be able to mitigate this since per Google's standard, those fields can be left empty (https://developers.google.com/transit/gtfs/reference#stop_timestxt).

michaz:
Thanks, will check.

michaz:
Ah! Yes, but...

```
t_315808_b_16374_tn_10,12:45:00,12:45:00,2394625,9,,0,0,5119.6002328,0,,,804,804,3,3,,,,
t_315808_b_16374_tn_11,13:00:00,13:00:00,2394625,1,,0,0,0,0,,,804,804,3,3,,,,
t_315808_b_16374_tn_11,,,771309,2,,2,3,769.711262941777,0,,,804,804,3,3,,,,
t_315808_b_16374_tn_11,,,771320,3,,2,3,1955.31007135351,0,,,804,804,3,3,,,,
t_315808_b_16374_tn_11,13:04:00,13:04:00,771344,4,,0,0,2254.29823408039,0,,,804,804,3,3,,,,
t_315808_b_16374_tn_11,,,2394636,5,,0,0,2310.25922616439,0,,,804,804,3,3,,,,
t_315808_b_16374_tn_11,,,771378,6,,2,3,2964.72103956919,0,,,804,804,3,3,,,,
t_315808_b_16374_tn_11,,,771372,7,,0,0,4320.58013401871,0,,,804,804,3,3,,,,
t_315808_b_16374_tn_11,,,2394642,8,,0,0,4759.44516215004,0,,,804,804,3,3,,,,
t_315808_b_16374_tn_11,13:15:00,13:15:00,2394625,9,,0,0,5119.6002328,0,,,804,804,3,3,,,,
```

... what makes you think this means the stops aren't in use?

I'm not doubting we are missing a feature here, or that throwing an Exception is okay, it's not. But I think this means that *there are no scheduled times for these stops* and the bus will just serve them as it comes along.

gcheval:
Do you see a way we could map those stops? Routing wise, it's hard to know when those stops will be reached.

gcheval:
Maybe average a time given the distance traveled? between each stops 🤔

michaz:
Yeah, and the code looks like someone started something like this and then stopped; or it was once there and someone removed it.

Can't fully rule out it was me.

michaz:
It was me.

What I'd recommend at the moment is this:

Use Conveyal's [GTFSlib](https://github.com/conveyal/gtfs-lib), or some other tool, to patch the missing times. Before import.

I'll open a ticket.

gcheval:
Great. Thanks!

michaz:
Thank you, too. I'm working on something else at the moment, but if you'd like to help, you could just try to paste the code that I apparently removed from GTFSFeed.java (compare with the version in GTFSlib) back in and see what happens.

I really can't remember why I did that exactly. Maybe the reason has gone away.

michaz:
I'd be happy to almost immediately merge that if it works. :-)

gcheval:
I'll take a look 👍

PR #2090, opened 2020-07-21
Set Unit public

boldtrn:
This PR addresses a minor issue. If I would like to use the method `NavigateResponseConverter#convertFromGHResponse(GHResponse ghResponse, TranslationMap translationMap, Locale locale, DistanceConfig distanceConfig)`, I need to pass a DistanceConfig. The constructor of the DistanceConfig is: `DistanceConfig(DistanceUtils.Unit unit, TranslationMap translationMap, Locale locale)`. `DistanceUtils.Unit` has only package access, so I cannot use it in a different package, therefore I cannot use the `NavigateResponseConverter` from a different package.


boldtrn:
Thanks for the fast merge :)

PR #2091, opened 2020-07-21
Improve tag extraction

otbutz:
- [x] support for parsing short ton weight limits
- [x] move speed limit parsing into OSMValueExtractor
- [x] utility to check parsers using JSON data provided by https://taginfo.openstreetmap.org/


core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -222,20 +223,35 @@ public double getMaxSpeed() {
...
+        return !Double.isNaN(speed) && speed >= 0;
+    }
+
+    /**
+     * @return <i>true</i> if the given speed is not {@link Double#NaN} and is
+     *         bigger than zero
+     */
+    protected boolean isPositiveSpeed(double speed) {
+        return !Double.isNaN(speed) && speed > 0;
+    }
```
karussell, line 254: why is this necessary?

otbutz: Yeah i was thinking about the same. Wouldn't it be better to ignore 0 speeds completely?

https://overpass-turbo.eu/s/WwG

Most of those ways have an accompanying access tags which already prohibit routing.

otbutz:
We should probably add support for parsing non-numerical maxheight values: https://wiki.openstreetmap.org/wiki/Key:maxheight#Non-numerical_values

Currently we miss out on a lot of valuable information:

* default: 18545
* none: 12264
* below_default: 3045
* unsigned: 136
* no_indications: 84

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
karussell: Yes, your overpass query is convincing, seems to be a good idea to exclude those ways and assume this is a tagging problem. Or is it used for other purposes (e.g. for other speed tags) too?

otbutz: I couldn't find any reasonable uses of speed 0 for `maxspeed:forward`, `maxspeed:backward` and `maxspeed:motorcycle`. It should be safe to ignore it as a tagging error.

Related: https://github.com/graphhopper/graphhopper/issues/367

otbutz:
IMHO it's better to ignore `default` and its equivalents and handle non existing limits later using spatial rules as we're able to properly differentiate between countries.

But i've the extended the extractor to deal with maxheight `below_default` returning a rather conservative height limit of 2m which should be enough for regular cars to pass and while blocking HGV traffic.

core/src/main/java/com/graphhopper/routing/ev/MaxSpeed.java
```diff
@@ -33,6 +33,10 @@
      * The speed value used for road sections without known speed limit.
      */
     public static final double UNSET_SPEED = Double.POSITIVE_INFINITY;
+    /**
+     * The lowest speed value which can be stored.
+     */
+    public static final double MINIMUM_STOREABLE_SPEED = 5;
```
karussell, line 39: This is not that easy and dependent on how the EncodedValue for the speed is configured. So we can easily store 0.1km/h with enough bits...

core/src/main/java/com/graphhopper/routing/util/parsers/helpers/OSMValueExtractor.java
```diff
@@ -80,6 +86,12 @@ public static double stringToMeter(String value) {
         value = FEET_PATTERN.matcher(value).replaceAll("ft");
         if (isInvalidValue(value))
             return Double.NaN;
+        
+        // only used for maxheight
+        if ("below_default".equals(value)) {
+            return 2; // assume that cars are able to pass
+        }
```
karussell, line 93: I would not do this and instead ignore all these soft tags. IMO tagging a height with such a value is a tagging issue and we should not use it or even convert it into an explicit max height.

otbutz: `below_default` is an accepted and documented value. Projects like [StreetComplete](https://github.com/westnordost/StreetComplete) make use of it: https://github.com/westnordost/StreetComplete/blob/ed6ab1426f5c26dc4779c2c84118483a4a84a6eb/app/src/main/java/de/westnordost/streetcomplete/quests/max_height/AddMaxHeight.kt#L68-L77

core/src/main/java/com/graphhopper/routing/ev/MaxSpeed.java
otbutz: I'm not sure how (or if?) we should deal with unstoreable values. If we know that we can't store a certain value, we could use alternative tags or fallback values.

core/src/main/java/com/graphhopper/routing/util/parsers/helpers/OSMValueExtractor.java
karussell: Ok, yes. I can see it in the wiki. It is not really useful IMO and seems to be mostly used by German mappers. My problem with this approach is for GraphHopper itself: if we would do this then some limits have not the meaning they have in real world. So I would keep this for another issue.

core/src/main/java/com/graphhopper/routing/ev/MaxSpeed.java
karussell: The tag extraction methods shouldn't care about it and always return precise numbers. Instead this issue can be handled in OSMMaxSpeedParser dependent on the configuration of the EncodedValue. As this is an important issue it is worth a separate issue.

karussell:
Sorry, I completely missed this ... will merge now.

karussell (approved):

Issue #2086, opened 2020-07-16
SPT with finite, non-zero turn costs much slower


karussell:
For routing with turn restrictions enabled we have a slow down of approx. 2x and so I would expect a maximum slow down of 2x (because the shortest path exploration is usually the fast part of the isochrone calculation). But it seems that the slow down factor is over 6x even for smaller time limits like 10min:

```
time curl 'localhost:8989/isochrone?point=52.550732,13.350878&profile=car&time_limit=600'
...
real	0m1,179s
user	0m0,000s
sys	0m0,018s

time curl 'localhost:8989/isochrone?point=52.550732,13.350878&profile=car_no_tc&time_limit=600'
...
real	0m7,464s
user	0m0,010s
sys	0m0,004s
```

Related to #1978 and #1458.

karussell:
Strange. I cannot reproduce this on my local machine (4.5s vs. 2.8s). Closing for now.

easbar:
Keep in mind that if turn costs are enabled the graph exploration is edge-based and so the shortest path tree is larger (contains edges that otherwise would be skipped as they point to already explored nodes). Maybe this could be an explanation? 

See this: https://discuss.graphhopper.com/t/missing-edges-in-spt-response/5586/4

otbutz:
You could profile the execution using JFR if you're using Java 11 or higher.

JMC: https://adoptopenjdk.net/jmc.html

easbar:
> the resulting shape is only slightly different

Yeah but the shortest path trees could still be very different in size? In your example the number of visited nodes is very different (486191 for car vs 170438 for car_no_tc) so to me it seems like the real question is why in one case much less nodes are explored (could be due to better/worse LM heuristic?). Maybe MiniGraphUI could help here :)

michaz:
Hmm. Seems elusive. Works fine on my machine.

```
2020-07-22 10:44:11.757 [dw-21 - GET /isochrone?point=48.839413,2.351074&profile=car_no_tc&time_limit=400] INFO  c.g.resources.IsochroneResource - took: 1.1765695, visited nodes:170247, consumed nodes:21612, {time_limit=[400], profile=[car_no_tc], point=[48.839413,2.351074]}
[...]

PR #2092, opened 2020-07-22
Work around slow PriorityQueue.remove

michaz:
See #2086.

### DIFF
--- a/isochrone/src/main/java/com/graphhopper/isochrone/algorithm/ShortestPathTree.java
+++ b/isochrone/src/main/java/com/graphhopper/isochrone/algorithm/ShortestPathTree.java
@@ -44,7 +44,9 @@ import static java.util.Comparator.comparingLong;
  *
  * IMPLEMENTATION NOTE:
  * util.PriorityQueue doesn't support efficient removes. We work around this by giving the labels
- * a deleted flag, not remove()ing them, k
+ * a deleted flag, not remove()ing them, and popping deleted elements off both queues.
+ * Not to self/others: If you think this optimization is not needed, please test it with a scenario
+ * where updates actually occur a lot, such as using finite, non-zero u-turn costs.
  *
  * @author Peter Karich
  * @author Michael Zilske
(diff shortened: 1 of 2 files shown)


isochrone/src/main/java/com/graphhopper/isochrone/algorithm/ShortestPathTree.java
```diff
@@ -78,6 +86,7 @@ public String toString() {
         }
     }
 
+    private IntHashSet queueDeleted = new IntHashSet();
```
otbutz, line 89: Is this used?

isochrone/src/main/java/com/graphhopper/isochrone/algorithm/ShortestPathTree.java
```diff
@@ -187,7 +194,11 @@ private double getExploreValue(IsoLabel label) {
 
     @Override
     protected boolean finished() {
-        return queueByZ.isEmpty() || getExploreValue(queueByZ.peek()) >= limit;
+        while (queueByZ.peek() != null && queueByZ.peek().deleted)
+            queueByZ.poll();
+        if (queueByZ.peek() == null)
+            return true;
+        return getExploreValue(queueByZ.peek()) >= limit;
```
otbutz, line 201:
```java
while (!queueByZ.isEmpty()) {
    IsoLabel head = queueByZ.poll();
    if (!head.deleted) {
        return getExploreValue(head) >= limit;
    }
}
return true;
```

Depending on the size of the underlying array it would probably be faster to sort it once and iterate over the array.

karussell: The Java PriorityQueue is an extreme fast implementation (also backed on an array) and so it is incredible hard to beat it even if alternatives are theoretically faster.

isochrone/src/main/java/com/graphhopper/isochrone/algorithm/ShortestPathTree.java
michaz: No.

isochrone/src/main/java/com/graphhopper/isochrone/algorithm/ShortestPathTree.java
otbutz: It would also allow to use a more finegrained Comparator and sort by deleted+z

PR #2093, opened 2020-07-23
add isochrone to measurement

karussell:
Related to #2086.


isochrone/src/main/java/com/graphhopper/isochrone/algorithm/ContourBuilder.java
```diff
@@ -152,4 +151,20 @@ public int compare(Polygon o1, Polygon o2) {
...
+            if (polygon.contains(point)) {
+                return polygon;
+            }
+            if (polygon.getNumPoints() > maxPoints) {
+                maxPoints = polygon.getNumPoints();
+                maxPolygon = polygon;
+            }
+        }
+        return maxPolygon;
+    }
```
karussell, line 169: Is this the correct place? (I did not want to duplicate it)

PR #2094, opened 2020-07-23
fix bucket size

### LINKED ISSUE #2083: isochrone problem in case of multiple polygons
Something is wrong for this request:
`/isochrone?point=43.15729250310914,-77.60433197021486&time_limit=1200&buckets=10`

the last polyline with bucket=9 seems fine:

image

but then it somehow does a big jump and the next bucket=8 (18minutes) is much smaller:

image

When I do this via a separate request:
`/isochrone?point=43.15729250310914,-77.60433197021486&time_limit=1080&buckets=1`
I get a more reasonable polyline:

image

karussell:
I think this fixes #2083.

TODO

 - [x] write test 
 - [x] better way to specify multiple time limits? Something like time_limit=x&time_limit=y&...? -> Probably best in a different issue

### DIFF
--- a/web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
+++ b/web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
@@ -123,8 +123,9 @@ public class IsochroneResource {
             shortestPathTree.setTimeLimit(limit + Math.max(limit * 0.14, 200_000));
         }
         ArrayList<Double> zs = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


michaz:
> Something like time_limit=x&time_limit=y&...?

Yeah. They can't be mixed, though.

karussell:
> If you do that, I'd suggest to convert one syntax to the other somewhere early in the code, and out of sight. Like other legacy parameters. (And then just test the parameter converter.)

Probably best in a different issue

PR #2099, opened 2020-07-28
Set OSMReader#process methods public

boldtrn:
I was browsing through the OSMReader and saw that the three methods:

- processRelation
- processWay
- processNode

Have three different visibility. In one project I am already using the processRelation method to add a hook. Now I wanted to use the processNode method as well and saw that this is not easily possible. I think there are good use-cases to extend all three methods.

WDYT? Would it make sense to set all 3 methods public? Personally I could very well live with setting them protected, but that way at least there is a way to modify it.

### DIFF
--- a/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -311,7 +311,7 @@ public class OSMReader implements DataReader, TurnCostParser.ExternalInternalMap
     /**
      * Process properties, encode flags and create edges for the way.
      */
-    void processWay(ReaderWay way) {
+    public void processWay(ReaderWay way) {
         if (way.getNodes().size() < 2)
             return;
 
@@ -479,7 +479,7 @@ public class OSMReader implements DataReader, TurnCostParser.ExternalInternalMap
             return Double.NaN;
     }
 
-    private void processNode(ReaderNode node) {
+    protected void processNode(ReaderNode node) {
         if (isInBounds(node)) {
             addNode(node);
 
--- a/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -414,7 +414,7 @@ public class OSMReader implements DataReader, TurnCostParser.ExternalInternalMap
         }
     }
 
-    public void processRelation(ReaderRelation relation) {
+    protected void processRelation(ReaderRelation relation) {
         if (tcs != null && relation.hasTag("type", "restriction"))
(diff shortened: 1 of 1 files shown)


boldtrn:
WDYT about this PR? Does it make sense?

boldtrn:
Thinking about this a bit more, it probably makes more sense to set these methods `protected`?

easbar:
I do not know the OSM parsing code too well, but even without knowing it I would expect that some parts of it deal with 'detecting' ways, nodes and relations and some other (hopefully independent) parts specify what is done when encountering these items. Ideally the latter part should be easy to customize. Speaking of making some methods protected I do not see strong reasons against this but I think the better option compared to relying on inheritance would be if we were able to pass in some callbacks instead.

boldtrn:
Ok sure, I am fine with Callbacks as well. Do you have something specific in mind or should I propose something in a new PR?

easbar:
Hm looking at it it seems like OSMReader is already doing too many things so just introducing callbacks in `writeOsmToGraph` probably doesn't really improve anything. So as far as I am concerned we can just go with making the methods protected for now. But maybe wait what @karussell has to say.

boldtrn:
Ok, I would be fine with setting the methods protected as well :).

One other idea I have would be to introduce something like this:

```
// kind of pseudo code
class ReaderCallback{

public void processNote(ReaderNote note)

public void processWay(ReaderWay way)

public void processRelation(ReaderRelation relation)

}
```

Each of these methods would be called first or last of each of the methods in OSMReader. I think I would prefer to call them before the actual method.

One more thing I just thought about would be to change void to boolean. The return value would be used to determine if the regular process code should be executed for this element or not.

easbar:
Hm yeah if the OSMReader would only *read* that would also make sense to me. The problem I see here is that OSMReader does not just read OSM, but it processes the OSM data in a specific way and builds the GraphHopper graph. So not sure if it makes sense to add a callback that somehow modifies this whole process? Especially if the whole process fails if the callbacks do not do certain things?

karussell:
> Hm looking at it it seems like OSMReader is already doing too many things so just introducing callbacks in writeOsmToGraph probably doesn't really improve anything. So as far as I am concerned we can just go with making the methods protected for now. 

Yeah, OSMReader needs to be refactored before we pretend there is a clean hook. I would also not make the methods public.

boldtrn:
Ok thanks for your feedback. So we could agree on setting these methods protected for now? And once the OSMReader is refactored, we could think about adding callbacks?

boldtrn:
Ok, I set processNode and processWay to protected. The method processRelation has been public before - I am not really sure why? Should I set it to protected as well?

karussell:
> Should I set it to protected as well?

Yes, let's try it :)

boldtrn:
Done :)

boldtrn:
Thanks a lot!

PR #2103, opened 2020-07-30
Interpolate Ferry Elevation for Skadi Elevation Provider

### LINKED ISSUE #2098: Interpolate Elevation for Ferries - Mapzen Elevation
Mapzen's elevation data contains bathymetric data (underwater elevation). For ferries it can therefore happen that major elevation changes happen on water. Probably we should interpolate the elevation for ferries like we do for tunnels/bridges.

boldtrn:
This PR fixes #2098. For now I set it up to only interpolate ferries if the SkadiElevationProvider is used. For the other providers this is probably not necessary.

Another improvement on top of this would be to not even lookup the elevation for any ferries (except for tower nodes) and then interpolate the elevation for the ferries. This might speed up the overall import, since no elevation file lookup needs to be done.

If found a nice example in Denmark: http://localhost:8989/maps/?point=57.437455%2C10.538464&point=57.295185%2C10.92041&locale=de-DE&elevation=true&profile=car&use_miles=false&layer=Omniscale

Before:
Screenshot from 2020-07-30 14-18-01
After:
Screenshot from 2020-07-30 14-18-10

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -970,8 +970,15 @@ public class GraphHopper implements GraphHopperAPI {
             float tunnel = sw.stop().getSeconds();
             sw = new StopWatch().start();
             new EdgeElevationInterpolator(ghStorage, roadEnvEnc, RoadEnvironment.BRIDGE).execute();
+            float bridge = sw.stop().getSeconds();
+            float ferry = -1;
+            if (eleProvider instanceof SkadiProvider) {
+                sw = new StopWatch().start();
+                new EdgeElevationInterpolator(ghStorage, roadEnvEnc, RoadEnvironment.FERRY).execute();
+                ferry = sw.stop().getSeconds();
+            }
             ghStorage.getProperties().put(INTERPOLATION_KEY, true);
-            logger.info("Bridge interpolation " + (int) sw.stop().getSeconds() + "s, " + "tunnel interpolation " + (int) tunnel + "s");
+            logger.info("Bridge interpolation " + (int) bridge + "s, " + "tunnel interpolation " + (int) tunnel + "s, ferry interpolation " + (int) ferry);
         }
     }
 


karussell:
Thanks for this! I'm a bit unsure if this special situation for a single provider belongs into core. Likely every elevation provider needs a special handling and it should be done outside of GH? (via `postProcessing`?)

boldtrn:
> Likely every elevation provider needs a special handling and it should be done outside of GH? (via postProcessing?)

Personally, I would not mind to do this outside of GH. However, I feel like this should be part of the core to be honest. Otherwise, we should be strict about this and remove the whole elevation provider handling out of the core?

IMHO, showing the bathymetric data for ferries is a bug and should be fixed. Otherwise everyone using the skadi provider needs to change this manually by overwriting the postProcessing method. For everyone else, this PR should not have any downsides.

karussell:
You are right it should be ignored and so we should have to do it for all providers. How long does it take for world wide? I guess it is under 20 seconds ... then we should remove the `if(instanceof)` and also adding a test for future maintenance :) ?

boldtrn:
> How long does it take for world wide?

To be honest, I haven't tested it for world wide yet. I expect that it is a lot faster than the tunnel interpolation. In OSM there are almost 1.000.000 tunnels but only 25.000 ferries.

boldtrn:
I removed the if, so it should now run for all ElevationProviders. I also renamed the method as well to reflect what it does now.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -962,16 +962,23 @@ private boolean hasInterpolated() {
...
+    void interpolateBridgesTunnelsAndFerries() {
         if (ghStorage.getEncodingManager().hasEncodedValue(RoadEnvironment.KEY)) {
             EnumEncodedValue<RoadEnvironment> roadEnvEnc = ghStorage.getEncodingManager().getEnumEncodedValue(RoadEnvironment.KEY, RoadEnvironment.class);
             StopWatch sw = new StopWatch().start();
             new EdgeElevationInterpolator(ghStorage, roadEnvEnc, RoadEnvironment.TUNNEL).execute();
             float tunnel = sw.stop().getSeconds();
             sw = new StopWatch().start();
             new EdgeElevationInterpolator(ghStorage, roadEnvEnc, RoadEnvironment.BRIDGE).execute();
+            float bridge = sw.stop().getSeconds();
+            float ferry = -1;
```
karussell, line 974: no need for that anymore :)

boldtrn: Thank, I removed it :)

karussell:
Thanks!

boldtrn:
Just to provide you with an update. Running the FerryInterpolation for the planet files takes 8s seconds. Here is the full log for reference of the other values: `Bridge interpolation 14s, tunnel interpolation 8s, ferry interpolation 8`

PR #2105, opened 2020-08-03
Allow highway=bridleway with bicycle=yes for BikeCommonFlagEncoder

ratrun:
See issue #504. Now allowing highway=bridleway with bicycle=yes for bicycles.


karussell:
Thanks a lot!

karussell:
With this we can close #504 or should we open it until we have country-dependent routing to allow bridleways in UK even without explicit bicycle permission?

ratrun:
As it was only requested in #504 with explicit tagging (bicycle=yes) I would close it.

PR #2100, opened 2020-07-29
Make it easier to extend Router

boldtrn:
This PR resulted from this discussion: https://discuss.graphhopper.com/t/make-router-extendable/5871

This should make it a lot easier to extend the Router

PR #2106, opened 2020-08-04
Enable overwriting of routeRoundTrip

fabudlx:
Hey together,

I tried to overwrite the lookup function by extending the routeRoundTrip function, since https://github.com/graphhopper/graphhopper/pull/2100 allows to overwrite this function now. There are still two methods that I really need to successfully create my own routeRoundTrip function. Is it possible to change these two methods from private to protected?

best regards,
Fabian


fabudlx:
Are there any comments/opinions on the proposed changes?

easbar:
I would prefer thinking about which parts of the GH code shall be reusable and are not reusable yet. For me there is a difference between 'library' kind of code that is 'meant' to be re-used and 'server/config' code that is more or less throw-away code that was only written once and in a specific way to implement the GH server.

The `routeRoundTrip` method itself is rather short IMO and I would expect it is not too much of a problem to overwrite and modify it entirely? Then again it calls `concatenatePaths`(which calls the private `createPathMerger` method) and `createFlexiblePathCalculator` which are private, so they would need to be copied and modified as well.

Don't get me wrong I do not think its a very big deal to make two more methods protected, especially if there is at least someone who thinks its useful. But on the other hand I do not think this is very sustainable, we should not open more and more methods just so it is easy to modify them a bit. The code in `Router` (just like `GraphHopper`) is still heavily mixing 'library'/useful/reusable code with server/configuration kind of code and I think this is the real problem here.

What to do with this PR? I do not have a strong opinion, I'd just like to say that IMO making 'random' methods protected should not be the way to go and there is no guarantee these methods will stay the way they are in the future. I am looking forward to hearing other's opinions here :)

boldtrn:
A lot of important logic that I regularly find needs adjustment to special problems is hidden in the Router. For example in #1713 I had an issue changing the snap prevention. I also remember a few other questions where similar things were asked or sought after.

So I think allowing to modify the Router is important and I think a lot of modifications are probably very problem specific, so adding them to GH core is probably not a good idea. I think in this case it would be the easiest to set these methods to protected for now. 

In the long run I am wondering if we might find a way to make the *Routing (RoundtripRouting, ViaRouting) classes easier to extend by defining a common interface, if possible? The used *Routing classes are not especially extension friendly - considering that all methods in the *Routing class are static. I know we just got rid of the RoutingTemplates ;). On the other hand, the code is now quite straight forward, which is a big plus as well. I agree, I feel a bit uncertain about what would be the best step in the long run.

michaz:
> So I think allowing to modify the Router is important and I think a lot of modifications are probably very problem specific, so adding them to GH core is probably not a good idea. I think in this case it would be the easiest to set these methods to protected for now.

I just want to emphasize that what @easbar says is important in full generality:

Extending the Router to "modify" it is not something we should support. Maybe this sounds drastic, but the correct way to modify the Router is to fork it, unless the modification is done by very specific extension points which we _support_. Every method that is not private and not final is a _commitment_ and causes serious work in the long run.

It basically dumps the work that maintaing a fork would be -- on us.

boldtrn:
> Every method that is not private and not final is a commitment and causes serious work in the long run.
> It basically dumps the same work that maintaing a private copy would be -- on us.

I know what you mean, but I don't agree with you on this. So far, even after the 1.0 release, GraphHopper changed a lot of interfaces, including the here discussed Router and RoutingTemplates. Don't get me wrong, change is good and change hurts these people that think it's necessary to extend some lower level feature. So I think no one can expect that every public method will stay the same forever (especially with the release of a major version ;)). I do understand, that once a method is public, one is hesitant to change it and I can see the issue with that. On the other hand maintaining a fork/copy of GraphHopper is a whole lot more work compared to changing a method.

michaz:
> RoutingTemplates

Don't want to dwell on this detail, we can agree to disagree (especially since we are sitting on opposite ends of the trade-off :-)), but RoutingTemplate was a particularly not-so-great interface. It is exactly an example of the thing I don't want to come back in salami slices, or disguised as protected methods. It had like three or four methods that were _totally ad-hoc_, each one just _pretending_ to be something general where you could just write your own implementation for and where it was perfectly clear what it meant, but in reality, it became 20 times clearer by just converting the exactly three implementations into exactly three procedures and removing this abstraction.

This _really_ held us back.

boldtrn:
No worries, I haven't been a big fan of the RoutingTemplates either. Also I really appreciate this great improvement by @easbar. The code is now indeed a lot cleaner and out right just better. So we fully agree on that.

My point is, don't let public or protected methods hold you back. Especially protected methods. It's OK if they change. 

Creating and maintaining a fork on the other hand just to change a minor thing seems like an anti-pattern to me. This leads to half maintained and outdated forks that don't improve anything and don't contribute anything back to GraphHopper. After several years of being outdated they will be thrown away and a new fork is created and the changes start again.

IMHO, a library should be open enough to be customizable, but close enough, that it's easier to contribute back an improvement instead of hacking it in a local fork. This allows to customize a few features that are important to the business case and allow the less business important features to be contributed.

easbar:
> @easbar Are you teasing me? :-)

No, not at all :) I just wanted to say that I do not want to decide myself if and how which parts of GH should be made extensible or not. I like to give my opinion, but I hope we can find some consensus on these things.

> Every method that is not private and not final is a commitment and causes serious work in the long run.

I am not sure if I agree. I mean I agree we should not expose an API that is not meant to be kept/maintained as an API, but I do not agree that this causes us lots of work (maybe some time ago, but not right now), because if we are honest we did not care too much about changing not-private or final methods recently. But I wonder if addng more protected/public/open code that just like private code might be changed in the next release anyway does not make it even worse for those who maintain a modified GH instance. How does it help not to have a clear boundary and no guarantees about what might and might not change?

To me it seems it very much comes down to how much code needs to be customized. If its just about making one method protected isn't it very easy to create a fork and keep it up to date? And if more changes are needed why does it help to make one method protected upstream when all this other code needs to be changed or added as well and could also break with the next release? 

> IMHO, a library should be open enough to be customizable, but close enough, that it's easier to contribute back an improvement instead of hacking it in a local fork. This allows to customize a few features that are important to the business case and allow the less business important features to be contributed.

Yes this sounds right, but how do we achieve this? Also the question is when a contribution is really an improvement. I think its an improvement if it helps most or many users of the code, but if it is just an adjustment that prevents work downstream it is not (I mean this just in general, I do not have any specific contribution in mind here).

> My point is, don't let public or protected methods hold you back. Especially protected methods. It's OK if they change.

Ok good to know, so you think having protected methods that are just as likely to change as private ones makes it easier to maintain a customized GraphHopper? And its easier because people who want to customize something do not have to create a fork and change the accessibility of this method this way?. Honestly, I do not really understand this, but since I never really maintained a custom GraphHopper instance I just take your word or maybe you can explain this a bit more.

My goal is rather making the GH code so decoupled and modular that it is easy to take the parts and create something interesting with them. My goal is not so much creating a stand-alone GH server that can just be pulled in as a maven artifact and be adjusted by overriding a few methods. And the reason is that I think the first approach is much more powerful while I do not really have an idea how the latter is even supposed to work. 

But apart from these general thoughts, like I said, I do not see a huge problem in making a high level method like `routeRoundTrip` protected instead of public for now. Especially since we are currently still far from the ideal.

> This really held us back.

Just my two cents on the routingtemplates. I agree this held us back, but not even so much because we did not want to break compatibility, but rather because the code was or became so twisted that it was even hard to remove it :) So I agree that just the attempt of making things 'pluggable' can come with significant costs. But obviously we should not compare the routing templates with making a few methods protected.

michaz:
> I agree this held us back, but not even so much because we did not want to break compatibility, but rather because the code was or became so twisted that it was even hard to remove it :) So I agree that just the attempt of making things 'pluggable' can come with significant costs.

Yes, exactly, that's what I meant as well. That interface and its methods were not a problem because we were hesitating to change their signatures, but because they _existed_, instead of just linearly writing down the three different alternatives.

boldtrn:
> To me it seems it very much comes down to how much code needs to be customized. If its just about making one method protected isn't it very easy to create a fork and keep it up to date? And if more changes are needed why does it help to make one method protected upstream when all this other code needs to be changed or added as well and could also break with the next release?

The biggest benefit from my perspective is that there is clear a boundary. I know exactly what is custom code and what is no custom code. This makes writing the code itself more complicated, but down the road it is easier to maintain.

Every GraphHopper update produces work to get everything compatible again, both for forks and for library extends. I maintain a few different projects that either fork GraphHopper or extend it as a library. From my experience, the library extends are a lot easier to maintain. That's why I try to get it working with extends and avoid forking. 

> Ok good to know, so you think having protected methods that are just as likely to change as private ones makes it easier to maintain a customized GraphHopper? And its easier because people who want to customize something do not have to create a fork and change the accessibility of this method this way?. Honestly, I do not really understand this, but since I never really maintained a custom GraphHopper instance I just take your word or maybe you can explain this a bit more.

Maintaining a fork is always ugly and when you update a fork after half a year for example you end up in merge hell. You might merge code for a few days, then fix all the obvious compiler errors. Then it's about fixing tests, both GraphHoppers own tests, then your own tests that test your customization. Just to give you an idea, I recently worked on a GraphHopper update with over 1000 commits, ~80k lines added and ~45k lines removed. There is no way that you can review something like this properly.

On the other hand, if you update library code, you could first comment out all your customization and get the basics back up an running, then your start to slowly add your custom code.

easbar:
Thanks for sharing some insights!

> On the other hand, if you update library code, you could first comment out all your customization and get the basics back up an running, then your start to slowly add your custom code.

Why can you take this approach when you extended the library, but not when you forked the upstream project? Maybe this example is too simple but I imagine it like this:

1) extend the library
- create new project
- add GH dependency
- add custom code, overwrite methods etc.

2) fork the library
- fork GH
- make stuff accessible (make methods public, maybe change some method signatures, add a hook etc.)
- create new project
- add forked GH dependency
- add custom code, overwrite methods etc.

So basically in the fork you open up methods just the way you need it and the (probably much heavier) customizations are in the actual custom project, but these you need to do (and later update) anyway. When you update to a new GH version you first need to check if your previous hooks/replacements/public/signatures still work or you need to find a new way to add them.

boldtrn:
> So basically in the fork you open up methods just the way you need it and the (probably much heavier) customizations are in the actual custom project, but these you need to do (and later update) anyway.

Ah, now I see, that would be indeed an option. When I talked about a fork I actually meant a fork with code changes and not just accessibility changes.

So in this case:
> Option 1) extend the library
> Option 2) fork the library

Both options are indeed comparable in effort.

easbar:
For now this is the recommended approach to add entry points / modification hooks etc. to the GraphHopper app/server code: https://github.com/graphhopper/graphhopper/pull/2106#issuecomment-688914649

Issue #655, opened 2016-01-30
Bike improvement: support for smoothness tag


ratrun:
The assumed highway speed should be adopted depending on http://wiki.openstreetmap.org/wiki/DE:Key:smoothness
The combination with all possible highway types, surface and grade tags needs to be considered.

karussell:
This seems to be important but do you have an example where considering smoothness would improve the routing? Changing this without a real world example could make it worse.

lukie80:
Smoothness is highly correlated with tracktype (grade#) but still allows for a refinement. The official definition (https://wiki.openstreetmap.org/wiki/Key:smoothness) is based on the physical usability of a way for wheeled vehicles:
- intermediate: good for regular bike
- bad: good for trekking bike / regular car
- very bad: only MTB and some effort  / high clearance car
- horrible: only MTB and high effort / Jeep

I would rate the following track as grade5 and very horrible if wet and grade4 and horrible if dry for example.
[Muddy forest track picture on pixabay.com](https://pixabay.com/en/forest-way-mud-track-tree-2771493/)

My personal statistical experience for German forest tracks in the flat country:
- grade1: good - intermediate, (default: good)
- grade2: good - intermediate, (default: intermediate)
- grade3: intermediate - bad, (no default)
- grade4: bad - very bad, (default: bad)
- grade5: bad (short grass on hard ground / lawn) - very horrible (overgrown, soft and uneven ground), (no default)

So basically, it mostly would alter the tracktype by +/-1 except for grade5 which could be lowered even more. One scoring possibility for bicycles could be:

- good and better: 24
- intermediate: 20
- bad: 15
- very bad: 6
- horrible and worse: 0

PR #2108, opened 2020-08-19
Add support for tag "smoothness", implements #655.

ratrun:
After a long time an implementation for #655.


karussell:
Thanks! Do you have examples where this PR leads to an improved routing?

ratrun:
Sorry, I do not have any real-world example. I tried to find one via overpass, but this is not that easy as most of the bad smoothness values are located within forest area.

karussell:
Without a real-world example it is not really possible to review the effect of it. Should we implement such a change at all if there is no such situation in real world?

ratrun:
I found a real world example: https://graphhopper.com/maps/?point=47.874663%2C16.120527&point=47.888809%2C16.110849&locale=de&vehicle=bike&weighting=fastest&elevation=true&turn_costs=false&use_miles=false&layer=Omniscale 
The two segments with bad surface are rather short (https://www.openstreetmap.org/way/543917919 and https://www.openstreetmap.org/way/805552627), but should be enough to take the route along the primary road with the modifications.
Even if it is hard to find real world examples where other ways are taken the modification improves the estimation of the travel time as lower speed values are assumed when the surface gets bad.

karussell:
Oh, it seems this tag occurs quite frequently: over 2 million times on ways.

Instead of creating the mechanism inside the FlagEncoder I would create a new `smoothness` encoded value that can be used in a custom model (and in the FlagEncoder). Would you have time to rewrite this? If yes, let me know if you need help with this.

ratrun:
I have managed to implement the encoded value for `smoothness` (see https://github.com/graphhopper/graphhopper/compare/master...ratrun:smoothnessEV). Shall I create a separate pull request for this? Unfortunatelly I have no idea how the new stuff could be used in the bike flag encoder and what you mean by that.

karussell:
> I have managed to implement the encoded value for smoothness (see master...ratrun:smoothnessEV). Shall I create a separate pull request for this? 

Thanks for this and yes a pull request is appreciated. We can then use the smootness EncodedValue in the FlagEncoder for the default bike vehicle. 

> Unfortunatelly I have no idea how the new stuff could be used in the bike flag encoder and what you mean by that.

Usually we would use the EncodedValue in a custom model. See e.g. the examples or custom-model.md documentation. But it is also possible to integrate it in the older FlagEncoder architecture. See e.g. 

> roundaboutEnc = getBooleanEncodedValue(Roundabout.KEY);

in AbstractFlagEncoder. And a similar thing could be done with the smoothness to access the value (which is filled before from the TagParser)

ratrun:
I have created the pull request for the enocded value, see #2303 as suggested. I also tried to use the encoded value in the BikeCommonFlagEncoder, but without success. I ran into "IllegalArgumentException: Cannot find EncodedValue smoothness in collection: null." when I tried to execute the enhanced tests. 
In the meantime I found a similar location in the car flag encoder: `applyBadSurfaceSpeed` is also acccessing the surface tag of the way directly like the code in this pull request above does.  There is no usage of the encoded value for surface, too. I would ask you to rewrite that car part first such that I can better understand what you mean. I'm sure that I'll manage to rewrite the BikeCommonFlagEncoder for usage of the smoothness encoded value with such an example.

karussell:
> I ran into "IllegalArgumentException: Cannot find EncodedValue smoothness in collection: null." when I tried to execute the enhanced tests.

Ah, I think we would need to automatically add the parser whenever a BikeFlagEncoder is added: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/EncodingManager.java#L336

ratrun:
Closing discussion here, we contine in #2303.

PR #2110, opened 2020-08-21
Memory efficient solutions

kubikrubikvkube:
This is minor refactoring to improve speed. See commits descriptional for additional info


karussell:
> This is minor refactoring to improve speed.

Did you benchmark the result? You can use our measurement procedure:

./

> System arraycopy is more memory efficient 

Why is this more memory efficient? It might be faster but probably only for long arrays and could also be slower or change nothing. Only a benchmark will show.

> String concatenation in loop is not memory efficient because it copies entire string. 

This would be only relevant for critical code and if it is critical JIT will likely replace it with `+=` internally, so making this less readable is IMO suboptimal.

PR #2113, opened 2020-08-26
Add weighted in/out edge explorers

easbar:
This PR adds two new `createEdgeExplorer`-like methods to `Graph`. One returns an explorer for outgoing and one an explorer for incoming edges and both take a `Weighting` as input. Basically `createOut/InEdgeExplorer(Weighting)` is a more general version of `createEdgeExplorer(DefaultEdgeFilter.out/InEdges(accessEnc))`. I am wondering if we should go in this direction.

* With the new explorers the access flags become less important, because the explorers rely on the weighting instead. The weighting can (and probably for some time will) use the access flags internally, but this way we use the weighting to ultimately decide whether or not an edge is accessible or not. For example in some custom weighting the edge weight might be infinite even though the access flags are set.

* Evaluating the weighting makes the graph iteration more expensive. This is unfortunate especially for cases where we do not need the weight and only want to check the accessibility of an edge. We could add something like `finiteEdgeWeight(EdgeIteratorState, boolean)` to `Weighting` to allow optimizing for such cases in the `Weighting` implementation.

This PR is related to #1835. I already did something similar in #1837 for CH preparation, but now I am wondering if it would not be better to have this in BaseGraph as well. Besides the possible performance related issues it seems cleaner to hide the access flags in Weighting instead of using the access flags and the weighting to determine edge access.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/AbstractNonCHBidirAlgo.java
+++ b/core/src/main/java/com/graphhopper/routing/AbstractNonCHBidirAlgo.java
@@ -46,7 +46,8 @@ public abstract class AbstractNonCHBidirAlgo extends AbstractBidirAlgo implement
     protected final NodeAccess nodeAccess;
     protected final Weighting weighting;
     protected final FlagEncoder flagEncoder;
-    protected EdgeExplorer edgeExplorer;
+    protected EdgeExplorer outEdgeExplorer;
+    protected EdgeExplorer inEdgeExplorer;
     protected EdgeFilter inEdgeFilter;
     protected EdgeFilter outEdgeFilter;
     protected EdgeFilter additionalEdgeFilter;
@@ -59,7 +60,8 @@ public abstract class AbstractNonCHBidirAlgo extends AbstractBidirAlgo implement
         this.flagEncoder = weighting.getFlagEncoder();
         this.graph = graph;
         this.nodeAccess = graph.getNodeAccess();
-        edgeExplorer = graph.createEdgeExplorer();
+        outEdgeExplorer = graph.createOutEdgeExplorer(weighting);
+        inEdgeExplorer = graph.createInEdgeExplorer(weighting);
         outEdgeFilter = DefaultEdgeFilter.outEdges(flagEncoder.getAccessEnc());
         inEdgeFilter = DefaultEdgeFilter.inEdges(flagEncoder.getAccessEnc());
         int size = Math.min(Math.max(200, graph.getNodes() / 10), 150_000);
@@ -85,12 +87,7 @@ public abstract class AbstractNonCHBidirAlgo extends AbstractBidirAlgo implement
         if (fromOutEdge == ANY_EDGE) {
             fillEdgesFrom();
         } else {
-            fillEdgesFromUsingFilter(new EdgeFilter() {
-                @Override
-                public boolean accept(EdgeIteratorState edgeState) {
-                    return edgeState.getOrigEdgeFirst() == fromOutEdge;
-                }
-            });
+            fillEdgesFromUsingFilter(edgeState -> edgeState.getOrigEdgeFirst() == fromOutEdge);
         }
     }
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/graphhopper/storage/Graph.java
```diff
@@ -100,16 +101,47 @@
...
+     *
+     * @see #createEdgeExplorer(EdgeFilter)
+     */
+    default EdgeExplorer createOutEdgeExplorer(Weighting weighting) {
+        final BooleanEncodedValue accessEnc = weighting.getFlagEncoder().getAccessEnc();
+        return createEdgeExplorer(edge -> {
+            boolean access = edge.get(accessEnc);
+            if (!access)
+                return false;
+            return Double.isFinite(weighting.calcEdgeWeight(edge, false));
```
easbar, line 127: This might make using this explorer too expensive for some cases, because if the access flag is set we would have to calculate the weight (possibly expensive e.g. for custom weighting) even though all we want to know is whether its finite or not. Something like `weighting.isEdgeWeightFinite(edge, false)` could help optimizing this. But probably there is no need to optimize this before there is an actual performance issue (so far this new explorer is not used anywhere).

core/src/main/java/com/graphhopper/routing/AbstractNonCHBidirAlgo.java
```diff
@@ -59,7 +60,8 @@ public AbstractNonCHBidirAlgo(Graph graph, Weighting weighting, TraversalMode tM
         this.flagEncoder = weighting.getFlagEncoder();
         this.graph = graph;
         this.nodeAccess = graph.getNodeAccess();
-        edgeExplorer = graph.createEdgeExplorer();
```
easbar, line 62: Another option would be using `graph.createEdgeExplorer()` 'everywhere' (unfiltered edges) and then filter them by whatever one needs to do. This is maybe the most flexible approach, but it is also a bit more messy to do the filtering 'everywhere'?

easbar:
On a more general note, can we agree on the following?

* The base graph without weighting can be seen as an unweighted graph (all edges are bidirectional)
* The `Weighting` makes the graph directed, because edges with weight=Infinite are not accessible. An edge with infinite weight is the same as an edge that does not exist.
* The access flags are mostly an optimization to realize this. For example they are used to 'disable' huge parts of the road network (for example inner cities for trucks and highways for pedestrians any bicycles). But conceptually this is the same as setting the edge weights for these parts of the network to Infinity.

karussell:
Can you explain what benefits we get? Is this cleaner for CH? At the moment I do not like it too much :) as for non-CH this would duplicate the potential expensive, "uncached" weight calculation and this fact would be hidden. My favourite is indeed your proposal to traverse all edges:

> Another option would be using graph.createEdgeExplorer() 'everywhere' (unfiltered edges) and then filter them by whatever is needed. This is maybe the most flexible approach ...

If I correctly used our visualizations you can see this already in our benchmark, e.g. for big_map_very_custom it is 91ms vs. ~74ms on master.

easbar:
> Can you explain what benefits we get? Is this cleaner for CH?

The benefit would be to have a clean way of getting/traversing all incoming/outgoing edges (this is relative to the weighting that is used) without duplicating the edge filtering logic. Kind of the same benefit we get when using `createEdgeExplorer(EdgeFilter)` instead of `createEdgeExplorer()`+doing the filterering using some `accessEnc`. But as of now I do not need this anymore (initally I thought its useful for edge-based CH preparation mostly). So I agree we should not add it as long as we do not really need it.

Then again with this logic we could either have something like `WeightedEdgeFilter` that uses a weighting just like we have `DefaultEdgeFilter` that uses the access flags. Or we remove `createEdgeExplorer(EdgeFilter)` entirely :). I think my point is that filtering edges based on access flags instead of the weighting is not really the right thing to do, because like I said the access flags are only an optimization/implementation detail and the weighting should be the single source of truth when we ask the question which edges are accessible (in any of the two directions).

But anyway, lets close here and probably reconsider this when we deal with #1835.

PR #2114, opened 2020-08-30
Update default config file name in graphhopper.sh output

jb2martel:
The name printed did not match the file used


karussell:
Thanks!

PR #2118, opened 2020-09-07
Remove graph compacting

michaz:
Let's see if this works out..


easbar:
I checked the performance stats and it looks like the CH preparation even runs a bit faster with this PR. The routing speed is a bit slower, but this is probably due to a weakness of our measurements: We use random query samples based on the node-ids, so if we change the node-ids we run the benchmarks with different routes unfortunately. It would be better to run the benchmarks using random gps coordinates.

michaz:
Can‘t we run all benchmarks long enough that enough routes are sampled so that it doesn‘t matter if they are different? Wouldn‘t it be even more meaningful then? If I understand it correctly, benchmarks need confidence intervals anyway, like in the „Microbenchmark Harness“..

easbar:
Yes that would be ideal, but in practice I feel like this is not so easy to do. Also we aren't necessarily interested in the 'real' average speed, but its enough to compare different versions on a fixed set of sample routes IMO.

michaz:
I‘m fine with something like „I‘m 95% sure that this is not more than 5% slower“. :-)

easbar:
> I‘m fine with something like „I‘m 95% sure that this is not more than 5% slower“. :-)

Yes me too, but its hard do make sure this is the case. I think we should just sample the gps coordinates instead of the node Ids (#2121), it should be no problem to do, I can do it today. This will 'break/change' our measurement results, but afterwards they should be more stable to such changes. But no need to hold back this PR because of this.

michaz:
What would happen if we just used a new random sample every time we run the benchmark?

 Then we would get some (large?) variability in results, and isn‘t that approximately the range we can (should?) ignore when deciding if some code change is bad or good?

Or is it that for any practical value, the variability would be so big that we would miss actually meaningful performance hits?

michaz:
Okay, the _amount of work_, measured in, like, sum of all route lengths, must have less variability than the phenomenon we‘re interested in.

easbar:
> Or is it that for any practical value, the variability would be so big that we would miss actually meaningful performance hits?

For CH maybe not so much, but for some of the slower measurements with smaller sample sizes yes (AStar and alternatives for example). Using different samples certainly has the potential to 'hide' real performance changes (or make you think there is a change when really there is not). What / how large a real performance change is is of course another question. 

Another benefit we get from using the same sample is that we can look at the distance checksum of all queries to see if the results/routes are still the same.

easbar:
> If I understand it correctly, benchmarks need confidence intervals anyway, like in the „Microbenchmark Harness“..

Especially for the routing queries it would be better to not only record the mean routing time but also some more details about the distribution (maybe 90,95,99% percentiles or variance etc.). We agreed on this many times but just did not do it yet :)

easbar:
This broke the measurements, because removed nodes (that are now not removed again) can cause connection-not-found errors. But this: 250834af724f4c192b91c36d14f0b1153a001ba2 should fix it until we use random GPS points in #2124. Also this commit c4a32d537768d433d81facd23e914d2d84eaf8de should make sure we notice when the measurements fail with connection-not-found in the future.

michaz:
Ahh. Got it. So now we're choosing node numbers randomly, and so origin/destination can end up in disconnected components. But one we go through the LocationIndex, this will be fixed, since that doesn't return nodes that only have access-restricted edges. Right?

easbar:
Yes exactly

Issue #2115, opened 2020-08-30
Too small interval in PT transfers


Octanas:
I’ve been having some problems with the PT routing but have been able to get around them. This one is tricky though.

In this particular case, the departure time of the second transport is just 44 seconds after the arrival time of the first transport, and since the stops aren’t really that close, it would be impossible in a real life scenario to do that transfer on time.

![image](https://user-images.githubusercontent.com/55537820/91663214-6b7a3880-eadf-11ea-8b34-97e2eb48e29b.png)

```
...
{
“type”: “pt”,
“departure_location”: “Carcavelos”,
“geometry”: {…},
“distance”: 0.0,
“feed_id”: “gtfs_0”,
“is_in_same_vehicle_as_previous”: false,
“trip_headsign”: “Cascais - Cais do Sodré Cais do Sodré”,
“travel_time”: 1590000,
“stops”: […],
“trip_id”: “LCP:27028974”,
“route_id”: “LCP:401”,
“arrival_time”: “2020-08-24T15:44:00.000+0000”,
“departure_time”: “2020-08-24T15:17:30.000+0000”
},
{
“type”: “pt”,
“departure_location”: “CORPO SANTO”,
“geometry”: {…},
“distance”: 0.0,
“feed_id”: “gtfs_0”,
“is_in_same_vehicle_as_previous”: false,
“trip_headsign”: “735 - Hosp. Sta. Maria ALAMEDA D. A. HENRIQUES”,
“travel_time”: 213000,
“stops”: […],
“trip_id”: “LCS:26991318-0”,
“route_id”: “LCS:191142”,
“arrival_time”: “2020-08-24T15:47:55.000+0000”,
“departure_time”: “2020-08-24T15:44:22.000+0000”
},
...
```

The transfers.txt file specifically states a min_transfer_time of 303 seconds for a transfer between those two stops, which is clearly not being respected. Is this a bug?

Octanas:
I believe I fixed the issue. The transfers.txt parsing is fine, min_transfer_time gets saved in memory. However, when adding the transfers to the departure timeline (`insertInboundTransfers` function in GtfsReader.java), every transfer object has min_transfer_time value 0. I traced the problem back to `getTransfersToStop` function in Transfers.java.

[...]

PR #2119, opened 2020-09-07
Fix to the lack of minimum transfer time in PT routes

Octanas:
Made the changes so that the PT routing takes the minimum transfer time between stops into account, as discussed on #2115.


michaz:
Awesome!

PR #2120, opened 2020-09-07
Add binary heap with efficient update operation and use it for node-queue in CH preparation

easbar:
So far we use a `TreeMap<Integer, Long>` to queue the nodes for CH preparation, but this wastes lots of memory. Using jol I find that 1mio nodes take about 64MB:

```java
   @Test
    public void testMemory() {
        GHTreeMapComposed heap = new GHTreeMapComposed();
        Random rnd = new Random(123);
        for (int i = 0; i < 1_000_000; i++) {
            heap.insert(i, rnd.nextFloat());
        }
        System.out.println(heap.getSize());
        System.out.println(GraphLayout.parseInstance(heap).toFootprint());
    }
```
```xml
<!-- pom.xml -->
<dependency>
  <groupId>org.openjdk.jol</groupId>
  <artifactId>jol-core</artifactId>
  <version>0.10</version>
</dependency>
```

and we use an additional float per node to store the `old_priorities` (another 4MB per 1mio nodes).
This PR adds a modified version of `IntDoubleBinaryHeap` that only uses two ints and a float per node (12MB instead of 68MB for 1mio nodes).

See inline comments for more details (The `BinaryHeapWithUpdate` class especially).


core/src/main/java/com/graphhopper/apache/commons/collections/IntFloatBinaryHeap.java
```diff
@@ -15,29 +15,26 @@
...
 /**
  * This class is a partial copy of the class org.apache.commons.collections.BinaryHeap for
- * just the min heap and primitive, sorted double keys and associated int elements.
+ * just the min heap and primitive, sorted float keys and associated int elements.
  * <p>
  * The library can be found here: https://commons.apache.org/proper/commons-collections/
  */
-public class IntDoubleBinaryHeap implements BinHeapWrapper<Number, Integer> {
-
+public class IntFloatBinaryHeap {
```
easbar, line 27: I renamed this from `IntDouble` to `IntFloat`, because this is what it actually uses. Shall we simply rename it to `(GH?)BinaryHeap`? Also I removed the `BinHeapWrapper` interface, because it looked like this was only needed for testing and I achieved the same with a testing interface.

core/src/test/java/com/graphhopper/routing/AlternativeRouteEdgeCHTest.java
```diff
@@ -141,9 +141,10 @@ public void testCalcOtherAlternatives() {
         hints.putObject("alternative_route.max_paths", 4);
         AlternativeRouteEdgeCH altDijkstra = new AlternativeRouteEdgeCH(g.getRoutingCHGraph(), hints);
         List<AlternativeRouteEdgeCH.AlternativeInfo> pathInfos = altDijkstra.calcAlternatives(10, 5);
-        assertEquals(2, pathInfos.size());
+        assertEquals(3, pathInfos.size());
         assertEquals(IntArrayList.from(10, 4, 3, 6, 5), pathInfos.get(0).path.calcNodes());
         assertEquals(IntArrayList.from(10, 4, 8, 7, 6, 5), pathInfos.get(1).path.calcNodes());
+        assertEquals(IntArrayList.from(10, 4, 8, 7, 6, 3, 2, 9, 1, 5), pathInfos.get(2).path.calcNodes());
```
easbar, line 147: @michaz  What do you think about this? The third alternative does not make so much sense because its just like the second but with an extra little detour. This probably now occurs because the shortcuts are slightly different. Or do you think I might have broken something? Defining what an 'acceptable' alternative is on such a small graph is probably rather hard and/or the heuristics aren't even meant to work for such small graphs?

michaz: Yeah, that‘s probably okay. Can‘t think of something you could have broken here.

karussell:
Thanks! 

> 12MB instead of 68MB for 1mio nodes

I tried the map bayern-200101.osm.pbf with 4 profiles (car, car4wd, foot and bike) and every node-based CH profile has more than 3 million nodes (it says), which would mean a memory difference of approx 3*4*56MB=672MB, which in practise does not seem to come close. This min_heap branch allowed me to do the import with `-Xmx1350m` (1300m failed) whereas for master I only needed `-Xmx1600m` to avoid OOM problems, the theory would have suggested that at least `-Xmx1950m` are necessary. Of course this is a really good improvement :) but likely the different start times reduces the memory pressure a lot. I.e. the map is small enough so that the first CH preparation is already started and reduced the queue, when the others preparations like foot have not yet fully build the queue and so peak memory demand do not come at the exact same time. (BTW: unfortunately I ran all tests with JDK 12 by accident but shouldn't matter much)

Now, the speed of the preparations seems to be the same. I only added 200MB to the minimum memory to avoid a too different memory pressure. Also our measurement tests on the server look fine speed-wise.

So this branch helps a lot for bigger maps and multiple preparations. Really good work!

easbar:
Thanks for testing!

Regarding the 'memory pressure': Its true that the TreeMap we used beforehand shrinks during the preparation (it gets smaller and smaller the more nodes get contracted, and especially at the beginning of the preparation this goes really quick as many nodes get contracted without adding shortcuts). The new heap implementation does not shrink, it just keeps its size. We could trim it after certain intervals as long as we keep the full `positions` int[] array, but not sure if this is really needed. 

Regarding the Xmx experiments we have to be careful about what causes the main memory usage. Its a good idea to use several CH profiles. For example I only tested with a single one and in this case often the `sortedNodes` do not cause the maximum memory usage. Instead it was either the OSM reading or, in some cases the subnetwork removal. Especially the subnetwork removal made testing a bit complicated because sometimes my Xmx limit was exceeded and sometimes it was not. Apparently the subnetwork search sometimes takes more memory than other times.

My main motivation for this PR was the fact that in this branch: `prepare_graph` I need to allocate additional memory *at the beginning* of the preparation, so the memory size of `sortedNodes` becomes more critical than in current master. 

So I think using the new heap here has no disadvantage, except that it does not shrink, but we can implement this if really needed. I also tried this new heap in `Node/EdgeBasedWitnessPathSearcher`, but unfortunately its not an improvement. Apparently the additional cost of keeping track of the positions in `percolateUp/Down` is not worth it to speed up the `update()` method in this case.

karussell:
> The new heap implementation does not shrink, it just keeps its size. 

Ah, good to know, but I don't think we need to change it for now. If we have lot of time and energy we could implement this. And independent from this we could implement some kind of a "delayed start" and start a new preparation in parallel only when enough memory is available. This way the number of threads is dynamic and we do not need to configure it and always have the shortest total preparation time possible.

> Its a good idea to use several CH profiles. 

Yeah, I also noticed the memory problems only happen for the import for 1 or 2 vehicle profiles :)

> I also tried this new heap in Node/EdgeBasedWitnessPathSearcher, but unfortunately its not an improvement. 

Ok

easbar:
Hm, after merging master (no more storage.optimize()) the edge-based CH preparation got really slow in this branch (too many lazy updates). Will have to investigate..

PR #2124, opened 2020-09-08
Use random GPS points instead of nodes in Measurement

### LINKED ISSUE #2121: Measurement results depend on node-ids
Currently we draw random source/target nodes for our sample queries in the `Measurement` class. However, this means our results depend on the node IDs. For example without `optimize()` we get different queries/results: #2118

Instead, we should:

1) Generate random GPS coordinates (within the boundaries of the measurement map) and check if we get a valid snapped node, if yes add this node to an array of query nodes. Do this until we got a certain amount of query nodes.

2) Draw random nodes from our array of query nodes to perform the routing queries.

easbar:
Fixes #2121.

### DIFF
--- a/tools/src/main/java/com/graphhopper/tools/Measurement.java
+++ b/tools/src/main/java/com/graphhopper/tools/Measurement.java
@@ -202,7 +202,10 @@ public class Measurement {
         StopWatch sw = new StopWatch().start();
         try {
             maxNode = g.getNodes();
-            generateRandomButValidPoints(hopper, preCalculatedPoints, new Random(seed));
+            // restricting the maximum snap distance is meant to prevent that it is overly likely that we draw query
+            // points from the 'dead' parts of the map (within the bounding box, but far away from the road network)
+            double maxSnapDistance = args.getDouble("measurement. max_snap_distance", 50_000);
+            generateRandomButValidPoints(hopper, preCalculatedPoints, maxSnapDistance, new Random(seed));
             final boolean runSlow = args.getBool("measurement.run_slow_routing", true);
             GHBitSet allowedEdges = printGraphDetails(g, vehicleStr);
             printMiscUnitPerfTests(g, encoder, count * 100, allowedEdges);
@@ -755,7 +758,7 @@ public class Measurement {
         print(prefix, miniPerf);
     }
 
-    private void generateRandomButValidPoints(GraphHopper hopper, int numPoints, Random rnd) {
+    private void generateRandomButValidPoints(GraphHopper hopper, int numPoints, double maxSnapDistance, Random rnd) {
         // create random but valid points to be used later (better than obtaining them during the measurement)
         StopWatch sw = new StopWatch().start();
         int numTries = 10 * numPoints;
@@ -771,7 +774,9 @@ public class Measurement {
             double lat = bBox.minLat + rnd.nextDouble() * (bBox.maxLat - bBox.minLat);
             double lon = bBox.minLon + rnd.nextDouble() * (bBox.maxLon - bBox.minLon);
             QueryResult qr = hopper.getLocationIndex().findClosest(lat, lon, edgeFilter);
-            if (qr.isValid()) {
+            if (qr.isValid() && DistancePlaneProjection.DIST_PLANE.calcDist(
+                    qr.getSnappedPoint().getLat(), qr.getSnappedPoint().getLon(),
+                    qr.getQueryPoint().getLat(), qr.getQueryPoint().getLon()) > maxSnapDistance) {
                 EdgeIterator iter = edgeExplorer.setBaseNode(qr.getClosestNode());
                 String hint = iter.next() ? iter.getName() : "";
                 randomPoints.add(new PointWithHint(new GHPoint(lat, lon), hint));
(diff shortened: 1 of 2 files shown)


easbar:
Oh nice, github auto-linked out previous comments here :+1: 
I just added a maximum snapping distance parameters to discard query points outside of the borders of the road network. Do you think 50km is a good default values?

tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -202,7 +202,10 @@ protected DataReader importData() throws IOException {
         StopWatch sw = new StopWatch().start();
         try {
             maxNode = g.getNodes();
-            generateRandomButValidPoints(hopper, preCalculatedPoints, new Random(seed));
+            // restricting the maximum snap distance is meant to prevent that it is overly likely that we draw query
+            // points from the 'dead' parts of the map (within the bounding box, but far away from the road network)
+            double maxSnapDistance = args.getDouble("measurement. max_snap_distance", 50_000);
```
karussell, line 207: note, there is a space in the key

karussell:
> Do you think 50km is a good default values?

I probably would use 500m if this does not increase the time it takes to collect these points as for points >1km we might even get a PointNotFoundException. (We should probably test lengthy snaps too, but in a different method?)

tools/src/main/java/com/graphhopper/tools/Measurement.java
easbar: you got the eagle's eyes ;)

easbar:
>  if this does not increase the time it takes to collect these points

that should not be a problem I did not try but the time this takes will be logged and I will check when its running

easbar:
Ok currently the problem with this PR is that it slows down the Measurement a lot. The reason seems to very slow LM routing and apparently the LM routing gets very slow when some of the (random) points snap onto a ferry connection (?) like this:

https://graphhopper.com/maps/?point=55.479581%2C14.767414&point=47.617199%2C12.06186&locale=en-us&vehicle=car&weighting=fastest&elevation=true&turn_costs=true&use_miles=false&layer=Omniscale

karussell:
Ah, this is interesting. Indeed if I remove the ferry part it is much faster

https://graphhopper.com/maps/?point=53.940389%2C10.862002&point=52.375599%2C11.689453&ch.disable=true (~200ms)

vs

https://graphhopper.com/maps/?point=55.479581%2C14.767414&point=52.375599%2C11.689453&ch.disable=true (~2600ms)

Could it be the rather slow speed of 28km/h over a long distance which causes the "to"-shortest-path-tree to grow before the finish condition is met?

easbar:
How did you measure the time? Yes using snap prevention should work, but of course 'hides' the fact that LM gets slow when snapping to ferry connections.

easbar:
But actually there is almost no need to measure its so much slower the delay can easily be noticed on GH maps

karussell:
> How did you measure the time?

Just in the browser log :D ... but oh ... it seems the took "switches" between ~2.7s and ~5.6s:

image

> Yes using snap prevention should work, but of course 'hides' the fact that LM gets slow when snapping to ferry connections.

It might also get much slower for e.g. destination-only areas as it is like a "hill" that is not considered for the other shortest-path-tree until it reaches the finish condition.

easbar:
This one is also really slow (and has no ferry snapping)
https://graphhopper.com/maps/?point=47.855474%2C7.732669&point=53.88226%2C13.281101&locale=en-US&vehicle=car&weighting=fastest&elevation=true&turn_costs=true&use_miles=false&layer=OpenStreetMap&ch.disable=true

easbar:
Actually I am seeing lots of timeouts on GH maps and now I think this might be related to the virtual edges and we only measured a much faster query time before, because the QueryGraph was 'skipped' as we set the start/target nodes to the node-IDs directly.

karussell:
Yes, certain long routes are still not optimal as we only have a limited set of landmarks for whole Europe (for car it is 24 I think). 

Please also note that (Dijkstra- and) LM-based requests have different behaviour for A to B depending on if the surrounding map is only Germany or including neighboring countries. So the response time for GH Maps should be worse than for our Measurement class.

> now I think this might be related to the virtual edges 

Hmmh, I do not think so. E.g. for the example you gave you can adjust start and end to land on the junction and it still times out. The main problem I see is that a lot of queries are started or ended on ferries and as the whole upper right and left corner of Germany will snap to them we get many of these lengthy queries.

karussell:
Sorry, this nice change seems to be unfortunately in an unmergable state now. Should we try to get this merged again?

For the histogram: it seems to be from Simon George - then likely the license header is wrong (it cannot be us) and we need to link to it from our NOTICE.md.

easbar:
The problem here was that for LM the measurement results strongly depend on the selection of the query points. Partly this is because of #2128 but even when we use snap_prevention to avoid query points on ferry routes there are still some outliers. I used the histogram to check this (actually I was not planning to keep this in the code). For CH we see a pretty nice Gaussian distribution by the way. I still think using GPS points here would be useful, but it will probably change the measurement results so they will be a bit harder to compare with results from the past. Then again it might be useful to investigate this a bit more to see why some routes take so much longer to calculate.

PR #2126, opened 2020-09-09
Add tolerance for Isochrones

boldtrn:
Calculating Isochrones for larger limits can take a lot of time. One major factor when calculating Isochrones is the triangulation.

If we accept less detailed results, the triangulation can be massively sped up. 

Currently the tolerance is passed into the triangulation without checking the input if it's a reasonable number. Not sure if we need something here? Also I found that the larger the limit, the higher tolerance can be without getting too coarse.

This can be especially helpful if you calculate big isochrones, like state or country size. Without speeding up the triangulation, this can take minutes. IMHO, for larger isochrones, details usually becomes less important, especially if you think about travel time, who knows how the traffic will be, so the isochrones are just an educated guess.

I will add an example below, an isochrone around Stuttgart. The detailed one is with tolerance=0 (the default), on my machine the calculation took 1.4s. The medium detailed one is with tolerance=0.002 and took 400ms. The less detailed on is with tolerance=0.005 and took 300ms. Obviously the result is worse, but there are applications that can live with less detail. For example if you use the time limit, but you don't know the traffic that well, so it's just a rough guess.

Screenshot from 2020-09-09 17-52-28
Screenshot from 2020-09-09 17-55-59
Screenshot from 2020-09-09 17-53-38


web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
```diff
@@ -141,18 +142,20 @@ public Response doGet(
...
-                isochrones.add(isochrone);
-            } else {
-                Polygon maxPolygon = heuristicallyFindMainConnectedComponent(isochrone, geometryFactory.createPoint(new Coordinate(point.get().lon, point.get().lat)));
-                isochrones.add(geometryFactory.createPolygon(((LinearRing) maxPolygon.getExteriorRing())));
+            if(!isochrone.isEmpty()){
+                if (fullGeometry) {
+                    isochrones.add(isochrone);
+                } else {
+                    Polygon maxPolygon = heuristicallyFindMainConnectedComponent(isochrone, geometryFactory.createPoint(new Coordinate(point.get().lon, point.get().lat)));
+                    isochrones.add(geometryFactory.createPolygon(((LinearRing) maxPolygon.getExteriorRing())));
```
boldtrn, line 157: For empty isochrones there was a null pointer exception here

web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
```diff
@@ -169,6 +172,7 @@ public Response doGet(
         }
         ObjectNode json = JsonNodeFactory.instance.objectNode();
 
+        sw.stop();
```
boldtrn, line 175: Took was always zero, because the sw wasn't stopped before reading it on line 183.

michaz:
Wow, interesting! Actually, I'm surprised that it looks as good as it does, thanks for sharing!

What is illustrated here is: The fewer points we put into the triangulation, the faster the computation. Allright.

However, if that is what we want, (speed up the computation by putting fewer points into the triangulation), there may be more straight-forward ways to do it.

What seems to happen when you use the tolerance value of the triangulator is this:
- Points (with travel time) are inserted in unspecified order
- If a point within radius `tolerance` was already inserted, the new point is dropped, including its position and travel time value. The first point just wins. Even if those two points are on different sides of a wall / river / lava stream.

(I think. Maybe points are merged in a more sophisticated way, but it doesn't look like it in JTS code.)

Another way of getting fewer points to triangulate is to comment out this block in JTSTriangulator:

```
            // add a pillar node to increase precision a bit for longer roads
            if (label.parent != null) {
                EdgeIteratorState edge = queryGraph.getEdgeIteratorState(label.edge, label.node);
                PointList innerPoints = edge.fetchWayGeometry(FetchMode.PILLAR_ONLY);
                if (innerPoints.getSize() > 0) {
                    int midIndex = innerPoints.getSize() / 2;
                    double lat2 = innerPoints.getLat(midIndex);
                    double lon2 = innerPoints.getLon(midIndex);
                    Coordinate site2 = new Coordinate(lon2, lat2);
                    site2.z = exploreValue;
                    sites.add(site2);
                }
            }
```

michaz:
(Also, the solution to the entire problem is to pre-compute the triangulation for the world. It isn't request-specific. Working on this.)

michaz:
Another idea: Drop nodes (don't put them in the triangulation at all) that are _most probably_ very far from the isochrone. Like, if I want a 1-hour-isochrone, I don't need to triangulate the entire disc, I only need nodes close to the isochrone (some inside, some outside). Tricky bit is finding out which ones I need.

michaz:
If we just put in the tolerance parameter, let's at least make it in meters and not in lat-lon.

boldtrn:
Thanks a lot for the detailed review @michaz.

> Also, the solution to the entire problem is to pre-compute the triangulation for the world. It isn't request-specific.

This actually sounds really cool, a preparation like CH for isochrones :+1:.

> If we just put in the tolerance parameter, let's at least make it in meters and not in latlongitude.

Yes that makes a lot of sense and is probably a lot easier for most users to handle and understand. How would you handle this internally? If a user passes for example 10m. Should we assume LatLong at the equator and use that for converting meter into latLong or should we use the starting point as reference? Or maybe there is a better way in JTS to do that?

> Another idea: Drop nodes (don't put them in the triangulation at all) that are most probably very far from the isochrone. Like, if I want a 1-hour-isochrone, I don't need to triangulate the entire disc, I only need nodes closely below 1-hour and closely over 1-hour, i.e. close to the isochrone (some inside, some outside). Tricky bit is finding out which ones I need.

This is a very interesting pointer. Finding these is probably quite tricky, yes. I was thinking if we could somehow use the ShortestPathTree for that to find points that are further up the tree that we could drop with a high possibility, but I think this is not easily possible. For example when searching for a Isochrone along a shoreline this might drop points that might still be relevant for the Isochrone.

michaz:
> I was thinking if we could somehow use the ShortestPathTree for that to find points that are further up the tree that we could drop with a high possibility, but I think this is not easily possible. For example when searching for a Isochrone along a shoreline this might drop points that might still be relevant for the Isochrone.

Yes, exactly. :-( No matter how you spin it, it's a "global" problem in the sense that you can never get everything you need from the SPT alone: In priciple, we would need to do even more work: Query the LocationIndex to get any points within the convex hull (or more?) of our point set to see if we missed any points that are far away on the road, but close on the map.

If you pre-compute the triangulation, that problem goes away.

michaz:
And *all* of this is only a problem if you want to *draw* the isochrone, in 2-D. If you only want to know where _on the graph_ the isochrone is, like, "where are the points on the road where I run out of time / gas", that's much easier (see SPT endpoint).

boldtrn:
Sorry for my late reply @michaz. Somehow the email must have gone under in my inbox.

You are right, having a global triangulation would be awesome and solve this issue for good.

In the meantime, WDYT about this PR? Is there something I should update?

boldtrn:
I update the PR and use the tolerance input in meters and convert it to coordinates. I had to use the distance² instead of the regular distance. This makes sense because we are calculating an area. WDYT?

michaz:
> I update the PR and use the tolerance input in meters and convert it to coordinates. I had to use the distance² instead of the regular distance. This makes sense because we are calculating an area. WDYT?

As discussed -- this is probably not quite right. Will merge and replace with a simple approximation.

michaz (approved):

Issue #1565, opened 2019-03-05
Speed up edge-based CH preparation.


easbar:
Goal would be to make it practical to use it for continental sized maps like Europe.

(See also #1810)

easbar:
I ran the measurement on `europe-140101.osm.pbf` using `Xmx/s29g`. 
|map|num nodes|num edges|preparation time|average query time|
|-|-|-|-|-|
|Germany|7.079.980|9.077.508|2.5h|26ms|
|Europe|38.307.787 (x5.4)|49.581.134 (x5.5)|10.4h (x4.2)|42ms (x1.6)|

The table also shows the relative increase of the different measures from Germany to Europe.
The scaling behavior is quite nice, although the size of the graph is about x5.5 the queries are less than x2 slower and also the import time increases less than proportional with the size of the graph.

karussell:
This sounds nice - thanks!

Do you also have the info regarding node based routing? Like preparation time and query time. Just for comparison on that hardware.

easbar:
> Do you also have the info regarding node based routing?

Will post once I have it.

easbar:
Here is the same measurement for node-based CH (no turn restrictions)

|map|num nodes|num edges|num shortcuts|preparation time|average query time|
|-|-|-|-|-|-|
|Germany|7.079.980|9.077.508|5.660.979 |12min|3ms|
|Europe|38.307.787 (x5.4)|49.581.134 (x5.5)|35.561.021 (x6.3)|1h12min (x6)|10ms (x3.3)|

For Europe the preparation for node-based is about 8.7x faster and the queries about 4x faster than for edge-based. The number of shortcuts is about x4 as high for edge-based.

easbar:
I think x4 shortcuts and query time is quite reasonable, but what probably could and should be improved is the preparation time.

karussell:
Thanks a lot!

> I think x4 shortcuts and query time is quite reasonable

Indeed, sounds good :)

easbar:
I ran the import for europe with the shortest weighting. The factors show the relation to the values I obtained with the fastest weighting above. The number of shortcuts increases by about 20-30% (for node&edge-based), and the import time increases drastically (x3.2/x4.6), two days for edge-based. Unfortunately I did not get the query times, because the `Measurement` class uses hard-coded  fastest weighting.

|map|num nodes|num edges|num shortcuts|preparation time|
|-|-|-|-|-|
[...]

PR #2132, opened 2020-09-15
Use temporary helper graph to speed up CH preparation

easbar:
This PR aims at reducing the CH preparation time by changing the contraction algorithm. The idea is pretty simple: We use a helper graph which at the beginning of the contraction is more or less a copy of the base graph. The witness searches are performed on this helper graph and the `CHGraph` is only used to store the resulting shortcuts. When we contract a node `x` with neighbors `p` and `q`: `p-x-q` we do not insert the p-q shortcut(s) into the CHGraph directly, but instead we add it to the helper graph. When the witness searches are finished we add the x-p and x-q edges to the `CHGraph` (if they are shortcuts). The p-q shortcut(s) are added only later when we contract node p/q. Importantly, we can also remove all edges adjacent to the node we are contracting from the helper graph after we have performed the witness search. This is the main reason why the preparation becomes faster using this method. So far measurements show that the preparation (node- and edge-based) on this branch is about two to three times faster than master (edge-based performance seems to benefit more from the changes in this branch, also because the turn costs are cached with the helper graph). 

The obvious downside of this approach is that we need additional memory for the helper graph. ~~As the helper graph becomes smaller and smaller during the preparation (while the CH graph becomes bigger and bigger) we need extra memory mostly at the beginning of the contraction~~. The size of the helper graph is initially proportional to the size of the base graph. First I thought there are 'as many' shortcuts as there are base graph edges and thus the memory size of the helper graph will be about equal to the memory we need to store the shortcuts. Unfortunately this is not exactly true for two reasons. First, since we are using 'bidirectional' shortcuts for node-based CH there are *less* shortcuts than base graph edges. Second, to support efficient removal of helper graph edges during the contraction we need to use a less memory efficient representation of the helper graph (this is also partly because of Java's wasteful way of storing heap data). Thus this PR increases the overall memory requirement for CH preparation (node- and edge-based).

I did some experiments on the Germany OSM for four (node-based) profiles (car,mtb,foot,bike) and was able to run the import with `Xmx7000m` (using `Xmx6500m` yielded an out of memory error) on master and needed `Xmx8500m` on this branch (using `Xmx8000m` yielded an out of memory error). The key thing to understand here is that the more profiles we prepare in parallel the more helper graph memory we need at the beginning of the import. So an obvious workaround would be using less threads. Another somewhat hacky possibility would be delaying the different preparations a bit such that the peak memory usages of the different threads do not happen at the same time. 

~~For edge-based CH we need a lot more memory than for node-based CH because there are more shortcuts, the shortcuts are all uni-directional and each shortcut needs more memory. Therefore the helper graph does not increase the overall memory usage (the helper graph memory is less than the memory we need to reserve for the shortcuts anyway). Actually I was able to speed up the preparation for edge-based even more when using a less memory efficient helper graph than the one I am using now.~~

Besides the speed up of the preparation using the approach of this PR has the following advantages:

1) The shortcuts are automatically sorted by the CH levels of their base nodes. This way we might be able to get rid of the `linkA/B` pointers for shortcuts (reduce the memory we need to store shortcuts) and the the shortcut storage is more cache-friendly. In fact the CH queries are faster on this branch.

2) After the shortcuts are written to the CHGraph they will not change again. So in principle it would be possible to start flushing some of the shortcut storage to disk already during the preparation (to reduce RAM usage).

3) Since we pre-calculate the base-edge weights and store them in the helper graph the preparation time is especially faster for 'expensive' weightings. I.e. if we use a custom weighting that does some costly edge-weight computations they are now done only once while the weightings get evaluated many many times during the witness searches on master. Obviously there would be other ways to 'cache' these weights, but with this PR this is already done and I checked that simply caching the weights does not nearly yield the speed up we see here.

Related #1565


core/src/main/java/com/graphhopper/routing/ch/CHPreparationGraph.java
```diff
@@ -0,0 +1,870 @@
...
+            return keyFwd;
+        }
+
+        @Override
+        public int getOrigEdgeKeyLastBA() {
+            int keyBwd = prepareEdge << 1;
+            if (nodeB > nodeA)
+                keyBwd += 1;
+            return keyBwd;
+        }
```
karussell, line 516: Why do we need those 4 methods? The methods getOrigEdgeKeyLastBA and getOrigEdgeKeyFirstBA
and then getOrigEdgeKeyFirstAB and getOrigEdgeKeyLastAB look identical and could we just omit them as this isn't used elsewhere? Also instead of
```
if (nodeA > nodeB) keyFwd += 1;
return keyFwd;
```

we could write:
```
return nodeA > nodeB? keyFwd + 1 : keyFwd;
```
but probably this is done from the JIT already

karussell: Ah, sorry. Forget this comment I just found another usage (EdgeBasedPrepareShortcut)...

core/src/main/java/com/graphhopper/routing/ch/CHPreparationGraph.java
```diff
@@ -0,0 +1,870 @@
...
+            OrigGraph build() {
+                int[] sortOrder = IndirectSort.mergesort(0, fromNodes.elementsCount, new IndirectComparator.AscendingIntComparator(fromNodes.buffer));
+                sortAndTrim(fromNodes, sortOrder);
+                sortAndTrim(toNodes, sortOrder);
+                sortAndTrim(edgesAndFlags, sortOrder);
+                return new OrigGraph(buildFirstEdgesByNode(), toNodes, edgesAndFlags);
+            }
+
+            private int getEdgeWithFlags(int edge, boolean fwd, boolean bwd) {
+                // we use only 30 bits for the edge Id and store two access flags along with the same int
```
karussell, line 757: This is a bit low. Can we avoid this reduction to 30 bits somehow?

easbar: Oh good point, I almost forgot about this. But 30bits should be enough, because this is only for base graph edges. And if they exceeded 30bits the edge-based preparation would fail anyway, because there are more than twice as many shortcuts than base edges. This is only used for edge-based.

core/src/main/java/com/graphhopper/routing/ch/AbstractNodeContractor.java
```diff
@@ -1,81 +0,0 @@
...
-    @Override
-    public void close() {
-        originalEdges.close();
-    }
-
-    boolean isContracted(int node) {
-        return prepareGraph.getLevel(node) != maxLevel;
-    }
-
-    void setOrigEdgeCount(int edgeId, int value) {
```
easbar, line 56: Thes orig edge counts are now included in the prepare graph's edges

core/src/main/java/com/graphhopper/routing/ch/CHParameters.java
```diff
@@ -12,7 +12,6 @@
     // node contraction, node-based
     public static final String EDGE_DIFFERENCE_WEIGHT = Parameters.CH.PREPARE + "node.edge_difference_weight";
     public static final String ORIGINAL_EDGE_COUNT_WEIGHT = Parameters.CH.PREPARE + "node.original_edge_count_weight";
-    public static final String CONTRACTED_NEIGHBORS_WEIGHT = Parameters.CH.PREPARE + "node.contracted_neighbors_weight";
```
easbar, line 15: I removed this heuristic for now, might be worth to re-implement it later, but not sure really

core/src/main/java/com/graphhopper/routing/ch/EdgeBasedWitnessPathSearcher.java
```diff
@@ -87,12 +85,11 @@
     // data structures used to build the shortest path tree
     // we allocate memory for all possible edge keys and keep track which ones have been discovered so far
     private double[] weights;
-    private int[] edges;
-    private int[] incEdges;
+    private int[] prepareEdges;
```
easbar, line 88: Here we got rid of one of the int[] arrays, i.et 4byte*N where N=2*number of base graph edges. We no longer need to store the incEdges, because we use the edge keys directly now and the index position equals the edge key for each element.

core/src/main/java/com/graphhopper/routing/ch/NodeBasedWitnessPathSearcher.java
```diff
@@ -30,41 +29,25 @@
...
     private static final int NOT_FOUND = -1;
-    private final PrepareCHGraph graph;
-    private final PrepareCHEdgeExplorer outEdgeExplorer;
+    private PrepareGraphEdgeExplorer outEdgeExplorer;
     private final IntArrayList changedNodes;
-    private final int maxLevel;
     private int maxVisitedNodes = Integer.MAX_VALUE;
     protected double[] weights;
-    private int[] parents;
-    private int[] edgeIds;
```
easbar, line 45: No need for these two arrays as we never need to find the actual shortest path (we just need to check its existence). This saved 2*4byte*N where N=number of nodes of the graph

core/src/main/java/com/graphhopper/routing/ch/PrepareContractionHierarchies.java
```diff
@@ -77,22 +76,18 @@ public static PrepareContractionHierarchies fromGraphHopperStorage(GraphHopperSt
...
         if (chConfig.getTraversalMode().isEdgeBased()) {
             TurnCostStorage turnCostStorage = chGraph.getBaseGraph().getTurnCostStorage();
             if (turnCostStorage == null) {
                 throw new IllegalArgumentException("For edge-based CH you need a turn cost storage");
             }
-            prepareGraph = PrepareCHGraph.edgeBased(chGraph, chConfig.getWeighting());
-            nodeContractor = new EdgeBasedNodeContractor(prepareGraph, pMap);
-        } else {
-            prepareGraph = PrepareCHGraph.nodeBased(chGraph, chConfig.getWeighting());
-            nodeContractor = new NodeBasedNodeContractor(prepareGraph, pMap);
```
easbar, line 95: Here we were allocating quite a bit of memory even when re-starting the server and just loading the preparations (no new preparation). We should really make PrepareContractionHierarchies about preparing CH and not use it as a 'handle' to the preparations available on the server...

core/src/main/java/com/graphhopper/routing/ch/PrepareEncoder.java
```diff
@@ -32,46 +32,21 @@
...
-     * There are two special cases when it returns 2.
-     */
-    public static final int getScMergeStatus(int existingScFlags, int newScFlags) {
-        if ((existingScFlags & scDirMask) == (newScFlags & scDirMask))
-            return 1;
-        else if ((newScFlags & scDirMask) == scDirMask)
-            return 2;
-
-        return 0;
-    }
```
easbar, line 76: This is no longer needed because after we wrote a shortcut to CHGraph we never change it again

core/src/main/java/com/graphhopper/routing/ch/SplitArray2D.java
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * This is a more memory-efficient version of two equal length `ArrayList<T>[]`s, i.e. this is a fixed size array where
+ * each element is a variable sized sub-array and each sub-array is divided into two parts.
+ * This is more memory efficient than two arrays of `ArrayList`s, because it saves the object-overhead of using an
+ * ArrayList object for each sub-array.
+ * <p>
+ * The elements in each sub-array are divided into two parts with ranges [0,mid[ and [mid,size[. We can add elements
+ * to either of the two parts.
+ */
+class SplitArray2D<T> {
```
easbar, line 32: Any idea for a better name? I am trying to say this is an array of arrays (2D) where each of the inner arrays is split into two parts (the elements of the two parts are separated from each other). But maybe not so important as this is only a package private class used by CHPreparationGraph...

core/src/main/java/com/graphhopper/routing/ch/CHPreparationGraph.java
easbar: Ok I forgot it already :) (but I made use of the ternary operator as its a bit more concise).

core/src/main/java/com/graphhopper/routing/ch/SplitArray2D.java
```diff
@@ -122,4 +122,17 @@ void clear(int n) {
         mids[n] = 0;
     }
 
+    private void grow(int n) {
+        // todo: think about grow factor: trimming and then doubling the size might be not what we want, rather
+        // increase by 50% or something?
```
karussell, line 127: How large are these sub-arrays on average? Could it make sense to initialize them with the average+20% then they do not need to grow that often and the size mismatch could be lower?

easbar: Ggood question. In this commit: https://github.com/graphhopper/graphhopper/pull/2132/commits/1c2ca0f9c421ff36b645c31fc2d66852717911cd I trimmed the arrays after the prepare graph is built for the first time to save some memory at the beginning of the contraction. For maximum speed it would of course be better to initialize them with some expected size (but no idea so far how important this is). I am currently running the import with this trimming to compare to the version without the trimming.

easbar: I did a quick check for Bayern (~6mio shortcuts) and Germany (~28mio shortcuts) and the array sizes seem to be relatively stable against changes of the map size. Before the contraction the average size of the arrays (number of in+out edges) is around 4.5 (makes sense: 2.25 in/out edges per node) and average mids[n] is almost exactly half of sizes[n] (on average the number of in edges is the same as out edges -> makes sense). 

The maximum size during the contraction is about 100 and the maximum mid is around 50 for both maps. 

So I think its already safe to say that we can store mid/size in a single int to save some (not a lot) memory (around 900MB for planet). For the resizing strategy I think it probably does not really matter, because the arrays never get very large and maybe its even the best approach to trim them as I do currently to make sure memory is minimized.

In my opinion if we change anything here we should try linking the edge objects to each other and remove the sub-arrays entirely, but I would postpone this until we feel like its definitely needed (and we are sure it improves anything).

karussell: > to trim them as I do currently to make sure memory is minimized.

Yes, this sounds reasonable

easbar:
This version shows even better performance for edge-based preparation, but takes even more memory: https://github.com/graphhopper/graphhopper/tree/prepare_graph_unidir 
I haven't really understood why its faster though. Just keeping this as a reminder here, the increased memory usage is def. not worth the performance gain atm.

easbar:
This is what it looks like in our measurements for Bayern:

image

and Germany (node-based only):

image

Note how the import time for the custom profiles is not bigger anymore, because all the weights get pre-calculated at the beginning of the preparation.

We also see a drop in the edge traversal time on the CH graph (due to better cache locality) and a related measurable reduction of the average query time (around 10%). 

image

Issue #2133, opened 2020-09-19
Add ability to parse OSM BoundingBox again


hegerdes:
**Add OSM BBox parsing**
It would be great if the OSMReader would support the OSM-Header-BoundingBox. Especially because osmosis lib supports this already. 

I know that you don't use the OSM BBox for the GH graph. The OSM BBox almost always has ways and relation that intercept the BBox and have nodes in- and outside the BBox. This can cause problems. But it would be nice if the Reader supports it for people who want to customize the Reader.

You already had the feature but disabled it: 
https://github.com/graphhopper/graphhopper/blob/c0aaffae8631efb8fa23c6ab5cf26e27027efd08/reader-osm/src/main/java/com/graphhopper/reader/osm/pbf/PbfBlobDecoder.java#L97-L116

I would like to reimplement this and add it to GH.

michaz:
What would you suggest should be GraphHopper‘s behavior when reading this bounding box? What should it do with it?

(I‘m not familiar with the commented-out code, so I don‘t know its history or original purpose..)

hegerdes:
For those how want to use GH as a library. Or for those how want hard Graph BB edges that are identical with the OSM BB. 

I know that the GH BB is calculated differently.
Couldn't find anything either to the old code. Just one old init commit

easbar:
> For those how want to use GH as a library. Or for those how want hard Graph BB edges that are identical with the OSM BB.

What do you mean? What are users who want to use GH as a library supposed to do with the OSM BB?

PR #2134, opened 2020-09-19
Added support for OSM BB Header to OSMReader

hegerdes:
The OSMReader is now able to use the OSM BBox Header.

I reimplemented and updated the old version that was already implemented but put in comments.
https://github.com/graphhopper/graphhopper/blob/c0aaffae8631efb8fa23c6ab5cf26e27027efd08/reader-osm/src/main/java/com/graphhopper/reader/osm/pbf/PbfBlobDecoder.java#L104-L112

This solvels #2133

### DIFF
--- a/reader-osm/pom.xml
+++ b/reader-osm/pom.xml
@@ -24,7 +24,7 @@
         <dependency>
             <groupId>org.openstreetmap.osmosis</groupId>
             <artifactId>osmosis-osm-binary</artifactId>
-            <version>0.47.3</version>
+            <version>0.48.3</version>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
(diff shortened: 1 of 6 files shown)


michaz:
I'm reluctant to have code around that produces information which is then not used. To be honest, if I came across that code, I would think "unused code" and remove it again. I would even think it might confuse users because it looks like it plays some role somewhere, which it doesn't.

Now I see what you meant by "use it as a library" -- good point, though I'm not sure we support e.g. extending `OSMReader` in your own code and then getting to the new element you're adding as library use. We just don't have it on our radar. (At least I don't..) It could just change back any time.

boldtrn:
Just a note: you could also use Osmosis or Osmconvert to cut the pbf file, which is probably easier than doing it in GraphHopper. For example here is the Osmconvert wiki page explaining how to do this: https://wiki.openstreetmap.org/wiki/Osmconvert#Keeping_Cross-Border_Ways_Complete

karussell:
It was indeed a feature at some point where the import was very memory intense and I did not know the existing tools ... unfortunately I did not fully remove the code it seems. For now I would indeed prefer to have the code removed as there seems to be no one with a need for it as it could be easily done externally.

PR #2004, opened 2020-04-18
Added docker push on every commit and tags

HarelM:
Related to #1985.
While dockerhub has this out-of-the-box it seems that there's an issue with configuring this properly there.
I have sent a support request but was not answered yet.
In the meantime, this should cover this. It can be removed if the dockerhub issue is resolved.

PR #2141, opened 2020-10-01
remove Dockerfile

karussell:
I won't have the resources to maintain a Dockerfile (see discussion e.g. in #2004) and I don't use it. As the image is already outdated for several months it is obvious that users are fine with our [current installation procedure](https://github.com/graphhopper/graphhopper/blob/master/docs/index.md#installation) that mainly consists of installing a JVM and then running our maintained graphhopper.sh bash script.

Feel free to point to your docker images here. Make sure you always mention the GraphHopper version that is used.

I also removed the pointer to the DEB or RPM packaging because also there no maintainer was found.


tomrod:
Great project and thanks for the work and dedication you clearly put in!

Sad to see the Dockerfile gone as this makes spinning a new build up very quick and consistent across Windows, Linux, and so on (what I'm facing now!).

Can the blog announcement be updated? [[link](https://www.graphhopper.com/blog/2018/09/25/getting-started-with-graphhopper-0-11-and-jdk-11/)]

PR #2143, opened 2020-10-01
move example code from docs into source code

### LINKED ISSUE #774: Update documentation
Update docs regarding new Java API (GraphHopperOSM, eclipse, etc)

Rough new structure of index.md:
## Get Started

Link to GraphHopper Maps with picture

Link to Dev setup & User web setup

Introductory slides & videos
## Web

web/quickstart.md
web/api-doc.md
core/deploy.md
## Configuration

explain roughly config.properties or link to source with all parameters?
elevation.md
## Translations

translations.md
## Developers

Maven snippet, snapshots, IDE setup

changelogs
### Technical

quickstart-from-source.md
technical.md 
routing.md
weighting.md
location-index.md
low-level-api.md
ch.md
create-new-flagencoder.md
### Android

Include picture

android/index.md
android/android-studio-setup.md
### Windows

windows-setup.md
### Eclipse

// hmmh I do not like it much that these docs are kind of a workaround where there has to be a better solution. Should we remove them completely?

eclipse-setup.md
eclipse-tomcat-setup.md

karussell:
This should fix #774 via creating a separate 'example' module with sources only for documentation purposes and at the same time it can be used as a start to create a Java project using GraphHopper through the Java library.

There were many incomplete examples (that probably converted into this state over the time and because they are not executable) and this is the main indication that we really need to do this when we want usable example Java code :)

TODOs:

 - [x] I was unsure if we should make it safe for refactoring via assertEquals or easy for beginners to copy and paste via System.out.println. But now I'm sure I need both and will e.g. comment all assertEquals with "// not necessary in real code".
 - [x] include isochrone


michaz:
>  I was unsure if we should make it safe for refactoring via assertEquals or easy for beginners to copy and paste via System.out.println. But now I'm sure I need both and will e.g. comment all assertEquals with "// not necessary in real code".

I would suggest not mixing "languages" here, or inventing your own conventions -- keep it simple, choose _one_ of two:

1. Write the examples as tests. Then assume that user knows what a test is and what assert means, and why they are there.
2. Write the examples as "main" code. Then don't put it in `src/test` but in `src/main`, and use _Java_'s `assert`, not JUnit's, and/or write tests for the examples.

karussell:
> Write the examples as "main" code. Then don't put them in src/test but in src/main,...
> then just run them from a one-line JUnit test.

Thanks, this is an interesting idea. Will think about which one makes more sense and also look into other open source projects how they do it. btw: do you know any with good examples?

karussell:
Will move the code into `main` and use your second approach. The projects that use this approach seem to have normal unit testing in place for the example app instead of asserts in the code. Will see if I can make it working this way, otherwise I will use `assert` like you proposed.

PR #2146, opened 2020-10-02
Refactor edge key

michaz:
This is how far I got. May be of use on its own (?) because I (think I) managed to get rid of one overloaded method in the CH class hierarchy which goes back and forth a lot..

What I did:

- Put `getEdgeKey()` on `EdgeIteratorState` and implement by calling `GHUtility.createEdgeKey`
- Replace most occurances of `GHUtility.createEdgeKey` by that call

What I would ultimately want to do:

- Replace remaining 3 occurances
- Remove it
- Replace the implementation (in `EdgeIteratorState`) by the one commented-out there. Because that is the _other_ notion of edge direction that we have, and I only want us to have one (the direction that was used to create the edge):

```
    default int getEdgeKey(boolean reverse) {
        return GHUtility.createEdgeKey(getBaseNode(), getAdjNode(), getEdge(), reverse);
        // What I really want to do is:
        // return getEdge() * 2 + (getBaseNode() != getAdjNode() && (get(EdgeIterator.REVERSE_STATE) ^ reverse) ? 0 : 1);
    }
```

What's the problem:

- I'm not sure how to get rid of the occurance in `CHPreparationGraph`:

```
        @Override
        public int getOrigEdgeKeyFirst() {
            int e = graph.edgesAndFlags.get(index);
            return GHUtility.createEdgeKey(node, getAdjNode(), e >> 2, false);
        }
```

- Particularly I don't know if we could just _inline_ the call there, and then the preparation uses a different traversal id than the rest of the system. Don't know if that matters, the important tests seem to pass, and it uses a different algorithm and different graph interface anyway, so it wouldn't even be wrong, technically. But maybe it does matter.

Feel free to ignore if not useful.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
+++ b/core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
@@ -27,6 +27,7 @@ import com.graphhopper.storage.RoutingCHEdgeIteratorState;
 import com.graphhopper.storage.RoutingCHGraph;
 import com.graphhopper.util.EdgeExplorer;
 import com.graphhopper.util.EdgeIterator;
+import com.graphhopper.util.EdgeIteratorState;
 import com.graphhopper.util.GHUtility;
 
 import static com.graphhopper.util.EdgeIterator.ANY_EDGE;
@@ -95,7 +96,7 @@ public abstract class AbstractBidirectionEdgeCHNoSOD extends AbstractBidirCHAlgo
                 : innerOutExplorer.setBaseNode(entry.adjNode);
         while (iter.next()) {
             final int edgeId = iter.getEdge();
-            int key = GHUtility.createEdgeKey(iter.getAdjNode(), iter.getBaseNode(), edgeId, !reverse);
+            int key = iter.getEdgeKey(reverse);
             SPTEntry entryOther = bestWeightMapOther.get(key);
             if (entryOther == null) {
                 continue;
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
```diff
@@ -131,8 +132,8 @@ protected int getIncomingEdge(SPTEntry entry) {
 
     @Override
     protected int getTraversalId(RoutingCHEdgeIteratorState edge, int origEdgeId, boolean reverse) {
-        int baseNode = getOtherNode(origEdgeId, edge.getAdjNode());
-        return GHUtility.createEdgeKey(baseNode, edge.getAdjNode(), origEdgeId, reverse);
+        EdgeIteratorState origEdge = graph.getBaseGraph().getEdgeIteratorState(origEdgeId, edge.getAdjNode());
+        return origEdge.getEdgeKey(reverse);
```
easbar, line 136: Using `getOtherNode` is better for performance here, I agree these `getTraversalId` methods are a bit twisted any hopefully this can be improved.

easbar (commented):
Thanks for this! It seems this is mostly about getting rid of `createEdgeKey(base, adj, edge, reverse)`. I guess we can do this on top of #2150, so let's come back here once #2150 is merged?

core/src/main/java/com/graphhopper/routing/AbstractBidirectionEdgeCHNoSOD.java
easbar: Then again we should not need this at all and the edge key should be known already. But then we need to store the first/last orig edge key (not id) for the edge-based CH shortcuts (but maybe this is better anyway...).

easbar:
For now we added  `graph.getEdgeIteratorStateForKey(edgeKey)` and `edgeState.getEdgeKey()`. Getting rid of `GHUtility.createEdgeKey(base, adj, edge, reverse)` would be nice, but rather low priority for now, see: https://github.com/graphhopper/graphhopper/issues/2152

PR #2147, opened 2020-10-03
Details (gh:mtb_scale, gh:surface and gh:road_class) are added to trackpoint extensions instead of route.

joe-akeem:


joe-akeem:
Wrong target repo 🙄

PR #2150, opened 2020-10-05
Add Graph#getEdgeIteratorStateForKey(edgeKey) and edgeState.getEdgeKey()

easbar:
This PR adds two new methods: 

* `EdgeIteratorState#getEdgeKey()` can be used to get an edge key from an edge state. The edge key is a combination of the edge ID and the direction of the edge, so its a unique identifier of a (directed) edge. It is equal to `2*edgeId` if the edge state matches the direction in which the edge is stored in BaseGraph and equal to `2*edgeId+1` otherwise.

* `Graph#getEdgeIteratorStateForKey(edgeKey) can be used to retrieve an edge state given an edge key.

Compared to #2146 this PR does not try to get rid of `GHUtility#createEdgeKey(int, int, int, boolean)` yet, but this should be at least easier to do after this PR.


core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
```diff
@@ -73,6 +74,16 @@ public void setBool(boolean reverse, IntsRef ref, boolean value) {
...
 
+    /**
+     * Returns the edge key of the current edge. The edge id can be derived from the edge key by calling
+     * {@link GHUtility#getEdgeFromEdgeKey(int)}, but the edge key also contains information about the
+     * direction of the edge. The edge key is even when the edge is oriented in storage direction and odd
+     * otherwise. You can use the edge key to retrieve an edge state in the associated direction using
+     * {@link Graph#getEdgeIteratorStateForKey(int)}. Loop edges are always returned in 'forward' direction even when
+     * you use an odd edge key.
+     */
+    int getEdgeKey();
```
easbar, line 85: In #2146 we call this `int getEdgeKey(boolean reverse)`. I would argue that the edge state has only one edge key and reversing the key is out of its scope, so I left out the reverse parameter. The reverse key can be obtained using `GHUtility.getReverseKey(edgeState.getEdgeKey())`, which of course is a bit less comfortable, especially in cases like this:

```java
int key = edgeState.getEdgeKey();
if (reverse)
    key = GHUtility.getReverseKey(key);
```

compared to:
```
int key = edgeState.getEdgeKey(reverse);
```

So not sure...

reader-gtfs/src/main/java/com/graphhopper/gtfs/WrapperGraph.java
```diff
@@ -112,6 +112,12 @@ public EdgeIteratorState getEdgeIteratorState(int edgeId, int adjNode) {
         }
     }
 
+    @Override
+    public EdgeIteratorState getEdgeIteratorStateForKey(int edgeKey) {
+        // todonow?
```
easbar, line 117: @michaz, can you check `WrapperGraph` here? I was not sure what is and what is not needed here.

core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
michaz: Oh yeah, excellent. I‘m not a fan of the (boolean reverse) parameter anywhere, I just kept to the convention. Let‘s drop it.

easbar: Ok so we keep `edgeState.getEdgeKey()` here :+1:

core/src/main/java/com/graphhopper/storage/Graph.java
```diff
@@ -82,6 +82,13 @@
      */
     EdgeIteratorState getEdgeIteratorState(int edgeId, int adjNode);
 
+    /**
+     * Returns the edge state for the given edge key
+     *
+     * @see EdgeIteratorState#getEdgeKey()
+     */
+    EdgeIteratorState getEdgeIteratorStateForKey(int edgeKey);
```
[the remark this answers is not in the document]
easbar: nope

reader-gtfs/src/main/java/com/graphhopper/gtfs/WrapperGraph.java
easbar: postponed until we need this method

PR #2151, opened 2020-10-06
Voice Instruction for Routing Requests

benedikt-brandtner-bikemap:
# Adds voice instructions to routing requests
This PR adds optional voice instructions to the `/route` and `/route-custom` endpoints via the`voice_instructions` query-param/post-data. The instructions are calculated exactly in the same way as for the `/navigate` endpoint and most of the logic has just been moved. The instructions are only calculated and added to the response if the parameter is provided.
## Why
Because the response of the `/navigate` endpoint is formatted differently than the routing responses and we want to implement more detailed voice instructions for our mobile clients.
Additionally most routing options are hardcoded for this endpoint and we needed more flexibility (eg.: disabled contraction hierachies).
## How
Two new classes have been introduced, `VoiceInstruction` and `VoiceInstructionList`, which are a representation of a single and multiple voice instructions. The `Instruction` class got a new field `voiceInstructions` which keeps track of the voice instructions relating to an instruction. Additionally all the logic previously contained in `NavigateResponseConverter.putVoiceInstructions` has been moved to `Instruction.setVoiceInstructions` and `NavigateResponseConverter.getThenVoiceInstructionpart` has been moved to `Instruction.getThenVoiceInstructionpart`. The generation of voice instructions happens within these two functions and should be exactly the same as before. 
The generation of voice instruction get triggered in `PathMerger.updateInstructionsWithContext` and i pass the state of the `voice_instructions` query-param/post-data in `Router.createPathMerger` in the same way as the `instructions` query-param/post-data.
## Changes

- Introduced `VoiceInstruction` and `VoiceInstructionList` classes
- Moved logic of `NavigateResponseConverter.putVoiceInstructions` to `Instruction.setVoiceInstructions`
- Moved `NavigateResponseConverter.getThenVoiceInstructionpart` to `Instruction.getThenVoiceInstructionpart`
- Moved graphhopper-navigation `VoiceInstructionConfig` to graphhopper-api
- Renamed graphhopper-navigation `DistanceConfig` to `VoiceInstructionDistanceConfig` and moved it to graphhopper-api
- Renamed graphhopper-navigation `DistanceUtils` to `VoiceInstructionDistanceUtils` and moved it to graphhopper-api
- Moved graphhopper-navigation `VoiceInstructionConfigTest` to graphhopper-core (would fit better in graphhopper-api but is dependend on `TranslationMap` which would trigger a circular dependency)
- Changed constructors of `VoiceInstructionConfig` and child-classes to accept a `Translation` instead of a `TranslationMap` and `Locale`
- Changed constructor of `VoiceInstructionDistanceConfig` accept a `Translation` instead of a `TranslationMap` and `Locale`
- Changed signature of `NavigateResource.calcRoute` to accept a `voiceInstructions` and `voiceUnit` parameter
- Removed `VoiceInstructionDistanceConfig` parameter from `NavigateResponseConverter.convertFromGHResponse` and subsequent functions
- Refactored `NavigateResponseConverter.putVoiceInstructions` to add the previously generated voice instructions to the response
- Added `testCustomWeightingVoiceInstructionsJson` and `testCustomWeightingVoiceInstructions` as well as extended `testCustomWeightingJson` and `testCustomWeighting` in `CustomWeightingRouteResourceLMTest`.
- Added `testVoiceInstructionsQuery` and `testVoiceInstructionsPostQuery` as well as extended `testBasicQuery` and `testBasicPostQuery` in `RouteResourceTest`

## Finishing Thoughts
I hope this feature is something that aligns with your plans and these changes would be acceptable for you. 
Also if anything could be improved please let me know.

greets

### DIFF
--- a/api/src/main/java/com/graphhopper/util/Instruction.java
+++ b/api/src/main/java/com/graphhopper/util/Instruction.java
@@ -18,6 +18,7 @@
 package com.graphhopper.util;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 public class Instruction {
@@ -42,6 +43,7 @@ public class Instruction {
     public static final int PT_START_TRIP = 101;
     public static final int PT_TRANSFER = 102;
     public static final int PT_END_TRIP = 103;
+    private static final int VOICE_INSTRUCTION_MERGE_TRESHHOLD = 100;
     protected PointList points;
     protected final InstructionAnnotation annotation;
     protected boolean rawName;
@@ -49,6 +51,7 @@ public class Instruction {
     protected String name;
     protected double distance;
     protected long time;
+    protected VoiceInstructionList voiceInstructions;
     protected Map<String, Object> extraInfo = new HashMap<>(3);
 
     /**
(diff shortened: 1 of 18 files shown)


karussell:
I do not see a reason to add voice instructions to all the endpoints. We only support two things:

1. a powerful /route endpoint so that you could create voice instructions on the client side. If not everything is possible, then we could add the required details e.g. via path details or tuned instructions.
2. the /navigate endpoint so that it can be consumed directly with [our open source navigation client](https://github.com/graphhopper/graphhopper-navigation-example). We won't support a third option.

karussell:
The only possibility for yet another format would be a better supported open source navigation SDK (that allows user to create a navigation app like we already have) but this would then replace the current /navigate endpoint and wouldn't be included in the /route endpoint. Is your mobile client open source and more active?

benedikt-brandtner-bikemap:
Thanks for your quick response and sorry for the delay i had other things to work on...

_a powerful /route endpoint so that you could create voice instructions on the client side. If not everything is possible, then we could add the required details e.g. via path details or tuned instructions._

Maybe i didn't state it clearly enough before but it did implement the voice instructions as tuned instructions and the reasoning was not having to do this on each client but having it optionally included in the instructions list of the routing response. But i also get that this could be done on the client and therefore no changes would be required to the `/route` endpoints response.

Unfortunately there are currently no plans to release a navigation sdk from our side due to limitations in resources.

boldtrn:
I think right now, returning the voice instructions does not really add value to the route endpoint - please correct me if I am wrong. This contains a few fixed distances at which an instruction is spoken. Deciding when to speak an instruction is probably very domain specific. For example for a hiking app it might not make sense to get voice instructions 2 km before the turn.

That said I think there are quite a few things that could be done on the route endpoint to improve voice guidance.

One example would be to recognize when there are no other turn options, so that a client can safely say: "Turn right on the next intersection" instead of "turn right in 300 m". Other examples might be recognizing more complex intersections like T-intersections or turn lanes, for which we could either return special markup on the instructions or provide special voice instructions.

PR #2154, opened 2020-10-11
Investigate Baremaps API

michaz:
(Making a Draft PR from this so I can track it more easily)


bchapuis:
Hi @michaz, all the tests are passing locally, but they are failing on travis and appveyor are failing. Any idea on what may cause this issue?

otbutz:
> Hi @michaz, all the tests are passing locally, but they are failing on travis and appveyor. Any idea on what may cause this issue?

```
Failed to execute goal on project graphhopper-reader-osm: Could not resolve dependencies for project com.graphhopper:graphhopper-reader-osm:jar:3.0-SNAPSHOT: Could not find artifact com.baremaps:baremaps-osm:jar:0.3.0 in central (https://repo.maven.apache.org/maven2)
```

Did you already publish that version?

bchapuis:
> Did you already publish that version?

Yes, I published it on the [Github Package registry](https://github.com/baremaps/baremaps/packages/149093). I somewhat believed that Github Package was public, but it seems that it requires some sort of [authentication](https://docs.github.com/en/free-pro-team@latest/packages/using-github-packages-with-your-projects-ecosystem/configuring-apache-maven-for-use-with-github-packages#authenticating-to-github-packages). 

Thanks a lot for your help, I should probably start publishing these artifacts on maven central.

otbutz:
The [Maven Super POM](https://maven.apache.org/ref/3-LATEST/maven-model-builder/super-pom.html) only comes with Maven Central by default. So that would make this IMHO a requirement if we want to use it in GH.

bchapuis:
@otbutz @michaz Thanks a lot for the tips! I published the artifacts on maven central and the builds are now passing. 

Notice that I had to increase the memory in the maven-surfire-plugin a bit. This is due to the way the pbf reader is implemented, it reads several OSM file blocks and decompress them in parallel, which improves performance to the detriment of memory usage. However, I think that the overall memory consumption when reading large files should improve as the producer's speed is now bound to the consumer's speed.

```
<argLine>-Xmx256m -Xms256m -Duser.language=en</argLine>
```

PR #2155, opened 2020-10-12
Extend /route endpoint to also return path details in GPX format

joe-akeem:
The `/route` endpoint allows to return path details in the JSON response by adding the `details` query parameter (e.g. `details=road_class`). Unfortunately the details were omitted so far in the GPX response (query parameter `type=gpx`).

This PR adds the path details to the `<extensions>` element of each `<trkpt>` element. Obviously the values apply to the segment between two track points and not to the points themselves. When reading the path details of a track point they need to be interpreted as the path details of the segment between this and the the previous track point.

### DIFF
--- a/core/src/main/java/com/graphhopper/util/gpx/GpxFromInstructions.java
+++ b/core/src/main/java/com/graphhopper/util/gpx/GpxFromInstructions.java
@@ -19,14 +19,13 @@
 package com.graphhopper.util.gpx;
 
 import com.graphhopper.util.*;
+import com.graphhopper.util.details.PathDetail;
 import com.graphhopper.util.shapes.GHPoint3D;
 
 import java.text.DateFormat;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Locale;
+import java.util.*;
 
 // todo: the code here does not really belong into core, but we moved it here for now so its available from
 // map-matching resource (it cannot be in the api module, because it uses AngleCalc). Probably we should separate the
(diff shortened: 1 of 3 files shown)


karussell:
Sorry for the late feedback. 

I do not think that we should not expand the features for the GPX format because either the client is powerful enough to interpret our JSON or the client only knows the standard GPX features. Or why can't you use the JSON?

joe-akeem:
I'd like to display the response of the call to `/route` on a Mapbox GL JS map. For that it should be in GeoJSON format. The GPX response can be easily converted into GeoJSON on the client e.g. by using https://github.com/mapbox/togeojson. At the same time I would like to display some information such as `surface` and `mtb_scale` (see #2158) which I currently only get in the Graphhopper JSON response. So with the current solution I guess I would have to call the `/route` endpoint twice to get all information that I need? Or is there some library to convert the Graphhoper JSON response to GeoJson?

An alternative could be to add `type=geojson` to the endpoint and get geojson from Graphhopper directly 🤔 

I have a working prototype at https://www.singletrail-map.eu/, based on a Graphhopper fork with the changes of this PR.

What do you think?

karussell:
> So with the current solution I guess I would have to call the /route endpoint twice to get all information that I need? Or is there some library to convert the Graphhoper JSON response to GeoJson?

Our JS client already converts the resonse into a GeoJSON (without the path details).

> An alternative could be to add type=geojson to the endpoint and get geojson from Graphhopper directly thinking

When you use points_encoded=false you can already easily add it as a geojson layer (again without the path details) or just use our JS client. 

And with the path details I think this should be done on the client side (at least for now) as the data can be massive especially for longer routes.

joe-akeem:
Sounds good! I didn't know about the JS client and that it can convert to GeoJSON. I think with that I can achieve what I need with just one call to the routing API 👍

PR #2157, opened 2020-10-16
MiniPerfTest: no subclassing, inject lambda instead

### LINKED ISSUE #2056: Measurement results are not stable against changes in Measurement class
There seems to be some strange behavior of our `Measurement` class as for example adding via-routing measurements here: 9eb65586ebb55aca603e1aba4788d16581a678ec changed the measured time for `routingCH(_edge)_full` even though this measurement runs *before* the new via-routing measurement. Ideally the performance / JVM optimizations (?) should not alter the measurement results like this when adding or skipping parts of the measurements.

A similar problem arose when I tried adding some more edge-traversal tests, see the `potential_jdk_issue` branch...

Another thing I was not able to explain was that in this commit: 6f07e5ce7ed0b72d35e3f38a182cdd1e2ef0788a the `routingCH_alt.mean` time increased by about 15% (I was not able to reproduce this consistently and I was wondering if the reason I could not reproduce it was that the slow down depends on some *other* code running before the test...).

karussell:
Let's see if this could fix #2056.

(Not sure about the name `MeasurementUnit`. Maybe just `Unit` or `UnderMeasurement`?)

### DIFF
--- a/core/src/main/java/com/graphhopper/util/MiniPerfTest.java
+++ b/core/src/main/java/com/graphhopper/util/MiniPerfTest.java
@@ -27,12 +27,12 @@ import java.util.Locale;
 /**
  * @author Peter Karich
  */
-public abstract class MiniPerfTest {
+public class MiniPerfTest {
 
-    private static final double NS_PER_S  = 1e9;
+    private static final double NS_PER_S = 1e9;
     private static final double NS_PER_MS = 1e6;
     private static final double NS_PER_US = 1e3;
-    
+
     protected Logger logger = LoggerFactory.getLogger(getClass());
     private int counts = 100;
     private long fullTime = 0;
@@ -40,15 +40,15 @@ public abstract class MiniPerfTest {
     private long min = Long.MAX_VALUE;
     private int dummySum;
 
-    public MiniPerfTest start() {
+    public MiniPerfTest start(MeasurementUnit m) {
         int warmupCount = Math.max(1, counts / 3);
         for (int i = 0; i < warmupCount; i++) {
-            dummySum += doCalc(true, i);
+            dummySum += m.doCalc(true, i);
         }
         long startFull = System.nanoTime();
         for (int i = 0; i < counts; i++) {
             long start = System.nanoTime();
-            dummySum += doCalc(false, i);
+            dummySum += m.doCalc(false, i);
             long time = System.nanoTime() - start;
             if (time < min)
                 min = time;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/graphhopper/util/MiniPerfTest.java
```diff
@@ -61,6 +61,15 @@ public MiniPerfTest start() {
         return this;
     }
 
+    public interface MeasurementUnit {
```
otbutz, line 64: Add `@FunctionalInterface` if it's only intended to be used as a lamba

karussell:
Unfortunately this does not seem to fix the perf problem :/

easbar:
> Not sure about the name

`Task`?

karussell:
Ok. I just tried with the [latest JDK16 build](https://github.com/AdoptOpenJDK/openjdk16-binaries/releases/tag/jdk-2020-10-16-06-40) where it is supposed to be fixed but it even got worse: instead of +7% I now see 15% ...

easbar:
Did you try

```java
MiniPerfTest miniPerf = new MiniPerfTest().setIterations(count).start(this::myTask);

// ...

private int myTask(boolean warmup, int run) {
            double lat = rand.nextDouble() * latDelta + bbox.minLat;
            double lon = rand.nextDouble() * lonDelta + bbox.minLon;
            int val = idx.findClosest(lat, lon, EdgeFilter.ALL_EDGES).getClosestNode();
            return val;
}

```

instead of 
```java
MiniPerfTest miniPerf = new MiniPerfTest().setIterations(count).start((warmup, run) -> {
            double lat = rand.nextDouble() * latDelta + bbox.minLat;
            double lon = rand.nextDouble() * lonDelta + bbox.minLon;
            int val = idx.findClosest(lat, lon, EdgeFilter.ALL_EDGES).getClosestNode();
            return val;
        });
```

No idea if this makes a difference, but the second case with the lamdba might be just another syntax for an anonymous class while the first is actually referencing a method.. (just having a guess here)

karussell:
> the second case with the lamdba might be just another syntax for an anonymous class while the first is actually referencing a method.. (just having a guess here)

I stumbled over contradicting statements, but it seems to be [implementation specific](http://wiki.jvmlangsummit.com/images/1/1e/2011_Goetz_Lambda.pdf). I.e. a lambda can be an anonymous class or a method handle or similar. Still it is likely a private static method created at runtime for recent JDKs.

E.g. When I try to print getClass() inside the method I only get the special class name ala "Test$1" for the anonymous class not for the lambda or for the method reference.

I think we should try to migrate to JMH somehow to avoid non-standard surprises and/or wait that this is fixed in the JVM instead of investing more time here.

easbar:
> I think we should try to migrate to JMH somehow

Ok, I think this PR is a good step into this direction. The methods we now pass into MiniPerfTest will be the annotated benchmark methods in JMH later.

PR #2158, opened 2020-10-17
Details support for mtb_scale, mtb_uphill_scale, sac_scale and horse_scale

joe-akeem:
Adds support for mtb:scale, mtb:scale:uphill, sac_scale and horse_scale as enum encoded details.

This PR replaces #1812 since that one hasn't been worked on for a long time and I couldn't get permission for the repository to finish it.

If the implementation is accepted like this in general, I will add a few tests too!


karussell:
Thanks for your contribution. My biggest problem with the PR before was the naming and unit. IMO it is confusing to name something xy_scale and then have different conventions and names (HorseScale:NONE,DEMANDING,..., MtbScale:NONE, MTB_0,..., SacScale: NONE, HIKING, ...). I know it is done in OSM but encoded values are not 1:1 copies of OSM tags in general because otherwise e.g. users of the encoded values would have to convert strings like from `maxspeed` into numbers and similar issues.

The encoded values that we store should be thought via its usage. E.g. when it comes to the CustomWeighting this would be simpler:

```yml
priority:
  some_scale:
    "< 5": 0.5
```

than:

```yml
priority:
  some_scale:
    none: 0.5
    name1: 0.5
    name2: 0.5
    name2: 0.5
```

boldtrn:
We could use -1 for "none" in this case?

joe-akeem:
Not quite sure if I understood you right @karussell 🤔 Is it just about the names? Would it be better to name them `mtb_rating` and `sac_rating` or similar when using `EnumEncodedValue` instead of `IntEncodedValue`?

I understand that it would be better to use `IntEncodedValue` instead of `EnumEncodedValue` because it is easier to compare e.g. with the `<` operator as pointed out above. However, I encountered a problem with using `IntEncodedValue`:

* the default of an `IntEncodedValue` is always `0`. So if the tag is missing, we would automatically end up with  `mtb_scale=0` in Graphhopper. (That's why unfortunately `-1` is not working for "none" @boldtrn)
* According to https://wiki.openstreetmap.org/wiki/Key:mtb:scale however, `0` is the lowest possible MTB Scale which is not the same as "none"!

Of course we could have another interpretation for the OSM `mtb:scale` tag in Graphopper like
* `0` means "none" in Graphhopper while in OSM it means "Difficulty S0"
* `1` means "S0" in  Graphhopper while in OSM it means "Difficulty S1"
* ...

This however I would find very confusing since I would always also look at https://wiki.openstreetmap.org/wiki/Key:mtb:scale to get an idea of how to interpret the values.

The problem only exists with `mtb:scale` and `mtb:scale:uphill` btw. since for them `0` is a valid value in OSM. `sac:scale` starts with "T1" which could be mapped to `IntEncodedValue` value `1` easily and `0` could be used for "none". `horse_scale` is a different story again because it doesn't use numbers in OSM. So we could leave this as an `EnumEncodedValue` or assign some numbers to the enums. This would be less confusing since it is then totally different from OSM.

I'm happy to change the PR but I couldn't think of a better solution that is not confusing to the user when dealing with `mtb_scale` as described above. Any proposals?

karussell:
> Would it be better to name them mtb_rating and sac_rating or similar when using EnumEncodedValue instead of IntEncodedValue?

One problem I have is naming, yes. All are named "xy_scale" although they mean slightly different things. So we could give them a different name or use the same values for all (I prefer this but names and values would then be different to OSM although we would preserve the meaning).

> So if the tag is missing, we would automatically end up with mtb_scale=0 in Graphhopper

A missing tag could be detected in the parser and set to any value. (Only for custom scenarios where the graph is used without OSMParsers the default value of 0 would kick in and could make problems.)

> This however I would find very confusing since I would always also look at https://wiki.openstreetmap.org/wiki/Key:mtb:scale to get an idea of how to interpret the values.

If the meaning is changed we must rename the encoded value IMO.

I think we should try to find a more consistent handling, so that our custom weighting can be simpler and more intuitive. E.g. we could define 0 means "missing", and then a value from 1 to 6? Then we can call every encoded values the same like `xy_rating` and nothing should be mixed up with OSM. 

If enums are better or more expressive this would work too (also due to `some_enum.ordinal()`). Here I would still favor to use the same enum for all (mtb, sac, horse etc) which we can name `Rating` to make it clear that it is different to OSM.

See e.g. the RouteNetwork enum that is used for biking and hiking routes and it is ordered from biggest to smallest route.

> horse_scale is a different story again because it doesn't use numbers in OSM

Sure, but is this different for historical reasons or does the difference has a real benefit?

> assign some numbers to the enums.

btw: in Java every enum already has a number (`some_enum.ordinal()`).

boldtrn:
> the default of an IntEncodedValue is always 0. So if the tag is missing, we would automatically end up with mtb_scale=0 in Graphhopper. (That's why unfortunately -1 is not working for "none" @boldtrn)

If one sets the value for every edge, this could work, but yes, that not really nice.

joe-akeem:
I renamed the enums from `*Scale` to `*Rating` and made the values more uniform. They are also parsed to `int` so they can be weighed with a range as described above.

karussell:
Thanks! 

I've tried a common Rating enum, see here https://github.com/graphhopper/graphhopper/pull/2158/commits/4070658a17e40342c278683987c36e79724c086f and then replaced this with an int which I currently prefer: https://github.com/graphhopper/graphhopper/pull/2158/commits/94a34bfd2eaf0ea32cd12396dedb29067567501b

Do you think this could work too?

BTW: I removed MtbUphillRating for now as we do not have elevation changes at the moment which would be necessary for it to be useful. So this would require some more work in a separate PR.

joe-akeem:
> Do you think this could work too?

Sure 👍 Although one thing I learnt while playing with this yesterday is that it currently seems not possible to have multiple ranges with `int`. So I think I couldn't set up a profile for my mountain biking preferences which are "I really want everything S2 and below, S3 is OK and everything above is too hard and should be avoided" 🤔 However, it seems like this would work and that should be enough for now:
``` yaml
priority:
  some_scale:
    "< 3": 1
    "*": 0.5
```

karussell:
> Sure 

:+1: 

> However, it seems like this would work and that should be enough for now

Yes, this is the way this currently works. However, if everything goes well, we will have a bigger (breaking) change for the custom weighting soon that would allow all boolean and comparison operators.

karussell:
@joe-akeem sorry, I forgot to ask. Can you [send us](https://www.graphhopper.com/contact-form/) a signed CLA or just a note so that we can contact you via Email and you electronically sign it? See [here](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md#license-agreement) for the details. We can then also add you as a contributor in CONTRIBUTORS.md.

easbar:
> I removed MtbUphillRating for now as we do not have elevation changes at the moment which would be necessary for it to be useful

Elevation changes are not strictly necessary for the `mtb:scale:uphill` tag. It is just a number 0-5: https://wiki.openstreetmap.org/wiki/Key:mtb:scale?uselang=en#mtb:scale:uphill.3D0-5 and the wiki encourages mappers to always use it in combination with the `incline` tag, which would tell us which direction of an edge  the 'uphill' tag is referring to.

PR #2159, opened 2020-10-19
TurnRestrictions use exceptions based on restrictions in FlagEncoder

boldtrn:
This PR improves the generation of TurnRestrictions.

So far, turn restrictions except parsing worked based on the name of the FlagEncoder. If the name did not match a given set of hard coded values, we assumed it's a motor vehicle and added these tags.

This PR uses the restrictions that are set in the FlagEncoder, this makes sure we actually use the right restrictions and also improves the separation of concerns, because the OSMTurnRelationParser should not have any knowledge about specific vehicle access rules.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
@@ -25,7 +25,8 @@ import com.graphhopper.reader.osm.conditional.DateRangeParser;
 import com.graphhopper.routing.ev.*;
 import com.graphhopper.routing.util.parsers.helpers.OSMValueExtractor;
 import com.graphhopper.storage.IntsRef;
-import com.graphhopper.util.*;
+import com.graphhopper.util.BitUtil;
+import com.graphhopper.util.EdgeIteratorState;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -237,7 +238,7 @@ public abstract class AbstractFlagEncoder implements FlagEncoder {
 
         return maxSpeed;
     }
-    
+
     /**
      * @return <i>true</i> if the given speed is not {@link Double#NaN}
      */
@@ -442,5 +443,10 @@ public abstract class AbstractFlagEncoder implements FlagEncoder {
     @Override
     public boolean hasEncodedValue(String key) {
         return encodedValueLookup.hasEncodedValue(key);
+    }
+
+    @Override
+    public Collection<String> getRestrictions() {
+        return restrictions;
     }
 }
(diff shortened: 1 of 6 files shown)


karussell:
What is your use case? Something country-dependent?

I would like to shrink the FlagEncoder interface until it dissolves e.g. PT routing is already working without it.
I'm a bit unsure about your use case but would [my proposal](https://github.com/graphhopper/graphhopper/compare/tra_mo_refactoring) be sufficient?

boldtrn:
This is already a lot better, thanks for having a look :+1:. My use case was a custom encoder that uses a custom name, that doesn't contain any of the predefined "words". 

So my particular use case would be fixed by your proposal. 

However, I think any profile (to not use the word FlagEncoder ;)) should be able to register a custom set of restrictions and not rely on a predefined set.

boldtrn:
A minor note regarding your proposal: Why rename bicycle to bike? Bicycle is the corresponding OSM tag and all other values use the corresponding OSM tag.

karussell:
> However, I think any profile (to not use the word FlagEncoder ;)) should be able to register a custom set of restrictions and not rely on a predefined set

In the future we need to separate two things better: 

 *  one thing is the parsing step where only a limited set of vehicle types can exist (defined e.g. in the data source or in the road traffic regulations)
 * the other thing is the vehicle profile which then uses bike or car turn restrictions (or any other information from the storage) to influence the (Custom)Weighting.

E.g. the thing "car" exists in OSM data as a concept and from that we feed our car-related encoded values (max_speed etc) but the vehicle profile "car" is the Weighting that then uses encoded values. So in the Weighting we can happily mix truck encoded values (truck_max_speed or turn restrictions) with car encoded values. Or e.g. define a "police car" that does not strictly follow the car encoded value "car.access" where e.g. oneways can be used in the opposite direction or e.g. the bike profile could use the car max speed to avoid certain roads.

> Why rename bicycle to bike?

IMO it is more consistent with our naming ("bike" etc) and the TransportationMode is something that should not be seen as OSM-specific like all the parsers.

boldtrn:
> one thing is the parsing step where only a limited set of transportation modes can exist (defined e.g. in the data source or in the road traffic regulations)

Yes, I do agree with you on this. I still think it should be possible for a transportation mode to define its restrictions :).

Looking at your proposal again I see that the OSMRoadAccessParser cannot be easily changed and is hard coded in the EncodingManager which makes it almost impossible to change the restrictions of TurnCosts without forking GraphHopper. 

> IMO it is more consistent with our naming ("bike" etc) and the TransportationMode is something that should not be seen as OSM-specific like all the parsers.

Ok sure, I was just wondering as it seemed inconsistent.

karussell:
> Yes, I do agree with you on this. I still think it should be possible for a transportation mode to define its restrictions :)

But what would be the use case? (And if there is a new transportation mode in the road traffic regulations we can add it to the method.)

boldtrn:
> In theory this could happen but what is a use case in reality?

I haven't had a real life issue with this yet :). My current issue was just about the name. 

There are quite some transportation modes that have their own turn restrictions, also laid out in the traffic regulations like:
- coach
- motorhome
- trailer
- agricultural
- hazmat

These are all rather exotic cases, but I think these are valid. Maybe for these cases it would be sufficient to make it possible to register a custom TurnCostParser in the EncodingManager? We could handle this in a separate issue.

karussell:
> Maybe for these cases it would be sufficient to make it possible to register a custom TurnCostParser in the EncodingManager?

I don't think this should be done as there was no request for it. And even if it is we can add it to the TransportationMode.

karussell:
Closing in favor of #2168

PR #2161, opened 2020-10-23
Remove the precision adjustment on the ContourBuilder coordinate creation.

gcheval:
Fixes the issue where parts of holes are placed outside of their shell following precision modifications.


gcheval:
Hey @michaz, might need your input on this. Thanks!

michaz (approved):
Hmpf. Yes, thanks.

PR #2162, opened 2020-10-24
CustomWeighting: ensure Map is ordered

karussell:
There is a bug which could occur for Java or the API when using multiple encoded values and/or multiple cases for one encoded value. The next goal should be to make CustomModel.getPriority/getSpeedFactor/getMaxSpeed type safe which we might reach when we limit the language to only "clause" as key and number as value (which is a next but much bigger step).

### DIFF
--- a/core/src/test/java/com/graphhopper/routing/weighting/CustomWeightingTest.java
+++ b/core/src/test/java/com/graphhopper/routing/weighting/CustomWeightingTest.java
@@ -36,6 +36,7 @@ import org.locationtech.jts.geom.GeometryFactory;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 import static com.graphhopper.routing.ev.RoadClass.*;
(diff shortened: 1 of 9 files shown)


core/src/test/java/com/graphhopper/routing/weighting/CustomWeightingTest.java
```diff
@@ -243,6 +244,19 @@ public void testIntEncodedValue() {
         assertEquals(3.70, weighting.calcEdgeWeight(fastEdge, false), 0.01);
     }
 
+    @Test
+    public void unorderedMapShouldFail() {
+        Map map = new HashMap<>();
+        map.put(" > 1.5", 0.2); // allow decimal values in range even for int encoded value
+        CustomModel vehicleModel = new CustomModel();
+        vehicleModel.getPriority().put("lanes", map);
+        try {
```
easbar, line 253: junit 5 has `assertThrows(...)`.

example/src/main/java/com/graphhopper/example/RoutingExample.java
```diff
@@ -97,10 +93,12 @@ public static void headingAndAlternativeRoute(GraphHopper hopper) {
                 addPoint(new GHPoint(42.508774, 1.535414)).addPoint(new GHPoint(42.506595, 1.528795)).
                 setHeadings(Arrays.asList(180d, 90d)).
                 // use flexible mode (i.e. disable contraction hierarchies) to make heading and pass_through working
-                putHint(Parameters.CH.DISABLE, true);
+                        putHint(Parameters.CH.DISABLE, true);
```
easbar, line 96: whitespace

easbar (commented):
Sounds like a good idea to use linked hash map :+1:

core/src/test/java/com/graphhopper/routing/weighting/CustomWeightingTest.java
karussell: Ah, thanks. Old habits are hard to kill :)

PR #2163, opened 2020-10-25
add turn cost example for Java API

karussell:
Should we add this :) ?


api/src/test/java/com/graphhopper/util/HelperTest.java
```diff
@@ -55,6 +55,8 @@ public void testRound() {
         assertEquals(100.94, Helper.round(100.94, 2), 1e-7);
         assertEquals(100.9, Helper.round(100.94, 1), 1e-7);
         assertEquals(101.0, Helper.round(100.95, 1), 1e-7);
+        // todonow
+        assertEquals(1040.02, Helper.round(1040.02125359, -2), 1.e-7);
```
easbar, line 59: I ran into this while working on the example? How is this meant to work? Currently `Helper.round(1040.02125359, -2)` returns `1000`.

example/src/main/java/com/graphhopper/example/RoutingExampleTC.java
```diff
@@ -8,55 +8,81 @@
...
         // handle errors
         if (rsp.hasErrors())
             // if you get: Impossible curbside constraint: 'curbside=right'
             // you either specify 'curbside=any' or Parameters.Routing.FORCE_CURBSIDE=false to ignore this situation
             throw new RuntimeException(rsp.getErrors().toString());
-
         ResponsePath path = rsp.getBest();
-        assert Helper.round(path.getDistance(), -2) == 1700;
+        assert Math.abs(expectedDistance - path.getDistance()) < 1 : "unexpected distance : " + path.getDistance() + " vs. " + expectedDistance;
+        assert Math.abs(expectedTime - path.getTime()) < 1000 : "unexpected time : " + path.getTime() + " vs. " + expectedTime;
```
easbar, line 62: I added these time checks to see the difference when changing u-turn costs.

example/src/main/java/com/graphhopper/example/RoutingExampleTC.java
```diff
@@ -8,55 +8,81 @@
...
     }
 
     // see RoutingExample for more details
     static GraphHopper createGraphHopperInstance(String ghLoc) {
         GraphHopper hopper = new GraphHopperOSM().forServer();
         hopper.setDataReaderFile(ghLoc);
         hopper.setGraphHopperLocation("target/routing-tc-graph-cache");
-        // to enable turn restriction and curbside support ensure that FlagEncoder and profile supports turn costs
-        FlagEncoder car = new CarFlagEncoder(new PMap().
-                putObject("u_turn_costs", 3).putObject("max_turn_costs", 1));
```
easbar: setting u_turn_costs on the encoder has no effect (?!). also I think for the example its better to use `|turn_costs=true` instead of `max_turn_costs=1` since its easier to understand (we either make or do not make use of the OSM turn restrictions)

example/src/main/java/com/graphhopper/example/RoutingExampleTC.java
```diff
@@ -8,55 +8,81 @@
...
                 setCurbsides(Arrays.asList(CURBSIDE_ANY, CURBSIDE_RIGHT)).
                 setProfile("car");
-        GHResponse rsp = hopper.route(req);
+        route(hopper, req, 1730, 112_000);
+    }
 
+    public static void routeWithTurnCostsAndOtherUTurnCosts(GraphHopper hopper) {
+        GHRequest req = new GHRequest(42.50822, 1.533966, 42.506899, 1.525372).
+                setCurbsides(Arrays.asList(CURBSIDE_ANY, CURBSIDE_RIGHT)).
+                putHint(Parameters.CH.DISABLE, true).
```
easbar, line 47: This is a bit ugly because we do not get an error if CH is still enabled. When we change the u-turn costs (and CH is enabled) this change is simply ignored without throwing an error or something.

easbar (commented):
I tried making this example a bit more clear

api/src/test/java/com/graphhopper/util/HelperTest.java
easbar: Ah now I get it: `-2` means the decimal point is shifted to the *left* (by two). Rather counter-intuitive if you ask me because setting the exponent to `-2` I would expect the precision is `10^-2`, not `10^2`.

example/src/main/java/com/graphhopper/example/RoutingExampleTC.java
karussell: We can start to be more restrictive here and throw errors for curbside&noneCH? But this might break clients. Or we have a strict mode (default) and legacy mode ;)

example/src/main/java/com/graphhopper/example/RoutingExampleTC.java
karussell: > setting u_turn_costs on the encoder has no effect (?!)

Ups. Now I start confusing profiles with encoders already myself :D 

> also I think for the example its better to use |turn_costs=true

Ok, sure.

api/src/test/java/com/graphhopper/util/HelperTest.java
karussell: Hmmh, not sure why I did it that way for Helper. Should we can change it?

easbar: Hm for me its fine like this (now that I know it :smile: ). Rounding to `2` means 'two decimal places', from this point of view it makes sense?

example/src/main/java/com/graphhopper/example/RoutingExampleTC.java
easbar: There is this todo here: https://github.com/graphhopper/graphhopper/blob/ff85f037d637e3813dc588b2105df6422afbfa7a/core/src/main/java/com/graphhopper/routing/Router.java#L230-L233

And yes I think we did not do this for some backward-compatibility reasons. Not sure what the exact implications are if we change this.

easbar: > and throw errors for curbside&noneCH

curbsides+nonCH is not the problem. But what is not possible is changing the u-turn costs with CH.

> Or we have a strict mode (default) and legacy mode ;)

Hm this sounds complicated? Or do you mean some flag that controls this somehow (ignore or throw error).

api/src/main/java/com/graphhopper/util/Helper.java
```diff
@@ -318,10 +318,11 @@ public static double keepIn(double value, double min, double max) {
     }
 
     /**
-     * Round the value to the specified exponent
+     * Round the value to the specified number of decimal places, i.e. decimalPlaces=2 means we round to two decimal
+     * places. Using negative values like decimalPlaces=-2 means we round to two places before the decimal point.
```
easbar, line 322: Maybe this is what confused me, does this make it more clear?

api/src/main/java/com/graphhopper/util/Helper.java
```diff
@@ -318,23 +318,24 @@ public static double keepIn(double value, double min, double max) {
...
-    public static double round(double value, int exponent) {
-        double factor = Math.pow(10, exponent);
+    public static double round(double value, int decimalPlaces) {
+        double factor = Math.pow(10, decimalPlaces);
         return Math.round(value * factor) / factor;
     }
 
     public static double round6(double value) {
-        return Math.round(value * 1e6) / 1e6;
+        return round(value, 6);
```
easbar, line 330: Was there any reason these methods did not delegate to round() before?

api/src/test/java/com/graphhopper/util/HelperTest.java
karussell: Yes, makes a bit sense :)

example/src/main/java/com/graphhopper/example/RoutingExampleTC.java
karussell: > curbsides+nonCH is not the problem. 

Ah, due to the request properties thought you mean the problem that this parameter combination does not work for node-based yet ...

> Or do you mean some flag that controls this somehow (ignore or throw error).

Yes. But not sure how ugly the code would get with that.

api/src/main/java/com/graphhopper/util/Helper.java
karussell: No, I think I initially thought two methods (round4 and round6) would be sufficient. But not 100% sure.

example/src/main/java/com/graphhopper/example/RoutingExampleTC.java
easbar: Honestly, as long as their is no complaint I do not really feel like changing anything here :) Adding some ignore parameter makes it more complicated and maybe at some point we no longer need the backward-compatibility here?

api/src/main/java/com/graphhopper/util/Helper.java
karussell: Yes

easbar (approved):

example/src/main/java/com/graphhopper/example/RoutingExampleTC.java
karussell: > Adding some ignore parameter makes it more complicated

Yes, I also do not like it.

> maybe at some point we no longer need the backward-compatibility here?

Unfortunately this goes not away the longer we wait. Instead it is more the opposite. Maybe we introduce restrictions step by step ... have no hard opinion here but it is tricky.

easbar: We would have to figure out how many requests there are that use `u_turn_costs=<some value>` but do *not* use `ch.disable=true`. (and similar with `short_fastest.time/distance_factor`

easbar: Ah and then there might be a problem with `ProfileResolver`, i.e. when `u_turn_costs` is given but no `profile` is specified. Before we had profiles we used `u_turn_costs` to determine which CH preparation shall be used in case there are multiple ones using the same vehicle/weighting, but different u_turn_costs.

PR #2164, opened 2020-10-26
Allow spatial rules to provide fallback values for toll

otbutz:
This should improve our toll information coverage for Germany and Austria. I'd really like to get this merged before i start to work on #1808 so that all intended features are already upstream.


core/src/main/java/com/graphhopper/routing/ev/Toll.java
```diff
@@ -35,11 +36,11 @@
 
     public static Toll find(String name) {
         if (name == null)
-            return NO;
+            return MISSING;
         try {
             return Toll.valueOf(Helper.toUpperCase(name));
         } catch (IllegalArgumentException ex) {
-            return NO;
+            return MISSING;
```
karussell, line 43: Good idea, thanks!

core/src/main/java/com/graphhopper/routing/util/spatialrules/AbstractSpatialRule.java
```diff
@@ -52,6 +53,11 @@ public RoadAccess getAccess(RoadClass roadClass, TransportationMode transport, R
         return currentRoadAccess;
     }
     
+    @Override
+    public Toll getToll(RoadClass roadClass, TransportationMode transport, Toll currentToll) {
+        return currentToll;
+    }
+
```
karussell, line 60: Hmmh it would be really nice if we do not need to introduce a new method for every encoded value where we need country-dependent behaviour. Do you have an idea how we could achieve this?

karussell (commented):
Thanks, see my comments. (note that it also needs to be updated to master)

core/src/main/java/com/graphhopper/routing/util/spatialrules/AbstractSpatialRule.java
otbutz: IMHO only a very limited subset of the encoded values is subject to spatial rules. For example, it wouldn't make sense to override or provide a fallback value for the road surface of a specific area . All the other encoded values fall into the same category as they are basically physical limits(e.g `max_height`) or are to specific to be covered by a countable amount of rules(e.g `hazmat`).

The next step is going to be the introduction of spatial rule files. Maybe in the style of the custom weighting:

```yml
id: DEU
priority: 100

max_speed:
  motor_vehicle:
    road_class:
      motorway: "unlimited"
      trunk: "unlimited"
      primary: 100
      secondary: 100
      tertiary: 100
      residential: 100
      unclassified: 100
      living_street: 4
      "*": "unknown"

road_access:
  motor_vehicle:
    road_class:
      track: "destination"
      path: "no"
      bridleway: "no"
      cycleway: "no"
      footway: "no"
      pedestrian: "no"
      "*": "yes"

toll:
  motor_vehicle:
    road_class:
      motorway: "hgv"
      trunk: "hgv"
      primary: "hgv"
      "*": "no"

area:
  { "type": "Feature", "properties": {},
     "geometry": { 
       "type": "Polygon", 
       "coordinates": [[
       ...
       ]]
     }
  }
```

core/src/main/java/com/graphhopper/routing/ev/Toll.java
otbutz: Do we even need this method?

core/src/main/java/com/graphhopper/routing/util/spatialrules/AbstractSpatialRule.java
karussell: > For example, it wouldn't make sense to override or provide a fallback value for the road surface of a specific area 

Sure. Still it would make sense for use to have a method e.g. for average speed value guessing (car.average_speed) and probably a few others. Probably a method like

```
Object getValue(String valueType, ReaderWay way)
```

would make it more flexible? Or we add methods until it gets more clear on how to solve this.

> or are to specific to be covered by a countable amount of rules(e.g hazmat).

what do you mean here? 

> The next step is going to be the introduction of spatial rule files. Maybe in the style of the custom weighting:

I would keep stuff for now inside Java as the rules could get complex. Later we can surely introduce a file-based approach when there are dozens of Java classes of this kind and we start to see similarities but also exceptions.

btw: see also my latest branch regarding increasing flexibility for CustomWeighting: https://github.com/graphhopper/graphhopper/tree/janino_scripting

core/src/main/java/com/graphhopper/routing/ev/Toll.java
karussell: It does not seem so. Not sure if it is required for consistency reasons?

core/src/main/java/com/graphhopper/routing/util/spatialrules/AbstractSpatialRule.java
otbutz: > Sure. Still it would make sense for use to have a method e.g. for average speed value guessing (car.average_speed) and probably a few others. Probably a method like
> 
> ```
> Object getValue(String valueType, ReaderWay way)
> ```
> 
> would make it more flexible? Or we add methods until it gets more clear on how to solve this.

I'll see if i can simplify the interface a bit :wink: 

> 
> > or are to specific to be covered by a countable amount of rules(e.g hazmat).
> 
> what do you mean here?

One could abuse the spatial rules to add hazmat information which are not part of the OSM material. But that would require a huge amount of tiny rules and defeats the intended purpose of this feature.

> I would keep stuff for now inside Java as the rules could get complex. Later we can surely introduce a file-based approach when there are dozens of Java classes of this kind and we start to see similarities but also exceptions.

Judging from my previous attempt to model the spatial rules for all European countries(#1799), we can safely aim for a simple approach which can be expressed with rule files. Being able to provide arbitrary rules and make them discoverable via PathDetails would be really nice. This is currently impossible without creating a fork and extending the Country enum.

> btw: see also my latest branch regarding increasing flexibility for CustomWeighting: https://github.com/graphhopper/graphhopper/tree/janino_scripting

Looks interesting

otbutz: My biggest gripe with this is typesafety as we would need to pass in the current value as an untyped Object.

otbutz:
Any idea why the edge weight in the CustomWeightingTest increases?

karussell:
You introduced `MISSING("missing")` and the rule their expects that all != NO have toll.

otbutz:
Gotcha. Should i change the unittest here? The OSMTollParser doesn't store this value and only uses it internally.

karussell:
As this is a different issue: can you revert this change for this PR?

And why do you think we should differentiate between NO and MISSING - is there a use case you had in mind or for consistency reasons?

otbutz:
There's a difference between a way which is explicitly tagged as `toll=no` and a way without any toll information.
Most rules will only provide fallback values and should not override `toll=no`.

We could pass `null` for `currentToll` instead but having to handle null in all country rules is uglier IMHO.

karussell:
> This is also consistent with other encoded values 

Yes. But are there really many streets that are tagged toll=no and if there are what would be the use case to differentiate between those two cases? (no and missing)

But again, I think this is a different issue and unnecessarily delays merging this PR.

otbutz:
> > This is also consistent with other encoded values
> 
> Yes. But are there really many streets that are tagged toll=no and if there are what would be the use case to differentiate between those two cases? (no and missing)

Basically `toll=no` accounts for 23% of all toll tags:  https://taginfo.openstreetmap.org/keys/toll#values

As this PR intends to add a fallback logic, we should not override existing mapping information. e.g motorways in country X *usually* require toll to be paid. Mappers might tag ways with `toll=no` if they are explicitly exempt from this.

karussell:
Thanks, now I understand the problem :)

The problem I have is that this would break existing custom models that use toll (they can be in the request too!) and so this needs more care. (I wonder how we should handle such changes in general?)

boldtrn:
+1 to differentiate between no and undefined. Some motorways are explicitly tagged with toll=no to show a difference from the default value of that country.

For example in Austria there are some parts of the motorways that have explicitly no toll and it would be good if that would not be overwritten by the default being yes.

otbutz:
> The problem I have is that this would break existing custom models that use toll (they can be in the request too!) and so this needs more care. (I wonder how we should handle such changes in general?)

Is this really a problem as we do not encode `MISSING` ?

karussell:
Yes, this is a problem and the reason why the unit tests failed.

otbutz:
> Yes, this is a problem and the reason why the unit tests failed.

As far as i understood this, the test doesn't set a value which causes the EV to pick up the default value which is now `MISSING`.
But this situation wouldn't arise for regular users as the parser will only encode `NO`,`HGV` or `YES`

otbutz:
Just checked my assumption by declaring `MISSING` as the last value in the Toll enum and keeping `NO` as the first one. With this change, the test completes without setting any toll values on the EdgeIteratorState.

So the only problem is the failing unittest itself, which blindly assumes that the default value is `NO` if no value has been explicitly set.

otbutz:
Existing custom models should also be fine as we encode only `NO`, `HGV` and `MISSING`:

* `"if": "toll == ALL"` works
* `"if": "toll == HGV"` works
* `"if": "toll == NO"` works
* `"if": "toll != NO"` works

karussell:
The problem is that the meaning of `toll != NO` changes when we introduce MISSING. E.g. for the custom model
```json
"priority": {
   "if": "toll != NO",
   "multiply_by": 0
}
```

the intent is to exclude tolls, but when we introduce MISSING this no longer works and you would need `toll == YES || toll == HGV`.

otbutz:
That would be true if we encounter the value `MISSING` while iterating the graph. But we prevent that in the OSMTollParser:

https://github.com/graphhopper/graphhopper/blob/6f052a2ad9d6bb5a9f57135f9d020a93c0e50ca7/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTollParser.java#L68-L72

karussell:
Ok, I see. So you introduced MISSING just for the purpose of country rules and in the final graph we never have MISSING. This is a bit confusing I think and not consistent with the other encoded values.

otbutz:
@karussell i'm now using `Optional` to avoid null handling or introducing new Toll values

karussell:
Hm, thanks for the effort here. Unfortunately it is not consistent with the other methods now :)

Ok. Let's introduce the MISSING value and do not overwrite it in the parser, so that everything is consistent and we really have MISSING values for missing tags. 

We must write this change as a breaking one into the changelog. Furthermore we need to fix all examples, tests and docs where we use `toll != NO`. Do you think this is a better solution?

otbutz:
IMHO it is better to break things here and issue a warning. On the positive side, the user gains more flexibility as the CustomModels can handle MISSING at runtime.

PR #2167, opened 2020-10-27
Remove InstructionAnnotation

otbutz:
HGV toll is the default for motorways and trunks in a lot of countries. Displaying a toll warning for `Toll.HGV` isn't that helpful for truck drivers and will confuse owners of non-HGV vehicles.


boldtrn:
Have a look at https://discuss.graphhopper.com/t/using-instruction-annotations/6005

I think we should get rid of annotations in general :).

otbutz:
Agreed but that would require changes on the client side if i'm not mistaken?

boldtrn:
If you want to show the same information on the client as it was before, yes. But I think for the change on the GraphHopper side it should be enough to just remove the annotation handling in the Instruction generation.

otbutz:
@boldtrn i've completely removed the InstructionAnnotation

karussell:
Thanks, good idea to remove that (have changed the title of this issue accordingly). Can you also have a look into the gh maps JS code? (we can later add support for toll, ferry, ... via path details)

otbutz:
@karussell i've removed the annotation handling from instructions.js and updated the API documentation

boldtrn:
Since this is a breaking change, should we mention it in the changelog?

otbutz:
@karussell @boldtrn done

karussell:
Thanks!

PR #2168, opened 2020-10-27
TransportationMode refactoring

karussell:
Should avoid new methods to FlagEncoder like in #2159


boldtrn (approved):
LGTM, thanks a lot for the PR :+1:

core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
```diff
@@ -60,10 +60,32 @@ public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, boolean fer
...
-            
+
         roadAccessEnc.setEnum(false, edgeFlags, accessValue);
         return edgeFlags;
     }
+
+    public static List<String> toOSMRestrictions(TransportationMode mode) {
+        switch (mode) {
+            case FOOT:
+                return Arrays.asList("foot", "access");
```
otbutz, line 73: I'd argue that the restrictions are a defining property of a TransportationMode and would move them to the Enum.

boldtrn: But only for OSM data. If you use the transportation mode with anything else, these values might be different?

karussell: The naming is IMO OSM-specific and should be in the parser

otbutz: > OSM-specific

Good point

karussell: Also some restrictions could be not existent in the TransportationMode that exist in OSM

core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
```diff
@@ -60,10 +60,32 @@ public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, boolean fer
...
+            case MOTOR_VEHICLE:
+                return Arrays.asList("motor_vehicle", "vehicle", "access");
+            case CAR:
+                return Arrays.asList("motorcar", "motor_vehicle", "vehicle", "access");
+            case MOTORCYCLE:
+                return Arrays.asList("motorcycle", "motor_vehicle", "vehicle", "access");
+            case HGV:
+                return Arrays.asList("hgv", "motor_vehicle", "vehicle", "access");
+            default:
+                return mode.isMotorVehicle() ?
```
otbutz, line 87: Shouldn't we add a case for each mode and throw an exception in default? We could add a unittest to ensure that all TransportationModes return a value for `toOSMRestrictions()`

karussell: Good question. Having this default might make it easier for forks but forcing a complete list of TransportationModes also sounds good.

otbutz: I'd argue that if a fork wants to extend TransportMode they need think about what restrictions are appropriate.

core/src/main/java/com/graphhopper/routing/util/TransportationMode.java
```diff
@@ -15,28 +15,26 @@
...
 public enum TransportationMode {
+    OTHER(false), FOOT(false), VEHICLE(false), BIKE(false),
+    MOTOR_VEHICLE(true), CAR(true), MOTORCYCLE(true), HGV(true);
 
-    OTHER(0), MOTOR_VEHICLE(1), BICYCLE(2), FOOT(3);
+    private final boolean motorVehicle;
 
-    private final int value;
-
-    TransportationMode(int value) {
```
otbutz, line 34: Do we want to keep the hierarchy here?

e.g

```java
    OTHER()
    FOOT(OTHER)
    VEHICLE(OTHER)
    BIKE(VEHICLE),
    MOTOR_VEHICLE(VEHICLE),
    CAR(MOTOR_VEHICLE),
    MOTORCYCLE(MOTOR_VEHICLE),
    HGV(MOTOR_VEHICLE);

    TransportationMode(TransporationMode parent) {
         // ...
```

karussell: Not sure. We could also create a second method isVehicle ... I even had this in the enum before but removed it as I did not find a meaningful use case.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
boldtrn: The problem is, that if you can't change this when using GraphHopper as a lib, you would actually have to fork it.

karussell: I do not see a big problem that this is not customizable (it wasn't before and there are currently no use cases known that need this).

otbutz: TransportationMode is an enum so it can't be extended by library users without creating a fork?

karussell:
Have removed TransportationMode.MOTOR_VEHICLE as we should include only concrete transportation modes in the enum. Furthermore we still have the method isMotorVehicle of the enum.

PR #2171, opened 2020-11-05
Refactor PathDetailsBuilderFactory

otbutz:
* only use `PathDetailsBuilderFactory` within `PathDetailsFromEdges.calcDetails()`
* use a Set for the request path details as duplicate entries make no sense
* create all PathDetailsBuilders within a switch statement for better readability
* provide varargs `GHRequest.setPathDetails()` variant for better readability when using hardcoded list of values(e.g unittests)

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ev/RouteNetwork.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/RouteNetwork.java
@@ -27,10 +27,8 @@ public enum RouteNetwork {
 
     MISSING("missing"), INTERNATIONAL("international"), NATIONAL("national"), REGIONAL("regional"),
     LOCAL("local"), OTHER("other");
-
-    public static String key(String prefix) {
-        return prefix + "_network";
-    }
+    
+    public static final String KEY_SUFFIX = "_network";
 
     private final String name;
 
--- a/core/src/main/java/com/graphhopper/routing/ev/BikeNetwork.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/BikeNetwork.java
@@ -1,5 +1,5 @@
 package com.graphhopper.routing.ev;
 
 public class BikeNetwork {
-    public static final String KEY = RouteNetwork.key("bike");
+    public static final String KEY = "bike" + RouteNetwork.KEY_SUFFIX;
 }
--- a/core/src/main/java/com/graphhopper/routing/ev/FootNetwork.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/FootNetwork.java
@@ -1,5 +1,5 @@
 package com.graphhopper.routing.ev;
 
 public class FootNetwork {
-    public static final String KEY = RouteNetwork.key("foot");
+    public static final String KEY = "foot" + RouteNetwork.KEY_SUFFIX;
 }
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -198,7 +198,7 @@ abstract public class BikeCommonFlagEncoder extends AbstractFlagEncoder {
         registerNewEncodedValue.add(avgSpeedEnc = new UnsignedDecimalEncodedValue(getKey(prefix, "average_speed"), speedBits, speedFactor, speedTwoDirections));
         registerNewEncodedValue.add(priorityEnc = new UnsignedDecimalEncodedValue(getKey(prefix, "priority"), 3, PriorityCode.getFactor(1), false));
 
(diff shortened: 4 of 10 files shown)


karussell:
Why removing the method GraphHopper.setPathDetailsBuilderFactory? Only with this method adding custom encoded values and path details is possible.

otbutz:
I wasn't aware that this is supported at the moment. I've rebased my branch on the current master including the mtb changes and removed the last commit.

otbutz:
@karussell anything missing?

karussell:
time :D

otbutz:
> time for a closer review :D

Fixed the conflicts

api/src/main/java/com/graphhopper/GHRequest.java
```diff
@@ -41,7 +44,7 @@
     private List<String> pointHints = new ArrayList<>();
     private List<String> curbsides = new ArrayList<>();
     private List<String> snapPreventions = new ArrayList<>();
-    private List<String> pathDetails = new ArrayList<>();
```
karussell, line 44: I'm unsure if we should make it a set, especially since the request order (see below) might be significant.

Same problem in Request.pathDetails.

core/src/main/java/com/graphhopper/routing/ev/RouteNetwork.java
```diff
@@ -27,10 +27,8 @@
 
     MISSING("missing"), INTERNATIONAL("international"), NATIONAL("national"), REGIONAL("regional"),
     LOCAL("local"), OTHER("other");
-
-    public static String key(String prefix) {
-        return prefix + "_network";
-    }
```
karussell, line 33: What does this change improve? We have similar handling e.g. with EncodingManager.

Also please note my recent change here: https://github.com/graphhopper/graphhopper/commit/71631dfb480ebac8c9035c34e0e08b8cadc45949

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -72,7 +73,7 @@ public PathMerger setDouglasPeucker(DouglasPeucker douglasPeucker) {
         return this;
     }
 
-    public PathMerger setPathDetailsBuilders(PathDetailsBuilderFactory pathBuilderFactory, List<String> requestedPathDetails) {
```
karussell, line 75: request order might be significant (I know that json shouldn't be considered ordered but some clients might still do)

karussell (changes requested):
Thanks, I like that setPathDetails is now usable without an array wrapper.

core/src/main/java/com/graphhopper/routing/ev/RouteNetwork.java
otbutz: Java allows switch-case statements for Strings only if they're constants. A value obtained from a static method doesn't work.

core/src/main/java/com/graphhopper/util/PathMerger.java
otbutz: The whole List vs Set boils down to how we should treat duplicates in the requested details. The current implementation doesn't really check for this case and will treat it the same way as a missing builder:

https://github.com/graphhopper/graphhopper/blob/598ecb0ff519a86112686e8a97adbfe9c3148aa7/core/src/main/java/com/graphhopper/util/details/PathDetailsBuilderFactory.java#L89-L91

core/src/main/java/com/graphhopper/routing/ev/RouteNetwork.java
karussell: I see. But for this single switch-case usage I wouldn't want to change it globally. You could just assign it like you did for the priority key?

core/src/main/java/com/graphhopper/util/PathMerger.java
karussell: IMO this is not a bug as duplicating a path detail can be seen as an illegal argument.
Can we still request path details in a list with the POST endpoint? I'm also worried about different ordered in the output JSON causing trouble with clients.

core/src/main/java/com/graphhopper/routing/ev/RouteNetwork.java
otbutz: Do we get any benefits by indirecting the key generation via `RouteNetwork.key()` ?

```java
public class BikeNetwork {
    public static final String KEY = RouteNetwork.key("bike");
}
```

```java
public class BikeNetwork {
    public static final String KEY = "bike" + RouteNetwork.KEY_SUFFIX;
}
```

At runtime there's no difference between these two but the first one is not compile time constant which blocks the usage in the switch statement.

otbutz:
@karussell i've switched back to using a List and implemented the duplicate detection within PathDetailsBuilderFactory

PR #2174, opened 2020-11-11
Remove linkA/B fields from CH shortcuts

easbar:
This PR removes the `E_LINKA/B` fields from CH shortcuts and thus reduces the shortcut storage size from 36 to 28 bit (-23%) for edge-based CH and from 28 to 20 (-29%) bits for node-based CH. The CH query speed is the same (if not faster) than current master.

The reason we do not need the `E_LINKA/B` fields anymore is that since we merged #2132 all shortcuts adjacent to the same node are already stored consecutively and such that `getLevel(nodeA) < getLevel(nodeB) if nodeA!=nodeB`. I had to adjust quite a few tests where we manually create shortcuts and violated this condition. I also inlined `EdgeAccess` /`CHEdgeAccess` into `BaseGraph`/`CHGraphImpl`, because the edge iteration is no longer the same for CH and `BaseGraph`. Previously we used the abstract `EdgeAccess` class to modify the behavior of `EdgeIteratorStateImpl` to make it work for CH, now this is separated.

Iterating the edges&shortcuts at a given node now works as follows. For each node we store the index of the last shortcut we added at this node (just as before). Starting with this shortcut we then iterate the shortcuts by decreasing the edge index by one each time we call `iter.next()`. When we are done iterating the shortcuts we continue iterating the base graph edges (using the double linked list as before). 

I also tried iterating the shortcuts going from lower to higher shortcut ids here: https://github.com/graphhopper/graphhopper/commit/4216091664765dabefcc78b1a5df6d0028689d6f, but this did not turn out to be significantly faster. The routing in this commit is correct, but doing it like this would require adjusting many more tests, because doing it this way we have to 'freeze' the CHGraph after adding the last shortcut.


karussell:
Wow, really nice :+1: 

> now this is separated

:+1:

Issue #2037, opened 2020-05-14
Allow specifying areas for custom weighting in a separate file


easbar:
When we use `areas` in a custom model file we might face the issue of having to copy and paste large amounts of geojson data into the custom model file (for example city borders). This could be improved by allowing to reference another GeoJSON file used to define the area(s)? Otherwise we'd also have to duplicate the GeoJSON data, say if we want to refer to the same areas in different custom model files.

easbar:
Currently we use an object where the keys are the area names and the values are GeoJSON features to specify areas in custom-models:

```
 "areas": {
    "custom1": {
      "type": "Feature",
      "id": "something",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [ 1.525, 42.511 ],
            [ 1.510, 42.503 ],
            [ 1.531, 42.495 ],
            [ 1.542, 42.505 ],
            [ 1.525, 42.511 ]
          ]
        ]
      }
    }
  }
```

Unfortunately, this `areas` object is not compatible GeoJSON. So we cannot just use standard GeoJSON tools to view/create/modify the areas. In contrast, for the custom areas in `custom_areas.directory` each file needs to be a JsonFeatureCollection (with potentially multiple features). 

So using a JsonFeatureCollection instead of the object we currently use also for the areas in custom models would make it easier to work with multiple areas and more consistent with the custom areas, but we would need some convention how to define the area names. We could use the `id` field (or the `properties` object) for this, e.g.:

```
"areas": {
 "type" : "FeatureCollection",
 "features" : [{ 
   "type" : "Feature", 
   "id": "custom1",
   "properties" : {},  
   "geometry": {
     "type": "Polygon",
     "coordinates": [
             [
                [ 1.525, 42.511 ],
                [ 1.510, 42.503 ],
                [ 1.531, 42.495 ],
                [ 1.542, 42.505 ],
                [ 1.525, 42.511 ]
             ]
    ]
   }
 }]
}
```

This way we could easily allow multiple areas per file. The main advantage seems that this way we don't need a long list of area file names in the custom model, i.e. something like this (?):

```
"area_files": [ "my_areas.json"]
```

vs. this:

```
"area_files": ["areas1.json", "areas2.json", ..., "area3000.json"]
```

[...]

PR #2177, opened 2020-11-12
Add support for custom SpatialRules

otbutz:
* recreated our current country borders file using ~naturalearth dataset~ OSM adminlevel 2 boundaries because the current one had topology errors(self intersecting polygons etc)
* decouple spatialrule handling and the encoding of the rule ids:
   * avoid having to reorder TagParsers within EncodingManager.Builder.build()
   * apply rules without having to encode them in the graph
* allow to provide arbitrary rules via `GraphHopperConfig`
* use the SpatialRuleLookup to retrieve IDs for encoded rules instead of relying on the Country enum
* retrieve encoded IDs via SpatialRuleDetails
* use the boundingbox provided by `spatial_rules.max_bbox` to directly obtain intersections with border polygons


otbutz:
Current state of our boundaries file:

intersections

e.g German baltic sea:

intersection-example

Same area in the new boundaries file:

new-boundaries

The new file doesn't cause issues with JTS processing and is intersection free:

new-boundaries-map

otbutz:
I also renamed the file so it can be displayed within Github:

https://github.com/graphhopper/graphhopper/blob/641e0c245e3536841544360fb323d6b99b6d9713/core/files/spatialrules/countries.geojson

otbutz:
@karussell could you take a look and give me some feedback?

otbutz:
I managed to improve the border polygon file a bit further:

Current master:
old

My branch:
new

otbutz:
Switched to admin level 2 boundaries:

admin2

boldtrn:
The last image looks really good. What is the size of the geojson? IIRC initially we had a file of about ~25mb and we had to simplify that because we didn't want to blow up the repository size.

web/src/test/java/com/graphhopper/http/SpatialRulesTest.java
```diff
@@ -49,11 +49,9 @@
 
     private static GraphHopperServerConfiguration createConfig() {
         GraphHopperServerConfiguration config = new GraphHopperServerTestConfiguration();
-        // The EncodedValue "country" requires the setting "spatial_rules.borders_directory" as "country" does not load via DefaultTagParserFactory
-        // TODO should we automatically detect this somehow and include a default country file?
         config.getGraphHopperConfiguration().
                 putObject("graph.flag_encoders", "car").
-                putObject("graph.encoded_values", "country,road_environment,road_class,road_access,max_speed").
+                putObject("graph.encoded_values", "spatial_rule,road_environment,road_class,road_access,max_speed").
```
boldtrn, line 54: Maybe worth mentioning this in the changelog?

boldtrn:
This PR looks really interesting. I haven't tried it yet but at least looking at the code, this looks really promising :).

otbutz:
> What is the size of the geojson?

Around 3.2MB which is IMHO still a reasonable size for that level of detail. Thanks to Github you can have a look at it yourself:

https://github.com/graphhopper/graphhopper/blob/3a5fdb217ab7613d7f75f1ae56cc8b771d6fbbdb/core/files/spatialrules/countries.geojson

Adminlevel 2 seems to be a good fit for this usecase as we get close to 100% landmass coverage. This eliminates most of the simplification artifacts which caused parts of smaller islands to be flagged as No Mans Land.

boldtrn:
> Adminlevel 2 seems to be a good fit for this usecase as we get close to 100% landmass coverage. This eliminates most of the simplification artifacts which caused parts of smaller islands to be flagged as No Mans Land.

Yes I fully agree, that is a massive improvement. There were several places along complex shorelines where the SpatialRuleLookup had too many "no mans land" artifacts.

web/src/test/java/com/graphhopper/http/SpatialRulesTest.java
otbutz: Seems like a good idea. I wasn't quite sure about the renaming but it makes no sense to stick with `country` as users could encode arbitrary things here.

otbutz:
> This PR looks really interesting. I haven't tried it yet but at least looking at the code, this looks really promising :).

Thanks :smiley: 

The OSM dataset is steadily growing but we still have to deal with missing information(e.g max_speed tagging) and SpatialRules could fill this gap. Our current fallback values are not bad but are far from perfect:

https://github.com/graphhopper/graphhopper/blob/598ecb0ff519a86112686e8a97adbfe9c3148aa7/core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java#L108-L142

Considering the fact that we could default to country specific values which should yield more realistic values:

https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxspeed#Default_speed_limits_by_country

These changes should allow users of the core lib to provide their own SpatialRules without having to fork the repo. The next big step will be the definition of a SpatialRule fileformat so non-coders are also able to reap the benefits and/or provide sophisticated rules for the upstream project.

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -663,6 +670,18 @@ public String toEncodedValuesAsString() {
...
+        StringBuilder str = new StringBuilder();
+        for (SpatialRule rule : spatialRuleLookup.getRules()) {
+            if (str.length() > 0)
+                str.append(",");
+
+            str.append(rule.getId());
+        }
+
+        return str.toString();
+    }
```
karussell, line 683: Why do we need this? To be independent of the original spatial rule files?

karussell (changes requested):
Thanks a lot for this. I like several things here. E.g. the file improvements, the different OSM data source, the cleanup in several parts and removal of the Envelope. Also that there is a possibility to customize the spatial_rules. 

Can you split this up that all the refactorings goes into a different PR (and merged before this) and only the customizability goes into this PR?

Regarding the refactoring I see the challenge that we have two or three different things:

1. we need countries, so that I have can do country-specific stuff in CustomWeighting ala `country==DEU` and this geometry is always Germany, picked from the geojson in this repo and IMO cannot be customized. And this is why I think we still need the separate Country enum and store it in 7 or 8 bits.

2. at the same time we need to support custom areas. Probably via some StringEncodedValues? (not really required as we could use `int`s for now)

3. we could need custom areas that we only need for the import and that I do not want or need to store in an encoded value

btw: I also think we need a broader SpatialRule interface. One point was that I need the full ReaderWay object but also I would like to have an area-dependent `average_speed` guessing and a country-dependent ferry handling. Did you had a chance to look into this?

otbutz:
> Can you split this up that all the refactorings goes into a different PR (and merged before this) and only the customizability goes into this PR?

I'll give it a shot :wink:

> Regarding the refactoring I see the challenge that we have two or three different things:
>
>  1. we need countries, so that I have can do country-specific stuff in CustomWeighting ala `country==DEU` and this geometry is always Germany, picked from the geojson in this repo and IMO cannot be customized. 

Shouldn't the user be able to use boundaries with a higher resolution if he wants to? We already use some kind of selector logic as we load the ID from a specific property e.g `ISO3166-1:alpha3` which clearly separates it from other boundaries the user might introduce.

> And this is why I think we still need the separate Country enum and store it in 7 or 8 bits.

IMHO an enum is perfectly fine as long it has only a few entries. But we're talking about +200 possible countries which would take up a good chunk of space if encoded and are probably never used afterwards. We need an approach which is able to scale. 

Mangling a Larry Wall quote: "~Easy~ Small things should be ~easy~ small, and ~hard~ large things should be possible."

> 2. at the same time we need to support custom areas. Probably via some StringEncodedValues? (not really required as we could use `int`s for now)

Yeah i'm still not sure about how we should deal with "dynamic" values or ones that might get quite large as our current implementation is more geared towards encoding things which can be expressed with a few bits. The approach in my PR works but might fall a bit short if we need to encode *a lot* of rules(countries,states,districts,etc)

>  3. we could need custom areas that we only need for the import and that I do not want or need to store in an encoded value

The PR adds the ability to use Spatialrules without encoding them if someone wants to save on diskspace but some kind of filter would be nice. But i'm not sure how to handle this if we want to avoid that the user has to paste 200 countrycodes into a YML file. 

> btw: I also think we need a broader SpatialRule interface. One point was that I need the full ReaderWay object but also I would like to have an area-dependent `average_speed` guessing and a country-dependent ferry handling. Did you had a chance to look into this?

I'm still a fan of the possibility to create rule files. Maybe we can establish some kind of middle ground here and allow for more in-depth handling in the code while using a simplified version for the filebased approach e.g SpatialRule/SpatialRuleFile

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -426,6 +427,11 @@ public DecimalEncodedValue getDecimalEncodedValue(String key) {
     public <T extends Enum> EnumEncodedValue<T> getEnumEncodedValue(String key, Class<T> enumType) {
         return encodedValueLookup.getEnumEncodedValue(key, enumType);
     }
+    
+    @Override
+    public SpatialRuleLookup getSpatialRuleLookup() {
+        return encodedValueLookup.getSpatialRuleLookup();
+    }
```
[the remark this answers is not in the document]
otbutz: Depends on the fact if we switch `PathDetailsBuilderFactory.createPathDetailsBuilders()` to accept an EncodingManager instead of the EncodedValueLookup interface.

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
otbutz: To check if the graph loaded from disk still fits the current configuration. Similar like how we treat changes to encoded values.

karussell:
What comes now to my mind is that a string comparison in weighting is suboptimal and our goal should be the likely faster enum or integer comparisons. The difference is likely very small but especially in the Weighting every nano second counts as it is executed millions of times, especially when we have a choice like now in the early state :)

And you are right that the ID for custom areas should be a flexible thing - so the String is the right choice (not an enum) and we can still reach the expressiveness of `custom_area == EAST_SAXONY` in CustomWeighting (btw: I would probably rename spatial_rule to custom_area), because we can include somehow a declaration step like `int EAST_SAXONY = getInt(eastSaxonyRule)` before.

Now this "custom area" code should be work identical to the "country-dependent" code. And so, I still think we need a **separate** EncodedValue and name it `country` :) (then we can write expressions like `country == DEU` additionally to `custom_area==EAST_SAXONY`). 

> But we're talking about +200 possible countries which would take up a good chunk of space if encoded and are probably never used afterwards.

We should store only the countries where we have custom rules for. And so IMO for a long time 7 bits or probably only 6 bits will be sufficient. And if you do not want this you could exclude the country from the encoded_values and use your own subset with `custom_area` or `custom_country` or similar. Although I'm unsure about this - maybe we need the `country` as another default EncodedValue as it could be required for instructions or similar. Not sure yet.

> Shouldn't the user be able to use boundaries with a higher resolution if he wants to?

Isn't replacing the file no longer possible?

> Yeah i'm still not sure about how we should deal with "dynamic" values or ones that might get quite large as our current implementation is more geared towards encoding things which can be expressed with a few bits. The approach in my PR works but might fall a bit short if we need to encode a lot of rules

Why?

> (countries,states,districts,etc)

This could be done either via storing only one area plus the hierarchy or storing multiple area IDs if they are unrelated.

> I'm still a fan of the possibility to create rule files

I do not think this is necessary now (we can introduce this later) and before we have custom rule files we need a common Java base class or interface.

otbutz:
> What comes now to my mind is that a string comparison in weighting is suboptimal and our goal should be the likely faster enum or integer comparisons. The difference is likely very small but especially in the Weighting every nano second counts as it is executed millions of times, especially when we have a choice like now in the early state :)
> 
> And you are right that the ID for custom areas should be a flexible thing - so the String is the right choice (not an enum) and we can still reach the expressiveness of `custom_area == EAST_SAXONY` in CustomWeighting (btw: I would probably rename `spatial_rule` to `custom_area`), because we can include somehow a declaration step like
> `int EAST_SAXONY = getInt(eastSaxonyRule)`
> before.

The String ID of a rule is currently only used as a label and to enforce uniqueness. The encoded value is purely based on the index of the rule which we could also use for the CustomWeighting to avoid String comparisons:

```java
int eastSaxony = spatialRuleLookup.getRuleIndex(eastSaxonyRule);
```

> 
> Now this "custom area" code should be work identical to the "country-dependent" code. And so, I think this should be a String, but we need a **separate** EncodedValue and name it `country` :) (then we can write expressions like `country == DEU` additionally to `custom_area==EAST_SAXONY`).
> 
> > But we're talking about +200 possible countries which would take up a good chunk of space if encoded and are probably never used afterwards.
> 
> We should store only the countries where we have custom rules for. And so IMO for a long time 7 bits or probably only 6 bits will be sufficient. And if you do not want this you could exclude the country from the encoded_values and use your own subset with `custom_area` or `custom_country` or similar. Although I'm unsure about this - maybe we need the `country` as another default EncodedValue as it could be required for instructions or similar. Not sure yet.

We should decided if we want to give countries a special treatment here. From the encoding standpoint they should be treated differently. Maybe i want to apply a few hundred custom rules(e.g modeling some traffic zones etc) but i don't need them to be encoded in the graph.

Maybe we should really push the country selection into the configuration. Nobody wants to list +200 countries but it might get **a lot** more readable if we also interpret region identifiers:

```
countries=deu,aut,asia
```

> 
> > Shouldn't the user be able to use boundaries with a higher resolution if he wants to?
> 
> Isn't replacing the file no longer possible?

The user only configures the location of the borders directory. We parse any GeoJSON in there and try to lookup fitting borders by the specified ID. As long as the user provides borders for all rules with a fitting properties(`ISO3166-1:alpha3`=`DEU`), this should be totally doable.

otbutz:
@karussell i've moved much of the general refactoring bits into its own PR: #2188

easbar:
> The last image looks really good. What is the size of the geojson? Edit: I just checked - a bit over 3MB

The geojson file uses [a lot more decimal places](https://www.explainxkcd.com/wiki/index.php/2170:_Coordinate_Precision) than should be relevant in practice, e.g. `[32.353613281250006,-21.136523437500003]`. Here is a tool that can be used to reduce the precision: https://github.com/perrygeo/geojson-precision. It was mentioned in this (maybe also otherwise interesting) article: https://docs.mapbox.com/help/troubleshooting/working-with-large-geojson-data/
Would it make sense to run Douglas-Peucker on the file (in hope of simplifying the geometry without losing features)?

otbutz:
> The geojson file uses [a lot more decimal places](https://www.explainxkcd.com/wiki/index.php/2170:_Coordinate_Precision) than should be relevant in practice, e.g. `[32.353613281250006,-21.136523437500003]`.

Thanks. I already experimented with different precision values in the mapshapper export but forgot about it later on. Gonna give it a try and check if i can reduce the filesize any further.

> Would it make sense to run Douglas-Peucker on the file (in hope of simplifying the geometry without losing features)?

The geometries have already been simplified using the Visvalingam algorithm which produces better results for this usecase. You can see the difference in my previous post: https://github.com/graphhopper/graphhopper/pull/2177#issuecomment-729625834

It's basically Douglas-Peucker(master) vs Visvalingam(PR)

> Probably not so relevant, but this looks strange:
> image

I only used admin level 2 boundaries with proper country codes to generate the geojson. This excludes some disputed territories which might explain a few gaps:

Screenshot_2020-11-26 graphhopper graphhopper

otbutz:
@karussell a few thoughts how we could tackle this:

* use separate EVs for countries and custom areas
* countries are a static thing so we should use an enum for them
* encoding all possible countries needs a whole byte but we could safe a few bits if we only encode a subset. This way we would only need to encode countries which are part of the datas bbox and/or are part of a List provided via configuration.  e.g an europe import(44 countries) would need 6bit for the country EV or even lower if the user is only interested in subset like `aut,che,deu`
* similar to this PR the selected countries and custom areas can be persisted via StoreableProperties
* refactor the SpatialRuleLookup to be more of a generic PIP lookup

karussell:
> countries are a static thing so we should use an enum for them

Previously I would have said "yes", but when we develop the custom spatial rules we will use Strings and so we should Strings for countries too. This could also help with this problem: 

> encoding all possible countries needs a whole byte but we could safe a few bits if we only encode a subset. This way we would only need to encode countries which are part of the datas bbox and/or 

Here we can continuously build a map ala `ObjectIntMap<String>`, i.e. the first encountered string gets int 1, the next gets 2 and so on. This could be done either with an already existing IntEncodedValue or a [StringEncodedValue](https://github.com/graphhopper/graphhopper/blob/ab92136a1a1f71f73315113f31d59c4a1387b765/core/src/main/java/com/graphhopper/routing/profiles/StringEncodedValue.java) (it would be necessary to make it a bit more dynamic than in this old branch from #1447)

If done via StringEncodedValue then storing this information would be done from the encoded value and no custom code is required. What do you think?

otbutz:
The StringEncodedValue definitely seems like the way to go.

I'm still not sure how to properly handle user provided GeoJSON files. Our current approach of scanning a single folder and retrieving the names from a hardcoded ID field(`ISO3166-1:alpha3`) isn't suitable for this usecase.

Do we want to enforce a single configurable property key for all user provided GeoJSON files or do we need to support this on a per file basis? The later one is more dynamic and allows the user to import existing GeoJSON files without modification but would require more configuration as we would need to know which property to use for which file.

boldtrn:
IMHO, we could decide on a fixed property name? Even someone with no coding experience should be able to rename the keys in a geojson file?

We could use something like `spatial_id`?

otbutz:
The countries.geojson is only around 3MB and it's already hard to handle in a texteditor.

boldtrn:
Well, then let's make it configurable :)

otbutz:
@karussell should we stick with the current priority solution for overlapping areas? Otherwise we'd have to encode multiple areas per way which is really nasty. The SpatialRules are unaffected as this limitation only bites in the encoding step.

My suggestion for a possible configuration approach:

```yml
  ##### Custom Areas & Spatial Rules #####

  # Custom Areas are loaded from user provided GeoJSON files. A registered SpatialRule is enforced for an area if it has the same ID.
  # The following fields can be configured for each custom area file:
  # - location (required): the path of the GeoJSON file
  # - id_field (required): they GeoJSON property of the (multi)polygons to be used as the ID of the resulting area
  # - priority (optional, default: -100): priority of the area. If multiple areas overlap the one with the highest priority according to the natural order
  # - encode (optional: false): if the ID of the area should be encoded in the graph for later retrieval
  # - max_bbox (optional, default: -180,180,-90,90): maximum BBox for which custom areas are loaded. Having less areas, might result in a smaller graph.
  custom_area_files: []
  
  # custom_area_files:
  #   - location: core/files/spatialrules/countries.geojson
  #     id_field: "ISO3166-1:alpha3"
  #     priority: -100
  #     encode: true
  #     max_bbox: -180,180,-90,90
```

otbutz:
@boldtrn @karussell any idea how to handle custom area configuration in the GraphHopperConfig? The yml with explicit GeoJSON file locations looks nice but i still need to create the actual CustomArea objects within GraphHopperManaged and pipe them back into the GraphHopperConfiguration if we want to avoid a jackson dependency in `core`.

So we would end up with something (ugly) like this:
```java
public class GraphHopperConfig {
    private List<Profile> profiles = new ArrayList<>();
    private List<CHProfile> chProfiles = new ArrayList<>();
    private List<LMProfile> lmProfiles = new ArrayList<>();
    // parsed configuration as explained above
    private List<CustomAreaFile> customAreaFiles = new ArrayList<>();
    // JTS polygons as they're loaded from the GeoJSON file(s) or provided by core lib users
    private List<CustomArea> customAreas = new ArrayList<>();
    // rules which can be provided by users of the core lib
    private List<SpatialRule> spatialRules = new ArrayList<>();
    private final PMap map;
```

It would be **a lot** easier and more readable if we don't avoid that jackson dependencies. This feels like threading a needle behind the back...

karussell:
> Otherwise we'd have to encode multiple areas per way which is really nasty.

What do you mean here? If we have multiple countries we would still need only one encoded value. And for custom areas we could specify the encoded value like:

```yml
custom_area_files:
     encoded_value: federal_state
     ...
```

> The yml with explicit GeoJSON file locations looks nice but i still need to create the actual CustomArea objects within GraphHopperManaged

We already do that in GraphHopperManaged so I currently see a reason to do it elsewhere.

easbar:
Is this related to #2037? If we use complex area geometries it would be a mess to duplicate them for multiple custom models.

otbutz:
> Is this related to #2037?

Exactly.

>If we use complex area geometries it would be a mess to duplicate them for multiple custom models.

The current proposal would load `CustomAreas` from multiple files and store them as EncodedValues for later retrieval.

otbutz:
> And for custom areas we could specify the encoded value like:
> 
> ```yaml
> custom_area_files:
>    - location: core/files/spatialrules/countries.geojson
>      encoded_value: federal_state
>      ...
> ```

Good solution but how does this integrate with our current EncodedValues? Should we create and register EVs on-the-fly as they appear in CustomAreaFiles in addition to those configured with `graph.encoded_values` ?

otbutz:
Closing in favor of #2201

PR #2179, opened 2020-11-13
Update README to point to an actively maintained morituri fork

otbutz:
* builds on recent OS versions
* fixed memory leaks: europe can be converted on a regular desktop
* IDs in the pbf are properly sorted
* properly converted road types

optitool-morituri


karussell:
Interesting, thanks!

PR #2180, opened 2020-11-13
move Java code from API documentation to code

### LINKED ISSUE #2175: move client-hc examples from documentation into source code
We should do this for the example here: https://github.com/graphhopper/graphhopper/tree/master/client-hc what we already did for other example code.

karussell:
fixes #2175


client-hc/src/test/java/com/graphhopper/api/Examples.java
```diff
@@ -0,0 +1,100 @@
...
+import com.graphhopper.util.shapes.GHPoint;
+import okhttp3.OkHttpClient;
+import org.junit.Ignore;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.concurrent.TimeUnit;
+
+@Ignore
```
karussell, line 18: For now include it in code for refactoring purposes only. Do not run the tests as we would need an API key etc.

PR #2183, opened 2020-11-23
Update Dropwizard to 2.0.15

boldtrn:
This PR updates Dropwizard to the latest version 2.0.15.

I checked the `slf4j-log4j12` version, that is still the version used in Dropwizard. I also checked the `dropwizard-configurable-assets-bundle`, but I couldn't find exact specifications with which version this is compatible, but it's still the latest version.

Tests are still green and a few manual tests did not reveal any issues.


karussell:
Have merged the latest version 2.0.16 to master and created a backport 2.3 for it as there are a couple of security related fixes included.

PR #2185, opened 2020-11-24
Location index without pt bloat

michaz:
For PT, I've been building a LocationIndex that contains the entire Graph, including all the inner nodes and edges of the PT model, even though they are never returned by a location query: Trips can only start and end on the street network, and I always use a Filter to ensure that.

My LocationIndex for Berlin is 100 MB. With only the street network, it is 6 MB. The rest is junk.

I can easily add the PT part of the Graph **after** building the LocationIndex (this PR). The only reason I haven't been doing that until now is because, later, the LocationIndex refuses to load with a Graph that has had stuff added afterwards, because it uses the number of nodes as a checksum.

May I remove this feature (happens in this PR)? I tried hard, but I can't think of any other way to avoid putting 90% junk information into the LocationIndex. Do you have any idea what else we could use as a checksum, that would survive adding additional stuff to the Graph?

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
+++ b/core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
@@ -254,6 +254,10 @@ public class LocationIndexTree implements LocationIndex {
         if (dataAccess.getHeader(0) != MAGIC_INT)
             throw new IllegalStateException("incorrect location index version, expected:" + MAGIC_INT);
 
+        if (dataAccess.getHeader(1 * 4) != calcChecksum())
+            throw new IllegalStateException("location index was opened with incorrect graph: "
+                    + dataAccess.getHeader(1 * 4) + " vs. " + calcChecksum());
+
         setMinResolutionInMeter(dataAccess.getHeader(2 * 4));
         prepareAlgo();
         initialized = true;
--- a/core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
+++ b/core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
@@ -307,9 +307,12 @@ public class LocationIndexTree implements LocationIndex {
     }
 
     int calcChecksum() {
-        // do not include the edges as we could get problem with CHGraph due to shortcuts
-        // ^ graph.getAllEdges().count();
-        return graph.getNodes();
+        // Just some property of the graph this LocationIndex was created for.
+        // Intention is to more-or-less ensure that we aren't loading a LocationIndex for a totally different
+        // Graph. This is a possibility because the LocationIndex is normally the first thing that's written to disk.
+        // And in that case we want to avoid loading it instead of regenerating it.
+        BBox bounds = graph.getBounds();
+        return Objects.hash(Helper.round6(bounds.minLon), Helper.round6(bounds.maxLon), Helper.round6(bounds.minLat), Helper.round6(bounds.maxLat));
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/graphhopper/storage/index/LocationIndexTree.java
```diff
@@ -254,10 +254,6 @@ public boolean loadExisting() {
         if (dataAccess.getHeader(0) != MAGIC_INT)
             throw new IllegalStateException("incorrect location index version, expected:" + MAGIC_INT);
 
-        if (dataAccess.getHeader(1 * 4) != calcChecksum())
-            throw new IllegalStateException("location index was opened with incorrect graph: "
-                    + dataAccess.getHeader(1 * 4) + " vs. " + calcChecksum());
```
karussell, line 259: This check is necessary in rare cases when GH crashes somewhere in the middle of a CH or LM preparation I think and no graph files are on disk yet (but the loc index is). If then a new import with a different PBF is done it will load the wrong LocationIndex files from disk.

We could try to flush the LocationIndex only at the end on success. But we don't do this at the moment, because we flush files as soon as they are ready to avoid e.g. a re-import of the PBF file if only the CH preparation failed. This was sometimes handy for debugging CH or LM for larger areas, but don't know if it is worth the trouble.

Another solution could be to:

 1. call basegraph.freeze before LocationIndex (and ensure basegraph.freeze was called) 
 2. write the basgraph node+edge count into the header of the `nodes` DataAccess (on basegraph.freeze)
 3. use this basegraph header information to get the checksum (the problem is that we would need to use a bit ugly code to get this header info `dir.find("nodes").getHeader()` instead of using the `Graph` interface)

michaz:
I put back the checksum check. I switched the checksum from the number of nodes to the bounding box of the graph -- also a (somewhat) arbitrary property, but one that doesn't cause problems for what I want to do (insert additional nodes and edges after index creation), while it should prevent the "found a different location index in the graph directory" type of problem equally well.

What do you say?

(I don't really want to touch the freeze-flush-close mechanics, or invent a new field.)

karussell:
> What do you say?

I tried to find a different solution as we already write the node count into the nodes DataAccess (see BaseGraph). But the header is only in-sync with reality after calling "flush" and this does not work for all scenarios. Similar problems exist for the alternative I suggested, when I set the header when we call "freeze" (I ca).

Now using the boundaries is not that handy especially for big graphs as the boundaries might be identical for different versions of the same geographical area.

But what if you could provide an EdgeIterator looping only through the street edges and skip the inner edges? Ala `locationIndex.prepareIndex(edgeIterator)` - would this work?

> And I found a comment that indicated that the number of edges was in the checksum at some point

Ah, yeah. This comment is outdated and we should include the edges now that CHGraph is decoupled from BaseGraph since a few versions.

michaz:
> But what if you could provide an EdgeIterator looping only through the street edges and skip the inner edges?

Hmmm, yeah, that should work..

karussell:
Can you check with this branch: https://github.com/graphhopper/graphhopper/compare/checksum_header

michaz:
> Can you check with this branch

@karussell Thanks, but how do I use that? I know how to use an `EdgeFilter` to get an `EdgeExplorer`, but not how to use it to filter an `AllEdgesIterator`.

karussell:
And would it be possible with an EdgeFilter? (have changed it accordingly)

michaz:
Yep, works for me, thanks! If you merge, I'll merge.

karussell:
Ok, merged it :+1:

PR #2187, opened 2020-11-24
make date parsing working for latest JDK16 again

### LINKED ISSUE #2186: Date parsing fails in latest JDK16
Currently we disabled JDK16 because it fails at date parsing [here](https://github.com/graphhopper/graphhopper/commit/598ecb0ff519a86112686e8a97adbfe9c3148aa7). Now it seems that the latest JDK16 integrated ["CLDR 38"](https://unicode-org.github.io/cldr-staging/charts/38/delta/en.html#Months%20-%20abbreviated%20-%20Standalone) and so the abbreviations for months changed from `Sep` to `Sept` and so the currently working code fails with the latest JDK16:

```
new SimpleDateFormat("yyyy MMM dd", Locale.UK).parse("2015 Sep 30");
```

And indeed, [according to the docs](https://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html#text), the 3 characters just mean that the month is abbreviated - not that exactly 3 characters are used:

> Text: For formatting, if the number of pattern letters is 4 or more, the full form is used; otherwise a short or abbreviated form is used if available. For parsing, both forms are accepted, independent of the number of pattern letters. 

This is a bit ugly as we cannot change the OSM tagging. It seems to work with Locale.US or Locale.ROOT.

See: https://bugs.openjdk.java.net/browse/JDK-8256837

**Update:**

> I'd suggest specifying Locale.US in the formatter. Still, there is a very slim chance that CLDR could change the abbreviated name to "Sept" in Locale.US though

karussell:
Fixes #2186

Breaking change as we have to use a different Locale.


easbar:
Why is this a breaking change?

karussell:
Breaking in the sense that the method is public and someone could use it and rely on the Locale.UK.

In general I'm unsure what to do now as using Locale.US could break soon too.

easbar:
And this is mainly (only?) relevant when parsing dates from OSM? Don't we have to use a certain formatter according to the OSM format anyway? If its unclear what we should use maybe make it configurable?

karussell:
> Don't we have to use a certain formatter according to the OSM format anyway?

Yes, basically that is what we do. We create a simple formatter and parse dates:

new SimpleDateFormat("yyyy MMM dd", Locale.UK).parse("2015 Sep 30");

And with JDK16 they intend to throw an exception because "Sep" will no longer be parsed (as well as "Jul" and "Jun" I think).

Making this configurable does not help, because if Locale.US is also not working it will no longer parse all conditional tags with 3 letter months.

easbar:
Does OSM use a consistent date format like `2015 Sep 30`? If not how are we supposed which format to expect when parsing it?

karussell:
We use a chain of parsers to handle multiple scenarios. But yes, for months they seem to use 3 letters (or all). See our DateRangeParser.

easbar:
Maybe I did not understand the problem yet. 

> surely the date formatter must depend on the Locale

Why? I thought whatever we use to parse the dates must work with the dates in our data/OSM, but why does it have to depend on the locale?

karussell:
Sorry, I probably misunderstood your question. In our DateRangeParser we use fixed Locale to parse the existing IMO all English dates in OSM. 

(But in general the date parser depends on the Locale to parse dates in different languages.)

> but not on how JDK interprets date strings.

Yes, sure! We can build our custom date parser but the easiest way to parse date strings in Java was to use the SimpleDateFormat with a fixed Locale.UK. I would have never expected that they change the behaviour and break existing applications. I mean in principle they are correct and the docs only say that the pattern `MMM` means "abbreviated month" and not "3 letter month". But search the internet for "java 3 letter month parser" and you'll see that this is a very widespread misunderstanding :). They argue that we could set  `-Djava.locale.providers=COMPAT` to keep backward compatibility, but this is not a sane solution for our case.

So for now I would use Locale.US (or ROOT or ENGLISH) and if they later also break we would have to rename certain months from 3 characters to 4 characters before we parse them?

easbar:
Ok this makes sense. I'm also surprised by such a change. Is this final yet?

> have to rename certain months from 3 characters to 4 characters before we parse them?

Yes, or maybe have a look if there is a parser that addresses this issue since we won't be alone.

karussell:
> Ok this makes sense. I'm also surprised by such a change. Is this final yet?

It seems so according to the JDK bug tracker - it is "resolved" already.

michaz:
> I would have never expected that they change the behaviour and break existing applications.

I think they do it because they consider Locales data, not code. `Locale.UK` means: "Default for what we currently (!) think dates should be formatted in the UK."

I'm sure many people use it like we do and use it to parse a _fixed_ format. But I think that wasn't the exact intention.

Theres another constructor, maybe we can use that instead?

```
public SimpleDateFormat(String pattern, DateFormatSymbols formatSymbols)
```

karussell:
> There's another constructor, maybe we can use that instead?

Unfortunately this will still use some locale under the hood and even worse it seems as it is using the default Locale (not what we want):

> System.out.println(Locale.getDefault(Locale.Category.FORMAT));
> System.out.println(Locale.getDefault());

michaz:
Weird.

karussell:
@michaz I tried this DateFormatSymbols and it worked. The problem is still that the default locale is used, but if we use this only for this use case it should be fine.

In general the change to use Locale.ENGLISH already fixes the issue but to avoid surprises in the future (e.g. later JDKs might change the CLDR again) and so the explicit usage of the short month name list makes this a more stable solution (I hope :))

PR #2188, opened 2020-11-24
Refactor spatialrule border handling

otbutz:
* intersection free border polygons with admin level 2
* only keep intersections of border polygons with the max bbox
* refactored CountriesSpatialRuleFactory


core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupHelper.java
```diff
@@ -41,7 +41,7 @@
 public class SpatialRuleLookupHelper {
 
     private static final Logger logger = LoggerFactory.getLogger(SpatialRuleLookupHelper.class);
-    static String JSON_ID_FIELD = "ISO_A3";
+    public static String JSON_ID_FIELD = "ISO3166-1:alpha3";
```
karussell, line 44: This change should be included in the changelog. 

And is this used elsewhere or why did you make it public? E.g. it could be used in Measurement?

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupBuilder.java
```diff
@@ -53,27 +54,27 @@ public static SpatialRuleLookup buildIndex(List<JsonFeatureCollection> jsonFeatu
                 List<Polygon> borders = new ArrayList<>();
                 for (int i = 0; i < jsonFeature.getGeometry().getNumGeometries(); i++) {
                     Geometry poly = jsonFeature.getGeometry().getGeometryN(i);
-                    if (poly instanceof Polygon)
-                        borders.add((Polygon) poly);
-                    else
+                    if (poly instanceof Polygon) {
+                        Geometry intersection = bboxGeometry.intersection(poly);
+                        if (!intersection.isEmpty()) {
+                            PolygonExtracter.getPolygons(intersection, borders);
```
karussell, line 60: Why is this necessary and not just `borders.add(intersection)`? Isn't intersection always a Polygon?

karussell (approved):
Thanks!

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupHelper.java
otbutz: Probably a relict from my other PR.

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupBuilder.java
otbutz: [Geometry.intersection(Geometry other)](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/Geometry.html#intersection-org.locationtech.jts.geom.Geometry-):
> If the result is empty, it is an atomic geometry with the dimension of the lowest input dimension

For two polygons this would result in an empty polygon being picked up by the PolygonExtracter.

otbutz:
@karussell i've fixed the constant and added a small but significant speedup for our Point-in-polygon logic.

According to my local benchmarks the PR only takes about 60% of time to finish the rule lookups in our Measurement class. The biggest benefit seems to come from the new polygons.

otbutz:
I was able to further reduce the size of the GeoJSON file to about 2.7MB following the advice of @easbar (https://github.com/graphhopper/graphhopper/pull/2177#issuecomment-734159127)

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupHelper.java
karussell: Thanks! IMO it should be still included in the changelog. GeoJSON files with ISO_A4 will now break (?)

otbutz:
Added Palestine:

https://github.com/graphhopper/graphhopper/blob/312e8c3d556225ef20d8f5a4301004b735f355f8/core/files/spatialrules/countries.geojson

core/src/main/java/com/graphhopper/routing/util/spatialrules/SpatialRuleLookupHelper.java
otbutz: Done

karussell:
Thanks! And btw the shape of the new GeoJSON looks indeed much better

Some countries seems to be still missing (?)

```
$ grep id countries.geojson | wc -l
217
$ grep ISO countries.geo.json | wc -l
255
```

But I coudn't find them quickly, except the Western Sarah that @easbar already mentioned.

otbutz:
> Thanks! And btw the shape of the new GeoJSON looks indeed much better
> 
> Some countries seems to be still missing (?)
> 
> ```
> $ grep id countries.geojson | wc -l
> 217
> $ grep ISO countries.geo.json | wc -l
> 255
> ```
> 
> But I coudn't find them quickly, except the Western Sarah that @easbar already mentioned.

Our current file only lists 238 countries with a real countrycode, the rest has stuff like `"ISO_A3":"1-99"`. If my bash scripting hasn't let me down:

```
ABW
ALA
ASM
ATA
ATF
BLM
CUW
ESH
GUM
HKG
HMD
MAC
MAF
MNP
NCL
NFK
PRI
PYF
SPM
SXM
UMI
VIR
WLF
```

otbutz:
Aruba, Åland Islands, American Samoa, French Southern Territories, Saint Barthélemy, Curaçao, Western Sahara, Guam, Hong Kong, Heard Island and McDonald Islands, Macao, Saint Martin (French part), Northern Mariana Islands, New Caledonia, Norfolk Island, Puerto Rico, Saint Pierre and Miquelon, Saint Martin (Netherlands part), United States Minor Outlying Islands, Virgin Islands, U.S., Wallis and Futuna

Looks like a lot of small islands and i suspect that a lot of them are covered by the countries they in fact belong to.

karussell:
Ok, nicely grepped and investigated :) - thanks!

karussell (approved):

otbutz:
@karussell can we merge this?

karussell:
Yes. Thanks again!

PR #2189, opened 2020-11-25
Minor comment updates

boldtrn:
Just a minor comment update.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ev/TrackType.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/TrackType.java
@@ -21,13 +21,14 @@ import com.graphhopper.util.Helper;
 
 /**
  * This enum defines the track type of an edge which describes how well-maintained a certain track is.
- * If not tagged the value will be MISSING and all tracktype values that do not fit get OTHER as value.
+ * If there is no value tagged or if the value does not fit the given values, the TrackType will be MISSING.
+ * grade1 is a very well-maintained road, grade5 is a poorly maintained road.
  *
  * @see <a href="https://wiki.openstreetmap.org/wiki/Tracktype">Tracktype Wiki</a>
  */
 public enum TrackType {
     MISSING("missing"), GRADE1("grade1"), GRADE2("grade2"),
-    GRADE3("grade3"), GRADE4("grade4"), GRADE5("grade5"), OTHER("other");
+    GRADE3("grade3"), GRADE4("grade4"), GRADE5("grade5");
 
     public static final String KEY = "track_type";
 
@@ -48,7 +49,7 @@ public enum TrackType {
         try {
             return TrackType.valueOf(Helper.toUpperCase(name));
         } catch (IllegalArgumentException ex) {
-            return OTHER;
+            return MISSING;
         }
     }
 }
--- a/core/src/main/java/com/graphhopper/routing/ev/TrackType.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/TrackType.java
@@ -21,8 +21,7 @@ import com.graphhopper.util.Helper;
 
 /**
  * This enum defines the track type of an edge which describes how well-maintained a certain track is.
- * If not tagged the value will be MISSING (the default and easiest grad) and all edges that do not fit get OTHER as
- * value (the worst grade).
+ * If not tagged the value will be MISSING and all tracktype values that do not fit get OTHER as value.
  *
  * @see <a href="https://wiki.openstreetmap.org/wiki/Tracktype">Tracktype Wiki</a>
  */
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/graphhopper/routing/ev/TrackType.java
```diff
@@ -21,8 +21,7 @@
 
 /**
  * This enum defines the track type of an edge which describes how well-maintained a certain track is.
- * If not tagged the value will be MISSING (the default and easiest grad) and all edges that do not fit get OTHER as
```
boldtrn, line 24: MISSING does not indicate the "easiest grade". The easiest / well maintained grade is grade1. Missing means that we don't know the grade and could assume some other grade, but I wouldn't assume that missing grades mean it's grade1.

Other is not really the worst grade either :).

boldtrn: Maybe we should add some more information on this in the comment defining each grade? But probably it is enough the way it is in the OSM wiki.

boldtrn: BTW: Other is refered as worst value in other comments as well. I would propose to change this for all occurrences. Other is just an undefined state IMHO.

karussell: Hmmh, I think what is meant here is that in some cases these enums could be also converted into the corresponding number and then an order is required. But probably this wasn't a good idea and with better scripting support we can now refer to potential bad track types via:
`track_type.ordinal() > 3 || track_type == MISSING`
instead of just:
`track_type.ordinal() > 3`

> But probably it is enough the way it is in the OSM wiki.

As we do not have a 1:1 mapping I would not do this. Also we should make this more explicit and improve our documentation independent of OSM wiki.

otbutz: The MISSING type is related to how encoded values with no value are handled AFAIK.

boldtrn: > Hmmh, I think what is meant here is that in some cases these enums could be also converted into the corresponding number and then an order is required.

Ah, now I see how this was meant. Thanks for the explanation.

> As we do not have a 1:1 mapping I would not do this. Also we should make this more explicit and improve our documentation independent of OSM wiki.

Don't we have a 1:1 mapping for tracktype? But sure I can add some more information on this, if you'd want this?

otbutz: 1. Do we really need a distinction between `MISSING` and `OTHER` ? We should probably treat them as the same thing and drop `OTHER` completely.
2. Should we reorder the constants according to the grade?

```
    MISSING("missing"), GRADE5("grade5"), GRADE4("grade4"),
    GRADE3("grade3"), GRADE2("grade2"),GRADE1("grade1");
```

karussell: Re 1: yes, you are right

Re 2: interesting idea. But then gradeX != X ... tricky :D 

> Don't we have a 1:1 mapping for tracktype? But sure I can add some more information on this, if you'd want this?

Yes, for now this is probably identical. But in general our aim should be to have values that can be easily used in our CustomWeighting.

karussell: @boldtrn can you remove OTHER for now? All other values:

https://taginfo.openstreetmap.org/keys/tracktype#values

should be interpreted as MISSING.

boldtrn: Ok sure, should we apply the same to the other encoded values as well? Like Surface, RouteNetwork, etc.? We have other and missing there as well.

karussell: No, for surface etc there are definitely many "other" tags that should not mean "missing". In case of tracktype no other values can occur according to wiki and not many occur in real world according to taginfo.

karussell: And e.g. for RouteNetwork there are plenty of tags in the wiki and real world that we cannot handle but should not mean MISSING.

karussell: Hmmh, no. For route=foot it seems there are only 4 possible values.

boldtrn: Ah now I understand your reasoning for removing OTHER, thanks. Ok I updated the PR.

karussell:
Thanks!

PR #2190, opened 2020-11-25
Remove Activation dependency

boldtrn:
Related to #1388 

According to https://github.com/dropwizard/dropwizard/issues/2318 the issue was fixed with DW 2.0, so we should not need this dependency anymore?

I couldn't see errors when starting the webserver locally.


karussell:
Looks good - thanks!

Issue #2181, opened 2020-11-18
Isochrone jumps to unreachable parts of the road network


tomass:
**Describe the bug**

When calculating isochrone (both time and distance based), resulting polygon sometimes jumps over to the parts of the road network unreachable in specified time/distance.

For example:
![paveikslas](https://user-images.githubusercontent.com/969513/99535269-595ab880-29b1-11eb-9787-f5e58419b383.png)
Starting point is a yellow dot. Nearest bridge is some 7km away (so roundtrip is 14km).
The isochrone of 4km manages to jump to unreachable parts „1“ and „2“ of the network on the other side of the river.

**To Reproduce**

Steps to reproduce the behaviour. For example:
1. Calculate isochrone with distance limit of 4-5km from this point: https://www.openstreetmap.org/#map=18/54.59485/24.02584

You will get isochrone reaching parts of the road network over the river without a bridge.

**Expected behaviour**
Isochrone should not cover road network parts unreachable in specified time/distance.

**System Information**

Graphhopper 2.0 on linux, OpenJDK Runtime Environment 20.9 (build 15+36).
Using example configuration, car profile, OSM data from Geofabrik.

michaz:
Yes, very true, this is a known limitation.

Unfortunately, the algorithm that draws the isochrones only "sees" the road segments that "pierce" from the reachable area to the unreachable area -- it doesn't *globally* look for road segments *elsewhere* that are unreachable, and it doesn't know what rivers are.

The solution is more-or-less clear: Globally look up more points that could be "encroaching" and which the isoline has to exclude. I do it in an alternative implementation for public transit (see PtIsochroneResource), but haven't integrated it into the main code yet..

PR #2195, opened 2020-11-27
Isochrone: Look for encroaching points that are not in the SPT

michaz:
Fixes issue #2181. Triangulates a couple more points. But, theoretically, we now don't need the slack in the SPT-size anymore, but can search just up to the asked z-value. So it'll be fewer points again.


michaz:
Even though I think this is "technically correct", I'm not merging it for now, since it degrades performance quite a bit. At the same time, in situations _other_ than those in #2181, it sometimes gives _worse_ results. Reason is a combination of not having _constrained triangulations_ yet, and only taking the largest component of the possibly disconnected area, meaning that we chop off pieces of road that should be inside.

Tricky.

michaz:
Okay, I removed the search-space slack -- now it's _not_ slower than master, but the disconnected area problem remains.

Try next:
- Add "exterior" nodes of isochrone edges (i.e. the first nodes that are _not_ inside the isochrone)
- _Don't_ add nodes from _inaccessible_ parts of the network (e.g. subnetworks, footpaths for car, etc.)
- _Then_ see if good enough
- Someday, get constrained triangulations

michaz:
Not convincing yet.

Rikuoja:
@michaz What is the main reason this pull request is closed? Does the branch work at least in principle, I have a use case in which I might need this behavior, even if I might need to verify the results manually? I.e. can you give an example of a disconnected area and chopping off pieces of isochrone?

PR #2199, opened 2020-12-04
Force english locale for reader-gtfs tests

otbutz:
We set it explicitly in the parent POM to avoid locale issues when matching message contents in unittests:

https://github.com/graphhopper/graphhopper/blob/f15a52c2b9ccfc3bd7aea39617fe0bec5486e860/pom.xml#L183-L190

As the reader-gtfs POM overrides the argLine, we need to set it here as well.


otbutz:
Cherrypicked into #2200

PR #2201, opened 2020-12-04
Custom area support

### LINKED ISSUE #2037: Allow specifying areas for custom weighting in a separate file

otbutz:
* fixes #2037
* allow users to provide custom areas as GeoJSON files
* custom areas can be stored as EncodedValues for later retrieval(PathDetails, CustomModel)
* SpatialRules target CustomAreas now
* ~add support for StringEncodedValues~ see #2204


otbutz:
I'm struggling a bit with dropwizard:

grafik

Any idea how i can prevent it from creating ConfigurationMetadata for the CustomArea instances? Annotating the Getter method in `GraphHopperConfigMixIn` with `@JsonIgnore` didn't help

otbutz:
Our unittests run without any of the Mixin interfaces which explains why my previous attempts failed. Switching to a GraphHopperConfig subclass with annotated method overrides seems to fix the issue.

otbutz:
@karussell @boldtrn @easbar

karussell:
Thanks. 

This is a big change. Would you mind to split this up? E.g. move the StringEncodedValue into a separate PR before?

otbutz:
@karussell i moved StringEncodedValue related changes into #2204

otbutz:
@karussell what do you think about the configuration approach?

```yml
 custom_area_files:
       id_field: "ISO3166-1:alpha3"
       encoded_value: country
       encoded_value_limit: 63
       max_bbox: -180,180,-90,90**
```

I'm not sure if it's better to define the StringEncodedValues with their size elsewhere but that would add additional bloat to the GraphHopperConfig

otbutz:
@karussell could you take a look and give me some feedback?

otbutz:
@karussell should i rebase?

karussell:
Thanks again and sorry for the silence here. Will review it in the next week.

Still I have currently no clear view on how to bring together the concepts of spatial rules, custom areas and custom models. See also #1280 and #1808. Especially after this change that can be combined with custom models the spatial rules look somehow redundant (?) or maybe a spatial rule is only a specific custom model at import time?

otbutz:
> the spatial rules look somehow redundant (?) or maybe a spatial rule is only a specific custom model at import time?

The real benefit of spatial rules stems IMHO from the ability to manipulate the encoded values. Custom models only affect the routing itself but don't reflect in e.g the PathDetails. #2164 is a good example for would *could* be done with spatial rules. One could come up with a custom model which provides country specific toll rules but a client won't be able to retrieve this information later on in the form of "how many toll kilometers did the vehicle travel?". Using a spatial rule this information could easily be gathered from the PathDetails.

> Maybe unrelated, but currently I have no clear view on how to bring together the concepts of spatial rules, custom areas and custom models.

Custom areas can easily handle very huge and detailed border polygons without a performance impact at runtime.

karussell:
> The real benefit of spatial rules stems IMHO from the ability to manipulate the encoded values.

Yes and IMO this is similar to a custom model. Sure, it does not store the calculated information into an encoded value but I could think of a custom model where this is possible:

```yaml
max_speed:
 - if: country == DE
   set to: 120
```

Especially as we have the list of areas where I could think of one "import-model" per country or something like this.

> but a client won't be able to retrieve this information later on in the form of "how many toll kilometers did the vehicle travel?"

This would be an interesting feature in general: on-the-fly statistics similar to path details but defined in the custom model:

```yaml
details:
 - toll distance:
   - if: toll != NO
     add: distance
```

> Custom areas can easily handle very huge and detailed border polygons without a performance impact at runtime. See #2037

Yes. Likely we need something separate, but it could be very similar to a custom model, especially when we allow to define spatial rules in a file, but also in the Java API.

otbutz:
> This would be an interesting feature in general: on-the-fly statistics similar to path details but defined in the custom model:
> 
> ```yaml
> details:
>  - toll distance:
>    - if: toll != NO
>      add: distance
> ```

That sounds like a workable approach. That would be enough to retire spatial rules and replace them with custom models.

So how do we move forward with this? I'm going to drop the spatial rules from the GraphHopperConfig in my PR. It doesn't make sense to put them there if we want to replace them later on.

web-bundle/src/main/java/com/graphhopper/http/GraphHopperManaged.java
```diff
@@ -72,24 +71,23 @@ public GraphHopperManaged(GraphHopperConfig configuration, ObjectMapper objectMa
...
-            List<JsonFeatureCollection> jsonFeatureCollections = new ArrayList<>();
-            try (DirectoryStream<Path> stream = Files.newDirectoryStream(bordersDirectory, "*.{geojson,json}")) {
-                for (Path borderFile : stream) {
-                    try (BufferedReader reader = Files.newBufferedReader(borderFile, StandardCharsets.UTF_8)) {
-                        JsonFeatureCollection jsonFeatureCollection = localObjectMapper.readValue(reader, JsonFeatureCollection.class);
-                        jsonFeatureCollections.add(jsonFeatureCollection);
-                    }
-                }
+        if (!configuration.getString("spatial_rules.borders_directory", "").isEmpty()) {
+            throw new RuntimeException("spatial_rules.borders_directory has been deprecated. Please use custom_area_files instead.");
```
karussell, line 77: Can you additionally point to this PR or to some documentation on how to fix it?

config-example.yml
```diff
@@ -159,18 +159,23 @@ graphhopper:
...
+  # - encoded_value_limit (optional, default: -1): expected number of encountered areas to be stored in the encoded value. Defaults to the number of areas if -1.
+  # - max_bbox (optional, default: -180,180,-90,90): maximum BBox for which custom areas are loaded. Having less areas, might result in a smaller graph.
+  custom_area_files: []
+  
+  # custom_area_files:
+  #   - location: core/files/spatialrules/countries.geojson
+  #     id_field: "ISO3166-1:alpha3"
+  #     encoded_value: country
+  #     encoded_value_limit: 63
+  #     max_bbox: -180,180,-90,90
```
karussell, line 178: Is it possible to move the encoded value definition to `graph.encoded_values`? Then we could define it like `encoded_values=country|limit=63,surface,...` and we have all explicit encoded values in one place and would need to use it for the custom area.

And is max_bbox still necessary? I think I saw some PR where this requirement was removed, but maybe I mix something up.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1048,18 +1054,13 @@ protected void loadOrPrepareLM(boolean closeEarly) {
...
+            CustomAreaLookup landmarkAreaLookup = landmarkAreas.isEmpty() ? CustomAreaLookup.EMPTY : new CustomAreaLookupJTS(landmarkAreas);
+
             for (PrepareLandmarks prep : getLMPreparationHandler().getPreparations()) {
                 // the ruleLookup splits certain areas from each other but avoids making this a permanent change so that other algorithms still can route through these regions.
-                if (ruleLookup != null && !ruleLookup.getRules().isEmpty()) {
-                    prep.setSpatialRuleLookup(ruleLookup);
+                if (landmarkAreaLookup != CustomAreaLookup.EMPTY) {
+                    prep.setCustomAreaLookup(landmarkAreaLookup);
                 }
             }
```
karussell, line 1065: IMO it would be a bit more clear if we do
```java
if(!landmarkAreas.isEmpty()) {
  CustomAreaLookup landmarkAreaLookup = new CustomAreaLookupJTS(landmarkAreas);
  for (PrepareLandmarks prep : getLMPreparationHandler().getPreparations()) {
     prep.setCustomAreaLookup(landmarkAreaLookup); 
  }
}
```
Or did I overlooked something?

core/src/main/java/com/graphhopper/config/CustomAreaFile.java
```diff
@@ -0,0 +1,84 @@
+package com.graphhopper.config;
+
+
+public class CustomAreaFile {
```
karussell, line 4: Isn't this more a CustomAreaConfig and why is this separate from CustomArea?

core/src/main/java/com/graphhopper/config/CustomArea.java
```diff
@@ -0,0 +1,94 @@
...
+import org.locationtech.jts.geom.Polygon;
+
+import com.graphhopper.routing.util.spatialrules.SpatialRule;
+
+/**
+ * Defines a named area which can be encoded in the graph and be targeted with a {@link SpatialRule}
+ * 
+ * @author Thomas Butz
+ */
+public class CustomArea {
```
karussell, line 34: As this replaces the SpatialRules should we rename the PR to `make spatial rules customizable (CustomArea)`?

And can we somehow avoid the name clash with the other CustomArea class?

For future reference can you also list the renamed classes here as github somehow fails to catch all some of them. E.g. SpatialRuleLookupJTS, SpatialRuleLookupHelper (?) This would also help which classes I can skip when reviewing.

And why did you choose to keep some of the classes the same name like SpatialRuleSet? Could it make sense to rename it also to CustomAreaRules or something? Also the name for CountriesSpatialRuleFactory seems not consistent.

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -613,6 +623,15 @@ public void handleTurnRelationTags(OSMTurnRelation turnRelation, TurnCostParser.
...
         IntsRef edgeFlags = createEdgeFlags();
+        if (customAreaLookup != CustomAreaLookup.EMPTY) {
+            GHPoint estimatedCenter = way.getTag("estimated_center", null);
+            if (estimatedCenter != null) {
+                LookupResult result = customAreaLookup.lookup(estimatedCenter.lat,
+                                estimatedCenter.lon);
+                way.setTag("custom_areas", result.getAreas());
+                way.setTag("spatial_rule_set", result.getRuleSet());
+            }
+        }
```
karussell, line 634: I like the separate handling so that the properties are set before the other encoded value (instead of the `insert=true` code), but can you still move the code back into the specific parser (outside of EncodingManager)?

karussell (commented):
Thanks a lot for this PR - I finally managed to start review here :) (sorry again!)

> SpatialRules target CustomAreas now

Can you explain what you mean here?

config-example.yml
otbutz: > Is it possible to move the encoded value definition to `graph.encoded_values`? Then we could define it like `encoded_values=country|limit=63,surface,...` and we have all explicit encoded values in one place and would need to use it for the custom area.

I wasn't sure which option is better. From a configuration standpoint it was nice to have it in one place. But i think you're right and it's easier to configure/maintain if we manage it with the other EVs.

> And is max_bbox still necessary? I think I saw some PR where this requirement was removed, but maybe I mix something up.

It's an optimization which allows to skip country boundaries which are not relevant for the imported dataset. e.g importing an europe.pbf while using the global country boundaries file in our repo.

core/src/main/java/com/graphhopper/config/CustomAreaFile.java
otbutz: Should i merge them and keep `location` in the CustomArea? This field is only used to point to file and is not needed anymore after the border polygons have been loaded in GraphHopperManaged.

core/src/main/java/com/graphhopper/config/CustomArea.java
otbutz: > As this replaces the SpatialRules should we rename the PR to `make spatial rules customizable (CustomArea)`?

The SpatialRules are still there - at least until we fully replace them with CustomModels later on. The only thing that changed is that they don't provide their own border polygons anymore. SpatialRules only get applied now if their ID matches the one of a CustomArea.

> And can we somehow avoid the name clash with the other CustomArea class?

I think you're referring to `com.graphhopper.routing.ev.CustomArea` ? I can move the prefix and EV key creation into the regular CustomArea and delete this one.

> For future reference can you also list the renamed classes here as github somehow fails to catch all some of them. E.g. SpatialRuleLookupJTS, SpatialRuleLookupHelper (?) This would also help which classes I can skip when reviewing.
> 
> And why did you choose to keep some of the classes the same name like SpatialRuleSet? Could it make sense to rename it also to CustomAreaRules or something? Also the name for CountriesSpatialRuleFactory seems not consistent.

CustomAreaRule sounds good. I'm going to rename the remaining classes accordingly.

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
otbutz: Something like this?

```java
    public IntsRef handleWayTags(ReaderWay way, AcceptWay acceptWay, IntsRef relationFlags) {
        IntsRef edgeFlags = createEdgeFlags();
        CustomAreaParser.handleWayTags(edgeFlags, way, acceptWay.isFerry(), relationFlags, customAreaLookup);
        for (TagParser parser : edgeTagParsers) {
            parser.handleWayTags(edgeFlags, way, acceptWay.isFerry(), relationFlags);
        }
        for (AbstractFlagEncoder encoder : edgeEncoders) {
            encoder.handleWayTags(edgeFlags, way, acceptWay.get(encoder.toString()));
        }
        return edgeFlags;
    }
```

config-example.yml
otbutz: @karussell should we keep some kind of prefix for all custom area EVs?

e.g `encoded_values=custom_area.country|limit=63,custom_area.state|limit=127,surface,...`

My current approach is to create EncodedValues with a namespace using `EncodingManger.getKey(String,String)`. This makes it easier to retrieve the correct EV for the PathDetailsBuilder later:

```java
for (String checkPrefix : requestedPathDetails) {
     if (checkPrefix.startsWith(CustomArea.key("")) && evl.hasEncodedValue(checkPrefix))
          builders.add(new StringDetails(checkPrefix, evl.getStringEncodedValue(checkPrefix)));
}
```
Adding a prefix looks a bit verbose but without a it we'd have to treat any unknown key as a custom area encoded value and that would also include typos...


If we want to add other types of "dynamic" EVs in the future, we have to come up with a manageable solution.
We *could* default to creating StringEncodedValues for unknown keys but

core/src/main/java/com/graphhopper/config/CustomAreaFile.java
otbutz: Removing CustomAreaFile leads to the Dropwizard issue i mentioned [earlier](https://github.com/graphhopper/graphhopper/pull/2201#issuecomment-738796938), except that i can't ignore the `customAreas` field as a workaround anymore.

config-example.yml
karussell: > It's an optimization which allows to skip country boundaries which are not relevant for the imported dataset. e.g importing an europe.pbf while using the global country boundaries file in our repo.

Can you explain what you mean here? Do you mean geometries outside of max_bbox will be skipped?

> should we keep some kind of prefix for all custom area EVs?

Yes, problem we need some identification like we have for RouteNetwork but a dot is not possible. I suggest a postfix like there e.g. `country_area`? But then it would be ugly in our script `in_area_modena_area` :D ... so probably we should change it there too? (probably even independent of this PR to merge it fast)

karussell: > in_area_country_area

Ah, no. This would not be a problem. Still I would change it to `in_xy_area` for consistency I guess.

otbutz: > > It's an optimization which allows to skip country boundaries which are not relevant for the imported dataset. e.g importing an europe.pbf while using the global country boundaries file in our repo.
> 
> Can you explain what you mean here? Do you mean geometries outside of max_bbox will be skipped?

Exactly. Editing large geojson files is rather error prone so the user might for example use a big geojson file with lots of features while he only needs a specific subset for the small pbf dataset he wants to import. e.g using our world boundaries file while importing a DACHL pbf

I've just added a commit to also take this into account if the user doesn't provide a `encoded_value_limit`. The size of the resulting encoded value is therefore determined by the number of custom areas intersecting the bounding box.

Note that the bbox is also used to modify the polygons by calculating an intersection:

https://github.com/graphhopper/graphhopper/blob/4f126833b0d962b68e2dc103e748903ec08951ff/core/src/main/java/com/graphhopper/routing/util/area/CustomAreaHelper.java#L110-L123

> 
> > should we keep some kind of prefix for all custom area EVs?
> 
> Yes, probably we need some identification like we have for RouteNetwork but a dot is not possible. I suggest a postfix like there e.g. `country_area`? But then it would be ugly in our script `in_area_country_area` :D ... so probably we should change it there too? (probably even independent of this PR to merge it fast)

`_area` looks good

easbar:
@otbutz, does this PR add anything that is not covered by #2353 or can we close here?

otbutz:
I think we got it all covered.

Issue #1679, opened 2019-08-02
Include elevation for foot/hiking profile


easbar:
The German Alpine Club suggests a formula how elevation affects walking speed, as discussed here: https://discuss.graphhopper.com/t/walking-duration-estimate-elevation-ignored/4621/4

Related: #1597

hungerburg:
You might want to reopen this, the calculation is wrong, it gives the same result, whatever the direction. Sample paths:

- https://graphhopper.com/maps/?point=47.312133%2C11.383832&point=47.29842%2C11.391975&point=47.286413%2C11.398326&locale=de&vehicle=hike&weighting=fastest&elevation=true&turn_costs=false&use_miles=false&layer=TF%20Outdoors
- https://graphhopper.com/maps/?point=47.286413%2C11.398326&point=47.29842%2C11.391975&point=47.312133%2C11.383832&locale=de&vehicle=hike&weighting=fastest&elevation=true&turn_costs=false&use_miles=false&layer=TF%20Outdoors

Downwards looks a good guess; Upwards though is much too quick. Reasonably, the hike should be close to 4hrs, which still is quite an accomplishment, that only can be got at with training, but might match the DAV rate too.

PS: Intermediate marker to take the preferred path, where the hiking route runs, which is even shorter a little

otbutz:
https://github.com/graphhopper/graphhopper/issues/1597#issuecomment-615365980 looks intended?

karussell:
For the PR this was intended, yes. But in general we should make a difference between up+downhill for hiking too.

hungerburg:
Thank you, I know, that there are people that use graphhopper for hiking, so this will be welcome news for sure. Another example close by to help in calibration, nearly the same specs:

- Guidepost - https://www.carto.net/andre.mw/photos/2008/10/05_hoetting_rauschbrunnen/20081005-145836_wegweiser_beim_gasthof_rauschbrunnen.shtml
- Hohe Warte 4:00h  vs 2:29h - https://graphhopper.com/maps/?point=47.279313%2C11.347407&point=47.301614%2C11.33346&locale=de&vehicle=hike&weighting=fastest&elevation=true&turn_costs=false&use_miles=false&layer=TF%20Outdoors
[...]

Issue #1631, opened 2019-06-01
Setting different speeds for both directions can cause problems with loops


easbar:
Routing can fail if we use `MotorCycleEncoder` instead of `CarFlagEncoder` and there are loops (for OSM they are excluded during import though), because motorcycle allows setting different speeds for both directions. See this branch: [`motorcycle_edge_based_ch_bug`]( https://github.com/graphhopper/graphhopper/tree/motorcycle_edge_based_ch_bug)

easbar:
The test only shows the problem for edge-based CH, but I am not sure if it could be a problem for other (edge-based ?) algorithms as well, not sure if this is fully covered by tests.

easbar:
> The test only shows the problem for edge-based CH, but I am not sure if it could be a problem for other (edge-based ?) algorithms as well, not sure if this is fully covered by tests.

Yes, this is a problem for all edge-based algorithms and its nothing specific to edge-based CH. We cannot reliably use different speeds for the two directions for loops. The current solution is still that loops are excluded during the import. Here is a recent commit where `RandomizedRoutingTest` fails sometimes (run it many times): eda2b53e20c935e4d14cd5338f7baebe2321427d

Interestingly the test is 'fixed' by bbb15dfb6b67ace6e060c846ba69fae15641eddb, but this is only because with this commit the speeds aren't properly set to different values due to a changed `randomGraph` method, see this comment: https://github.com/graphhopper/graphhopper/pull/2202#discussion_r541996509

easbar:
In our current model loops with different properties for the two directions are not well defined. Every edge has an edge ID, a base node ID (A) and an adj node ID (B). A loop is an edge where A=B. The direction of an edge is the relation between A and B, but since A=B a loop has only one direction. Therefore setting different edge properties for the 'fwd' and 'bwd' direction of a loop is just wrong. Currently there are a few places (like `DefaultEdgeFilter`) where we do special handling for loop edges, but these would not be necessary if we did not use direction-dependent properties for loop edges in the first place. Can we somehow deny setting different values for fwd/bwd directions on loops?

This should not be confused with loop shortcuts in edge-based CH. The direction of these loop shortcuts is actually well defined, because of the first/last orig edges.

easbar:
Maybe we could replace, for example:

```java
       @Override
        public EdgeIteratorState set(BooleanEncodedValue property, boolean value) {
            property.setBool(reverse, getFlags(), value);
[...]

PR #2202, opened 2020-12-06
removed AbstractFlagEncoder.speedDefault and flagsDefault()

karussell:
It took a bit time, but finally I got rid of this method :) (usage was approx 2000 times). This will likely make it easier to move average_speed and access parsing into a separate TagParser.

My best friend was regex replace functionality of IntelliJ ala:

`graph.edge\((\d+)..(\d+)..(\d+),` (or variations with `(\d+.\d+)` or `(\w+)`)
`GHUtility.setProperties(graph.edge($1,$2).setDistance($3),encoder,60,true,`

The new mechanism is a bit clumsy, but if you think about this then it is really necessary to initialize the edge with a specific speed (and vehicle). And in production we do not use this method, only for tests.

TODO: replace more single method calls of GHUtility with the new `GHUtility.setProperties(List<EdgeIteratorState>)`?

### DIFF
--- a/core/src/main/java/com/graphhopper/util/GHUtility.java
+++ b/core/src/main/java/com/graphhopper/util/GHUtility.java
@@ -623,48 +623,46 @@ public class GHUtility {
         return edgeKey / 2;
     }
 
-    public static IntsRef setProperties(IntsRef edgeFlags, FlagEncoder encoder, double averageSpeed, boolean fwd, boolean bwd) {
-        if (averageSpeed < 0.0001 && (fwd || bwd))
-            throw new IllegalStateException("Zero speed is only allowed if edge will get inaccessible. Otherwise Weighting can produce inconsistent results");
+    public static IntsRef setSpeed(double fwdSpeed, double bwdSpeed, FlagEncoder encoder, IntsRef edgeFlags) {
+        if (fwdSpeed < 0 || bwdSpeed < 0)
+            throw new IllegalArgumentException("Speed must be positive but wasn't! fwdSpeed:" + fwdSpeed + ", bwdSpeed:" + bwdSpeed);
 
         BooleanEncodedValue accessEnc = encoder.getAccessEnc();
-        DecimalEncodedValue avSpeedEnc = encoder.getAverageSpeedEnc();
-        accessEnc.setBool(false, edgeFlags, fwd);
-        accessEnc.setBool(true, edgeFlags, bwd);
-        if (fwd)
-            avSpeedEnc.setDecimal(false, edgeFlags, averageSpeed);
-        if (bwd)
-            avSpeedEnc.setDecimal(true, edgeFlags, averageSpeed);
+        DecimalEncodedValue avgSpeedEnc = encoder.getAverageSpeedEnc();
+
+        avgSpeedEnc.setDecimal(false, edgeFlags, fwdSpeed);
+        if (fwdSpeed > 0)
+            accessEnc.setBool(false, edgeFlags, true);
+
+        if (avgSpeedEnc.isStoreTwoDirections())
+            avgSpeedEnc.setDecimal(true, edgeFlags, bwdSpeed);
+        if (bwdSpeed > 0)
+            accessEnc.setBool(true, edgeFlags, true);
         return edgeFlags;
     }
 
-    public static EdgeIteratorState setProperties(EdgeIteratorState edge, FlagEncoder encoder, double fwdSpeed, double bwdSpeed) {
-        BooleanEncodedValue accessEnc = encoder.getAccessEnc();
-        edge.set(accessEnc, true).setReverse(accessEnc, true);
-        DecimalEncodedValue avSpeedEnc = encoder.getAverageSpeedEnc();
-        return edge.set(avSpeedEnc, fwdSpeed).setReverse(avSpeedEnc, bwdSpeed);
+    public static void setSpeed(double fwdSpeed, double bwdSpeed, FlagEncoder encoder, EdgeIteratorState... edges) {
+        setSpeed(fwdSpeed, bwdSpeed, encoder, Arrays.asList(edges));
     }
 
-    public static IntsRef setProperties(IntsRef edgeFlags, FlagEncoder encoder, double fwdSpeed, double bwdSpeed) {
+    public static void setSpeed(double fwdSpeed, double bwdSpeed, FlagEncoder encoder, Collection<EdgeIteratorState> edges) {
+        if (fwdSpeed < 0 || bwdSpeed < 0)
+            throw new IllegalArgumentException("Speed must be positive but wasn't! fwdSpeed:" + fwdSpeed + ", bwdSpeed:" + bwdSpeed);
         BooleanEncodedValue accessEnc = encoder.getAccessEnc();
-        accessEnc.setBool(false, edgeFlags, true);
-        accessEnc.setBool(true, edgeFlags, true);
-        DecimalEncodedValue avSpeedEnc = encoder.getAverageSpeedEnc();
-        avSpeedEnc.setDecimal(false, edgeFlags, fwdSpeed);
-        avSpeedEnc.setDecimal(true, edgeFlags, bwdSpeed);
-        return edgeFlags;
-    }
+        DecimalEncodedValue avgSpeedEnc = encoder.getAverageSpeedEnc();
+        for (EdgeIteratorState edge : edges) {
+            edge.set(avgSpeedEnc, fwdSpeed);
+            if (fwdSpeed > 0)
+                edge.set(accessEnc, true);
 
-    public static void setProperties(List<EdgeIteratorState> edges, EncodingManager em, boolean fwd, boolean bwd) {
-        for (FlagEncoder encoder : em.fetchEdgeEncoders()) {
-            double s = encoder.getMaxSpeed() / 2;
-            for (EdgeIteratorState edge : edges) {
-                GHUtility.setProperties(edge, encoder, s, fwd, bwd);
-            }
+            if (avgSpeedEnc.isStoreTwoDirections())
+                edge.setReverse(avgSpeedEnc, bwdSpeed);
+            if (bwdSpeed > 0)
+                edge.setReverse(accessEnc, true);
         }
     }
 
-    public static EdgeIteratorState setProperties(EdgeIteratorState edge, FlagEncoder encoder, double averageSpeed, boolean fwd, boolean bwd) {
+    public static EdgeIteratorState setSpeed(double averageSpeed, boolean fwd, boolean bwd, FlagEncoder encoder, EdgeIteratorState edge) {
         if (averageSpeed < 0.0001 && (fwd || bwd))
             throw new IllegalStateException("Zero speed is only allowed if edge will get inaccessible. Otherwise Weighting can produce inconsistent results");
 
--- a/core/src/main/java/com/graphhopper/util/GHUtility.java
+++ b/core/src/main/java/com/graphhopper/util/GHUtility.java
@@ -168,8 +168,11 @@ public class GHUtility {
                 "graph.edge(%d, %d, %f, %s); // edgeId=%s\n", from, to, edge.getDistance(), fwd && bwd ? "true" : "false", edge.getEdge());
     }
 
+    /**
+     * @param speed if null a random speed will be assign to every edge
+     */
     public static void buildRandomGraph(Graph graph, Random random, int numNodes, double meanDegree, boolean allowLoops,
-                                        boolean allowZeroDistance, DecimalEncodedValue randomSpeedEnc,
+                                        boolean allowZeroDistance, BooleanEncodedValue accessEnc, DecimalEncodedValue speedEnc, Double speed,
                                         double pNonZeroLoop, double pBothDir, double pRandomDistanceOffset) {
         if (numNodes < 2 || meanDegree < 1) {
             throw new IllegalArgumentException("numNodes must be >= 2, meanDegree >= 1");
@@ -204,14 +207,13 @@ public class GHUtility {
             maxDist = Math.max(maxDist, distance);
             // using bidirectional edges will increase mean degree of graph above given value
             boolean bothDirections = random.nextDouble() < pBothDir;
-            EdgeIteratorState edge = graph.edge(from, to, distance, bothDirections);
-            double fwdSpeed = 10 + random.nextDouble() * 110;
-            double bwdSpeed = 10 + random.nextDouble() * 110;
-            if (randomSpeedEnc != null) {
-                edge.set(randomSpeedEnc, fwdSpeed);
-                if (randomSpeedEnc.isStoreTwoDirections())
-                    edge.setReverse(randomSpeedEnc, bwdSpeed);
-            }
+            EdgeIteratorState edge = graph.edge(from, to).setDistance(distance).set(accessEnc, true);
+            if (bothDirections) edge.setReverse(accessEnc, true);
+            double fwdSpeed = speed != null ? speed : 10 + random.nextDouble() * 110;
+            double bwdSpeed = speed != null ? speed : 10 + random.nextDouble() * 110;
+            edge.set(speedEnc, fwdSpeed);
+            if (speedEnc.isStoreTwoDirections())
+                edge.setReverse(speedEnc, bwdSpeed);
             numEdges++;
         }
         LOGGER.debug(String.format(Locale.ROOT, "Finished building random graph" +
@@ -651,6 +653,15 @@ public class GHUtility {
         avSpeedEnc.setDecimal(false, edgeFlags, fwdSpeed);
         avSpeedEnc.setDecimal(true, edgeFlags, bwdSpeed);
         return edgeFlags;
+    }
+
+    public static void setProperties(List<EdgeIteratorState> edges, EncodingManager em, boolean fwd, boolean bwd) {
+        for (FlagEncoder encoder : em.fetchEdgeEncoders()) {
+            double s = encoder.getMaxSpeed() / 2;
+            for (EdgeIteratorState edge : edges) {
+                GHUtility.setProperties(edge, encoder, s, fwd, bwd);
+            }
+        }
     }
 
     public static EdgeIteratorState setProperties(EdgeIteratorState edge, FlagEncoder encoder, double averageSpeed, boolean fwd, boolean bwd) {
--- a/core/src/main/java/com/graphhopper/util/GHUtility.java
+++ b/core/src/main/java/com/graphhopper/util/GHUtility.java
@@ -819,6 +819,11 @@ public class GHUtility {
         }
 
         @Override
+        public EdgeIteratorState set(BooleanEncodedValue property, boolean fwd, boolean bwd) {
+            throw new UnsupportedOperationException("Not supported. Edge is empty.");
+        }
+
+        @Override
         public int get(IntEncodedValue property) {
             throw new UnsupportedOperationException("Not supported. Edge is empty.");
         }
@@ -835,6 +840,11 @@ public class GHUtility {
 
         @Override
         public EdgeIteratorState setReverse(IntEncodedValue property, int value) {
+            throw new UnsupportedOperationException("Not supported. Edge is empty.");
+        }
+
+        @Override
+        public EdgeIteratorState set(IntEncodedValue property, int fwd, int bwd) {
             throw new UnsupportedOperationException("Not supported. Edge is empty.");
         }
 
@@ -859,6 +869,11 @@ public class GHUtility {
         }
 
         @Override
+        public EdgeIteratorState set(DecimalEncodedValue property, double fwd, double bwd) {
+            throw new UnsupportedOperationException("Not supported. Edge is empty.");
+        }
+
+        @Override
         public <T extends Enum> T get(EnumEncodedValue<T> property) {
             throw new UnsupportedOperationException("Not supported. Edge is empty.");
         }
@@ -875,6 +890,11 @@ public class GHUtility {
 
         @Override
         public <T extends Enum> EdgeIteratorState setReverse(EnumEncodedValue<T> property, T value) {
+            throw new UnsupportedOperationException("Not supported. Edge is empty.");
+        }
+
+        @Override
+        public <T extends Enum> EdgeIteratorState set(EnumEncodedValue<T> property, T fwd, T bwd) {
             throw new UnsupportedOperationException("Not supported. Edge is empty.");
         }
 
--- a/core/src/main/java/com/graphhopper/util/GHUtility.java
+++ b/core/src/main/java/com/graphhopper/util/GHUtility.java
@@ -634,8 +634,12 @@ public class GHUtility {
         if (fwdSpeed > 0)
             accessEnc.setBool(false, edgeFlags, true);
 
-        if (avgSpeedEnc.isStoreTwoDirections())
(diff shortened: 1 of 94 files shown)


core/src/test/java/com/graphhopper/routing/AlternativeRouteCHTest.java
```diff
@@ -57,23 +58,23 @@ public GraphHopperStorage createTestGraph(EncodingManager tmpEM) {
...
+        GHUtility.setProperties(graph.edge(1, 9).setDistance(10000), carFE, 60, true, true);
+        GHUtility.setProperties(graph.edge(9, 2).setDistance(10000), carFE, 60, true, true);
+        GHUtility.setProperties(graph.edge(2, 3).setDistance(10000), carFE, 60, true, true);
 
-        graph.edge(4, 11, 9000, true);
-        graph.edge(11, 12, 9000, true);
-        graph.edge(12, 10, 10000, true);
+        GHUtility.setProperties(graph.edge(4, 11).setDistance(9000), carFE, 60, true, true);
+        GHUtility.setProperties(graph.edge(11, 12).setDistance(9000), carFE, 60, true, true);
+        GHUtility.setProperties(graph.edge(12, 10).setDistance(10000), carFE, 60, true, true);
```
easbar, line 77: Hm I think for tests like this `graph.edge(b, b, dist, reverse)` was quite useful? Can we not keep it somehow?

easbar: At least I would add `GHUtility.edge(graph, a, b, dist, bidir)`?

core/src/main/java/com/graphhopper/util/GHUtility.java
```diff
@@ -653,6 +655,15 @@ public static IntsRef setProperties(IntsRef edgeFlags, FlagEncoder encoder, doub
         return edgeFlags;
     }
 
+    public static void setProperties(List<EdgeIteratorState> edges, EncodingManager em, boolean fwd, boolean bwd) {
```
easbar, line 658: The name of this method seems rather confusing. The old `setProperties` method should probably be called `setAccessAndSpeed`? And this new one just sets the access and speed for all encoders, but uses half the max speed as speed? So its something like `setAccessAndHalfTheMaxSpeedForAllEncoders?`

Btw. what I also think is confusing is that we have `graph.edge(a, b)` and not `graph.edge(a, b, dist)`. Do we ever need an edge without a distance? Maybe distance should go in these `setProperties` kind of methods as well then?

core/src/test/java/com/graphhopper/reader/dem/BridgeElevationInterpolatorTest.java
```diff
@@ -75,17 +73,19 @@ public void interpolatesElevationOfPillarNodes() {
...
-        EdgeIteratorState edge34 = graph.edge(3, 4, 10, true);
-        EdgeIteratorState edge56 = graph.edge(5, 6, 10, true);
-        EdgeIteratorState edge67 = graph.edge(6, 7, 10, true);
-        EdgeIteratorState edge78 = graph.edge(7, 8, 10, true);
-        EdgeIteratorState edge89 = graph.edge(8, 9, 10, true);
-        EdgeIteratorState edge17 = graph.edge(1, 7, 10, true);
-        EdgeIteratorState edge27 = graph.edge(2, 7, 10, true);
-        EdgeIteratorState edge37 = graph.edge(3, 7, 10, true);
+        FlagEncoder encoder = encodingManager.getEncoder("car");
+        EdgeIteratorState edge01 = GHUtility.setProperties(graph.edge(0, 1).setDistance(10), encoder, 60, true, true);
```
easbar, line 77: What I like here is that the speed is now explicit, while previously one always had to keep in mind that its 60 in these tests. Maybe I just have to get used to this verbose code to create a testing graph, but also see my other comment about `setProperties` which I do not find intuitive.

core/src/test/java/com/graphhopper/routing/AlternativeRouteTest.java
```diff
@@ -69,28 +71,28 @@ public AlternativeRouteTest(TraversalMode tMode) {
...
          \   /   \
          5--6-7---8
         
          */
-        graph.edge(1, 9, 1, true);
-        graph.edge(9, 2, 1, true);
-        graph.edge(2, 3, 1, true);
-        graph.edge(3, 4, 1, true);
-        graph.edge(4, 10, 1, true);
+        GHUtility.setProperties(graph.edge(1,9).setDistance(1),encoder,60,true, true);
```
easbar, line 82: We should probably run code formatting here. For example you can undo all changes and commit them again with the "format  on commit" checkbox enabled in intellij.

core/src/test/java/com/graphhopper/storage/TurnCostStorageTest.java
```diff
@@ -52,12 +53,13 @@ public void setup() {
...
-        g.edge(1, 3, 1, true);
-        g.edge(2, 3, 1, true);
-        g.edge(2, 4, 1, true);
+    public static void initGraph(GraphHopperStorage g) {
+        GHUtility.setProperties(Arrays.asList(
+                g.edge(0, 1).setDistance(3),
+                g.edge(0, 2).setDistance(1),
+                g.edge(1, 3).setDistance(1),
+                g.edge(2, 3).setDistance(1),
+                g.edge(2, 4).setDistance(1)), g.getEncodingManager(), true, true);
```
easbar, line 62: I am not sure yet if this is an improvement compared to several calls to setProperties? Especially setting `speed=maxspeed/2`  seems not very obvious (or maybe I do not understand it yet)?

core/src/main/java/com/graphhopper/util/GHUtility.java
karussell: Ah, ok. Sure.

> Do we ever need an edge without a distance?

Yes. Barrier nodes and I think also for PT we do not set a distance.

core/src/test/java/com/graphhopper/routing/AlternativeRouteCHTest.java
karussell: It is not easy as the information which encoder should be used is required.

core/src/test/java/com/graphhopper/reader/dem/BridgeElevationInterpolatorTest.java
karussell: I also do not find it intuitive and for tests we should think about a better solution. Maybe we can use the method you proposed above but with a GraphHopperStorage which gives us access to the EncodingManger where we then by default fetch car?

core/src/test/java/com/graphhopper/routing/AlternativeRouteTest.java
karussell: Ok, will do. (I usually do not use IntelliJ for git)

easbar: I just meant this is an easy way to apply the formatting to the changed files only. I think we never fixed the formatting for the entire project so its hard to run it everywhere ;)

core/src/test/java/com/graphhopper/storage/TurnCostStorageTest.java
karussell: The idea was basically to have an alternative to the previously used `graph.edge(from, to, dist, bothDir)` (like you said before it was pretty useful). So we set some default speed for *all* encoders. And `maxspeed/2` is approximatively the previously used default speed for all encoders.

We should rename the method as a first step. But I'm unsure if we should make it more explicit as it will then get more verbose?

core/src/test/java/com/graphhopper/routing/AlternativeRouteCHTest.java
easbar: Ah true we would need `GHUtility.edge(graphHopperStorage...)` then. Hm, maybe we do not need this after all when we tweak the GHUtility method(s) a bit.

core/src/main/java/com/graphhopper/util/GHUtility.java
easbar: What is the reason for `speed=maxspeed/2`?

> Yes. Yes, for barrier nodes and I think also for certain PT edges we do not set a distance.

For barrier nodes there is a distance (its just  zero). For PT I am not sure. But if its just zero (but exists in storage) also for PT I would rather add the distance parameter and set it to zero than have this `setDistance()` method. Even when we do not have it we could add `graph.edge(a, b, dist)` additionally?

core/src/test/java/com/graphhopper/storage/TurnCostStorageTest.java
easbar: Can we not just remove it entirely?

karussell: What do you want to remove entirely? The method? I would rather expand its usage and make speed+encoder more explicit? I like it more as the edge creation is less hidden within GHUtility calls.

core/src/main/java/com/graphhopper/util/GHUtility.java
karussell: > What is the reason for speed=maxspeed/2?

Some arbitrary speed we use to initialized this. Similar to flagEncoder.speedDefault before. But surely we can get rid of this. See my comment below.

> For PT I am not sure. But if its just zero (but exists in storage) also for PT I would rather add the distance parameter

What I meant is that distance is just an edge property like speed and for certain edges of an edge-expanded model in PT the distance property does not make sense, not even 0.

easbar: > What I meant is that distance is just an edge property like speed 

Yes excepts its not stored in flags?

>  for certain edges of an edge-expanded model in PT the distance property does not make sense, not even 0.

This is interesting because it means we could save memory for PT if we removed the distance field (if this is relevant in any way).

Would it be a problem to add `graph.edge(a, b, dist)`?

core/src/test/java/com/graphhopper/storage/TurnCostStorageTest.java
easbar: > What do you want to remove entirely? The method? 

Yes, `GHUtility.setProperties(List<EdgeIteratorState>...` and setting maxspeed/2

core/src/main/java/com/graphhopper/util/GHUtility.java
karussell: > Yes excepts its not stored in flags?

Currently an implementation detail :D 

> This is interesting because it means we could save memory for PT if we removed the distance field (if this is relevant in any way).

Yes, definitely. (btw it is time-expanded not edge-expanded like I said before)

> Would it be a problem to add graph.edge(a, b, dist)?

Not sure. IMO it is not a method that belongs to the Graph interface.

core/src/test/java/com/graphhopper/storage/TurnCostStorageTest.java
easbar: ... or maybe even (thinking out loud now):

```java
EdgeIteratorState edge = graph.edge(3, 5).setDistance(200).set(accessEnc, true, false).set(speedEnc, 60, 60);
```

i.e. add setter methods for encoded values to set both directions with one call.

core/src/main/java/com/graphhopper/util/GHUtility.java
easbar: Ok convinced, lets keep it as is and use `setDistance()`.

core/src/test/java/com/graphhopper/storage/TurnCostStorageTest.java
easbar: I think I favor the last option I proposed right now because it is very explicit but not even that verbose. I am just not sure how `set(encValue, boolean, boolean)` behaves for encoded values that are do not support two different directions (like road_class). Would be we be able to throw an error when we call `set(encValue, fwd, bwd)`, `encValue` is not bidirectional and `fwd!=bwd`?

karussell: > `set(encValue, boolean, boolean)`

Sounds like a good solution.

> I am just not sure how set(encValue, boolean, boolean) behaves for encoded values that do not support two different directions

This should work the way you described

> Its only used in three places.

Only because I introduced it very late in the process as I stumbled over the fact that often multiple edges are initialized with the same speeds and IMO this "bulk" method improves readability of the edge creation.

What about a method like:

`setSpeeds(Double fwdSpeed, Double bwdSpeed, FlagEncoder encoder, EdgeIteratorState... edges)`

It is short and less verbose for many edges. 

And it would be already compatible to a version without "access"-bits. With that I mean the following: currently we have the "access"-bits defined only to use less bits (2 bits + speed bits intead of 2*speed_bits) and something that only the Weighting can define. Hmmh, or before or additionally to #1835 it might be interesting to investigate if we in general move away from access bits and just use `2*speed_bits`? So for higher precision of 7 bits this means 5 bits more per vehicle profile but even for hiking or walking this would make sense (#1679). Not sure.

easbar: > setSpeeds(Double fwdSpeed, Double bwdSpeed, FlagEncoder encoder, EdgeIteratorState... edges)

Where would you put this method and can you make give an example for its usage? 

> And it would be already compatible to a version without "access"-bits. 

Ah because `fwd/bwdSpeed=infinity` would mean 'no access'? 

>  if we in general move away from access bits

This is a hard question, but so far it seems useful to me to e.g. separate motorized and outdoor profiles even without specifying the weighting already. Is this something we have to decide here? My guess is the refactoring here will not make it harder to remove the access flags. So yes we can introduce a helper method that already hides them, but we can just as well do this when we actually make this move? Delaying this until later seems to be the better option to me because we do not really know what we will do yet :)

karussell: > Ah because fwd/bwdSpeed=infinity would mean 'no access'?

yes (for speed==0)

> Is this something we have to decide here?

No :) ... just thought that having a method without access bits specified is of an advantage for the future :)

> Delaying this until later seems to be the better option to me because we do not really know what we will do yet :)

ok

> Where would you put this method and can you give an example for its usage?

Hmmh, it does not work well with the encoder list it seems, but with one encoder it would be a bit more powerful than the existing setProperties method (as it has two speed parameters and allows multiple edges) and still is not more verbose. This method is used +1000 times and we can clean up some of its usages like in BridgeElevationInterpolatorTest, where we currently have:

```java
EdgeIteratorState edge01 = GHUtility.setProperties(graph.edge(0, 1).setDistance(10), encoder, 60, true, true);
EdgeIteratorState edge12 = GHUtility.setProperties(graph.edge(1, 2).setDistance(10), encoder, 60, true, true);
EdgeIteratorState edge23 = GHUtility.setProperties(graph.edge(2, 3).setDistance(10), encoder, 60, true, true);
EdgeIteratorState edge34 = GHUtility.setProperties(graph.edge(3, 4).setDistance(10), encoder, 60, true, true);
EdgeIteratorState edge56 = GHUtility.setProperties(graph.edge(5, 6).setDistance(10), encoder, 60, true, true);
EdgeIteratorState edge67 = GHUtility.setProperties(graph.edge(6, 7).setDistance(10), encoder, 60, true, true);
EdgeIteratorState edge78 = GHUtility.setProperties(graph.edge(7, 8).setDistance(10), encoder, 60, true, true);
EdgeIteratorState edge89 = GHUtility.setProperties(graph.edge(8, 9).setDistance(10), encoder, 60, true, true);
EdgeIteratorState edge17 = GHUtility.setProperties(graph.edge(1, 7).setDistance(10), encoder, 60, true, true);
EdgeIteratorState edge27 = GHUtility.setProperties(graph.edge(2, 7).setDistance(10), encoder, 60, true, true);
EdgeIteratorState edge37 = GHUtility.setProperties(graph.edge(3, 7).setDistance(10), encoder, 60, true, true);
```

and could replace it with:

```java
EdgeIteratorState edge01, edge12, edge23, edge34, edge56, edge67, edge78, edge89, edge17, edge27, edge37;
GHUtility.setProperties(60, 60, encoder,
        edge01 = graph.edge(0, 1).setDistance(10),
        edge12 = graph.edge(1, 2).setDistance(10),
        edge23 = graph.edge(2, 3).setDistance(10),
        edge34 = graph.edge(3, 4).setDistance(10),
        edge56 = graph.edge(5, 6).setDistance(10),
        edge67 = graph.edge(6, 7).setDistance(10),
        edge78 = graph.edge(7, 8).setDistance(10),
        edge89 = graph.edge(8, 9).setDistance(10),
        edge17 = graph.edge(1, 7).setDistance(10),
        edge27 = graph.edge(2, 7).setDistance(10),
        edge37 = graph.edge(3, 7).setDistance(10));
```

easbar: Ok, either way seems fine. Is it a lot of work to turn the single edge creations into calls to `(GHUtility.setSpeeds(fwdSpeed, bwdSpeed, encoder, EdgeIteratorState...)`? I am happy with [this](https://github.com/graphhopper/graphhopper/pull/2202#discussion_r537024992) format and maybe only use `GHUtility.setSpeeds` later in places we feel it makes sense?

karussell: Yes, it is a lot of work to migrate the current stuff. Probably we should follow a similar approach like migrating to junit5 :)

> I am happy with this format

Should I introduce the methods `set(encValue, fwd, bwd)` in this PR or later?

easbar: > Yes, it is a lot of work to migrate the current stuff. Probably we should follow a similar approach like migrating to junit5 :)

I agree.

> Should I introduce the methods set(encValue, fwd, bwd) in this PR or later?

Is it harder or easier to search/replace GHUtility.setProperties here than it would be to search/replace graph.edge and do change it to set(encValue, fwd, bwd) in master? Hm yes why not do it here directly?

karussell: When migrating the current code to:

```java
EdgeIteratorState edge = graph.edge(3, 5).setDistance(200).set(accessEnc, true, true).set(speedEnc, 60, 60);
```

there are two problems: if we throw an explicit error we cannot easily refactor the current usages as `speedEnc` is often not for two directions. Except we set the backward value in a silent manner which I did in https://github.com/graphhopper/graphhopper/commit/418454a9c9e070a283e1fab2037b2f11bd51f50f#diff-305b1d577b70e729c3845b447da93e1bc9668716a1c4b2e0928987b8623380b0R174, but I do not like it that much for our Java API.

And the code will be longer:

```java
EdgeIteratorState edge = graph.edge(3, 5).setDistance(200).set(encoder.getAccessEnc(), true, true).set(encoder.getSpeedEnc(), 60, 60);
```

So for now I would prefer to keep it (simpler?) like:

```java
EdgeIteratorState edge = GHUtilities.setSpeed(60, 60, encoder, graph.edge(3, 5).setDistance(200));
```

This code then uses the implicit setReverse:

```java
if (avgSpeedEnc.isStoreTwoDirections()) edge.setReverse(avgSpeedEnc, bwdSpeed);
```

which I find acceptable for the utility function and it has also the advantage that no access encoded value is mentioned and a speed of 0 would indicate that the direction is blocked.

Tricky stuff :)

easbar: > Except we set the backward value in a silent manner 

Yeah that's not ideal, for example setting `.set(speedEnc, 60, 40)` should be an error if there is only one direction?
And I guess you right if we do just search&replace we a) do not know if the encoder is bidirectional and b) we do not even know if speed=60 is correct because that depends on the encoder? But maybe we can assume its a car encoder with one direction and speed 60 and only fix the remaining (hopefully not so many) cases?

> So for now I would prefer to keep it (simpler?) like: 

Yes sure if it makes the refactoring easier I would also use the utility method for now.

karussell: > Yes sure if it makes the refactoring easier I would also use the utility method for now.

Yes, this 'regex refactoring' was relative time intense :)

karussell:
I have also committed the changes with formatting like you suggested. Is this in a shape where I can merge this or should we e.g. remove the addition of the new set methods? If we keep them I'll add a few tests at least :)

easbar (approved):
Looks good and I would keep the setter methods.

karussell:
Ok, replaced a few places so that `set(enc, fwd, bwd)` is used in tests too.

btw: I like the cleanup here :) https://github.com/graphhopper/graphhopper/pull/2202/commits/84f673a09747df92c0e66428aad8484f2f2060c9#diff-9fa225a474dad457f320fc9e51c0337f69b560f7c9b676387e8d2155d8ef1d21

core/src/main/java/com/graphhopper/util/GHUtility.java
```diff
@@ -168,8 +168,11 @@ private static void printUnitTestEdge(FlagEncoder encoder, EdgeIteratorState edg
                 "graph.edge(%d, %d, %f, %s); // edgeId=%s\n", from, to, edge.getDistance(), fwd && bwd ? "true" : "false", edge.getEdge());
     }
 
+    /**
+     * @param speed if null a random speed will be assign to every edge
+     */
     public static void buildRandomGraph(Graph graph, Random random, int numNodes, double meanDegree, boolean allowLoops,
-                                        boolean allowZeroDistance, DecimalEncodedValue randomSpeedEnc,
+                                        boolean allowZeroDistance, BooleanEncodedValue accessEnc, DecimalEncodedValue speedEnc, Double speed,
                                         double pNonZeroLoop, double pBothDir, double pRandomDistanceOffset) {
```
easbar, line 176: Why did you add the `speed` parameter here? The previous behavior was that the default speed was used if `randomSpeedEnc` was null and random speeds were generated otherwise. In case `randomSpeedEnc.isStoreTwoDirections` different speeds were used for fwd/bwd direction. Ok now the default speed is gone we need some speed parameter (or just set it to 60 inside this method?), but we should still use random speeds were we did before. Right now speed is `!= null` almost everywhere. Also its a bit problematic to change this method because it might break older seeds (that we should not rely on, but I think in some cases we still do...)

karussell: I did not intend to change this and tried to be careful ... likely I misunderstand something. The intention was to call with speed=null where before it was called with randomSpeedEnc=null. 

> Right now speed is != null almost everywhere.

Only for EdgeBasedTarjanSCCTest there were calls with randomSpeedEnc=null. Those two calls are done via speed=null now. 

> Also its a bit problematic to change this method because it might break older seeds

Hmmh, `random.nextDouble()` is still called regardless of whether it is used like before. For me currently it is not obvious where I changed behaviour. Probably we need a test for this :) ? I'll have a look as you observed behaviour change due to this in #1631

> but we should still use random speeds were we did before

I checked twice but if speed is null the speeds are random (which also feels more logical than before).

easbar: > I did not intend to change this and tried to be careful ...

Yeah, I saw. I also checked and though its the same. But now I realized that once we set `speed != null` we always have `fwdSpeed==bwdSpeed`, while before we had `fwdSpeed!=bwdSpeed` in case `speedEnc!=null` :roll_eyes: ... Ugh I will have a look again tomorrow ;)

PR #2204, opened 2020-12-08
StringEncodedValues

otbutz:
* minor code cleanup: add generic parameters to Enum references
* add StringEncodedValue which encodes Strings based on their index within a predefined list
* StringDetails for retrieval via PathDetails

### DIFF
--- a/core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
+++ b/core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
@@ -191,6 +191,16 @@ public interface EdgeIteratorState {
     <T extends Enum<?>> EdgeIteratorState setReverse(EnumEncodedValue<T> property, T value);
 
     <T extends Enum<?>> EdgeIteratorState set(EnumEncodedValue<T> property, T fwd, T bwd);
+    
+    String get(StringEncodedValue property);
+    
+    EdgeIteratorState set(StringEncodedValue property, String value);
+    
+    String getReverse(StringEncodedValue property);
+    
+    EdgeIteratorState setReverse(StringEncodedValue property, String value);
+    
+    EdgeIteratorState set(StringEncodedValue property, String fwd, String bwd);
 
     String getName();
 
--- a/core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
+++ b/core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
@@ -182,15 +182,15 @@ public interface EdgeIteratorState {
 
     EdgeIteratorState set(DecimalEncodedValue property, double fwd, double bwd);
 
-    <T extends Enum> T get(EnumEncodedValue<T> property);
+    <T extends Enum<?>> T get(EnumEncodedValue<T> property);
 
-    <T extends Enum> EdgeIteratorState set(EnumEncodedValue<T> property, T value);
+    <T extends Enum<?>> EdgeIteratorState set(EnumEncodedValue<T> property, T value);
 
-    <T extends Enum> T getReverse(EnumEncodedValue<T> property);
+    <T extends Enum<?>> T getReverse(EnumEncodedValue<T> property);
 
-    <T extends Enum> EdgeIteratorState setReverse(EnumEncodedValue<T> property, T value);
+    <T extends Enum<?>> EdgeIteratorState setReverse(EnumEncodedValue<T> property, T value);
 
-    <T extends Enum> EdgeIteratorState set(EnumEncodedValue<T> property, T fwd, T bwd);
+    <T extends Enum<?>> EdgeIteratorState set(EnumEncodedValue<T> property, T fwd, T bwd);
 
     String getName();
 
--- a/core/src/main/java/com/graphhopper/util/details/StringDetails.java
+++ b/core/src/main/java/com/graphhopper/util/details/StringDetails.java
@@ -1,0 +1,48 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.util.details;
+
+import com.graphhopper.routing.ev.StringEncodedValue;
+import com.graphhopper.util.EdgeIteratorState;
+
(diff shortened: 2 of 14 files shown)


karussell:
Would it make sense to make this dynamic? I.e. only provide a maximum number of strings at construction time and feed strings as they come while import? This way we could easier introduce new tags. Hmmh, but it would also mean that a smaller area is fine and the larger area requires a higher number and breaks.

core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
```diff
@@ -171,13 +171,21 @@ public void setBool(boolean reverse, IntsRef ref, boolean value) {
...
-    <T extends Enum> EdgeIteratorState setReverse(EnumEncodedValue<T> property, T value);
+    <T extends Enum<?>> EdgeIteratorState setReverse(EnumEncodedValue<T> property, T value);
+    
+    String get(StringEncodedValue property);
+    
+    EdgeIteratorState set(StringEncodedValue property, String value);
+    
+    String getReverse(StringEncodedValue property);
+    
+    EdgeIteratorState setReverse(StringEncodedValue property, String value);
```
karussell, line 188: Here also the new method `set(property, fwd, bwd)` would be required now

core/src/main/java/com/graphhopper/util/details/StringDetails.java
```diff
@@ -0,0 +1,48 @@
...
+    private String currentVal = null;
+
+    public StringDetails(String name, StringEncodedValue ev) {
+        super(name);
+        this.ev = ev;
+    }
+
+    @Override
+    protected Object getCurrentValue() {
+        return currentVal;
```
karussell, line 35: Here we would need to avoid `null` values (due to JSON limitation). See DecimalDetails how we do this for infinity.

otbutz: Are you sure about that? I can't find a call to the DecimalDetails constructor which doesn't just pass `null` as the infinity value.

karussell: Ups. You are right. I misremembered the illegal chars. It is infinity and NaN, not infinity and null :)

otbutz:
> Would it make sense to make this dynamic? I.e. only provide a maximum number of strings at construction time and feed strings as they come while import? This way we could easier introduce new tags. Hmmh, but it would also mean that a smaller area is fine and the larger area requires a higher number and breaks.

I though a bit about dynamic approaches but encountered a few problems:
* we need the number of bits at construction time so we can't really safe space
* if we'd solve this somehow and only encode values we encounter while importing, we would still need to store the encountered values. Otherwise it would be impossible to recreate the StringEncodedValue when loading from an already existing graph.

I dodged the problem in my custom area PR by providing two filter mechanisms(bbox+id) which should allow to conveniently limit the number of areas without having to deal with GeoJSON file itself.

core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
otbutz: Rebased on master

otbutz:
Anything missing?

karussell:
> we need the number of bits at construction time so we can't really safe space

If bits are full we could either throw an exception or everything goes into a special "OTHER" string (?)

> we would still need to store the encountered values

Yes, it would be necessary to serialize this into the StorableProperties.

But let us do this later in a separate issue (if required at all). 

Can you add an integration test where you collect e.g. "surface" or some other string tag via one of our test OSM PBFs? Because it might be useful to introduce some automatic normalization features like white space removal or lower case conversion etc? Hmmh, maybe we don't add such a test here and have a real integration test using the new Janino scripting (an important optimization there would be to do the comparison based on the index not on the String).

otbutz:
> If bits are full we could either throw an exception

From my custom area PR:

```yaml
   custom_area_files:
       id_field: "ISO3166-1:alpha3"
       encoded_value: country
       filter: deu,aut
       max_bbox: -180,180,-90,90
```

Would you add an extra option to define the maximum number of encoded entries?

> or everything goes into a special "OTHER" string (?)

I fear that this might cause erratic behavior and problems which are hard to reproduce.

e.g if we encode countries and reach our maximum number of entries, the resulting EV would completely depend on the order of the processed ways

> Yes, it would be necessary to serialize this into the StorableProperties.

Don't we store the toString representation of the EncodedValues? Maybe we should extend this to a load/store mechanic.

> Can you add an integration test where you collect e.g. "surface" or some other string tag via one of our test OSM PBFs? Because it might be useful to introduce some automatic normalization features like white space removal or lower case conversion etc?

The first usecase would be custom areas and i think it's ok to expect the IDs in the GeoJSON to be already normalized.

>Hmmh, maybe we don't add such a test here and have a real integration test using the new Janino scripting (an important optimization there would be to do the comparison based on the index not on the String).

Not sure if feasible but you could lookup the String within the EncodedValue array once and rely on reference checks later on if you want to avoid String.equals()

otbutz:
We would also need to switch to a hashmap based approach as the Strings passed in via setString() are unordered, so we can't rely on binary search.

otbutz:
Something like this?

```java
public final class StringEncodedValue extends UnsignedIntEncodedValue {
    private final List<String> values;
    private final ObjectIntMap<String> indexMap;
    private final int maxEntries;
    
    public StringEncodedValue(String name, int maxEntries) {
        this(name, maxEntries, false);
    }
    
    public StringEncodedValue(String name, int maxEntries, List<String> values, boolean storeTwoDirections) {
        super(name, 32 - Integer.numberOfLeadingZeros(maxEntries), storeTwoDirections);
        
        this.values = new ArrayList<>(values);
        this.indexMap = new ObjectIntHashMap<>(values.size());
        int index = 1;
        for (String value : values) {
            indexMap.put(value, index++);
        }
        this.maxEntries = maxEntries;
    }
    
    public StringEncodedValue(String name, int maxEntries, boolean storeTwoDirections) {
        super(name, 32 - Integer.numberOfLeadingZeros(maxEntries), storeTwoDirections);
        
        this.values = new ArrayList<>(maxEntries);
        this.indexMap = new ObjectIntHashMap<>(maxEntries);
        this.maxEntries = maxEntries;
    }

    public final void setString(boolean reverse, IntsRef ref, String value) {
        int index = indexMap.get(value);
        if (index == 0) {
            if (values.size() == maxEntries) {
                throw new IllegalStateException("Maximum number of entries reached for " + getName() + ": " + maxEntries);
            }
            values.add(value);
            index = values.size();
            indexMap.put(value, index);
        }
        super.setInt(reverse, ref, index);
    }

    public final String getString(boolean reverse, IntsRef ref) {
        int value = super.getInt(reverse, ref);
        if (value == 0) {
            return null;
        }
        return values.get(value);
    }

    public List<String> getValues() {
        return Collections.unmodifiableList(values);
    }

    @Override
    public boolean equals(Object obj) {
        if (this == obj) {
            return true;
        }
        if (!(obj instanceof StringEncodedValue)) {
            return false;
        }
        StringEncodedValue other = (StringEncodedValue) obj;
        return Objects.equals(values, other.values);
    }

    @Override
    public int getVersion() {
        return 31 * super.getVersion() + staticHashCode(values);
    }
}
```

karussell:
Yes, thanks :) 

An additional indexOf method would be nice, so we can easier get the int (for fast usage inside Weighting.calc)

otbutz:
Changed my implementation and added a few unittests

karussell (approved):
Thanks! Also for the additional tests regarding the edge cases.

PR #2206, opened 2020-12-09
Extend the list of restricted encoded value names

otbutz:
Added the following entries: 

- `_` (underscore is a keyword since Java 9)
- false, true
- null
- sealed, non-sealed, permits (JEP 360)
- record (JEP 395)
- var, yield (restricted identifiers)

https://docs.oracle.com/javase/specs/jls/se15/html/jls-3.html#jls-3.9


karussell:
Thanks!

PR #2210, opened 2020-12-11
Properly decode lat/lon for DenseNodes

### LINKED ISSUE #2197: Use PbfFieldDecoder for DenseNodes
Citing the OSM wiki:

> In addition to granularity, the primitive block also encodes a latitude and longitude offset value. These values, measured in units of nanodegrees, **must** be added to each coordinate.
> 
> ```
> latitude = .000000001 * (lat_offset + (granularity * lat))
> longitude = .000000001 * (lon_offset + (granularity * lon))
> ```
> 
> Where latitude is the latitude in degrees, granularity is the granularity given in the PrimitiveBlock, lat_offset is the offset given in the PrimitiveBlock, and **lat/lon is encoded in a Node or delta-encoded in a DenseNode**. The explanation of the equation for longitude is analogous. 

The PbfBlobDecoder ignores the offset and granularity completely for DenseNodes:
https://github.com/graphhopper/graphhopper/blob/c03a54c212da1719c901d82526a644bc6a39b1c2/reader-osm/src/main/java/com/graphhopper/reader/osm/pbf/PbfBlobDecoder.java#L240

otbutz:
Fixes #2197

### DIFF
--- a/reader-osm/src/main/java/com/graphhopper/reader/osm/pbf/PbfBlobDecoder.java
+++ b/reader-osm/src/main/java/com/graphhopper/reader/osm/pbf/PbfBlobDecoder.java
@@ -237,7 +237,7 @@ public class PbfBlobDecoder implements Runnable {
                 tags.put(fieldDecoder.decodeString(keyIndex), fieldDecoder.decodeString(valueIndex));
             }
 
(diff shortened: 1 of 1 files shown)


karussell:
Thanks!

PR #2209, opened 2020-12-10
Java/Janino scripting to replace our interpreter

karussell:
This change is intended to make custom profiles more powerful and also more readable. It breaks backward compatibility. It uses a very powerful and fast library called [Janino](https://janino-compiler.github.io/janino/) that allows on-the-fly compilation of partial expressions, blocks or whole classes written in Java.

With Janino scripting we can satisfy much more use cases. As soon as custom profiles were released there was often the need to combine expressions like "use factor=0.5 only if road_class is primary AND in a certain area" or "surface is unpaved AND road_class is not unclassified". This is now possible. Still I do not think we should use more power of Janino and keep the scripting simple (expressions only).

The big advantage with Janino scripting is now that we do not have the burden in the future to extend or improve our own interpreter. Performance-wise the routing with Janino scripting is at least of the same speed (for preparation it is even faster!). For query time this sounds a bit strange as the compilation takes some time. But when we assume that a custom profile is re-used for future requests of the same profile the class cache helps (see ExpressionBuilder, but the cache can be disabled). Now the most surprising thing for me was that even without the cache the slow down is not much (20% for LM queries? precise numbers needs to be re-done for the most recent commit). Janino is able to compile many expressions and a whole class in only a few milliseconds. And currently we create a cached version automatically for all server-side profiles due to GraphHopper.checkProfilesConsistency, i.e. for them there is no compilation overhead. btw: our cache is not that bad or complex as the cached object for key X is always the same (unlike for DBs with user data).

We have to add Janino as dependency to the core (1MB). I find this acceptable for what we get.

A custom profile in the current master looks like:

```yml
priority:
  max_height: {"<3.8": 0}

speed_factor:
  road_class:
    motorway: 0.85
    primary: 0.9

max_speed:
  road_class:
    living_street: 2

max_speed_fallback: 80
```

with this branch this gets more readable and you do not need to study the docs and still know what it does:

```yml
priority:
- if: max_height < 3.80
  multiply by: 0

speed:
- if: road_class == MOTORWAY: 
  multiply by: 0.85
- if: road_class == PRIMARY
  multiply by: 0.9
  # if you wish to keep compatibility you apply max_speed after all multiplications, but in general you can use "limit to" at any time
- if: road_class == LIVING_STREET
  limit to: 2
  # replacement for max_speed_fallback which is now implicit
  # it could be also done via the else statement of the last if ... or via:
- if: true
  limit to: 80
```

It is slightly more verbose for the enum case, but also more clear. 

[...]

PR #2211, opened 2020-12-13
Custom weighting helper cleanup

easbar:
Here I tried separating the custom model parsing from the custom weighting creation a bit, such that e.g. we can test extracting speeds and priorities from the custom model definition without creating a `CustomWeighting`.

Follow-up of this comment: https://github.com/graphhopper/graphhopper/pull/2209/files#r541899846


core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeighting.java
```diff
@@ -169,4 +165,46 @@ public long calcEdgeMillis(EdgeIteratorState edgeState, boolean reverse) {
     public String getName() {
         return NAME;
     }
+
+    @FunctionalInterface
+    public interface EdgeToDoubleMapping {
+        double apply(EdgeIteratorState edge, boolean reverse);
```
karussell, line 171: In current master this is named `EdgeToValueEntry` and `getValue` but the class name you have is better, unsure about apply :D

karussell (approved):
Thanks I now better understand what you mean. Especially testing priority calculation without CustomWeighting is useful to avoid relying entirely on some magic weight numbers :)

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeighting.java
easbar: `java.util.Function` calls it `apply`, but honestly this never sounded very intuitive to me either. Maybe call it `get` like `java.util.Supplier`? Or `map`? Don't know really.

karussell: `get` sounds better, yes :)

easbar:
After merging this maybe we can migrate some tests from `CustomWeightingTest` to `ExpressionBuilderTest`, but no need to rush and we can also do this later in master.

PR #2214, opened 2020-12-16
Log MultiException properly

easbar:
Our `MultiException` constructor calls the default `RuntimeException` constructor , so the `detailsMessage` field in `Throwable` is always `null`. Therefore we log `unknown reason` for many errors. I changed this so we see more details on e.g. 400 errors like this:

```
# before
com.graphhopper.http.MultiExceptionMapper: bad request: unknown reason

# after
com.graphhopper.http.MultiExceptionMapper: bad request: [java.lang.IllegalArgumentException: The 'heading' parameter is currently not supported for speed mode, you need to disable speed mode with `ch.disable=true`. See issue #483]
```

Not sure, we could also properly set the message in `MultiException` and skip `java.lang.IllegalArgumentException` for the output, but this seemed like the minimal fix..

### DIFF
--- a/web-bundle/src/main/java/com/graphhopper/http/MultiExceptionMapper.java
+++ b/web-bundle/src/main/java/com/graphhopper/http/MultiExceptionMapper.java
@@ -33,7 +33,9 @@ public class MultiExceptionMapper implements ExceptionMapper<MultiException> {
 
     @Override
     public Response toResponse(MultiException e) {
-        logger.info("bad request: " + (Helper.isEmpty(e.getMessage()) ? "unknown reason" : e.getErrors()));
+        logger.info("bad request: " + (Helper.isEmpty(e.getMessage())
+                ? (e.getErrors().isEmpty() ? "unknown reason" : e.getErrors().toString())
+                : e.getErrors()));
         return Response.status(Response.Status.BAD_REQUEST)
                 .entity(e)
                 .build();


michaz:
Thanks! Wait, could we instead change MultiException, please? I was lazy there.

It's a common pattern, we have at least two MultiExceptions in our classpath, one from Jetty and one from HK2 via Jersey via Dropwizard.

Could you try to copy/paste more construction from there to our MultiException until it behaves nicely? (Not a lot I think..)

michaz:
(No reason _not_ to do also do something in the Logger, of course..)

easbar:
The one from jetty sets the message field to `"Multiple exceptions"`, and overwrites `toString`, which just prints a list of the nested exceptions (`List.toString()`). The one from HK2 uses the first of the multiple nested exceptions as message and cause (but why?) but also overrides `getMessage()` where the information of all nested exceptions is included. We also use the message of the first nested exception as message field in `MultiExceptionSerializer` (but all the others are still included in `hints`).

So I guess the question is whether or not we want to mess with the `message` field and/or the `getMessage()` method. I think I decided against this because so far the only problem is this server side log (and apparently even this did not seem to bother anyone for many years) and I do not want/risk to break the public API :)

easbar:
So if we change `MultiException` I would probably go with the jetty approach: Use `"Multiple exceptions"` as message and override `toString()`.

easbar:
Hm, but the usages look fine and the output should be controlled by `MultiExceptionSerializer` and `MultiExceptionGPXMessageBodyWriter`?

michaz (approved):
Ah okay. Thanks for the analysis!

Issue #1916, opened 2020-02-17
Rare failure of RandomizedRoutingTest for edge-based CH


easbar:
For this commit an CH (edge-based) test failed during a CI build: 79f3ecd606695f2510b67de06bcd28ed16affce9

```
[INFO] Running com.graphhopper.routing.weighting.RandomizedRoutingTest
expected: [50, 8, 35, 22]
given:    [50, 1, 22]
seed: 130339338847
[ERROR] Tests run: 60, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5.595 s <<< FAILURE! - in com.graphhopper.routing.weighting.RandomizedRoutingTest
[ERROR] randomGraph_withQueryGraph[CH_ASTAR, EDGE_BASED](com.graphhopper.routing.weighting.RandomizedRoutingTest)  Time elapsed: 0.069 s  <<< FAILURE!
java.lang.AssertionError: 
wrong weight: 50->22
expected: 69.53626249436314
given:    71.0755881986408
seed: 130339338847
	at com.graphhopper.routing.weighting.RandomizedRoutingTest.comparePaths(RandomizedRoutingTest.java:353)
	at com.graphhopper.routing.weighting.RandomizedRoutingTest.runWithQueryGraph(RandomizedRoutingTest.java:310)
	at com.graphhopper.routing.weighting.RandomizedRoutingTest.randomGraph_withQueryGraph(RandomizedRoutingTest.java:282)
```

easbar:
In this commit: dfa19a686c6efb44d5373a00059eff151a563704 some issues of this test were changed and now turn costs are also included. Also the encoder was changed from motorcycle to car. However the test still fails as of now (fb3eed814b567cfebde3bed20c3c273e7344c913) if we change the encoder back to Motorcycle (and use the above seed).

easbar:
Here is another... d353297209ec871eca0468ce04f269739452beff

easbar:
> However the test still fails as of now (fb3eed8) if we change the encoder back to Motorcycle (and use the above seed).

This is just a duplicate of #2061 

> Here is another... d353297

This is a rare case of a certain soft violation:

```
[
wrong distance 33->50, expected: 6745.514924478206, given: 7031.102924478206, wrong nodes 33->50
expected: [33, 44, 18, 39, 35, 16, 5, 23, 28, 43, 15, 0, 21, 2, 32, 50]
[...]

PR #2215, opened 2020-12-17
Do not cancel search for edge-based CH AStar too early

### LINKED ISSUE #2061: Wrong weight for edge-based CH when using AStar
cacd3d7cdeef8eb1480ca592a2d888f8637d53ec

This caused [the build](https://travis-ci.org/github/graphhopper/graphhopper/builds/695750595?utm_source=github_status&utm_medium=notification) to fail in #2060:

easbar:
Fixes #2061 and #1916.

The first problem was a real bug in the (currently disabled) CH/AStar implementation: We cancelled the entire search under a certain condition when we should only stall the current node. The other problem were some soft validations in the tests were the shortest path was not well defined because of alternative paths with same distance (see #1916).

I now enabled AStar again for edge-based CH which improves the query speed from about 3.3ms to 2.4ms (about 25-30%).


karussell:
Nice! It's going to be a great 3.0 release :) - so many goodies!

PR #2216, opened 2020-12-18
LocationIndexTree cleanup

michaz:
Project:

- [x] - Store edge ids, not node ids, because that's what's conceptually already happening
- [x] - Remove findNClosest; use the much better defined bounding box query instead. It doesn't matter at all that the query shape is then a square, not a circle. A ton of code goes away.
- [ ] - Clarify what the parameters mean, particularly "precision" and "maxRegionSearch". I mean, if you have high precision aka super-small tiles, then you possibly have to search many tiles to find a nearest neighbor. You can't have it both ways. But that's okay because if there are no bugs, a low "precision" doesn't actually change the results, it just changes the performance. Etc. Actually, "precision" and "maxRegionSearch" *together* are something like a precision, because that determines how far away from a street I can click and still get a result.
- [x] - Possibly:
- [x]   - Use the line scan algorithm that we want to be using (put the line in all the tiles that it intersects)
- [x]   - Why don't we do the calculations for that in pixels, like the original line scan algorithms? We are already using fixed-size cells, aka pixels, with fixed-length spatial keys, aka pixel coordinates. Our line scan can just take point a and point b directly as spatial keys, and emit all pixels on the line as spatial keys as well.


michaz:
Done -- LocationIndex now stores edgeIds, and `findNClosest` is gone. MapMatching now does a bounding box query (expecting that it's correct, so modulo the open Bresenham problem), and does the "snapping" in its own code. It's even a bit faster, apparently.

easbar:
This sounds great :+1: . We probably need to adjust/fix measurement as well. Currently it fails with this:

```sh
2020-12-20 02:59:49,335 [main] ERROR com.graphhopper.tools.Measurement - Problem while measuring /ext_data/tmp/measurement-small-gh
java.lang.ArrayIndexOutOfBoundsException: Index 35 out of bounds for length 32
	at com.graphhopper.storage.RAMIntDataAccess.getInt(RAMIntDataAccess.java:208)
	at com.graphhopper.storage.GHNodeAccess.getLatitude(GHNodeAccess.java:66)
	at com.graphhopper.storage.GHNodeAccess.getLat(GHNodeAccess.java:89)
	at com.graphhopper.storage.index.LocationIndexTree.calcMinDistance(LocationIndexTree.java:565)
	at com.graphhopper.storage.index.LocationIndexTree.findNetworkEntries(LocationIndexTree.java:545)
	at com.graphhopper.storage.index.LocationIndexTree.findClosest(LocationIndexTree.java:589)
	at com.graphhopper.tools.Measurement.lambda$printLocationIndexQuery$1(Measurement.java:461)
	at com.graphhopper.util.MiniPerfTest.start(MiniPerfTest.java:46)
	at com.graphhopper.tools.Measurement.printLocationIndexQuery(Measurement.java:458)
	at com.graphhopper.tools.Measurement.start(Measurement.java:207)
	at com.graphhopper.tools.Measurement.main(Measurement.java:93)
```

See `Measurement#printLocationIndexQuery`.

michaz:
Oops! No, that was a proper mistake by me, which is only found by the measurement and not by the tests. :-} Will fix.

michaz:
One round trip routing test now finds a slightly different round trip, but that's okay, I guess, if it's also a nice round trip.

PR #2218, opened 2020-12-21
Elevation output

braito4:
https://github.com/graphhopper/map-matching/pull/187


karussell:
@michaz should we use a ConfiguredCommand instead to read the configuration from the config file?

braito4:
My workflow for map-matching was basically over cli, but originally you can only add elevation if you use the web interface, so; this add add elevation on cli. 
I didn't add cli for importing; i haven't figure it out yet

michaz:
@karussell Yeah, if anything. I don't really feel like maintaining this, it's just been there forever and at one point I ported it to a Dropwizard Command.

michaz:
MatchCommand now requires a config file and a profile name. If elevation can go in the config file, it can now go into map-matching.

Issue #1984, opened 2020-04-13
Problem with via-point (virtual node) when cutting an area via osmosis


karussell:
Problematic test case that gives different answer for route Q to P when requesting "Q, P, R" vs. "Q, P" only:

```java
@Test
public void edge_splitting_bug() {
    GraphHopperOSM hopper = new GraphHopperOSM();
    hopper.setEncodingManager(EncodingManager.create("car"));
    hopper.setGraphHopperLocation(GH_LOCATION);
    hopper.setOSMFile("../core/files/lauf-200115.osm.pbf");
    hopper.setProfiles(new ProfileConfig("profile").setVehicle("car").setWeighting("fastest"));
    hopper.importOrLoad();
    GraphHopperStorage graph = hopper.getGraphHopperStorage();
    LocationIndex locationIndex = hopper.getLocationIndex();
    FlagEncoder encoder = hopper.getEncodingManager().fetchEdgeEncoders().get(0);
    final FastestWeighting weighting = new FastestWeighting(encoder);
    final BooleanEncodedValue accessEnc = weighting.getFlagEncoder().getAccessEnc();
    EdgeFilter edgeFilter = new EdgeFilter() {
        @Override
        public boolean accept(EdgeIteratorState edgeState) {
            return edgeState.get(accessEnc) && !Double.isInfinite(weighting.calcEdgeWeight(edgeState, false))
                    || edgeState.getReverse(accessEnc) && !Double.isInfinite(weighting.calcEdgeWeight(edgeState, true));
        }
    };
    GHPoint p = new GHPoint(49.529355099741494, 11.296203940756207);
    GHPoint q = new GHPoint(49.523147, 11.293001);
    GHPoint r = new GHPoint(49.53027363148555, 11.293370881574472);

    // works
    List<GHPoint> points = Arrays.asList(p, q/*, r*/);
    // does not work
//        List<GHPoint> points = Arrays.asList(p, q, r);

    List<QueryResult> queryResults = new ArrayList<>(points.size());
    for (GHPoint point : points) {
        queryResults.add(locationIndex.findClosest(point.lat, point.lon, edgeFilter));
    }
    QueryGraph queryGraph = QueryGraph.lookup(graph, queryResults);
    Dijkstra dijkstra = new Dijkstra(queryGraph, queryGraph.wrapWeighting(weighting), TraversalMode.NODE_BASED);
[...]

Issue #19, opened 2013-03-04
Problems importing OSM data


sweco-semtne:
I'm having trouble digesting a cloudmade OSM download of Berlin.

http://downloads.cloudmade.com/europe/western_europe/germany/berlin#downloads_breadcrumbs

Exception in thread "main" java.lang.AssertionError: Conversation pillarNode to towerNode already happended!? osmId:99128942 pillarIndex:13589

PR #2221, opened 2020-12-21
OSMReader has problems with ways that reference missing nodes

easbar:
Sometimes OSM ways reference nodes that do not exist in the OSM file. This problem was partly fixed in #19 (yes long ago), but only for the case where the missing nodes where somewhere in the middle of a way. However, when the missing nodes are at the beginning of the way I ran into two problems:

1) In some cases the entire way is left out and not added to the graph at all
2) Even worse: In some cases an edge is added with 
`edge.getDistance() != calcDistance(edge.fetchWayGeometry(FetchMode.ALL))`.

Such missing nodes appear frequently when we extract a box using Osmosis. In this case often the first or last nodes are missing, because the way starts/ends outside of the bounding box. See #1984. I am not sure if this can also happen with original OSM files.

If this data issue only arises when using certain tools like Osmosis I think we should simply reject such data. Its easy to throw an appropriate error in this case and It should be easy enough use some tool to fix the issue before importing it with GH. In this case maybe we can also get rid of the code that was added for #19. Otherwise we would need to add some code that handles this situation correctly. Even leaving out such ways entirely would be an option IMO, but there should definitely be no edges with inconsistent distances, because this leads to problems like #1984.

### DIFF
--- a/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -690,7 +690,10 @@ public class OSMReader implements DataReader, TurnCostParser.ExternalInternalMap
 
         // sample points along long edges
         if (this.longEdgeSamplingDistance < Double.MAX_VALUE && pointList.is3D())
-            pointList = EdgeSampling.sample(wayOsmId, pointList, longEdgeSamplingDistance, distCalc, eleProvider);
+            pointList = EdgeSampling.sample(pointList, longEdgeSamplingDistance, distCalc, eleProvider);
+
+        if (doSimplify && pointList.size() > 2)
+            simplifyAlgo.simplify(pointList);
 
         double towerNodeDistance = distCalc.calcDistance(pointList);
 
@@ -715,9 +718,6 @@ public class OSMReader implements DataReader, TurnCostParser.ExternalInternalMap
         }
 
         EdgeIteratorState iter = graph.edge(fromIndex, toIndex).setDistance(towerNodeDistance).setFlags(flags);
-
-        if (doSimplify && pointList.size() > 2)
-            simplifyAlgo.simplify(pointList);
 
         // If the entire way is just the first and last point, do not waste space storing an empty way geometry
         if (pointList.size() > 2)
--- a/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -720,11 +720,32 @@ public class OSMReader implements DataReader, TurnCostParser.ExternalInternalMap
         EdgeIteratorState iter = graph.edge(fromIndex, toIndex).setDistance(towerNodeDistance).setFlags(flags);
 
         // If the entire way is just the first and last point, do not waste space storing an empty way geometry
-        if (pointList.size() > 2)
+        if (pointList.size() > 2) {
+            // the geometry consists only of pillar nodes, but we check that the first and last points of the pointList
+            // are equal to the tower node coordinates
+            checkCoordinates(fromIndex, pointList.get(0));
+            checkCoordinates(toIndex, pointList.get(pointList.size() - 1));
             iter.setWayGeometry(pointList.shallowCopy(1, pointList.size() - 1, false));
+        }
 
+        checkDistance(iter);
         storeOsmWayID(iter.getEdge(), wayOsmId);
         return iter;
+    }
+
+    private void checkCoordinates(int nodeIndex, GHPoint point) {
+        final double tolerance = 1.e-6;
+        if (Math.abs(nodeAccess.getLat(nodeIndex) - point.getLat()) > tolerance || Math.abs(nodeAccess.getLon(nodeIndex) - point.getLon()) > tolerance)
+            throw new IllegalStateException("Suspicious coordinates for node " + nodeIndex + ": (" + nodeAccess.getLat(nodeIndex) + "," + nodeAccess.getLon(nodeIndex) + ") vs. (" + point + ")");
+    }
+
+    private void checkDistance(EdgeIteratorState edge) {
+        final double tolerance = 1;
+        final double edgeDistance = edge.getDistance();
+        final double geometryDistance = distCalc.calcDistance(edge.fetchWayGeometry(FetchMode.ALL));
+        if (Math.abs(edgeDistance - geometryDistance) > tolerance)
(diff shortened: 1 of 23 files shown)


easbar:
This PR now fixes the issue: When there are ways that start/end with references to non-existing nodes the missing nodes are ignored. I also added two checks that run during the import and make sure that the edge distance always matches the edge geometry and the coordinates of the first/last points in the temporary pointList match those of the tower nodes. The first check is now possible because of #2222.

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -573,10 +573,30 @@ int addTowerNode(long osmId, double lat, double lon, double ele) {
...
-        final int lastIndex = osmNodeIds.size() - 1;
         int lastInBoundsPillarNode = -1;
         try {
-            for (int i = 0; i < osmNodeIds.size(); i++) {
+            // #2221: ways might include nodes at the beginning or end that do not exist -> skip them
+            int firstExisting = -1;
+            int lastExisting = -1;
+            for (int i = 0; i < osmNodeIds.size(); ++i) {
+                final long tmpNode = getNodeMap().get(osmNodeIds.get(i));
+                if (tmpNode > -TOWER_NODE || tmpNode < TOWER_NODE) {
```
easbar, line 583: Btw one of the major difficulties I had when trying to understand `OSMReader` is that the node map is used in two different ways during the import. In the first pass it simply separates between nodes that occur in two or more, only one or zero ways. For this we use the `TOWER/PILLAR/EMPTY_NODE` constants. But then in the second run we use the same map and assign pillar/tower node ids and this leads to very confusing code like this:

```java
  // skip osmIds with no associated pillar or tower id (e.g. !OSMReader.isBounds)
                if (tmpNode == TOWER_NODE)
                    continue;

                if (tmpNode == PILLAR_NODE) {
```

where `tmpNode == TOWER/PILLAR_NODE` does **not** mean that `tmpNode` is a tower/pillar node but it means something very different(!): Because no 'real' tower/pillar node id was assigned for this node it means that `tmpNode` does not exist. 

This is probably the first thing I would clean up in OSMReader, because it really tripped me up. I understand that we probably want to re-use the map for performance reasons, but at least we should use a second reference to it (using a different name) or avoid this whole confusion by using some helper methods. Then again I wonder if we could exploit the fact that during the first pass we only need to distinguish three values (zero ways, one way, multiple ways) to achieve faster lookups somehow.

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -573,10 +573,30 @@ int addTowerNode(long osmId, double lat, double lon, double ele) {
         final PointList pointList = new PointList(osmNodeIds.size(), nodeAccess.is3D());
         final List<EdgeIteratorState> newEdges = new ArrayList<>(5);
         int firstNode = -1;
-        final int lastIndex = osmNodeIds.size() - 1;
         int lastInBoundsPillarNode = -1;
         try {
-            for (int i = 0; i < osmNodeIds.size(); i++) {
+            // #2221: ways might include nodes at the beginning or end that do not exist -> skip them
+            int firstExisting = -1;
+            int lastExisting = -1;
```
easbar, line 580: The actual fix is a bit lengthy, but rather trivial: We scan the osmNodeIds list from the beginning and end to find the first node that does exist (and skip those that do not exist).

easbar:
> If this data issue only arises when using certain tools like Osmosis I think we should simply reject such data. Its easy to throw an appropriate error in this case and It should be easy enough use some tool to fix the issue before importing it with GH. 

This still seems like a good solution to the problem to me, but I am not sure if these data errors occur in e.g. Geofabrik extracts. Also some of our testing maps contain these kind of errors so we would have to replace them as well. For now I think we should  apply the fix.

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: > if we could exploit the fact that during the first pass we only need to distinguish three values 

For example we could use a `LongHashSet` to store all OSM node Ids that occur in any way and a second (smaller) one to keep track which of these nodes are tower nodes. Not sure if this would be an improvement, because before pass2 we would still have to build the mapping between OSM node Ids and GH tower/pillar IDs.

karussell: > but at least we should use a second reference to it (using a different name) or use some helper methods to make this more clear.

:+1: 

> For example we could use a LongHashSet to store all OSM node Ids that occur in any way

The problem might for big maps with millions of nodes as sets behave very ugly then: when increasing the size you double the size (additionally to the load factor of usually 0.75) and so you will need 3*1/0.75 = 4 and you copy such big arrays whenever it is too small. But surely we can just try as the memory usage requirements changed significantly over the last years. Still at some point not too far in the past I tried with a database like mapdb but it was still much slower and used more resources (disk+memory).

karussell (approved):
> Sometimes OSM ways reference nodes that do not exist in the OSM file.

The assumption before was that this will never happen (in geofabrik PBFs this did not occur to my knowledge). But is nice that we know and partially fix this now when it happens.

> Also some of our testing maps contain these kind of errors 

Interesting. Whenever I created a test PBF I used the export feature of openstreetmap.org and if it is a pbf converted the xml to a pbf with probably osmosis. So maybe the most tools are not 100% strict.

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: Ok yes this was just an idea, its probably hard to say without trying it. Also I am not even sure if there is really much need to speed up the OSM parsing and improving readability seems more important.

easbar:
> The assumption before was that this will never happen (in geofabrik PBFs this did not occur to my knowledge). But is nice that we know and partially fix this now when it happens.

We actually already handled a similar case before, see #19. But this was only for missing nodes in the middle of a way. The worst part was that we did not throw an error in case some nodes were missing unexpectedly.

reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java
karussell: Yes, a speed up is not necessary. Still I would like it, if the new version is not much slower :) (maybe max 2x slower? probably depends on the gained readability)

PR #2222, opened 2020-12-21
Simplify edge geometry before the distance is calculated

easbar:
During import we calculate `edge.getDistance()` before we simplify the edge geometry. Therefore the edge distance can be different to the distance we would calculate from the edge geometry:

```java
edge.getDistance() != new DistanceCalcEarth().calcDistance(edge.fetchWayGeometry(FetchMode.ALL));
```

This doesn't seem right, so in this PR I moved the edge simplification in front of the distance calculation.

### DIFF
--- a/reader-osm/src/test/java/com/graphhopper/GraphHopperTest.java
+++ b/reader-osm/src/test/java/com/graphhopper/GraphHopperTest.java
@@ -114,8 +114,8 @@ public class GraphHopperTest {
         assertEquals(expectedVisitedNodes, rsp.getHints().getLong("visited_nodes.sum", 0));
 
         ResponsePath res = rsp.getBest();
-        assertEquals(3587.9, res.getDistance(), .1);
-        assertEquals(277173, res.getTime(), 10);
+        assertEquals(3586.9, res.getDistance(), .1);
+        assertEquals(277112, res.getTime(), 10);
         assertEquals(91, res.getPoints().getSize());
 
         assertEquals(43.7276852, res.getWaypoints().getLat(0), 1e-7);
@@ -137,10 +137,10 @@ public class GraphHopperTest {
                 setAlgorithm(ASTAR).setProfile(profile));
 
         // identify the number of counts to compare with CH foot route
-        assertEquals(699, rsp.getHints().getLong("visited_nodes.sum", 0));
+        assertEquals(700, rsp.getHints().getLong("visited_nodes.sum", 0));
 
         ResponsePath res = rsp.getBest();
-        assertEquals(3437.6, res.getDistance(), .1);
+        assertEquals(3437.1, res.getDistance(), .1);
         assertEquals(85, res.getPoints().getSize());
 
         assertEquals(43.7276852, res.getWaypoints().getLat(0), 1e-7);
@@ -280,7 +280,7 @@ public class GraphHopperTest {
             long sum = rsp.getHints().getLong("visited_nodes.sum", 0);
             assertNotEquals(sum, 0);
             assertTrue("Too many nodes visited " + sum, sum < 120);
-            assertEquals(3437.6, bestPath.getDistance(), .1);
+            assertEquals(3437.1, bestPath.getDistance(), .1);
             assertEquals(85, bestPath.getPoints().getSize());
         }
 
@@ -296,7 +296,7 @@ public class GraphHopperTest {
             long sum = rsp.getHints().getLong("visited_nodes.sum", 0);
             assertNotEquals(sum, 0);
             assertTrue("Too many nodes visited " + sum, sum < 120);
-            assertEquals(3437.6, bestPath.getDistance(), .1);
+            assertEquals(3437.1, bestPath.getDistance(), .1);
             assertEquals(85, bestPath.getPoints().getSize());
         }
 
@@ -311,7 +311,7 @@ public class GraphHopperTest {
         long sum = rsp.getHints().getLong("visited_nodes.sum", 0);
         assertNotEquals(sum, 0);
         assertTrue("Too few nodes visited " + sum, sum > 120);
-        assertEquals(3437.6, bestPath.getDistance(), .1);
+        assertEquals(3437.1, bestPath.getDistance(), .1);
         assertEquals(85, bestPath.getPoints().getSize());
 
         hopper.close();
@@ -361,7 +361,7 @@ public class GraphHopperTest {
         assertEquals(2, rsp.getAll().size());
 
         assertEquals(1310, rsp.getAll().get(0).getTime() / 1000);
-        assertEquals(1432, rsp.getAll().get(1).getTime() / 1000);
+        assertEquals(1431, rsp.getAll().get(1).getTime() / 1000);
 
         req.putHint("alternative_route.max_paths", 3);
         req.putHint("alternative_route.min_plateau_factor", 0.1);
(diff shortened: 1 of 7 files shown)


reader-osm/src/test/java/com/graphhopper/GraphHopperTest.java
```diff
@@ -887,7 +887,7 @@ public void testSRTMWithInstructions() {
                 setAlgorithm(ASTAR).setProfile(profile));
 
         ResponsePath res = rsp.getBest();
-        assertEquals(1625.4, res.getDistance(), .1);
+        assertEquals(1614.3, res.getDistance(), .1);
```
easbar, line 890: This is the only test where my change lead to a distance different above `1m`, is this ok?

easbar:
To me the biggest question is whether the edge simplification should run before or after the long edge sampling. Or even before the elevation smoothing? @msbarry can you comment?

otbutz:
Simplification boils down to loss of precision. There _might_ be corner cases where we pick up wrong elevation data because we work with simplified paths.

msbarry:
> To me the biggest question is whether the edge simplification should run before or after the long edge sampling. Or even before the elevation smoothing? @msbarry can you comment? Probably it should go before the long edge sampling otherwise we might not sample a long edge, because it only turns out to be a long edge during the simplification?

Edge simplification should run after long edge sampling, since long edge sampling inserts elevation detail on long straight edges that influences simplification (since simplification is now elevation-aware), see the example before/after elevation profile here https://github.com/graphhopper/graphhopper/pull/1953#issue-387694399. 

It seems safe to calculate the edge distance before simplification since we use such a tight threshold is used to simplify before storing edge geometries, especially if the test diffs are so small.

easbar:
Thanks for your response!

> Edge simplification should run after long edge sampling, 

Ok you mean first we should add all possible sample points and then simplify afterwards? That sounds reasonable, but actually I do not even expect much of a difference quality-wise(?). When we simplify first this just means some points might get removed, but then we actually add points again during the long edge sampling. Initially I was concerned that if we ran the sampling first we might miss adding some elevation details, because the long edge sampling would not run for all edges. But contrary to what I thought the long edge sampling always runs (and not just on long edges)? So yes let's run the sampling before the simplification (this is the current state of this PR anyway).

> It seems safe to calculate the edge distance before simplification since we use such a tight threshold is used to simplify before storing edge geometries, especially if the test diffs are so small.

Do you mean we should or should not calculate the edge distance before the simplification?

And by the way two more questions:

1) We long-edge-sample with a fixed interval. Would it not make sense to e.g. sample with a higher frequency but only keep the points with significant elevation changes (so we hopefully cover all elevation details)? Hm but as long as the simplification runs afterwards we can probably add as many points we want, because the simplification will remove those that are unnecessary. Maybe this was the idea?

2) Do you remember what the `wayOsmId` parameter here: https://github.com/graphhopper/graphhopper/blob/bde99a8ccd143f3852b3bc87b3bcb8b5328f8f0f/core/src/main/java/com/graphhopper/reader/dem/EdgeSampling.java#L37 was for? It seems unused, can we remove it?

msbarry:
> Ok you mean first we should add all possible sample points and then simplify afterwards?

Yes, the long edge sampling is just meant to keep a minimum distance between elevation samples along all edges (it just has a shortcut to avoid touching short edges), then 3d-aware simplification throws out any edge samples that really aren't needed to retain elevation details in the profile.  If we don't simplify after the long edge sampling, we'll keep too much extra data around.

> Do you mean we should or should not calculate the edge distance before the simplification?

It seems safe to calculate the edge distance before simplification like you have already in this PR.

> 1. We long-edge-sample with a fixed interval. Would it not make sense to e.g. sample with a higher frequency but only keep the points with significant elevation changes (so we hopefully cover all elevation details)? Hm but as long as the simplification runs afterwards we can probably add as many points we want, because the simplification will remove those that are unnecessary. Maybe this was the idea?

Yes exactly, they need to run together. Long edge sampling adds in detail, then simplification throws out unnecessary detail.

> 2. Do you remember what the `wayOsmId` parameter here: https://github.com/graphhopper/graphhopper/blob/bde99a8ccd143f3852b3bc87b3bcb8b5328f8f0f/core/src/main/java/com/graphhopper/reader/dem/EdgeSampling.java#L37
>     was for? It seems unused, can we remove it?

Oops, I must have had it in there for debug logging. Looks like it's not used anymore, so safe to remove. Sorry about that!

otbutz:
> When we simplify first this just means some points might get removed, but then we actually add points again during the long edge sampling.

Removing points changes the path of the edge. In extreme cases the long edge sampling _might_ pick up completely wrong elevation data because the inserted points are not part of the original edge. But i'm not sure if we really run into this kind of problems as we don't simplify that hard and our elevation data usually has low resolution.

grafik

easbar:
Ok

> It seems safe to calculate the edge distance before simplification like you have already in this PR.

This whole PR is about calculating the edge distance **after** the simplification, but I guess this is just a typo :) 

> Removing points changes the path of the edge.

Yes sure. When we first sample the elevation details the edges will be protected from over-simplification. And yes the simplification threshold is very small anyway (1m) and it cannot really be larger so we keep details like sharp turns of footways or roundabouts etc.

msbarry:
> This whole PR is about calculating the edge distance **after** the simplification, but I guess this is just a typo :) 

Sorry yes. That's what I meant!

> In extreme cases the long edge sampling _might_ pick up completely wrong elevation data because the inserted points are not part of the original edge. But i'm not sure if we really run into this kind of problems as we don't simplify that hard and our elevation data usually has low resolution.


"Long edge sampling" might be a bad name. We only add intermediate points along long edge segments, but never actually change the shape of a line. "Interpolate" might make more sense here? I think QGis calls it "points along geometry?"

reader-osm/src/test/java/com/graphhopper/GraphHopperTest.java
easbar: I had a look at this and the difference of about `11m` is the result of multiple edges that have a different distance due to the change in this PR. The biggest contribution is an edge with a `6.9m` difference and the full geometry is shown here:

image

Due to the simplification it looks like this:

image

Because point 2 is removed we get a longer less steep segment instead of a short steep and a long flat one and this leads to the different distance with and without simplification. Note that the long edge sampling that would normally prevent the loss of these elevation details is not enabled in this test.

So yes this PR changes the distance by a significant amount, but no this is nothing to worry about, because all we do now is that we store the distance of the geometry we store for this edge instead of storing a distance for a geometry that we do not keep anyway.

easbar:
@karussell any objections? I think this is ready to merge.

karussell (approved):
Thanks!

easbar:
> "Long edge sampling" might be a bad name. We only add intermediate points along long edge segments, but never actually change the shape of a line. "Interpolate" might make more sense here? I think QGis calls it "points along geometry?"

Hm 'sampling' works for me I guess. What confuses me more is the 'long edge' part of it: It sounds like we only do the sampling for 'long' edges. But really this parameter sets a minimum sampling interval. For example setting it to 30m means that whenever there is a segment longer than 30m we add additional sampling points every 30m along this segment. This way we make sure the elevation is sampled at least every 30m (unless the 3D simplification removes the sample point again because it is not needed). The total length of the edge does not even play a role.

PR #2225, opened 2020-12-29
random test for LocationIndex

karussell:
I always wanted to have random tests for LocationIndex like we have for routing algorithms. Now due to the nice work in #2216 I invested a bit time :) 

This seems to fail for master too, but for branch `locindex` more frequent. We need to investigate if it is the "full search" test method which is not yet perfect or one or more bugs in LocationIndexTree.

### DIFF
--- a/core/src/test/java/com/graphhopper/storage/index/LocationIndexTreeTest.java
+++ b/core/src/test/java/com/graphhopper/storage/index/LocationIndexTreeTest.java
@@ -23,6 +23,8 @@ import com.graphhopper.storage.*;
 import com.graphhopper.util.*;
 import com.graphhopper.util.shapes.BBox;
 import com.graphhopper.util.shapes.GHPoint;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
 
 import java.io.Closeable;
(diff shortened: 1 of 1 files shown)


michaz:
Hmm. "wayIndex" was off-by-one in some cases, and also, locationIndex doesn't continue looking for a closer position on the same edge when the distance is smaller than a threshold.

Hence the difference / failing test.

I replaced the logic by a call to the logic used in LocationIndex.

The test can now only fail when the correct edge is not found. For what happens _on_ the edge, it uses the logic it is supposed to be testing. If we want to improve that, we need a more careful specification of what _should_ be happening exactly. (In the test.)

michaz:
In this branch, I now removed the "early aborting" in LocationIndex, because then your test would succeed.

Not sure if we also want that in reality, because then _other_ tests fail. (Because then we only snap to nodes when the node is _really_ the closest position on the edge. Which is different from what we do now..)

michaz:
All of this also means that for a location index that is trivial enough (all the edges in the same tile), this test doesn't test a lot. The test and the testee basically do the same loop over all the edges.

karussell:
> Hence the difference / failing test

ok :) 

> Snap to a node iff it is really the node that is the closest location, typically behind a dead-end.

Can you explain what I need to do to see the failing tests then?

> Not sure if we are making sure that the random graph's spatial extent is big enough to get an interesting location index.

Oh, yeah, we need to make sure it creates also tests for a bigger area to test also other code of the location index.

michaz:
> Can you explain what I need to do to see the failing tests then?

Remove one or both of the conditional `return` statements in `LocationIndexTree.traverseEdge`.

michaz:
Screenshot 2021-01-04 at 15 24 57

Whoa! The future is now.

PR #2193, opened 2020-11-26
Improve pbf parser performance

otbutz:
This should speedup pbf parsing by about 10%

PR #2230, opened 2021-01-04
Remove handling of node tag "ele"

easbar:
The node tag "ele" is only meant to record above sea level elevation for special points like mountain peaks, but its not really meant to be used to provide elevation data for road networks:

From the [osm wiki](https://wiki.openstreetmap.org/wiki/Key:ele):

>  OpenStreetMap does not try to be a general elevation database so you should not tag elevation of nodes with no other tags and no specific meaning that suggests an elevation value is significant information in this case. 

Handling of this tag was only used in tests so far, so I removed it here. Similar changes to the ones here were already done by @otbutz in #2193.


gberaudo:
Hi @easbar, @otbutz, @karussell,

We were relying on this hidden feature in our project. Here is how and why it could become more and more common usecase for other graphhopper users.

The normal way of adding elevation in graphhopper is to use 2d raster sources like SRTM. It is straightforward and great but has drawbacks in the context of bridges or tunnels because the elevation will be taken from the terrain instead of the road network itself. As a consequence the resulting network will go down in the rivers or go up above the mountains, making any routing based on relief moot.

So we created our own OSM files with proper elevation on each node, and gave that to graphhopper, configured with a [special provider](https://github.com/camptocamp/graphhopper/blob/camptocamp_2.3/core/src/main/java/com/graphhopper/reader/dem/OSMNodeElevationProvider.java) and a [small hack](https://github.com/camptocamp/graphhopper/blob/camptocamp_2.3/reader-osm/src/main/java/com/graphhopper/reader/osm/OSMReader.java#L527-L532). It worked very well.

The road network and elevation we used at that time was "proprietary", but it is freely available now thanks to the trend of governments publishing opendata. So I am pretty sure there will be more and more demand of this kind of high quality routing and having it in standard in Graphhopper would be certainly attractive for more and more users.

We will invest the necessary time to restore that functionnality and maintain it in our branch. But we can as well do the work upstream if you think it would be valuable for Graphhopper.

otbutz:
I would rather prefer a dedicated elevation provider which is able to handle street level data to be honest. The current implementations are all tile based and if we really want to add support for this, we should do it in a clean and extensible way.

If we want to support it upstream, the following points are IMHO important:
* has to work with a regular OSM pbf
* elevation data is stored in an open and preferably standardized format
* free available test data in said format or at least easily convertible to it

karussell:
> As a consequence the resulting network will go down in the rivers or go up above the mountains

No, this is not the case. See: https://github.com/graphhopper/graphhopper/pull/798 and it is independent of the FlagEncoder since a few releases

easbar:
Thanks for describing your use case. I also think that there should be a way to integrate elevation data on the street level with GraphHopper. But I rather think it should be treated as a separate data resource so it won't be necessary to manipulate the OSM file. There are no official plans of adding elevation data to every OSM (street) node, right?

@gberaudo Do you use the OSM road network and attached elevation data to this? And if yes, how does this work? I mean assuming you have a separate data source containing elevation data, how did you match this to the OSM nodes? Is there a freely available example of this data? Or do you use a separate road network that already includes elevation data and you just converted this to the OSM format so you can feed it to GraphHopper's OSMReader?

gberaudo:
Thank you @otbutz, @karussell, @easbar for your quick replies.

Indeed, from a user perspective, I see 2 really different use cases related to elevation:
- A: enrich existing OSM data containing bad or incomplete elevation values with external rasters
Here the user has to choose which raster source to use during the graph generation, thus the elevationProvider class that handle this diversity.
- B: use 3D OSM data having all nodes with a correct ele value
Here it can be preprocessed OpenStreetMap data, or it can be data from other source, in the OSM format.
For this case the user has already decided the source of elevation and Graphhopper should simply use it.

@karussell, I was not aware of the intelligent bridget/tunnel interpolation capability of Graphhopper. It is interesting. The Swiss federation also publishes [very precise DEM rasters as open data](https://www.swisstopo.admin.ch/en/geodata/height/alti3d.html#download), which could probably give interesting results. However, the most precise dataset is around 1 terabytes, which may be challenging to use.
With the proper elevation provider, this is perfect for case A.

The federation also publishes a [road network](https://www.swisstopo.admin.ch/en/geodata/landscape/tlm3d.html) where all coordinates are already in 3D. @easbar, we convert that to the OSM format, keeping the elevation of each node. Then Graphhopper simply reads elevation from the nodes. It is very convenient because we are relying on the OSM standard and there are many tools to handle it.
Road and elevation are already present in the dataset, this is case B.

@otbutz, woudn't it be possible to stick with the OSM standard and simply expose a config option? would it cause a performance issue?
In my code, I only introduced the special OSMNodeElevationProvider to avoid messing with the Graphhopper configuration. But it does nothing. Ideally, I would have implemented it with a configration option: "use_osm_elevation=true". That option (false by default), would activate the code in the OSM parsing to extract the ele tag and then use the value.

otbutz:
> @otbutz, woudn't it be possible to stick with the OSM standard and simply expose a config option?

Strictly speaking, we would not follow the OSM standard, but only the OSM pbf file format. The usage of the `ele` tag is clearly defined and even contradicts how you're planing to use it:

> Elevation (height above sea level) of a point, in metres. This is mainly intended for mountain peaks but could also be used for elevation of airport runways and many other objects. For OpenStreetMap, this value should be in metres above above mean sea level as defined by the EGM96 geoid model. This elevation is usually very close to national "above sea level" systems with differences < 1m. **This is not the height above the WGS84 ellipsoid (see Geoid) which is shown as raw elevation by some satellite navigation devices and which can differ from geoid elevation by up to 100 m.**

gberaudo:
@otbutz, thanks I overlooked that information.

So here is the situation as I understand it:

- the OSM PBF (and XML) formats allow to specify elevation on the geoid;
- in the context of the OpenStreetMap project / database, the rule is to not use ele systematically, but only in a few cases;
- in Grapphopper, OSM ele should not be used, an elevationprofider doing a lookup by node id from an external file should be used instead
- user should rely on an ad hoc tool to populate that file from their actual source of data (shapefile, osm, ...)

I looked at existing formats but could not find one. @otbutz, do you have some idea?
Maybe an sqlite table (elevations(id number, ele: number) indexed by id) would do?

otbutz:
> @otbutz, thanks I overlooked that information.
> 
> So here is the situation as I understand it:
> 
> * the OSM PBF (and XML) formats allow to specify elevation on the geoid;

https://en.wikipedia.org/wiki/Geoid#Relationship_to_GPS/GNSS

OSM: height over the mean sea level (orthometric height)
SRTM and other: satellite data (ellipsoidal height)

>* in the context of the OpenStreetMap project / database, the rule is to not use ele systematically, but only in a few cases;
> 
>* in Grapphopper, OSM ele should not be used, an elevationprofider doing a lookup by node id from an external file should be used instead

Our current tilebased provider implementations are doing their lookup based on lat/lon:

https://github.com/graphhopper/graphhopper/blob/3324a85f7f1870781438ecebe69eaacbbfe6143b/core/src/main/java/com/graphhopper/reader/dem/ElevationProvider.java#L60-L63


>* user should rely on an ad hoc tool to populate that file from their actual source of data (shapefile, osm, ...)
> 
> 
> I looked at existing formats but could not find one. @otbutz, do you have some idea?
> Maybe an sqlite table (elevations(id number, ele: number) indexed by id) would do?

You might want to check if OpenLR would work for this usecase: https://www.openlr-association.com/method.html

It's opensource, has a compatible license, a Java implementation exists and is designed to augment one map dataset with values from a different one.

gberaudo:
@otbutz,

Regarding the height reference. Do you mean users should only use Graphhopper with ellipsoid heights only?
Is something wrong happening when using orthometric heights?

> Our current tilebased provider implementations are doing their lookup based on lat/lon:
Should a new base class acceptig the node id as parameter be created then?
`double getEle(number id)`

> You might want to check if OpenLR would work for this usecase: https://www.openlr-association.com/method.html
> It's opensource, has a compatible license, a Java implementation exists and is designed to augment one map dataset with values from a different one.

This method may be interesting to add elevation from a precise "map" to a map from OpenStreetMap data. Question: should it be part of Graphhopper or of an independant preprocessing tool? The later would offer the larger possibilities (nut just handling elevation) and avoid having to maintain the functionnality in Graphhopper; but we would be back to the initial use case: supporting OSM data with ele tag.

easbar:
> Indeed, from a user perspective, I see 2 really different use cases related to elevation:

Yes, I understand. In one case we can provide an elevation source that returns elevation data given a lat,lon pair and in the other case we explicitly provide elevation data for every OSM (highway) node. I imagine there could even be a combination of both. For example GraphHopper supports elevation interpolation where additional helper nodes are created to maintain elevation details (see #1953). This requires the first kind of elevation source, but could be useful also when we use the second. Or, alternatively, such interpolation is done beforehand.

Anyway, I think both use cases should be supported by GraphHopper and at the very least they should be possible with minor adjustments when they are needed. Since the first kind is already supported via `ElevationProvider` the question is how we can support he second kind. We'd need a mechanism to include elevation data based on OSM node IDs. Note that GraphHopper IDs won't work, because only tower nodes (junctions) have an ID and pillar nodes (those inbetween) do not. So in theory a lookup table OSM node ID -> elevation would work, but I also like to think of the elevation data as node tags, because then this lookup won't be required. We should just use any tag but *not* the `ele` tag for this (because it has another meaning). Or even better I think this shouldn't even matter. Ideally, the OSMReader would be configurable (or easy to modify) to consider arbitrary node tags. And since this is a goal for the upcoming OSMReader rewrite anyway I don't think it will be hard to support this elevation use case as well.

And @gberaudo isn't it almost trivial to do this already with the current OSMReader? It looks like all you need to do is get whatever tag you use (can be `ele` as well) here and use this for the elevation:

https://github.com/graphhopper/graphhopper/blob/8243cf2c21b4e8c4b4a06c3b186c57f96b3b1a60/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java#L495-L504

It might be as simple as this:

```java
//      double ele = eleProvider.getEle(lat, lon);
        double ele = node.getTag("my_ele", myDefaultValueInCaseTheMyEleTagIsMissing);
```

otbutz:
Is the mapping even done by OSM ID? e.g the road network [link](https://www.swisstopo.admin.ch/en/geodata/landscape/tlm3d.html) provided by @gberaudo seems to be a regular ESRI shapefile. I think we still need to lookup elevation based on lat/lon. The difference for me is that instead of a tilebased approach we would have something like the LocationIndex which would lookup the nearest line/point with elevation data.

easbar:
No, as far as I understood @gberaudo has a complete road network (including elevation for all nodes) and converts this to the OSM format so it can be read by GraphHopper (I think a very reasonable approach).

gberaudo:
@otbutz, the lines in the shapefile are 3D; the resulting osm.pbf have an elevation for every node. Since all the data is in the network, there is no lookup during the Graphhopper import.

> And @gberaudo isn't it almost trivial to do this already with the current OSMReader? It looks like all you need to do is get whatever tag you use (can be ele as well) here and use this for the elevation:

Indeed, it is actually quite easy to implement! Thanks @easbar.
I went forward and implemented your idea in my local fork. Here is the draft PR: #2373.
I have a few issues with the tests; will fix that if you think it is a good addition to Graphhopper.

otbutz:
I'm not sure if we should mix the `ele` tag with our default ElevationsProviders as those are two different models.

Again citing the OSM wiki:

> This is not the height above the WGS84 ellipsoid (see Geoid) which is shown as raw elevation by some satellite navigation devices and which can differ from geoid elevation **by up to 100 m**.

If someone uses regular OSM pbfs with this workaround, i'd expect breakage.

gberaudo:
@otbutz, I understand that:
- the OSM `ele` tag contains orthometric heights (height from the geoid);
- while the SRTM data contains ellipsoid heights (like the GPS).
These are incompatible and clearly should not be used together.

That is why the draft PR does not use the `ele` tag, but requires to provide an explicit tag and is disabled by default.
I can describe shortly the difference in the documentation.

Then it is the choice of the user:
- either preprocess to have osm.pbf files with all data (then she will not activate elevation providers);
- or choose compatible heights (like the TLM and raster from Swisstopo, which are all relative to the geoid) and activate the read from node + fallback mechanism proposed by @easbar.

@otbutz, does it clarify?

otbutz:
Let's move this discussion to #2373

PR #2233, opened 2021-01-14
Remove DataReader interface, stop using GraphHopperOSM

easbar:
Now that OSMReader is in core, we do not really need GraphHopperOSM anymore. Also the DataReader interface is no longer useful. Custom data reading is best implemented by overriding `GraphHopper#readData` now.


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -322,19 +333,19 @@ public GraphHopper setGraphHopperLocation(String ghLocation) {
     }
 
     public String getDataReaderFile() {
-        return dataReaderFile;
+        return osmFile;
```
easbar, line 336: Or should we rename it to `getOSMFile`? This is how `GraphHopperOSM` calls this method.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -623,53 +636,38 @@ private GraphHopper process(String graphHopperLocation, boolean closeEarly) {
...
-        }
     }
 
-    protected DataReader importData() throws IOException {
-        ensureWriteAccess();
-        if (ghStorage == null)
-            throw new IllegalStateException("Load graph before importing OSM data");
-
-        if (dataReaderFile == null)
+    protected void readData() {
```
easbar, line 641: Should we rename it to `read/importOSM`? It used to be private.

karussell (approved):
Thanks!

easbar:
Ok, lets do further changes in subsequent commits:

1) rename get/setDataReaderFile -> get/setOSMFile
2) datareader.file -> osm.file
3) remove GraphHopperOSM

1) seems simple enough to me. 2) I am not sure because it will break so many/all configurations and maybe we will have to rename it again in case we make the configuration more hierarchic/modular? Sure considering that the java variable is called `osmFile` this might be a bit counter-intuitive, but then again it always was? Same thing for 3). Removing it forces lots of down-stream code to change while keeping it means little extra cost for us at the moment. I think we should just deprecate it here and remove it for GH 4.0?

koch-t:
> Now that OSMReader is in core, we do not really need GraphHopperOSM anymore. Also the DataReader interface is no longer useful. Custom data reading is best implemented by overriding `GraphHopper#readData` now.

Sorry for kicking an old PR, so mostly not relevant, but if `GraphHopper#readData` is the preferred way to read data, shouldn't this method be public instead of protected?

easbar:
Which version are you looking at? `GraphHopper#readData` no longer exists in the current master branch.

PR #2237, opened 2021-02-01
Move frontend code to web-bundle

easbar:
Here I moved the frontend code (maps, pt, isochrone) from the web module to the web-bundle module. This way it can be included by any server that uses web-bundle as dependency.

I moved everything from 

web/src/main/resources/assets to 
web-bundle/src/main/resources/com/graphhopper/maps/assets

I think assets is the default path for Dropwizard resources, but since web-bundle is not a Dropwizard server we should rather put the files into some namespaced path (com/graphhopper). We could also just use ../com/graphhopper/maps/ instead of ../com/graphhopper/maps/assets?


web/src/main/java/com/graphhopper/http/GraphHopperApplication.java
```diff
@@ -45,9 +45,11 @@ public void initialize(Bootstrap<GraphHopperServerConfiguration> bootstrap) {
         bootstrap.addCommand(new MatchCommand());
 
         Map<String, String> resourceToURIMappings = new HashMap<>();
-        resourceToURIMappings.put("/assets/", "/maps/");
+        resourceToURIMappings.put("/com/graphhopper/maps/assets", "/maps/");
         resourceToURIMappings.put("/META-INF/resources/webjars", "/webjars"); // https://www.webjars.org/documentation#dropwizard
         bootstrap.addBundle(new ConfiguredAssetsBundle(resourceToURIMappings, "index.html"));
+        // todonow: shall we remove configured assets bundle (and use dropwizard-assets instead)?
+//        bootstrap.addBundle(new AssetsBundle("/com/graphhopper/maps/assets/", "/maps/", "index.html"));
```
easbar, line 52: I was hoping to fix the problem that pt and isochrone require a tailing slash, but using AssetsBundle instead of ConfiguredAssetsBundle did not help unfortunately. Still, why do we need configured assets bundle? When we start the server from the IDE we can still change the code without creating a new jar.

karussell: Did you try to change something in JavaScript and did this change appear with `npm watch` & refresh? I can remember that there was a problem with this.

easbar: Yes: https://github.com/graphhopper/graphhopper/pull/2238
But I could not make it work for src/main/resources/assets and if that does not work I see no reason to keep configured-assets-bundle.

easbar: Ah no wait. It works for index.html even for src/main/resources/assets. The reason I thought it does not work is that the javascript code is copied to target/classes/assets (using `npm run watch`)... The 'override' is not really an override because dropwizard still reads from target/classes/assets (by default).

easbar: I think we should remove it as in #2238. I have already merged #2238 into this PR.

michaz (approved):

web/src/main/java/com/graphhopper/http/GraphHopperApplication.java
easbar: See #2238

easbar:
I removed the intermediate subfolder, maps is now in web-bundle/src/main/resources/com/graphhopper/maps

PR #2238, opened 2021-02-01
Remove configurable-assets-bundle

easbar:
The configurable-assets-bundle was used so we can edit the maps JavaScript code while the server is running(?). However, we were already pointing the bundle override into web/target/classes in config-example.yml and if we change the files there our changes will be reflected in the browser already without the configurable-assets-bundle. Therefore I do not see why we need this?

With this PR we (still) can edit main-template.js or any of the other javascript files of the map client and see the changes after a page reload byrunning `npm run watch` which will browserify all the files and copy the resulting `main.js` file to web/target/classes/assets whenever we make a change. Changes to `index.html` are not reflected and we would need some mechanism that updates `index.html` in the target folder after making a change to achieve the same effect. In any cas the configurable-resource-bundle would only be beneficial if we pointed it to web/src/main/resources/assets, but I could not make this work and apparently config-example.yml was changed to point to the target folder (maybe for the same reason).

### DIFF
--- a/web/src/main/java/com/graphhopper/http/GraphHopperApplication.java
+++ b/web/src/main/java/com/graphhopper/http/GraphHopperApplication.java
@@ -42,6 +42,8 @@ public final class GraphHopperApplication extends Application<GraphHopperServerC
         bootstrap.addCommand(new ImportCommand());
         bootstrap.addCommand(new MatchCommand());
         bootstrap.addBundle(new AssetsBundle("/assets/", "/maps/", "index.html"));
+        // see this link even though its outdated?! // https://www.webjars.org/documentation#dropwizard
+        bootstrap.addBundle(new AssetsBundle("/META-INF/resources/webjars", "/webjars/", null, "webjars"));
     }
 
     @Override
--- a/web/src/main/java/com/graphhopper/http/GraphHopperApplication.java
+++ b/web/src/main/java/com/graphhopper/http/GraphHopperApplication.java
@@ -22,14 +22,12 @@ import com.graphhopper.http.cli.MatchCommand;
 import com.graphhopper.http.resources.RootResource;
 import com.graphhopper.navigation.NavigateResource;
 import io.dropwizard.Application;
-import io.dropwizard.bundles.assets.ConfiguredAssetsBundle;
+import io.dropwizard.assets.AssetsBundle;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
 
 import javax.servlet.DispatcherType;
 import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.Map;
 
 public final class GraphHopperApplication extends Application<GraphHopperServerConfiguration> {
 
@@ -43,11 +41,7 @@ public final class GraphHopperApplication extends Application<GraphHopperServerC
         bootstrap.addBundle(new RealtimeBundle());
         bootstrap.addCommand(new ImportCommand());
         bootstrap.addCommand(new MatchCommand());
-
-        Map<String, String> resourceToURIMappings = new HashMap<>();
-        resourceToURIMappings.put("/assets/", "/maps/");
-        resourceToURIMappings.put("/META-INF/resources/webjars", "/webjars"); // https://www.webjars.org/documentation#dropwizard
-        bootstrap.addBundle(new ConfiguredAssetsBundle(resourceToURIMappings, "index.html"));
+        bootstrap.addBundle(new AssetsBundle("/assets/", "/maps/", "index.html"));
     }
 
     @Override
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ android/libs/graphhopper-*-android.jar
 .idea/
 *iml
 /web/src/main/resources/assets/js/config/options_prod.js
+/web/src/main/resources/assets/js/main.js
 /web/dependency-reduced-pom.xml
 /web/node
 *.pbf
--- a/config-example.yml
+++ b/config-example.yml
@@ -172,15 +172,6 @@ graphhopper:
   # Having less rules, might result in a smaller graph. The line below contains the world-wide bounding box, uncomment and adapt to your need.
   # spatial_rules.max_bbox: -180,180,-90,90
 
-
-# Uncomment the following to point /maps to the source directory in the filesystem instead of
-# the Java resource path. Helpful for development of the web client.
-# Assumes that the web module is the working directory.
-#
-# assets:
-#  overrides:
-#    /maps: web/target/classes/assets/
-
 # Dropwizard server configuration
 server:
   application_connectors:
--- a/docs/core/quickstart-from-source.md
+++ b/docs/core/quickstart-from-source.md
@@ -96,15 +96,6 @@ as those versions are not in maven central:
 
 ### JavaScript
 
-When developing the UI for GraphHopper you need to enable serving files
-directly from local disc via your config.yml:
-
-```yml
-assets:
-  overrides:
-    /maps: web/src/main/resources/assets/
-```
-
 To setup the JavaScript development environment install the [node package
 manager](https://github.com/nvm-sh/nvm):
 
@@ -122,7 +113,7 @@ production:
 ```bash
 cd web
 
-# For development just use watchify and all changes will be available on refresh:
+# For development just use watchify and all changes (except those in index.html) will be available on refresh:
 npm run watch
 
 # bundle creates the main file
--- a/docs/core/quickstart-from-source.md
+++ b/docs/core/quickstart-from-source.md
@@ -113,7 +113,12 @@ production:
 ```bash
 cd web
 
-# For development just use watchify and all changes (except those in index.html) will be available on refresh:
+# For development just use watchify, which updates the main.js file whenever you change one of the .js files:
+# To see your changes in the browser without restarting the server you can either run the GH server in debug mode from
+# IntelliJ and use `Run->Debugging Actions->Reload Changed Classes` (and refresh your browser window). Or you serve
+# GH maps from the web/src/main/resources/assets folder directly using e.g. `live-server` (npm install -g live-server).
+# For the latter case make sure GH maps points to your running GH instance (set the routing host in js/config/options.js
+# to e.g. http://localhost:8989/).
 npm run watch
 
 # bundle creates the main file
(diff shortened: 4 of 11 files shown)


easbar:
Sorry no, its a bit more complicated than this...
I think we have two:

1) use the configurable-assets-bundle and point it to `src/main/resources/assets`. This way the resources are loaded from the location on disk and we can make changes directly. *However*, the javascript code will *not* be updated because it is copied to `target/classes/assets` by `npm run watch`. We could fix this by adjusting the `watch` script maybe. At least in theory. When I tried this I still did not see updates probably because Dropwizard was *still* reading the `main.js` file from target/classes/assets. Basically it reads from both paths and things get complicated.

2) remove the configurable-assets-bundle and always read from `target/classes/assets`. Since our `npm run watch` command compiles and copies our javascript to `target/classes/assets` anyway we will see our changes of the java script code. However, we do not see changes of the `index.html` file, because `npm run watch` does not copy it to the target folder (but we could add something like this to the watch script).

easbar:
To me it does not seem to be worth the extra plugin and configuration (and further changes) just to be able to hot-reload index.html. Also there are better ways to do this. For example running a separate hot-reloading server from the assets folder. There is not really a need to serve GH maps from the GH server while developing the UI.

web/src/main/java/com/graphhopper/http/GraphHopperApplication.java
```diff
@@ -43,11 +41,7 @@ public void initialize(Bootstrap<GraphHopperServerConfiguration> bootstrap) {
         bootstrap.addBundle(new RealtimeBundle());
         bootstrap.addCommand(new ImportCommand());
         bootstrap.addCommand(new MatchCommand());
-
-        Map<String, String> resourceToURIMappings = new HashMap<>();
-        resourceToURIMappings.put("/assets/", "/maps/");
-        resourceToURIMappings.put("/META-INF/resources/webjars", "/webjars"); // https://www.webjars.org/documentation#dropwizard
```
easbar, line 49: Apparently this is no longer needed with the default assets bundle. At least maps/pt/ and maps/isochrone/ are working without it.

karussell:
It is not only index.html but likely also any change to the css file or icon etc (?)

We can surely do this change, but we need a documented and working way to develop the UI. Maybe we decouple developing the UI as you suggested and serve the UI e.g. via http-server? (for the end product this won't be required as we bundle it in the maven build)

easbar:
> It is not only index.html but likely also any change to the css file or icon etc (?)

Yes. `npm run watch` bundles and copies the javascript files to target/classes/assets, while all the static files like *.css, index.html, image files etc. simply remain in src/main/resources. Actually I think this is where most of the confusion comes from. Why do we copy the bundled file (`main.js`) to target/classes/assets manually? This should be done by maven instead. Maven does this just like it does for index.html etc.

> We can surely do this change, but we need a documented and working way to develop the UI. Maybe we decouple developing the UI as you suggested and serve the UI e.g. via http-server? (for the end product this won't be required as we bundle it in the maven build)

Yes there are many options. If we fix the paths in package.json (see my last commit) we can already get 'hot reload' without the configured assets bundle. It can be done like this:

* npm run watch # this bundles the javascript files and updates `main.js` (in src/main/resources) whenever we change a javascript file
* start the GH server in debug mode
* make changes to static resources, javascript (or even Java files)
* in IntelliJ: run -> debugging actions -> reload changed classes
* refresh your browser window

This works and we can apply changes to all files (even Java), but the hot reload is rather slow and we need to manually refresh the browser. Therefore a better option (in case we are *only* developing the UI) is this:

* start the GH server (say on port 8989 which is the default)
* npm install -g live-server
* npm run watch (as before) 
* in web/src/main/resources/assets/js/options point GH maps to your local GH server using this line: `routing: {host: 'http://localhost:8989', api_key: ''}`
* cd web/src/main/resources/assets && live-server # this will serve GH maps from this folder from port 8080 by default

This way we can develop GH maps and we do not even have to reload the browser page, but it will reload automatically if we focus the browser window.

My favorite solution for this, however, is using VSCode with the Live Server plugin. It works the same way except that we do not run live-server from the command line but simply start live server for the assets folder from VSCode. I like this better because the browser window reloads even without focusing the browser and we can just keep developing and watch our changes on another screen for example. I don't know why this works and does not work with the live-server package from npm.

karussell:
> Actually I think this is where most of the confusion comes from. Why do we copy the bundled file (main.js) to target/classes/assets manually?

Not sure I understand this. We do not copy main.js manually IMO, we create it from the other JS files and put it into the target folder so that it gets served and live editing works. How would maven help us here?

Or do you mean why we put main.js into `target/..` and not `src/..`? I think the problem is that main.js is not really a source and so we do not want it in our git repo. Ok, we could ignore it via git but it still would feel strange to me as the sources and main.js will then be mixed.

> Yes there are many options

I would describe a solution that can be used with any editor like with `live-server` (or even `http-server` as I would not care too much that I have to reload to see changes -> this is necessary in the current situation too).

> in web/src/main/resources/assets/js/options point GH maps to your local GH server using this line

Why is this necessary? Isn't this possible to do with the npm watch command?

easbar:
> Not sure I understand this. We do not copy main.js manually IMO, we create it from the other JS files and put it into the target folder so that it gets served and live editing works. How would maven help us here?

Maven copies resource files to the target folder as part of our 'build'. If we copy main.js to target we 'manually' interfere with this build system. This is why the hot reload does not work properly: If we start e.g. the GH server from IntelliJ it tries to simulate our build (maven), but once we manipulate the target folder some other way (e.g. by running `npm run watch` which copies a file (main.js) to target that does not exist in src) there is confusion because the files in target and src are no longer in sync.

> Or do you mean why we put main.js into target/.. and not src/..? I think the problem is that main.js is not really a source and so we do not want it in our git repo. Ok, we could ignore it via git but it still would feel strange to me as the sources and main.js will then be mixed.

No we should not add main.js to src (as it was some time ago). Still, I think browserify/watchify should create the file in src (and we should leave it up to maven to copy it to target). Its true its not a source file in the sense that we write it ourselves, but it is a source file in the sense that it is something maven should find and add to our build target. The fact that we use browserify to build  the main.js file from some other js files is just an implementation detail so to say. For maven the only important thing is that this file exists and that it should be added to the jar. To do this in a really clean way I would even argue that the 'source' javascript files should not go into src/main/resources, because they are just 'source' files for main.js and should *not* go to the jar.

> I would describe a solution that can be used with any editor

Yes I was also hesitant to mention another tool. `live-server` would be ok probably because we need to install node anyway. Do you mean `python http.server`? So far we have no python dependencies and on Windows its probably not installed by default? I would prefer `live-server` also because of the automatic reload.

> I would not care too much that I have to reload to see changes 

You should really give it a try, I think its very useful (but only if you do not even have to focus the browser as with the VSCode live-server plugin) :)

> Why is this necessary? Isn't this possible to do with the npm watch command?

Hm the npm watch command only re-creates main.js from the js files whenever there is a change. But once we serve GH maps from *some* server other than the GH server (e.g. using live-server or http.server) we have to set the url of the GH server or how else should the routing requests end up in the right place?

easbar:
I updated the docs mentioning both the IntelliJ and the live-server methods.

karussell:
> Maven copies resource files to the target folder as part of our 'build'

Do you mean IntelliJ? I mean, for maven the build process is already working, but here we need to copy the main.js file to `target` whenever there is a change and this is not something that maven does for us (?)

> To do this in a really clean way I would even argue that the 'source' javascript files should not go into src/main/resources, because they are just 'source' files for main.js and should not go to the jar.

Probably the most correct place (in maven-terms) for main.js would be `target/generated-sources`?

> do you mean python http.server

I meant http-server from npm

> only if you do not even have to focus the browser as with the VSCode live-server plugin) :)

Ah, I think I only know this from deck.gl and this "late auto reload" was quite ugly and not really a time safer.

> But once we serve GH maps from some server other than the GH server

Ok, understood

easbar:
> Do you mean IntelliJ? 

No I meant maven, but well, its kind of the same thing. Maven (run from the command line) builds and copies to target folder (and adds the same files to the jar). But IntelliJ simulates the same as it adds all the files to the classpath when e.g. starting the server.

> I mean, for maven the build process is already working, but here we need to copy the main.js file to target whenever there is a change and this is not something that maven does for us (?)

No! If we change main.js in src/main/resources and reload the classes in intellij the updated file is refreshed in the classpath of the running server as well. If we copy main.js to target/assets in theory we do not need to reload the classes, but then we have inconsistent behavior, because things aren't like they are expected to be. Or to put it the other way around. Assuming we put it into target/assets at which folder would you point the configurable assets plugin? If we point it to target/assets we no longer get live updates for index.html and if we point it to src/main/resources we no longer see updates of main.js. 

> Probably the most correct place (in maven-terms) for main.js would be target/generated-sources?

main.js is a non-Java file that is supposed to be added to the class path so it should be in src/main/resources IMO. We should not touch the target folder at all. The target folder is just maven's output folder. And like I said if anything then we should move the javascript 'source' files out of src/main/resources because they are not resources.

> I meant http-server from npm

Ah ok, but its probably similar. Hm I like the live reload of live-server but never tried http-server. Personally I would use VSCode with the live server plugin anyway for the reasons I mentioned earlier.

> Ah, I think I only know this from deck.gl and this "late auto reload" was quite ugly and not really a time safer.

Ok :)

easbar:
Should we do smaller steps and:

* make browserify create main.js in src/main/resources
* add main.js to .gitignore
* keep the plugin for now and point it to src/main/resources

This way we can move on with #2237 and keep the same workflow for now (and it will actually work in contrast to current master).

karussell:
Ok, have made some minor changes to make it working with live-server

> Ok :) What do you mean with 'late' reload? 

The browser reload only happened when I viewed/focussed the browser window. So not much faster then quickly switching app windows+F5. 
btw: it seems that live-server sometimes reloads it immediately and sometimes requires the focus. Maybe I did something wrong.

karussell (approved):

easbar:
Just for later reference. We now removed the configurable-assets-bundle and build main.js in src/main/resources for now. Hot-reload can be achieved either using IntelliJ reload classes or the npm live-server package. See docs for instructions.

Later we should think about:

* moving javascript sources to src/main/js or similar so they do not end up in the jar and also live-server does not get triggered multiple times.
* creating main.js in yet another location, because its not really a source file...

Issue #2232, opened 2021-01-11
Clarify questions related to customizable routing


karussell:
Follow up issue of #2209 

* [x] Remove space characters from `multiply by`, `limit to` and `else if` and use _
* [x] Allow `"else": ""` not only `"else": null` (important for JSON)
* [x In documentation: should we focus on JSON format so that users of the web API can directly use it and only mention YAML for simpler creation of server-side models? -> opened #2275
* [x] If production format is decided, eventually move /route-custom into /route via a separate `custom_model` entry. This would also allow integration e.g. into /isochrone -> #2252 
* [x] improve UI and auto complete feature (I found [Ajv](https://ajv.js.org/) and [this example for the editor] (https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-configure-json-defaults)) (see #2239) 
* [x] make server-side models without separate file possible
* [x] should we add the `set_to` operator (yet)? -> later
* [x] should we add description/comment fields to statements? -> later
* [x] send custom request when markers are moved and custom model box is open? -> opened #2276
* [x] keep current selection of path-detail diagram in ui-state so it remains when we send another custom request (and it is still available) -> #2258

easbar:
> improve UI and auto complete feature

I think this is the most important thing to improve by far. Users need to get proper feedback for their actions: Syntax checking is important to not get frustrated with annoying little errors, auto-completion is important to easily see what can be customized and interactively learn the custom model language. Ultimately we should also be able to 'understand' why some route was chosen over another. For this we will need to be able to visualize the edge weights and the edge properties (but this won't really happen without a proper rewrite of GH maps, so probably for another issue).

> which points are missing for a production-state format?

I guess this becomes easier to find out the more people are use the feature (therefore the first point is so important IMO)

> In documentation: should we focus on JSON format so that users of the web API can directly use it and only mention YAML for simpler creation of server-side models?

This is a tricky question. I expect most users to start with the GH maps UI and enter some custom models in the flex box to see how their example routes change. The format that is entered there should be the same as in the documentation. On the other hand I expect it will be easiest to give proper hints / provide some kind of 'autocomplete' if we use the YAML format (with one statement per line) in the flex box. When a user is satisfied with the result it should be easy enough to translate the YAML format to JSON. We could even offer this in the UI: 'show current model as JSON'.

So with this reasoning I would suggest to:

[...]

PR #2239, opened 2021-02-16
Interactive custom model text box

easbar:
This is work in progress to realize a text input box that allows entering a custom model with error highlighting and auto-completion. The goal here is to provide a very flat learning curve for the custom model language. The custom model is written in YAML that can contain boolean expression strings that refer to encoded values like `road_class == PRIMARY` etc.

Since the encoded values are unknown to most users they are provided as auto-completion suggestions. The auto-completion can be triggered by pressing Ctrl+Space. Note that the auto-completion suggestions depend on the cursor position. If the cursor is positioned within some word the conditions are filtered by the characters between the beginning of the word and the cursor.

`|road_class` + Ctrl+Space yields `surface, road_class, road_class_link, toll, ...`, but `ro|ad_class` + Ctrl+Space only yields `road_class, road_class_link, ...` (all values starting with `ro`)! Is this what most people would expect?

Usage: see the README.md file in the textbox folder

Currently the boolean expressions are colored in blue, errors in boolean expressions are red and errors in yaml format are indicated by red underlines.

todo:

- [x] calculate error ranges also for YAML errors related to null values 
- [x] allow multiple 'else if' (there is a minor bug that prevents this currently)
- [x] move YAML error messages inside text editor (tooltip or maybe error indicator at line number). for a tooltip maybe take a look here: https://github.com/codemirror/CodeMirror/blob/master/addon/tern/tern.js#L595
- [x] support boolean and numeric encoded values
- [x] support boolean literals `true`, `false`
- [x] integrate with GH maps
- [x] ~~adjust /info endpoint to return encoded value types (numeric/boolean/enum)~~ not needed -> show ev from /info
- [x] fix CI build: `Error: Can't walk dependency graph: Cannot find module 'custom-model-editor' `
- [x] replace string.startsWith (done) ~~and also check IE/browser compatibility in general~~ IE won't be supported
- [x] set initial content to make it easier to get started? could be comments as well...
- [x] read-only view for json (to be able to copy&paste the model to request)
- [x] colored frame of the editor on valid/invalid input? or at least disable search button as long as the input is invalid.
- [x] support area encoded values (?)
- [x] ~~minify codemirror css files~~ (not worth it), add version information as well
- [x] update 'Help' link to point somewhere the new syntax is explained
- [x] remove js-yaml (we only need one yaml parser)
- [x] update documentation for separate editor module

further improvements (maybe for a follow up PR):

- [ ] auto-complete for YAML (?), especially speed/priority/areas/distance_influence/limit to/multiply by would be nice
- [ ] support negation operator `!`(?)
- [ ] support boolean encoded values without `==/!= true/false`?
- [ ] auto-complete for numeric encoded values? maybe just show a hint saying that one needs to enter a number?
- [ ] simplify error messages (now that they are shown at their origin) ?
- [ ] maybe trigger autocomplete automatically without pressing Ctrl+Space? At least in certain cases? This way maybe users not familiar with auto-complete get to know it more easily.
- [ ] maybe use tooltips to provide some documentation, e.g. for speed/priority/distance_influence
- [ ] check syntax for areas section (geojson) (?)
- [ ] maybe use code mirror bracket matching
- [ ] adjust colors, theme and styles
- [ ] make it easy to translate error messages? but first make them as simple as possible.
- [ ] check code mirror options to see if we need anything besides the defaults, maybe customize the 'mode'?
- [ ] handle YAML multiline syntax like `>+` and `|`... https://stackoverflow.com/a/21699210
- [ ] dark theme and VIM keybindings (just kidding...)


easbar:
@karussell I found this code in `main-template.js`:

https://github.com/graphhopper/graphhopper/blob/16a325382dc783352c07393317b8737f2d68f7c9/web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js#L176-L183

It looks like it takes a custom model from the url parameters? Is this an actual working feature or can it be removed? So far I thought it would be nice to be able to store custom models inside the url, but it was not possible...

karussell:
Thanks for this incredible work :)

> Is this an actual working feature or can it be removed? 

For development it should work but we can also remove it.

easbar:
> For development it should work but we can also remove it.

Yes it does. One can use the load_custom parameter to point to an external yaml file that will be copied to the custom model box :+1: Speaking of hidden features: Pressing Ctrl+Enter while the cursor is in the box to update the current route is also quite useful. The only annoying thing can be the automatic zoom-to-route in this case.

I now integrated the new text box in GH maps and both features are still available (but flex=true is no longer necessary for the first one).

easbar:
The CI builds fail, because I am including the custom model editor as local npm module like this:

```json
 "dependencies": {
    "d3": "5.11.0",
    "flatpickr": "4.4.6",
    "custom-model-editor": "file:src/main/js/custom-model-editor",
```

This way a symlink pointing to the given path will be created in `node_modules`. Apparently on travis, appveyor and github packages this symlink cannot be resolved so the build fails. Not sure what to do here :(

karussell:
> I now integrated the new text box in GH maps 

Nice :+1: 

> This way a symlink pointing to the given path will be created in node_modules.

Instead of the package.json why can't we do a relative require statement like

easbar:
> Instead of the package.json why can't we do a relative require statement like

Yes, good idea, this works :+1: Its not ideal, because this way we have to run `npm install && npm run build` from the custom-model-editor directory separately, but this can be done via the maven frontend plugin at least. See my last commit.

easbar:
In the last commit I added a button that switches the editor to a read-only json view

image

I'd like to disable it as long as the yaml input is invalid. WDYT is this useful? Can/should we change these three 'buttons' from anchors to buttons? Or how can we make it appear disabled? Maybe use ~~strikethrough~~, or even remove it entirely?

Btw I also disable the search button as long as the input is invalid now, does this make sense?

karussell:
Over one thing I stumbled: before I was able to do the usual `npm run watch` I had to either build the jar via maven or run `cd src/main/js/custom-model-editor/ && npm install && npm run build` and restart the GH server.

> In the last commit I added a button that switches the editor to a read-only json view

Nice idea!

And should I do a reimport of our customizable routing server in its current state? (and rewrite the examples&screenshots?)

btw: I just picked a random route and thought something is really broken as it did not avoid MOTORWAY at all. No matter what I did. The problem was that although in the map it looked like a motorway it was a TRUNK (I have a strange luck ;) ) ... maybe we should send `details:[road_class]` ... or better ALL used encoded values :D ? 

And previously it worked when I included `details: [road_class]` in the yaml but we are now too strict :) 

(also a bit ugly that the elevation widget currently only works if elevation is enabled :) )

> Can/should we change these three 'buttons' from anchors to buttons?

text is fine IMO. Only a minor style adaption so that it matches the sidebar footer? (greenish text without underscore)

easbar:
> Over one thing I stumbled: before I was able to do the usual npm run watch I had to either build the jar via maven or run cd src/main/js/custom-model-editor/ && npm install && npm run build and restart the GH server.

Since we now include custom-model-editor via require(...) only instead of package.json we have to build it separately yes. You can also do `npm run watch` in the custom-model-editor folder to get live updates.

> And should I do a reimport of our customizable routing server in its current state? (and rewrite the examples&screenshots?)

Sure! We cannot use areas yet though.

> ... maybe we should additionally send details:[road_class] ... or better ALL used encoded values :D ?

You mean we should show all encoded values when we hover edges in the mvt view? Yes!

> And previously it worked when I included details: [road_class] in the yaml but we are now too strict :)

What do you mean? What are details in yaml?

> (also a bit ugly that the elevation widget currently only works if elevation is enabled :) )

Yes, since it is also showing path details this feature should not depend on elevation...

> text is fine IMO. Only a minor style adaption so that it matches the sidebar footer? (greenish text without underscore)
And for disabling: gray it out?

Ok will try.

karussell:
> What do you mean? What are details in yaml?

Since we (currently) have the properties like `speed` on the same JSON root level like the other request parameters we would request path details via `details:[...]`. But as the plan is to change this (in #2232) it is probably not worth a thought :)

> Since we now include custom-model-editor via require(...) only instead of package.json we have to build it separately yes. You can also do npm run watch in the custom-model-editor folder to get live updates.

We would need to have this documented or include the editor JS files like we include other JS files?

> You mean we should show all encoded values when we hover edges in the mvt view? Yes!

I didn't mean this but it would be nice :)

What I meant is that we collect all encoded values used in the script and include it in the query as path details. This way we can easily see the e.g. the road_class for the route and how it changes if we change the number.

easbar:
> Since we (currently) have the properties like speed on the same JSON root level like the other request parameters we would request path details via details:[...]. But as the plan is to change this (in #2232) it is probably not worth a thought :)

Ah ok. Btw so far you can use ctrl+enter to send the request regardless of the validation. Not sure if this should be a 'feature' :) So far the validation check is just not implemented in this case...

> We would need to have this documented or include the editor JS files like we include other JS files?

Ok sure it should be documented. We cannot include editor files like other files because the editor uses an additional transpilation step.

> What I meant is that we collect all encoded values used in the script and include it in the query as path details. This way we can easily see the e.g. the road_class for the route and how it changes if we change e.g. the multiply-by number. (if the elevation widget is shown)

Ok yes that makes sense. It should not be too hard to return the used encoded values from the expression parser.

easbar:
Its still not working with IE11, I get a `SCRIPT5021` error that is due to a character range `\u{D800}-\u{DFFF}` used in the yaml parser. Apparently it is supposed to support IE11 though: https://github.com/eemeli/yaml/pull/73. Maybe I am not correctly making use of it's 'browser build' yet.

Issue #2085, opened 2020-07-15
Graph Storage Improvements


easbar:

easbar:
From @karussell (originally in #2084):

I'm was investigating how to store the graph more compact and found something that improves the situation for wayGeometry. This is still in its early shape but basically uses a "vint" (see Lucene) with delta encoding like described in #234. To make this more effective the whole graph structure would need a change so that the "edge rows" would not need to be of same length. And this new graph storage (basically a persistent linked list that avoids duplicating edge information) would also allow a more efficient turn cost storage as 1. only the turn cost needs to be stored (no additional node id info) and 2. it would likely need only one byte per tc info. So we would want this if we start storing and reading turn costs for every junction.

>   It creates a new IntsRef object on every call, even though all it returns is a double

The funny or sad thing here is that the JVM seems to heavily optimize these things and for the wayGeometry work I found it did not make a difference if I reused or created the byte array. But here it could make a difference when we call it for every node in the algorithm and not only for the location lookup.

easbar:
Ok cool :) Maybe a bit off-topic here but another thing that I thought could be useful in different places would be separating the graph topology (information about which edges are adjacent to which nodes) from the edge properties. So we could have a data structure that allows fast graph traversal and only gives us edge ids/keys (basically the `nodeA/B` and `nextA/B` columns from the edge storage) and then a separate storage for 'edge data'. Maybe this way it would be easier/more natural to a) 'attach' additional data to edges (simply by id) and b) to optimize the graph traversal for different use-cases. For example think of Dijkstra running for the car encoder. In principle we could build the (rather small) 'traversal' graph only including roads that are accessible for cars. This is a bit contrary to making the routing more 'flexible' (no access flags), but I do not yet see any disadvantage of this separation and at least things like this would be possible. Not sure just some thoughts, maybe you can take this into account when designing this new storage.

karussell:
> So we could have a data structure that allows fast graph traversal and only gives us edge ids/keys (basically the nodeA/B and nextA/B columns from the edge storage) and then a separate storage for 'edge data'.

Interestingly the new storage could also allow that. Let me sketch the storage roughly:

 * nodeRefs: a fixed length list of "pointers" into the edge area
 * edgeRefs: a fixed length list of "pointers" into the edge area
 * edge area: 
```
[...]

PR #2243, opened 2021-02-22
Store encoded String values

otbutz:
Stores the encoded String values as part of the toString representation of the EncodedValue in the StorableProperties.

As discussed in https://github.com/graphhopper/graphhopper/pull/2204#issuecomment-746513867

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ev/StringEncodedValue.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/StringEncodedValue.java
@@ -8,6 +8,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.SortedMap;
+import java.util.TreeMap;
 
 /**
  * This class holds a List of up to {@link #maxValues} encountered Strings and stores
@@ -17,6 +19,14 @@ import java.util.Objects;
  * @author Thomas Butz
  */
 public final class StringEncodedValue extends UnsignedIntEncodedValue {
+    private static final String LIST_SEPARATOR = ";";
+    private static final List<String> ILLEGAL_VALUE_CHARS;
+    static {
+        List<String> illegalChars = new ArrayList<>(ILLEGAL_PROPERTY_CHARS);
+        illegalChars.add(LIST_SEPARATOR);
+        ILLEGAL_VALUE_CHARS = Collections.unmodifiableList(illegalChars);
+    }
+    
     private final int maxValues;
     private final List<String> values;
     private final ObjectIntMap<String> indexMap;
@@ -45,7 +55,7 @@ public final class StringEncodedValue extends UnsignedIntEncodedValue {
         this.indexMap = new ObjectIntHashMap<>(values.size());
         int index = 1;
         for (String value : values) {
-            indexMap.put(value, index++);
+            indexMap.put(validate(value), index++);
         }
     }
 
@@ -59,7 +69,7 @@ public final class StringEncodedValue extends UnsignedIntEncodedValue {
             if (values.size() == maxValues)
                 throw new IllegalStateException("Maximum number of values reached for " + getName() + ": " + maxValues);
 
-            values.add(value);
+            values.add(validate(value));
             index = values.size();
             indexMap.put(value, index);
         }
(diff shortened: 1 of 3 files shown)


otbutz:
@karussell i'm afraid we have to drop the specialized `StringEncodedValue.getVersion()` implementation and revert back to the default `getVersion()` of `UnsignedIntEncodedValue`. Currently testing this approach for #2201 and my problem is that we can't compare the EncodedValue created from the configuration with the one saved in the StoreableProperties as the first one lacks the List of String values which causes a version mismatch.

Another problem is the simple String comparison in GraphHopperStorgage.loadExisting(). We would need to skip additional properties here:

https://github.com/graphhopper/graphhopper/blob/aa069d6d8ac7bb3744f4fda27095fa81eeace761/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java#L185-L196

I'm also a bit confused as we seem to handle EncodingManager creation from existing graphs differently for GraphHopper and GraphHopperManaged:

GraphHopper(non-managed): importOrLoad() -> load() -> EncodingManager.create(EncodedValueFactory, FlagEncoderFactory, String)

GraphHopperManaged: GraphHopper.init() -> GraphHopper.createEncodingManager()

We can't offer any safety checks for the non-managed usecase as we have no configuration to compare to but we should probably try to unify code paths here.

karussell:
> implementation and revert back to the default getVersion() of UnsignedIntEncodedValue

Can you describe why this is not possible?

What I also would find interesting is if we could say: hey please store the values, if they are in the following list, and if not store as "other" and if missing store as "missing". 

Now another question for me is if we should we make strings behave like enums in custom models? Then we could write 
`something == MISSING` in a condition and convert this internally to a string comparison with "missing" (we already do some index magic for strings) and if later the encoded value changes into an enum (or an enum into a string) the custom model is still the same and users won't notice. 

Obviously this won't work for tags where case is important like `name` but our StringEncodedValues are not appropriate for that due to the many values anyway. (for that the [StringIndex](https://github.com/graphhopper/graphhopper/issues/2085#issuecomment-752080073) seems better)

otbutz:
> > implementation and revert back to the default getVersion() of UnsignedIntEncodedValue
> 
> Can you describe why this is not possible?

Currently we populate the `values` only while initially creating the graph. If we load an existing graph we don't have access to them anymore, so the StringEncodedValue is not really usable at the moment. If we add this persistence functionality we shouldn't break the detection of configuration mismatches which is handled here:

https://github.com/graphhopper/graphhopper/blob/aa069d6d8ac7bb3744f4fda27095fa81eeace761/core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java#L241-L247

Basically we create an EncodingManager from our configuration(read: no knowledge about values) and compare it with the String representation of the EncodingManager used while importing. If we use the approach proposed in this PR we end up with a mismatch:

Graph:
```
country_area|version=-1739096844|bits=6|index=0|shift=9|store_both_directions=false|values=deu;aut;cze
```

Config:
```
country_area|version=-1837704844|bits=6|index=0|shift=9|store_both_directions=false|values=
```

One solution could be to move from String to Object comparison(e.g `configEncodingManager.equals(graphEncodingManager)`). But this would still fail without reverting the changes to getVersion() and equals() of the StringEncodedValue as we can only reliably compare their basic properties like `bits`.

> 
> What I also would find interesting is if we could say: hey please store the values, if they are in the following list, and if not store as "other" and if missing store as "missing".

Maybe we could add an additional filter? This wouldn't break the usecase of custom areas where we need to populate this List of the fly, as we could pass a filter which accepts all values.

> 
> Now another question for me is if we should we make strings behave like enums in custom models? Then we could write
> `something == MISSING` in a condition and convert this internally to a string comparison with "missing" (we already do some index magic for strings) and if later the encoded value changes into an enum (or an enum into a string) the custom model is still the same and users won't notice.

A missing value should be represented as `null` which is already mapped to `0`:

https://github.com/graphhopper/graphhopper/blob/2b47a7a5b6488a0934c6d8c5396747da12509e8d/core/src/main/java/com/graphhopper/routing/ev/StringEncodedValue.java#L52-L67

I think we could avoid String comparisons for missing values if we check for null instead.

> 
> Obviously this won't work for tags where case is important like `name` but our StringEncodedValues are not appropriate for that due to the many values anyway. (for that the [StringIndex](https://github.com/graphhopper/graphhopper/issues/2085#issuecomment-752080073) seems better)

The StringIndex looks like it's more geared towards being used a key-value store. I'm missing the ability to store multiple values for a single key. Something like this:

```java
stringIndex.add(Collections.singletonMap(ev.getName(), ev.getValues()));
```

PR #2245, opened 2021-03-04
Incorrectly classified oneways

boldtrn:
A [Kurviger](https://kurviger.de/) user just reported an interesting bug. A [road](https://www.openstreetmap.org/way/115560610) was classified as oneway from the routing engine even though it is not marked as a oneway.

The issue is created in the isOneway method, where we only check if a `*:forward ` or `*:backward` tag exists but not if this tag actually marks it as a oneway.

Therefore I propose to check if these tags are actually indicating that a restrictive value is set.

This situation doesn't seem to happen frequently, but there are a few ways like this.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
@@ -277,8 +277,8 @@ public class CarFlagEncoder extends AbstractFlagEncoder {
      */
     protected boolean isBackwardOneway(ReaderWay way) {
         return way.hasTag("oneway", "-1")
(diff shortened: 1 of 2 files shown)


boldtrn:
I just saw that there is a similar check for bicycles as well, if you agree with these changes I could add the same for bicycle as well?

karussell:
Thanks!

karussell:
btw: interesting that I coincidentally stumbled over this when I played with custom models, where I needed to remove `no` from restrictedValues (to implement an "allowing" vehicle that can be further restricted with the custom model). And I did not need to debug this and just thought: oh, nice - this is the problem :)

PR #2248, opened 2021-03-08
Rename road_access to car_access

karussell:
In the branch [node_parser_extraction](https://github.com/graphhopper/graphhopper/tree/node_parser_extraction) I learned we really need a more precise name here. E.g. we later introduce an access encoded value for every TransportationMode.


karussell:
Hmmh, no. We need to first clarify how to handle all the accesses (for the different TransportationMode car+bike+motorcycle+wheelchair) and maybe even store it depending on the direction. Then it could even mean that car_access is just a boolean used in combination with motor_vehicle_access (string or enum) or similar. Cannot be decided yet but a rename is not a safe solution at the moment.

PR #2249, opened 2021-03-08
rename in_area_custom1 to in_custom1

karussell:
rename to `in_` as it might be confusing when `area` is part of the area name. See e.g. https://github.com/graphhopper/graphhopper/pull/2201#discussion_r578242801


easbar (approved):
LGTM thanks!

PR #2251, opened 2021-03-11
Use json instead of yaml in custom model text box

easbar:
Here I changed the custom model text box so it only accepts json instead of yaml in GH maps. The reasons to use json instead of yaml were:

* consistency with GH api (only accepts json) -> less potential for confusion about the syntax for (client-side) users
* consistency in docs (no need to describe both yaml and json in custom model documentation)
* json is less error-prone, because indentation is not significant. yaml can be especially annoying when using copy&paste and the indentation is broken. also indentation issues might be not intuitive for users not familiar with yaml
* Using json is more restrictive. Allowing yaml later is possible while going back from yaml to json would break previous examples, documentation, blog articles etc.
* yaml is incredibly complex and allows a lot more things than we need. This complexity means yaml parsers are harder to write, need more code etc.

Of course there were multiple reasons why we used yaml so far:

* yaml allows commenting out parts of the custom model which can be useful to experiment with a model. For GH maps we could allow this even with this PR, because the underlying json parser has an option to allow javascript style comments. Comments could also be useful to provide some inline documentation right in the custom model box, but we can probably achieve the same with some popup windows.

* yaml looks nice / clean, because in most cases no brackets are needed and strings do not need quotation. However, relying on indentation to define the structure also can make it hard to find syntax errors.

* yaml is used for GH server configuration. This also means it is possible to write a custom model on the server side using comments etc., but then we cannot simply paste it in GH maps. The workaround would be converting it to json first (losing all the comments). Or maybe we should write custom models in json on the server side as well and keep them in separate files (just like area files, also written in (geo)json). 

Currently it seems the advantages of json outweigh those of yaml, but we could still allow yaml later. Comments could be realized by adding a (or multiple) 'description' field(s) to the custom model if this is really something we need.

The new version of the custom model text box uses another parser (https://github.com/microsoft/node-jsonc-parser). First I was hoping we can run the yaml parser we were using in 'json mode', but this did not work very well as it still allows yaml-like lists, single-quoted strings etc. Also it seems unnecessary to use the bigger yaml parser when all we need is json.

Please feel free to try and test the new editor. It supports all features of the yaml version in master, but probably there are many things that can still be improved :) 

Todo:

- [ ] Update documentation so client-side users to not get in touch with yaml stuff that is only needed for server configuration
- [ ] Update [demo server and blog post](https://www.graphhopper.com/blog/2020/05/31/examples-for-customizable-routing/) @karussell


karussell:
Yeah, this dual format was very confusing to understand but also to explain :) And so I highly prefer - thanks!

The only big 2 problems I have with JSON is its bigger verbosity (only a minor problem due to the nice autocompletion :+1: and especially 0 problem as an API) and the lack of comments.

btw: should we remove the statement in the example JSON with `road_class == RESIDENTIAL` so that `max_weight < 3` shows up on smaller screens too? (the second road_class statement would not show much new features IMO)

> Comments could also be useful to provide some inline documentation right in the custom model box, but we can probably achieve the same with some popup windows

Comments would be indeed helpful especially for server-side files. Should we just ignore `comment` entries in all statements?

```json
{
 "comment": "bla bla bla bla",
 "//": "or a very short notation for a comment ;) ?",
 "if": "road_class == MOTORWAY",
 "multiply_by": 0.8
}
```

> Or maybe we should write custom models in json on the server side as well and keep them in separate files

yes, especially if comments are allowed :) (and we could even drop yaml support completely)

> has an option to allow javascript style comments.

How would that look like? Is it the same style like jackson could allow?

> objectMapper.configure(JsonParser.Feature.ALLOW_COMMENTS, true); // or ALLOW_YAML_COMMENTS

```
{
 // C-style comments
 /* even two line comments
  */
 # YAML comments
 "if": "road_class == MOTORWAY",
 "multiply_by": 0.8
}
```

Of course we should allow this only for server-side profiles.

easbar:
> btw: should we remove the statement in the example JSON with road_class == RESIDENTIAL so that max_weight < 3 shows up on smaller screens too? (the second road_class statement would not show much new features IMO)

Sure we can do that. I was also wondering if we should enable soft line wrapping so the entire text is always visible. Especially since we show line numbers and indentation is no longer significant this could work. Or we go the other way and stop showing the line numbers so there is a bit more space. But should we also remove the red error icons in the left gutter then? Errors would be still visible by the red squiggly underlines.

> Comments would be indeed helpful especially for server-side files. Should we just ignore comment entries in all statements?

I think even in the maps UI commenting out something can be useful to experiment without deleting a statement for example. Adding a comment/description/"//" field in the statements could indeed be useful yes. And maybe a global description field as well?

> yes, especially if comments are allowed :) (and we could even drop yaml support completely)

For config.yml I think yaml is still very useful (because of the comments). But we could also move the comments in `config-example.yml` somewhere else. Then again whenever I have to use json to write some config file I find it very annoying to not be able to comment... Or we just keep config.yml like it is but move custom models into separate files (no longer allow inline custom models) and add the `custom_model_folder` instead.

> How would that look like? Is it the same style like jackson could allow?

I think the jsonc parser simply treats all lines starting with `//` and all text sections wrapped in `/* */` as comments.

> Of course we should allow this only for server-side profiles not for the endpoint as I'm unsure if this could introduce security problems if certain "unsafe" statements could be injected via "safe" comments (I don't think it is possible but still :) )

Yes allowing comments for Jackson would be a nice way to allow comments in server-side config files, but they should not be allowed for the http api. Using dedicated json fields for comments seems like the most portable/consistent approach, but unfortunately does not allow 'commenting out' a statement for example. Not sure if this is a real problem though. And maybe we do not really need comments at all and we could start without them and think about allowing comments later... :)

easbar:
Shall we merge here and do further improvements in future commits/PRs?

karussell:
Ok, yes - let's merge :)

> I was also wondering if we should enable soft line wrapping so the entire text is always visible.

I do not like softline wrapping as I loose the "structural overview". Maybe we use the fullscreen feature from CodeMirror :) ?

> Or we go the other way and stop showing the line numbers so there is a bit more space. 

Yes, line numbers seem currently not that important and the red underlines are enough (?) Or are they not obvious enough?

> but they should not be allowed for the http api.

Yes, of course :) ! For server-side models and the editor this seems like a good solution the more I think about it, but as you said later would be also fine.

easbar:
Ok thanks! So many decisions haha :) So we do not use softline wrappings but remove the line numbers and error icons instead :)

PR #2252, opened 2021-03-15
Move /route-custom into /route endpoint

easbar:
Here I moved the functionality of /route-custom into the route endpoint. The custom model now lives under a dedicated node called `custom_model`.


easbar:
I'm still wondering what the new API is supposed to be like. Currently with this PR a custom (POST) routing request looks like this: 

```json
{
 "profile": "some_custom_profile",
 "custom_model": { ... }
}
```
i.e. we have to pick a certain profile (that uses `weighting: custom`) *and* set the custom model (even though it might be just empty `{}`). For a 'classic' profile the request only includes the `profile` parameter and no custom model. Using a 'classic' profile *and* setting a custom model yields an error. Using a custom profile and not setting a custom model (instead of an empty model `{}` also yields an error). Can we finalize on this API? API users need to know which profiles are 'custom' and which are not. Since we probably always need to distinguish between 'customizable' profiles and 'fixed' profiles this seems reasonable?! However, it can also be a bit more complicated than this because sending a custom model is only possible for custom profiles that allow non-CH requests.

easbar:
>  Using a custom profile and not setting a custom model (instead of an empty model {}) also yields an error. 

This shouldn't be. Using a custom profile without a custom model makes sense and should be allowed.

easbar:
Update: The behavior is now like this:

* request uses custom profile and no custom model (or null) -> just use the custom profile as it was configured on the server-side
* request uses custom profile and custom model -> the query custom model will be merged with the server custom model
* BUT this does not work with CH, so we need to to add ch.disable=true (this is consistent to e.g. block_area)
* custom_model={} is treated like any other custom model *not* like custom_model=null or a missing custom_model field
* the maps UI adds the ch.disable=true option for custom requests via the custom model box
* for the java API we now do `request.setCustomModel(customModel) instead of `request.putHint(CustomModel.KEY, customModel)

karussell (approved):
Nice, thanks!

PR #2253, opened 2021-03-15
Do not close geometry if CH + custom profile

karussell:


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1026,7 +1026,8 @@ protected void prepareCH(boolean closeEarly) {
 
             if (closeEarly) {
                 locationIndex.close();
-                ghStorage.flushAndCloseEarly();
+                boolean includeWayGeometry = getProfiles().stream().noneMatch(p -> p instanceof CustomProfile);
+                ghStorage.flushAndCloseEarly(includeWayGeometry);
```
karussell, line 1030: Not sure if it is worth the effort without wayGeometry or if we should exclude everything in case of a CustomProfile. Or maybe we should explicitly check for areas?

PR #2254, opened 2021-03-15
Do not close geometry if CH + custom profile

karussell:


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1026,7 +1026,8 @@ protected void prepareCH(boolean closeEarly) {
 
             if (closeEarly) {
                 locationIndex.close();
-                ghStorage.flushAndCloseEarly();
+                boolean includeWayGeometry = getProfiles().stream().noneMatch(p -> p instanceof CustomProfile);
+                ghStorage.flushAndCloseEarly(includeWayGeometry);
```
karussell, line 1030: Not sure if it is worth the effort without wayGeometry or if we should exclude everything in case of a CustomProfile. Or maybe we should explicitly check for areas?

easbar: Is it really necessary to be so specific here? Is it critical to flush the string index when we keep the way geometry anyway? Can we not even use the string index in a custom profile so we have to keep it anyway? Maybe we should not call `ghStorage.flushAndCloseEarly` whenever there is a CustomProfile?

karussell: Yes, you are right. Especially in later versions where we store more than the name into the StringIndex.

easbar: I just removed the flag from flushAndCloseEarly

easbar (approved):

easbar:
Just for later reference: Flushing/closing the base graph's geometry or StringIndex is not possible before we build the CH preparation graph in case the weighting depends on them. Custom weightings can depend on the geometry via `area` definitions and potentially they can depend on the StringIndex as well.

karussell:
Nice - thanks!

PR #2256, opened 2021-03-16
Make LM preparation working with custom profiles

karussell:
The LM preparation will get an extreme slow down for custom profiles that exclude e.g. motorways (occurs only for world wide instances). Related to #1835.

Usually we just need to pick one node of the returned subnetwork in LM preparation and use that to calculate the coordinates of the landmarks. But likely because of some old code that tried to make this more robust, we loop over all nodes of this subnetwork, to try again with nodes that couldn't be explored.

Interestingly even with this loop it is fast as we exit the loop as soon as a largish subnetwork was found. But now with the CustomWeighting the Tarjan algorithm incorrectly returns a large subnetwork that is splitted by a oneway motorway (Tarjan only sees access not the weight) and then the initial search takes 30 sec (in the large subnetwork) and then due to the next landmarks searches it gets stuck in the small subnetwork (due to the oneway motorway and the filter just uses access flags) and so it returns too few nodes for the last explorer (explorer.getFromCount) so that the next node of the large subnetwork is tried where the same happens again and again likely until all nodes of the large subnetwork were tried. Somewhere [here](https://graphhopper.com/maps/?point=22.365107%2C114.180932&point=22.3804%2C114.203421).

This PR should fix this problem (not yet tried for world wide instances).

Performance seems to be fine.


core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -897,57 +893,31 @@ static GHPoint createPoint(Graph graph, int nodeId) {
...
         @Override
-        public final boolean accept(EdgeIteratorState iter) {
-            boolean blocked = blockedEdges.contains(iter.getEdge());
+        public final boolean accept(EdgeIteratorState edgeState) {
+            boolean blocked = blockedEdges.contains(edgeState.getEdge());
             if (blocked)
                 return false;
-            return fwd && iter.get(accessEnc) || bwd && iter.getReverse(accessEnc);
+            return fwd && Double.isFinite(weighting.calcEdgeWeightWithAccess(edgeState, false)) ||
+                    bwd && Double.isFinite(weighting.calcEdgeWeightWithAccess(edgeState, true));
```
easbar, line 915: Can we use FiniteWeightFilter here?

easbar:
Very nice you figured this out! Thanks :)

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
karussell: I currently do not see how. It would only be equal if `fwd` and `bwd` are `true`.

easbar: Ah yes sorrry nevermind. I just saw this pattern and thought its the same...

karussell:
It still does not work in rare cases. The reason is that the custom profile can create islands. E.g. "exclude motorway" can disconnect otherwise accessible roads from the main network. To make Tarjan algo aware of this and handle islands that are partially connect we need to replace FiniteWeighting with a directed EdgeFilter:

```
EdgeFilter accessFilter = edgeState -> edgeState.get(encoder.getAccessEnc()) && Double.isFinite(weighting.calcEdgeWeight(edgeState, false));
```

Otherwise the algo won't split e.g. the graph 1-2-3->4-5 into two networks 1-2-3 and 4-5. Which can lead to a problem in the next step where LandmarkExplorer selects landmarks and picks a random start node from a network like 1-2-3 but ends with e.g. 4 as next node and because 4-5 is a network from which 1-2-3 is not reachable and which is also too small it will be skipped and a new start node e.g. 2 is tried and and could fail again and again. Which is the problem here as for the planet-wide import the main network "1-2-3" consists of millions of nodes and so exploration takes half a minute for every node leading to a too long waiting time.

Unfortunately this is not enough to fix the slow down for world wide LM preparation.

The second problem is that LandmarkExplorer is not aware of the created islands (from the customization) and a graph exploration there can cross "island boundaries" leading to the same problem. To fix that we mark the main road network and force the exploration (in LandmarkExplorer) to stay there.

Unfortunately this causes the routing requests to return incorrect approximations. I think there is a bug which was mostly covered until now. Will need to investigate this further.

Likely these problems won't occur when #2273 is done.

easbar:
> To make Tarjan algo aware of this and handle islands that are partially connect we need to replace FiniteWeighting with a directed EdgeFilter:

Yes, currently in master we use : `DefaultEdgeFilter.outEdges(accessEnc)`. Using `FiniteWeightFilter` would indeed me wrong because it accepts an edge even when only the reverse direction has finite weight.

karussell:
This finally goes through also for a planet-wide import and I was able to remove a bit more code :)

The minor question is now if we should require the subnetwork EncodedValue and adapt the tests. Or if we should keep it like it currently is (only include subnetwork EV if it exists).

For that let me explain why the subnetwork EncodedValue is used. Because it is not required as we do a Tarjan in LandmarkStorage again. And we could do this TarjanSCC also edge-based and then use the created subnetworks instead of relying on the work of PrepareRoutingSubnetwork (which could even use a different minimum network size). 
I tried this but as everything regarding landmarks is still node-based this opened up a rabbit hole (required to change the stored LM weights to edge-based and maybe even the LM query to be clean). IMO this is a separate issue and it is not clear if this would improve something as memory usage could increase (higher edge key than nodes?). So I decided to keep using the node-based Tarjan but use the subnetwork EncodedValue from PrepareRoutingSubnetwork that is also correct regarding turn restrictions. So basically what we did before but now cleaner and based on the Profile not FlagEncoder only.

core/src/test/java/com/graphhopper/routing/RandomizedRoutingTest.java
```diff
@@ -132,15 +133,15 @@ public void init() {
...
         graph = new GraphBuilder(encodingManager)
                 .setCHConfigStrings("p1|car|fastest|node", "p2|car|fastest|edge")
                 .setDir(dir)
                 .create();
         turnCostStorage = graph.getTurnCostStorage();
         chConfigs = graph.getCHConfigs();
         // important: for LM preparation we need to use a weighting without turn costs #1960
-        lmConfig = new LMConfig("config", chConfigs.get(0).getWeighting());
+        lmConfig = new LMConfig("car", chConfigs.get(0).getWeighting());
         weighting = traversalMode.isEdgeBased() ? chConfigs.get(1).getWeighting() : chConfigs.get(0).getWeighting();
```
karussell, line 145: @easbar do you think it is fine when we force to add a subnetwork EncodedValue? See also below for the other tests that needed a change due to the introduced check in LandmarkStorage.

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -777,30 +777,30 @@ private LandmarkExplorer findLandmarks(int[] landmarkNodeIdsToReturn, int startN
      */
     private static class LandmarkExplorer extends DijkstraBidirectionRef {
         private EdgeFilter accessFilter;
-        // todo: rename 'from' to 'reverse' (and flip it) ? 'from' is used in many places for node ids and 'reverse' is mostly used for the direction
-        private final boolean from;
```
karussell: @easbar was able to remove your todo/question added in #1341 :)

core/src/test/java/com/graphhopper/routing/RandomizedRoutingTest.java
easbar: Ok it seems for now this is a good choice. And if this makes problems we can still change it later (use separate edge-based TarjanSCC for LM or so).

easbar (approved):

PR #2258, opened 2021-03-17
Keep elevation diagram selection between requests

easbar:
I added a new `selected_detail` url parameter to the maps UI. It stores the selected detail in the elevation diagram so we can keep the selection in between requests (e.g. when we shift the markers or update the custom model).

This PR depends on #2257

### DIFF
--- a/web-bundle/package.json
+++ b/web-bundle/package.json
@@ -30,7 +30,7 @@
     "leaflet": "1.5.1",
     "leaflet-contextmenu": "1.4.0",
     "leaflet-loading": "0.1.24",
-    "leaflet.heightgraph": "0.2.0",
+    "leaflet.heightgraph": "1.4.0",
     "leaflet.vectorgrid": "1.3.0",
     "moment": "2.22.1"
   },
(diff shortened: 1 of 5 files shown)


easbar:
@karussell WDYT is `selected_detail` too clunky for an url parameter? Or maybe we should keep the state without storing it in the url? I just did this the same way we use `use_miles` or `layer`. Otherwise I think this is quite useful.

karussell:
Sounds useful, yes.

> is selected_detail too clunky for an url parameter?

And what if we "store" this info in the order of details parameter and pick the first?

easbar:
> And what if we "store" this info in the order of details parameter and pick the first?

Yes this would be nice, but not really sure if this is easy to do. In principle you are right and the `details` parameters should be in sync with the elevation diagram (and even the order could be significant). But currently this is not what we do, i.e. the details parameters can only be entered via the url and the elevation diagram is purely based on the routing response. So the url affects the request, but not how the response is displayed in the UI (including the diagram). Furthermore the the /route endpoint does not maintain the order for path details, so the order in the url is not the same as in the response. With the custom request it is even more inconsistent, because in this case we take the details from the custom model conditions and the ones from the url will be ignored. 

But maybe first things first. I just noticed something really strange: Before this commit: https://github.com/graphhopper/graphhopper/pull/2258/commits/6514fb71f40c5599511d1a9849301bb5d44b9209 which only added a console.log the feature did not work (I probably removed a similar console.log just before I pushed this PR). What I did was: 

1) start the server
2) click to markers to calculate the route
3) add `details=time&details=road_class` to the url
4) choose time in the elevation diagram
5) move one of the markers

With the console log the diagram keeps showing time, but without the console.log it shows road_class? Can you confirm this? I got the same behavior in chrome and firefox.

easbar:
I could not find out what the problem is with this console.log and @karussell could not reproduce the issue, so I leave it as it is now. Keeping the `selected_detail` parameter as well.

PR #2259, opened 2021-03-17
Color gradient for numeric path details

easbar:
For numeric path details like `time` or `average_speed` its not very useful to use unrelated, random colors in the elevation diagram. Here I am using a linear color gradient for such numeric path details. The gradient uses two fixed colors to represent the minimum/maximum of the encoded value range of the current path (green to red). For discrete values I use a fixed 10-color palette now (leaflet.heightgraph was using a random choice of six color palettes, but to me it seems confusing to see different colors each time).

Not entirely sure about this change yet, any opinions @karussell, @boldtrn?

Here are screenshots of a numeric and a discrete path detail:

image

image

By the way I use the 10-color palette from d3, but maybe not really worth keeping the entire library when all we need is an array of 10 hex codes...

And thanks again for these changes in heightgraph, they are very useful @boldtrn!

This PR is based on #2258.

### DIFF
--- a/web-bundle/src/main/resources/com/graphhopper/maps/js/map.js
+++ b/web-bundle/src/main/resources/com/graphhopper/maps/js/map.js
@@ -1,6 +1,7 @@
 var mainTemplate = require('./main-template.js');
 var tileLayers = require('./config/tileLayers.js');
 var translate = require('./translate.js');
+var d3 = require('d3');
 
 var routingLayer;
 var map;
@@ -292,6 +293,7 @@ module.exports.addElevation = function (geoJsonFeature, details, selectedDetail,
         expandCallback: function (expand) {
             expandElevationDiagram = expand;
         },
+        mappings: {},
         selectedAttributeIdx: 0,
         chooseSelectionCallback: detailSelected
     };
@@ -304,7 +306,9 @@ module.exports.addElevation = function (geoJsonFeature, details, selectedDetail,
         detailIdx++;
         if (detailKey === selectedDetail)
             selectedDetailIdx = detailIdx;
-        GHFeatureCollection.push(sliceFeatureCollection(details[detailKey], detailKey, geoJsonFeature))
+        GHFeatureCollection.push(sliceFeatureCollection(details[detailKey], detailKey, geoJsonFeature));
+        options.mappings[detailKey] = getColorMapping(details[detailKey]);
+
     }
     if (selectedDetailIdx >= 0)
         options.selectedAttributeIdx = selectedDetailIdx;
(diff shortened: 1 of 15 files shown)


otbutz:
Can we avoid red/green color shades for colorblind users?

easbar:
Ah and @boldtrn suggested we should use a special color (like grey) for missing values? I think this is a great idea, but how can we reliably recognize missing details?

easbar:
> Can we avoid red/green color shades for colorblind users?

Good point! What colors would be better?

web-bundle/src/main/resources/com/graphhopper/maps/js/map.js
```diff
@@ -325,6 +338,62 @@ module.exports.addElevation = function (geoJsonFeature, details) {
     elevationControl.addData(GHFeatureCollection);
 };
 
+function getColorMapping(detail) {
+    var detailInfo = analyzeDetail(detail);
+    if (detailInfo.numeric === true && detailInfo.minVal !== detailInfo.maxVal) {
+        // for numeric details we use a color gradient
+        var colorMin = [47, 115, 54];
+        var colorMax = [170, 58, 56];
```
easbar, line 346: @otbutz The min/max colors are specified here

boldtrn:
> Ah and @boldtrn suggested we should use a special color (like grey) for missing values? I think this is a great idea, but how can we reliably recognize missing details?

Using the current GH API you can't. GH uses the country fallback speed. This would require a change in the MaxSpeedParser. 

> Good point! What colors would be better?

As far as I know, any other color should work, I think you can even use red or green, but should avoid both in the same visualization.

So for example we could use blue instead of green (but you are using blue already :)).

easbar:
> Using the current GH API you can't. GH uses the country fallback speed. This would require a change in the MaxSpeedParser.

Ah ok. For some encoded values (like road environment) we use 'other', but e.g. for surface we have 'missing'...

>  any other color should work,

How about this:

image

I found this here: https://uigradients.com/#Superman

otbutz:
> > Can we avoid red/green color shades for colorblind users?
> 
> Good point! What colors would be better?

https://thenode.biologists.com/data-visualization-with-flying-colors/research/

otbutz:
Funnily enough, this was posted on hackernews just recently: https://blog.datawrapper.de/which-color-scale-to-use-in-data-vis/

easbar:
Thanks! I like it :) And it even says "Pale grey is meant for bad data in maps.", so a perfect fit for us ;)

I now hard-coded 'missing' and 'unclassified' to use grey. We can add more values to the 'value missing' list of course.

image

PR #2260, opened 2021-03-18
Update maps hot reload setup

easbar:
* live-server is now in dev dependencies so we do not need to install it separately
* the live-server path in the docs was broken and I fixed it
* updated the docs

Essentially we can develop maps by running the GH server, `npm run watch` and `npm run liveServer` in a separate terminal. However, we still need to set the GH server port in options.js. I was hoping to get rid of this using browserify-swap (see my first commit), but this is not working reliably (at least on my machine). After a few updates the wrong options.js is pulled in and we have to restart the terminal with `npm run watch`. Without browserify-swap the UI is also sometimes in a broken state (probably because live-server loads main.js before it was finished building), but this can easily be fixed with a browser refresh.

Related: https://github.com/graphhopper/graphhopper/pull/2238


easbar:
Maybe these options could be useful for live-server:

* --wait=MILLISECONDS - (default 100ms) wait for all changes, before reloading
* --no-browser - suppress automatic web browser launching
* --browser=BROWSER - specify browser to use instead of system default

and actually if we do not want the automatic browser reload we could use a simple server and hit refresh ourselves. The important part is just that we serve GH maps from the file system where `main.js` is updated.

docs/core/quickstart-from-source.md
```diff
@@ -96,29 +96,26 @@ as those versions are not in maven central:
...
+   use [nvm-windows](https://github.com/coreybutler/nvm-windows).
+2. Build the custom-model-editor
+   component: `sh -c "cd web-bundle/src/main/js/custom-model-editor && npm install && npm run build"`
+3. Build the Web UI: `cd web-bundle && npm install && npm run bundle` which results in the `main.js` file
+4. Restart the GH server so it picks up the latest version of the UI bundle
+
+You can achieve an even faster development cycle by running `npm run watch` which will update `main.js` whenever you
+make changes to one of the .js files. The same command works in the custom-model-editor folder in case you want to make
+changes there. To hot-reload your changes in the browser the best option is to serve GH maps from a separate server like
+live-server. You can do this by running `npm run liveServer` from a separate terminal, but you need to adjust the
```
karussell, line 116: Maybe `live-server` or `liveserver` or just `server`?

docs/core/quickstart-from-source.md
```diff
@@ -96,29 +96,26 @@ as those versions are not in maven central:
...
+routing host in options.js and point it to your GH server. It is also a good idea to remove `main.js`
+from the IntelliJ index. Right-click the file and choose `Mark as plain text`.
```
karussell, line 118: This is not 100% clear to me. I'll have to try it :)

easbar: Yes, please feel free to try and adjust

docs/core/quickstart-from-source.md
easbar: Ok `serve` is used by angular for example

docs/core/quickstart-from-source.md
karussell: > It is also a good idea to remove `main.js` from the IntelliJ index

This is only necessary when some auto magic is enabled (?) I would prefer to omit this and formulate this part editor agnostic. Otherwise it could sound that this only only possible with IntelliJ.

> but you need to adjust the routing host in options.js and point it to your GH server

What do you think about creating a third `serve` or `localrouting` mode (additionally to `development` and `production`), then this extra step is not necessary. To make this working we only need to create a file `options_localrouting.js` with the correct routing host and a new entry:

```
   ... ,
   "localrouting": {
      "config/options.js$": "./src/main/resources/com/graphhopper/maps/js/config/options_localrouting.js"
   }
```

in package.json and then `BROWSERIFYSWAP_ENV='localrouting' npm run watch` will automatically point to the local routing host.

karussell: I.e. the doc could be reduced to:

To hot-reload your changes in the browser you run 
`BROWSERIFYSWAP_ENV='localrouting' npm run watch` 
and from a separate terminal run 
`npm run serve`.

easbar: > Is this only necessary when some auto magic is enabled (?) I would prefer to omit this and formulate this part editor agnostic. Otherwise it sounds like if this only possible with IntelliJ.

when we run `npm run watch` the `main.js` file is rebuilt and so IntelliJ keeps indexing the file all the time which takes lots of cpu power, battery and fan speed goes up :)

> then this extra step is not necessary.

like I said this is exactly what I did but it is not working very well so I removed it again (see my initial comment)

karussell: > when we run npm run watch the main.js file is rebuilt and so IntelliJ keeps indexing the file all the time which takes lots of cpu power, battery and fan speed goes up :)

Ok

> like I said this is exactly what I did 

Oh, ok. Then I misunderstood your approach and thought you set it somehow directly in options.js. For me using a separated file worked reliable (but likely you tried longer :) )

Now, when re-reading your description - what did you mean with:

> Without browserify-swap the UI is also sometimes in a broken state 

How do you pick the different routing host without browserify-swap?

> (probably because live-server loads main.js before it was finished building)

Oh, ok. Then the combination of live-server & watchify & browserify-swap is probably the issue. How are other projects doing this? Seems like a simple required to inject some command line parameter into JS.

easbar: > For me using a separated file worked reliable (but likely you tried longer :)

Did you try with my first commit? I got the problem very consistently. I just changed the custom model box example text in `main-template.js` a few times and then the wrong options file was used so I got this error:

image

But maybe something about my browserify-swap setup wasn't correct, maybe you can compare it to yours?

> what did you mean with:

I got all kind of errors (syntax errors etc. just like you would get if you loaded only half of main.js). Probably because live-server loaded main.js before it was finished building. The situation can probably be improved with the wait option: https://github.com/graphhopper/graphhopper/pull/2260#issuecomment-801999719 but I did not try this. Also its not really an issue we just have to refresh the browser window.

> How do you pick the different routing host without browserify-swap? 

I just comment out one and comment in the other line in options.js

> Oh, ok. Then the combination of live-server & watchify & browserify-swap is probably the issue. How are other projects doing this? Seems like a simple requirement to inject some command line parameter into JS.

It is not related to live-server, I got the same problem when I served the file without live-server. It's probably some problem when we combine watchify with browserify-swap.

> How are other projects doing this? Seems like a simple requirement to inject some command line parameter into JS.

I don't know. For live-reloading webpack-dev-server is quite popular, but I don't know how it is done commonly with browserify. browserify-swap sounds like the right tool for the job but I could not make it work properly.

karussell: I just tried it with your first commit (that looks identical to what I tried) and it worked for me (changing the example JSON in main-template.js). Can you try again without IntelliJ? Or maybe you need to increase the watchify `delay` or `poll` or live-server its `wait`?

> It's probably some problem when we combine watchify with browserify-swap.

Ah, ok

> browserify-swap sounds like the right tool for the job but I could not make it work properly.

It looks very outdated (6 years) and as this is a common requirement I can imagine that this is done differently. But no idea :)
Maybe a simple sed command that runs before watchify in a shell pipe and replaces options.js with options_xy.js works better ;)

easbar: > Maybe a simple sed command that runs before watchify in a shell pipe and replaces options.js with options_xy.js works better ;)

Yes that would work probably but not sure if if works on different OS then. If browserify-swap works for you we can just use it and either it works for everyone except me, then I will just change it locally or we see it makes problems and revert it.

easbar: I reverted to use browserify-swap again. But it does not work for me. After three changes to the example json the wrong options file is used.

easbar: I also tried again. Removed node_modules, main.js, ran mvn clean, closed intellij, started GH server, npm run watch and npm run server from the command line and changed main-template.js via another text editor. After a few (like 3-5) changes the wrong options file is used and the error I posted above is shown (because the routing host is taken from the wrong options file)

easbar: I am pretty sure something about browserify-swap is broken. I now ran into a case where the source mapping broke onto a new (non-comment) line:

image

I will just add another script so we can use the watch command with or without browserify-swap (I won't use it, I rather comment out the line in options.js)

karussell: Oh sorry, didn't want to waste your time with this. Maybe my setup is broken or I'm doing something wrong. Although I also tried from scratch git clone and with a different editor and only with FF without a problem. Maybe timing is important and my laptop is just too slow to hit the race condition :D So yes, we should use the (default) setup that works for you too.

easbar: I found another way to change the configuration: We can just use another entry file for the watchify command: e58c34e6991b38da7780ca8ca197d21829ff3f90

Do you also like this better? If so I will remove browserify-swap and go this way instead.

easbar: > Maybe timing is important and my CPU or SSD is just too slow to hit the race condition :D

Don't worry. Maybe something weird is going on my computer as well, it is a bit like this console.log thing earlier today :) Still I think using different entry files (my previous comment) seems like a good solution and we do not have to rely on this browserify swap thing.

easbar: ... except this does not work either. Even with this approach I got the wrong routing host. Ok, I am giving up, I will just set the routing host in options.js manually, it's easy enough.

web-bundle/src/main/resources/com/graphhopper/maps/js/config/options.js
```diff
@@ -11,7 +11,7 @@
 exports.options = {
     with_tiles: true,
     environment: "development",
-    // for local development use this (assuming your GH server is running on the standard port)
+    // if you server GH maps from another server than the GH server use this (assuming your GH server is running on the standard port)
```
karussell, line 14: minor typo: server -> serve ?

easbar: Ok I wrote: 'Use this if you *serve* the web UI from ...'

docs/core/quickstart-from-source.md
easbar: All we have to do is comment in one line in options.js manually as described in the docs. I think this is easy enough and it seems like a waste of time to try and automate this. Especially with GH maps 2.0 and its new toolchain on the horizon :) Merging here for now.

PR #2268, opened 2021-03-29
Use MaxSpeed EV in the AbstractFlagEncoder

### LINKED ISSUE #2266: Use max_speed EV for the AbstractFlagEncoder
I'm currently testing to switch the AbstractFlagEncoder to use the max_speed encoded value and have already a working prototype. The remaining problem seems to be related to how we handle spatial rules and fallback values.

Let's take the following route which is part of our unittests as an example:

grafik

https://graphhopper.com/maps/?point=49.995905%2C11.547918&point=50.004871%2C11.517191vehicle=car&weighting=fastest&elevation=true

The route changes if the encoded max_speed values in conjunction with spatial rules are used:

grafik

The detour seems to be caused by the high fallback values of the spatial rule which assumes 100kmh for unclassified/residential. This is higher than the tagged speedlimit of 70kmh for the B85 and causes the routing to bypass it.

I can lower the limits in the GermanySpatialRule so that they fit the `defaultSpeedMap` values of the CarFlagEncoder but i'm not sure if that's the best course of action.

otbutz:
Fixes #2266 

* avoids parsing the max speed tags multiple times by using the encoded value
* respects SpatialRules as they are used in the OSMMaxSpeedParser
* lowered some max speed fallback values for Germany and Austria to match the limits of urban areas
* adapted unittests

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
@@ -24,7 +24,6 @@ import com.graphhopper.reader.osm.conditional.ConditionalOSMTagInspector;
 import com.graphhopper.reader.osm.conditional.DateRangeParser;
 import com.graphhopper.routing.ev.*;
 import com.graphhopper.routing.util.parsers.OSMRoadAccessParser;
-import com.graphhopper.routing.util.parsers.helpers.OSMValueExtractor;
 import com.graphhopper.storage.IntsRef;
 import com.graphhopper.util.EdgeIteratorState;
 
@@ -55,6 +54,7 @@ public abstract class AbstractFlagEncoder implements FlagEncoder {
     private long encoderBit;
     protected BooleanEncodedValue accessEnc;
     protected BooleanEncodedValue roundaboutEnc;
+    protected DecimalEncodedValue maxSpeedEnc;
     protected DecimalEncodedValue avgSpeedEnc;
     // This value determines the maximal possible speed of any road regardless of the maxspeed value
     // lower values allow more compact representation of the routing graph
@@ -142,6 +142,7 @@ public abstract class AbstractFlagEncoder implements FlagEncoder {
         // define the first 2 bits in flags for access
         registerNewEncodedValue.add(accessEnc = new SimpleBooleanEncodedValue(EncodingManager.getKey(prefix, "access"), true));
         roundaboutEnc = getBooleanEncodedValue(Roundabout.KEY);
+        maxSpeedEnc = getDecimalEncodedValue(MaxSpeed.KEY);
         encoderBit = 1L << index;
     }
 
@@ -209,16 +210,14 @@ public abstract class AbstractFlagEncoder implements FlagEncoder {
     /**
      * @return {@link Double#NaN} if no maxspeed found
      */
-    protected double getMaxSpeed(ReaderWay way) {
-        double maxSpeed = OSMValueExtractor.stringToKmh(way.getTag("maxspeed"));
-        double fwdSpeed = OSMValueExtractor.stringToKmh(way.getTag("maxspeed:forward"));
-        if (isValidSpeed(fwdSpeed) && (!isValidSpeed(maxSpeed) || fwdSpeed < maxSpeed))
-            maxSpeed = fwdSpeed;
-
-        double backSpeed = OSMValueExtractor.stringToKmh(way.getTag("maxspeed:backward"));
-        if (isValidSpeed(backSpeed) && (!isValidSpeed(maxSpeed) || backSpeed < maxSpeed))
-            maxSpeed = backSpeed;
-
+    protected double getMaxSpeed(IntsRef edgeFlags) {
+        double fwdSpeed = maxSpeedEnc.getDecimal(false, edgeFlags);
+        double bwdSpeed = maxSpeedEnc.getDecimal(true, edgeFlags);
+        
+        double maxSpeed = Math.min(fwdSpeed, bwdSpeed);
+        if (maxSpeed == MaxSpeed.UNSET_SPEED) {
+            return Double.NaN;
+        }
         return maxSpeed;
     }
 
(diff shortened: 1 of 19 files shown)


core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GermanySpatialRule.java
```diff
@@ -59,13 +59,10 @@ public double getMaxSpeed(RoadClass roadClass, TransportationMode transport, dou
...
-                return 100;
             case TERTIARY:
-                return 100;
             case UNCLASSIFIED:
-                return 100;
             case RESIDENTIAL:
-                return 100;
+                return 50;
             case LIVING_STREET:
                 return 4;
```
karussell, line 67: This does not look right. My understanding of the spatial rules was that they mirror the laws a country has regarding the maximum default speed. But 50km/h is only required inner city. Did you analyse the underlying cause you had to change this?

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
```diff
@@ -209,16 +210,14 @@ public double getMaxSpeed() {
...
-        if (isValidSpeed(backSpeed) && (!isValidSpeed(maxSpeed) || backSpeed < maxSpeed))
-            maxSpeed = backSpeed;
-
+    protected double getMaxSpeed(IntsRef edgeFlags) {
+        double fwdSpeed = maxSpeedEnc.getDecimal(false, edgeFlags);
+        double bwdSpeed = maxSpeedEnc.getDecimal(true, edgeFlags);
+        
+        double maxSpeed = Math.min(fwdSpeed, bwdSpeed);
+        if (maxSpeed == MaxSpeed.UNSET_SPEED) {
+            return Double.NaN;
```
karussell, line 219: Could we refactor the code so that it always use MaxSpeed.UNSET_SPEED instead of two "magic" values?

karussell (commented):
Thanks for your pull request!

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GermanySpatialRule.java
otbutz: The SpatialRule acts as a fallback here. We don't know if we're in urban area or not so we have to go with the more defensive assumption that we are.

Otherwise things like #2266 happen: The routing prefers an urban residential road as the maxspeed fallback is 100km/h  which is higher than the 70km/h of the B85. The routing acts accordingly and favors a detour.

core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java
otbutz: A bit unsure about it as `MaxSpeed.UNSET_SPEED` is rather specific to our encoding and the rest of the logic rolls with `NaN` which is IMHO a better representation of an unknown value. I'd rather see the default value of the MaxSpeed changed to `NaN`.

core/src/main/java/com/graphhopper/routing/util/spatialrules/countries/GermanySpatialRule.java
karussell: Hm, this sounds not the way to go. A spatial rule should stay like we defined and be identical to the local laws. If we mix in ETA improvements already here it won't work in the long term. So this PR is surely a step forward but we likely need to start with the ETA improvements before.

otbutz: > A spatial rule should stay like we defined and be identical to the local laws.

I agree but the law clearly differentiates between urban and non-urban speed limits. Using 100km/h as the fallback for all residential roads is a clear violation of the law for most urban areas.

karussell: I see what you mean but then we need to expand the spatial rules to support inner and outer limits. But IMO this is a bigger undertaking that we need to plan a bit. E.g. the first step would be to identify from OSM and other sources where "inner city" is, from there we could populate an encoded value `inner_city` and finally apply the correct rural or urban spatial rule.

For the first step we should be able to use vector (define our own areas or grab data from OSM) and raster data (using e.g. satellite imagery).

otbutz: Right now, SpatialRules are in a Schrödinger's cat situation: they are both right and wrong depending on which road segment they handle. I agree that we should add the distinction to GH at some point in the future, but for now, defaulting to the lower boundaries is a workable solution.

otbutz:
Ultimately we need to untangle this thicket of default values and derived ones from maximum values.

To summarize our current approach for routing:

* use a value from the [defaultSpeedMap](https://github.com/graphhopper/graphhopper/blob/be7b2476aaff55c7c266fd517c8cf2c37ae91b43/core/src/main/java/com/graphhopper/routing/util/CarTagParser.java#L109-L141) as basis for the average speed
* if we have a valid maxspeed tag for the way, we set the average speed to [90% of the limit](https://github.com/graphhopper/graphhopper/blob/be7b2476aaff55c7c266fd517c8cf2c37ae91b43/core/src/main/java/com/graphhopper/routing/util/CarTagParser.java#L259-L267)

This leads to a few quirks:
* the MaxSpeed EV is only used for details. e.g changing anything via `CountryRules.getMaxSpeed()` doesn't influence routing
* we have a single defaultSpeedMap which tries to cover multiple countries

We could gut the CountryRules, remove `getMaxSpeed()` and roll with parsed MaxSpeed but that doesn't improve the average speed calculation.

My proposal would be to replace `CountryRules.getMaxSpeed()` with `CountryRules.getAverageSpeed()`, finally promoting CountryRules from an optional feature to a core component. Our current defaultSpeedMap could be serve as the default implementation. So we would end up with the same average speeds but with the opportunity to tune them on a per-country basis.

easbar:
> Ultimately we need to untangle this thicket of default values and derived ones from maximum values.

I'd like to see the current VehicleTagParsers split into (vehicle) access and speed parsers. With this in place the speed parsers could use a country rule method that modifies the (global) speed in a country-specific way, just like the other country rule methods modify some encoded value in a country-specific way. Ok, splitting the vehicle parsers into access and speed is probably not absolutely necessary for this, but since this needs to be done anyway...

otbutz:
Let's close this one and see how things turn out with #2810, which should solve the gist of the problems described in https://github.com/graphhopper/graphhopper/pull/2268#issuecomment-1141349955.

PR #2278, opened 2021-04-07
Replace some usages of DefaultEdgeFilter in PT, leave some open questions

easbar:
I took a look at the usage of access flags in PT. We are currently trying to move access flags into the weighting as much as possible (#1835). The biggest issue I ran into here seems to be that PT uses the `Weighting` but does not call `calcEdgeWeight`, but only `calcEdgeMillis` *and* `FastestWeighting` returns infinite weight in case speed is zero, but not infinite time *and* PT does not set speeds, for example for the `VirtualEdgeIteratorStates` that are created in the overlayGraph in `RealtimeFeed`.

Maybe we can get rid of some edge filter usages entirely, because in a few cases we first filter by access flags, but then pick only certain PT edge types right after anyway. See the comments in the code.

It would be probably best to consider some of the changes in the `pt_graph` branch first before we make changes here. For example it moves all instances where PT edges are created to a common method.


floyedxman:
great performance

Issue #2273, opened 2021-04-06
Run subnetwork removal for each profile not just for each flag encoder.


easbar:
Connection not found errors often occur when the start/target locations belong to 'islands' that are are separated from the main road network.  This problem can be circumvented by excluding such islands such that the request points do not snap onto edges disconnected parts of the network in the first place. We currently do exactly this during the import, but we do remove the subnetworks depending on the flag encoder. With custom models this can be problematic, because custom model rules can disconnect otherwise connected parts of the road network, for example when we exclude bridges, tunnels, motorways etc.

Therefore we should  run the subnetwork removal separately for each profile, instead of running it for each flag encoder. This means we need access flags for each profile, not just for each encoder.

easbar:
Actually we need to distinguish between access flags we set based on OSM tags and access flags we set during subnetwork removal. We should keep the first kind as they are for now and use an additional bit for each profile that is used to mark subnetworks: A boolean encoded value that is true when an edge belongs to a subnetwork and false otherwise. Let's call it `is_subnetwork` for now? It should probably be added by default somewhere in the GraphHopper class (maybe unless min_network_size is 0?), but it should not be mandatory. Our `Weighting`s will need to receive the `is_subnetwork` EV as well.

karussell:
> Let's call it is_subnetwork for now?

I'm working on this right now and it seems that we can consider the small subnetwork exclusion only for the LocationIndex lookup, but a couple of tests needs to be adapted then. In the Weighting we then know that we are in one of the main networks and that the route should be found (or fails for obvious reason like "Connection not found" when searching a route from the US to EU). We could even store which of the subnetwork every edge is but we already do this for landmarks and so it would be only a minor speed up for "connection not found" when doing CH or "pure" A* requests and so I would not want to invest the additionally required memory in this.

The advantage of this separation is that we can disable this "minimal subnetwork exclusion" per request to allow routing on a real small island. And then we can easily show the roads to the user that are inaccessible e.g. via path details (or in the debug MVT view).

easbar:
> Our Weightings will need to receive the is_subnetwork EV as well.

Yes, I think you are right, the weihtings actually do not have to be aware of the flag, because once we 'snap' to a certain subnetwork the routeing automatically stays within this subnetwork :+1:

karussell:
> Ok, but it could be a really big improvement, no?
> Does this not mean we explore the entire main network before we can be sure the connection was not found? But 

[...]

PR #2280, opened 2021-04-13
Always use edge-based subnetwork removal

karussell:
In preparation for #2273

(from production statistics it seems that speed is same or even faster compared to node-based subnetwork removal)


easbar (approved):
Cool thanks! For later reference also see this comment: https://github.com/graphhopper/graphhopper/pull/2044#discussion_r428512822

PR #2284, opened 2021-04-17
Bug wrong exception type

hoofstephan:

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -35,6 +35,7 @@ Here is an overview:
  * HelgeKrueger, modularization of javascript, #590
  * henningvs, doc improvements
  * highsource, more efficient geometry update, UI fixes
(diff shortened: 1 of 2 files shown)


easbar:
Can you explain why we need to throw `IllegalArgumentException` instead of `IllegalStateException` or why the current code is problematic?

hoofstephan:
One should throw exceptions which represent the cause as close as possible. In this case the cause is a wrong argument value (empty string). Note that I throw a NullPointerException if the argument is null. I choose to separate these two to make the code more readable. And, according to Joshua Blog, one should throw NullPointerExceptions for prohibeted null arguments.

An IllegalStateException is (for example) appropriate when you call a method on an object whereas another method should have been called first (e.g. an init() call, which places the object in "correct" state)

easbar:
Ok yes this makes sense. Thanks!

Issue #2005, opened 2020-04-20
Web UI does not properly handle different screen sizes / make it responsive


HarelM:
**Describe the bug**
The sidebar appear at the top of the screen when using firefox to surf to http://localhost:8989/
Also there's no icon for car4wd profile... (small bug)

**To Reproduce**
1. Surf to http://localhost:8989/ using firefox on windows

**Expected behavior**
The sidebar should be at the side like in chrome

**System Information**
Used latest code: 1.0-pre33.4

**Screenshots & Logs**
![image](https://user-images.githubusercontent.com/3269297/79726635-c9a63c00-82f3-11ea-8211-6ed4e27919af.png)

easbar:
FF/Ubuntu works fine (for me at least). Why is the map not visible on your screenshot?

HarelM:
I'm using FF on windows.
Not sure why I don't see any tiles, but switching to another source shows the map, it might be an issue but not the one reported here :-)

easbar:
Ok I was asking because its a bit hard to tell from your screen shot what the problem is. You mean the side bar is on the left but does not go all the way to the bottom of the screen?

HarelM:
Sorry, it happens also in chrome. I saw it on the a server I have which has poor resolution when using remote desktop.
When I make the browser "zoom" in to 150% in chrome and 120% in firefox the map moves down and the sidebar stays at top:
Chrome:
This happens in https://graphhopper.com/maps as well:
![image](https://user-images.githubusercontent.com/3269297/81612400-70f32c00-93e5-11ea-8979-17b75e83830d.png)
Firefox:
![image](https://user-images.githubusercontent.com/3269297/81612438-7badc100-93e5-11ea-8000-bb6485b87cb7.png)

easbar:
Ah so its a zooming issue. I can reproduce this too, but if I keep zooming more the map moves to the right of the sidebar again :) Generally, the web UI does not properly deal with different screen sizes, for example I cannot use it on my phone because the sidebar is so large I cannot even see the map.

HarelM:
I would suggest to adopt a responsive framework from one of the following top 5:
https://www.themexpert.com/blog/top-5-responsive-css-framework
I've had the chance to work with bootstrap and it is very good, but it's a bit old in terms of web and I think less and less companies decide to work with it nowadays.
A mobile site can have the search bar at the top and the map at the bottom to facilitate for this. a responsive design is basically the definition of this issue :-).
[...]

PR #2285, opened 2021-04-18
Improved responsiveness of Web UI #2005

jguddas:
This PR fixes the following issues:

**the line graph being cut of on small screens**  
solved by adding dynamic width adjustment

before:  
image

after:
:warning: I was unable to test this with europe_germany_berlin.pbf :warning:

**the route information disappearing on small screens**  
solved by adding a min height of 100px

before:  
image

after:  
image

**missing support for small screens**  
solved by tweaking the css in a media query

before:  
image

after:  
image


easbar:
> I was unable to test this with europe_germany_berlin.pbf 

I just tried this and the elevation widget seems to be invisible. All we see are the open/close buttons:

image

image

jguddas:
@easbar I think I fixed the issue.

I'm still unable to test it after changing the config and doing a clean rebuild, thanks for your help anyway.

karussell:
Unfortunately I have the same problem as @easbar.

> I'm still unable to test it after changing the config and doing a clean rebuild

Can you describe the problem why you are unable to test this?

What I did to test this: I switched to your branch. Then I downloaded the berlin-latest.osm.pbf, used the sample config and enabled elevation like  @easbar described. Then removed the `graph-cache` folder. Then `cd web-bundle` and `npm run watch` and finally started GH with IntelliJ to force a fresh import. Or instead you can also build the jar via command line `mvn clean install` and run as described in [the readme](https://github.com/graphhopper/graphhopper/#installation).

easbar:
> Unfortunately I have the same problem as @easbar.

After the latest commits the elevation widget was visible for me. But it did not adjust its width. Another problem was that after using the full screen button some parts of the screen remained white, so the space was not used.

easbar:
Thanks for your work and also your latest fix. The full screen issue seems to be resolved. I cannot really say much about the css code, but just by testing the results I think this PR is an improvement compared to master. I tested for both Chrome and Firefox.

easbar:
Ah but now I got another error (in the browser's console):

```
map.js:285 Uncaught ReferenceError: infinity is not defined
    at Object.module.exports.addElevation (map.js:285)
    at HTMLLIElement.<anonymous> (main-template.js:659)
    at HTMLLIElement.dispatch (jquery.js:5429)
    at HTMLLIElement.elemData.handle (jquery.js:5233)
    at Object.trigger (jquery.js:8715)
    at HTMLLIElement.<anonymous> (jquery.js:8793)
    at Function.each (jquery.js:381)
    at jQuery.fn.init.each (jquery.js:203)
    at jQuery.fn.init.trigger (jquery.js:8792)
    at jQuery.fn.init.jQuery.fn.<computed> [as click] (jquery.js:10738)
```

It depends on the screen size, but this `infinity` variable in `map.js` seems to be non-existing.

PR #2286, opened 2021-04-19
Restructure code in Router

easbar:
The route calculation in Router now uses different 'Solver's depending on request. Plugging in different solvers allows modifying profile selection, weighting creation, snapping behavior and routing algorithm selection.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -974,6 +974,11 @@ public class GraphHopper implements GraphHopperAPI {
                             ? lmPreparationHandler.getPreparation(lmp.getPreparationProfile()).getLandmarkStorage()
                             : lmPreparationHandler.getPreparation(lmp.getProfile()).getLandmarkStorage());
         }
+        return doCreateRouter(ghStorage, locationIndex, profilesByName, pathBuilderFactory,
+                trMap, routerConfig, createWeightingFactory(), chGraphs, landmarks);
+    }
+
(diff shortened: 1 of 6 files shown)


boldtrn:
Thanks for providing this PR. I was just wondering, what are the reasons to have the solver as static inner class? Wouldn't it be cleaner to break this down into multiple classes?

easbar:
Hm no real reason I think, just no real reason to separate it, yet. You mean we should make the solver classes separate public (?) classes? Btw currently you can also override `createSolver` if you want to provide your own.

boldtrn:
> I am still not sure about these abstractions, so I am trying to be careful, I guess.

Yes I totally get that. Unless it is required, code can be easier without abstraction, so makes sense 👍 .

> You mean we should make the solver classes separate public (?) classes?

Yes :). I don't have any strong points for that, it's more a feeling:
- The Router is now at ~550 lines, this makes it more complicated, so having this split into smaller parts would be beneficial (The solvers are in total about 230 lines).
- Somehow the inner class feels a bit weird. Usually I try to avoid inner classes.

To be honest I was mostly wondering about the usage of static at the inner class. So I took a closer look.

easbar:
> To be honest I was mostly wondering about the usage of static at the inner class.

Ah I make every inner class static if its possible. When it's static it means its unrelated or decoupled to the actual class, so it can be easily moved somewhere else if need be. I usually try to avoid non-static inner classes.

boldtrn:
> When it's static it means its unrelated or decoupled to the actual class, so it can be easily moved somewhere else if need be.

Yes, and I think that was the reason why I thought it should be moved out :).

otbutz:
non-static inner classes also have a slight overhead ;)

https://rules.sonarsource.com/java/RSPEC-2694

easbar:
I don't think the router is more complicated because the file is longer. It would be more complicated if it contained more entangled code, but since the solvers are more or less independent its easy to reason about them or about the rest of the router in isolation. But sure maybe the next step would be moving them to separate files. That would be an indication for me that they should/can be used by themselves, but currently I am not sure about it :) Well, it does not matter that much, basically a static inner class just belongs to another 'namespace'. But obviously if there is any valid use-case that requires moving the solvers that should not be a big deal.

PR #2289, opened 2021-04-20
Split custom model docs into client and server parts, update to JSON

### LINKED ISSUE #2275: Update custom model and profile documentation
See #2232

easbar:
Fix #2275.

### DIFF
--- a/docs/core/custom-models.md
+++ b/docs/core/custom-models.md
@@ -133,12 +133,6 @@ Here is a complete request example for a POST /route query in berlin that includ
 } 
 ```
 
-If you put this JSON into a file `custom.json` you can send the request like this:
-
-```shell
-curl -XPOST -H "Content-Type: application/json" "http://localhost:8989/route" --data custom.json
-```
-
 Note that this only works for custom profiles and so far only for POST /route (but not GET /route or /isochrone, /spt or
 /map-matching).
 
--- a/docs/core/custom-models.md
+++ b/docs/core/custom-models.md
@@ -1,0 +1,538 @@
+# Custom Models
+
+GraphHopper provides an easy-to-use way to customize its route calculations: Custom models allow you to modify the
+default routing behavior by specifying a set of rules in JSON language. Here we will first explain some theoretical
+background and then show how to use custom models in practice.
+
+## How GraphHopper's route calculations work
+
+One of GraphHopper's most important functionality is the calculation of the 'optimal' route between two locations. To do
+this, GraphHopper subdivides the entire road network into so called 'edges'. Every edge represents a certain road
+segment between two junctions. Therefore finding the optimal route between two locations means finding the optimal
+sequence of edges that connect the two locations. GraphHopper stores certain attributes (so called 'encoded values') for
+every edge and applies a formula (the so called 'weighting') to calculate a numeric 'weight' for every edge. The total
+weight of a route is the sum of all the edge's weights. The optimal route is the one where the total weight is the
+smallest.
+
+For example you can imagine the edge weight to be the time you need to travel from one junction to another. Finding the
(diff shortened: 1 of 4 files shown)


docs/core/custom-models.md
```diff
@@ -0,0 +1,538 @@
...
+Using custom models is simple. You just need to add the `custom_model` property to your request JSON. This only works
+for certain profiles, however, and so far only POST /route (but not GET /route or /isochrone, /spt or /map-matching)
+are supported.
+
+### via GraphHopper maps
+
+GraphHopper maps offers an interactive text editor that can be used to comfortably enter custom models. You can open it
+by pressing the 'custom' button. It will check the syntax of your custom model and mark errors in red. You can press
+Ctrl+Space or Alt+Enter to retrieve auto-complete suggestions. Pressing Ctrl+Enter will send a routing request for the
+custom model you entered.
```
karussell, line 538: I would even move this section to the top of the description?

What I find also important is the show a complete example e.g. "The POST body for a query in Berlin looks like:"

```json
{
  "points": [
    [ 13.31543, 52.509535 ],
    [ 13.29779, 52.512434 ]
  ],
  "custom_model": {
    "priority": [
      {
        "if": "road_class == MOTORWAY",
        "multiply_by": 0
      }
    ]
  },
  "profile": "car"
} 
```

"If you put this into a file route.json you can immediately try this:"

```
curl -XPOST -H "Content-Type: application/json" "http://localhost:8989/route" --data route.json
```

karussell (approved):
Thanks a lot!

docs/core/custom-models.md
easbar: Ok yes good idea

easbar: > curl -XPOST -H "Content-Type: application/json" "http://localhost:8989/route" --data route.json

This requires a local server, shall we change localhost to the custom model demo server IP?

karussell: I would like to get rid of these demo servers, when we get this feature on graphhopper.com ( hopefully soonish ;) )

And I think that the reader here will have the ability to set up a local routing server. For our docs at docs.graphhopper.com we probably have to copy these docs (?)

easbar: Instead of copying the docs we could also just move them entirely? But for docs.graphhopper.com maybe we want to be specific about which encoded values exist etc. too. So far I thought custom-models.md is for client-side only users (like on GH maps) so they do not need to know about setting up a local server. Setting up a local server is described in profiles.md. But maybe we can just remove the comment about curl here because how to send a GH request is described elsewhere anyway.

karussell: Yes, feel free to remove the curl command.

> Instead of copying the docs we could also just move them entirely? 

In general we cannot use the same docs here as we use in docs.graphhopper.com. I'm not sure if we would need a separate documentation for custom models and although introducing this separation causes lots of heading it also allows us to independently move forward here without "production" limitations.

easbar: Ok so let's see when we actually do it and then probably copy it (or most parts of it)?

karussell: :+1:

docs/core/custom-models.md
```diff
@@ -100,8 +100,56 @@ of this you need to specify a so called 'custom model', which is a set of rules
...
+    ],
+    "distance_influence": 50
+  }
+} 
+```
+
+If you put this JSON into a file `custom.json` you can send the request like this:
+
+```shell
+curl -XPOST -H "Content-Type: application/json" "http://localhost:8989/route" --data custom.json
```
karussell: Just for later reference - there was a mistake and it should be `--data @custom.json`.

PR #2291, opened 2021-04-22
Remove GraphHopper#loadGraph and Princeton performance test

easbar:
`GraphHopper#loadGraph` was used in some tests to 'sneak in' some kind of manually built graph. But this defeats the purpose of the `GraphHopper` class I think. We have some high-level tests in `GraphHopperTest` that work with real OSM imports (as intended by the `GraphHopper` class), but if all we want to do is test some routing requests on some manually built graph we should rather write tests that just to that (like `RoutingAlgorithmTest`). 

The Princeton performance test has been replaced by Measurement now and is no longer needed IMO. It only ran with `N=10` runs anyway and it is hard to use a hard limit, because the performance is also hardware dependent.


karussell:
Nice, thanks for this clean up :+1:

PR #2295, opened 2021-04-25
Clean up PrepareRoutingSubnetwork and tests

easbar:
I reviewed #2290 again and put my suggestions into this PR. See comments inline.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -1096,7 +1096,7 @@ public class GraphHopper implements GraphHopperAPI {
         for (Profile profile : profilesByName.values()) {
             // if turn costs are enabled use u-turn costs of zero as we only want to make sure the graph is fully connected assuming finite u-turn costs
             Weighting weighting = createWeighting(profile, new PMap().putObject(Parameters.Routing.U_TURN_COSTS, 0));
-            jobs.add(new PrepareJob(profile.getName(), encodingManager.getBooleanEncodedValue(Subnetwork.key(profile.getName())), weighting));
+            jobs.add(new PrepareJob(encodingManager.getBooleanEncodedValue(Subnetwork.key(profile.getName())), weighting));
         }
         return jobs;
     }
--- a/core/src/main/java/com/graphhopper/routing/subnetwork/EdgeBasedTarjanSCC.java
+++ b/core/src/main/java/com/graphhopper/routing/subnetwork/EdgeBasedTarjanSCC.java
@@ -21,7 +21,6 @@ package com.graphhopper.routing.subnetwork;
 import com.carrotsearch.hppc.*;
 import com.carrotsearch.hppc.cursors.IntCursor;
 import com.graphhopper.routing.util.AllEdgesIterator;
-import com.graphhopper.routing.weighting.Weighting;
 import com.graphhopper.storage.Graph;
 import com.graphhopper.util.BitUtil;
 import com.graphhopper.util.EdgeExplorer;
@@ -46,7 +45,7 @@ import java.util.List;
  */
 public class EdgeBasedTarjanSCC {
     private final Graph graph;
-    private final Weighting weighting;
+    private final EdgeTransitionFilter edgeTransitionFilter;
     private final EdgeExplorer explorer;
     private final BitUtil bitUtil = BitUtil.LITTLE;
     private final IntArrayDeque tarjanStack;
@@ -67,26 +66,25 @@ public class EdgeBasedTarjanSCC {
     /**
      * Runs Tarjan's algorithm using an explicit stack.
      *
-     * @param weighting                   Only edges accepted by this filter will be considered when we explore the graph.
-     *                                    If a turn has infinite costs the corresponding
-     *                                    path will be ignored (edges that are only connected by a path with such a turn will not
-     *                                    be considered to belong to the same component)
+     * @param edgeTransitionFilter        Only edge transitions accepted by this filter will be considered when we explore the graph.
+     *                                    If a turn is not accepted the corresponding path will be ignored (edges that are only connected
+     *                                    by a path with such a turn will not be considered to belong to the same component)
      * @param excludeSingleEdgeComponents if set to true components that only contain a single edge will not be
      *                                    returned when calling {@link #findComponents} or {@link #findComponentsRecursive()},
      *                                    which can be useful to save some memory.
      */
-    public static ConnectedComponents findComponents(Graph graph, Weighting weighting, boolean excludeSingleEdgeComponents) {
-        return new EdgeBasedTarjanSCC(graph, weighting, excludeSingleEdgeComponents).findComponents();
+    public static ConnectedComponents findComponents(Graph graph, EdgeTransitionFilter edgeTransitionFilter, boolean excludeSingleEdgeComponents) {
+        return new EdgeBasedTarjanSCC(graph, edgeTransitionFilter, excludeSingleEdgeComponents).findComponents();
     }
 
     /**
-     * Like {@link #findComponents(Graph, Weighting, boolean)}, but the search only starts at the
-     * given edges. This does not mean the search cannot expand to other edges, but this can be controlled using by
-     * the edgeFilter. This method does not return single edge components (the excludeSingleEdgeComponents option is
+     * Like {@link #findComponents(Graph, EdgeTransitionFilter, boolean)}, but the search only starts at the
+     * given edges. This does not mean the search cannot expand to other edges, but this can be controlled by the
+     * edgeTransitionFilter. This method does not return single edge components (the excludeSingleEdgeComponents option is
      * set to true).
      */
-    public static ConnectedComponents findComponentsForStartEdges(Graph graph, Weighting weighting, IntContainer edges) {
-        return new EdgeBasedTarjanSCC(graph, weighting, true).findComponentsForStartEdges(edges);
+    public static ConnectedComponents findComponentsForStartEdges(Graph graph, EdgeTransitionFilter edgeTransitionFilter, IntContainer edges) {
+        return new EdgeBasedTarjanSCC(graph, edgeTransitionFilter, true).findComponentsForStartEdges(edges);
     }
 
     /**
@@ -94,15 +92,15 @@ public class EdgeBasedTarjanSCC {
      * which can be set like `-Xss1024M`. Usually the version using an explicit stack ({@link #findComponents()}) should be
      * preferred. However, this recursive implementation is easier to understand.
      *
-     * @see #findComponents(Graph, Weighting, boolean)
+     * @see #findComponents(Graph, EdgeTransitionFilter, boolean)
      */
-    public static ConnectedComponents findComponentsRecursive(Graph graph, Weighting weighting, boolean excludeSingleEdgeComponents) {
-        return new EdgeBasedTarjanSCC(graph, weighting, excludeSingleEdgeComponents).findComponentsRecursive();
+    public static ConnectedComponents findComponentsRecursive(Graph graph, EdgeTransitionFilter edgeTransitionFilter, boolean excludeSingleEdgeComponents) {
+        return new EdgeBasedTarjanSCC(graph, edgeTransitionFilter, excludeSingleEdgeComponents).findComponentsRecursive();
     }
 
-    private EdgeBasedTarjanSCC(Graph graph, Weighting weighting, boolean excludeSingleEdgeComponents) {
+    private EdgeBasedTarjanSCC(Graph graph, EdgeTransitionFilter edgeTransitionFilter, boolean excludeSingleEdgeComponents) {
         this.graph = graph;
-        this.weighting = weighting;
+        this.edgeTransitionFilter = edgeTransitionFilter;
         this.explorer = graph.createEdgeExplorer();
         tarjanStack = new IntArrayDeque();
         dfsStackPQ = new LongArrayDeque();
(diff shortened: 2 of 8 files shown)


core/src/main/java/com/graphhopper/routing/subnetwork/EdgeBasedTarjanSCC.java
```diff
@@ -46,7 +45,7 @@
  */
 public class EdgeBasedTarjanSCC {
     private final Graph graph;
-    private final Weighting weighting;
+    private final EdgeTransitionFilter edgeTransitionFilter;
```
easbar, line 48: I wanted to reduce the dependency between the tarjan algorithm and weighting. Now I have this interface which specifies whether or not an edge-transition is allowed or not, similarly to node-based tarjan where we specify whether an edge is allowed or not (using EdgeFilter)

core/src/main/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworks.java
```diff
@@ -22,24 +22,36 @@
...
+ * one-way subnetworks it is important to search for strongly connected components on the directed graph and not do a
+ * simple connectivity check for one direction.
+ * <p>
+ * Note that it depends on the weighting whether or not edges belong to a subnetwork or not. For example if a weighting
+ * 'closes' a bridge to an island the island might become a subnetwork, but if the bridge was open it would belong to
+ * the main network. There can even be subnetworks that are due to turn restrictions.
+ * <p>
+ * We always run an edge-based connected component search, because this way we retrieve the edges (not the nodes) that
+ * belong to each component and can include turn restrictions as well. Node-based component search is faster, but since
+ * the subnetwork search goes relatively fast anyway using it has no real benefit.
```
easbar, line 54: I updated the docs and tried to get rid of everything that mentioned the 'removal' of edges or nodes, because we do not (or do no longer) remove anything. All this class now does is it sets the subnetwork encoded value.

core/src/main/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworks.java
```diff
@@ -56,155 +68,118 @@ public PrepareRoutingSubnetworks(GraphHopperStorage ghStorage, List<PrepareJob>
     }
 
     /**
-     * The subnetwork removal removes components with less than {@link #minNetworkSize} nodes from the graph if it is
-     * run node-based. For edge-based subnetwork removal it removes components with less than 2*{@link #minNetworkSize}
-     * (directed) edges.
+     * All components of the graph with less than 2*{@link #minNetworkSize} directed edges (edge keys) will be marked
+     * as subnetworks. The biggest component will never be marked as subnetwork, even when it is below this size.
```
easbar, line 72: All subnetwork preparation is now edge-based so no need to talk about node-based anymore.

core/src/main/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworks.java
```diff
@@ -56,155 +68,118 @@ public PrepareRoutingSubnetworks(GraphHopperStorage ghStorage, List<PrepareJob>
...
-     *
-     * @return number of removed edges
-     */
-    int setSubnetworks(PrepareJob job) {
-        return setSubnetworks(job.name, job.weighting, job.inSubnetworkEnc);
+        int total = 0;
+        for (PrepareJob job : prepareJobs)
+            total += setSubnetworks(job.weighting, job.subnetworkEnc);
+        logger.info("Finished finding and marking subnetworks for " + prepareJobs.size() + " jobs, took: " + sw.stop().getSeconds() + "s, " + Helper.getMemInfo());
+        return total;
```
easbar, line 96: I now return the total number of marked subnetwork edges here, so we can make setSubnetworks private

core/src/main/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworks.java
```diff
@@ -56,155 +68,118 @@ public PrepareRoutingSubnetworks(GraphHopperStorage ghStorage, List<PrepareJob>
...
     }
 
-    private int setSubnetworks(String jobName, Weighting weighting, BooleanEncodedValue inSubnetworkEnc) {
+    private int setSubnetworks(Weighting weighting, BooleanEncodedValue subnetworkEnc) {
         // partition graph into strongly connected components using Tarjan's algorithm
         StopWatch sw = new StopWatch().start();
-        EdgeBasedTarjanSCC.ConnectedComponents ccs = EdgeBasedTarjanSCC.findComponents(ghStorage, weighting, false);
+        EdgeBasedTarjanSCC.ConnectedComponents ccs = EdgeBasedTarjanSCC.findComponents(ghStorage,
+                (prev, edge) -> Double.isFinite(GHUtility.calcWeightWithTurnWeightWithAccess(weighting, edge, false, prev)),
+                false);
```
easbar, line 104: Here we setup the EdgeTransitionFilter based on the weighting. We could go further and do this already in PrepareJob, but so far I do not see much benefit of this.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1096,7 +1096,7 @@ protected void cleanUp() {
         for (Profile profile : profilesByName.values()) {
             // if turn costs are enabled use u-turn costs of zero as we only want to make sure the graph is fully connected assuming finite u-turn costs
             Weighting weighting = createWeighting(profile, new PMap().putObject(Parameters.Routing.U_TURN_COSTS, 0));
-            jobs.add(new PrepareJob(profile.getName(), encodingManager.getBooleanEncodedValue(Subnetwork.key(profile.getName())), weighting));
+            jobs.add(new PrepareJob(encodingManager.getBooleanEncodedValue(Subnetwork.key(profile.getName())), weighting));
         }
```
easbar, line 1100: The prepare job name was only used for logs and I now use subnetworkEnc.getName() instead.

core/src/main/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworks.java
```diff
@@ -56,155 +68,118 @@ public PrepareRoutingSubnetworks(GraphHopperStorage ghStorage, List<PrepareJob>
...
+        return markedEdges;
     }
 
-    /**
-     * This method checks if the node is removed or inaccessible for ALL encoders.
-     *
-     * @return true if no edges are reachable from the specified nodeIndex for any flag encoder.
-     */
-    boolean detectNodeRemovedForAllEncoders(EdgeExplorer edgeExplorerAllEdges, int nodeIndex) {
-        // we could implement a 'fast check' for several previously marked removed nodes via GHBitSet
```
easbar, line 170: This method was not used anymore. We probably forgot to remove it in a previous refactoring

core/src/main/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworks.java
```diff
@@ -56,155 +68,118 @@ public PrepareRoutingSubnetworks(GraphHopperStorage ghStorage, List<PrepareJob>
...
-     *
-     * @return true if no edges are reachable from the specified nodeIndex for any flag encoder.
-     */
-    boolean detectNodeRemovedForAllEncoders(EdgeExplorer edgeExplorerAllEdges, int nodeIndex) {
-        // we could implement a 'fast check' for several previously marked removed nodes via GHBitSet 
-        // removedNodesPerVehicle. The problem is that we would need long-indices but BitSet only supports int (due to nodeIndex*numberOfEncoders)
+    private int setSubnetworkEdge(int edgeKey, Weighting weighting, BooleanEncodedValue subnetworkEnc) {
+        // edges that are not accessible anyway are not marked as subnetworks additionally
+        if (!Double.isFinite(weighting.calcEdgeWeightWithAccess(ghStorage.getEdgeIteratorStateForKey(edgeKey), false)))
+            return 0;
```
easbar, line 159: I think this is useful to get a useful number of 'subnetwork edges', otherwise e.g. for bike every motorway edge would be included. But for our routing code it probably does not matter whether or not edges with access=false or infinite weight are additionally marked as subnetwork edges.

core/src/main/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworks.java
```diff
@@ -56,155 +68,118 @@ public PrepareRoutingSubnetworks(GraphHopperStorage ghStorage, List<PrepareJob>
...
-                if (edgeFilter.accept(iter))
-                    return false;
-            }
+        int edge = EdgeBasedTarjanSCC.getEdgeFromKey(edgeKey);
+        EdgeIteratorState edgeState = ghStorage.getEdgeIteratorState(edge, Integer.MIN_VALUE);
+        if (!edgeState.get(subnetworkEnc)) {
+            edgeState.set(subnetworkEnc, true);
+            return 1;
+        } else {
+            return 0;
```
easbar, line 167: Previously we used the EdgeFilter from createEdgeFilter here, but I think this way the code is more clear.

core/src/test/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworksTest.java
```diff
@@ -83,103 +71,111 @@ private static GraphHopperStorage createSubnetworkTestStorage(String flagEncoder
...
+        assertEquals(0, instance.doWork());
+        assertEquals(IntArrayList.from(), getSubnetworkEdges(g, encoder));
     }
 
     @Test
-    public void testRemoveNode() {
-        GraphHopperStorage g = createSubnetworkTestStorage("car,bike");
-        EncodingManager em = g.getEncodingManager();
-        FlagEncoder carEncoder = g.getEncodingManager().getEncoder("car");
-        FlagEncoder bikeEncoder = g.getEncodingManager().getEncoder("bike");
```
easbar, line 145: This test tested a method that was not used anyway and did not even use PrepareRoutingSubnetworks to prepare the subnetworks, so it basically tested nothing :) I probably forgot deleting it in a previous refactoring.

core/src/test/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworksTest.java
```diff
@@ -226,35 +226,38 @@ GraphHopperStorage createSubnetworkTestStorageWithOneWays(String flagEncodersStr
...
         PrepareRoutingSubnetworks instance = new PrepareRoutingSubnetworks(g, Collections.singletonList(job)).
                 setMinNetworkSize(3);
-        int removed = instance.setSubnetworks(job);
+        int subnetworkEdges = instance.doWork();
 
-        assertEquals(5, removed);
-        assertTrue(isConsistent(g));
-        g.edge(7, 8);
-        assertTrue(isConsistent(g));
+        // this time also the (8,9,10) component is a subnetwork
```
easbar, line 258: I think since we are not removing edges or nodes anymore there is no need to check the 'consistency' of the graph or whether or not we can add edges after subnetwork search (which we do not do in production anyway). All we need to test is that the right edges get marked as subnetwork edges.

core/src/test/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworksTest.java
```diff
@@ -34,32 +39,15 @@
 import java.util.List;
 
 import static com.graphhopper.routing.weighting.TurnCostProvider.NO_TURN_COST_PROVIDER;
-import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 /**
  * @author Peter Karich
  */
 public class PrepareRoutingSubnetworksTest {
```
easbar, line 48: I tried cleaning up this test, but honestly I still do not like it because it does not really systematically test what PrepareRoutingSubnetworks (now) does. I will probably rewrite it from scratch.

easbar: Ok I improved the test a bit further in my last commit, now it also includes a test with turn restrictions and I think it is in a good state.

core/src/main/java/com/graphhopper/routing/subnetwork/EdgeBasedTarjanSCC.java
karussell: Good idea. What about deprecating EdgeFilter and use this new more powerful interface instead in general?

core/src/main/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworks.java
karussell: Thanks!

core/src/main/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworks.java
karussell: Ah, ok. I saw this as well but thought it was used in tests after doWork to see if nodes were removed but as you pointed out the doWork method wasn't called at all. Ups.

core/src/test/java/com/graphhopper/routing/subnetwork/PrepareRoutingSubnetworksTest.java
karussell: Nice

karussell (approved):
Thanks a lot for this!

core/src/main/java/com/graphhopper/routing/subnetwork/EdgeBasedTarjanSCC.java
karussell: But maybe for a separate PR :)

easbar: Hm the 'transition' part is not always needed. For example `LocationIndex#findClosest(lat, lon, EdgeFilter)` or `Graph#createEdgeExplorer(EdgeFilter)` are just filtering edges without any context what the 'previous' edge was.

easbar:
Thanks for the review!

Issue #2207, opened 2020-12-09
Conflicting protobuf-java versions


otbutz:
The situation at the moment:

* reader-osm: protobuf-java **3**.11.4 as a dependency of osmosis-osm-binary
* reader-gtfs: protobuf-java **2**.6.1as dependency of gtfs-realtime-bindings

This causes problems in my draft PR (#2205) as protobuf v3 classes are not available.

We should probably move to a newer version of the gtfs-realtime-bindings to avoid this.

otbutz:
![grafik](https://user-images.githubusercontent.com/22315436/101611165-752e0900-3a09-11eb-8e0f-ddf46f24bab4.png)

karussell:
Thanks for pointing this out. Unfortunately gtfs realtime 0.0.7 seems to use only [protobuf 3.6.1](https://github.com/MobilityData/gtfs-realtime-bindings/blob/master/java/pom.xml). And protobuf 3.14.0 seems to be the most recent version.

But maybe 3.x is compatible from a Java standpoint? Can you try if 0.0.7 passes tests and still makes the problems you mentioned?

otbutz:
Switching to gtfs-realtime-bindings 0.0.7 provided by io.mobilitydata.transit while excluding the transient dependency on protobuf 3.6.1 seems to fix the issue. At least the unittests in my draft PR pass now.

Should i create a separate PR for the fix or can i add this to my existing POM cleanup PR? https://github.com/graphhopper/graphhopper/pull/2200

karussell:
Let's wait what @michaz says :)

michaz:
Oh, please go ahead, as long as it works. There's nothing special happening there.

otbutz:
Fixed by #2200

PR #2297, opened 2021-04-26
use locationtech in isochrone instead of mapbox-vector-tile

karussell:


otbutz:
https://github.com/graphhopper/graphhopper/issues/2207

easbar:
The motivation for this PR is that the isochrone module does not really need mapbox-vector-tile, but only a dependency of it (jts). Therefore we removed the mapbox-vector-tile dependency in the isochrone pom. However, mapbox-vector-tile actually *is* needed for web-bundle, so we added it there.

The problem is/was that mapbox-vector-tile pulls in protobuf 3.5.1, but osmosis-osm-binary (used in core) uses protobuf 3.11.4 and stops working with protobuf 3.5.1. Therefore we exclude the protobuf version of mapbox-vector-tile now. Luckily, it continues to work with the newer version (3.11.4) we get from osmosis.

@otbutz Thanks for the reminder. This is another version clash: gtfs-realtime-bindings 0.0.4 pulls in yet another (and even older) protobuf dependency (2.6.1). For this PR it is enough to exclude the protobuf artifact brought in by mapbox-vector-tile though.

easbar (approved):

PR #2301, opened 2021-04-28
Make custom model box work with dragging markers, changing vehicle, adding points etc.

easbar:
This is my attempt to better integrate the custom model box in the web client. I think it 'works', but for example it kind of breaks the url history as we cannot store the custom model via the url parameters.

### DIFF
--- a/web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js
+++ b/web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js
@@ -78,17 +78,21 @@ $(document).ready(function (e) {
     // todo: the idea was to highlight everything so if we start typing the example is overwritten. But unfortunately
     // this does not work. And not even sure this is so useful?
     showCustomModelExample();
-    cmEditor.validListener = function(valid) {
+    cmEditor.validListener = function (valid) {
         $("#custom-model-search-button").prop('disabled', !valid);
     };
-    $("#custom-model-button").click(function() {
+    ghRequest.cmEditor = cmEditor;
+    ghRequest.cmEditorActive = false;
+    $("#custom-model-button").click(function () {
         $("#custom-model-box").toggle();
+        ghRequest.cmEditorActive = !ghRequest.cmEditorActive;
         // avoid default action, so use a different search button
         $("#searchButton").toggle();
         mapLayer.adjustMapSize();
         cmEditor.cm.refresh();
         cmEditor.cm.focus();
-        cmEditor.cm.setCursor(cmEditor.cm.lineCount())
+        cmEditor.cm.setCursor(cmEditor.cm.lineCount());
+        sendCustomData();
     });
     function showCustomModelExample() {
         cmEditor.value =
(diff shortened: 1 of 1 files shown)


web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js
```diff
@@ -78,17 +78,21 @@ $(document).ready(function (e) {
     // todo: the idea was to highlight everything so if we start typing the example is overwritten. But unfortunately
     // this does not work. And not even sure this is so useful?
     showCustomModelExample();
-    cmEditor.validListener = function(valid) {
+    cmEditor.validListener = function (valid) {
         $("#custom-model-search-button").prop('disabled', !valid);
     };
-    $("#custom-model-button").click(function() {
+    ghRequest.cmEditor = cmEditor;
```
easbar, line 84: ghRequest now keeps a reference to the custom model editor. rather hacky, but it 'works' :)

web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js
```diff
@@ -78,17 +78,21 @@ $(document).ready(function (e) {
...
+    ghRequest.cmEditorActive = false;
+    $("#custom-model-button").click(function () {
         $("#custom-model-box").toggle();
+        ghRequest.cmEditorActive = !ghRequest.cmEditorActive;
```
easbar, line 88: here we simply keep track whether or not the custom model box is opened or not. if it is we send a custom request, otherwise we ignore the custom model box and send the same request we did before.

web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js
```diff
@@ -78,17 +78,21 @@ $(document).ready(function (e) {
...
         // avoid default action, so use a different search button
         $("#searchButton").toggle();
         mapLayer.adjustMapSize();
         cmEditor.cm.refresh();
         cmEditor.cm.focus();
-        cmEditor.cm.setCursor(cmEditor.cm.lineCount())
+        cmEditor.cm.setCursor(cmEditor.cm.lineCount());
+        sendCustomData();
```
easbar, line 95: I thought it is more intuitive to update the route whenever we open/close the custom model box

web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js
```diff
@@ -115,72 +119,23 @@ $(document).ready(function (e) {
...
     }
-    $("#custom-model-example").click(function() {
+
+    $("#custom-model-example").click(function () {
         showCustomModelExample();
         return false;
     });
 
-    var sendCustomData = function() {
-       mapLayer.clearElevation();
```
easbar, line 124: clicking the custom model search button now triggers just a normal search, but before we send the request we check whether it should be custom or not.

web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js
```diff
@@ -115,72 +119,23 @@ $(document).ready(function (e) {
...
+        mySubmit();
     };
 
-    // todo: prevent this as long as custom model is invalid?
-    // so far it is useful so we can send the request regardless of the invalidation
-    cmEditor.setExtraKey('Ctrl-Enter', sendCustomData);
+    var sendCustomDataIgnoreErrors = function () {
+        ghRequest.ignoreCustomErrors = true;
+        mySubmit();
+    }
```
easbar, line 136: I thought it might be useful to ignore errors when pressing ctrl-enter, so this way we can force sending the request even when the custom model box reports an invalid custom model.

web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js
```diff
@@ -901,19 +856,71 @@ function routeLatLng(request, doQuery) {
...
 
-    var urlForAPI = request.createURL();
     routeResultsDiv.html('<img src="img/indicator.gif"/> Search Route ...');
-    request.doRequest(urlForAPI, createRouteCallback(request, routeResultsDiv, urlForHistory, doZoom));
+    if (request.cmEditorActive) {
+        doCustomRequest(request, routeResultsDiv);
+    } else {
+        var urlForAPI = request.createURL();
+        request.doRequest(urlForAPI, createRouteCallback(request, routeResultsDiv, urlForHistory, doZoom));
+    }
```
easbar, line 867: Here we finally check whether the custom box was active or not and send either the custom or the standard request. Since this logic is now down here, we can use the custom model box when we drag markers, change the marker order, change the vehicle, add/delete points etc.

easbar: Maybe we should not run the code block above that starts with `if (!doQuery && History.enabled) {` when the custom model editor is active?

karussell (approved):
Thank you - works for me :D

web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js
karussell: I really like this idea and solution! Then it is also easy to see what is changed due to the custom model.

PR #2305, opened 2021-05-06
Skip shapes which don't block any edges

otbutz:
BlockArea currently treats an empty GHIntHashSet the same way as a non-existing cache for blocked edges. But a small shape which doesn't cover edges in the graph would also yield an empty GHIntHashSet. For those cases BlockArea would resort to the more expensive bbox/shape intersection logic:

https://github.com/graphhopper/graphhopper/blob/0522c495ccad5f22fe19c6e858a40a7bd6f74eae/core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java#L198-L223

I changed the logic to explicitly store `null` as the cache for blocked edges if the shape is to large so we can properly determine if we have no cache or if it's just empty.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
+++ b/core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
@@ -28,6 +28,7 @@ import org.locationtech.jts.algorithm.RectangleLineIntersector;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 import static com.graphhopper.util.shapes.BBox.toEnvelope;
 
@@ -38,7 +39,7 @@ import static com.graphhopper.util.shapes.BBox.toEnvelope;
  */
 public class GraphEdgeIdFinder {
 
-    private final static int P_RADIUS = 5;
+    private static final int P_RADIUS = 5;
     private final Graph graph;
     private final LocationIndex locationIndex;
 
@@ -47,6 +48,16 @@ public class GraphEdgeIdFinder {
         this.locationIndex = locationIndex;
     }
 
+    static double calculateArea(Shape shape) {
+        if (shape instanceof BBox)
+            return calculateArea((BBox) shape);
+        if (shape instanceof Polygon)
+            return calculateArea((Polygon) shape);
+        if (shape instanceof Circle)
+            return calculateArea((Circle) shape);
+        throw new IllegalStateException("Unsupported shape: " + shape);
+    }
+    
     /**
      * @param bBox
      * @return an estimated area in m^2 using the mean value of latitudes for longitude distance
@@ -68,9 +79,10 @@ public class GraphEdgeIdFinder {
     }
 
     /**
-     * This method fills the edgeIds hash with edgeIds found inside the specified shape
+     * This method creates an edgeIds hashset with edgeIds found inside the specified shape
      */
-    public void findEdgesInShape(final GHIntHashSet edgeIds, final Shape shape, EdgeFilter filter) {
+    private GHIntHashSet findEdgesInShape(final Shape shape, EdgeFilter filter) {
+        GHIntHashSet edgeIds = new GHIntHashSet();
         locationIndex.query(shape.getBounds(), new LocationIndex.Visitor() {
             @Override
             public void onEdge(int edgeId) {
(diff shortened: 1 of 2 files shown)


karussell:
If a shape does not cover edges - maybe we should even throw an exception? (it might be too ugly?)

Did you had a chance to measure the query difference? Because this API is kind of legacy and if the difference is huge it will be critical to implement this for the new option via custom model too.

otbutz:
Screenshot_20210506_111718

otbutz:
Noticed this while benchmarking requests with a few blocked points. And e.g a point might not always cover an edge in the graph if the road is not present in OSM.

otbutz:
We could also drop these kind of shapes. There's no point to do the checks if we know that we won't find any intersections?

karussell:
Yes, either this or throwing an exception.

otbutz:
Apart from `GraphHopperTest:556` everything seems to work.

otbutz:
Shouldn't have skipped my :coffee: Readded the `continue` statement in the intersect check loop and everything is working as expected

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
```diff
@@ -165,23 +175,26 @@ public BlockArea(Graph g) {
...
 
         public String toString(int shapeIndex) {
+            GHIntHashSet blockedEdges = edgesList.get(shapeIndex);
+            if (blockedEdges == null) {
+                return Collections.emptyList().toString();
+            }
+            
             List<Integer> returnList = new ArrayList<>();
-            for (IntCursor intCursor : edgesList.get(shapeIndex)) {
+            for (IntCursor intCursor : blockedEdges) {
```
karussell, line 188: Are these changes necessary anymore? Because the edgesList contains now only HashSets that are != null and with size > 0 after your recent changes. (?)

otbutz: Yes. We either store `null` HashSets for large Shapes or non-empty ones for small shapes.

otbutz: Yes. We either store a `null` HashSet for large shapes or non-empty HashSets for smaller ones. We only drop empty HashSets for small shapes.

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
```diff
@@ -200,8 +213,8 @@ public final boolean intersects(EdgeIteratorState edgeState) {
             BBox bbox = null;
             for (int shapeIdx = 0; shapeIdx < blockedShapes.size(); shapeIdx++) {
                 GHIntHashSet blockedEdges = edgesList.get(shapeIdx);
-                // blockedEdges acts as cache that is only useful when filled and for non-virtual edges
-                if (!blockedEdges.isEmpty() && edgeState.getEdge() < baseEdgeCount) {
+                // blockedEdges acts as cache if present and is only useful for non-virtual edges
+                if (blockedEdges != null && edgeState.getEdge() < baseEdgeCount) {
```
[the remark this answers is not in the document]
otbutz: Throwing an exception for empty Sets sounds good

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
karussell: > We either store a null HashSet for large shapes

Why not an empty set where we can avoid the special `null` handling?

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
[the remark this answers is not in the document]
karussell: Hm. I just thought about this. The problem then is that users might rely on the silence here. E.g. if users draw arbitrary areas to exclude and before accepting a new blocked area they would need to query us via a route, which sounds ugly. Also it might happen that the area does work now but fails in the future because road is under construction or certain roads are excluded by default and then all routing requests with the empty block area will fail.

karussell (approved):
Thank you - LGTM!

otbutz:
@karussell the id lookup currently only works for non-virtual edges. Could we use the same mechanism as in `EdgeIdDetails` ?

```java
int edge;
if (edgeState instanceof VirtualEdgeIteratorState)
   edge = GHUtility.getEdgeFromEdgeKey(((VirtualEdgeIteratorState) edgeState).originalEdgeKey());
else
   edge = edgeState.getEdge();
```

PR #2308, opened 2021-05-07
Rewrite and fix CustomModelParser#findMaxSpeed

easbar:
Meant as a replacement for #2307.
I tried rewriting the method so it now includes LIMIT and MULTIPLY (I hope it works as good or better than the old method).

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomModelParser.java
+++ b/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomModelParser.java
@@ -471,9 +471,9 @@ public class CustomModelParser {
         for (Statement st : statements) {
             final Statement.Keyword keyword = st.getKeyword();
             if (ELSEIF.equals(keyword) && ELSE.equals(prev))
-                throw new IllegalArgumentException();
+                throw new IllegalArgumentException("After each else-statement there must be an if-statement");
             if (ELSE.equals(keyword) && ELSE.equals(prev))
-                throw new IllegalArgumentException();
+                throw new IllegalArgumentException("After each else-statement there must be an if-statement");
             if (IF.equals(keyword)) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/graphhopper/routing/weighting/custom/CustomModelParser.java
```diff
@@ -405,40 +408,79 @@ private static SimpleCompiler createCompiler(long counter, Java.AbstractCompilat
...
+            if (IF.equals(keyword)) {
+                block = new ArrayList<>();
+                result.add(block);
             }
+            block.add(st);
+            prev = keyword;
         }
-
-        return globalMin_maxSpeed;
+        return result;
```
karussell, line 484: Not sure if we should do this. We would duplicate the work of the compiler. There is no need for this code to fail here early as it would fail later.

karussell (approved):
Cool, thanks for this improvement :)

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomModelParser.java
easbar: Ok, I thought it makes it more clear to first split the statements into blocks and then find the max per block instead of doing everything in one loop. It's true we do not necessarily need to check the order of else/else_if here, but I thought this way we get a more useful error message. The downside of course would be false positives in case there is still a bug here. Don't know, please feel free to adjust the code.

karussell: Would you be ok with the latest commit?

easbar: Sure, go ahead :)

PR #2310, opened 2021-05-10
Stop using junit 4's `@Theory` in FareTest

easbar:
If we want to switch to JUnit 5 entirely (remove the JUnit 4 vintage engine) one of the harder things to migrate are usages of `@Theory`. We only use this in `FareTest` currently. I migrated the test to JUnit 5. I don't think it is an improvement for this specific test, but at least this way we can switch to JUnit 5. 

`FareTest` uses a couple of fare/trip combinations (using `@DataPoint`) that are used to run the different tests. Running tests for different paramaters can easily be done with JUnit 5 as well using `@ParameterizedTest`. I simply collect the cartesian product of all trips and fares and pass the result to the different tests.  `FareTest` then filters out some (many actually) of the data points using JUnit assumptions (also no problem with JUnit 5). 

The somewhat annoying thing is that JUnit 5 counts every input data point as single test and even counts every failed assumption like an ignored test and even logs an error like:

`org.opentest4j.TestAbortedException: Assumption failed: There are fares.`

image

If there are only a couple of different scenarios we could of course filter the input parameters already `ArgumentProvider`s so we would not need the `assume` statements. 

Otherwise it seems like the idea of property-based testing is not something the Jupiter engine wants to support and in fact there are *other* JUnit 5 test engines we could think about using: https://stackoverflow.com/questions/51543612/junit-5-theory-and-datapoint
(and of course simply keeping the vintage engine is also an option (but somewhat ugly as well).


reader-gtfs/src/test/java/com/graphhopper/gtfs/fare/FareTest.java
```diff
@@ -23,168 +23,190 @@
...
+        public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
+            List<Object[]> dataPoints = new ArrayList<>();
+            for (int i = 0; i < fares.size(); i++) {
+                Map<String, Map<String, Fare>> fare = fares.get(i);
+                for (int j = 0; j < trips.size(); j++) {
+                    Trip trip = trips.get(j);
+                    dataPoints.add(new Object[]{fare, trip, "fare " + i + ", trip: " + j});
+                }
+            }
+            return dataPoints.stream().map(Arguments::of);
```
easbar, line 127: This is just a cartesian product of all trips/fares and each tests will reject some of themm using `Assumptions`. But maybe we can also create multiple such providers that filter out certain trip/fare data points already here.

reader-gtfs/src/test/java/com/graphhopper/gtfs/fare/FareTest.java
```diff
@@ -23,168 +23,190 @@
...
+        }
     }
 
-    @Theory
-    public void irrelevantAlternatives(Map<String, Map<String, Fare>> fares, Trip trip) {
-        assumeThat("There are at least two fares.", fares.get("only_feed_id").entrySet().size(), is(greaterThanOrEqualTo(2)));
+    @ParameterizedTest(name = "{2}")
+    @ArgumentsSource(DataPointProvider.class)
+    public void irrelevantAlternatives(Map<String, Map<String, Fare>> fares, Trip trip, String displayName) {
```
easbar, line 133: The displayName parameter is only needed to make the `name = "{2}"` display name template work. Another solution could be using a `DataPoint` class as test method parameter and overwiting `toString()` there. But I did not feel like it was worth it. JUnit 5.8 will have better support here: https://stackoverflow.com/a/66994980 but it has not been released yet and again I don't think this is such a big issue here.

michaz:
I'm not opposed to this. The test does what I wanted it to do, but I was never 100% convinced this was the _best way of writing it down_, so writing it down in a different way is totally okay for me.

michaz (approved):

PR #2315, opened 2021-05-12
Speedup handling of small areas in custom models

### LINKED ISSUE #2313: Speedup handling of small areas in custom models
The current implementation defaults to costly Polygon intersection checks which requires that the tower nodes of each examined edge are loaded to create a BBox:

https://github.com/graphhopper/graphhopper/blob/0522c495ccad5f22fe19c6e858a40a7bd6f74eae/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java#L60-L65

The old BlockArea implementation preloads all edge ids for small areas which allows for faster id based filtering of non-virtual edges:

https://github.com/graphhopper/graphhopper/blob/e806a5c55750555c1c8031752b6ceae601a33eb3/core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java#L214-L239

We should implement something similar for the CustomModel handling.

otbutz:
Fixes #2313 

This PR refactors our older BlockArea logic and reuses it to enable faster intersection checks for small polygons.

* renamed BlockArea to ShapeFilter as it's not only used for blocking anymore
* create a ShapeFilter per JsonFeature and pass the filters via `CustomWeightingHelper.init()`
* adapt CustomModelParser and ExpressionVisitor to use the ShapeFilter
* pass `Graph` and `LocationIndex` into DefaultWeightingFactory and CustomModelParser in order to lookup edge ids

TODO
- [ ] allow to configure max area size

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -943,7 +943,7 @@ public class GraphHopper implements GraphHopperAPI {
     }
 
     protected WeightingFactory createWeightingFactory() {
-        return new DefaultWeightingFactory(ghStorage, getEncodingManager());
+        return new DefaultWeightingFactory(ghStorage, locationIndex, getEncodingManager());
     }
 
     @Override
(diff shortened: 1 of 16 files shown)


otbutz:
I'm not sure if it's worth to pass a parameter all the way down just to configure the max area for edge lookups.

otbutz:
Closing in favor of #2338

PR #2316, opened 2021-05-12
Remove GraphHopperAPI#load

easbar:
This method is not really used in a meaningful way. The main users of `GraphHopperAPI`:  `RouteResource` and `NavigateResource` do not use it either.


client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
```diff
@@ -125,12 +124,6 @@ public OkHttpClient getDownloader() {
         return downloader;
     }
 
-    @Override
-    public boolean load(String serviceUrl) {
-        this.routeServiceUrl = serviceUrl;
-        return true;
-    }
-
```
easbar, line 133: Any concerns about backward-compatiblity here? client-hc users now need to set the url via the constructor, but this will become apparent immediately, because there will be a compile error.

karussell: Not a problem, I think.

PR #2320, opened 2021-05-14
Use entire road geometry for NameSimilarityEdgeFilter

### LINKED ISSUE #2319: Point hint is not working as expected for curved roads
Doing a location lookup for the blue marker on the picture with `point_hint=Doubtfire` should clearly snap to `Doubtfire Crescent`, because it is quite close to the marker and it matches the `point_hint`. However, currently it snaps to `Denison Street`. 

image

The reason is that here:

https://github.com/graphhopper/graphhopper/blob/0522c495ccad5f22fe19c6e858a40a7bd6f74eae/core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java#L163-L166

we only accept edges for which the bounding lies within a circle around the query point (we use 100m radius currently). The problem is that in `GHUtility#createBBox(iter)` we only consider the tower nodes to calculate the bounding box. Since the road is curved (it has a u-like shape) this is very different to the bounding box we would obtain if we took into account the entire road geometry (pillar nodes). See the boxes of `Doubtfire Crescent` and `Denison Street` in the screenshot.

easbar:
Fixes #2319

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
+++ b/core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
@@ -175,12 +175,8 @@ public class NameSimilarityEdgeFilter implements EdgeFilter {
         // we should include the entire geometry, see #2319
         PointList geometry = edgeState.fetchWayGeometry(FetchMode.ALL);
         BBox bbox = new BBox(180, -180, 90, -90);
-        for (int i = 0; i < geometry.getSize(); i++) {
-            bbox.minLon = Math.min(bbox.minLon, geometry.getLon(i));
-            bbox.maxLon = Math.max(bbox.maxLon, geometry.getLon(i));
-            bbox.minLat = Math.min(bbox.minLat, geometry.getLat(i));
-            bbox.maxLat = Math.max(bbox.maxLat, geometry.getLat(i));
-        }
+        for (int i = 0; i < geometry.getSize(); i++)
+            bbox.update(geometry.getLat(i), geometry.getLon(i));
         return bbox;
     }
 
--- a/core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
+++ b/core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
@@ -19,9 +19,9 @@ package com.graphhopper.routing.util;
 
 import com.graphhopper.apache.commons.lang3.StringUtils;
 import com.graphhopper.debatty.java.stringsimilarity.JaroWinkler;
-import com.graphhopper.storage.NodeAccess;
 import com.graphhopper.util.EdgeIteratorState;
-import com.graphhopper.util.GHUtility;
+import com.graphhopper.util.FetchMode;
+import com.graphhopper.util.PointList;
 import com.graphhopper.util.shapes.BBox;
 import com.graphhopper.util.shapes.Circle;
 import com.graphhopper.util.shapes.GHPoint;
@@ -160,7 +160,7 @@ public class NameSimilarityEdgeFilter implements EdgeFilter {
             return false;
         }
 
-        BBox bbox = GHUtility.createBBox(iter);
+        BBox bbox = createBBox(iter);
         if (!pointCircle.intersects(bbox)) {
             return false;
         }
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -171,9 +171,17 @@ public final boolean accept(EdgeIteratorState iter) {
         return isJaroWinklerSimilar(pointHint, edgeName);
     }
 
-    private static BBox createBBox(NodeAccess na, EdgeIteratorState edgeState) {
-        return BBox.fromPoints(na.getLat(edgeState.getBaseNode()), na.getLon(edgeState.getBaseNode()),
-                na.getLat(edgeState.getAdjNode()), na.getLon(edgeState.getAdjNode()));
+    private static BBox createBBox(EdgeIteratorState edgeState) {
```
easbar, line 174: This method was no longer used since #1950, but it did the same wrong thing (exclude pillar nodes)

core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -171,9 +171,17 @@ public final boolean accept(EdgeIteratorState iter) {
...
+        // we should include the entire geometry, see #2319
+        PointList geometry = edgeState.fetchWayGeometry(FetchMode.ALL);
+        BBox bbox = new BBox(180, -180, 90, -90);
+        for (int i = 0; i < geometry.getSize(); i++) {
+            bbox.minLon = Math.min(bbox.minLon, geometry.getLon(i));
+            bbox.maxLon = Math.max(bbox.maxLon, geometry.getLon(i));
+            bbox.minLat = Math.min(bbox.minLat, geometry.getLat(i));
+            bbox.maxLat = Math.max(bbox.maxLat, geometry.getLat(i));
+        }
```
easbar, line 183: Finding the bounds of a point list is  maybe something we do elsewhere already and we could create a helper method for this otherwise. Unfortunately I could not find something like it. 

Btw. why has PointList `getSize()` and `size()`? And looking at point list I was not even sure if `size()` really returns the number of elements or rather the capacity?!

easbar: Maybe we should store the bounds of each edge including the geometry for such use cases. There are two very similar todos here: 

https://github.com/graphhopper/graphhopper/blob/e806a5c55750555c1c8031752b6ceae601a33eb3/core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java#L226-L228

https://github.com/graphhopper/graphhopper/blob/0522c495ccad5f22fe19c6e858a40a7bd6f74eae/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java#L61-L63

core/src/main/java/com/graphhopper/routing/util/NameSimilarityEdgeFilter.java
```diff
@@ -171,9 +171,17 @@ public final boolean accept(EdgeIteratorState iter) {
         return isJaroWinklerSimilar(pointHint, edgeName);
     }
 
-    private static BBox createBBox(NodeAccess na, EdgeIteratorState edgeState) {
-        return BBox.fromPoints(na.getLat(edgeState.getBaseNode()), na.getLon(edgeState.getBaseNode()),
-                na.getLat(edgeState.getAdjNode()), na.getLon(edgeState.getAdjNode()));
+    private static BBox createBBox(EdgeIteratorState edgeState) {
+        // we should include the entire geometry, see #2319
+        PointList geometry = edgeState.fetchWayGeometry(FetchMode.ALL);
```
otbutz, line 176: I've seen this usage pattern a few times now. We create a PointList only to do some logic with each point and throw it away later. Wouldn't it be better to be able to consume them on-the-fly with a visitor pattern?

e.g `edgeState.visitWayGeometry(FetchMode.ALL, visitor);`

easbar: Yes, maybe. But for this case I wonder if we should simply store the bounding box for each edge.

otbutz: > I wonder if we should simply store the bounding box for each edge.

My europe import logs roughly 92 million edges. That's about 2.8GB of bbox information if we stored minlon,minlat,maxlon,maxlat per edge?

easbar: Yes, if we used four doubles (4*8=32bytes). But that would be very wasteful. We can definitely reduce precision there (4*4bytes is more than enough), and maybe we can even just store the indices of the point list. Then maybe 4bytes per edge is enough.

easbar: ... and we could probably even somehow restrict this to 'curved' edges only, because for straight edges using the tower node coordinates as bounding box is often sufficient.

otbutz: Implemented visitor access: https://github.com/graphhopper/graphhopper/pull/2321

easbar:
I had to fix the routing measurement for point_hint, because so far we always set the point_hint to the name of the very first edge. But to test the speed of the look-ups with point hints I think we should rather use some point hint that cannot be found (worst case scenario in terms of performance). I fixed this here: b8a25d775b4578b9e37912736cfbccdd49ea9f25 and it changed the measured routing CH speed with point hints from about 1.5ms to 1.8ms, so we can probably say that point_hints make route calculation (including location lookup, but no instructions, serialization etc.) about 20% slower.

After fixing this I compared the performance (`metrics.routingCH_with_hints.mean`) of this PR vs. master:

|map|master|this branch|
|-|-|-|
|germany|1.78ms|1.83ms|
|bayern|1.19ms|1.21ms|

The difference is less than 3% even on the smaller map where I assume the relative amount of time spent to lookup the locations is larger (because the routes are shorter). I think we can conclude that the bbox/geometry check does not play an important role and not too surprisingly the majority of the additional time we need to use `point_hint` is spent on the string comparison checks.

Btw I also did a quick performance comparison for the current scenario where we only use the tower nodes. Not using a `PointList` via `fetchWayGeometry` and instead accessing the tower node coordinates directly like `nodeAccess.getLat(iter.getBaseNode())` did not make a significant difference here either, <2%.

PR #2324, opened 2021-05-17
Remove junit vintage engine, no more JUnit 3/4

easbar:
After replacing `@Theory` (#2310) and class-wide parameters (#2312) moving from JUnit 4 to 5 entirely is 'only' a matter of changing the import statements and assertion method calls and I did this here.


karussell:
Oh, wow. Really nice - didn't notice this :) :+1:

AnLogLabster:
I saw the earlier thread but just hadn't had time to refresh my memory on JUnit in order to mount a meaningful response, but it seems like you've got it all covered 👍

Anvoker:
I saw the earlier thread but just hadn't had time to refresh my memory on JUnit in order to mount a meaningful response, but it seems like you've got it all covered 👍

easbar:
Thanks a lot four your initial work on this @Anvoker! Can I add you to the contributors list?

Anvoker:
If you think that's enough to warrant it, then sure thing! Don't see why not.

easbar:
I just saw you are already on the list :) But anyway I updated it here: d141dbde06c3d1d813c71d3204258b5b7a3fca15

PR #2326, opened 2021-05-18
Use bbox intersection logic for rectangles

otbutz:
If a polygon describes a rectangle, we don't need to load the way geometry to check for an intersection.
This should be especially beneficial for large areas, for which we previously had to load all intersecting edge geometries.


otbutz:
@karussell any objections?

karussell:
Looks good!

(There could be rare problems with this approach in general, when the bbox of the tower nodes is smaller than the bbox of the entire geometry, but this is not a new problem.)

PR #2321, opened 2021-05-14
Use visitor to access way geometry

otbutz:
As discussed in #2320

PR #2327, opened 2021-05-18
Reuse edge bbox and geometry for multiple areas

otbutz:
Solves the remaining `TODO` by caching the edge bbox and way geometry for multiple areas.

TODO: combine with #2321 to reset und update the cached bbox which would eliminate most object creation


karussell:
Currently we do not store edge state inside the Weighting and I would like to keep it this way, except there is a huge benefit. What is the speed up in your measured cases?

otbutz:
Hamburg(53.548416,10.003159) to Berlin(52.517997,13.396111) using flexible mode with a dummy custom model targeting 4 areas.

## Test 1: Areas which don't intersect any edges

```json
{
    "priority": [
     {
      "if": "in_a || in_b || in_c || in_d",
      "multiply_by": 0.0
     }
    ],
     "areas": {
      "a": {
       "type": "Feature", 
       "properties": {},
       "geometry": { 
         "type": "Polygon", 
         "coordinates": [[
          [ 14.31, 51.75 ],
          [ 14.34, 51.75 ],
          [ 14.34, 51.77 ],
          [ 14.31, 51.77 ],
          [ 14.31, 51.75 ]
         ]]
       }
      },
      "b": {
       "type": "Feature", 
       "properties": {},
       "geometry": { 
         "type": "Polygon", 
         "coordinates": [[
          [ 14.31, 51.75 ],
          [ 14.34, 51.75 ],
          [ 14.34, 51.77 ],
          [ 14.31, 51.77 ],
          [ 14.31, 51.75 ]
         ]]
       }
      },
      "c": {
       "type": "Feature", 
       "properties": {},
       "geometry": { 
         "type": "Polygon", 
         "coordinates": [[
          [ 14.31, 51.75 ],
          [ 14.34, 51.75 ],
          [ 14.34, 51.77 ],
          [ 14.31, 51.77 ],
          [ 14.31, 51.75 ]
         ]]
       }
      },
      "d": {
       "type": "Feature", 
       "properties": {},
       "geometry": { 
         "type": "Polygon", 
         "coordinates": [[
          [ 14.31, 51.75 ],
          [ 14.34, 51.75 ],
          [ 14.34, 51.77 ],
          [ 14.31, 51.77 ],
          [ 14.31, 51.75 ]
         ]]
       }
      }
    }
   }
```

Unpatched: 370ms
Patched: 320ms

Diff: -14.5%


## Test 2: Areas intersecting with edges

Same model except for the coordinates:
```
         "coordinates": [[
          [ 11.19, 52.92 ],
          [ 12.35, 52.92 ],
          [ 12.35, 53.64 ],
          [ 11.19, 53.64 ],
          [ 11.19, 52.92 ]
         ]]
```

Unpatched: 1280ms
Patched: 1060ms

Diff: -20%

otbutz:
Note that the second test should also be a lot faster with #2326 as we don't need to load any edge geometries to detect intersections.

karussell:
Under these circumstance I would not want to introduce this caching (yet).

otbutz:
> Under these circumstance I would not want to introduce this caching (yet).

Should we cache the bbox in the EdgeIteratorState? That would also allow us to avoid PointList creation as we could use NodeAccess to create the BBox directly.

otbutz:
Implemented bbox caching as part of the EdgeIteratorState: https://github.com/graphhopper/graphhopper/compare/master...otbutz:iter_bbox

The biggest benefit is that we're able to reuse the BBox instance for all visited edges. This completely avoids object creation and brings my second test case down to about 940ms which is a whooping 30% faster than our current master.

otbutz:
Closing in favor of #2338

PR #2329, opened 2021-05-18
Set road_class=motorway when motorroad=yes

easbar:
The motorroad tag is documented here: https://wiki.openstreetmap.org/wiki/Key:motorroad

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - added smoothness encoded value, used to determine bike speed (#2303)
 - migrated tests from junit 4 to 5 (#2324)
+- changed handling of motorroad tag (#2329)
 
 ### 3.0 [17 May 2021]
 
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadClassParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadClassParser.java
@@ -49,6 +49,10 @@ public class OSMRoadClassParser implements TagParser {
         String roadClassTag = readerWay.getTag("highway");
         if (roadClassTag == null)
             return edgeFlags;
+        if (!roadClassTag.isEmpty() && readerWay.hasTag("motorroad", "yes")) {
+            roadClassEnc.setEnum(false, edgeFlags, RoadClass.MOTORWAY);
+            return edgeFlags;
+        }
         RoadClass roadClass = RoadClass.find(roadClassTag);
         if (roadClass == OTHER && roadClassTag.endsWith("_link"))
             roadClass = RoadClass.find(roadClassTag.substring(0, roadClassTag.length() - 5));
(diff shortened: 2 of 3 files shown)


otbutz:
> The motorroad tag is used to describe highways that have motorway-like access restrictions (i.e. cyclists and pedestrians banned) **but that are not a motorway.**

boldtrn:
IMHO this should be stored in a separate encoded value to handle things like access. But for example max speed might differ.

easbar:
If motorroad=yes was a separate encoded value would you use it to model access and speed? And if yes how?

boldtrn:
Sorry I am a bit late with my reply @easbar.

For example in Germany there are some highway=primary with motorroad=yes. 

For cars I guess the average speed might be higher on these roads.

For bicycles, foot, etc. access should be blocked (or made very unlikely).

easbar:
> For cars I guess the average speed might be higher on these roads.

We use the `motorroad` tag to determine the average speed (via the average_speed encoded value), e.g. here: 

https://github.com/graphhopper/graphhopper/blob/0522c495ccad5f22fe19c6e858a40a7bd6f74eae/core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java#L109-L115

> For bicycles, foot, etc. access should be blocked (or made very unlikely).

We do this via the access encoded value:

https://github.com/graphhopper/graphhopper/blob/8ace2228fb9cfd9cbd089184df8e3141bd490c56/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java#L262-L263

Changing the road_class for motorroad=yes does not affect access and speed encoded values. But of course if motorroad was a separate EV we could use it in more sophisticated ways via custom models. That's why I was wondering what kind of custom models that use such a motorroad encoded value people would come up with.

boldtrn:
Yes for example this could help when developing a custom model for Mofa or Scooter. 

One other idea might be if you have a vehicle where you want to avoid complicated turns (like an RV), you might want to prefer motorroads.

I think it might be helpful for custom models.

easbar:
Ok yes it is probably helpful, but at least I am just speculating, I don't know what we would actually do with motorroad in a custom model. The meaning of the motorroad tag is not entirely clear to me at also seems to be country-dependent. That's why I was wondering if you or anyone can already tell me how exactly it could be used (and especially why it definitely should *not* simply be used to change the road_class like in this PR).

boldtrn:
If you look at the highway=primary motorrad=yes, they are different from motorways. Also different from trunk as well (but less different). So for me these are different types of road. For example when I look at the path details and GraphHopper would return trunk/motorway for a primary that would feel wrong.

On the other hand, why is OSM not just tagging these as highway=trunk :).

> also seems to be country-dependent

Yes, IMHO that's one more reason not do something that might make sense for Country A, but might be wrong for Country B :).

> Can anyone give an example where we want to distinguish between e.g. road_class == trunk && motorroad=yes and road_class == motorway?

There are probably cases where they are different. Maybe toll calculation/avoidance? I guess speed is probably not the same, motorways will be higher. If you look at: https://wiki.openstreetmap.org/wiki/Key:motorroad most countries use motorroad for access restrictions but don't assume any max speed changes (only min speed).

I think they should not be the same. I am not sure if anyone would want to use in a custom model, my guess is yes, but probably not a high prio tag for custom models.

otbutz:
> Probably we should ask this on the tagging list?

Definitely.

easbar:
This is somewhat related to #2692

PR #2330, opened 2021-05-19
Maps: make export link working with custom model

karussell:
With this PR you can easily share GH links with your custom models


ratrun:
Could you please explain how the user is supposed to trigger this functionality? I would expect that if I add a custom rule in the web UI and press search that the custom rule parameters magically make its way into the URL. But this is not the case for me.

karussell:
Should have explained this better :) 

We cannot make this working via updating the URL, otherwise every key pressed would end up in the URL history. Instead you click on the chain icon when you are ready:

image

ratrun:
Thank you, I never used that symbol before. Very handy!

PR #2332, opened 2021-05-24
Allow other profiles than car for isochrones.

ratrun:
Currently the profile string is hardcoded in the web ui. This change allows to modify it via the web UI.


easbar:
Thanks, this is definitely useful. Could we turn the input box into a selection box and query the available profiles via the `/info` endpoint of the server?

ratrun:
@easbar I have tried it in e15af0c and now it works for me. Please note that I'm not familiar with js, so there are for sure better ways to do this.

easbar:
Thanks :+1: It's actually quite fun to take a look at the same isochrone for different profiles :)

Btw I noticed something interesting: The isoline for bike with fastest weighting covers a bigger area than the one for bike with shortest weighting?!

image
image

PR #2333, opened 2021-05-27
Include priority EncodedValue in CustomWeighting

### LINKED ISSUE #2292: custom requests doesn't consider `priority` EncodedValue
A request against bike with weighting=custom should result in the same response like with bike and weighting=fastest. Currently the custom model only considers `vehicle$access` and `vehicle$average_speed`. But for bike, foot and motorcycle the EncodedValue `encoder$priority` should be considered.

Current workaround could be:

```
"priority": [{
  "if": "bike$priority < 0.2",
  "multiply_by": 0.2
},{
  "else if": "bike$priority < 0.5",
  "multiply_by": 0.5
},{
  "else if": "bike$priority < 0.7",
  "multiply_by": 0.7
}]
```

To make this simpler we could allow an EncodedValue as value e.g.

```
"priority": [{
  "if": true,
  "multiply_by": "bike$priority"
}]
```

We would not be able to support this at query time (only server-side) as this is per-edge and could make the LM incorrect even if we would know e.g. the maximum.

karussell:
Fixes #2292.

Instead of introducing a new API with "set_to" or a "value expressions" (which we will likely introduce later, see [discussion](https://discuss.graphhopper.com/t/custom-rule-question-regarding-relationship-to-existing-flag-encoders/6539/7?u=karussell)) it is more consistent to use the `encoder$priority` in CustomWeighting (additionally to `encoder$average_speed` and `encoder$access`) if available. It is used here as the start value for priority instead of 1.

To do this I moved the 0.5 into the encoder (to avoid this arbitrary number in CustomWeighting) and had to change the factor as well as to increase the bits used for `encoder$priority`. This resulted in only some minor test changes. See RoutingAlgorithmWithOSMTest.java and GraphHopperTest.java where I will investigate if the new routing results are still reasonable.

The advantage is that with the `encoder$priority` we can now also "hard exclude" certain edges not only "soft avoid" them.

 - [x] use #2335 here
 - [x] check again: racingbike should be faster than bike for
    http://localhost:8989/maps/?point=43.730864%2C7.420771&point=43.727687%2C7.418737&profile=racingbike (shows ugliness only without elevation) -> this seems to be a different issue: https://discuss.graphhopper.com/t/6554


core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
```diff
@@ -208,10 +207,8 @@ public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, EncodingManager.A
             setSpeed(true, edgeFlags, ferrySpeed);
         }
 
-        priorityWayEncoder.setDecimal(false, edgeFlags, PriorityCode.getFactor(handlePriority(way)));
-
-        // Set the curvature to the Maximum
-        curvatureEncoder.setDecimal(false, edgeFlags, PriorityCode.getFactor(10));
+        priorityWayEncoder.setDecimal(false, edgeFlags, PriorityCode.getValue(handlePriority(way)));
+        curvatureEncoder.setDecimal(false, edgeFlags, 10.0 / 7.0);
```
easbar, line 211: Previously we had `PriorityCode.getFactor(10)=10/10.0=1`, but now we have `10.0/7.0`=1.43...`? Should this not be `1.0+0.5=1.5` or `PriorityCode.getValue(BEST)`?

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
```diff
@@ -24,14 +24,14 @@
...
-    VERY_NICE(6),
-    BEST(7);
+    WORST(5),
+    AVOID_AT_ALL_COSTS(6),
+    REACH_DEST(8),
+    AVOID_IF_POSSIBLE(9),
+    UNCHANGED(10),
+    PREFER(12),
+    VERY_NICE(13),
+    BEST(15);
```
easbar, line 34: Maybe off-topic, but where do these names come from? `AVOID_AT_ALL_COSTS` sounds like its entirely forbidden, but it is still allowed and `WORST` is even more not allowed (but still not forbidden) :) Could we now also set it to `0`? Since we use four bits I would assume all values from 0 to 15 are possible?

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
```diff
@@ -42,11 +42,11 @@ public int getValue() {
         return value;
     }
 
-    /**
-     * This method returns the PriorityCode.value in a range between 0 and 1 suitable for direct usage in a Weighting.
-     */
-    public static double getFactor(int val) {
-        return (double) val / BEST.getValue();
+    public static double getFactor(int value) {
+        return (double) value / 10.0;
```
easbar, line 46: Maybe put `value / UNCHANGED` here ?

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java
```diff
@@ -57,6 +59,14 @@ protected final double getRawSpeed(EdgeIteratorState edge, boolean reverse) {
         return speed;
     }
 
+    protected final double getRawPriority(EdgeIteratorState edge, boolean reverse) {
+        if (priority_enc == null) return 1;
+        double priority = reverse ? edge.getReverse(priority_enc) : edge.get(priority_enc);
+        if (Double.isInfinite(priority) || Double.isNaN(priority) || priority < 0)
+            throw new IllegalStateException("Invalid priority " + priority);
```
easbar, line 66: Can we / should we also check `priority < 1.5` here? And don't we have to adjust `CustomWeighting#getMinWeight` as well, now that priority can be max. `1.5` in case we do not apply any rules in custom model?

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
karussell: > Could we now also set it to 0? 

We should probably change/refactor the name and introduce new codes. But just setting them to different values will change routing behaviour where it is not intended.

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
karussell: IMO this is unrelated to the PriorityCode itself and only how we store them.

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java
karussell: The priority above 1 was possible for server-side profiles in versions before (3.0) so I consider this a different issue.

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
karussell: I could think about:

```
    EXCLUDE(0),
    VERY_BAD(3),
    BAD(5), // WORST
    REACH_DEST(6), // AVOID_AT_ALL_COSTS
    AVOID(8), // REACH_DEST
    AVOID_IF_POSSIBLE(9),
    UNCHANGED(10),
```

I'm unsure if we should find a name for every number. If values between are really required then they could be used directly?

karussell: It seems the naming is taken from this OSM wiki: https://wiki.openstreetmap.org/wiki/Key:class:bicycle#Values

core/src/main/java/com/graphhopper/routing/util/MotorcycleFlagEncoder.java
easbar: Sorry, my bad. You are right it was `10.0/7.0` also before and makes sense not to use `PriorityCode` for curvature.

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
easbar: > But just setting them to different values will change routing behaviour where it is not intended.

No, I did not mean to set different values. I was just wondering if it is now possible to set 0 as well.

> We should probably change/refactor the name and introduce new codes, yes

> Something like this?

Looks better yes.

Ok, or maybe use no names at all and just the value?

> I'm unsure if we should find a name for every number. If values between are really required then they could be used directly?

Yes, me neither :) Also 'best' and 'worst' sounds a bit confusing to me, I rather think of 'low' and 'high' priority.

> It seems the naming with 7 values is taken from this OSM wiki

Ok, I don't see much benefit in using these names, but we can also just leave it as is.

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java
easbar: Ok, but now that CustomWeighting does use priority and now that priority can be 1.5 I think we have to adjust `CustomWeighting#getMinWeight` otherwise the routing could be wrong, don't you think?

karussell: Have created #2334

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
easbar: Yes maybe just use the value directly? And maybe keep a constant for the max (15)? Not sure either way is fine for me, we can also leave this as it is, because it is not directly related to this issue.

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java
easbar: Ok thanks :+1:. Before this PR we did not have to take into account priority in `CustomWeighting#getMinWeight`, because CustomWeighting simply ignored the `vehicle$priority` values and priority was always `<1`, right? With this PR it should be easy to construct a test case that produces connection not found if we set `Priority.BEST`?

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
karussell: When thinking about this a bit then I would keep the enum values (as they also describe the intend a bit) although the current naming is indeed a bit confusing. Maybe @ratrun has an idea/opinion how to handle the increased possible values?

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java
karussell: > and priority was always <1, right?

Before this PR when I created a JSON that increased the priority above 1 then this was also a problem.

karussell: > With this PR it should be easy to construct a test case that produces connection not found if we set Priority.BEST?

It will find the connection, I think, but could give a suboptimal route as the heuristic with A* is no longer underestimate

easbar: > Before this PR when I created a JSON that increased the priority above 1 then this was also a problem.

Ok in the server-side config producing this problem was also possible, right. 

> It will find the connection, I think, but could give a suboptimal route as the heuristic with A* is no longer guaranteed to be underestimating

I think with edge-based CH AStar it is also possible to get connection not found, but I will try to come up with a test case to proof this :)

And fixing this is actually quite easy, right? We just need `encoder#getMaxPriority` like `encoder#getMaxSpeed` and need to use this in `getMinWeight()`? But sure we can also do this in a follow-up PR.

karussell: > I think with edge-based CH AStar it is also possible to get connection not found, but I will try to come up with a test case to proof this :)

Aha, ok. No need to proof this - I just believe you :D ... I think finding those problems is rather time consuming and not sure if worth the effort :) 

> And fixing this is actually quite easy, right? We just need encoder#getMaxPriority like encoder#getMaxSpeed and need to use this in getMinWeight()? But sure we can also do this in a follow-up PR.

Yes, see https://github.com/graphhopper/graphhopper/pull/2335 :) (where we would replace the 1 with priorityEnc.getMax or something to integrate it here)

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
ratrun: I took the original naming from https://wiki.openstreetmap.org/wiki/DE:Key:class:bicycle. With 3 bits available this was kind of logical. Now that there is one more bit available I would like to use the most backward compatible solution.

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java
easbar: > Aha, ok. No need to proof this - I just believe you :D ... I think finding those problems is rather time consuming and not sure if worth the effort :)

Ah I think actually quite the opposite. I probably spent too much time thinking and talking about all the possibilities and it is more efficient  (and sustainable) to start with a failing test to identify the (or one of the) problems :)

>  Yes, see #2335 :) (where we would replace the 1 with priorityEnc.getMax or something to integrate it here)

:+1:

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
karussell: Ok, will try to come up with a better naming that does not use the old names so that not different values are used accidentially. What I like more about the new range is that we are able to exclude certain edges and that UNCHANGED is really keeping the value where before it was a "slight prefer".

core/src/test/java/com/graphhopper/routing/PriorityRoutingTest.java
```diff
@@ -0,0 +1,126 @@
...
+import com.graphhopper.util.EdgeIteratorState;
+import com.graphhopper.util.PMap;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class PriorityRoutingTest {
+
+    @Test
+    void testMaxPriority() {
```
easbar, line 49: This test fails because CustomWeighting does not consider the priority in `getMinWeight` yet, but should be fixed by #2335

karussell: Yes, this works when #2335 is merged and `priorityEnc.getMaxDecimal()` is used for findMaxPriority

karussell: Works now

core/src/main/java/com/graphhopper/routing/weighting/PriorityWeighting.java
```diff
@@ -53,6 +52,9 @@ public double calcEdgeWeight(EdgeIteratorState edgeState, boolean reverse) {
         double weight = super.calcEdgeWeight(edgeState, reverse);
         if (Double.isInfinite(weight))
             return Double.POSITIVE_INFINITY;
-        return weight / (0.5 + edgeState.get(priorityEnc));
+        double priority = edgeState.get(priorityEnc);
+        if (priority > 1.5)
+            throw new IllegalArgumentException("priority cannot be bigger than 1.5 but was " + priority);
```
easbar, line 57: Can we use `PriorityCode.getFactor(BEST.getValue())` also here?

easbar:
Looks good :+1:

core/src/main/java/com/graphhopper/routing/weighting/PriorityWeighting.java
karussell: Sure

PR #2338, opened 2021-06-02
Cache tower node bbox in EdgeIteratorState

otbutz:
* allows to check for tower node bbox intersection via the EdgeIteratorState
* caches the created BBox to speedup handling of multiple area checks
* resets the BBox within `EdgeIteratorImpl.goToNext()` to avoid recreating it for every visited edge

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/querygraph/VirtualEdgeIteratorState.java
+++ b/core/src/main/java/com/graphhopper/routing/querygraph/VirtualEdgeIteratorState.java
@@ -124,7 +124,7 @@ public class VirtualEdgeIteratorState implements EdgeIteratorState {
         if (pointList.isEmpty())
             return false;
         
-        return shape.intersects(pointList);
+        return shape.intersects(pointList.clone(false).makeImmutable());
     }
 
     @Override


core/src/main/java/com/graphhopper/routing/querygraph/VirtualEdgeIteratorState.java
```diff
@@ -117,6 +119,17 @@ else if (mode == FetchMode.PILLAR_ONLY) {
     public EdgeIteratorState setWayGeometry(PointList list) {
         throw new UnsupportedOperationException("Not supported for virtual edge. Set when creating it.");
     }
+    
+    @Override
+    public ReadableBBox getTowerBBox() {
```
otbutz, line 124: It's probably not worth to cache the bbox for virtual edges

otbutz:
@karussell the changes proposed here should be less intrusive than those in #2327

karussell:
Although we do some caching already in EdgeIteratorState (`flags`) I would not want this (yet). I can see the requirements but I can also see the big problems with caching. Also this change requires a separate interface (and when is the cached value reset - shouldn't it be done when setWayGeometry is called?).

The least invasive change could be probably to cache the PointList in the EdgeIteratorState somehow or store the bbox like proposed in https://github.com/graphhopper/graphhopper/pull/2320#discussion_r632416236. But even if this would be possible I (currently!) do not see the need to have this cached. E.g. we currently should focus on making longer route requests much faster with simple custom models (avoid motorway, ferry, ...) instead of making routing a bit faster for specific use cases (many `areas`).

otbutz:
> shouldn't it be done when setWayGeometry is called?

Good catch. Currently the reset happens here:

https://github.com/graphhopper/graphhopper/pull/2338/files#diff-267f44bcfcb3b930af5a645649c8c73fed2c7ecfc2a1998d6ce4af276f800de5R941-R943

> The least invasive change could be probably to cache the PointList in the EdgeIteratorState

Gave this one a little bit of thought but we need the full way geometry pretty late as the bbox check is sufficient for most examined edges.

> or store the bbox like proposed in [#2320 (comment)](https://github.com/graphhopper/graphhopper/pull/2320#discussion_r632416236).

Depends on which kind of bbox we're talking about. The bbox for the complete way geometry is a lot more taxing. The one of the tower nodes is rather cheap but we still suffer from having to create `visited edges * areas` BBox instances for each run.

> specific use cases (many `areas`).

The caching is already notable for a single area as we create and modify a single BBox instead of a BBox *per edge*.

otbutz:
@karussell i gave it a bit of thought lately and we could skip all of the `ReadableBBox` stuff i've we do the intersection check within the EdgeIteratorState. e.g:

```java
boolean intersectsTowerBBox(BBox bbox);
```

otbutz:
@karussell this version should be a lot less invasive and keeps the cached BBox within the iterator.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -1220,6 +1221,26 @@ public EdgeIteratorState setWayGeometry(PointList pillarNodes) {
...
+            
+            if (!towerBBox.intersects(shape.getBounds())) {
+                return false;
+            }
+            
+            if (shape instanceof Polygon && ((Polygon) shape).isRectangle()) {
+                return true;
+            }
+            
+            return shape.intersects(fetchWayGeometry(FetchMode.ALL).makeImmutable());
```
otbutz, line 1242: We could also cache the way geometry here and drop it when onNext() or setWayGeometry() is called.

otbutz:
@karussell any chance to get this merged?

Issue #407, opened 2015-05-08
Bicycle/MTB routing should observe bicycle=yes on a barrier


arminus:
Presently, a node with a tag barrier=kissing_gate and bicycle=yes blocks bicycle or mtb routing through that node. 

IMHO, a tag bicycle=yes should override any barrier on a node (same goes for stile, turnstile, etc). Plus, the bicycle=yes tag should allow routing for all "bike-types".

karussell:
@NopMap and @ratrun can you confirm this change proposal?

ratrun:
I fully support the propsual. It makes sense, especially for vehicle type mountainbike. After having read the definition for a kissing_gate barrier in the wiki I would even allow it for the mounainbike vehicle case also without the "bicycle=yes" tag combination. I would implement this in one step, this probably shouldn't be too difficult.
When we get the possibility to add costs for nodes (#75 ??) , the cost for such a barrier should be quite high, because you definetly have to get off the bike.
Additionally the instructions should eventually tell something like "Get off the bike to pass barrier". But this is another step. After a short glimpse into the code I don't think that this feature is already implemented for other types of barrieres.

ratrun:
I prepared a patch at https://github.com/ratrun/graphhopper/commit/731422eef37cc7b4b30229b3cd20f21f857278f5
which removes the absolute `kissing_gate` barrier for the mountainbike encoder. For other bike encoders this barrier remains absolute.

karussell:
Thanks, looks good! Would you create a PR from this commit?

arminus:
I think the current implementation is too limiting. If there's a tag bicycle=yes on a kissing_gate barrier, it should allow all bike types, not only mountain bikes. Here's an overpass query which shows 34 such barriers in Munich - I seriously doubt one needs an MTB there ;-)

http://overpass-turbo.eu/s/9su

karussell:
(reopened this issue due to this and the discussion on the mailing)

ratrun:
Could you please provide a picture of a kissing_gate which can be passed easily by bike and where you put would set the tag `biclycle=yes`?

Here is the "reference" picture from the OSM wiki: http://wiki.openstreetmap.org/w/images/thumb/7/7b/TL0452_stile.jpg/300px-TL0452_stile.jpg

I doubt that the 34 kissing_gates in Munich are tagged correctly. 

[...]

PR #2340, opened 2021-06-11
Allow barrier kissing_gate with bicycle=yes for all bicycle profiles,…

ratrun:
… see discussion in #407.

I additionally moved the turn_stile from the list of absolute barriers to the potentials.

### DIFF
--- a/core/src/test/java/com/graphhopper/routing/util/AbstractBikeFlagEncoderTester.java
+++ b/core/src/test/java/com/graphhopper/routing/util/AbstractBikeFlagEncoderTester.java
@@ -64,7 +64,7 @@ public abstract class AbstractBikeFlagEncoderTester {
         IntsRef relFlags = encodingManager.handleRelationTags(rel, encodingManager.createRelationFlags());
         IntsRef edgeFlags = encodingManager.handleWayTags(way, new EncodingManager.AcceptWay().put(encoder.toString(), WAY), relFlags);
         DecimalEncodedValue enc = encodingManager.getDecimalEncodedValue(EncodingManager.getKey(encoder.toString(), "priority"));
-        assertEquals((double) expectedPrio / BEST.getValue(), enc.getDecimal(false, edgeFlags), 0.01);
+        assertEquals(PriorityCode.getValue(expectedPrio), enc.getDecimal(false, edgeFlags), 0.01);
         return edgeFlags;
     }
 
@@ -283,7 +283,7 @@ public abstract class AbstractBikeFlagEncoderTester {
 
         osmWay.setTag("highway", "secondary");
         osmWay.setTag("tunnel", "yes");
-        assertPriority(AVOID_AT_ALL_COSTS.getValue(), osmWay);
+        assertPriority(AVOID_MORE.getValue(), osmWay);
 
         osmWay.setTag("bicycle", "designated");
         assertPriority(PREFER.getValue(), osmWay);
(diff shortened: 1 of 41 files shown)


karussell:
Thanks!

core/src/test/java/com/graphhopper/routing/util/AbstractBikeFlagEncoderTester.java
```diff
@@ -410,6 +410,24 @@ public void testBarrierAccess() {
         node.setTag("bicycle", "yes");
         // no barrier!
         assertTrue(encoder.handleNodeTags(node) == 0);
+
+        // Issue 407
+        // kissing_gate without bicycle tag
+        node.setTag("barrier", "kissing_gate");
+        // barrier!
+        assertFalse(encoder.handleNodeTags(node) == 0);
```
karussell, line 418: Can you have a look - this test seems to fail for mtb and racingbike according to travis: https://travis-ci.org/github/graphhopper/graphhopper/jobs/774275860
Also down on line 425 for bike and bike2

ratrun: I'm sorry for the inconvenience. I had accidentially disabled the tests and therefore didn't notice :disappointed:.  bba2e67 is a bit more complex now, but does the intended blocking of `barrier=kissing_gate` without `bicycle=yes` except for the mountainbike encoder, where  this is allowed.

karussell: Hmmh, this is indeed a bit involved. As we run `blockByDefault==false` in production (see our reasoning for this [here](https://forum.openstreetmap.org/viewtopic.php?id=58238)) would it make the code easier if that is enabled in tests too?

ratrun: Sorry, I'm a bit confused now. `blockByDefault` is false for all bicycle encoders, also for the tests. So I don't see how setting it to false once more would simplify the code.
The only change which would simplify the code (and the tests as well) is allowing of a barrier `kissing_gate` per default, without the addition of `bicycle=yes`. But I don't like to do this.

karussell: The currently handling with the blockByDefault is already not that easy to understand and introducing another concept with different defaults makes it harder to reason about the behaviour. Also having the 3rd absoluteBarriers list is a bit inflexible as tagging cannot influence access.

My proposal: create another pull request where we only have two lists:

 * one is collecting all barriers with a pass-by-default behaviour (passByDefaultBarriers) and 
 * one with a block-by-default behaviour (blockByDefaultBarriers) (basically this is the previous absoluteBarriers)

Would this make sense to you too? 

There or in another issue we should also think about removing the blockByDefault config parameter. E.g. in production we already use `block_barriers=false` even for car.

ratrun: Thank you, this makes sense! I have started with the proposed PR, but I'm facing a couple of car specific high level tests which are failing now. I'll need some help to fix some of them. I'll comment and mark these if this is OK.

karussell: Sure, will do - thank you!

karussell:
Would you mind updating this PR due to the merged changes in #2345?

ratrun:
Commit 96c0e1b contains all changes I had in mind. Please check. With this change the result of the discussion in #407 should be implemented.

karussell:
LGTM - thanks!

ratrun:
Could you please finally close #407?

karussell:
I would wait until it is deployed - should happen in the next 2 days.

PR #2344, opened 2021-06-17
Replace absoluteBarriers with blockByDefaultBarriers and replace potentialBarriers with passByDefaultBarriers

ratrun:
Removal of configure option "blockByDefault".
This is the result of the discussion in #2340.
The high level car testcases need to be checked, especially those 3 marked with "FIXME to be checked!!!"
Please note that this only provides the basis for the implementation of #407.


karussell:
Sorry, I usually know how to fetch foreign branches but replacing this PR was a bit faster ;). See #2345.

PR #2345, opened 2021-06-19
make absolute barriers passable

karussell:
Replaces #2344

 - make absolute barriers passable (now called `blockByDefaultBarriers`)
 - car now also allows passing "potential barriers" by default (those barriers are now called `passByDefaultBarriers`)
 - removal of configuration option "blockByDefault".

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
@@ -71,29 +71,28 @@ public class CarFlagEncoder extends AbstractFlagEncoder {
 
         blockPrivate(properties.getBool("block_private", true));
         blockFords(properties.getBool("block_fords", false));
-        blockBarriersByDefault(properties.getBool("block_barriers", true));
         setSpeedTwoDirections(properties.getBool("speed_two_directions", false));
 
         intendedValues.add("yes");
         intendedValues.add("designated");
         intendedValues.add("permissive");
 
-        potentialBarriers.add("gate");
-        potentialBarriers.add("lift_gate");
-        potentialBarriers.add("kissing_gate");
-        potentialBarriers.add("swing_gate");
-        potentialBarriers.add("cattle_grid");
-        potentialBarriers.add("chain");
+        passByDefaultBarriers.add("gate");
+        passByDefaultBarriers.add("lift_gate");
+        passByDefaultBarriers.add("kissing_gate");
+        passByDefaultBarriers.add("swing_gate");
+        passByDefaultBarriers.add("cattle_grid");
+        passByDefaultBarriers.add("chain");
 
-        absoluteBarriers.add("fence");
-        absoluteBarriers.add("bollard");
-        absoluteBarriers.add("stile");
-        absoluteBarriers.add("turnstile");
-        absoluteBarriers.add("cycle_barrier");
-        absoluteBarriers.add("motorcycle_barrier");
-        absoluteBarriers.add("block");
-        absoluteBarriers.add("bus_trap");
-        absoluteBarriers.add("sump_buster");
+        blockByDefaultBarriers.add("fence");
+        blockByDefaultBarriers.add("bollard");
+        blockByDefaultBarriers.add("stile");
+        blockByDefaultBarriers.add("turnstile");
+        blockByDefaultBarriers.add("cycle_barrier");
+        blockByDefaultBarriers.add("motorcycle_barrier");
+        blockByDefaultBarriers.add("block");
+        blockByDefaultBarriers.add("bus_trap");
+        blockByDefaultBarriers.add("sump_buster");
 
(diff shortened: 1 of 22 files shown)


core/src/test/java/com/graphhopper/GraphHopperTest.java
```diff
@@ -614,7 +614,7 @@ public void testCustomModel() {
         final String customCar = "custom_car";
         final String emptyCar = "empty_car";
         CustomModel customModel = new CustomModel();
-        customModel.addToSpeed(Statement.If("road_class == TERTIARY", Statement.Op.MULTIPLY, 0.1));
```
karussell, line 617: the problem is now that previously barriers hindered using the track and when we avoid tertiary roads by so much it is forced into using tracks. In general we should think about having a bit slower speed on tracks for car, but this is for another issue.

core/src/test/resources/com/graphhopper/reader/osm/test-barriers.xml
```diff
@@ -4,16 +4,16 @@
         <tag k="name" v="Wiesbaden-Naurod"/>
     </node>
     <node id="20" lat="52" lon="9.4" version="1">
-        <tag k="barrier" v="gate"/>
```
karussell, line 7: gate is passable by default so we either need a blockByDefaultBarrier or an additional access tag

core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
```diff
@@ -71,29 +71,28 @@ public CarFlagEncoder(PMap properties) {
...
-        absoluteBarriers.add("motorcycle_barrier");
-        absoluteBarriers.add("block");
-        absoluteBarriers.add("bus_trap");
-        absoluteBarriers.add("sump_buster");
+        passByDefaultBarriers.add("gate");
+        passByDefaultBarriers.add("lift_gate");
+        passByDefaultBarriers.add("kissing_gate");
+        passByDefaultBarriers.add("swing_gate");
+        passByDefaultBarriers.add("cattle_grid");
+        passByDefaultBarriers.add("chain");
```
karussell, line 85: Not sure if kissing_gate and chain should be really pass by default.

core/src/test/java/com/graphhopper/routing/RoutingAlgorithmWithOSMTest.java
```diff
@@ -170,13 +170,13 @@ public void testMoscow() {
...
-        queries.add(new Query(55.818994, 37.595354, 55.819175, 37.596931, 1052, 14));
+        queries.add(new Query(55.818891, 37.59515, 55.81997, 37.59854, 1052, 14));
         // should choose the closest road not the other one (opposite direction)
         // http://localhost:8989/?point=55.818898%2C37.59661&point=55.819066%2C37.596374
-        queries.add(new Query(55.818898, 37.59661, 55.819066, 37.596374, 24, 2));
+        queries.add(new Query(55.818536, 37.595848, 55.818702, 37.595564, 24, 2));
         // respect one way!
         // http://localhost:8989/?point=55.819066%2C37.596374&point=55.818898%2C37.59661
-        queries.add(new Query(55.819066, 37.596374, 55.818898, 37.59661, 1114, 23));
+        queries.add(new Query(55.818702, 37.595564, 55.818536, 37.595848, 1114, 23));
```
karussell, line 179: I needed to move the coordinates (so that they snap differently) as some nearby roads are now accessible due to open barriers

core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
ratrun: Here I was wondering as well, but I didn't want to mix changes.

karussell: Ok, let's do this in a separate PR. Do you think the remaining changes look fine?

boldtrn:
Isn't this a breaking change, as it removes the option to block by default and only allow access if explicitly allowed. Might be worthwhile to add this to the changelog?

karussell:
> as it removes the option to block by default 

This feature is not removed and can be controlled via a list of tags (that subclasses can use).

boldtrn:
Thanks for the clarification, yes you are right. So now a fence with motorcar=yes would be also accepted in the routing. I thought these tags would be still these "absolute barriers" that just block access.

karussell:
> So now a fence with motorcar=yes would be also accepted in the routing.

Yes, exactly. So IMO an improvement probably also for other cases.

core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java
ratrun: There is a similar case for the `stile` and `turn_stile` barrier, which should be moved in a seperate PR. Besides that it looks fine to me now.

PR #2352, opened 2021-06-25
UI improvements - Added/Enabled alternate route option

sharmibas:
Added alternate route option in the default map

Please read [our contributing guide](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md) and note that also your contribution falls under the Apache License 2.0 as outlined there.

Your first contribution should include a change where you add yourself to the CONTRIBUTORS.md file.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
     <groupId>com.graphhopper</groupId>
     <artifactId>graphhopper-parent</artifactId>
     <name>GraphHopper Parent Project</name>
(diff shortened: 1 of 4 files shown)


karussell:
Welcome to GraphHopper and thanks for your pull request.

Unfortunately this change is not acceptable. We cannot change the server-side defaults only for UI improvements. E.g. there are many use cases where path details are unnecessary. You can add them on the request side easily instead.

Issue #2246, opened 2021-03-04
Consolidate concepts of spatial rules, custom models and custom areas


easbar:
We have an ongoing discussion about different concepts that are all about area-dependent data (often not originating from OSM) that is used in different parts of GraphHopper. We have spatial rules and now also custom models and in #2201 there is a proposal for custom areas. The discussion about how these concepts relate to each other is spread over multiple issues and PRs.

Can we clarify first what we are trying to achieve so we can derive a (hopefully) consistent approach how such functionality should be dealt with in the future? I'll start with a small list of things we'd like to have, but please feel free to modify and extend it.

1) Modify edge weights for route calculations based on the location of the edge per request / from the client side. Currently we can do this by specifying `areas` in the custom model and use the area in the `speed` or `priority` section.

2) Modify edge weights for route calculations depending on whether or not the edges are located in areas that are defined on the server side.
 
3) Choose values of encoded values depending on the location of edges (e.g. max_speed depends on the country)

Related: #1280, #1808, #2201, #1799, #2164 (and probably many more?!)

otbutz:
Also related: #1799, #2164

> spatial rules, custom models and custom areas

I'd go with @karussell naming proposal: custom models, custom areas and custom rules

> 3. Choose values of encoded values depending on the location of edges (e.g. max_speed depends on the country)

We should differentiate a bit more here:

* override
* cap
* fallback

I'd really appreciate if we could finally come up with a concept that everyone agrees with. I've been pushing this since November 2019 and fatigue is setting in to be honest. Yes it's complex and i don't expect to be spoon fed or get reviews within the minute. But getting zero feedback for multiple weeks is a real mood killer. I'm aware that reviews of code written by some thirdparty take up quite some time and even more so if it's in a part of the codebase you haven't touched recently. Again, i understand that.

Drafting a PR already takes requires some effort but it gets even worse if i don't get any feedback for a longer period of time. Reacting to change requests after over 2 weeks adds additional workload because i have to start from scratch. Any knowledge of that changes or the code it touches is gone at this point.

Unanswered questions like "how do you want to handle X" also adds to this latency. Yes, i can draft a workable solution but i might end up putting a lot of work into something which doesn't fit into the concept of the project maintainers and gets therefore rejected.

Sorry if this is off topic but i needed to vent.

karussell:
> doesn't fit into the concept of the project maintainers and gets therefore rejected

[...]

PR #2353, opened 2021-06-29
Replace spatial rules by custom areas, include country borders in bundle and add Country EncodedValue

easbar:
Here I tried to demonstrate some of the ideas described in this comment: https://github.com/graphhopper/graphhopper/issues/2246#issuecomment-870096236

I haven't really removed any code yet and rather added custom areas the way I think they should be handled. Especially this commit: d3b037b3d159827cfb05ee6e72f36c46698b1532 shows how we can use custom areas. I have added a country encoded value that only requires a few changes and can immediately be used with custom models / auto-complete in the custom model box and path details. To try this yourself just add 

```
custom_areas_folder: core/files/spatialrules/
```

to `config.yml` and add `new_country` to `graph.encoded_values`. 

Like I said, I would rather include `countries.geojson` in the main bundle, but this just demonstrates how we could use arbitrary geojson files. I also think we can use the current spatial rules the same way: We just add geojson to the `custom_areas_folder` and adjust the maxspeed parser or whatever parser we want to make area specific. 

### Update: 

This PR replaces spatial rules with 'custom areas'. This means that we can read arbitrary geojson polygons during import which are then made available to the tag parsers and flag encoders. This allows setting location-dependent encoded values, for example we could draw a Polygon around Paris and do special handling of the max speed encoded value just for this area. 

This PR also adds the `countries.geojson` file to the core bundle and uses above mechanism to provide the country for each edge. There is a new `CountryParser` which sets the country encoded value for every edge and the country is also passed to so called `CountryRules`. `CountryRules` provide an easy way to apply country-specific rules for e.g. `RoadAccess` or `MaxSpeed`.


easbar:
Here: f5748e00740171b5135d8652e53230b5aeb3fab8 I added a proof of concept how we can transfer the current spatial rules to country rules. There could be one file per country and the rules could be used in the various tag parsers (like in the example OSMMaxSpeed, but of course also others. If someone wanted to do something similar for custom areas this could also be the way to go, up to the point that the equivalent to these `CountryRule`s could be moved to some config files (depending on the specific use-case).

core/src/main/java/com/graphhopper/routing/util/spatialrules/CustomAreaIndex.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public List<CustomArea> query(double lat, double lon) {
+        Envelope searchEnv = new Envelope(lon, lon, lat, lat);
+        Point point = gf.createPoint(new Coordinate(lon, lat));
+        @SuppressWarnings("unchecked")
+        List<CustomArea> result = index.query(searchEnv);
+        return result.stream().filter(
+                c -> c.getBorders().stream()
+                        // todo: SpatialRuleLookupJTS uses SpatialRuleContainer#covers, but why?
+                        .anyMatch(point::within)).collect(Collectors.toList());
```
karussell, line 51: Maybe @otbutz can comment on this as there is some optimization going on there. The documentation says:

> This predicate is similar to within(org.locationtech.jts.geom.Geometry), but is more inclusive (i.e. returns true for more cases).

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -346,7 +353,8 @@ protected void processWay(ReaderWay way) {
             }
         }
 
-        IntsRef edgeFlags = encodingManager.handleWayTags(way, acceptWay, relationFlags);
+        List<CustomArea> customAreas = estimatedCenter == null ? emptyList() : customAreaIndex.query(estimatedCenter.lat, estimatedCenter.lon);
+        IntsRef edgeFlags = encodingManager.handleWayTags(way, acceptWay, relationFlags, customAreas);
```
easbar, line 357: For now, we can also just put the custom areas into an artificial tag like `estimated_center`.

core/src/main/java/com/graphhopper/routing/util/parsers/TagParser.java
```diff
@@ -33,4 +34,10 @@
     void createEncodedValues(EncodedValueLookup lookup, List<EncodedValue> registerNewEncodedValue);
 
     IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, boolean ferry, IntsRef relationFlags);
+
+    // todo: this way subclasses have to overwrite both methods in case they need to make use of custom areas, but on the
+    // otherhand existing subclasses can remain unchanged...
+    default IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, boolean ferry, IntsRef relationFlags, List<CustomArea> customAreas) {
+        return handleWayTags(edgeFlags, way, ferry, relationFlags);
+    }
 }
```
easbar, line 43: We should rather remove the old method instead of adding another one, and actually no signature change is necessary if we put the custom areas into an artificial way tag (see above).

core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxSpeedParser.java
```diff
@@ -54,11 +53,26 @@ public void createEncodedValues(EncodedValueLookup lookup, List<EncodedValue> li
...
+        for (CustomArea customArea : customAreas) {
+            Object countryCode = customArea.getProperties().get("ISO3166-1:alpha3");
+            if (countryCode != null) {
+                NewCountry country = NewCountry.valueOf(countryCode.toString());
+                CountryRule countryRule = CountryRule.getCountryRule(country);
+                if (countryRule != null) {
+                    maxSpeed = countryRule.getMaxSpeed(roadClass, TransportationMode.CAR, maxSpeed);
+                }
+            }
+        }
```
easbar, line 73: We can move most of this into OSMReader and pass the country as an artificial way tag instead.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -665,6 +690,38 @@ protected void importOSM() {
...
+        List<JsonFeatureCollection> jsonFeatureCollections = new ArrayList<>();
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(bordersDirectory, "*.{geojson,json}")) {
+            for (Path borderFile : stream) {
+                try (BufferedReader reader = Files.newBufferedReader(borderFile, StandardCharsets.UTF_8)) {
+                    JsonFeatureCollection jsonFeatureCollection = localObjectMapper.readValue(reader, JsonFeatureCollection.class);
+                    jsonFeatureCollections.add(jsonFeatureCollection);
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
otbutz, line 706: `UncheckedIOException` ?

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -346,6 +355,19 @@ protected void processWay(ReaderWay way) {
             }
         }
 
+        List<CustomArea> customAreas = estimatedCenter == null ? emptyList() : customAreaIndex.query(estimatedCenter.lat, estimatedCenter.lon);
+        CountryRule countryRule = null;
+        for (CustomArea customArea : customAreas) {
+            Object countryCode = customArea.getProperties().get("ISO3166-1:alpha3");
+            if (countryCode != null) {
+                NewCountry country = NewCountry.valueOf(countryCode.toString());
```
otbutz, line 363: Doesn't `Enum.valueOf()` yield an `IllegalArgumentException` if it can't find the value?

```suggestion
                NewCountry country = NewCountry.find(countryCode.toString());
```

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
```diff
@@ -0,0 +1,265 @@
...
+    ZAF("South Africa"),
+    ZMB("Zambia"),
+    ZWE("Zimbabwe"),
+    MISSING("missing");
+
+    public static final String KEY = "new_country";
+
+    private final String name;
+
+    NewCountry(String name) {
```
otbutz, line 245: Can we also include the two letter country code?

core/src/main/java/com/graphhopper/routing/util/countryrules/AustriaCountryRule.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+package com.graphhopper.routing.util.countryrules;
+
+import com.graphhopper.routing.ev.RoadAccess;
+import com.graphhopper.routing.ev.RoadClass;
+import com.graphhopper.routing.util.TransportationMode;
+
+public class AustriaCountryRule implements CountryRule {
+    public final static GermanyCountryRule RULE = new GermanyCountryRule();
```
otbutz, line 26: Copy&Paste mistake?

core/src/main/java/com/graphhopper/routing/util/parsers/CountryParser.java
```diff
@@ -0,0 +1,53 @@
...
+       registerNewEncodedValue.add(countryEnc);
+    }
+
+    @Override
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, boolean ferry, IntsRef relationFlags) {
+        List<CustomArea> customAreas = way.getTag("custom_areas", Collections.emptyList());
+        for (CustomArea customArea : customAreas) {
+            Object countryCode = customArea.getProperties().get("ISO3166-1:alpha3");
+            if (countryCode != null) {
+                NewCountry country = NewCountry.valueOf(countryCode.toString());
```
otbutz, line 47:
```suggestion
                NewCountry country = NewCountry.find(countryCode.toString());
```

core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxSpeedParser.java
```diff
@@ -56,9 +53,13 @@ public void createEncodedValues(EncodedValueLookup lookup, List<EncodedValue> li
     public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, boolean ferry, IntsRef relationFlags) {
         double maxSpeed = OSMValueExtractor.stringToKmh(way.getTag("maxspeed"));
 
+        RoadClass roadClass = RoadClass.find(way.getTag("highway", ""));
+        CountryRule countryRule = way.getTag("country_rule", null);
+        if (countryRule != null)
+            maxSpeed = countryRule.getMaxSpeed(roadClass, TransportationMode.CAR, maxSpeed);
```
otbutz, line 59: As noted in https://github.com/graphhopper/graphhopper/issues/2246#issuecomment-878078733 we should probably pass this down to the value extractor.

core/src/main/java/com/graphhopper/routing/util/spatialrules/CustomAreaIndex.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class CustomAreaIndex {
+    private final GeometryFactory gf;
+    private final STRtree index;
+
+    public CustomAreaIndex(List<CustomArea> customAreas) {
+        gf = new GeometryFactory();
+        index = new STRtree();
+        for (CustomArea customArea : customAreas) {
+            for (Polygon border : customArea.getBorders()) {
```
otbutz, line 35: We should use `PreparedPolygon` which is a lot faster for point-in-polygon checks

core/src/main/java/com/graphhopper/routing/util/spatialrules/CustomAreaIndex.java
otbutz: `PreparedPolygon.intersects()` should be preferred

otbutz (changes requested):

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
```diff
@@ -0,0 +1,265 @@
...
+    VGB("British Virgin Islands"),
+    VNM("Vietnam"),
+    VUT("Vanuatu"),
+    WSM("Samoa"),
+    XKX("Kosovo"),
+    YEM("Yemen"),
+    ZAF("South Africa"),
+    ZMB("Zambia"),
+    ZWE("Zimbabwe"),
+    MISSING("missing");
```
otbutz, line 239: Can we move this one to the top? The other EV enums use the first entry to encode the most permissive or indicate a missing/unknown value.

core/src/main/java/com/graphhopper/routing/util/countryrules/AustriaCountryRule.java
```diff
@@ -0,0 +1,77 @@
...
+
+        if (!transport.isMotorVehicle()) {
+            return RoadAccess.YES;
+        }
+
+        switch (roadClass) {
+            case LIVING_STREET:
+                return RoadAccess.DESTINATION;
+            case TRACK:
+                return RoadAccess.FORESTRY;
```
otbutz, line 66: Not subject of this PR but we should discuss this. This blocks all tracks in Austria for cars which is a bit too restrictive IMHO.

otbutz (changes requested):

karussell:
> PreparedPolygon.intersects() should be preferred

Can you expand on this @otbutz ?

core/src/main/java/com/graphhopper/routing/util/spatialrules/CustomAreaIndex.java
karussell: Can you expand on this?

otbutz: See my comment above. PreparedPolygon should be used instead of a regular JTS Polygon here as it's optimized for PIP checks.

easbar:
@otbutz Thanks for your review, I'll take a closer look at your suggestions. But from what I understand you like/agree with the general idea?

core/src/main/java/com/graphhopper/routing/util/countryrules/AustriaCountryRule.java
easbar: Oh good catch, yes!

otbutz:
Absolutely :wink:

core/src/main/java/com/graphhopper/routing/util/countryrules/AustriaCountryRule.java
easbar: Ok, please open an issue for this as it also applies to current master.

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
easbar: Ok done.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxSpeedParser.java
easbar: Ok, see discussion in #2246.

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: I just took this from `GraphHopperManaged`, but yes I agree `UncheckedIOException` seems to be what we should use (since Java 8).

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: Yes, it does. For country this should be ok, because I will move the countries.geojson into the source bundle and the NewCountry enum should then always be in sync with the country codes used in this file. But otherwise I agree it would be more useful to return `null` or something like a 'missing' value in case some value is missing (like `road_class=my_404_class`).

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
easbar: So far I used `ISO3166-1:alpha3` as country code, but no idea honestly. What are the pros and cons of the different codes? If there is some use-case for it we can of course also include other fields.

core/src/main/java/com/graphhopper/routing/util/spatialrules/CustomAreaIndex.java
easbar: Ok, so I should just create a prepared polygon using the borders of each custom area? And what about this 'filled lines' code in `SpatialRuleContainer`? Is this another optimization?

otbutz: PreparedPolygon is the fastest method for full point-in-polygon checks. These are still costly though and the "filled lines" logic helps to avoid them. We basically rasterize the polygon and mark those rectangles which are completely within the polygon so we can skip PIP checks for them.

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
otbutz: Depends on the data. Our geojson uses alpha3 and some tags in OSM like e.g `maxspeed=RU:urban` reference the alpha2 code. I doesn't really hurt to have both representations.

core/src/main/java/com/graphhopper/routing/util/countryrules/AustriaCountryRule.java
otbutz: Done: https://github.com/graphhopper/graphhopper/issues/2362

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
easbar: Ok I see. So should we use alpha3 for the enum constants and add alpha2 as additional field, just like name? Or maybe alpha3 as enum constant, but also as additional field?

Btw the countries properties currently contain all these fields:

```
  "id": -2186665,
  "admin_level": "2",
  "name": "Tonga",
  "local_name": "Tonga",
  "name_en": "Tonga",
  "flag": "http://upload.wikimedia.org/wikipedia/commons/9/9a/Flag_of_Tonga.svg",
  "name:af": "Tonga",
  "name:am": "ቶንጋ",
  "name:an": "Tonga",
  "name:ar": "تونجا",
  "name:az": "Tonqa",
  "name:ba": "Тонга",
  "name:be": "Тонга",
  "name:bg": "Тонга",
  "name:bn": "টোঙ্গা",
  "name:bo": "ཊོང་ག",
  "name:br": "Tonga",
  "name:bs": "Tonga",
  "name:ca": "Tonga",
  "name:ce": "Тонга",
  "name:cs": "Tonga",
  "name:cu": "Тонга",
  "name:cy": "Tonga",
  "name:da": "Tonga",
  "name:de": "Tonga",
  "name:dv": "ޓޮންގާ",
  "name:el": "Τόνγκα",
  "name:en": "Tonga",
  "name:eo": "Tongo",
  "name:es": "Tonga",
  "name:et": "Tonga",
  "name:eu": "Tonga",
  "name:fa": "تونگا",
  "name:fi": "Tonga",
  "name:fo": "Tonga",
  "name:fr": "Tonga",
  "name:fy": "Tonga",
  "name:ga": "Tonga",
  "name:gd": "Tonga",
  "name:gl": "Tonga",
  "name:gv": "Yn Tongey",
  "name:he": "טונגה",
  "name:hi": "टोंगा",
  "name:hr": "Tonga",
  "name:ht": "Tonga",
  "name:hu": "Tonga",
  "name:hy": "Տոնգա",
  "name:ia": "Tonga",
  "name:id": "Tonga",
  "name:io": "Tonga",
  "name:is": "Tonga",
  "name:it": "Tonga",
  "name:ja": "トンガ",
  "name:jv": "Tonga",
  "name:ka": "ტონგა",
  "name:kk": "Тонга",
  "name:kn": "ಟೋಂಗಾ",
  "name:ko": "통가",
  "name:ku": "Tonga",
  "name:kw": "Tonga",
  "name:la": "Tonga",
  "name:lb": "Tonga",
  "name:li": "Tonga",
  "name:lt": "Tonga",
  "name:lv": "Tonga",
  "name:mk": "Тонга",
  "name:ml": "ടോങ്ക",
  "name:mn": "Тонга",
  "name:mr": "टोंगा",
  "name:ms": "Tonga",
  "name:mt": "Tonga",
  "name:my": "တုံဂါနိုင်ငံ",
  "name:nl": "Tonga",
  "name:nn": "Tonga",
  "name:no": "Tonga",
  "name:nv": "Tʼónga",
  "name:oc": "Tònga",
  "name:or": "ଟୋଙ୍ଗା",
  "name:os": "Тонгæ",
  "name:pa": "ਟੋਂਗਾ",
  "name:pl": "Tonga",
  "name:pt": "Tonga",
  "name:qu": "Tunqa",
  "name:rm": "Tonga",
  "name:ro": "Tonga",
  "name:ru": "Тонга",
  "name:rw": "Tonga",
  "name:sa": "टोंगा",
  "name:se": "Tonga",
  "name:sh": "Tonga",
  "name:sk": "Tonga",
  "name:sl": "Tonga",
  "name:sm": "Toga",
  "name:sq": "Tonga",
  "name:sr": "Тонга",
  "name:su": "Tonga",
  "name:sv": "Tonga",
  "name:sw": "Tonga",
  "name:ta": "தொங்கா",
  "name:tg": "Тонга",
  "name:th": "ประเทศตองกา",
  "name:tl": "Tonga",
  "name:to": "Tonga",
  "name:tr": "Tonga",
  "name:ts": "Tonga",
  "name:tw": "Tonga",
  "name:ty": "To’a",
  "name:ug": "Ton’ga",
  "name:uk": "Тонга",
  "name:ur": "ٹونگا",
  "name:uz": "Tonga",
  "name:vi": "Tonga",
  "name:vo": "Tonguäns",
  "name:wo": "Tonga",
  "name:yo": "Tóngà",
  "name:zh": "東加",
  "boundary": "administrative",
  "maritime": "yes",
  "name:ace": "Tonga",
  "name:als": "Tonga",
  "name:arz": "تونجا",
  "name:ast": "Tonga",
  "name:bcl": "Tonga",
  "name:bjn": "Tonga",
  "name:bpy": "টঙ্গা",
  "name:bxr": "Тонга",
  "name:ceb": "Tonga",
  "name:chy": "Tonga",
  "name:ckb": "تۆنگا",
  "name:crh": "Tonga",
  "name:diq": "Tonga",
  "name:ext": "Tonga",
  "name:frp": "Tonga",
  "name:hak": "Tûng-kâ",
  "name:hif": "Tonga",
  "name:ilo": "Tonga",
  "name:lij": "Tonga",
  "name:lmo": "Tonga",
  "name:mrj": "Тонга",
  "name:nah": "Tonga",
  "name:nan": "Tonga",
  "name:nds": "Tonga",
  "name:pam": "Tonga",
  "name:pih": "Tonga",
  "name:pms": "Tònga",
  "name:pnb": "ٹونگا",
  "name:sah": "Тонга",
  "name:scn": "Tonga",
  "name:sco": "Tonga",
  "name:smn": "Tonga",
  "name:sms": "Tonga",
  "name:szl": "Tůnga",
  "name:tpi": "Tonga",
  "name:tzl": "Tonga",
  "name:war": "Tonga",
  "name:xal": "Тонһлмудин Нутг",
  "name:yue": "湯加",
  "timezone": "Pacific/Tongatapu",
  "way_area": "3.34815",
  "wikidata": "Q678",
  "ISO3166-1": "TO",
  "wikipedia": "en:Tonga",
  "name:simple": "Tonga",
  "driving_side": "left",
  "name:bat-smg": "Tuonga",
  "name:fiu-vro": "Tonga",
  "name:be-tarask": "Тонга",
  "is_in:continent": "Oceania",
  "ISO3166-1:alpha2": "TO",
  "ISO3166-1:alpha3": "TON",
  "official_name:cs": "Království Tonga",
  "official_name:de": "Königreich Tonga",
  "official_name:en": "Kingdom of Tonga",
  "official_name:es": "Reino de Tonga",
  "official_name:ru": "Королевство Тонга",
  "ISO3166-1:numeric": "776"
}
```
Since we want to move this into the bundle we should probably get rid of what we do not use (just in case this actually reduces the filesize).

core/src/main/java/com/graphhopper/routing/util/spatialrules/CustomAreaIndex.java
easbar: Ok I see.

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
otbutz: nv-i18n handles this in a similar way: https://github.com/TakahikoKawasaki/nv-i18n/blob/master/src/main/java/com/neovisionaries/i18n/CountryCode.java

One code representation as the name of the constant and the other one as a regular value.

easbar: Ok, so we will include the alpha2 and alpha3 codes, and a readable name I would say. Which name should we keep? `official_name:en`, `name:simple`, `name`, `local_name`, `name_en`, `name:en`?  I'm a bit lost here... What about `ISO3166-1:numeric`? Some other interesting fields are: `driving_side`, `is_in:contintent`, `timezone`? I just checked, not including any properties the `countries.geojson` file will be 1.3MB vs 2.7MB currently. So not too important, but getting rid of the many country name translations could be worth it.

karussell: > Since we want to move this into the bundle

What do you mean here? Isn't the file already in the jar? And I do not think that the properties matter much size-wise, especially since everything is compressed in the jar.

easbar: No, it's in core/files/spatialrules currently. The compressed size of the countries file is 900kb currently and 450kb if no properties were included. For a comparison the core jar is 1.1MB currently.

I was asking myself two questions: Can we get rid of some properties we aren't using anyways (just to optimize jar size)? And which fields should we include in the country enum, and which ones might be useful?

otbutz: Most of these tags don't work for countries with oversea territories like France: https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=2202162&class=boundary

I'd go with these:

* name:en
* ISO3166-1:alpha2
* ISO3166-1:alpha3

ISO3166-1:numeric won't hurt but i don't see a usecase for it within GraphHopper.

otbutz: > No, it's in core/files/spatialrules currently. The compressed size of the countries file is 900kb currently and 450kb if no properties were included. For a comparison the core jar is 1.1MB currently.

We could probably reduce the file size further if we simplified all the non-shared borders. We don't need to maintain full precision along coastlines or the shape of individual islands in an archipelago. But I'm afraid this will be a manual process, or at least I don't know of any tools or algorithms to automate this. And it's probably not worth the time as the current filesize without additional properties is good enough IMHO.

easbar: > I'd go with these:

Ok thanks and good to know. You mean things like timezone and continent don't work for oversea territories?

> And it's probably not worth the time as the current filesize without additional properties is good enough IMHO.

Yes, I'm not too worried about the file size and we should definitely not compromise on the border precision or something. I just thought removing the many translated country names would be a quick win.

karussell: > No, it's in core/files/spatialrules currently. 

Ah, ok. Thanks. Strange, always thought we have it in the bundle already. 

> So should we use alpha3 for the enum constants and add alpha2 as additional field, just like name? 

This sounds good to me :+1: 

> driving_side

This is indeed interesting but not sure if we should attach it to the enum or if it belongs to the country rule.

otbutz: Timezones are a completely different caliber:

Timezone map

For example:

* USA: 6 timezones
* Russia: 11(!) timezones

easbar: Ok, sure. Russia actually does not even have a timezone field (probably for this reason).

core/src/main/java/com/graphhopper/routing/util/spatialrules/CustomAreaIndex.java
easbar: I tried this and you are right it is much faster. I did a test with 100k random queries in Europe against an area index that uses the countries file and this took around 18s, but only like 0.2s when using prepared geometry :+1:
However, the filled lines optimization does not seem to make the queries faster and sometimes it is even slower. Therefore I would rather not include this, or is this especially useful for some corner cases (maybe smaller areas?)

easbar: My test code is here if you want to try yourself: https://github.com/graphhopper/graphhopper/pull/2353/commits/436821e718c2db9e595d4c7dd1c3d7546a061fbb

otbutz: > However, the filled lines optimization does not seem to make the queries faster and sometimes it is even slower. Therefore I would rather not include this, or is this especially useful for some corner cases (maybe smaller areas?)

Interesting. If it's not really faster, i'd agree that we should remove it.

easbar: In my tests it did not seem to be significant, maybe like 10%, but not sure to be honest. I'll remove it for now.

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -346,6 +355,26 @@ protected void processWay(ReaderWay way) {
             }
         }
 
+        List<CustomArea> customAreas = estimatedCenter == null ? emptyList() : customAreaIndex.query(estimatedCenter.lat, estimatedCenter.lon);
+        // special handling for countries: since they are built-in with GraphHopper they are always fed to the
+        // encodingmanager
+        NewCountry country = null;
```
otbutz, line 361: If we later substitute `null` with `NewCountry.MISSING`, i'd rather use it as the initial value.

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -346,6 +355,26 @@ protected void processWay(ReaderWay way) {
...
+            if (countryCode != null) {
+                country = NewCountry.valueOf(countryCode.toString());
+                if (countryRule != null)
+                    LOGGER.warn("Multiple country rules found for way {}", way.getId());
+                countryRule = CountryRule.getCountryRule(country);
+            }
+        }
+        way.setTag("country", country != null ? country : NewCountry.MISSING);
+        if (countryRule != null)
+            way.setTag("country_rule", countryRule);
```
otbutz, line 374: Isn't it the other way around? In this case the way would belong to multiple countries.

I'd prefer to lookup the country in the loop and retrieve the rule afterwards:

```suggestion
        NewCountry country = NewCountry.MISSING;
        for (CustomArea customArea : customAreas) {
            Object countryCode = customArea.getProperties().get("ISO3166-1:alpha3");
            if (countryCode == null)
                continue;
            
            if (country != NewCountry.MISSING)
                LOGGER.warn("Multiple countries found for way {}", way.getId());
            
            country = NewCountry.valueOf(countryCode.toString());
        }
        way.setTag("country", country);
        
        CountryRule countryRule = CountryRule.getCountryRule(country);
        if (countryRule != null)
            way.setTag("country_rule", countryRule);
```

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
```diff
@@ -0,0 +1,284 @@
...
+
+    public static EnumEncodedValue<NewCountry> create() {
+        return new EnumEncodedValue<>(NewCountry.KEY, NewCountry.class);
+    }
+
+    public static NewCountry find(String name) {
+        if (name == null || name.isEmpty())
+            return MISSING;
+
+        for (NewCountry country : values()) {
```
otbutz, line 276: This is usually fast enough but this will be executed for each way in the OSMReader. Could we add a static HashMap and fill it once?

core/src/main/java/com/graphhopper/routing/util/parsers/CountryParser.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public void createEncodedValues(EncodedValueLookup lookup, List<EncodedValue> registerNewEncodedValue) {
+        registerNewEncodedValue.add(countryEnc);
+    }
+
+    @Override
+    public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay way, boolean ferry, IntsRef relationFlags) {
+        NewCountry country = way.getTag("country", null);
```
otbutz, line 44:
```suggestion
        NewCountry country = way.getTag("country", NewCountry.MISSING);
```

otbutz (changes requested):

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: Yes, I probably forgot updating this when I added the `MISSING` value.

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: Yes, fully agree.

core/src/main/java/com/graphhopper/routing/util/parsers/CountryParser.java
easbar: Yes, done

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
easbar: I'll add a todo and then we should measure this before we add a hashmap. Maybe also calling `valueOf` and catching the exception so we can return `MISSING` if there is no match is fast.

easbar:
Next thing we should try is actually replacing spatial rules with custom areas.

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
otbutz: Logging both countrycodes would also be good.

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
otbutz: > Maybe also calling `valueOf` and catching the exception so we can return `MISSING` if there is no match is fast.

Instantiation of exceptions is quite costly. I'd avoid that in any hot code path.

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -346,6 +355,26 @@ protected void processWay(ReaderWay way) {
...
+        List<CustomArea> customAreas = estimatedCenter == null ? emptyList() : customAreaIndex.query(estimatedCenter.lat, estimatedCenter.lon);
+        // special handling for countries: since they are built-in with GraphHopper they are always fed to the
+        // encodingmanager
+        NewCountry country = NewCountry.MISSING;
+        for (CustomArea customArea : customAreas) {
+            Object countryCode = customArea.getProperties().get("ISO3166-1:alpha3");
+            if (countryCode == null)
+                continue;
+            if (country != NewCountry.MISSING)
+                LOGGER.warn("Multiple countries found for way {}", way.getId());
```
otbutz, line 367:
```suggestion
                LOGGER.warn("Multiple countries found for way {}: {}, {}", way.getId(), country, countryCode);
```

otbutz (changes requested):

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: done

core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
```diff
@@ -57,11 +57,12 @@ public IntsRef handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, boolean fer
...
 
-        SpatialRuleSet spatialRuleSet = readerWay.getTag("spatial_rule_set", null);
-        if (spatialRuleSet != null && spatialRuleSet != SpatialRuleSet.EMPTY) {
-            RoadClass roadClass = RoadClass.find(readerWay.getTag("highway", ""));
-            accessValue = spatialRuleSet.getAccess(roadClass, TransportationMode.CAR, YES);
-        }
+        RoadClass roadClass = RoadClass.find(readerWay.getTag("highway", ""));
+        // todo: for spatial rules we got the entire set of rules for this way, but now we throw an error in case
+        //       there are multiple rules for the same way. is this ok?
+        CountryRule countryRule = readerWay.getTag("country_rule", null);
```
otbutz, line 63: The SpatialRule interface is more generic. You could have multiple rules with different areas which target the same way. e.g country/state/area. Our previous country rules are just a special case which happens to be overlapping free.

But for dedicated CountryRules we can give that guarantee.

The usecase for a set of rules would be point 5 of your comment: https://github.com/graphhopper/graphhopper/issues/2246#issuecomment-870096236. With custom areas there is no such guarantee that the borders don't overlap. We should keep this in mind if we tackle those later on.

core/src/main/java/com/graphhopper/routing/util/spatialrules/AreaIndex.java
```diff
@@ -0,0 +1,79 @@
...
+import org.locationtech.jts.geom.prep.PreparedGeometry;
+import org.locationtech.jts.geom.prep.PreparedGeometryFactory;
+import org.locationtech.jts.index.strtree.STRtree;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class AreaIndex<T extends AreaIndex.HasBorders> {
+
+    public interface HasBorders {
```
otbutz, line 32: An `AreaIndex` should IMHO keep an index of `Areas`:

```suggestion
    public interface Area {
```

otbutz (changes requested):

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -483,26 +481,27 @@ private boolean createLandmarksForSubnetwork(final int startNode, final byte[] s
...
+            List<SplitArea> areas = areaIndex.query(na.getLat(adjNode), na.getLon(adjNode));
+            SplitArea areaAdj = areas.isEmpty() ? null : areas.get(0);
 
             int baseNode = allEdgesIterator.getBaseNode();
-            set = ruleLookup.lookupRules(na.getLat(baseNode), na.getLon(baseNode));
-            SpatialRule ruleBase = set.getRules().isEmpty() ? null : set.getRules().get(0);
-            if (ruleAdj != ruleBase) {
+            areas = areaIndex.query(na.getLat(baseNode), na.getLon(baseNode));
+            SplitArea areaBase = areas.isEmpty() ? null : areas.get(0);
+            // todo: what if both areaAdj and areaBase are null?
```
otbutz, line 503: We mark only those as inaccessible which don't have the same area. `null` is also some kind of area which describes the unassigned landmass: https://github.com/graphhopper/graphhopper/blob/master/core/src/main/resources/com/graphhopper/routing/lm/map.geo.json

* Europe -> Russia = inaccessible
* Russia -> Kazakhstan(null) = inaccessible
* Kazakhstan(null) -> Uzbekistan(null) = accessible

easbar: Yes so this is fine.

core/src/main/java/com/graphhopper/routing/util/spatialrules/AreaIndex.java
easbar: Thanks for reviewing this. I'm in the process of replacing spatial rules with custom areas, trying to find a good way to do so. Therefore there might be some back and forth changes.

Here I wasn't sure yet if we actually need to make the AreaIndex generic this way, but so far I think it has some advantages. We already use it for two different things: CustomAreas and the Landmark splitting areas.

Regarding the rename: I also like `Area`, so I left a little todo and we can decide later.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
easbar: > But for dedicated CountryRules we can give that guarantee.

Ok yes, because our countries.geojson has this property.

> With custom areas there is no such guarantee that the borders don't overlap. We should keep this in mind if we tackle those later on.

The custom areas are already added as artificial/temporary way tag by the OSMReader, so the complete `List<CustomArea>` is available to all tag parsers and flag encoders. I don't think there is anything else we need to do there and the rest is up to the single tag parsers (they can either throw an error if multiple matches were not expected or make use of all matches etc.).

easbar: I'll remove the this todo, because like you said there is only one country match.

core/src/test/java/com/graphhopper/routing/util/spatialrules/AreaIndexTest.java
```diff
@@ -177,6 +169,29 @@ else if (areas.isEmpty())
...
+            customAreas = jsonFeatureCollection.getFeatures().stream().map(CustomArea::fromJsonFeature).collect(Collectors.toList());
+        }
+        return new AreaIndex<>(customAreas);
+    }
+
+    @Test
+    public void testCountries() throws IOException {
+        AreaIndex<CustomArea> countryIndex = createCountryIndex();
+        // Berlin
+        assertEquals("Deutschland", countryIndex.query(52.5243700, 13.4105300).get(0).getProperties().get("name"));
```
otbutz, line 188: Looks like that the name is currently referring to the local name. IMHO we should probably prefer the english version or what do you think?

easbar: Well for this test it does not matter which property we look at. Or do you want to change the name field in `countries.geojson`?

otbutz: This is more about the properties we want to keep in the `countries.json`

easbar:
@otbutz In `GraphHopper#setupCountrySpatialRules` we take care to only include countries for which there are SpatialRules (Germany + Austria) in the `SpatialRuleLookup` and additionally there is a `maxBounds` parameter that can be used to limit the extent of the lookup. Is this something we need? With the new `NewCountry` enum we now consider every country and it would be easiest to just add all countries to the new `AreaIndex`, or do you think this would result in a significant slow down for smaller maps?

otbutz:
IMHO the best approach would be to handle this based on the data we import. Doesn't the pbf have metadata which describes the bounding box? We could simply drop any country from the index which doesn't intersect the bbox.

easbar:
The OSM bounding box is not very reliable. I think we do not use it for some reason. Otherwise this seems like a good idea, but then again if it does not really provide an important speed up let's not do it :) I'd say we continue any once we like the functionality we will check performance and add optimizations like this if they are required.

core/src/test/java/com/graphhopper/routing/util/spatialrules/AreaIndexTest.java
easbar: Ah yeah did we decide which ones we will keep yet? Anyway, I'll change this test to `name:en` already.

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
easbar: Should we keep `alpha2`, `alpha3` `name:en`, `ISO3166-1:numeric`, `flag` and `wikidata` in `countries.geojson` and remove everything else? The last three wouldn't be used by default but I thought they might be useful to somebody?

easbar:
My current todo list:

- [ ] Move OSMReader constructor argument (AreaIndex) into separate setter similar to others (?)
- [ ] Change CountryRule method signatures, maybe just pass ReaderWay? What is TransportationMode for?
- [ ] todo in CountryRuleTest (add a test that calculates a route in a country without a rule and one in a country with a rule)
- [ ] Move AreaIndexTest performance test to measurement, or remove if it is not required, test Measurement
- [ ] Check/compare performance compared to master/spatial rules

docs/core/spatial-rules.md
```diff
@@ -1,21 +0,0 @@
...
-`highway=track` in Austria is `access=yes`, whereas the default accessibility is `access=destination` in Germany. Other 
-examples are different max speeds for different countries. More information on different road rules can be found in the 
-OSM wiki, different [accessibilities](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restrictions) and 
-different [speeds](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxspeed#Motorcar).
-
-## Enabling Rules
-
-If you have a working GraphHopper setup it is easy to enable Spatial Rules.
-We provide a set of approximate country borders, within the GraphHopper repository. If you need exact borders you can
-get the exact borders from [here](https://github.com/datasets/geo-countries). Go to your `config.yml` file and
```
easbar, line 13: in `spatialrules/README.md` we describe how we got the borders from osm-boundaries.com. How is this different to the resource described here? And which one should we recommend? Should we still describe our borders as 'approximate' or are they more or less exact? @otbutz?

easbar:
Regarding the 'proof of concept', I think this branch proofs the concepts works. Spatial rules were completely replaced by custom areas and country-specific things can be handled via country rules. By modifying tag parsers and flag encoders we are very flexible in modifying encoded values.

karussell:
> What is TransportationMode for?

those country rules are currently car-centric but we could also make access to road classes dependent on the transportation mode e.g. allow track per default in one country for bike and forbid it in another.

easbar:
> those country rules are currently car-centric but we could also make access to road classes dependent on the transportation mode e.g. allow track per default in one country for bike and forbid it in another.

Ah ok, but then we need multiple encoded values, like multiple `RoadAccess` for the different transportation modes?

I mean currently we do this in `OSMRoadAccessParser`:

```java
        CountryRule countryRule = readerWay.getTag("country_rule", null);
        if (countryRule != null)
            accessValue = countryRule.getAccess(roadClass, TransportationMode.CAR, RoadAccess.YES);
```

and if `OSMRoadAccessParser` set multiple encoded values (one for each transportation mode) we would have to obtain multiple access values.

karussell:
Yes, either we need multiple transportation modes in the parser or one parser for every transportation mode (see the road_access and rename_road_access and node_parser_extraction branches where I played with this a bit)

easbar:
Ok so the country rule methods should probably have three arguments: `ReaderWay`, `TransportationMode` and the current value.

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
easbar: I did this

easbar:
I think this PR is getting close to a merge-able state. I decided to disable country rules by default for this PR, so it is more or less just a refactoring. We can enable country rules (so far there is only Germany+Austria) by default in a follow-up PR.

easbar:
Currently the only thing I want to check is performance, because this PR removed the 'filled lines' optimization that we used in `SpatialRuleLookup` and there is no longer an option to restrict the lookup (now `AreaIndex`) to a bounding box.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -418,6 +410,24 @@ public GraphHopper setTagParserFactory(TagParserFactory factory) {
...
+    public GraphHopper setCustomAreasDirectory(String customAreasDirectory) {
+        this.customAreasDirectory = customAreasDirectory;
+        return this;
+    }
+
+    public String getCustomAreasDirectory() {
+        return this.customAreasDirectory;
+    }
+
+    public GraphHopper setCountryRulesEnabled(boolean countryRulesEnabled) {
```
otbutz, line 422: Shouldn't we allow users to add or overwrite CountryRules? This should IMHO be doable without having to create a fork.

otbutz (changes requested):

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: Yes, I think we can just offer a setter for a `CountryRuleFactory`.

core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
```diff
@@ -0,0 +1,79 @@
...
+import org.locationtech.jts.geom.prep.PreparedGeometry;
+import org.locationtech.jts.geom.prep.PreparedGeometryFactory;
+import org.locationtech.jts.index.strtree.STRtree;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class AreaIndex<T extends AreaIndex.HasBorders> {
+
+    // todo: rename to just 'Area'?
```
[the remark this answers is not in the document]
easbar: Ok yes. I don't have a strong preference here... `Area` sounds like more than it really is? But it's nice and short... Should we keep it an inner class? `AreaIndex.Area` works for me. Just `Area`, not sure. Or just `HasBorders`?

easbar:
> Currently the only thing I want to check is performance, because this PR removed the 'filled lines' optimization that we used in SpatialRuleLookup and there is no longer an option to restrict the lookup (now AreaIndex) to a bounding box.

With the current benchmark the graph import for Germany (just the default countries, no custom areas) is more or less the same: Something like 151s vs 141s (~ 7% slower), so nothing to worry about IMO. The `AreaIndex` queries take around `0.08ms` on average, so one million lookups take around 80s.

@otbutz do you have some kind of datasets with lots of custom areas? I wonder if performance becomes more critical when there are lots of areas.

easbar:
Btw there actually *is* a major performance problem: CH preparation becomes around x3 slower when country rules are enabled. I think this is related to the handling of `track` ways that use `RoadAccess.Destination` in `GermanyCountryRule`. But this problem already exists in master and is beyond the scope of the present PR.

core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
```diff
@@ -0,0 +1,79 @@
...
+        index.build();
+    }
+
+    public List<T> query(double lat, double lon) {
+        Envelope searchEnv = new Envelope(lon, lon, lat, lat);
+        @SuppressWarnings("unchecked")
+        List<IndexedCustomArea<T>> result = index.query(searchEnv);
+        Point point = gf.createPoint(new Coordinate(lon, lat));
+        return result.stream()
+                .filter(c -> c.covers(point))
```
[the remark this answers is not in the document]
easbar: Ok, I'll try and re-run the benchmark.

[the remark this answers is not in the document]
easbar: Ah, but `IndexedCustomArea#covers` already uses `preparedGeometry#intersects`. Is this what you meant? We can rename `covers` to `intersects` if this makes it more clear.

core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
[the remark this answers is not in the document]
otbutz: I'd roll with `AreaIndex.Area`.

otbutz:
> @otbutz do you have some kind of datasets with lots of custom areas? I wonder if performance becomes more critical when there are lots of areas.

The STRtree is usually doing a good job keeping the number of PIP checks low. I'd expected performance only to degrade if a lot of overlapping areas are involved and/or if the polygons have a high resolution.

> Btw there actually *is* a major performance problem: CH preparation becomes around x3 slower when country rules are enabled. I think this is related to the handling of `track` ways that use `RoadAccess.Destination` in `GermanyCountryRule`. But this problem already exists in master and is beyond the scope of the present PR.

We need to take a closer look on our rules and their impact on routing in general. See https://github.com/graphhopper/graphhopper/issues/2266

One of the problems is that we would need to differentiate between optimistic(upper bound) and pessimistic(lower bound) fallback values. IMHO this is in the same vein as distinguishing the TransportationMode. We need to find a balance between too much detail and oversimplification.

Talking about simplfication: did you remove the other properties from the countries.geojson?

Could you take a look on memory/storage consumption? I fear that the large enum comes with a price tag.

core/src/main/java/com/graphhopper/routing/ev/NewCountry.java
easbar: Do we even need `Country#find`? Currently we use `valueOf` in `OSMReader`, because the values are coming from the bundled countries.geojson file, so there is no need to handle the case where the value does not exist.

easbar:
> We need to take a closer look on our rules and their impact on routing in general.

Yes, sure, and hopefully this PR will be the foundation for this. I disabled the country rules by default in this PR and we can improve the rules, fix CH preparation performance and enable the rules by default afterwards.

> Talking about simplfication: did you remove the other properties from the countries.geojson?

Yes, see here: https://github.com/graphhopper/graphhopper/pull/2353#discussion_r676033086

> Could you take a look on memory/storage consumption? I fear that the large enum comes with a price tag.

In which context? Do you mean the Country EV? The Country enum has 218 entries, so it should take one byte per edge (8bits, 2^8=256). For a global import with currently around 250mio edges this would mean `250MB` of additional memory is required for the Country EV. That should not be a problem and the Country EV can also be disabled.

otbutz:
250MB sounds very reasonable

docs/core/spatial-rules.md
otbutz: > in `spatialrules/README.md` we describe how we got the borders from osm-boundaries.com. How is this different to the resource described here? And which one should we recommend?

I think i actually missed to update this file when i recreated our borders. The one referring to osm-boundaries is the correct one.

> Should we still describe our borders as 'approximate' or are they more or less exact? @otbutz?

They gained a lot more precision but should still be considered an approximation which should be sufficient for 99.9% of routing related decisions:

grafik

easbar: Ok I think we can leave the documentation as it is then.

core/src/test/java/com/graphhopper/routing/util/spatialrules/AreaIndexTest.java
```diff
@@ -0,0 +1,182 @@
...
+                new Coordinate(9, 9), new Coordinate(10, 9), new Coordinate(10, 10), new Coordinate(9, 10),
+                new Coordinate(9, 9)});
+        AreaIndex<CustomArea> index = new AreaIndex<>(Arrays.asList(
+                createCustomArea("1", border1),
+                createCustomArea("2", border2),
+                createCustomArea("3", border3),
+                createCustomArea("4", border2, border3)));
+        testQuery(index, 0, 0);
+        testQuery(index, 1.5, 1.5, "1");
+        // todo: is it somehow necessary that coordinates right on the border are included?
```
otbutz, line 60: I'm not sure we'll ever face problems with this one in reality. JTS uses full double-precision for its checks(See [FAQ](https://locationtech.github.io/jts/jts-faq.html#C3)) and the chance that the center of a way is *exactly* on the border is close to zero.

Or, to put it another way: the margin of error introduced by simplified boundaries is orders of magnitude higher

easbar: Yes fully agree. I think I just left this comment there because in the corresponding spatial rule test we had a check like this. Will remove the comment.

core/src/main/java/com/graphhopper/util/Constants.java
```diff
@@ -67,7 +67,7 @@
     private static final int JVM_MINOR_VERSION;
 
     public static final int VERSION_NODE = 5;
-    public static final int VERSION_EDGE = 19;
+    public static final int VERSION_EDGE = 20;
```
easbar, line 70: Is there anything else we need to take care of since the `country` EncodedValue now changed it's meaning? Or do we even need to called it something else?

otbutz: Do we even need to do this? Changing the enum constants should yield a different hash of the EV.

otbutz: https://github.com/graphhopper/graphhopper/blob/86d23a3131fb7f65706f2a52ff23bfe250d27c48/core/src/main/java/com/graphhopper/routing/ev/EnumEncodedValue.java#L60-L63

easbar: Ok, I'll revert this then. This way no new import is required in case the country encoded value was not used at all.

boldtrn:
> Btw there actually is a major performance problem: CH preparation becomes around x3 slower when country rules are enabled. I think this is related to the handling of track ways that use RoadAccess.Destination in GermanyCountryRule. But this problem already exists in master and is beyond the scope of the present PR.

Have you tested why this happens? 

I haven't had a chance to look at the code changes yet (sorry :( - summer is always pretty busy).

My understanding from reading your proposal a few weeks ago:
- For every way we check **once** which rules apply and we store this in an encoded value
- Everytime we use the country rule afterwards, we only need to read the EV, so there should be no major impact on CH preparation

Is the current approach different from this?

easbar:
>  Have you tested why this happens?

Yes, it is because witness path searches that include destination edges take very long because of the large weight of these edges. `GermanyCountryRule` sets access=destination for all tracks, which are very many edges.

> My understanding from reading your proposal a few weeks ago:
>  For every way we check once which rules apply and we store this in an encoded value

Yes, for every way we find the matching country rule (and all other custom areas) and this information is forwarded to all tag parsers. Currently only the max speed and road access parsers (and `CountryParser`, see below) make use of this information, so it affects the values of the max speed and road access encoded values. This happens only for the countries for which there are country rules, of course. These are still only Germany and Austria. There is also a new `CountryParser` which simply stores the country for every edge (`Country` EncodedValue). 

>  Everytime we use the country rule afterwards, we only need to read the EV, so there should be no major impact on CH preparation. Is the current approach different from this?

After the import (when the encoded values are set) the country rules are never applied again. Like I said, CH preparation is still affected, because of the increased weights of the access=destination edges. For Germany I measured a x3 slow down for both node- and edge-based preparation. I am working on a fix for this.

easbar:
Is everybody happy? I will merge this then.

otbutz:
Just one last thought: do we want a default country rule? I'd really like to get rid of hardcoded fallback values in FlagEncoders in the long run:

https://github.com/graphhopper/graphhopper/blob/86d23a3131fb7f65706f2a52ff23bfe250d27c48/core/src/main/java/com/graphhopper/routing/util/CarFlagEncoder.java#L108-L141

I'd avoid any logic for now and just go for a NOOP implementation first. The reason why i bring this up now is that this would allow us to remove a bit of null handling. We would *always* have a CountryRule which should be applied.

easbar:
>  I'd really like to get rid of hardcoded fallback values in FlagEncoders in the long run:

Yes, the thing we need to do here is that instead of setting average speed and access in `FlagEncoders` there really should be tag parsers that create these speed and access encoded values. But this is a very different issue and removing `FlagEncoders` this way has been on the list for years now.

> The reason why i bring this up now is that this would allow us to remove a bit of null handling. We would always have a CountryRule which should be applied.

I really don't mind the null handling here. For some countries there is no rule, and the most direct way to state this in code is that the rule is null (=missing) in this case. For country rule it could kind of work, because so far all `CountryRule` methods have a clear way of 'doing nothing'. But generally I prefer `null` for these things. For example I would totally remove `ElevationProvider.NOOP`, because `getEle` has no clear way of 'doing nothing', it has to provide a value. So it provides `Double.NaN` and one has to handle `Double.NaN`, i.e. nothing is improved by having this `NOOP` provider. For the same reason I did not add something like a `CountryRuleFactory.EMPTY`.

easbar:
>  And once we have this in place we can of course use country rules for average_speed and access, too.
Btw in principle with the current code we can already make use of the country information in `FlagEncoder#handleWayTags` to set country-specific values for speed&access.

otbutz:
LGTM then :)

easbar:
> LGTM then :)

Cool, and thanks a lot for your initial work on this and all your help with this PR.

boldtrn:
> Yes, it is because witness path searches that include destination edges take very long because of the large weight of these edges. GermanyCountryRule sets access=destination for all tracks, which are very many edges.

Ah thanks yes that makes sense. I thought there would be direct slow down because of too many lookups.

otbutz:
@easbar any updates on the CH slowdown? Could we enable countryrules by default now?

PR #2355, opened 2021-07-01
Remove FlagEncoder#getVersion

easbar:
We do not need to keep track of the versions in that much detail, base graph's version (nodes, edges, geometry etc.) should be enough. Also we want to get rid of `FlagEncoder` anyway.


karussell (approved):
:+1:

Issue #2365, opened 2021-07-14
Remove workarounds for missing jackson dependency in core


otbutz:
With #2354 we finally have jackson available in core so we can get rid of a few warts like GraphHopperConfigMixIn or JSON handling in GraphHopperManaged

easbar:
This is somewhat 'sketchy territory', because often it leads to the question what the core module really is about. For example should parsing `config.yml` (GraphHopperConfigMixIn) be part of the core? Or is this a feature of the 'webapp' (web-bundle)? Is the `GraphHopper` class an application in and of itself or is there only really one (web) application/server (which uses Dropwizard and config.yml as configuration?). 

I think I will start with this JsonFeatureCollection constructor parameter, because this is very similar to what we need for custom areas.

michaz:
Thanks! Just want to say I appreciate *both* you going forward *and* your level of architectural caution here. ,-)

easbar:
Yes, I think core should be 'allowed' to use Jackson to read GeoJSON (just generally), but e.g. the `GraphHopperConfigMixIn` that is solely used to read the `GraphHopperConfig` from `config.yml` is rather not a part of core.

michaz:
Of course, most of that discussion would be conveniently over if the JRE just had JSON internally.

Issue #2354, opened 2021-06-29
Move Jackson into core and maybe remove the web-api module?


easbar:
For a long time we tried to avoid Jackson in core due to Android compatibility, but now we do not have this restriction anymore we can finally add the Jackson dependency to core? This way we should be able to avoid some ugly workarounds like the `jsonFeatureCollection` argument in the `GraphHopper` class and the Jackson mixins. To me it looks like we can just make core a depend on the web-api module, but probably we can then just as well move the entire module into core?!

otbutz:
100% agree

easbar:
Should we do this now/before #2353? For #2353 we have to make a choice: Either we add `countries.geojson` to the `web-bundle` just like landmark splitting collection and inject the custom areas via the GraphHopper constructor in `GraphHopperManaged`. This seems really odd because we (only) need the custom areas for a core class (OSMReader) and the entire reason we cannot read the custom areas at the place we need them would be that we restrict ourselves to not having Jackson in core. Or we fix this issue first, then we can move `countries.geojson` into core. Honestly, so far I do not see a single reason not to have Jackson in core. In core we already have file IO, so if we read and write files why should we not read/write serialize/deserialize json files? We parse OSM in binary format, so why can we not read json files? Also for XML we would probably not hesitate to use it since it is included in java already? So just because Jackson is used at the web interface that does not mean Jackson is something that is strictly related to web-stuff and we should keep it out of core because we are trying to keep web-stuff out of core. Jackson is *also* used at the web interface, but we just as well need it to read custom areas, custom models, landmark splitting collections and would probably even use it for more things like base graph properties etc. if we did not have this restriction in the past. 

Regarding the `web-api` module I don't have a strong opinion, but also I do not see a real reason to keep it at the moment.

otbutz:
IMHO the whole injection solution is just a dirty workaround and i would be happy if we could use core as-is without doing the injection thing manually.

easbar:
There is one place where we use web-api, but not core: client-hc

otbutz:
We could switch the jackson dependencies from transitive to explicit ones in the client-hc project.

easbar:
I think I want to do this the minimal invasive way, otherwise there will be just too much to discuss :) So I think I want to keep the web-api module and only add Jackson dependencies to core. Like I said I see no reason why core should not depend on Jackson, because it is simply something used to do disk I/O which we do a lot in core.

[...]

PR #2366, opened 2021-07-14
Move landmark splitting and spatial rule border parsing from web-app into core

easbar:
After #2354 we can now move some of the code in `GraphHopperManaged` closer to where it is actually used. Particularly we were parsing the landmark splitting borders geojson and spatial rule borders geojson files in `GraphHopperManaged` and 'injected' this code into `GraphHopper`. But the only reason we did this was that we weren't able to read geojson in core.

Also for #2353 I'd like to read the countries.geojson and other custom area geojson files directly from the GraphHopper class.

Related to #2365

### DIFF
--- a/docs/core/spatial-rules.md
+++ b/docs/core/spatial-rules.md
@@ -8,11 +8,11 @@ different [speeds](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxs
 
 ## Enabling Rules
 
-I you have a working GraphHopper setup it is easy to enable Spatial Rules, **but they only work with the DataFlagEncoder**.
-We provide a set of approximate country borders, within the GraphHopper repository. If you need exact borders you can
-get the exact borders from [here](https://github.com/datasets/geo-countries). Go to your
-`config.yml` file and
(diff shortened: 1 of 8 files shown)


docs/core/spatial-rules.md
```diff
@@ -8,11 +8,11 @@ different [speeds](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxs
 
 ## Enabling Rules
 
-I you have a working GraphHopper setup it is easy to enable Spatial Rules, **but they only work with the DataFlagEncoder**.
-We provide a set of approximate country borders, within the GraphHopper repository. If you need exact borders you can
-get the exact borders from [here](https://github.com/datasets/geo-countries). Go to your
-`config.yml` file and
-uncommend the line: `spatial_rules.borders_directory` and point it to the directory where your rules are. You need to re-import your graph after that.
+I you have a working GraphHopper setup it is easy to enable Spatial Rules, **but they only work with the
+DataFlagEncoder**. We provide a set of approximate country borders, within the GraphHopper repository. If you need exact
```
karussell, line 12: We can remove the bold phrase as it should work regardless of the FlagEncoder.

easbar: -> https://github.com/graphhopper/graphhopper/pull/2367

PR #2367, opened 2021-07-14
Remove some outdated mentions of the DataFlagEncoder

easbar:
DataFlagEncoder was removed already for 1.0, but it is still mentioned in a few places

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/InstructionsOutgoingEdges.java
+++ b/core/src/main/java/com/graphhopper/routing/InstructionsOutgoingEdges.java
@@ -138,9 +138,6 @@ class InstructionsOutgoingEdges {
 
         for (EdgeIteratorState edge : visibleAlternativeTurns) {
             tmpSpeed = getSpeed(edge);
-            if (tmpSpeed < 1) {
-                return false;
-            }
             if (tmpSpeed > maxSurroundingSpeed) {
                 maxSurroundingSpeed = tmpSpeed;
             }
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/graphhopper/routing/InstructionsOutgoingEdges.java
```diff
@@ -139,7 +139,6 @@ public boolean outgoingEdgesAreSlowerByFactor(double factor) {
         for (EdgeIteratorState edge : visibleAlternativeTurns) {
             tmpSpeed = getSpeed(edge);
             if (tmpSpeed < 1) {
-                // This might happen for the DataFlagEncoder, might create unnecessary turn instructions
                 return false;
```
easbar, line 142: Do you know if this is still necessary?

karussell: No, unfortunately I do not know. Maybe remove it if no tests fail due to removing?

easbar: @boldtrn maybe?

easbar: The tests are green also after removing it. So let's remove it and put it back if someone can show a case where this is important.

boldtrn: Not sure if this case can still happen, probably not, so makes sense to remove it :)

easbar: Ok thanks for taking a look :+1:

PR #2371, opened 2021-07-22
Avoid "bicycle=use_sidepath" ways instead of preferring it.

### LINKED ISSUE #1454: bicycle=use_sidepath not recognized; preference for cycleways is off
https://www.openstreetmap.org/directions?engine=graphhopper_bicycle&route=51.51773%2C9.94051%3B51.52776%2C9.93919#map=18/51.52173/9.94200
I'm not sure if they have their own graphhopper (and what version) or whether it is delegated to an external instance. I hope it does not matter.

The route so determined includes e.g. way 262459057, which is not allowed for bicycles due to `bicycle=use_sidepath`. Even so, the routing engine IMO should even prefer highway=cycleway over highway=primary if the time difference is miniscule.

https://www.openstreetmap.org/directions?engine=graphhopper_bicycle&route=51.51659%2C9.94031%3B51.52776%2C9.93919#map=16/51.5231/9.9442
This routing result is also a strange one. Seems like a cycle road is given (unnecessarily) higher preferrence over a cycleway.

ratrun:
Fixes #1454.
Prefering ways with "bicycle=use_sidepath" is the opposite to what is correct for bicycles. Now we are avoiding these ways.
See https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Duse_sidepath.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -472,12 +472,13 @@ abstract public class BikeCommonFlagEncoder extends AbstractFlagEncoder {
             weightToPrioMap.put(100d, PREFER.getValue());
         }
 
+        if (way.hasTag("bicycle", "use_sidepath")) {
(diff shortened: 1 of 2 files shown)


karussell:
Thank you!

PR #2372, opened 2021-07-24
Move api module into web-api, remove Jackson MixIns

easbar:
As suggested by @michaz here I moved the api module into web-api. Core now depends on web-api instead of api, so the only thing that really changed is that all classes that were previously in web-api are now included in core as well. These classes were mostly Jackson Serializers/Deserializers and the Jackson MixIns. Since Jackson is already a core dependency since #2354 this does not really change anything. 

I then also removed the Jackson MixIns and added the equivalent annotations to the corresponding classes directly. These were `CH/LM/Profile`, `GraphHopperConfig` and `GHRequest`. Except `GHRequest` these are all server-side classes, so actually not related to the web interface at all. 

So the only relevant change might be that `GHRequest` has now some Jackson annotations while previously we used a mix in instead.

Btw we now have these modules (dependencies in square brackets)

```
// GraphHopper + Server
core [web-api] -> GraphHopper without add-ons and the web-server part
web-api -> things we also use on the (Java) client side
client-hc [web-api] -> our hand-crafted Java client, the reason why we need web-api
web-bundle [core, reader-gtfs, map-matching] -> web/server related stuff, but without the server
web [web-bundle, navigation] -> GraphHopper as web-server

// add-ons
reader-gtfs [core] -> public transport
map-matching [core, hmm-lib] -> map matching
hmm-lib -> required by map matching
navigation [core] -> an additional server endpoint for the navigation SDK

// other
tools [core] -> our Swing GUI (and some other things which probably could be in core)
example [core] -> example code for documentation purposes
```

Related: #2365


michaz:
Sweet! Now we can move classes that aren't (and shouldn't) used by the client at all out of there, either to `web-*` or, if used by `core`, to `core`, and then we're done and it is what it is, right?

michaz (approved):

easbar:
I think there aren't any classes in web-api that are *not* used by client-hc. Except `GraphHopperAPI` which I will remove anyway.

PR #2373, opened 2021-07-26
Read elevation from OSM node then fallback to elevation provider

gberaudo:
Draft implementation of the idea discussed in https://github.com/graphhopper/graphhopper/pull/2230#issuecomment-885450232.

TODO with questions:
- [ ] validate with @otbutz, @easbar, @karussell this is the way to go;
- [ ] have `mvn clean test verify` pass as required by CONTRIBUTING.md (is there an easy way to set a breakpoint in OSMReader code?)
- [ ] add a dedicated test (should I edit one of the OSM test files or add a new one? Should I use another tag name or stick with "ele"?)

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -501,7 +501,7 @@ public class OSMReader implements TurnCostParser.ExternalInternalMap {
             if (eleString != null)
                 ele = Double.parseDouble(eleString);
         }
-        if (ele == Double.NaN) {
+        if (Double.isNaN(ele)) {
             ele = eleProvider.getEle(lat, lon);
         }
         if (nodeType == TOWER_NODE) {
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -109,6 +109,7 @@ public class OSMReader implements TurnCostParser.ExternalInternalMap {
     private Date osmDataDate;
     private final IntsRef tempRelFlags;
     private final TurnCostStorage tcs;
+    private String readElevationFromTag = null;
 
     public OSMReader(GraphHopperStorage ghStorage) {
         this.ghStorage = ghStorage;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -494,7 +495,15 @@ boolean addNode(ReaderNode node) {
...
         double lat = node.getLat();
         double lon = node.getLon();
-        double ele = eleProvider.getEle(lat, lon);
+        double ele = Double.NaN;
+        if (readElevationFromTag != null) {
+            String eleString = node.getTag(readElevationFromTag);
+            if (eleString != null)
+                ele = Double.parseDouble(eleString);
+        }
+        if (Double.isNaN(ele)) {
```
otbutz, line 504: I'm not really happy with this fallback mechanic. We might end up mixing values from different height models.
*If* we want to support this it's either tag based elevation or from an external provider.

Or to cite Ghostbuster: don't cross the streams

otbutz (changes requested):

easbar:
In the current form I also don't think this should be included in this repository. Mostly because it adds additional complexity as there are now two ways to read elevation. At the same time it can be easily done in a forked version of GraphHopper.

But I also wonder if we should pass the `ReaderNode` to `ElevationProvider`:

```java
public interface ElevationProvider {
    default double getEle(ReaderNode node) {
         return getEle(node.getLat(), node.getLon());
    }

    double getEle(double lat, double lon);
}
```

So for @gberaudo's use-case one would only implement the first method (and maybe just throw an error for the second). But this won't work with the long edge-sampling, which has to be disabled in this case. So not entirely clean, but I also do not see a real reason why `ElevationProvider`s should not get hold of the `ReaderNode`, just like `TagParsers` have access to the `ReaderWay`.

easbar:
Regarding the `ElevationProvider` interface, I think `setInterpolate`, `setDAType`, `setAutoRemoveTemporaryFiles` and `setBaseURL` all should not be there. They should rather be constructor parameters of the implementing subclasses (if needed). The abstraction of an 'elevation provider' is that it is something that provides an elevation, not something that you can set a `DAType` for etc. 

`getInterpolate` could be useful here, though. We could make sure that the long edge sampling never runs if `getInterpolate` is false. So for @gberaudo's use-case this would simply return false.

otbutz:
The ElevationProvider interface could really need some overhaul. Most if not all of the other methods should be moved into an abstract baseclass.

@easbar i think your solution is a lot cleaner and can be supported upstream.

gberaudo:
There is no consensus around the approach in this PR so I will close it.
Thanks @easbar. @otbutz for the discussions and feedbacks.

otbutz:
Not entirely correct. The consensus is described in #2374 which aims to provide a better `ElevationProvider` interface. This should allow you to easily implement your usecase without having to maintain a fork :wink:

Issue #2374, opened 2021-07-27
Cleanup ElevationProvider interface


otbutz:
The ElevationProvider interface carries methods which only make sense for certain implementations. We should move all methods except for `getEle(double, double)` into a fitting abstract baseclass for file/tile based providers.

otbutz:
Based on @easbar draft in https://github.com/graphhopper/graphhopper/pull/2373#issuecomment-887282112

```java
public interface ElevationProvider {

    double getEle(double lat, double lon);

    default double getEle(ReaderNode node) {
         return getEle(node.getLat(), node.getLon());
    }

    boolean isInterpolating();
}
```

easbar:
Yes. We might need `void release()` as well. Maybe `canInterpolate` instead of `isInterpolating`? And regarding the other setters I wonder why they even should be setters, they could just be constructor parameters. Anyway the most important part is cleaning up the interface.

otbutz:
I'd rather replace `release()` with extending the `AutoCloseable` interface.

easbar:
Ok, but maybe this is something for another issue, because we do not do this for our other resources (DataAccess etc.) yet either.

PR #2381, opened 2021-08-23
Refactor elevation provider

gberaudo:
Refactor elevation provider:

- clean up interface;
- add a getEle(node) variant with a default implementation.

Fixes: #2374.
Ping @otbutz, @easbar.

I did not introduce constructor parameters as it was a bit tricky with multiple constructors like in MultiplSourceElevationProvider.
The `AbstractElevationProvider` was alreading containing filesystem related code so I moved the methods there.
I have not renamed `getInterpolate` to `canInterpolate` yet. I will work on it later.

Please let me know your feedback of this PR.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -570,22 +570,10 @@ public class GraphHopper {
         if (ghConfig.has("graph.elevation.calcmean"))
             throw new IllegalArgumentException("graph.elevation.calcmean is deprecated, use graph.elevation.interpolate");
 
-        boolean interpolate = ghConfig.has("graph.elevation.interpolate")
-                ? "bilinear".equals(ghConfig.getString("graph.elevation.interpolate", "none"))
-                : ghConfig.getBool("graph.elevation.calc_mean", false);
-
         String cacheDirStr = ghConfig.getString("graph.elevation.cache_dir", "");
         if (cacheDirStr.isEmpty() && ghConfig.has("graph.elevation.cachedir"))
             throw new IllegalArgumentException("use graph.elevation.cache_dir not cachedir in configuration");
 
-        String baseURL = ghConfig.getString("graph.elevation.base_url", "");
-        if (baseURL.isEmpty() && ghConfig.has("graph.elevation.baseurl"))
-            throw new IllegalArgumentException("use graph.elevation.base_url not baseurl in configuration");
-
-        boolean removeTempElevationFiles = ghConfig.getBool("graph.elevation.cgiar.clear", true);
-        removeTempElevationFiles = ghConfig.getBool("graph.elevation.clear", removeTempElevationFiles);
-
-        DAType elevationDAType = DAType.fromString(ghConfig.getString("graph.elevation.dataaccess", "MMAP"));
         ElevationProvider elevationProvider = ElevationProvider.NOOP;
         if (eleProviderStr.equalsIgnoreCase("srtm")) {
             elevationProvider = new SRTMProvider(cacheDirStr);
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -596,16 +596,17 @@ private static ElevationProvider createElevationProvider(GraphHopperConfig ghCon
...
             elevationProvider = new SRTMGL1Provider(cacheDirStr);
         } else if (eleProviderStr.equalsIgnoreCase("multi")) {
-            elevationProvider = new MultiSourceElevationProvider(cacheDirStr);
+            MultiSourceElevationProvider provider = new MultiSourceElevationProvider(cacheDirStr);
+            provider
+                    .setAutoRemoveTemporaryFiles(removeTempElevationFiles)
+                    .setInterpolate(interpolate)
+                    .setDAType(elevationDAType);
+            if (!baseURL.isEmpty())
+                provider.setBaseURL(baseURL);
```
easbar, line 605: So far we called `setAutoRemoveTemporaryFiles`, `setInterpolate`, `setDAType` and `setBaseURL` for all the different elevation providers, but now you call these methods only for the `MultiSourceElevationProvider`? Is this intended, and why?

core/src/main/java/com/graphhopper/reader/dem/AbstractElevationProvider.java
```diff
@@ -68,35 +72,44 @@ void setSleep(long sleep) {
...
-        return cacheDir;
-    }
-
-    @Override
-    public ElevationProvider setBaseURL(String baseUrl) {
+    /**
+     * For the MultiSourceElevationProvider you have to specify the base URL separated by a ';'.
+     * The first for cgiar, the second for gmted.
+     */
+    public AbstractElevationProvider setBaseURL(String baseUrl) {
```
easbar, line 79: Can we keep the less specific javadoc of this method that we had in `ElevationProvider` here and put this more specific javadoc at the override in `MultiSourceElevationProvider`?

easbar (changes requested):
Thanks for this PR. See my comments inline.

> The AbstractElevationProvider was alreading containing filesystem related code so I moved the methods there.

Can we rename it to `F(T?)ileBasedElevationProvider` then?

Please also add a line to `CHANGELOG.md` and add yourself to `CONTRIBUTORS.md`.

otbutz:
I'd vote for **T**ileBasedElevationProvider

core/src/main/java/com/graphhopper/GraphHopper.java
gberaudo: @easbar, I addressed your requested changes. Is the PR good for you now? Let me know so that we can move forward.

gberaudo:
@otbutz, I am willing to address the concern you expressed. Could you please give me some inputs so that I can move forward?

otbutz:
Scratch that. Had a look at the current implementation and the MultiSourceElevationProvider only ties together a CIGAR and a GMTED provider. So this is far from a generic solution.

I'd go with `MultiSourceTileBasedElevationProvider extends TileBasedElevationProvider` and butcher the `TileBasedElevationProviderInterface`

So we'd end up with the following interfaces/classes:

* ElevationProvider (interface)
* TileBasedElevationProvider (abstract class)

gberaudo:
OK @otbutz, I reworked the PR with your idea.
@easbar, does it also look good for you?

core/src/main/java/com/graphhopper/reader/dem/MultiSourceElevationProvider.java
```diff
@@ -93,10 +101,25 @@ public void release() {
...
+    }
+
+    @Override
+    String getFileName(double lat, double lon) {
+        return null;
+    }
+
+    @Override
+    String getDownloadURL(double lat, double lon) {
+        return null;
```
otbutz, line 122: If these two methods are not supposed to be called, i'd rather throw an `UnsupportedOperationException`

otbutz (changes requested):

core/src/main/java/com/graphhopper/reader/dem/MultiSourceElevationProvider.java
gberaudo: OK @otbutz, done

config.yml_old
```diff
@@ -0,0 +1,100 @@
+graphhopper:
```
otbutz, line 1: Looks like an accidental addition

otbutz (changes requested):
Apart from the config.yml_old the PR is looking fine :)

config.yml_old
gberaudo: Sorry for that one. Now removed.

gberaudo:
Hi @easbar, you requested changes to the PR. I think I addressed all your demands. If the code looks good to you now would you be willing to merge the PR?

easbar:
> Hi @easbar, you requested changes to the PR. I think I addressed all your demands. If the code looks good to you now would you be willing to merge the PR?

Thanks for your changes. Yes, this PR looks good to me. There are a few minor things that we should fix (like missing overrides, duplicate javadocs in multi source provider and some whitespace/formatting), but I'll do this in a follow-up commit.

> Had a look at the current implementation and the MultiSourceElevationProvider only ties together a CIGAR and a GMTED provider. So this is far from a generic solution.

Yes, the name is a bit misleading indeed, I'll check if this can be improved, but this is unrelated to this PR.

easbar (approved):

gberaudo:
Thanks @easbar. I saw your follow up PR. Will also manage this part next time.

PR #2385, opened 2021-09-13
Remove InternalGraphEventListener

easbar:
More cleanups in the `CHGraph/RoutingCHGraph/GraphHopperStorage/GraphInterface/XYIteratorImpl` universe are on the way, but I think we can already get rid of this one :)


core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
```diff
@@ -65,22 +65,7 @@ public GraphHopperStorage(Directory dir, EncodingManager encodingManager, boolea
         this.dir = dir;
         this.properties = new StorableProperties(dir);
         this.segmentSize = segmentSize;
-        InternalGraphEventListener listener = new InternalGraphEventListener() {
-            @Override
-            public void initStorage() {
-                for (CHGraphImpl cg : chGraphs) {
-                    cg.initStorage();
-                }
-            }
```
easbar, line 74: Probably after some previous refactorings the `initStorage` call did not do anything that we cannot do in the constructor anymore so it is not needed at all.

core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
```diff
@@ -65,22 +65,7 @@ public GraphHopperStorage(Directory dir, EncodingManager encodingManager, boolea
...
-
-            @Override
-            public void freeze() {
-                for (CHGraphImpl cg : chGraphs) {
-                    cg._prepareForContraction();
```
easbar, line 79: No need for this there-and-back loop anymore. Right now we call freeze in GraphHopperStorage, which calls freeze in BaseGraph, which then executes the listener it got in the constructor earlier, and the listener calls `prepareForContraction`. But GraphHopperStorage can just as well simply call `prepareForContraction` directly. This is probably left over from some previous refactorings.

core/src/main/java/com/graphhopper/storage/CHGraphImpl.java
```diff
@@ -387,8 +394,13 @@ void initStorage() {
...
     public CHGraph create(long bytes) {
-        nodesCH.create(bytes);
-        shortcuts.create(bytes);
+        // creating the DataAccess here does not make much sense, because nodesCH has to have a certain size that we
+        // only know in prepareForContraction(). also the size of shortcuts is unknown and if anything it should be
+        // approximated when the number of base graph edges can be determined -> also in prepareForContraction()
+        // Anyway, we have to create the DataAccess here, because otherwise we get an error when calling flush() and
+        // loadExisting() later, see #2384
+        nodesCH.create(0);
+        shortcuts.create(0);
```
karussell, line 403: Maybe we should in general remove the size argument of this create method? And only create the basics or already one segment in DataAccess.

karussell (approved):
Thank you :) :+1:

core/src/main/java/com/graphhopper/storage/CHGraphImpl.java
easbar: Yes, good question. `x.create(bytes)` would already be possible by using `x.create(); x.ensureCapacity(bytes)`?

karussell: Yes, indeed. This should be identical.

easbar: -> https://github.com/graphhopper/graphhopper/issues/2386

Issue #1067, opened 2017-05-23
Allow decoupling of normal graph weights when using CH


karussell:
When edge weights update then CH still uses them although shortcuts are not updated. So after an update a recalculation of the shortcuts is necessary **or** a separate storage of the base graph edge weights is necessary. The last option requires more RAM (~4bytes*edges) but could also speed up CH routing as the Weighting.calcWeight method must not be called and instead something like the edge.getWeight could be used where a frozen value is stored.

See discussion [here](https://discuss.graphhopper.com/t/estimated-ram-needed-to-instantiate-two-graphhoppers-simultaneously/1939) and related to #240.

easbar:
Another (big I think) advantage of a separate storage for the base graph is that we can disconnect edges leading to or coming from lower levels, but this requires not only a separate storage for the weights, but rather for the edge connection information (nextA/nextB).

karussell:
Yes, this would be interesting. For a world wide graph caching the weights would add 0.5GB (per CH profil as we have around 100M edges) to currently 6-8GB and having nextA+nextB would add 3*0.5GB, which might be worth it if the speed is >20% (which tests indicates it is)

karussell:
Can we close here?

easbar:
> When edge weights update then CH still uses them although shortcuts are not updated.

Do we even update edge weights anywhere? 

> Another (big I think) advantage of a separate storage for the base graph is that we can disconnect edges leading to or coming from lower levels, but this requires not only a separate storage for the weights, but rather for the edge connection information (nextA/nextB).

We already did this, but only to some extent as we learned that the combined storage also has some benefits.

As far as I am concerned we can close here yes.

boldtrn:
> Do we even update edge weights anywhere?

Would be a really cool feature for stuff like live traffic 👍  But I think we are dreaming about this for some years now :)

karussell:
> Do we even update edge weights anywhere?

Not anymore. We had an endpoint for this purpose once, but we removed it. So the mistake could be done only if manually implemented, but then the responsibility would not be on our side and one needs to update the CH preparation.

Issue #1810, opened 2019-11-30
Bugs in node-based CH preparation heuristics


easbar:
So far there are two todos in `NodeBasedNodeContractor`, that refer to odd/false calculations regarding the CH preparation heuristics (also pointed out here: https://github.com/graphhopper/graphhopper/pull/1225#issue-153508946):

https://github.com/graphhopper/graphhopper/blob/0a967b3dd4f8a78d3cbe3bc0b64e820b950995d4/core/src/main/java/com/graphhopper/routing/ch/NodeBasedNodeContractor.java#L81-L86

https://github.com/graphhopper/graphhopper/blob/0a967b3dd4f8a78d3cbe3bc0b64e820b950995d4/core/src/main/java/com/graphhopper/routing/ch/NodeBasedNodeContractor.java#L312-L314

These are easy to fix, but fixing them probably requires re-optimizing the CH preparation parameters.

I think there is another one: One of the CH preparation heuristics counts the number of contracted neighbor nodes. Currently this is done like this:

https://github.com/graphhopper/graphhopper/blob/0a967b3dd4f8a78d3cbe3bc0b64e820b950995d4/core/src/main/java/com/graphhopper/routing/ch/NodeBasedNodeContractor.java#L116-L123

with:

https://github.com/graphhopper/graphhopper/blob/0a967b3dd4f8a78d3cbe3bc0b64e820b950995d4/core/src/main/java/com/graphhopper/routing/ch/NodeBasedNodeContractor.java#L69-L75

and:

https://github.com/graphhopper/graphhopper/blob/5c980b4627578626268873f9424de35188187441/core/src/main/java/com/graphhopper/routing/util/LevelEdgeFilter.java#L40-L52

A 'contracted neighbor' should be a node that is contracted already and connected to the current node via an edge or a shortcut. The `super.accept(edgeState)` in the `remainingNodesFilter` accepts all shortcuts and all edges that go to equal or higher CH level (we can ignore the `maxNodes` condition because there are no virtual edges during CH preparation). We increase the `contractedNeighbors` counter for every shortcut and do not increase it for non-shortcut edges. This seems wrong to me: there can be shortcuts leading to un-contracted nodes and there can be non-shortcuts leading to contracted nodes. 

So it should be something like this:

```java
// just an edge explorer using `DefaultEdgeFilter.allEdges`
EdgeIterator iter = allEdgeExplorer.setBaseNode(node);
while (iter.next()) {
[...]

PR #2387, opened 2021-09-15
Cleanup CH iterators, remove CHGraph etc.

easbar:
At some point `CHGraphImpl` was meant to act like `BaseGraph` with some additional shortcuts. The corresponding edge iterators like `CHEdgeIteratorState` inherit from the base graph iterators like `EdgeIteratorState`, even though many methods like `get(encoded_value)` are not available for shortcuts.

The routing algorithms use the `RoutingCHGraph` instead for some time now, see #1850, #1780, #1067, but the `CHGraph` classes are still used internally, which is quite a mess. This PR will finally remove `CHGraph(Impl)`, `CHEdgeExplorer`, `CHEdgeIterator` and `CHEdgeIteratorState`.


core/src/main/java/com/graphhopper/routing/ch/NodeBasedNodeContractor.java
```diff
@@ -30,7 +33,9 @@
 class NodeBasedNodeContractor implements NodeContractor {
     private final CHPreparationGraph prepareGraph;
     private final Params params = new Params();
-    private ShortcutHandler shortcutHandler;
+    // todonow: maybe use a set to prevent duplicates instead?
+    private List<Shortcut> shortcuts = new ArrayList<>();
```
karussell, line 37: A HashSet is more memory intense as the underlying array is 30%-60% empty but might be indeed faster. We would have to try this :)

karussell (commented):
Nice, thanks! Can't comment much but sounds really great :)

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -947,7 +947,7 @@ protected void postProcessing(boolean closeEarly) {
 
         if (sortGraph) {
             if (ghStorage.isCHPossible() && isCHPrepared())
-                throw new IllegalArgumentException("Sorting a prepared CHGraph is not possible yet. See #12");
+                throw new IllegalArgumentException("Sorting a prepared CHGraphImpl is not possible yet. See #12");
```
karussell, line 950: probably RoutingCHGraph?

core/src/main/java/com/graphhopper/storage/CHBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+
+    public int getShortcutForPrepareEdge(int prepareEdge) {
+        if (prepareEdge < origEdges)
+            return prepareEdge;
+        int index = prepareEdge - origEdges;
+        return shortcutsByPrepareEdges.get(index);
+    }
+
+    public void buildCH() {
+        // during contraction the skip1/2 edges of shortcuts refer to the prepare edge-ids *not* the CHGraphImpl (shortcut)
```
karussell, line 106: RoutingCHGraphImpl?

easbar: Yes, I thought about renaming `RoutingCHGraph` back to `CHGraph` once `CHGraph` is gone? Otherwise yes, there is still some clean up to do here.

core/src/main/java/com/graphhopper/routing/ch/NodeBasedNodeContractor.java
easbar: This todo has also been there for some time :) I just moved it here from another place.

karussell: ah, I thought you changed it from "todo" to "todonow"?

easbar:
> Nice, thanks! Can't comment much but sounds really great :)

Thanks. Yes this is hard to review and also not finished. I wanted to open this as a draft actually. Where is the 'convert to draft' button?

core/src/main/java/com/graphhopper/routing/ch/NodeBasedNodeContractor.java
easbar: Ah very true :) I probably thought it is time to try this and fix the todo. I ran into many things that I thought aren't ideal and it is hard to keep the focus on this specific refactoring. For now I'm happy all tests are green and `CHGraphImpl` is gone, but now I need to clean up more.

core/src/main/java/com/graphhopper/routing/ch/NodeBasedNodeContractor.java
```diff
@@ -131,29 +137,52 @@ public IntContainer contractNode(int node) {
      * with them.
      */
     private void insertShortcuts(int node) {
-        shortcutHandler.startContractingNode();
+        shortcuts.clear();
         {
```
otbutz, line 141: Can we remove those code blocks?

otbutz (changes requested):
NodeBaseNodeContractor.prepareContraction() is divides two ints which is probably not intended.

```java
meanDegree = (double) prepareGraph.getOriginalEdges() / prepareGraph.getNodes();
```

easbar:
> NodeBaseNodeContractor.prepareContraction() divides two ints which is probably not intended.

Yes, this is part of #1810.

core/src/main/java/com/graphhopper/routing/ch/NodeBasedNodeContractor.java
easbar: We could, but I think they are useful to limit the scope of the `iter` variable. But anyway I put both blocks into separate methods.

core/src/main/java/com/graphhopper/storage/CHBuilder.java
easbar: Hm, not sure yet, maybe do this later.., or at least not in this PR

easbar:
1107 LOC gone :)

PR #2389, opened 2021-09-20
Refine OSMGetOffBikeParser

OlafFlebbeBosch:
* No need to get off bike on paths
  * Adding bicycle/official and bicycle/permissive for riding

Please read [our contributing guide](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md) and note that also your contribution falls under the Apache License 2.0 as outlined there.

Your first contribution should include a change where you add yourself to the CONTRIBUTORS.md file.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMGetOffBikeParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMGetOffBikeParser.java
@@ -17,7 +17,7 @@ import java.util.List;
 public class OSMGetOffBikeParser implements TagParser {
 
     private final HashSet<String> pushBikeHighwayTags = new HashSet<>();
-    private final List<String> accepted = Arrays.asList("designated", "yes");
+    private final List<String> accepted = Arrays.asList("designated", "yes", "official", "permissive");
     private final BooleanEncodedValue offBikeEnc;
 
     public OSMGetOffBikeParser() {
--- a/core/src/test/java/com/graphhopper/routing/util/parsers/OSMGetOffBikeParserTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/parsers/OSMGetOffBikeParserTest.java
@@ -50,6 +50,10 @@ public class OSMGetOffBikeParserTest {
         assertFalse(isGetOffBike(way));
         way.setTag("bicycle", "designated");
         assertFalse(isGetOffBike(way));
+        way.setTag("bicycle", "official");
+        assertFalse(isGetOffBike(way));
+        way.setTag("bicycle", "permissive");
+        assertFalse(isGetOffBike(way));
 
         way = new ReaderWay(1);
         way.setTag("railway", "platform");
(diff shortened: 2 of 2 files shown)


karussell:
Thanks @OlafFlebbeBosch this looks good! This would need a review from @ratrun :) as I'm unsure why we had getting of the bike for paths.

boldtrn:
> I'm unsure why we had getting of the bike for paths

I always thought that the idea was that most paths are narrow and might require to get off the bike if you are an insecure rider or if there is traffic coming the opposite way - a bit like "you might have to get off the bike".

I would be in favour of removing path here :)

karussell:
Yes, although the wiki even explicitly states: "highway=path | A generic multi-use path open to non-motorized vehicles, implicitly allowed for pedestrians, cyclists, …"

if a path has foot=designated and no bicycle tags we should still mark get_off_bike as true.

boldtrn:
> if a path has foot=designated and no bicycle tags we should still mark get_off_bike as true (i.e. the default access is a bit tricky for path?)

Yes, probably you are right, as this would be identical to a `highway=footway` with no bicycles tags?

OlafFlebbeBosch:
@boldtrn @karussell : As I understand the code already works the way you are proposing ? Should I add explicit tests for the scenarios at hand ?

ratrun:
I didn't know that the default accessibility for `highway=path` implicitly allows bicycles. Here in Austria most small ways in the forest are tagged like this, but accesss for bikes is usually forbidden unlike it is explicitely allowed. So this was a localisation issue - see https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions#Austria.

karussell:
> I didn't know that the default accessibility for highway=path implicitly allows bicycles

Ah, ok. So this PR is fine for you?

> Should I add explicit tests for the scenarios at hand ?

Yes, if I understand it correctly, it would be good to have something like this:

```
way = new ReaderWay(1);
assertFalse(isGetOffBike(way));

way.setTag("footway", "yes");
assertTrue(isGetOffBike(way));
```

otbutz:
Sounds like this would be a better fit for CountryRules to be honest.

OlafFlebbeBosch:
Thanks @ratrun  for the insights into Austria specific rules. I wasn't aware of these!

I will update to reinstantiate the "path" in the get off bike sections again and will see into looking the proposal of @otbutz in an follow-up PR.

OlafFlebbeBosch:
Revoked the change to "path" for now. Adding a bit more tests. Hope it was ok to squash the changes

ratrun:
@OlafFlebbeBosch 
Thanks for changing the code for `path` again. I believe that the correct global default is implemented and therefore I'm fine with this PR now. 
@karussell Thanks for asking!

karussell:
> I believe that the correct global default is implemented 

How do you think should it behave without bicycle tags?

I do not think this works in this PR:

```
way.setTag("footway", "yes");
assertTrue(isGetOffBike(way));
```

boldtrn:
> way.setTag("footway", "yes");

Do you mean `foot=designated` or `highway=footway`?

karussell:
Ah, yes "foot=designated" is probably what I meant - and if this works already then I'm fine :)

OlafFlebbeBosch:
Hi,  I am adding @karussell proposed test explicitly.
Sorry something went wrong while pushing the previous update. The change should be in merge-able state now.

karussell:
Thanks @OlafFlebbeBosch !

PR #2391, opened 2021-09-22
Remove EncodingManager from BaseGraph, cleanup nodes and edges header

easbar:
Looks like we do not need the EncodingManager in BaseGraph, see comments inline

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/BaseGraph.java
+++ b/core/src/main/java/com/graphhopper/storage/BaseGraph.java
@@ -58,7 +58,6 @@ class BaseGraph implements Graph {
     // can be null if turn costs are not supported
     final TurnCostStorage turnCostStorage;
     final BitUtil bitUtil;
-    final EncodingManager encodingManager;
     private final int intsForFlags;
     // length | nodeA | nextNode | ... | nodeB
     // as we use integer index in 'egdes' area => 'geometry' area is limited to 4GB (we use pos&neg values!)
@@ -89,11 +88,9 @@ class BaseGraph implements Graph {
     private long maxGeoRef;
     private boolean frozen = false;
 
-    public BaseGraph(Directory dir, final EncodingManager encodingManager, boolean withElevation,
-                     boolean withTurnCosts, int segmentSize) {
+    public BaseGraph(Directory dir, int intsForFlags, boolean withElevation, boolean withTurnCosts, int segmentSize) {
         this.dir = dir;
-        this.encodingManager = encodingManager;
-        this.intsForFlags = encodingManager.getIntsForFlags();
+        this.intsForFlags = intsForFlags;
         this.bitUtil = BitUtil.get(dir.getByteOrder());
         this.wayGeometry = dir.find("geometry");
         this.stringIndex = new StringIndex(dir);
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -175,51 +172,45 @@ void checkInitialized() {
             throw new IllegalStateException("The graph has not yet been initialized.");
     }
 
-    protected int loadNodesHeader() {
-        nodeEntryBytes = nodes.getHeader(1 * 4);
-        nodeCount = nodes.getHeader(2 * 4);
```
easbar, line 180: Don't know why we start the header at `1*4` not `0*4`, so I changed this, and it seems to work.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -175,51 +172,45 @@ void checkInitialized() {
...
 
         if (bounds.hasElevation()) {
-            bounds.minEle = Helper.intToEle(nodes.getHeader(7 * 4));
-            bounds.maxEle = Helper.intToEle(nodes.getHeader(8 * 4));
+            bounds.minEle = Helper.intToEle(nodes.getHeader(6 * 4));
+            bounds.maxEle = Helper.intToEle(nodes.getHeader(7 * 4));
         }
 
-        frozen = nodes.getHeader(9 * 4) == 1;
-        return 10;
```
easbar, line 192: The return value of these methods was never used anyway, they can also be private.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -175,51 +172,45 @@ void checkInitialized() {
...
 
-    protected int setNodesHeader() {
-        nodes.setHeader(1 * 4, nodeEntryBytes);
-        nodes.setHeader(2 * 4, nodeCount);
-        nodes.setHeader(3 * 4, Helper.degreeToInt(bounds.minLon));
-        nodes.setHeader(4 * 4, Helper.degreeToInt(bounds.maxLon));
-        nodes.setHeader(5 * 4, Helper.degreeToInt(bounds.minLat));
-        nodes.setHeader(6 * 4, Helper.degreeToInt(bounds.maxLat));
+    private void setNodesHeader() {
+        nodes.setHeader(0 * 4, nodeEntryBytes);
```
easbar, line 192: Same here: We can start at `0` instead of `4`

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -175,51 +172,45 @@ void checkInitialized() {
...
         edgeCount = edges.getHeader(1 * 4);
-        return 5;
     }
 
-    protected int setEdgesHeader() {
+    protected void setEdgesHeader() {
         edges.setHeader(0, edgeEntryBytes);
         edges.setHeader(1 * 4, edgeCount);
-        edges.setHeader(2 * 4, encodingManager.hashCode());
-        edges.setHeader(3 * 4, supportsTurnCosts() ? turnCostStorage.hashCode() : -1);
```
easbar, line 221: Here we can get rid of the encodingManager, because we write this header, but never read it again. Or am I missing something?

core/src/main/java/com/graphhopper/util/Constants.java
```diff
@@ -66,8 +66,8 @@
     private static final int JVM_MAJOR_VERSION;
     private static final int JVM_MINOR_VERSION;
 
-    public static final int VERSION_NODE = 5;
-    public static final int VERSION_EDGE = 19;
+    public static final int VERSION_NODE = 6;
+    public static final int VERSION_EDGE = 20;
```
easbar, line 70: I updated these because the nodes/edges header was changed.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -175,51 +172,45 @@ void checkInitialized() {
...
-        nodes.setHeader(9 * 4, isFrozen() ? 1 : 0);
-        return 10;
+        nodes.setHeader(8 * 4, isFrozen() ? 1 : 0);
     }
 
-    protected int loadEdgesHeader() {
+    protected void loadEdgesHeader() {
         edgeEntryBytes = edges.getHeader(0 * 4);
         edgeCount = edges.getHeader(1 * 4);
-        return 5;
```
easbar, line 214: We never use this return value (just like the others), but `5` does not even seem to be correct?  It should be `2` maybe (the number of ints loaded).

PR #2402, opened 2021-09-27
Add Austria specific rule for dismounting bike at paths

OlafFlebbeBosch:
Working on #2389 there was the insight we should add a contry specific rule for Austria:
* Bicycle riders have to push their bike on "highway=path" , unless it is explicitly allowed in Austria.
* Changing the rules in a way to push bikes on "highway=path" when it is a "foot=designated" path, unless biking is explicitly allowed.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/countryrules/AustriaCountryRule.java
+++ b/core/src/main/java/com/graphhopper/routing/util/countryrules/AustriaCountryRule.java
@@ -72,4 +72,12 @@ public class AustriaCountryRule implements CountryRule {
                 return RoadAccess.YES;
         }
     }
+
+    @Override
+    public boolean getOffBike(ReaderWay readerWay, boolean currentOffBike) {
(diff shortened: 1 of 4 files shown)


austrianton:
I would like to add, that due to Austrian forestry law, bicycle=no and vehicle=forestry should be treated the same way.

The only way to legally have a bike on a forest road, is to carry it on your back (so actually also no pushing)

austrianton:
The routing seems to now avoid paths, that don't have an explicit bicycle=yes tag, for bikes. So it seems to be implemented.

oflebbe:
Oops yes, I have an idea 👍 .  I wanted to merge accounts, but the outcome is not what I was expecting. It is both me.
Nevertheless, my contribution in this PR went into git as 2e0016d84d992d3dddb0d7d4d06d2ffe360fed4a . 
We already implemented a countryspecfic rule for dismounting bikes IIRC in this commit .

PR #2405, opened 2021-09-27
Load CountryRules via ServiceLoader

### LINKED ISSUE #2404: Improve CountryRuleFactory
Currently the association between Country and CountryRule is hardcoded in the CountryRuleFactory with both available rules carrying a static instance of themselves like e.g

```java
public final static AustriaCountryRule RULE = new AustriaCountryRule();
```

This is fine for a few countries but gets clunky if we add coverage for more. I'd propose that we switch to using ServiceLoader instead and let the CountryRules tell us which country they are valid for.

```java
public class AustriaCountryRule implements CountryRule {
    
    @Override
    public Country getCountry() {
        return Country.AUT;
    }
```

The resulting factory code would be a lot more dynamic:

```java
public class CountryRuleFactory {
    
    private final Map<Country, CountryRule> rules;
    
    public CountryRuleFactory() {
        Map<Country, CountryRule> rulesMap = new EnumMap<>(Country.class);
        ServiceLoader<CountryRule> loader = ServiceLoader.load(CountryRule.class);
        for (Iterator<CountryRule> iter = loader.iterator(); iter.hasNext(); ) {
            CountryRule rule = iter.next();
            if (rule.getCountry() == null) {
                throw new IllegalStateException("CountryRule needs to define a countrycode.");
            }
            rulesMap.put(rule.getCountry(), rule);
        }
        this.rules = Collections.unmodifiableMap(rulesMap);
    }

    public CountryRule getCountryRule(Country country) {
        return rules.get(country);
    }
}
```

otbutz:
Fixes #2404

I've only added  a check for the CountryCode not being null as ServiceLoader.iterator() throws an exception for malformed entries and missing or uninstantiable classes: https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html#iterator--

We can also add a unittest to check if all rules are added to the service definition file.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/countryrules/AustriaCountryRule.java
+++ b/core/src/main/java/com/graphhopper/routing/util/countryrules/AustriaCountryRule.java
@@ -25,7 +25,6 @@ import com.graphhopper.routing.ev.RoadClass;
 import com.graphhopper.routing.util.TransportationMode;
 
 public class AustriaCountryRule implements CountryRule {
-    public final static AustriaCountryRule RULE = new AustriaCountryRule();
     
     @Override
     public Country getCountry() {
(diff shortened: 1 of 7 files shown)


otbutz:
See #2418

PR #2407, opened 2021-09-28
Fix loading localizations from ios bundle

OlKir:
Because in iOS app bundle resources (including localization files) placed differently as in web version `TranslationMap.class.getResourceAsStream` returns `null` instead of stream. To fix this add "/" if the first attempt failed.


karussell:
Would it be better if we allow setting TranslationMap in the GraphHopper class? Then you could use `TranslationMap.doImport(File folder)` directly.

OlKir:
Thank you for the hint! Yes, proposed approach will be more universal and future-proof.
With my quick fix I just tried to be as less intrusive as possible. I will try your suggestion and if I manage to implement it I will close this PR and create another one.

PR #2409, opened 2021-09-28
Allow setting custom TranslationMap for GraphHopper.

OlKir:
Default implementation of GraphHopper uses TranslationMap.doImport() without
parameters which tries to load localization files from path based on class
file location. In case of usage GraphHopper on different platform
(e.g. static binary library on iOS) we need some way to provide
a different path. TranslationMap already contains doImport(File folder)
method and this change allows to pass created TranslationMap to GraphHopper.


karussell:
I just introduced a smaller change: https://github.com/graphhopper/graphhopper/commit/b50e82d4188f30f18d6d8b10ba52384430fd44e4
If this is fine let's close here.

PR #2411, opened 2021-09-29
Consider heading when snapping

### LINKED ISSUE #1463: If heading is provided change EdgeFilter for LocationIndexTree
The issue can be reproduced with oneway streets like [this](https://graphhopper.com/maps/?point=51.211414%2C14.566305&point=51.211521%2C14.561906&ch.disable=true&heading=270), ie. the heading does not work. Expected is the snapping to the other side of the motorway but it is unclear if the heading should already be applied at the snapping step. (The ordinary case works like [here](https://graphhopper.com/maps/?point=51.192725%2C14.56094&point=51.192819%2C14.567678&ch.disable=true&heading=270))

karussell:
Fixes #1463


core/src/main/java/com/graphhopper/routing/Router.java
```diff
@@ -159,7 +160,7 @@ private void checkIfPointsAreInBounds(List<GHPoint> points) {
     }
 
     private void checkHeadings(GHRequest request) {
-        if (request.getHeadings().size() > 1 && request.getHeadings().size() != request.getPoints().size())
```
karussell, line 162: was this intended or a bug?

core/src/main/java/com/graphhopper/routing/Router.java
```diff
@@ -380,10 +384,15 @@ protected void checkProfileCompatibility() {
...
 
-        protected EdgeFilter getSnapFilter() {
+        protected EdgeFilter createSnapFilter() {
             return new DefaultSnapFilter(weighting, lookup.getBooleanEncodedValue(Subnetwork.key(profile.getName())));
         }
 
+        protected EdgeFilter createDirectedSnapFilter() {
+            BooleanEncodedValue inSubnetworkEnc = lookup.getBooleanEncodedValue(Subnetwork.key(profile.getName()));
+            return edgeState -> !edgeState.get(inSubnetworkEnc) && Double.isFinite(weighting.calcEdgeWeightWithAccess(edgeState, false));
+        }
```
karussell, line 394: unfortunately we need yet another EdgeFilter as we need to check if in which direction the Weighting works

core/src/test/java/com/graphhopper/routing/HeadingRoutingTest.java
```diff
@@ -199,11 +258,11 @@ private GraphHopperStorage createSquareGraph() {
         EncodingManager encodingManager = new EncodingManager.Builder().add(carEncoder).add(Subnetwork.create("profile")).build();
         GraphHopperStorage g = new GraphBuilder(encodingManager).create();
 
-        //   2---3---4
-        //  /    |    \
-        //  1----8----5
-        //  /    |    /
-        //  0----7---6
```
karussell, line 206: important fix to know the angle from the ascii chart ;)

core/src/main/java/com/graphhopper/routing/Router.java
easbar: You mean it should be `>0`? I think we want to check that there are either no headings or as many as there are points.

core/src/main/java/com/graphhopper/routing/HeadingResolver.java
```diff
@@ -59,16 +60,46 @@ public IntArrayList getEdgesWithDifferentHeading(int baseNode, double heading) {
...
+            }
+        }
+        if (closestPoint >= 0) {
+            double fromLat = points.getLat(closestPoint - 1), fromLon = points.getLon(closestPoint - 1);
+            double toLat = points.getLat(closestPoint), toLon = points.getLon(closestPoint);
+            double orientation = AngleCalc.ANGLE_CALC.calcOrientation(fromLat, fromLon, toLat, toLon);
+            orientation = AngleCalc.ANGLE_CALC.alignOrientation(xAxisAngle, orientation);
+            return Math.abs(orientation - xAxisAngle) < Math.toRadians(angleDifference);
+        }
+        return false;
```
otbutz, line 95: Early out?

```suggestion
        if (closestPoint == -1) {
            return false;
        }
        double fromLat = points.getLat(closestPoint - 1), fromLon = points.getLon(closestPoint - 1);
        double toLat = points.getLat(closestPoint), toLon = points.getLon(closestPoint);
        double orientation = AngleCalc.ANGLE_CALC.calcOrientation(fromLat, fromLon, toLat, toLon);
        orientation = AngleCalc.ANGLE_CALC.alignOrientation(xAxisAngle, orientation);
        return Math.abs(orientation - xAxisAngle) < Math.toRadians(angleDifference);
```

core/src/main/java/com/graphhopper/routing/HeadingResolver.java
```diff
@@ -59,16 +60,46 @@ public IntArrayList getEdgesWithDifferentHeading(int baseNode, double heading) {
...
+     * This method returns true if the specified heading is parallel to the specified edgeState (antiparallel isn't tested).
+     * Note that only the road segments near the specified pointNearHeading are checked for parallelism (<20m) and that
+     * a angle difference of 30° is accepted.
+     */
+    public static boolean isHeadingNearlyParallel(EdgeIteratorState edgeState, double heading, GHPoint pointNearHeading) {
+        DistanceCalc calcDist = DistanceCalcEarth.DIST_EARTH;
+        double xAxisAngle = AngleCalc.ANGLE_CALC.convertAzimuth2xaxisAngle(heading);
+        PointList points = edgeState.fetchWayGeometry(FetchMode.ALL);
+        int closestPoint = -1;
+        // TODO should we use the same default like for gpx_accuracy which is 40m?
```
otbutz, line 73: I'd say yes because the heading is most likely reported from a device using GPS data.

core/src/main/java/com/graphhopper/routing/Router.java
easbar: Ah no this is correct. We can specify a start heading also for routing with via points (without specifying a heading for all via points).

core/src/main/java/com/graphhopper/routing/HeadingResolver.java
karussell: IMO the `if` is short enough and the `return`s are closer this way.

core/src/main/java/com/graphhopper/routing/HeadingResolver.java
karussell: Yes, the data also comes from GPS. Still the use case is different IMO (not 100%). For map matching the number of candidates increases with a bigger gpx_accuracy which could lead to a more precise overall match (let's neglect performance for this). For heading the quality can decrease arbitrarily when we increase this too high (as we reject closer roads with a different heading). Let's assume the heading is also picked from the device than it might be bad for slow speed (ok, we could argue that the application could determine these bad situations and remove the heading parameter)

I propose to use a different value as a start (and increase this to e.g. 30m and later 40m if not sufficient in practice).

otbutz: > For heading the quality can decrease arbitrarily when we increase this too high (as we reject closer roads with a different heading).

Isn't that the point of the heading parameter?

> Let's assume the heading is also picked from the device than it might be bad for slow speed (ok, we could argue that the client could determine these bad situations and remove the heading parameter)

Bad GPS precision is far more likely e.g if the signal is reflected/blocked

It would be nice if the heading parameter would be used to prioritize fitting edges and not enforced as a hard filter.

core/src/main/java/com/graphhopper/routing/HeadingResolver.java
```diff
@@ -87,9 +97,18 @@ public static boolean isHeadingNearlyParallel(EdgeIteratorState edgeState, doubl
...
-            return Math.abs(orientation - xAxisAngle) < Math.toRadians(angleDifference);
+            if (Math.abs(orientation - xAxisAngle) < Math.toRadians(angleDifference))
+                return true;
+
+            if (parallel && directedEdgeFilter.accept(edgeState.detach(true))) {
+                orientation = AngleCalc.ANGLE_CALC.calcOrientation(toLat, toLon, fromLat, fromLon);
+                orientation = AngleCalc.ANGLE_CALC.alignOrientation(xAxisAngle, orientation);
+                return Math.abs(orientation - xAxisAngle) < Math.toRadians(angleDifference);
+            }
         }
```
karussell: @easbar can you have a look if this sounds better to you? (now the distance calculation needs to be done once and the angle calculation too ... still in rare cases the opposite angle calculation is required, see test)

core/src/main/java/com/graphhopper/routing/HeadingResolver.java
```diff
@@ -78,6 +86,8 @@ public static boolean isHeadingNearlyParallel(EdgeIteratorState edgeState, doubl
             double distance = calcDist.validEdgeDistance(pointNearHeading.lat, pointNearHeading.lon, fromLat, fromLon, toLat, toLon)
                     ? calcDist.calcDenormalizedDist(calcDist.calcNormalizedEdgeDistance(pointNearHeading.lat, pointNearHeading.lon, fromLat, fromLon, toLat, toLon))
                     : calcDist.calcDist(fromLat, fromLon, pointNearHeading.lat, pointNearHeading.lon);
+            if (i == points.size() - 1)
+                distance = Math.min(distance, calcDist.calcDist(toLat, toLon, pointNearHeading.lat, pointNearHeading.lon));
```
karussell, line 90: geometry is a bitch :D

core/src/main/java/com/graphhopper/routing/HeadingResolver.java
easbar: Hm, `isHeadingParallel` looks really complicated? And there is no test (that just tests this method) either. I still wonder why there is no method that just calculates the angle of the geometry near the query point. It would be immediately clear what this method does and it would be really easy to test as well.

`isHeadingNearlyParallel` on the other is hard to understand (for me). Like, what is this directedEdgeFilter argument supposed to do? For me this method just does too many things and therefore I find it hard to understand. Maybe this is more complicated than I thought and I am missing some details, will send a PR with my suggestion.

karussell: (Fix via #2411)

karussell:
Thanks @easbar & @otbutz!

easbar:
Thanks! This is a nice feature improvement!

Issue #2386, opened 2021-09-13
Remove the `bytes` argument in `Storable.create(bytes)`


easbar:
See here: https://github.com/graphhopper/graphhopper/pull/2385/files#r707181772

Possibly related: #1792

PR #2412, opened 2021-09-29
Convert `Directory#find` to `Directory#create`

### LINKED ISSUE #1792: Move setting of segment size into `DataAccess` constructors

easbar:
We currently have `Directory#find` which should rather be called `Directory#findOrCreate`. However, there are only very few cases where we actually make use of the `find` part and these look rather like `find` was used this way accidentally. The underlying problem is that generally there is a 1:1 correspondence between `DataAccess` objects and files on disk, so for every file there can only be one such object. There are some exceptions to this rule: 

* `RAM(_INT)` DataAccess aren't really related to files at all because they can never be flushed to disk
* even `RAM_STORE` DataAccess objects might live only in memory and only if they are flushed they actually correspond to a file on disk.

But anyway, making sure each name is used only once does make some sense, because this way we definitely avoid the case where multiple DataAccess could be flushed to the same file (and overwrite each other). Well, at least as long all DataAccess objects are created using the same `Directory` that is.

Our `Directory#find` method does *not* give such a guarantee though. It makes sure that no two DataAccess objects with the same name are created, but code calling `Directory#find` might still accidentally receive a reference to a DataAccess object that is already in use. In this case we might accidentally overwrite data while it is *in memory*.

There is another issue with `Directory#find`: It has a `daType` parameter, but the `daType` is only useful for the case `find()` **creates** a new object. If it just returns an existing one it is useless, and we are forced to pass 'the right' daType which is just a very annoying API. To make things worse there is also `segmentSize` which we would have to deal with the same way if we wanted to remove the (rather pointless) `setSegmentSize` setter. `setSegmentSize` is problematic, because the segment size cannot be changed after the DataAccess was created (see #1792).

In this PR I removed the `find` part of `Directory#find` and turned it into `Directory#create`. It is now illegal to use the same name multiple times. This way there cannot be multiple DataAccess objects with the same name (as long as they are created using Directory) and flushing to disk should be save in this regard.

I do not think this is the end of this story, but certainly a step into the right direction. I think after this PR we can easily fix #1792 and #2386,  and should be able to remove `DataAccess#create` and `DataAccess#loadExisting` (c.f. #2403).

### DIFF
--- a/reader-gtfs/src/main/java/com/graphhopper/gtfs/GraphHopperGtfs.java
+++ b/reader-gtfs/src/main/java/com/graphhopper/gtfs/GraphHopperGtfs.java
@@ -125,7 +125,8 @@ public class GraphHopperGtfs extends GraphHopper {
                 throw new RuntimeException("Error while constructing transit network. Is your GTFS file valid? Please check log for possible causes.", e);
             }
             streetNetworkIndex.close();
-            LocationIndexTree locationIndex = new LocationIndexTree(getGraphHopperStorage(), getGraphHopperStorage().getDirectory());
+            // we have to set an explicit location index name, because the default one is already taken
+            LocationIndexTree locationIndex = new LocationIndexTree(getGraphHopperStorage(), getGraphHopperStorage().getDirectory(), "location_index_pt");
             PtEncodedValues ptEncodedValues = PtEncodedValues.fromEncodingManager(getEncodingManager());
             EnumEncodedValue<GtfsStorage.EdgeType> typeEnc = ptEncodedValues.getTypeEnc();
             locationIndex.prepareIndex(edgeState -> edgeState.get(typeEnc) == GtfsStorage.EdgeType.HIGHWAY);
--- a/core/src/main/java/com/graphhopper/coll/OSMIDMap.java
+++ b/core/src/main/java/com/graphhopper/coll/OSMIDMap.java
@@ -45,9 +45,9 @@ public class OSMIDMap implements LongIntMap {
     public OSMIDMap(Directory dir, int noNumber) {
         this.dir = dir;
         this.noEntryValue = noNumber;
-        keys = dir.find("osmid_map_keys");
+        keys = dir.create("osmid_map_keys");
         keys.create(2000);
-        values = dir.find("osmid_map_values");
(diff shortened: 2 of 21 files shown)


reader-gtfs/src/main/java/com/graphhopper/gtfs/GraphHopperGtfs.java
```diff
@@ -125,7 +125,8 @@ protected void importPublicTransit() {
                 throw new RuntimeException("Error while constructing transit network. Is your GTFS file valid? Please check log for possible causes.", e);
             }
             streetNetworkIndex.close();
-            LocationIndexTree locationIndex = new LocationIndexTree(getGraphHopperStorage(), getGraphHopperStorage().getDirectory());
+            // we have to set an explicit location index name, because the default one is already taken
+            LocationIndexTree locationIndex = new LocationIndexTree(getGraphHopperStorage(), getGraphHopperStorage().getDirectory(), "location_index_pt");
```
easbar, line 129: This is the most interesting change: For PT we first create a location index (just in memory, but it still has a DataAccess name, so a corresponding, hypothetical file), then use it, but before actually flushing it we close/remove it and then build **another** location index. 

Previously we were able to `find` the underlying (closed) DataAccess in the directory again and then create it again. And although this might not have been so obvious we were overwriting the old one. Now we need to set a new name here or, alternatively, we would not only have to close `streetNetworkIndex` but also remove it from the Directory.

Actually, this change might not be ok yet, because this way the location index file will end up as "location_index_pt" where previously it was just "location_index". Will have to check this again.

easbar: > Actually, this change might not be ok yet, 

This is now fixed.

karussell: :+1:

PR #2413, opened 2021-09-30
HMM-Lib: Bump Java version to 8

boldtrn:
We switched to Junit5 some time ago. Junit5 requires Java 8. The hmm-lib is still on Java 7. 

When you try to execute the `ViterbiAlgorithmTest` in InteliJ, there is a build error, as the pom sets the language level to Java 7.

```
/graphopper/graphhopper/hmm-lib/src/test/java/com/bmw/hmm/ViterbiAlgorithmTest.java:153:9
java: cannot access java.util.function.BooleanSupplier
  class file for java.util.function.BooleanSupplier not found
```

Bumping the version to Java 8 fixes that problem. Running through the command line works fine, so I am a bit unsure about this change, but I guess bumping the version should not hurt?

### DIFF
--- a/hmm-lib/pom.xml
+++ b/hmm-lib/pom.xml
@@ -57,17 +57,4 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     </properties>
-
-    <build>
-        <plugins>
-            <plugin>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.1</version>
-                <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
-                </configuration>
-            </plugin>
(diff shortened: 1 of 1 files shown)


easbar:
Hm, I can run the test in IntelliJ without a problem. But I also think it is/should be no problem to upgrade to Java 8

hmm-lib/pom.xml
```diff
@@ -64,8 +64,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
```
easbar, line 64: Can you just remove the entire plugin, please? It will be taken from parent pom then.

boldtrn:
Yes I always thought it might be related to my setup, but I double checked on a second device and I saw the same result. I use the Maven integration of IntelliJ.

I updated the PR to remove the compiler plugin :)

easbar:
Thanks

PR #2414, opened 2021-09-30
country_rules: fix documentation in example config

karussell:
I think that the documentation does not match the code. At the moment the default is "disabled country rules".


easbar:
Yes, thanks!

PR #2418, opened 2021-10-01
Register country rules in a map

otbutz:
* avoids the need for static `RULE` instances
* allows lib users to easily manipulate registered rules without having to implement their own factory

See discussion in https://github.com/graphhopper/graphhopper/issues/2404


easbar:
Thanks!

PR #2420, opened 2021-10-01
Don't ignore parsed access

otbutz:
The OSMRoadAccessParser currently ignores the parsed access and always passes RoadAccess.YES to the CountryRule

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
@@ -59,7 +59,7 @@ public class OSMRoadAccessParser implements TagParser {
 
         CountryRule countryRule = readerWay.getTag("country_rule", null);
         if (countryRule != null)
-            accessValue = countryRule.getAccess(readerWay, TransportationMode.CAR, YES);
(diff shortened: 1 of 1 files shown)


otbutz:
@easbar i think this might be related to the CH slowdown you mentioned in https://github.com/graphhopper/graphhopper/pull/2353#issuecomment-886518367

Not sure if this would make it better or worse though.

karussell:
Thanks! (this has probably not something to do with the slow down, but will get to it at a later point I think :) )

easbar:
The slowdown occurs, because I think GermanyCountryRule sets all tracks to `access=destination`.

karussell:
Should we avoid this destination access (I'm unsure if this is really the correct legal situation). And assume a slower speed in CarFlagEncoder.trackTypeSpeedMap?

otbutz:
I'm not that happy with the current rule. Blindly assuming `access=destination` for *all* tracks isn't really the best idea. If OSM doesn't have any restriction, we shouldn't artificially create one which are not covered by any law etc.

Driving on an unrestricted track of grade1 or even grade2 is perfectly fine IMHO:

https://wiki.openstreetmap.org/wiki/Key:tracktype#Values

PR #2424, opened 2021-10-05
Handle way splitting at junctions and barriers in one place

easbar:
So far we first split the OSM ways in OSMReader into segments at barrier nodes and then called `addOSMWay` for each segment. In `addOSMWay` we further split these segments at junctions and loop nodes. Here I changed this such that we first split the ways at junctions **and** barrier nodes. The loop handling is still done afterwards (not sure about this yet).

This change auto-fixed the issue that we were not able to handle certain types of loop edges. We now split all loops into two pieces except loop segments consisting of only two nodes, which we now ignore. Previously we had to stop adding edges for the current way when we ran into such a case.

With this change we also ignore barrier nodes at junctions, which are mapping errors. We used to add a single barrier edge also in this case, even though it is not well-defined which one should be added. With this PR we simply print a warning and ignore the barrier.

I think the PR makes OSMReader at least a bit easier to understand, but the main motivation is getting rid of the node flags map.


easbar:
This still ignores too many barrier nodes, mostly barrier nodes between ways like this:

```
way1     barrier node     way2
-----------------x-----------
```

They are tower nodes, but not 'junctions'.

PR #2427, opened 2021-10-08
explicitly treat unhewn_cobblestone as a bad surface

matkoniecz:
https://wiki.openstreetmap.org/wiki/Tag:surface%3Dunhewn_cobblestone
https://taginfo.openstreetmap.org/tags/surface=unhewn_cobblestone#chronology

My contribution falls under the Apache License 2.0

Untested.

I still need to follow https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md fully - I just noticed that this surface is still missing while updating my presentation slides

Disclaimer: I am author of this OSM wiki page and contributor to StreetComplete, one of main promoters of that tag

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -48,7 +48,7 @@ Here is an overview:
  * kodonnell, adding support for CH and other algorithms (#60) and penalizing inner-link U-turns (#88)
  * legraina, improved docker for dockerhub
  * lmar, improved instructions
- * matkoniecz, tweaking documentation
+ * Mateusz Konieczny (matkoniecz), tweaking documentation and minor fixes
  * manueltimita, fixes like #1651
  * mathstpierre, fixes like #1753
  * michalmac, fixes like #1467
--- a/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
+++ b/core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
@@ -112,7 +112,8 @@ abstract public class BikeCommonFlagEncoder extends AbstractFlagEncoder {
 
         setSurfaceSpeed("paved", 18);
         setSurfaceSpeed("asphalt", 18);
-        setSurfaceSpeed("cobblestone", 8);
+        setSurfaceSpeed("unhewn_cobblestone", 4);
+        setSurfaceSpeed("cobblestone", 10);
         setSurfaceSpeed("cobblestone:flattened", 10);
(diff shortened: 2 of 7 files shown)


matkoniecz:
Note that I am unsure how

```
core/src/main/java/com/graphhopper/routing/ev/Surface.java
30:    PAVED("paved"), ASPHALT("asphalt"), CONCRETE("concrete"), PAVING_STONES("paving_stones"), COBBLESTONE("cobblestone"),
```

should be handled

easbar:
> no idea is it related or not

this is unrelated

> but https://github.com/matkoniecz/graphhopper/blob/master/core/src/test/java/com/graphhopper/routing/util/Bike2WeightFlagEncoderTest.java has not place where 8 is used. Or "cobblestone"

The test fails, because you changed the speed for `cobblestone` from 8 to 10 in `BikeCommonFlagEncoder`, maybe this was accidental?

> Note that I am unsure how

The `Surface` encoded value is not directly related to the speeds you adjusted in the bike flag encoders. The `Surface` encoded value simply stores the surface of a way so it can be displayed as path detail for example. The speed used for the route calculation is taken from the flag encoders, however. The surface can be used to define the speed and affect the route calculations via custom models though. We are currently working on making this more consistent.

Also note that the memory required to store the surface depends on the number of distinct surface values. Currently there are 15 values so 4 bits are sufficient. Once this exceeds 16 we need 8 bits. We currently support the [14 most surface tags](https://taginfo.openstreetmap.org/keys/surface#values) except the 12th (sett), which we map to `cobblestone` plus 'missing' and 'other'. The next three popular ones are: `wood` (which we map to `unpaved`), `earth` (which we map to `dirt`), and `concrete:plates` (which is the first one that we treat as `other`). `unhewn_cobblestone` is currently number 26 in the list (and mapped to `other` just like all the others).

Our goal is not to use a 1:1 relationship between GraphHopper encoded values and OSM tag values, but rather consolidate the different tags so they can be used more easily in custom models. So memory usage is not the only reason we are trying to keep the number of values low. Since you are one of the main promoters of this tag, maybe you can suggest which of our existing values is a good fit for `unhewn_cobblestone`, or do you think it should be treated separately (for the purpose of routing)? And if we add a new value, maybe the existing ones can be consolidated in another way. You can find the tag-to-value conversion in `OSMSurfaceParser`.

matkoniecz:
if I understand it well then unknown surface is already treated as a very bad surface, so in this case explicit support is not needed

PR #2428, opened 2021-10-08
Minor improvement to avoid changing the url

boldtrn:
This is a minor change to actually refresh the page and don't modify the url if the server is offline.


karussell:
Thanks!

PR #2430, opened 2021-10-12
OSMReader: Keep node tags until we create edges

easbar:
Instead of compressing the node and relation tags into node/relation'flags' early with this PR we keep the full node and relation tag data until we parse the ways in the second pass (but only for 'barrier' nodes and 'route' relations). This requires more memory, but it is actually not that bad, because there aren't so many barrier nodes and route relations. I did a little test and both master and this branch failed to run the import with `Xmx=1100m`, but succeeded for `Xmx=1200m`.

I keep the node tags in a `List<Map<String, Object>>` and use a separate `LongIntMap` to associate OSM node IDs with the indices into this list. The relations are grouped by the member way IDs `LongMap<List<ReaderRelation>>`. There is still room for improvement memory-wise, for example we could remove some of the tags like 'created' etc.

The goal here is to make the relation and node tag handling more flexible. With this PR we should be able to delay the creation of edge flags until we actually create the edges and get rid of `EncodingManager#handleNodeTags` and some others. Ultimately, this should help us in getting rid of `FlagEncoder` altogether.

Putting this as a Draft here, because maybe we should continue in this direction until we actually can remove e.g. `handleNodeTags` before merging.

**Update**:

We reverted the changes to relation flags for now as this increased memory usage during import in some cases.

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/ReaderElement.java
+++ b/core/src/main/java/com/graphhopper/reader/ReaderElement.java
@@ -68,11 +68,6 @@ public abstract class ReaderElement {
         return properties;
     }
 
-    public void _setTags(Map<String, Object> properties) {
-        this.properties.clear();
-        this.properties.putAll(properties);
-    }
-
     public void setTags(Map<String, String> newTags) {
         properties.clear();
         if (newTags != null)
--- a/core/src/test/java/com/graphhopper/reader/osm/OSMReaderTest.java
+++ b/core/src/test/java/com/graphhopper/reader/osm/OSMReaderTest.java
@@ -347,7 +347,7 @@ public class OSMReaderTest {
                 importOrLoad();
 
         Graph graph = hopper.getGraphHopperStorage();
-        // we ignore the barrier at node 50, but not the one at node
+        // we ignore the barrier at node 50, but not the one at node 20
         assertEquals(7, graph.getNodes());
         assertEquals(7, graph.getEdges());
 


easbar:
I am now confident that this is the way we should proceed. So far we just keep the `barrier` node tags and add extra edges for all nodes with this tag. Later we can of course extend this approach so we can keep other node tags as well. Later we could pass node tags other than `barrier` to the tag parsers without create extra edges as for `barrier`.

I'm still working on a rewrite of the `OSMReader` here: https://github.com/graphhopper/graphhopper/tree/reader2, but this PR is maybe the biggest functional (and performance-critical) change, so I kept it separate here.

This PR is a bit slower (<10%), but I think this is very reasonable considering that it also opens a lot of opportunities. The memory usage is increased as well and currently the biggest contribution is due to the relations. This is because there are a lot more route relations than barrier nodes. If we kept other node tags than `barrier` this might change.

Here is the memory footprint for the node and relation data for Germany (foot,bike+car):

```
java.util.ArrayList@13da7ab0d footprint:
     COUNT       AVG       SUM   DESCRIPTION
    441929        26  11608640   [B
         1   2160888   2160888   [Ljava.lang.Object;
    406758        34  13886704   [Ljava.util.HashMap$Node;
    441929        24  10606296   java.lang.String
         1        24        24   java.util.ArrayList
    406758        48  19524384   java.util.HashMap
    862227        32  27591264   java.util.HashMap$Node
   2559603            85378200   (total)


com.graphhopper.coll.GHLongIntBTree@52423092d footprint:
     COUNT       AVG       SUM   DESCRIPTION
      4067       416   1692336   [I
      4067       816   3319600   [J
        40       428     17144   [Lcom.graphhopper.coll.GHLongIntBTree$BTreeEntry;
         1        48        48   com.graphhopper.coll.GHLongIntBTree
      4067        40    162680   com.graphhopper.coll.GHLongIntBTree$BTreeEntry
     12242             5191808   (total)


com.carrotsearch.hppc.LongObjectScatterMap@5672a040d footprint:
     COUNT       AVG       SUM   DESCRIPTION
    339743        45  15377312   [B
         1  33554456  33554456   [J
   2801274        64 181154992   [Ljava.lang.Object;
    117057        75   8872112   [Ljava.util.HashMap$Node;
         1        16        16   com.carrotsearch.hppc.HashOrderMixing$3
         1        56        56   com.carrotsearch.hppc.LongObjectScatterMap
    117057        32   3745824   com.graphhopper.reader.ReaderRelation
    339743        24   8153832   java.lang.String
   2801273        24  67230552   java.util.ArrayList
    117057        48   5618736   java.util.HashMap
    985279        32  31528928   java.util.HashMap$Node
   7618486           355236816   (total)
```

The biggest contribution is due to the `LongObjectScatterMap<List<ReaderRelation>>` that maps each OSM way IDs for which there are route relations to the list of relations that include this way as a member (~355MB). For the (barrier only) node tags only around 85MB plus around 5MB for the mapping are needed. Thus the total is around 455MB. Assuming that the planet graph is around a factor x20 bigger than Germany this would mean around 9GB of additional memory is required during the import. There is room for improvement of course because the relations map isn't memory-friendly at all. Still, not sure if this should be a priority at the moment considering that the CH preparation requires more memory anyway, especially when we run it for multiple profiles in parallel.

PR #2432, opened 2021-10-13
MMAP load config

### LINKED ISSUE #2416: When using MMAP provide an option to force load the graph
Currently loading big graphs is possible and very fast with MMAP. The problem starts with the first requests: they are very slow.

To reduce this we could query the graph (and force the OS to load specific DataAccess objects from disk into memory) but it is unclear how to do that for arbitrary road data. Instead we can directly call `MappedByteBuffer.force()` for the important DataAccess objects like the ones of the BaseGraphStorage. We could even allow to specify a percentage for how many DataAccess objects we should call `force`. And have different values for base graph and for CH/LM (maybe even per-profile).

karussell:
Fixes #2416.

These new MMAP config options shall help to get cold query times similar to the case when the RAM option is used. They however do not help to warm up the JVM (JIT compiler). I.e. the first query will be still much slower than follow up queries but the difference should be reduced.

I tested this configuration on a rather slow Intel Xeon E5-1650V3 but with 256GB physical memory and an SSD disk.
Using a configuration with several edge-based CH custom profiles (no LM) leading to a graph of 247GB including a 39GB base graph plus 2.6GB location index and 8 CH profiles.

For java I used `-Xmx140g -Xms140s` (and ZGC) and the MMAP config option for GraphHopper.

Not really scientific measurements as I only repeated some numbers twice but as the values were very reproducible this should be fine (I ran `sudo sysctl vm.drop_caches=3` after I stopped the JVM after every measurement to avoid cache effects).

The query was a 3000km route against car with avoid motorway and only the first response time was used. (The response time of the second query was then always below 0.35s.)

config | response time (seconds) | server start (seconds)
---------|-----|-----
all options set to 0% | 39 | 15
all 0% except `graph.dataaccess.base.load: 100` | 13 | 94
all 0% except `graph.dataaccess.index.load: 100` | 38 | 21
all 0% except `graph.dataaccess.ch.load: 100` | 38 | 455
all 100% | 39 | 537
all 80% (~198g) | 38 | 430
all 50% (~124g) | 30 | 277
base=100 and ch=50 (index=0) (~142g) | 30 | 316
base=100 and ch=30 (index=0) (~101g) | 7 | 225
base=100 and ch=20 (index=0) (~80g) | 8.5 | 181
base=100 and ch=20 and index=100 (~83g)  | 8.5 | 186
RAM instead of MMAP  | | (does not load with Xmx248g to Xmx252)

The results are interesting. Especially when we try to load everything. This seems to have minimal or even no effect on the route response time although the startup time can suffer a lot. Which very likely that the required data for the route is already removed from physical memory before we can use it as the memory pressure is high when mostly everything is loaded (keep in mind the available off-heap memory is smaller than 116g=256g-140g).

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -102,7 +102,7 @@ public class GraphHopper {
     // for routing
     private final RouterConfig routerConfig = new RouterConfig();
     // for index
-    private LocationIndex locationIndex;
+    private LocationIndexTree locationIndex;
     private int preciseIndexResolution = 300;
     private int maxRegionSearch = 4;
     // for prepare
@@ -317,7 +317,7 @@ public class GraphHopper {
         return locationIndex;
     }
 
-    protected void setLocationIndex(LocationIndex locationIndex) {
+    protected void setLocationIndex(LocationIndexTree locationIndex) {
         this.locationIndex = locationIndex;
     }
 
@@ -773,6 +773,7 @@ public class GraphHopper {
 
             postProcessing(false);
             ghStorage.loadMMap(mmapLoadConfig);
+            locationIndex.loadMMap(mmapLoadConfig.getIndexPercentage());
             lmPreparationHandler.loadMMap(mmapLoadConfig.getLMPercentage());
             setFullyLoaded();
             return true;
@@ -1013,7 +1014,7 @@ public class GraphHopper {
         );
     }
 
-    protected LocationIndex createLocationIndex(Directory dir) {
+    protected LocationIndexTree createLocationIndex(Directory dir) {
         LocationIndexTree tmpIndex = new LocationIndexTree(ghStorage, dir);
         tmpIndex.setResolution(preciseIndexResolution);
         tmpIndex.setMaxRegionSearch(maxRegionSearch);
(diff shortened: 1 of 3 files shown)


easbar:
> We could query with some random requests but they might fail due to connection not found or point not found.

But what is the problem with failing requests? You could just keep sending random requests until a certain amount was successful. And usually you should know what data was imported and so it shouldn't be hard to send requests that work? Or is it crucial to use requests that actually load a large part of the graph? For JIT warmup this probably does not matter?

> Or we implement a special endpoint that has access to the graph directly.

But then we still need this new API internally (and even expose it via http)

easbar:
> Ah, really nice, then the most likely used edges would be already loaded into physical memory :) (?)

This depends on the query. Typical CH queries use only a few dozens or maybe hundreds of shortcuts (out of hundreds of millions for planet), some of which are on lower levels and some of which are on higher levels. It might be even more useful to just load some random segments (just guessing here).

easbar:
> So the "warming" with random example queries seems to be improve neighboring areas a bit but likely thousands of random requests would be required for the world wide setup to reduce the initial slowdown in a broad area like MMAP preloading does.

Ok, a thousand (CH?) queries does not sound so bad, but I can imagine that it is hard to cover a sufficient amount of the graph storage by running random queries. So let's use the extended configuration options in your new branch?

PR #2440, opened 2021-10-25
DataAccess config for type and load

### LINKED ISSUE #2416: When using MMAP provide an option to force load the graph

karussell:
Fixes #2416 and was discussed in #2432. Use in the config via:

```yml
graphhopper:

  graph.dataaccess.shortcuts_mycar: MMAP;30
  graph.dataaccess.nodes_ch_mycar: MMAP;30
  graph.dataaccess.shortcuts_car_rarely_used: MMAP;0
  graph.dataaccess.nodes_ch_car_rarely_used: MMAP;0

  # the default type:
  graph.dataaccess: RAM_STORE
  # so all other DataAccess objects like nodes, edges, geometry, location_index, string_index_keys, string_index_vals of the base graph get this default type
```


easbar:
Thanks!

Issue #2392, opened 2021-09-22
Infer default speed limits


westnordost:
This issue ticket's purpose is to
 
1. get feedback from you guys if inferring default speed limits is desired at all
2. get feedback from you guys if the default speed limit data as provided is in a format you are able to make use of etc.
3. finally, serve as a starting-area / feature request to implement this

## Assumptions

The maximum allowed speed on any given road is an important factor to rate the costing of edges in the graph and travel time.

`maxspeed` coverage in OSM is sketchy and likely will always be: Currently, **only about 12%** of all roads have a speed limit tagged.

Additionally, `maxspeed` is just for cars. Other vehicles such as trucks, buses, motorhomes, cars with trailers or motorcycles will, depending on the road type, have different maximum speed limits that are often below the signed one.

## Implicit speed limits

When speed limits are not signed, the default speed limits apply. _(Maybe more so outside the United States. I heard that in the US, things are usually always explicitly signed.)_ Regardless of the reason why `maxspeed` is missing - nothing is signed or information is still missing - the actual speed limits can be inferred by looking at the legislation. 

**For example**, in Hungary, the default speed limits are as follows
- 20 km/h on a living street
- 30 km/h on a bicycle road
- 50 km/h on any other urban road
- 90 km/h on a non-urban road, except for vehicles pulling a trailer, with a [GVWR](https://en.wikipedia.org/wiki/Vehicle_weight#Gross_vehicle_weight_rating) of 3.5t, buses or trucks for which it is 70 km/h
- 110 km/h on motorroads, except for the vehicle types mentioned above for which it is still 70 km/h
- 130 km/h on a motorway. For buses however, it's 100 km/h. For vehicles, pulling a trailer, with a GVWR of 3.5t or more and trucks, it is 80 km/h

So, if a road in Hungary is tagged with e.g. `highway=living_street` or `motorroad=yes`, we can easily infer the implicit speed limit from the above information. Hungary however is a rather easy example.

### Sourcing implicit speed limits

The legislations as the one above have been collected over the years on the OSM wiki page [Default speed limits](https://wiki.openstreetmap.org/wiki/Default_speed_limits) in a well-researched (with links to the actual legislation) and machine-readable form. @ianthetechie and me have been working on [a Python parser for this table](https://github.com/westnordost/osm-default-speeds) that outputs the information in it as a JSON-file so that it can easily be consumed by (routing) software.

[...]

PR #2450, opened 2021-10-29
European toll fallback rules

otbutz:
Added toll fallback rules for most European countries. I skipped countries like the UK or Ireland, which have toll roads but cannot be addressed by simple road type matching like e.g "toll on all motorways by default", because they only charge tolls on certain roads or even road sections.

Some of them also simply express that the country in question doesn't collect any toll like Finland or Iceland:

https://github.com/graphhopper/graphhopper/blob/2ec8e348ce2a641df7365c1c09fa83e93e28e2f6/core/src/main/java/com/graphhopper/routing/util/countryrules/europe/FinlandCountryRule.java#L26-L35

Those rules are still useful as we know that any untagged way can safely be assumed to be toll free in those countries.


otbutz:
@karussell anything missing?

easbar:
> The main problem I have with continuing this path for CountryRules is that I'm unsure if we should add a new method for every encoded value we want to modify - see also e.g. https://github.com/graphhopper/graphhopper/pull/2402.

But what would be the alternative? We could keep the encoded values separate from each other by putting all the country-specific stuff into the parsers directly, i.e. the TollParsers knows everything about Toll and also about Toll in the different countries. But then the country-specific stuff is spread over several parsers. Or we use separate country rules for each country **and** for each encoded value, like `GermanyTollCountryRule, GermanyMaxSpeedCountryRule, ...`. But then we have even more files and I do not really see why this would be better. Changing the country rules will require recompilation anyway (by design) so I do not really see a problem with having the different methods in one file. After all, I think this is exactly what we wanted initially: One file per country.

> Adding more methods is not really problematic because they all come with a default implementation. Child classes only need to implement them when it makes sense to do so.

Yes, this saves some work when someone wants to add such a method (for another or custom encoded value).

> Making the interface more generic just moves the ugliness into the rules themselves. I'd rather have a larger but well-defined interface than having to write boilerplate code in every rule.

You mean we could use some kind of generic method in the `CountryRule` interface and then do some runtime logic in the implementations of this method that checks which encoded value shall be modified? Something like:

```java
interface CountryRule {
   Object getXYZ(ReaderWay way, ...);
} 

class MyCountryRule implements CountryRule {
    Object getXYZ(ReaderWay way, ...) {
       if (thisIsAboutToll(way, ...)) {
            Toll toll = determineToll(way, ...);
            return toll;
       } else if (thisIsAboutMaxSpeed(way, ...)) {
          double maxSpeed = determineMaxSpeed(way, ...);
            return maxSpeed;
       }
    }
}
```

Yes, I wouldn't like this either.

easbar:
Actually I am more worried about the signature of the CountryRule methods. What is the TransportationMode supposed to do here? These encoded values aren't meant to be 'vehicle' specific, are they? And what is `currentXYZ` supposed to mean? Can we agree that this shall be the value that the parser would return if the country rule did nothing at all? Maybe we can find a name that expresses this more clearly? Also `getAccess` should probably be called `getRoadAccess`. Each method is meant to alter the value of exactly one encoded value in a country-specific way, right?

easbar:
Maybe 

```java
    default double modifyMaxSpeed(ReaderWay readerWay, double maxSpeed) {
        return maxSpeed;
    }

    default RoadAccess modifyRoadAccess(ReaderWay readerWay, RoadAccess roadAccess) {
        return roadAccess;
    }
    
    default Toll modifyToll(ReaderWay readerWay, Toll toll) {
        return toll;
    }
}
```

I know, not directly what this PR is about, but now we are at it...

otbutz:
> These encoded values aren't meant to be 'vehicle' specific, are they?

MaxSpeed and RoadAccess are both dependant on the TransportationMode. Or better, they *should* be.

e.g maxspeed for [motorcar](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxspeed#Motorcar) vs [hgv](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxspeed#HGV)

> And what is `currentXYZ` supposed to mean? Can we agree that this shall be the value that the parser would return if the country rule did nothing at all?

CountryRules could also cap values. So they're not restricted to providing fallback values.

> Each method is meant to alter the value of exactly one encoded value in a country-specific way, right?

Correct.

easbar:
> MaxSpeed and RoadAccess are both dependant on the TransportationMode. Or better, they should be.

Right. But so far they are not :) And certainly not all encoded values depend on the transportation mode. Then again there is no reason all the CountryRule methods need to have the same signature. We should just not add the TransportationMode parameter to all methods when it is not needed, just because some others have it.

> CountryRules could also cap values. So they're not restricted to providing fallback values.

Sure, this not what I meant. I meant that the parsers determine some kind of value (what we currently call `currentMaxSpeed` etc.) and the country rules provide a hook that allows modifying the value (cap, etc.) one last time before it is returned.

otbutz:
> > e.g maxspeed for [motorcar](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxspeed#Motorcar) vs [hgv](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxspeed#HGV)
> 
> Wouldn't this be two different EncodedValues that we want to fill? (max_speed & max_speed_hgv)

Not sure if it's worth to keep one EV per TransportationMode but at least HGV should be treated differently.

easbar (approved):

core/src/main/java/com/graphhopper/routing/util/countryrules/europe/AlbaniaCountryRule.java
```diff
@@ -0,0 +1,36 @@
...
+import com.graphhopper.routing.util.TransportationMode;
+import com.graphhopper.routing.util.countryrules.CountryRule;
+
+public class AlbaniaCountryRule implements CountryRule {
+    
+    @Override
+    public Toll getToll(ReaderWay readerWay, TransportationMode transportationMode, Toll currentToll) {
+        if (!transportationMode.isMotorVehicle() || currentToll != Toll.MISSING) {
+            return currentToll;
+        }
```
easbar, line 32: Does the Toll EV depend on the TransportationMode or not? Can we not just interpret the Toll EV as representing the toll for motorized vehicles (and assume there is no such things for non-motorized vehicles)? And if yes, can we remove the transportation mode parameter from this method already? I will adjust the signatures further as we discussed, but since the `getToll` method is added here so many times can we remove the transportation mode parameter here already?

core/src/main/java/com/graphhopper/routing/util/countryrules/europe/BelarusCountryRule.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class BelarusCountryRule implements CountryRule {
+    
+    @Override
+    public Toll getToll(ReaderWay readerWay, TransportationMode transportationMode, Toll currentToll) {
+        if (!transportationMode.isMotorVehicle() || currentToll != Toll.MISSING) {
+            return currentToll;
+        }
+
+        RoadClass roadClass = RoadClass.find(readerWay.getTag("highway", ""));
```
easbar, line 35: Sooner or later I think the country rules should have available the encoded values that were written already at this point so we could just get the road class as encoded value rather than repeating the road class parsing here. We would probably just add the encoding manager to the country rules via their constructor. Whether or not this will work will depend on the road class parser being executed before the toll parser, but this is exactly what we were planning to do (see #2470): The order of the parsers matters and allows encoded values to depend on others.

Again, since we are adding so many countries here, can we make this change already?

easbar:
Btw did you take the toll information for the different countries from a wiki table? Is there such a table? Where? Or did you take this data from somewhere else? I still think we should write a (semi-automatic) code generator to parse these wiki tables (same for max speed and access) at some point (#https://github.com/graphhopper/graphhopper/issues/2392), but that is for another issue.

core/src/main/java/com/graphhopper/routing/util/countryrules/europe/AlbaniaCountryRule.java
otbutz: Agreed

otbutz:
> Btw did you take the toll information for the different countries from a wiki table? Is there such a table? Where?

Sadly not. I had to compile the rules from various sources. But i think it might be worth to create a wikipage for it and maintain it there.

core/src/main/java/com/graphhopper/routing/util/countryrules/europe/BelarusCountryRule.java
otbutz: So we would need to add dependency resolution?

```java
public interface TagParser {
    
    Set<String> getProvidedEncodedValues();
    
    default Set<String> getEncodedValueDependencies() {
        return Collections.emptySet();
    }
```

We could do a topological sort within `EncodingManager.Builder.build()`. With the parsers in the right order, we can safely pass the EncodedValueLookup into the CountryRules.

easbar: Hm, not sure the original idea was just doing this manually, i.e. the order of the list of tag parsers counts. But actually if all you need to do here is reading the highway tag, maybe let's postpone this.

core/src/main/java/com/graphhopper/routing/util/countryrules/CountryRuleFactory.java
```diff
@@ -25,14 +25,64 @@
...
+        rules.put(ISL, new IcelandCountryRule());
+        rules.put(ITA, new ItalyCountryRule());
+        rules.put(JEY, new JerseyCountryRule());
+        rules.put(LIE, new LiechtensteinCountryRule());
+        rules.put(LTU, new LithuaniaCountryRule());
+        rules.put(LUX, new LuxembourgCountryRule());
+        rules.put(LVA, new LatviaCountryRule());
+        rules.put(MCO, new MonacoCountryRule());
+        // MDA
+        // MKD
```
easbar, line 70: I guess these are just missing countries for which we have no toll rules yet? Can you just add an empty country rule for these? Or doesn't this make sense?

otbutz: I can add empty CountryRules for them. Probably easier to fill the gaps for potential contributors.

otbutz:
Germany was at least a prominent example in the past, but maybe things have improved? We (OPTITOOL GmbH) tried running GH without the fallback rules about two years ago and quickly reintroduced the rules when customers reported high deviations.

otbutz:
> Ok, so this is more about HGV toll than about missing `toll=yes` in general?

Absolutely

otbutz:
> Did you add these countries just for completeness?

Yes.

> And can we set Toll.NO there instead?

If applicable, sure.

Issue #1898, opened 2020-02-11
Introduce edge-length dependent EncodedValues


karussell:
When we store the absolute elevation change in an encoded value then this would not properly work for the virtual edges as it needs to be changed proportional to the length of the edge. Similar to what we do for the [distance now](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/querygraph/GraphModificationBuilder.java#L229).

With this change we could then introduce elevation and an absolute "time" or "congestion" term for CustomWeighting (#1776).

With this change we could also make the distance an EncodedValue.

PR #2451, opened 2021-10-29
Add geometry encoded value

easbar:
I was trying to make some progress towards #1898. See comments inline.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ev/ElevationGain.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/ElevationGain.java
@@ -1,0 +1,55 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.graphhopper.routing.ev;
+
+import com.graphhopper.storage.IntsRef;
+import com.graphhopper.util.PointList;
+
+public class ElevationGain {
+    public static final String KEY = "elevation_gain";
+
+    public static GeometryEncodedValue create() {
+        return new ElevationGainEncodedValue();
+    }
+
+    private static class ElevationGainEncodedValue extends UnsignedDecimalEncodedValue implements GeometryEncodedValue {
+        public ElevationGainEncodedValue() {
+            // todonow: how many bits etc.
+            super(KEY, 32, 1, true);
+        }
+
+        @Override
+        public double calculateDecimal(boolean reverse, IntsRef ref, PointList geometry) {
+            // todonow: handle !3D
+            // todonow: we probably need to handle several edge cases, like very short edges etc. see also Bike2WeightFlagEncoder
+            double gain = 0;
+            double prev = geometry.getEle(0);
+            for (int i = 1; i < geometry.size(); i++) {
+                double ele = geometry.getEle(i);
+                double delta = ele - prev;
+                if (!reverse && delta > 0)
+                    gain += delta;
+                if (reverse && delta < 0)
+                    gain -= delta;
+                prev = ele;
+            }
+            return gain;
+        }
+    }
+}
--- a/core/src/main/java/com/graphhopper/routing/ev/GeometryEncodedValue.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/GeometryEncodedValue.java
@@ -1,0 +1,26 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.graphhopper.routing.ev;
+
+import com.graphhopper.storage.IntsRef;
+import com.graphhopper.util.PointList;
+
+public interface GeometryEncodedValue extends DecimalEncodedValue {
+    double calculateDecimal(boolean reverse, IntsRef ref, PointList geometry);
+}
--- a/core/src/main/java/com/graphhopper/routing/ev/Distance.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/Distance.java
@@ -1,0 +1,46 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.graphhopper.routing.ev;
+
+import com.graphhopper.storage.IntsRef;
+import com.graphhopper.util.DistanceCalc;
+import com.graphhopper.util.DistanceCalcEarth;
+import com.graphhopper.util.PointList;
+
+public class Distance {
+    public static final String KEY = "distance";
+
+    public static GeometryEncodedValue create() {
+        return new DistanceGeometryEncodedValue();
+    }
+
+    private static class DistanceGeometryEncodedValue extends UnsignedDecimalEncodedValue implements GeometryEncodedValue {
+        private static final DistanceCalc distCalc = DistanceCalcEarth.DIST_EARTH;
+
+        public DistanceGeometryEncodedValue() {
+            // todonow: bits, factor etc.
+            super(KEY, 32, 1, false);
+        }
+
+        @Override
+        public double calculateDecimal(boolean reverse, IntsRef ref, PointList geometry) {
+            return distCalc.calcDistance(geometry);
+        }
+    }
+}
--- a/core/src/main/java/com/graphhopper/routing/ev/UnsignedDecimalEncodedValue.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/UnsignedDecimalEncodedValue.java
@@ -23,7 +23,7 @@ import com.graphhopper.storage.IntsRef;
  * This class holds a decimal value and stores it as an unsigned integer value via a conversion factor and a maximum number
  * of bits.
  */
-public final class UnsignedDecimalEncodedValue extends UnsignedIntEncodedValue implements DecimalEncodedValue {
+public class UnsignedDecimalEncodedValue extends UnsignedIntEncodedValue implements DecimalEncodedValue {
     private final double factor;
     private final double defaultValue;
     private final boolean useMaximumAsInfinity;
--- a/core/src/main/java/com/graphhopper/routing/querygraph/VirtualEdgeIterator.java
+++ b/core/src/main/java/com/graphhopper/routing/querygraph/VirtualEdgeIterator.java
@@ -17,11 +17,7 @@
  */
 package com.graphhopper.routing.querygraph;
 
-import com.graphhopper.routing.ev.BooleanEncodedValue;
-import com.graphhopper.routing.ev.DecimalEncodedValue;
-import com.graphhopper.routing.ev.EnumEncodedValue;
-import com.graphhopper.routing.ev.IntEncodedValue;
-import com.graphhopper.routing.ev.StringEncodedValue;
+import com.graphhopper.routing.ev.*;
 import com.graphhopper.routing.util.EdgeFilter;
 import com.graphhopper.storage.IntsRef;
 import com.graphhopper.util.EdgeIterator;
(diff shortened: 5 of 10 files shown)


core/src/main/java/com/graphhopper/routing/ev/ElevationGain.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.graphhopper.storage.IntsRef;
+import com.graphhopper.util.PointList;
+
+public class ElevationGain {
+    public static final String KEY = "elevation_gain";
+
+    public static GeometryEncodedValue create() {
+        return new ElevationGainEncodedValue();
+    }
```
easbar, line 29: Just a quick sketch of an encoded value that holds the total elevation gain (or loss if evaluated in reverse) of an edge.

easbar:
I'm not sure if this is actually needed yet and should we really try to replace `edge.getDistance()` with `edge.get(distanceEnc)`? The obvious alternative would be adding `edge.getElevationGain(reverse)` instead... How many of such geometry encoded values can we come up with and are there any other 'length-dependent' EVs? Maybe something like maximum slope?

Also average slope is interesting, because this is just `elevation_gain(forward)-elevation_gain(backward)/distance`, so basically a geometry encoded value that does not store anything but is calculated based on other geometry encoded values...

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -653,7 +655,10 @@ void addEdge(int fromIndex, int toIndex, PointList pointList, IntsRef flags, Rea
         if (doSimplify && pointList.size() > 2)
             simplifyAlgo.simplify(pointList);
 
-        double towerNodeDistance = distCalc.calcDistance(pointList);
+        // todonow: how should we manage geometry encoded values / where do they come from, probably EncodingManager somehow?
+        //          here they are very similar to tag parsers. (handleWayTags -> handleGeometry)
+        GeometryEncodedValue distanceEnc = Distance.create();
+        double towerNodeDistance = distanceEnc.calculateDecimal(false, flags, pointList);
```
easbar, line 661: If we ever replaced distance with the distance EV this would be the place to calculate the distance (and same for elevation gain etc.) in OSMReader.

core/src/main/java/com/graphhopper/routing/ev/GeometryEncodedValue.java
```diff
@@ -0,0 +1,26 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.graphhopper.routing.ev;
+
+import com.graphhopper.storage.IntsRef;
+import com.graphhopper.util.PointList;
+
+public interface GeometryEncodedValue extends DecimalEncodedValue {
```
easbar, line 24: The name might be confusing. It is not meant as an EncodedValue that stores the 'geometry/pillar nodes/point list', but rather a road attribute that depends on this geometry.

karussell:
Thank you! Interesting to think about this topic having this code sketch :)

> I'm not sure if this is actually needed yet and if we should actually try to replace edge.getDistance() with edge.get(distanceEnc)? The obvious alternative would be adding edge.getElevationGain(reverse) instead... How many of such geometry encoded values can we come up with and are there any other 'length-dependent' EVs? Maybe something like maximum slope?

I'm not sure :) and not sure if it is worth the effort in general :)

I would view this "geo-dependent calculation" also more as an "external procedure" where we do not need to adapt **getting the value**. And only adapt **setting the value** when we create the edge 1. in OSMImport and 2. for VirtualEdgeIteratorState. The advantage would be that there is no performance implication at query time and no need to add a new API for this to custom models etc.

easbar:
>  The advantage would be that there is no performance implication at query time

But at query time we have to calculate distance, elevation gain etc. for the virtual edges. There is no way around this? And where else would you put the code that does this calculation. It is the same code we use for the real edges during import.

> and no need to add a new API for this to custom models etc.

Can you explain what you mean here? What needs to be done for custom models?

And what was the original idea of 'length-dependent' encoded values? Did you have something different in mind?

karussell:
> But at query time we have to calculate distance, elevation gain etc. for the virtual edges. There is no way around this? 

Yes. But this calculation has to be done only once and then we store the value in the virtual edge. Then, for fetching this value, it does not matter if it is a relative or absolute decimal value.

> Can you explain what you mean here? What needs to be done for custom models?

If we would introduce a new edge.get() or edge.getElevationGain() we would need to adapt the current code in custom models.

> And what was the original idea of 'length-dependent' encoded values? Did you have something different in mind?

The idea for this was not really concrete.

But I also do not think that the code is far away from this idea. I think that it would be nice to avoid a new edge.get method and instead have e.g. code in VirtualEdgeIteratorState (and in OSMReader/TagParser) that sets the final value:

```
for(encVal : encodingManager.getGeoDependentEncodedValues()) {
    encVal.setDecimal(false, intsRef, pointsList)
}
new VirtualEdgeIteratorState(..., intsRef, ...)
```

And then, for reading, the "geo-dependent" EncodedValues are identical like the other DecimalEncodedValues

easbar:
> Yes. But this calculation has to be done only once and then we store the value in the virtual edge.

Ok. I thought the getter would not be called many times (or even only once) for virtual edges. 

> Then, for fetching this value, it does not matter if it is a relative or absolute decimal value.
> If we would introduce a new edge.get() or edge.getElevationGain() we would need to adapt the current code in custom models.

Ok, I see. I thought the problem with this would be that the virtual edges would have to be aware of which geometry values exist. Currently `VirtualEdgeIteratorState` has a `distance` field and then it would also need an `elevationGain` field etc. and they would have to be (statically) associated with the according `GeometryEncodedValues` somehow. But maybe we could just use a dynamic `Map<String, Double>` that maps encoded values names to double values instead?

> and instead have e.g. code in VirtualEdgeIteratorState (and in OSMReader/TagParser) that sets the final value:

I already do this for import: 

```java
    // todonow: how should we manage geometry encoded values / where do they come from, probably EncodingManager somehow?
        //          here they are very similar to tag parsers. (handleWayTags -> handleGeometry)
        GeometryEncodedValue distanceEnc = Distance.create();
        double towerNodeDistance = distanceEnc.calculateDecimal(false, flags, pointList);
        // ...
        EdgeIteratorState iter = graph.edge(fromIndex, toIndex)
                // this is just the standard setter for `DecimalEncodedValue` that is currently inherited by `GeometryEncodedValue`
                .set(distanceEnc, towerNodeDistance)
   
```

For `VirtualEdgeIteratorState` we could do the same, but like I said then we need this `Map<String, Double>` or similar.

> And then, for reading, the "geo-dependent" EncodedValues are identical like the other DecimalEncodedValues

Ok, yes. I can see the benefits of this.

Or... we do not do any of this and just add `EdgeIteratorState.get/setElevationGain`... and keep `EdgeIteratorState.get/setDistance` as well.

karussell:
Yes, indeed. We should try to avoid the knowledge of the geometry values and could indeed do it in the getter for virtual edges (as you say: it shouldn't matter performance-wise). But I would still try to avoid a different getter method. We could do the mechanics under the hood like (?)

```java
double getDecimal(EncodedValue enc) {
    if(enc instance GeoEncValue) return ((GeoEncValue)enc).calculateValue(reverse, edgeFlags, pointList)
    enc.getDecimal(reverse, edgeFlags);
}
```

> I already do this for import

Sorry, you already did this how meant it (I think), but I did not fully understand the "todonow" (i.e. wasn't sure which of this code is a sketch)

> But maybe we could just use a dynamic Map<String, Double> that maps encoded values names to double values instead?

Yes, this would be a possibility.

> Or... we do not do any of this and just add EdgeIteratorState.get/setElevationGain... 

Another option would be to use "elevation gain per meter" :) but of course the precision would be lost.

easbar:
I removed `get(GeometryEncodedValue)` and do the instanceof check that you proposed in `VirtualEdgeIteratorState` now. This should work the same way as my first version, but we don't need to add another get method. We could still somehow store or cache the calculated values in `VirtualEdgeIteratorState` to replace the `calculateDecimal` call.

> Sorry, you already did this how meant it (I think), but I did not fully understand the "todonow" (i.e. wasn't sure which of this code is a sketch)

Once we know how to get/set values for geometry EVs the next question is where we keep the geometry EVs. The code where I put the todonow simply creates the distance EV but that won't cut it, because we also need the EV to read the values later.

> Another option would be to use "elevation gain per meter" :) but of course the precision would be lost.

What kind of elevation EVs we actually use or provide is another question I guess. The good news is that once we established the framework for this we can use any implementation that calculates some double values (but only unsigned so far!) from the point list. The idea of the `ElevationGain` I added here was that it is not just the average slope of the entire edge but rather the sum of all positive elevation changes so an edge that starts/ends on the same elevation, but represents a road that still goes up and down between the start and end can be distinguished from a road that is just flat. Note that we can get the `ElevationLoss` by simply reading the `ElevationGain` in reverse. See also my comment about the average slope here: https://github.com/graphhopper/graphhopper/pull/2451#issuecomment-954983104

karussell:
Nice!

> The good news is that once we established the framework for this we can use any implementation that calculates some double values from the point list.

Yes. My point was more that there is a possibility to consider elevation in an encoded value (and in custom models) and still avoiding establishing a new framework but this surely comes with your mentioned drawbacks.

Still at the moment we have to focus on improving the elevation data or precision first IMO as otherwise we won't get the intended benefits. See e.g. [this route](https://graphhopper.com/maps/?point=50.904684%2C14.207039&point=50.889824%2C14.232402&profile=bike) close a mountain which is a smooth slightly downward route in reality.

In Bike2FlagEncoder I explicitly used the "average slope" (elevation difference of tower nodes only) as the more detailed "elevation gain" produced worse results.

---

What just came to my mind is that with this new framework one could block a very specific location at an edge without introducing a new barrier edge. I.e. assume tower nodes A and B with pillar nodes in-between `A-x-y-z-B` and the elevation is `0 0 BIG 0 0`. And now if a destination is at x then the routes goes via A. But if a destination is at z then the route goes via B. Or to put it differently: elevation is just one of the many possible per-pillar-node data we could store and make available like construction data, pedestrian crossing, traffic light or even speed.

easbar:
> My point was more that there is a possibility to consider elevation in an encoded value (and in custom models) and still avoiding establishing a new framework but this surely comes with your mentioned drawbacks.

But how? You mean using `edgeState.getSlope()` or something?  I don't really see how this would be easier even.

> Still, at the moment, we have to focus on improving the elevation data or precision first IMO as otherwise we won't get the intended benefits. 

Ok. But maybe  `graph.elevation.long_edge_sampling_distance` and `graph.elevation.way_point_max_distance` already help here? Hm or maybe not your example looks like we rather need some kind of smoothing. But anyway, I'm not sure if such artifacts are a good reason not to allow such geometry-dependent EVs already, because once we have this at least we can start experimenting with elevation EVs. But sure there is also no need to rush with this issue.

> What just came to my mind is that with this new framework one could block a very specific location at an edge without introducing a new barrier edge.

Not sure if I understand this yet. Do you mean blocking a specific location during import or at query time?

karussell:
> But how? You mean using edgeState.getSlope() or something? I don't really see how this would be easier even.

Using a DecimalEncodedValue like `edgeState.get(averageSlope)`. We would have to feed this value at import time from what you wrote above:

```
average_slope := (elevation_gain(forward) - elevation_gain(backward)) / distance
```

As this slope is "smeared" over the entire edge (it is "per meter") we can split the edge anywhere, i.e. there is no special handling required for this average slope to create virtual edges. (We could even have gain and loss as separate EV)

And this average_slope is different to what we currently do in Bike2FlagEncoder:

```
rough_average_slope := (elevation at A - elevation at B) / distance
```

(and in my memories I only compared rough_average_slope with average_slope and I picked the rough values but this might have changed.)

> But anyway, I'm not sure if such artifacts are a good reason not to allow such geometry-dependent EVs already,

Currently I think that the "average_slope" will even give better results (as it hides these artifacts). So I would play with this first in custom models and make it easy there (something like `"multiply_by": "slope"` to avoid hills?). And I would see this here as a later step. btw: how would we use it in custom models?

> Not sure if I understand this yet. Do you mean blocking a specific location during import or at query time?

Not really related to this issue. But yes, I meant at import time without creating a separate edge. And this showed me a generic version of store "something per pillar node".

easbar:
> As this slope is "smeared" over the entire edge (it is "per meter") we can split the edge anywhere, i.e. there is no special handling required for this average slope to create virtual edges.

Ok, I see, yes this is true and because of this we would not need any special handling for split edges.

> Currently I think that the "average_slope" will even give better results (as it hides these artifacts)

I think this entirely depends on the data and the use-case. For good mountainbike profiles (at least some I can imagine), or wheelchair profiles that need to prevent steep sections I don't think the average slope would be sufficient, but for car even not considering the elevation can give useful results. This is kind of the idea of this draft: It does not matter and everybody can just implement whatever encoded values they like.

> btw: how would we use it in custom models?

So far I did not (only) intend to use this for custom models, because one could still implement some weighting in Java that makes use of this. But let's see: Just off the top of my head a few ideas. 

I would find the maximum slope interesting (can be used to prevent or prefer steep sections):

```json
"priority": [
   {
       "if": "max_slope > 0.1"
       "multiply_by": 0.1
   }    
]
```

or maybe the average upward/downward slope to find/exclude steep roads

```json
"priority": [
   {
       "if": "average_upward_slope > 0.05"
       "multiply_by": 0.1
   }    
]
```

or just count how many meters of an edge are above a certain slope threshold:

```json
"priority": [
   {
       "if": "steep_sections > 30"
       "multiply_by": 0.1
   }    
]
```

The simple `average_slope` could be used in a similar way (`average_slope > 0.05`), but would be less useful.

> something like "multiply_by": "slope" to avoid hills

Hm, but this requires "slope" as a right-hand value and it would be hard to tell how this influences the weight, because the absolute value of "slope" might be unknown or hard to compare to the factors used in other rules.

> And this average_slope is different to what we currently do in Bike2FlagEncoder:

Not really related to this issue, but I think the two formulae are equivalent. The difference difference between the elevation gain and the elevation loss is equal to the difference between the elevation at A and B?!

karussell:
> but this requires "slope" as a right-hand value and it would be hard to tell how this influences the weight

Yes, but IMO we need this. Otherwise introducing and tweaking "avoid or prefer hills" might be too complicated. We could allow a factor in-between `"multiply_by": 0.5 * slope` or even a math function `"multiply_by": sqrt(slope)` and have an exact definition. We need this to not only avoid steep sections, but pick one of many possible routes with different elevation changes e.g. assume you have a route with mild incline and then one with sharper incline. (The alternative would be to use some kind of a "step function" with many if's)

> because the absolute value of "slope" might be unknown or hard to compare to the factors used in other rules

Why? Couldn't we define the slope like it is usually defined in meters of elevation gain or loss per 100m?

> Not really related to this issue, but I think the two formulae are equivalent. The difference difference between the elevation gain and the elevation loss is equal to the difference between the elevation at A and B?!

Ah, yes, indeed. Probably I tried something else with involving pillar nodes too basing the speed changes on the total elevation gain + loss.

> For good mountainbike profiles (at least some I can imagine), or wheelchair profiles that need to prevent steep sections I don't think the average slope would be sufficient

Probably. But before we come to this challenge we need to tune the elevation data.

karussell:
> I don't see how this draft is any more complicated than introducing right-hand encoded values.

Probably my comments were a bit too intense ... I'm not against this draft. I do not want to hinder something you would like to push :)

> Using "if": "max_slope > 0.1" would be a very simple approach to avoiding hills and not require any changes to the custom model language

With "avoiding hills" I also mean to prefer less hilly roads. With this simple "max_slope>0.1" condition you won't optimize the elevation changes on a route and only avoid or exclude too big elevation changes.

> I think quite the opposite: We need to be able to include the elevation in the modeling to find out in which way we need to tune the elevation data.

Yes. I meant that we do not necessarily need this PR and could feed an ordinary average_slope EV.

easbar:
>  I do not want to hinder something you would like to push :)

Sure, I did not feel like you wanted to hinder something :) It's certainly good to make sure something is actually needed before we introduce it.

> Yes. I meant that we do not necessarily need this PR and could feed an ordinary average_slope EV

Yes, maybe this is the best thing to start with. I was just about to write an issue just for average_slope. I'm sure we will learn something along the way and everything we need for this we probably need for more advanced elevation use-cases as well.

PR #2455, opened 2021-11-02
Load CH preparations in parallel

### LINKED ISSUE #2454: Load CH preparations in parallel
We currently load LM preparations with `prepare.lm.threads` threads. For CH we load the preparations one by one (not in parallel). I did a quick check and loading in parallel actually seems faster (up to x2, but also large fluctuations, using four threads when I tried it on my laptop with four files of size 3GB). 

Since (at least edge-based) CH preparations are comparable to LM preparations shall we load CH in parallel as well? The reason I ran into this was, again, that I was trying to disentangle the loading from the import code and in `LMPreparationHandler` the same parallel loop does the loading and import. For CH it is different, because we load the CH preparations with the base graph before `CHPreparationHandler` is doing anything (for no actual reason).

Here is my testing code:

```java
package com.graphhopper.tools;

import com.graphhopper.storage.DAType;
import com.graphhopper.storage.DataAccess;
import com.graphhopper.storage.GHDirectory;
import com.graphhopper.util.StopWatch;

import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.ExecutionException;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.Future;

public class ParallelLoad {
    public static void main(String[] args) throws ExecutionException, InterruptedException {
        {
            StopWatch swWrite = StopWatch.started();
            GHDirectory dir = new GHDirectory("parallel-load-test-gh", DAType.RAM_STORE);
            dir.create();
            DataAccess[] das = new DataAccess[4];
            for (int i = 0; i < 4; i++) {
                das[i] = dir.create("file" + i);
                das[i].create(3_000_000_000L);
                das[i].setInt(13, 54);
                das[i].flush();
                das[i].close();
            }
            System.out.println("Writing took: " + swWrite.stop().getTimeString());
        }
        {
            StopWatch swLoad = StopWatch.started();
            GHDirectory dir = new GHDirectory("bla", DAType.RAM_STORE);
            DataAccess[] das = new DataAccess[4];
            List<Future<String>> futures = new ArrayList<>(4);
            ExecutorService executorService = Executors.newFixedThreadPool(4);
            for (int i = 0; i < 4; i++) {
                final int index = i;
                futures.add(executorService.submit(() -> {
                    String name = "file" + index;
                    das[index] = dir.create(name);
                    das[index].loadExisting();
                    return name;
                }));
            }
            executorService.shutdown();
            for (int i = 0; i < 4; i++)
                futures.get(i).get();
            // just a quick check
            for (int i = 0; i < 4; i++)
                if (das[i].getInt(13) != 54)
                    throw new IllegalStateException();
            System.out.println("Loading took: " + swLoad.stop().getTimeString());
        }
    }
}
```

Using four threads:

```
Writing took: 15.74s
Loading took: 6.35s
```

Without ExecutorService (just one thread)
```
Writing took: 15.53s
Loading took: 10.90s
```

easbar:
Fixes #2454


core/src/main/java/com/graphhopper/routing/ch/CHPreparationHandler.java
```diff
@@ -116,37 +115,26 @@ public int getPreparationThreads() {
...
-
-        threadPool.shutdown();
-
-        try {
-            for (int i = 0; i < preparations.size(); i++) {
-                completionService.take().get();
-            }
-        } catch (Exception e) {
-            threadPool.shutdownNow();
-            throw new RuntimeException(e);
```
easbar, line 148: I moved this code to a helper method in `GHUtility`.

core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
```diff
@@ -270,10 +268,14 @@ public boolean loadExisting() {
...
-                if (!cg.chStore.loadExisting())
-                    throw new IllegalStateException("Cannot load " + cg);
-            });
+            List<Callable<String>> callables = chEntries.stream().map(c -> (Callable<String>) () -> {
+                        if (!c.chStore.loadExisting())
+                            throw new IllegalStateException("Cannot load " + c);
+                        return c.chConfig.getName();
+                    })
+                    .collect(Collectors.toList());
+            int numThreads = Math.max(1, Math.min(4, callables.size()));
```
easbar, line 277: We used to only use one thread here. The `prepare.ch.threads` parameter is not available here, so I hard-coded the default to 4 (max). Should be ok, because the memory usage won't be increased because of this.

core/src/main/java/com/graphhopper/util/GHUtility.java
```diff
@@ -723,6 +727,20 @@ public static long calcMillisWithTurnMillis(Weighting weighting, EdgeIteratorSta
...
+        ExecutorCompletionService<String> completionService = new ExecutorCompletionService<>(executorService);
+        callables.forEach(completionService::submit);
+        executorService.shutdown();
+        try {
+            for (int i = 0; i < callables.size(); i++)
+                completionService.take().get();
+        } catch (Exception e) {
+            executorService.shutdownNow();
+            throw new RuntimeException(e);
+        }
```
otbutz, line 741:
```suggestion
        try {
            for (int i = 0; i < callables.size(); i++)
                completionService.take().get();
        } catch (InterruptedException e) {
            Thread.currentThread().interrupt();
            throw new RuntimeException(e);
        } catch (ExecutionException e) {
            throw new RuntimeException(e);
        } finally {
            executorService.shutdown();
        }
```

core/src/main/java/com/graphhopper/util/GHUtility.java
```diff
@@ -723,6 +727,20 @@ public static long calcMillisWithTurnMillis(Weighting weighting, EdgeIteratorSta
         }
     }
 
+    public static void runConcurrently(List<Callable<String>> callables, int threads) {
```
otbutz, line 730:
```suggestion
    public static void runConcurrently(List<Callable<String>> callables, int threads, String threadPrefix) {
```

Would be nice if we could remove the Thread renaming from the Callables and let the ThreadFactory of the Executor handle this.

otbutz (changes requested):

core/src/main/java/com/graphhopper/util/GHUtility.java
easbar: I tried to simply keep it as it was before. There is a long, possibly related discussion about this stuff here: #980

core/src/main/java/com/graphhopper/storage/GraphHopperStorage.java
otbutz: If memory consumption doesn't scale with the threadcount here, why don't we use all available cores?

```java
int numThreads = Math.max(1, Math.min(Runtime.availableProcessors(), callables.size()));
```

easbar: Yes, this would also be possible, I guess. I was just trying to be conservative. If we did that we could also just use `.parallelStream()`.

otbutz: Using `parallelStream()` has some sideeffects as it's ForkJoinPool is shared by the whole JVM. Probably not that bad for standalone GraphHopper but a pitfall people which integrate it as a library.

easbar: Ok, good to know. :+1:

karussell:
LGTM - thanks!

easbar:
I'll merge this so we can move forward. @otbutz if you still think something should be improved please don't hesitate to open a separate PR.

PR #2458, opened 2021-11-05
Ignore too large mtb:scale values

easbar:
Sometimes mappers seem to use `mtb:scale=12` or `mtb:scale=34`. They probably mean "S1/2", or "S3/4", but this is not recommended in the OSM wiki: https://wiki.openstreetmap.org/wiki/Key:mtb:scale?uselang=en. The wiki encourages mappers to use `1+` or `2-` instead of `S1/2` for example. There are less than 20 occurrences of `12`, `23` and `34` for the whole planet according to taginfo, so I think we can just ignore this. We already ignore the finer separation using `+/-` which is recommended in the wiki (and somewhat popular) btw.

This PR ignores values above 6. In current master we get an exception during import because the EV's maximum value is exceeded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### 5.0 [not yet released]
 
+- fixed handling of too large mtb:scale tags
 - use GraphHopper#setGraphHopperLocation before calling load() instead of GraphHopper#load(graphHopperLocation) (#2437)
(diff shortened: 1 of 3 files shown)


karussell (approved):
:+1:

easbar:
A bit surprising nobody noticed this yet...

PR #2464, opened 2021-11-17
Improve exceptions

### LINKED ISSUE #1188: Create a GHException for "maximum nodes exceeded"

otbutz:
* adds a dedicated exception if the maximum number of nodes is reached: `MaximumNodesExceededException`
* added support for it in the `ResponsePathDeserializer` to handle it properly
* `PointNotFoundException`/`PointOutOfBoundsException` are now a subclass of `DetailedIllegalArgumentException`
* added missing `serialVersionUID` to all exceptions

grafik

Regarding compatibility:

Users of the GH library currently need to process the messages of IllegalArgumentExceptions if they want to check if the number of nodes has been reached while processing a request:

```java
if (e instanceof IllegalArgumentException && e.getMessage() != null && e.getMessage().startsWith("No path found due to maximum nodes exceeded"))
```

This is cumbersome and might break silently if the message is changed in future GH versions.

The new excpetion is still a subclass of `IllegalArgumentException` and the message is untouched, so this shouldn't break existing code.

But with this change it's a lot easier to deal with routing related errors:

```java
if (t instanceof ConnectionNotFoundException) {
    // ...
} else if (e instanceof PointDistanceExceededException) {
    // ...
} else if (e instanceof MaximumNodesExceededException) {
    // ...
} else {
    // something else
}
```

Fixes #1188


easbar:
Possibly related to #1188, #1201

boldtrn (approved):
That's great, thanks 👍 

Not sure if this is something we can write a test for to make sure this works (but maybe a bit of an overhead for only a very small amount of code :)

otbutz:
@michaz should we use the same exception in the gtfs project?

https://github.com/graphhopper/graphhopper/blob/cb38af67b2d3c0e092babc221d66ceb1c0c6cb4b/reader-gtfs/src/main/java/com/graphhopper/gtfs/PtRouterFreeWalkImpl.java#L283

https://github.com/graphhopper/graphhopper/blob/cb38af67b2d3c0e092babc221d66ceb1c0c6cb4b/reader-gtfs/src/main/java/com/graphhopper/gtfs/PtRouterImpl.java#L392

michaz:
Okay yes.

michaz:
Long-term, I think we should have the Java-API behave more Java-y. Exceptions should be thrown, not put in a list. If something is not thrown, it's .. a custom response object, not an Exception.

Also the convention that IllegalArgumentExceptions map to 4xx errors is bogus anyway. (The Java standard library uses them. You get an IllegalArgumentException for all kinds of stuff. It totally doesn't mean that *the web service user* supplied an illegal argument.)

michaz:
> I agree that the current exception model seems to be build with the REST framework in mind.

Just as a side note: That's impossible, since GHResponse and the Exceptions all predate our use of Dropwizard. It was all hand-written Servlet back then. So whatever conventions are in there, they are in there deliberately, not to please the framework. :-)

easbar:
Can you please add a test and a line in changelog.md for this PR?

otbutz:
#2478

web-api/src/main/java/com/graphhopper/util/exceptions/DetailedRuntimeException.java
```diff
@@ -24,6 +24,8 @@
  */
 public class DetailedRuntimeException extends RuntimeException implements GHException {
 
+    private static final long serialVersionUID = 1L;
```
easbar, line 27: @otbutz Is there any specific reason you added this? From what I understand this is recommended for all classes implementing `Serializable`, and since we extend from `Throwable` here we do implement `Serializable`. Then again I don't see how we would ever serialize an exception, let alone run into a problem because we serialize an exception and it doesn't have a `serialVersionUID`?!

otbutz: > @otbutz Is there any specific reason you added this?

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Serializable.html

> it is *strongly recommended* that all serializable classes explicitly declare serialVersionUID values, since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations, and can thus result in unexpected InvalidClassExceptions during deserialization. 

> Then again I don't see how we would ever serialize an exception

We don't but that doesn't imply that a user of the grahhopper lib won't :)

easbar:
> We don't but that doesn't imply that a user of the graphhopper lib won't :

Yes, I meant 'we' as in 'some random Java devs'. Honestly, who serializes an exception? But yeah if we want to stick to the Java conventions it seems like the recommended thing to do...

michaz:
> Honestly, who serializes an exception?

I'm aware this was a rhetorical question, but here we are:

This used to be quite common. Some time ago, it was considered a good idea to pretend that calling a method locally and calling a service on a server are almost the same thing.

You would have a "service interface", and you would write your code so that it pretends it doesn't care whether that interface is implemented by a regular Java class in your JVM, ...

... OR by a huge machinery that serializes the parameters (with Java's Serializable, of course), calls a remote application server where the interface is ostensibly implemented, and serialize back the return value or any Exception that was thrown, where a million things can go wrong, such as the connection can time out and you get an Exception, but the server still executed the request, because the error happened on the way back, which you have no way of knowing. But hey, it's almost the same thing as calling a method.

michaz:
It's called Enterprise Java Beans and is exactly as much fun as it sounds.

easbar:
Aha good to know :D I'm glad I'm too young to remember these times.

otbutz:
@michaz as someone who has to work with JavaEE....have my angry upvote

PR #2471, opened 2021-11-19
Remove acceptWay optimization

### LINKED ISSUE #2462: Remove acceptWay optimization
It goes like this:

* During pass1 of the OSM import we call `EncodingManager#acceptWay` for every OSM way (except those with only one node or no tags). `acceptWay` calls `getAccess` for each `FlagEncoder` which returns either `WAY`, `FERRY` or `CAN_SKIP` for each way. The value `OTHER` is never used (at least not here). However, it then only checks whether or not any of the encoders returned something other than `CAN_SKIP`. So basically there is nothing to optimize here: We need to call `getAccess` once per way for each encoder. The distinction between `FERRY` and `WAY` is not used.

* pass2 is more interesting: We do the same as in pass1 (filter out all ways for which `EncodingManager#acceptWay` is false, but besides filtering the ways we also need to calculate the edge flags for each way. This is done in `EncodingManager#handleWayTags`, which takes an `AcceptWay` (the result of `EncodingManager#acceptWay` as input parameter. This allows a potential optimization: We can reuse the result of the first `acceptWay` call here. 

However, we do not even do this optimization anymore since #2448. I therefore tried putting it back here: fa7c84b040a903b748d840a4ffa5842c6581eb28 and it turns out there is no measurable speedup, and even more so there is no significant speedup that really justifies this added complexity it seems. I also tried doing this with more (5) flag encoders. Even when I repeated the `acceptWay` check eleven times the slowdown was still only around 50%. I think we should remove this.

The first step would probably be adding `public IntsRef handleWayTags(ReaderWay way, IntsRef relationFlags) ` to `EncodingManager` (without the acceptWay parameter) and calling `acceptWay` inside. But then we should also remove the `ferry` parameter from `TagParser#handleWayTags` and remove the `acceptWay` parameter from `AbstractFlagEncoder#handleWayTags`.

easbar:
Fixes #2462


core/src/test/java/com/graphhopper/routing/util/BikeFlagEncoderTest.java
```diff
@@ -567,18 +567,20 @@ public void testSacScale() {
...
+        IntsRef flags = encodingManager.handleWayTags(osmWay, relFlags);
         assertEquals(PriorityCode.getValue(BEST.getValue()), priorityEnc.getDecimal(false, flags), .1);
 
         // important: UNCHANGED should not get 0 priority!
         osmWay = new ReaderWay(1);
-        osmWay.setTag("highway", "somethingelse");
-        flags = encodingManager.handleWayTags(osmWay, accessMap, encodingManager.createRelationFlags());
+        osmWay.setTag("highway", "track");
+        flags = encodingManager.handleWayTags(osmWay, encodingManager.createRelationFlags());
         assertEquals(PriorityCode.getValue(UNCHANGED.getValue()), priorityEnc.getDecimal(false, flags), .1);
```
easbar, line 583: @karussell do you remember what this comment 

```
// important: UNCHANGED should not get 0 priority!
```

is supposed to say? The issue with this test is that previously we called `AbstractFlagEncoder#getAccess` before `handleWayTags` and passed the `AcceptWay` into `handleWayTags`. This allowed us in this test to pass `Access=WAY` and test the priority even for ways that aren't even accessible. This is why the test used to work even for `highway=somethingelse`. I now changed it to `track` because then we also get `priority==UNCHANGED`. But I don't understand why the comment says the priority should not be 0, but then we do not even test whether the priority is not 0...

core/src/test/java/com/graphhopper/routing/util/CarFlagEncoderTest.java
```diff
@@ -634,9 +620,9 @@ public void testCombination() {
         way.setTag("highway", "cycleway");
         way.setTag("sac_scale", "hiking");
 
-        EncodingManager.AcceptWay map = new EncodingManager.AcceptWay();
-        assertTrue(em.acceptWay(way, map));
-        IntsRef edgeFlags = em.handleWayTags(way, map, em.createRelationFlags());
+        assertEquals(EncodingManager.Access.CAN_SKIP, encoder.getAccess(way));
+        assertNotEquals(EncodingManager.Access.CAN_SKIP, ((AbstractFlagEncoder) em.getEncoder("bike")).getAccess(way));
+        IntsRef edgeFlags = em.handleWayTags(way, em.createRelationFlags());
```
easbar, line 625: Not sure why we use a bike encoder for this car encoder test, but I just kept the logic the same.

core/src/test/java/com/graphhopper/routing/util/RacingBikeFlagEncoderTest.java
```diff
@@ -206,75 +208,92 @@ public void testHandleWayTagsInfluencedByRelation() {
...
-    public void testAvoidanceOfHighMaxSpeed() {
+    public void testPriority_avoidanceOfHighMaxSpeed() {
+        // here we test the priority that would be calculated if the way was accessible (even when it is not)
+        // therefore we need a modified encoder that always yields access=WAY
+        BikeCommonFlagEncoder encoder = new RacingBikeFlagEncoder(new PMap("block_fords=true")) {
+            @Override
+            public EncodingManager.Access getAccess(ReaderWay way) {
+                return WAY;
+            }
+        };
```
easbar, line 219: This is the same problem I described in the other comment: We used to be able to test the priority by ignoring access by just passing in `Access.WAY`. Since this is no longer possible I worked around this by using a modified encoder here. The much better approach would of course be testing the priority calculation in isolation, but this is not possible yet, because `handleWayTags` does a bit of everything.

easbar (commented):
This was pretty straight-forward with the exception of a few tests that test the priority calculation under the assumption that there is access for ways that do not have access, something that never happens in the production code currently. However, in the future this is what we actually might do: Calculate priorities regardless of the access flags as the access flags might change depending on the custom model.

core/src/test/java/com/graphhopper/routing/util/BikeFlagEncoderTest.java
karussell: I think I meant that the default value is 0. And the default meaning should be "unchanged priority" but we do not have an option for this (anymore) and we need to set this UNCHANGED explicitly. But not sure.

Should we ensure that `priority` is UNCHANGED for unknown highway values (?)

core/src/test/java/com/graphhopper/routing/util/CarFlagEncoderTest.java
karussell: Likely because initially there were a couple of bit handling bugs in EncodingManager (and/or EncodedValue) when using multiple flag encoders like overlapping ranges or similar and we needed the test against that somewhere I guess but yes those tests do not belong here.

core/src/test/java/com/graphhopper/routing/util/BikeFlagEncoderTest.java
easbar: > I think I meant that the default value is 0.

Ok, this makes sense. It *does* get the default value 0 as long as the access is false though.

> Should we ensure that priority is UNCHANGED for unknown highway values (?)

Probably, but maybe this is not needed yet (at least for this PR), because for unknown highway values the access will be false anyway. The test only works in master because we pass `Access.WAY` even though in reality it would be `CAN_SKIP`.

core/src/test/java/com/graphhopper/routing/util/CarFlagEncoderTest.java
easbar: Ok, I'll leave it as it is then

PR #2472, opened 2021-11-19
Use a point along the edge geometry to determine areas

### LINKED ISSUE #2467: Use a point along the edge geometry to determine the country/custom_areas
See: https://github.com/graphhopper/graphhopper/pull/2457#discussion_r748023083

otbutz:
Fixes #2467

@cosmin-petrescu i added you as co-author

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -224,12 +224,16 @@ public class OSMReader {
         way.removeTag("country_rule");
         way.removeTag("custom_areas");
 
-        double middleLat = pointList.getLat(pointList.size() / 2), middleLon = pointList.getLon(pointList.size() / 2);
         List<CustomArea> customAreas;
-        if (areaIndex == null || Double.isNaN(middleLat) || Double.isNaN(middleLon)) {
-            customAreas = emptyList();
+        if (areaIndex != null) {
+            double middleLat = pointList.getLat(pointList.size() / 2), middleLon = pointList.getLon(pointList.size() / 2);
+            if (!Double.isNaN(middleLat) && !Double.isNaN(middleLon)) {
+                customAreas = areaIndex.query(middleLat, middleLon);
+            } else {
+                customAreas = emptyList();
+            }
         } else {
-            customAreas = areaIndex.query(middleLat, middleLon);
+            customAreas = emptyList();
         }
 
         // special handling for countries: since they are built-in with GraphHopper they are always fed to the EncodingManager
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -199,7 +199,6 @@ public class OSMReader {
         // Estimate length of ways containing a route tag e.g. for ferry speed calculation
         double firstLat = pointList.getLat(0), firstLon = pointList.getLon(0);
         double lastLat = pointList.getLat(pointList.size() - 1), lastLon = pointList.getLon(pointList.size() - 1);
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -225,9 +223,15 @@ protected void setArtificialWayTags(PointList pointList, ReaderWay way) {
         way.removeTag("country");
         way.removeTag("country_rule");
         way.removeTag("custom_areas");
-        List<CustomArea> customAreas = estimatedCenter == null || areaIndex == null
-                ? emptyList()
-                : areaIndex.query(estimatedCenter.lat, estimatedCenter.lon);
+
+        double middleLat = pointList.getLat(pointList.size() / 2), middleLon = pointList.getLon(pointList.size() / 2);
+        List<CustomArea> customAreas;
+        if (areaIndex == null || Double.isNaN(middleLat) || Double.isNaN(middleLon)) {
```
easbar, line 229: Is it actually expected that the coordinates happen to be `NaN`? Otherwise we could just remove this check (and, very minor, not fetch the coordinates at all when `areaIndex==null`?

otbutz: I just followed the existing code to be honest. Calculating the middle point only if the index is present makes sense.

easbar: Oh, ok. I did not see this. So let's keep it like this (check for NaN) for now.

otbutz:
```java
        List<CustomArea> customAreas;
        if (areaIndex != null) {
            double middleLat = pointList.getLat(pointList.size() / 2), middleLon = pointList.getLon(pointList.size() / 2);
            customAreas = areaIndex.query(middleLat, middleLon);
        } else {
            customAreas = emptyList();
        }
```

The WaySegmentParser seems to be the only place where we deliberately create points with NaN coordinates.

otbutz: Done

easbar: > The WaySegmentParser seems to be the only place where we deliberately create points with NaN coordinates.

Where do you mean exactly?

otbutz: Just checked the codebase if NaN coordinates are a thing and found this:

https://github.com/graphhopper/graphhopper/blob/1e5f4e74f8365bf322bd3ab001afb78e638198e1/core/src/main/java/com/graphhopper/reader/osm/WaySegmentParser.java#L368-L374

easbar: Ah ok, this can probably be removed anyway, now we merged #2457.

easbar: I think the reason some points could be missing in theory was that we left out parts of OSM ways when they weren't within a given bounding box that could be specified for the import at some point.

easbar (approved):

easbar:
Thanks!

cosmin-petrescu:
Thank you!

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: I removed this. cb38af67b2d3c0e092babc221d66ceb1c0c6cb4b

easbar:
I don't have a real-world example right now, but probably we should still use the geometric center for edges with only two nodes? Otherwise we might face the problem that we use a node that is close to the border again.

otbutz:
I'll craft a PR

PR #2474, opened 2021-11-19
Use geometric center to lookup areas for ways with only two points

otbutz:
As discussed with @easbar in https://github.com/graphhopper/graphhopper/issues/2472#issuecomment-974087575


easbar:
Cool, thanks!

PR #2478, opened 2021-11-30
Cover MaximumNodesExceededException with unittest

otbutz:
* extended the existing unittest to check the type of the exception and if the details are properly populated
* mentioned the new exception in the changelog

see https://github.com/graphhopper/graphhopper/pull/2464#issuecomment-982378307


easbar:
Thanks!

Issue #2453, opened 2021-11-01
Allow adding CH preparations to existing graph folder


easbar:
We cannot add CH preparations to an existing graph folder. However, this **does** work for LM preparations. I'd like to get rid of this asymmetry. There are three levels to support this kind of 'delta import' that I can think of:

1) Do not allow adding anything to an existing graph folder. This is the simplest approach, but also the most frustrating one for users. To change anything about the preparations one needs to remove the graph folder and run the import again.
2) Allow adding CH (LM) preparations to an existing graph folder, but only as long as there are no CH (LM) preparations.
3) Allow adding CH (LM) preparations no matter whether there are some already or not.

Currently we support 3) for LM (we can just add more LM preparations), but under no circumstances can we add CH preparations.

I would not have mentioned 2) if there weren't these `prepare.ch/lm.done` flags in the properties file. These flags imply that LM or CH preparation is either "done" or not, so adding preparations would only be possible if there weren't any others yet (case 2)). However, the flag for LM is only written, but never read. The flag for CH is written and read, but I think before it is read adding CH preparations already fails in `checkIfConfiguredAndLoadedWeightingsCompatible()` which checks the `graph.ch.profiles` field of the properties file during `GraphHopperStorage#loadExisting`.

We also have these version flags like `graph.profiles.car.version=-1833937352`, which, are used for LM **and** CH and overwrite each other: https://github.com/graphhopper/graphhopper/blob/e82ca457d03387b54acce638f81ecd533011e53b/core/src/main/java/com/graphhopper/GraphHopper.java#L1057-L1060

I think we should just use `graph.profiles.ch/lm.car.version` (store one separate version hash for each CH and LM preparation, if not just store the **actual** profile) instead of `graph.profiles.car.version` that is currently used for both CH and LM. We should also remove `prepare.lm.done` because it is never read anyway and then we should also remove `prepare.ch.done` so we can add CH preparations just like LM. `graph.ch.profiles` we should also just get rid of.

Here is a little todo list:

* [x] decide whether or not we want to support this whole idea of 'delta imports' -> we do

if yes:
* [x] remove `prepare.lm.done` because it is not used anyway
* [x] replace `graph.profiles.car.version` with `graph.profiles.ch.car.version` and `graph.profiles.lm.car.version`
* [x] remove `prepare.ch.done` and use `graph.profiles.ch.car.version` instead
[...]

PR #2481, opened 2021-12-04
Remove CH from GraphHopperStorage, allow delta import of CHs

### LINKED ISSUE #2388: Remove everything related to CH from `GraphHopperStorage`
We currently create CHGraphs in a rather strange way:

```java
FlagEncoder encoder = new CarFlagEncoder();
EncodingManager encodingManager = EncodingManager.create(encoder);
Weighting fastestWeighting = new FastestWeighting(encoder);
CHConfig chConfig = CHConfig.nodeBased("car", weighting);
GraphHopperStorage ghStorage = new GraphBuilder(encodingManager).addCHConfig(chConfig).create();
CHGraph chGraph = ghStorage.getCHGraph("car");
```

`GraphHopperStorage` is not really related to CH though. All it really does is holding a reference to the CHGraphs and writing 

```
put("shortcuts.version", Constants.VERSION_SHORTCUT);
properties.put("graph.ch.profiles", chProfileNames.toString()); 
```

to the properties file. Other than that it just forwards everything to `BaseGraph`.

So all we really need is a `Map<String, RoutingCHGraph>` somewhere. Maybe we could use a small class that has this map and manages the properties file. It is definitely unrelated to all the `BaseGraph`-like code in `GraphHopperStorage`. The `shortcuts.version` could probably just be included in the `header` of the shortcuts `DataAccess`.

easbar:
Fixes #2388 and #2453.

Our current API to create CH graphs is this:

```java
// encoding manager
FlagEncoder encoder = new CarFlagEncoder();
EncodingManager encodingManager = EncodingManager.create(encoder);
// before we even create the base graph we have to setup the CH graphs and corresponding Weightings
Weighting fastestWeighting = new FastestWeighting(encoder);
CHConfig chConfig = CHConfig.nodeBased("car", weighting);
GraphHopperStorage ghStorage = new GraphBuilder(encodingManager).addCHConfig(chConfig).create();
// run CH preparation
PrepareContractionHierarchies pch = PrepareContractionHierarchies.fromGraphHopperStorage(ghStorage, chConfig);
pch.doWork();
// get the resulting (now no longer empty) CH graph from GHStorage
RoutingCHGraph chGraph = ghStorage.getRoutingCHGraph("car");
```

with the present PR this becomes:

```java
// encoding manager
FlagEncoder encoder = new CarFlagEncoder();
EncodingManager encodingManager = EncodingManager.create(encoder);
// create the base graph (not related to CH at all)
GraphHopperStorage ghStorage = new GraphBuilder(encodingManager).create();
ghStorage.freeze(); // once all edges were added to the base graph it is finished and we freeze it
// now we can add CHs on top of our base graph 
Weighting fastestWeighting = new FastestWeighting(encoder);
CHConfig chConfig = CHConfig.nodeBased("car", weighting);
PrepareContractionHierarchies pch = PrepareContractionHierarchies.fromGraphHopperStorage(ghStorage, chConfig);
PrepareContractionHierarchies.Result result = pch.doWork();
// ... the result of the preparation is the CHStorage, i.e. basically the CH levels and created shortcuts 
CHStorage chStorage = result.getCHStorage();
// we can obtain a CH graph from this:
RoutingCHGraph chGraph = ghStorage.createCHGraph(chStorage, chConfig);

// instead of using `PrepareContractionHierarchies` to build a CH we can also obtain a `CHStorage` using:
ghStorage.createCHStorage(chConfig);
```

I think this is a big improvement, because we no longer need to setup CHs even before the base graph was built. With this PR and some recent commits it was easy to implement CH delta imports (#2453). The base graph is also no longer coupled to CH. 

I'm still not entirely happy with this API, because we still need to call `ghStorage.createCHStorage/Graph` to create the CH structures. This is because `CHStorage` depends on the number of nodes of the frozen base graph and `RoutingCHGraph` needs a reference to the `BaseGraph`, which is private and only visible for `GraphHopperStorage`. Long-term I actually think we should remove `GraphHopperStorage` entirely and make `BaseGraph` public instead, but I'd like to do this in another PR.


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1131,6 +1127,9 @@ public void close() {
         if (ghStorage != null)
             ghStorage.close();
 
+        chGraphs.values().forEach(RoutingCHGraph::close);
+        landmarks.values().forEach(LandmarkStorage::close);
```
easbar, line 1131: So far we only called `ghStorage.close()` here which 'closed' the base graph and the CH graphs, but not LM! Now we explicitly close CH and LM.

Issue #1646, opened 2019-06-27
Tolerance of `heading` is fixed to 100 degrees.


easbar:
The tolerance should be configurable (maybe both as a config parameter and on a per-request basis (?)).

Related: #680

easbar:
A fix for this was contributed in #2482, but currently it is unclear why this feature is needed.

PR #2482, opened 2021-12-06
Added tolerance adjustment in HeadingResolver

kylemikableh:
This references issue #1646 which wanted to be able to adjust the tolerance of the heading. This is my first contribution I am sure I did something wrong!

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/HeadingResolver.java
+++ b/core/src/main/java/com/graphhopper/routing/HeadingResolver.java
@@ -26,10 +26,11 @@ import com.graphhopper.util.shapes.GHPoint;
 
 public class HeadingResolver {
     private final EdgeExplorer edgeExplorer;
-    private double toleranceRad = Math.toRadians(100);
+    private double toleranceRad;
 
     public HeadingResolver(Graph graph) {
         this.edgeExplorer = graph.createEdgeExplorer();
+        toleranceRad = Math.toRadians(100); //default tolerance
     }
 
     //Added constructor to allow configuration of the tolerance (issue #1646)
--- a/core/src/main/java/com/graphhopper/routing/HeadingResolver.java
+++ b/core/src/main/java/com/graphhopper/routing/HeadingResolver.java
@@ -32,6 +32,12 @@ public class HeadingResolver {
         this.edgeExplorer = graph.createEdgeExplorer();
     }
 
+    //Added constructor to allow configuration of the tolerance (issue #1646)
+    public HeadingResolver(Graph graph, int toleranceDec) {
+        this.edgeExplorer = graph.createEdgeExplorer();
+        toleranceRad = Math.toRadians(toleranceDec);
+    }
+
     /**
      * Returns a list of edge IDs of edges adjacent to the given base node that do *not* have the same or a similar
      * heading as the given heading. If for example the tolerance is 45 degrees this method returns all edges for which
--- a/core/src/main/java/com/graphhopper/routing/Router.java
+++ b/core/src/main/java/com/graphhopper/routing/Router.java
@@ -224,12 +224,13 @@ public class Router {
         PathCalculator pathCalculator = solver.createPathCalculator(queryGraph);
         boolean passThrough = getPassThrough(request.getHints());
         boolean forceCurbsides = getForceCurbsides(request.getHints());
+        int tolerance = request.getTolerance();
         if (passThrough)
             throw new IllegalArgumentException("Alternative paths and " + PASS_THROUGH + " at the same time is currently not supported");
         if (!request.getCurbsides().isEmpty())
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/graphhopper/routing/HeadingResolver.java
```diff
@@ -26,10 +26,17 @@
...
 
     public HeadingResolver(Graph graph) {
         this.edgeExplorer = graph.createEdgeExplorer();
+        toleranceRad = Math.toRadians(100); //default tolerance
+    }
+
+    //Added constructor to allow configuration of the tolerance (issue #1646)
+    public HeadingResolver(Graph graph, int toleranceDec) {
+        this.edgeExplorer = graph.createEdgeExplorer();
+        toleranceRad = Math.toRadians(toleranceDec);
```
easbar, line 39: Such comments, like "Added constructor..." are nice for a reviewer to understand what your attention was, but you can just add them here in github and they typically should not remain in the code directly. I mean stating your intention in comments can be a good thing, but for something simple like a new constructor it is not necessary, it is already clear when looking at the constructor parameters that this constructor allows setting the tolerance.

Another thing: It is usually a good idea to chain the constructors. Here the first one could just call the other like `this(graph, 100)` for example.

core/src/main/java/com/graphhopper/routing/Router.java
```diff
@@ -224,12 +224,13 @@ protected GHResponse routeAlt(GHRequest request, Solver solver) {
...
         boolean passThrough = getPassThrough(request.getHints());
         boolean forceCurbsides = getForceCurbsides(request.getHints());
+        int tolerance = request.getTolerance();
         if (passThrough)
             throw new IllegalArgumentException("Alternative paths and " + PASS_THROUGH + " at the same time is currently not supported");
         if (!request.getCurbsides().isEmpty())
             throw new IllegalArgumentException("Alternative paths do not support the " + CURBSIDE + " parameter yet");
 
-        ViaRouting.Result result = ViaRouting.calcPaths(request.getPoints(), queryGraph, snaps, solver.weighting, pathCalculator, request.getCurbsides(), forceCurbsides, request.getHeadings(), passThrough);
+        ViaRouting.Result result = ViaRouting.calcPaths(request.getPoints(), queryGraph, snaps, solver.weighting, pathCalculator, request.getCurbsides(), forceCurbsides, request.getHeadings(), passThrough, tolerance);
```
easbar, line 233: In this context it would be good to indicate the tolerance is related to the heading, I would just call it `headingTolerance` (here and in `GHRequest`).

core/src/test/java/com/graphhopper/routing/HeadingRoutingTest.java
```diff
@@ -164,6 +166,52 @@ public void headingTest5() {
         assertArrayEquals(new int[]{5, 4, 3, 8, 7, 6, 5, 4, 3, 2}, calcNodes(graph, response.getAll().get(0)));
     }
 
+    @Test
+    public void headingTestWithToleranceRoundTrip() {
+        // Test enforce start direction
+        GraphHopperStorage graph = createSquareGraph();
+        int tolerance = 80;
+        Router router = createRouter(graph);
```
easbar, line 174: In this test I would calculate two routes with two different values for the tolerance. This way we can be sure the tolerance actually makes a difference (now and in the future when some of the code changes).

web-api/src/main/java/com/graphhopper/GHRequest.java
```diff
@@ -37,6 +37,7 @@
 public class GHRequest {
     private List<GHPoint> points;
     private String profile = "";
+    private final int DEFAULT_TOLERANCE = 100; //Default tolerance that can be changed
```
easbar, line 40: Typically such constants should be static. But since this one is not really used anywhere else you could also just write `private int tolerance = 100` below.

easbar (commented):
Thanks a lot for contributing to GraphHopper and sending your first contribution. I think you did a good job here. You picked an open issue, and even one that has the `good first issue` label (Is this the reason you picked this specific issue?) You sent a PR that provides an isolated fix for this issue, without making changes to unrelated code. You also provided some tests for your new feature and explained what you did. :+1:

However, quite unfortunately, things kept moving since we opened the corresponding issue #1646 three years ago. Therefore, we currently do not want to add the heading tolerance, mostly because we do not think it is of much use (contrary to what we thought back then). This is probably rather disappointing for you, maybe even more so, because it took us a while to respond. But I hope you can understand that this is how it goes sometimes in open source programming. I think it is mostly our fault to not have removed the 'good first issue' label for this issue. The only thing you could do better next time is asking if a certain feature is still of interest next time.

I tried my best giving you some feedback in the review comments below, I hope this will be of some help for you.

I'm closing here now, but if you still think this feature is important for some use-cases please feel free to explain and we can re-open then. I'm currently assuming you picked this issue, because it was there, not because you are using the feature yourself (?!)

PR #2483, opened 2021-12-07
Add route_type to public transport leg

ainar:
Hello,
I just added the route_type field to public transport legs. I think this was necessary. See more on [my post on the forum](https://discuss.graphhopper.com/t/get-the-type-of-vehicle-when-querying-public-transport-itinerary/6964).
There is still the test to update/to add.
Aina.


ainar:
Hey, sorry I was not active for a while. Is there any plan to add such a feature?

Issue #2053, opened 2020-06-05
JDK9+ Jigsaw modules support


oli0044:
Hi,

I'm trying to add the maven dependency graphhopper-reader-osm as suggested in the docs to a JDK 11 Java project using the new Jigsaw module system (i.e. module-info.java exists) ... [, because the Jigsaw module system is needed by other dependencies].

Up to graphhopper version 0.10.3 it works peachy, but since version 0.11.0 up to the current 1.0 the project does not compile anymore with the error:
`reads package com.graphhopper from both graphhopper.api and graphhopper.core`

This is due to the fact, that since 0.11.0 some classes from graphhopper-core were split off into graphhopper-api, but the new Jigsaw module system needs the different modules/dependency-jars to be disjoint package-wise. So for example: com.graphhopper.util has some classes in maven dependency graphhopper-ore and some in graphhopper-api, but this is not supported anymore when using the Jigsaw module system since JDK9+ because it needs to have com.graphhopper.util only be exported by one module/dependency.

 - Is there a plan to support the Jigsaw module system in the future with embedded module-info.java files?
 - Is it possible to make a third (optional) maven dependency/module "graphhopper-core-plus-api" which combines graphhopper-core and graphhopper-apiback into one dependency again? (Which - I think - would solve the problem above)
 - Or am I just being dumb and doin something wrong... and the support for the Jigsaw module system is working already all along?

I might maybe be able to suggest a pull request with the second option "combined graphhopper-core-plus-api". But as an only intermediate maven user I would highly appreciate any push in the right direction, or maybe a "stop" if this is the wrong direction.

Thanks for the great work with Graphhopper in any case :-)

Greetings from Germany

Oli4

easbar:
If jigsaw standard does not allow same package names in different modules I think we should not do it. Especially (core) `com.graphhopper.util` is quite a mess anyway (it has many unrelated classes) and should be easy to rename. `com.graphhopper` only has a few classes so they could be moved to some sub-package? Are these the only two packages that cause a problem?

oli0044:
Hi easbar,

thanks for taking this seriously :-).

As far as I see it, It's the packages:

`com.graphhopper`
`com.graphhopper.col`
`com.graphhopper.routing.util`
`com.graphhopper.util`

[...]

PR #2487, opened 2021-12-10
Rename web/com.graphhopper.http to web/com.graphhopper.application

easbar:
For Jigsaw compatibility we must not use the same package across different Maven modules, see #2053. Therefore we cannot use `com.graphhopper.http` in both `web` and `web-bundle` as we currently do. Here I renamed `web/com.graphhopper.http` to `web/com.graphhopper.app` for this reason. *Update* it is now called `web/com.graphhopper.application`.

The only real downside is that now `com.graphhopper.http.GraphHopperApplication` is `com.graphhopper.app.GraphHopperApplication` which means the main class used to start the GH server changed its name and `java` commands using the old name will no longer work. Maybe this is ok, because we just removed the `graphhopper.sh` script and therefore the procedure of starting the server changes between 4.x to 5.0 anyway.

Another alternative would be keeping `web/com.graphhopper.http` as  is and renaming `web-bundle/com.graphhopper.http` instead. I decided against this because it leads to more (downstream) Java code changes and the name `app` seems to be a better fit for the classes in the `web` module (for example it not only has the server but also `ImportCommand`.


boldtrn:
I personally find `.app` a bit misleading. App often implies some sort of frontend. For me the main purpose of the .http package is to wire the server, so I would propose to go with `com.graphhopper.server`.

karussell:
GraphHopper**App**lication :wink:

otbutz:
> main class used to start the GH server changed its name and `java` commands using the old name will no longer work.

I think most people simply start GH via `java -jar` which retrieves the main class from the jars Manifest. I wouldn't be to concerned about possible breakage here.

boldtrn:
> GraphHopperApplication 😉 (i.e. we already use this "app" naming)

I already thought that you would reply like that :). I still think the naming is not really speaking for itself. 

If you see a package:
`com.example.app`

Would you expect this to contain server code or UI code? I would expect UI code.

easbar:
> I think most people simply start GH via java -jar which retrieves the main class from the jars Manifest. I wouldn't be to concerned about possible breakage here.

Ok, yes, this is true. I already updated the name in the manifest file.

> Would you expect this to contain server code or UI code? I would expect UI code.

Honestly, I would not expect either. I would just see there is some application that I can run. And in fact there **is** a CLI application and a server, so just calling it 'server' would also be misleading to some extent. Then again the maven module is called `web` which causes the same confusion :)

otbutz:
If `web-bundle` is the `GraphHopper Dropwizard Bundle` we could also reflect this in to package naming. eg `com.graphhopper.dropwizard.GraphHopperBundle`

easbar:
> If web-bundle is the GraphHopper Dropwizard Bundle we could also reflect this in package naming. eg com.graphhopper.dropwizard.GraphHopperBundle

It is to some extent related to Dropwizard, but I wouldn't call it like this, because `RouteResource` etc. are close to JAX-RS and in theory can be used with other web frameworks etc.

otbutz:
I wouldn't touch the `resource` package to be honest. But the `com.graphhopper.http` package within `web-bundle` seems to be mostly concerned with dropwizard.

easbar:
Ok, that's true it is just `com.graphhopper.resources` not `com.graphhopper.http.resources` currently.

easbar:
How about `com.graphhopper.application`? That would still be general enough to cover cli+server, it is just a bit longer. And maybe people don't associate this with a frontend/mobile "app" as much as they do when they read 'app'?

boldtrn:
Sounds good, I personally would prefer application over app in this case.

easbar:
Unfortunately this does not work yet. For example MapMatchingResourceTest fails with an NPE (the `gpx` parameter is `null`). Probably we rely on some kind of Dropwizard magic that requires certain things to be in the same package, at least with our current setup.

easbar:
> Probably we rely on some kind of Dropwizard magic that requires certain things to be in the same package, at least with our current setup.

Ah never mind. I just forgot renaming the `src/test/resources/http` folder, and got further confused, because there were two `tour-with-loop2.gpx` files...

PR #2489, opened 2021-12-10
Avoid Pattern compilation

otbutz:
The old code would need to compile two Pattern instance for each call of `simpleXMLEscape()` as `String.replaceAll()` requires that a Pattern is compiled on each call. The PR switches one of the replaceAll() calls to a simple replace() as it doesn't require regex functionality. The other one is compiled once and gets reused.


karussell:
Did you have a problem in production due to this?

> The PR switches one of the replaceAll() calls to a simple replace() as it doesn't require regex functionality. 

:+1:

otbutz:
Just Sonarlint screaming on top of its lungs :smile:

otbutz:
Any objections?

easbar:
Thanks

Issue #2375, opened 2021-07-27
CH preparation becomes very slow when there are large-weight edges


easbar:
For example if we enable `GermanyCountryRule` many track highways are set to use `RoadAccess.Destination` which strongly increases the corresponding edge weights. This slows down CH preparation as witness searches explore large parts of the graph.
For Germany I measured a x3 slow down for node- and edge-based CH when enabling country rules.

See also comment here: https://github.com/graphhopper/graphhopper/pull/2353#issuecomment-886980293

easbar:
I did not notice it back then, but #2390 already improved the situation quite a bit. I just measured the CH preparation time with and without the German country rule  before and after #2390, using the bayern-190101.osm.pbf map:

||before #2390|after #2390|
|-|-|-|
|without country rules|60s|52s|
|with country rules|178s|63s|

So #2390 actually improved the CH preparation time by a factor of about three in case the country rule is included. Since #2390 the slow down that is due to the German country rule is less than 20%, while it was around x3 before, according to this little experiment.

However, since #2390 only affects the node-based CH preparation this improvement only exists for node-based CH. For edge-based CH I just repeated the same experiment and found that the preparation takes around 800s without the country rule and around 2040s with the country rule (2.5x as long).

otbutz:
Looking forward to the country rules being enabled by default :smiley:

easbar:
Update: After #2491 the preparation including the country rule is now faster than it was before without the country rule. The country rule still slows down the preparation by about 20%, but for node-based I consider this issue as solved for now. I we really wanted to improve this further we could try tuning the parameters introduced in #2491 including the country rules, but for now improving edge-based CH is a lot more critical.

PR #2491, opened 2021-12-11
Use a fixed number of maximum visited nodes for node-based CH

### LINKED ISSUE #1810: Bugs in node-based CH preparation heuristics

easbar:
This PR fixes #1810 and improves the node-based CH performance.

So far we were limiting the number of visited nodes for witness path searches based on the (estimated) 'mean degree' of the graph. However, as pointed out in #1810 there were two critical bugs in this calculation. First, the visited nodes limit was zero when building the initial queue (all witness searches were cancelled immediately) and second the cast to an int meant that the limit was also zero if `meanDegree < 1` and otherwise it would only ever be a multiple of `100`. The third bug pointed out in #1810 is obsolete as we removed the 'contracted neighbor' heuristic in the meantime.

I did new parameter studies for the node-based CH preparation and found that we get better results using a simpler approach. Instead of adjusting the visited node limit based on the mean degree of the graph I just set it to a fixed number. However, I use a different number for the heuristic calculation for the node ordering and for the actual node contraction. This way we can save time during the witness path searches that are only done to determine the node ordering without skipping too many witness path searches when actually contracting a node. I found that using a value of 30 for the heuristics and a value of 200 for the contraction work quite well. This can depend on the map and exact use-case, so it might be worthwhile to tune these values further, especially when there is a preference for a faster preparation, faster queries or less shortcuts/memory usage. 

The other thing I found out is that using (only) neighbor updates, and no periodic and no lazy updates at all works better than what we are currently doing in master. With the set of parameters I use in this PR there are less shortcuts, the preparation is faster and the query times are equal (if not faster). I tested this using maps of different sizes (Bavaria, Germany and planet). To give an example: For master I ran the preparation for planet in 4943s vs. this branch 4556s (-8%), the number of shortcuts decreased from 221mio to 194mio (-14%) and the average query time was around 1.51ms for both. Pretty good, I would say, considering that I only removed code or simplified the algorithm :)

With the new parameters the preparation is faster than before, even when we enable the `GermanyCountryRule` that uses access=destination for all tracks (related to #2375).

The only 'downside' to this PR is that now the gap between node- and edge-based CH performance is even bigger. But hopefully some of these ideas will work for edge-based CH as well.


easbar:
Btw these two parameters:

```
private float edgeDifferenceWeight = 10;
private float originalEdgesCountWeight = 1;
```

seem to be a lot less critical for the CH performance than the two new ones. I am still running a few tests changing these, but it seems like they don't make such a big difference at all.

Issue #2498, opened 2021-12-12
setAllowWrites(false) makes GraphHopper.load() fail even though no write access is needed.


oli0044:
**Describe the bug**
When setting `setAllowWrites(false)` for non-writable file systems the `GraphHopper.load()` function fails in the subfunction `prepareCH()` because `prepareCH()` requires write access, even when there are no CHProfiles to be prepared.

The prepareCH() function throws an `IllegalStateException("Writes are not allowed!")`.

**To Reproduce**

```
GraphHopperConfig config = new GraphHopperConfig(); // iniitalize parameter map
config.putObject("index.max_region_search", 8); // define more iterations to find the points more accurately
config.getProfiles().add(new Profile("car").setVehicle("car").setWeighting("fastest")); // add the profile
config.getCHProfiles().add(new CHProfile("car")); // set the rofile for speed mode

GraphHopper hopper = new GraphHopper().setGraphHopperLocation(distancesFolder.getAbsolutePath()) // set working directory
                                   .setPreciseIndexResolution(1000) // define higher resolution to find the points more accurately
                                   .init(config) // start graphhopper instance
                                   .setAllowWrites(false) // only read the file
                                   .setGraphHopperLocation(distancesFolder.getAbsolutePath());
hopper.load(); // load gaphhopper
```

**Expected behavior**
The `prepareCH()` function should not throw an `IllegalStateException("Writes are not allowed!")` if no CHProfile has to be prepared.

**System Information**

operating system: Ubuntu 20.04
JDK: java-11-openjdk-amd64
GraphHopper: current master branch

oli0044:
See pull request #2499

easbar:
Yes, I think this is a bug. But before we come up with a fix we need to clarify what this `isAllowedWrites` flag is even supposed to do. This part of GraphHopper needs an overhaul anyway, if you ask me. Related to: #1895.

So why should we `setAllowWrites(false)` when we call `load`? If we were **loading** why should we expect anything to be **written** and then disallow such writing? This sounds obvious, but actually you are right in a way, because the `load()` method **can** actually write. For LM (and since #2481 also for CH) **load** means: "load  the available preparations, but create the others when they are still missing". So it kind of makes sense to disallow writing when calling load, because a somewhat reasonable assumption would be that calling `setAllowWrites(false)` just before `load()` would mean that the loading will be done as usual, but this 'delta-import' feature was disabled. However, this would be a rather strange, not to say terrible, API for doing this.

[...]

PR #2499, opened 2021-12-12
Issue 2498: Prevent write access check in prepareCH if it's not needed.

oli0044:
see #2498


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -1039,7 +1039,9 @@ else if (prepared.containsKey(profile.getProfile())) {
     }
 
     protected Map<String, PrepareContractionHierarchies.Result> prepareCH(boolean closeEarly, List<CHConfig> configsToPrepare) {
-        ensureWriteAccess();
+        if (!closeEarly && !configsToPrepare.isEmpty()) { // do not check write access if no ch profiles have to be prepared and no ealry closing is set
+            ensureWriteAccess();
+        }
```
easbar, line 1044: I think we should deal with this situation at an earlier point. When `prepareCH` is called it means that CHs should be prepared (written). So if you meant to do a 'pure load' this should be prevented on a higher level. Either you need to take care of this in your configuration (do not add more CH profiles than there are on disk so no delta imports will run), or `loadOrPrepareCH` should offer something like `deltaImport=true`.

oli0044: Well that is exactely the problem: Although there are no additional CH profiles added, `loadOrPrepareCH` still calls `prepareCH(closeEarly, configsToPrepare)` with an empty `List<CHConfig> configsToPrepare`.

So we could check if `List<CHConfig> configsToPrepare` is empty before calling `prepareCH` altogether... however, inside of `prepareCH` there is some `closeEarly` functionality that flushes and closes the `ghStorage`, which is used when called via `importAndClose() -> process(closeEarly = true) -> postProcessing(closeEarly = true) -> loadOrPrepareCH(closeEarly = true) -> prepareCH(closeEarly = true, configsToPrepare)`.

If I prevent `prepareCH` from being called when there are no ` configsToPrepare`, then the `closeEarly` code is not applied.

Can the `closeEarly` code be ignored? Does it belong in `prepareCH`?

michaz:
I think an incremental improvement (so my customer can load their contraction hierarchies on their read-only filesystem again) is to just remove the `ensureWriteAccess()`.

This now means that in the case where the user said `setAllowWrites(false)` *and* there are CHs to build, this will now be violated.

But if the intention of this option was to let GraphHopper be well-behaved on an actual read-only file system (and it was!), that doesn't matter -- it will just fail when it tries to write, and that is less often than _now_, where it always fails, even if it doesn't actually want to write.

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: Ok I see.

> Can the closeEarly code be ignored?  Does it belong in prepareCH?

We should probably not ignore it, but it definitely does not belong in `prepareCH`.

easbar:
Superseded by #2500

PR #2503, opened 2021-12-20
broken wheelchair routing - priority and elevation improvements

### LINKED ISSUE #2497: wheelchair does not work for fastest weighting
See https://discuss.graphhopper.com/t/strange-wheelchair-routing/6982

karussell:
Should fix #2497.

The problem was that the priority encoder was never written too leading to an infinite weight. Additionally, when using elevation, the same could happen for short routes with a too big "elevation noise". 

Can you try again with these changes @zdila?


zdila:
Thanks! I will try it within couple of days.

zdila:
Works for me. Thank you!

zdila:
BTW Graphopper is able to find a route now, but routes are often strange, not using footways but roads for cars.

Also for example check this: https://www.freemap.sk/?map=18/48.724220/21.236095&layers=X&tool=route-planner&points=48.724433/21.236876%2C48.723818/21.234326&transport=foot-stroller - it is using latest GH from yesterday with this PR merged and also data from yesterday. It navigates the wheelchair from a footway to the road in opposite direction, then "u-turn" and then it jumps up on the bridge on the road in the construction (there is no shared node). Uses wheelchair with any weighting.

karussell:
> then it jumps up on the bridge on the road in the construction

I can't reproduce this behaviour - a highway=construction way cannot be visited except there is an explicit wheelchair=yes tag. Are you using an unmodified version of GH? Which config are you using?

btw: the opposite direction of a road should be legal for a wheelchair.

> there is no shared node

What do you mean here?

zdila:
I am actually using modified GH only to give marked trails higher priority:
https://github.com/FreemapSlovakia/graphhopper/commit/16f084a48aa44d204cdab8b4fefdee1d9ed51cb7#diff-ff39f67e2307d6318e51faed22356a33d312eccd88ff854234aa8e89d470a3ef

And also different elevation provider: https://github.com/FreemapSlovakia/graphhopper/commit/87884a2defe6553e6aad43dc3650e60253c6edd7

>> there is no shared node
> What do you mean here?

I mean that bridge and road beneath doesn't share a node:

https://www.freemap.sk/?map=19/48.724215/21.236521&layers=X&point=48.724226/21.236518

zdila:
Actually I was wrong, it is not using the bridge, but this area https://www.openstreetmap.org/way/561084192

I don't think that border of amenity=hospital should be used for any routing, even if it has wheelchair=yes. Should I report it as a different issue?

karussell:
Ah, thanks. Ugly. This is another bug and relates to this PR here - will fix it here.

core/src/main/java/com/graphhopper/routing/util/WheelchairFlagEncoder.java
```diff
@@ -75,12 +75,15 @@ protected WheelchairFlagEncoder(int speedBits, double speedFactor) {
         safeHighwayTags.add("service");
         safeHighwayTags.add("platform");
 
-        excludeHighwayTags.add("trunk");
-        excludeHighwayTags.add("trunk_link");
-        excludeHighwayTags.add("primary");
-        excludeHighwayTags.add("primary_link");
-        excludeHighwayTags.add("steps");
-        excludeHighwayTags.add("track");
```
karussell, line 88: `excludeHighwayTags` wasn't properly working anyway, so I removed this difference and it is now using the same highways foot uses.

karussell:
I think my last change should fix the problem - can you try again?

zdila:
Tested now, but is still seems to use the edge of that area for wheelchair routing.

karussell:
The way will be excluded for the most recent version:

image

The Wheelchair encoder now uses the same whitelist approach for highway types like the other encoders and this shouldn't happen.

Can you test the original GH code and with this map extract? [map.zip](https://github.com/graphhopper/graphhopper/files/7801334/map.zip)

zdila:
I tried graphopper from master (d358d4b8a) with data from https://download.geofabrik.de/europe/slovakia.html and wheelchair routing doesn't use the area edge (but doesn't work at all with `fastest` weighting). Is it enough? Or is your `map.zip` somewhat special?

karussell:
Can you try this branch `wheelchair_fix`? The change is not yet in `master`.

zdila:
I am slightly confused now :).

[I tried this branch today morning](https://github.com/graphhopper/graphhopper/pull/2503#issuecomment-1003946960) with https://download.geofabrik.de/europe/slovakia.html and the routing on area edge was still present. It is commit 29a73c3035c680a8fa26dd01e7ead63013682297.

zdila:
hm, tried again with clean-build and now it works properly with this branch.

karussell:
Thanks!

Issue #196, opened 2014-05-06
Allow oneways and foot-only roads for bike


karussell:
http://graphhopper.com/maps/?point=49.952383%2C11.578732&point=49.953391%2C11.578404&vehicle=bike

Allow opposite oneways and foot-only roads and set `get_off_bike` to `true`. But assign not only low speed like 5km/h but also low priority.

fleipold:
This is an important issue for urban use cases. I am not sure whether the link above is still demonstrating the problem.
I used this one (where the suggested route is more than twice as long): https://graphhopper.com/maps/?point=52.496249%2C13.420917&point=52.497328%2C13.415364&locale=en-GB&elevation=true&profile=bike&use_miles=false&selected_detail=Elevation&layer=Omniscale

From reading the discussions it seems that workarounds are possible, but none is given. Perhaps it would make sense to link a workaround from here, if it exists.

jp-lopez:
I'm trying to solve that [here](https://github.com/graphhopper/graphhopper/pull/2747/files). Can you take a look when you can @karussell ? Still some tests missing, but to know if is the good approach. Thanks

PR #2504, opened 2022-01-03
Add restricted access values "agricultural", "forestry", and "delivery" for BikeCommonFlagEncoder

ratrun:
See discussion in  https://discuss.graphhopper.com/t/bicycle-allowed-on-vehicle-delivery-or-vehicle-forestry/6983/3


karussell:
Thanks for the quick fix!

The problem with such a hard constraint is probably very similar to #196 as the person could get off the bike in case there would be very large detours. But this is a different issue IMO.

karussell:
Moved the test into AbstractBikeFlagEncoderTester.

Issue #2512, opened 2022-01-31
Support for superroutes in bicycle routing


ratrun:
This is not a bug, but rather a feature request: For bicycle routing we make a difference if a highway belongs to a regional, a national, or an international cycleroute: https://github.com/graphhopper/graphhopper/blob/87130a37c8147f060ef5560dbe979b0ceeecbad4/core/src/test/java/com/graphhopper/routing/util/BikeFlagEncoderTest.java#L520-L526 
This takes into regard that usually international or national cycle routes are better equipped compared to regional ones.

For Eurovelo routes it becomes more and more popular to use relations of `type=superroute` , which contain references to other relations of `type=route`.  See https://wiki.openstreetmap.org/wiki/Relation:superroute . The problem in graphhopper is that currently we treat the superroute only with the `network` priority for their members and not of the superroute itself.
One example is the EV9 superroute for Austria: https://www.openstreetmap.org/relation/2766147. Its member relations are tagged as `network=icn` and contain references to regional cycle routes such as https://www.openstreetmap.org/relation/48137. These member relations are tagged as `network=rcn` . Graphhopper currently uses this `rcn` value. The request of this issue is to use the value of the superroute, if there is one.

The implementation of this is not trivial as the member relations could "loop", if incorretly referenced in OSM. 
I believe that it is sufficient to support the nesting of one level: Support for a super-super relation like https://www.openstreetmap.org/relation/2938 is not needed. Maybe this information helps: https://wiki.openstreetmap.org/wiki/Super-relation/Implementation

ratrun:
Closing now, because the PR would add to much complexity, see discussion in #2516.

PR #2516, opened 2022-02-08
Implementation and test of superroute support for bicycle routes.

ratrun:
See #2512.
This implementation is avoiding a third pass, but requires some memory as it keeps all ways, which are referenced by bicycle route relations. The memory for the storage of the superroutes itself should be neglectable.


easbar:
Thanks for working on this.

> This implementation is avoiding a third pass, but requires some memory as it keeps all ways, which are referenced by bicycle route relations

Can you be a bit more specific here? Did you test the memory usage for, say, Germany or Europe (or planet)? How large are these collections going to be for Germany/Europe/planet? Also using `LongObjectMap<String>` and `LongObjectMap<LongArrayList>` will reduce memory, but first we should check the size / number of elements of these collections.

```
    // stores route relation OSM ID and the "network" tag value of a superroute
    private HashMap<Long, String> superRouteRouteMembers = new HashMap<Long, String>();
    // stores all the ways of a route relation for route=bicycle relations in order to process superroutes
    private HashMap<Long, ArrayList<Long>> bicycleRouteWayMembers = new HashMap<Long, ArrayList<Long>>();
```

Depending on the additional memory usage this feature should probably be opt-in so the import performance does not change in case one does not need it (?).

ratrun:
I checked the result for Germany: `bicycleRouteWayMembers.size(): 26117 superRouteRouteMembers.size(): 1377`
Do you think this is worth a new configuration parameter such that the default (e.g. for car) is to not process the bicycle superroutes?
The optimal solution would be to release the memory after the pass1 has finished. But so far I was not able to figure out how this could be implemented.

ratrun:
Now I implemented the releasing of the memory after pass1: [715c7d4](https://github.com/graphhopper/graphhopper/pull/2516/commits/715c7d453b8bc4d175b7db80348a7575b69c2f7d).

ratrun:
@karussell Please provide feeback on this one, it looks to get lost. What do you want me to change or is there no chance that this gets merged?

easbar:
I don't think this PR will get lost in that it does change so much code and it should be easy to update to master. There are lots of code changes in master currently, but `OSMReader` won't be strongly affected.

What would really help merging this PR is some statistics about the number of relations, collection sizes, memory usage and import time when using planet OSM. If the processing is potentially heavy-weight in terms of memory or import time an option to turn off the feature would be useful as well. And maybe some examples that clearly show how this feature improves the bike routing. If this sounds like a lot of work please keep in mind that if you don't do this we have to do it and since it is currently not our priority this PR did not get so much attention so far. Some easy to understand examples would also help increasing our priority for this PR :)

ratrun:
@easbar Thanks for the feedback.

> What would really help merging this PR is some statistics about the number of relations, collection sizes, memory usage and import time when using planet OSM.

Sorry, I don't have the means to run it for the planet. But I don't expect noteable increase in memory usage now that all the memory gets released after pass1. Therefore I don't believe that an option to turn off the feature is needed. But if you want me to add it I'll do it.

> And maybe some examples that clearly show how this feature improves the bike routing.

Please note that I already have included an artificial testcase which exactly does that. See core/files/test-superroute.osm here opened in JOSM:

image

My testcase routes from North to South and checks that the zig-zag route and not the straight line is taken.

ratrun:
Thanks for checking.

Here is one prominent real world example: The EV9 in Austria for this route: https://graphhopper.com/maps/?point=47.552028%2C16.090164&point=47.511694%2C16.042957&locale=de&elevation=true&profile=bike&use_miles=false&selected_detail=Elevation&layer=TF%20Cycle . 

If you look more closely, I'm sure you will find lots of more differences. With the suggested change the EV9 in the west and not the main road is taken. 

It is true that the superroute tag is not very often used, but you need to be aware that each bicycle superroute contains route members with lots of indiviual way segments. They are mainly used in Europe in order to be able to reasonably manage the EuroVelos (See https://de.eurovelo.com). So in the end it has an effect on thousands of kilometers of the most promintent cycle ways.

> you avoid a 3rd pass because you store all bike relations in bicycleRouteWayMembers? 

Yes

> Testing this against planet would be required but might be possible due to your nice cleanup after pass1.

I would ask you to give it a try, if possible.

> why do you have to call osmParsers.handleRelationTags twice? (one call for type=superroute and one new call in type=route)

The IDs in the OSM file are sorted with lower values first. Besides for the nodes and ways this is also true for the OSM relation IDs. Here there are two cases: The more likely situation is that intitialy all the member relations are already there when the superroute is created. Here the route member IDs are all lower compared to the ID of the supperoute. But if somebody changes the type of a bicycle route into a bicycle superroute later this is might not be the case any longer. I discovered both situations when I looked at real-world superroutes in Austria.

> So really not sure if we should include this feature.

I hope that I was able to convice you. 😄

ratrun:
> As this is now following the official bike route ... I think the route you meant was [this](https://graphhopper.com/maps/?point=47.552028%2C16.090164&point=47.531748%2C16.071882&point=47.511694%2C16.042957&profile=bike&layer=TF%20Cycle). 

Yes, exactly. Please note that somebody changed back the EV9 from a super route back to an ordinary route [5 days ago](https://www.openstreetmap.org/changeset/124742215#map=8/47.711/16.233&layers=C). This is the reason for the changed routing in the meantime.

>why should we fix these cases via introducing superroutes instead of just e.g. increasing the priority for the official bike routes?

Because the superroutes will most likely contain the best route compared to other minor cycle routes nearby. I consider the Eurovelos as the "motorways for bicycles".

ratrun:
>If we would introduce e.g. area routing that require more complex relations parsing or a third pass or similar, then I would lean towards of accepting this complexity as this gives us an important new feature, but here I fear it will only give us an improvement that we similarly can get by an easier approach.

I understand your argument that you don't want to add too much complexity to the OSM reader, therefore I'm closing this PR now.

PR #2517, opened 2022-02-15
Node out of bounds error when using average_speed path detail

easbar:
For a `car` profile with turn costs, e.g. this route http://localhost:8989/maps/?point=50.018339%2C11.502782&point=50.017515%2C11.501698&point=50.017129%2C11.504467&locale=en-US&elevation=false&profile=car&use_miles=false&layer=Omniscale&details=average_speed fails with an error like

```
node: 1942 out of bounds [0,1942[
```

(depending on the size of the map). The reason is that the turn cost storage is queried with a virtual node ID (node 1942 is a virtual node and thus beyond the range of valid node IDs). The error only occurs when the `average_speed` path detail is enabled and a virtual node is part of the route between two via points. So in the simplest case a route with one via-point A-B-C is such that we visit A again when we go from B to C. Otherwise the error does not occur, because we do not consider the turn costs at via points when calculating `average_speed` (and the same for instructions). I already fixed this here: 0c23614503fdc940a09192ac43bc5f0d48fb32c2.

### DIFF
--- a/core/src/test/java/com/graphhopper/GraphHopperTest.java
+++ b/core/src/test/java/com/graphhopper/GraphHopperTest.java
@@ -2452,4 +2452,26 @@ public class GraphHopperTest {
         }
     }
 
+    @Test
+    void averageSpeedPathDetailBug() {
+        final String profile = "profile";
+        GraphHopper hopper = new GraphHopper()
+                .setProfiles(new Profile(profile).setVehicle("car").setWeighting("fastest").setTurnCosts(true).putHint(U_TURN_COSTS, 80))
+                .setGraphHopperLocation(GH_LOCATION)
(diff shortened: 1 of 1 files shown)


easbar:
This bug had been around for quite some time, but still we never noticed it. I think the reason is that `details=average_speed` is not used too often and additionally the problem seems to happen a lot less frequently with CH, because due to the CH shortcuts it is a lot more likely that we 'skip' the virtual node. For GH maps `details=average_speed` is disabled by default so we have to enable it explicitly to reproduce the error. For GH maps 2 it is in fact enabled, but so far we did not use `ch.disable=true` which is probably why we only saw this error now. Finally, the error would be easier to find if it resulted in a 500 server error instead of a 400 response.

PR #2522, opened 2022-02-25
Improve edge-based CH import performance

### LINKED ISSUE #2375: CH preparation becomes very slow when there are large-weight edges

easbar:
This PR improves the node contraction algorithm and parameters for edge-based CH and fixes #2375.

For the car/fastest profile with `u_turn_costs=80` and using a recent OSM extract for Europe it reduced the CH preparation time by around 40% from 7.5h to 4.5h in my test. It slightly reduced the number of shortcuts as well (from 222mio to 209mio). The average query time for random queries in Bavaria was slightly slower (2.6ms vs 2.4ms). With this PR the preparation also takes slightly less memory, because I got rid of one integer array of size `2*edges` (for Europe this saves around `2*95mio*4bytes = 760MB`).

However, the most important improvement here is that edge-based CH now works better for large u-turn costs and large edge weights as they occur when we enable e.g. the `GermanyCountryRule` (this rule sets access=destination for tracks). For example in my test with master it took 221s to complete a car/fastest edge-based CH import for bayern-190101.osm.pbf with `u_turn_costs=0`, 720s for `u_turn_costs=80` and 1280s with `u_turn_costs=200`. With the improvements in this PR this is reduced to 181s (-19%), 440s (-39%) and 516s (-60%), so the import is faster especially for large u-turn costs. 

Including country rules the speed up is even better: For example for germany-190101.osm.pbf the car/fastest/u_turn_costs=40  import took 2733s without country rules in master and 10776s with country rules (almost four times slower). With this PR the import times stays around the same when enabling country rules! This means that including country rules this PR can yield a 10x speedup.

The most important changes leading to these improvements were:

* Do a mini search to first find the bridge paths before running the extra witness searches. This way we can make use of an explicit weight limit for each witness search
* Limit witness searches as much as possible (similar to #2390)
* Use a dynamic limit for the number of polled edges based on an approximated mean degree of the graph as @karussell (:+1:) did for node-based CH. Previously I was setting such a limit based on the settled nodes of previous searches. While this works quite well normally it leads to too long witness searches when there are too many long ones (as it happens for large u-turn costs or large weight edges).
* Use different limits for the number of polled edges during heuristic searches and actual node contraction
* Use neighbor updates rather than periodic updates
* Some heuristic parameter tuning

Related: #1565


karussell:
:rocket:

PR #2523, opened 2022-02-28
Roads as base

michaz:
New profile which uses an unrestricted road network. To be used as a base network for customization.

Experimental.

cc @karussell


core/src/main/java/com/graphhopper/routing/ev/DefaultEncodedValueFactory.java
```diff
@@ -32,6 +32,10 @@ public EncodedValue create(String string) {
 
         if (Roundabout.KEY.equals(name)) {
             enc = Roundabout.create();
+        } else if ("car_access".equals(name)) {
+            enc = new SimpleBooleanEncodedValue("car_access", true);
+        } else if ("bike_access".equals(name)) {
+            enc = new SimpleBooleanEncodedValue("bike_access", true);
```
karussell, line 38: Here it is a bit ugly that we have now this kind of "generic transportation mode" access and additionally the "flag encoder specific" access (`car$access`). Probably the generic access (`xy_access`) should be refined and at least `foot_access` should be added and the `xy$access` should be removed or for now at least removed from being exposed in `/info`.

PR #2524, opened 2022-03-01
Make sure encoder max speed cannot be exceeded, add DecimalEncodedValue#getNextStorableValue

### LINKED ISSUE #2322: Max speed can get exceeded due to rounding error
For example `FootFlagEncoder` has a maximum speed of 15km/h. When we store the speed it is converted to an int using `speedFactor`. However, with our current logic it can happen that we set the speed to `15`, but `get(averageSpeedEnc)` returns e.g. `16`. This way the speed of an edge can exceed the maximum even when we did not exceed the maximum speed when we set the value.

This obviously can be problematic when routing with AStar, but also it can result in a 'max speed violated' error in `CustomWeighting`. See the tests here: https://github.com/graphhopper/graphhopper/commit/944d95ae5e0553e9fa74a758ae2ba17cd6d1da8e

The problematic code is in `UnsignedIntDecimalEncodedValue#toInt`:

```java
    private int toInt(double val) {
        return (int) Math.round(val / factor);
    }
```

In the example above the speed factor is 2, so 15/2=7.5 -> round: 8 -> (int) 8, and when later it is multiplied with the factor again we get 16 > 15. Rounding might be the right thing to do for the decimal encoded value, but in the context of the flag encoder's max speed we somehow have to make sure that the maximum speed is definitely never exceeded. 

Btw currently we can even do something like:
```java
edge.set(encoder.getAverageSpeedEnc(), encoder.getMaxSpeed() * 2);
```
without getting an error. But this is maybe for another issue. The main issue here is that we get back a value that exceeds the max speed even when the value we set was within the valid bounds.

Note that the failing tests are *slightly* artificial, because they deliberately use the `FootFlagEncoder`, because it has an uneven maxSpeed (15) *and* set a non-default speed factor (2 instead of 1).

easbar:
Trying to fix #2322, but see inline comments.


core/src/main/java/com/graphhopper/routing/util/BikeCommonFlagEncoder.java
```diff
@@ -106,6 +106,8 @@ protected BikeCommonFlagEncoder(int speedBits, double speedFactor, int maxTurnCo
         unpavedSurfaceTags.add("wood");
 
         maxPossibleSpeed = 30;
+        // todo: would like to do it like this but avgSpeedEnc is not created yet
+//        maxPossibleSpeed = avgSpeedEnc.getNextStorableValue(30);
```
easbar, line 110: We set the max possible speed before the `avgSpeedEnc` is created so we have no chance of using `avgSpeedEnc` to find a value that works for `maxPossibleSpeed` (depends on `speedFactor`).

karussell: And can we instead move the initialization of maxPossibleSpeed after avgSpeedEnc is created (as it is not final)?

easbar: I think we should not do this. Not initializing avgSpeedEnc in the constructor is bad already. Not initializing the max speed because of this would be even worse. I changed the code so the avgSpeedEnc is initialized in the constructor now.

Issue #2469, opened 2021-11-19
Remove AbstractFlagEncoder#applyWayTags


easbar:
Since we calculate edge flags per edge (no longer per way) since #2457 we should now be able to remove `applyWayTags` from `EncodingManager` and the flag encoders.

easbar:
Probably related to #2470, because we set a few encoded values that depend on others in `applyWayTags`.

Issue #2463, opened 2021-11-17
Remove FlagEncoder


easbar:
This will be a big refactoring task that we will have to split into several smaller issues. We can keep a list of related issues here and also discuss the general direction this should be going here.

* #2462

otbutz:
IMHO #2460

easbar:
> IMHO #2460

Yes, everything related to access and access flags is related. Just not sure if this is required to remove the FlagEncoders (or vice versa).

otbutz:
Removing FlagEncoders requires #2460 to be solved but we can implement it without removing FlagEncoders. At least that is my understanding of it.

The benefit is that users can perform proper truck routing without touching the code :)

easbar:
I think we are getting there :) What we probably should do next is:

* [ ] allow storing the maximum, actually used values for encoded values, so we can replace `encoder.getMaxSpeed()` with `speedEnc.getMaxValue()` or similar. This would be an improvement, because we no longer have to keep the max value synchronized between the vehicle tag parsers and the flag encoders.
* [ ] replace the `FlagEncoder` parameters from the weighting constructors with references to the encoded values they need. For example instead of receiving a `FlagEncoder` and calling `encoder.getAverageSpeedEnc()`,  the `FastestWeighting` should receive the speed encoded value (and encoded values it needs besides this of course). 
* [ ] For `FastestWeighting` we also need to replace `encoder.isMotorVehicle()/HGV()` with some boolean constructor parameters representing the same values (or just a PMap). Maybe we should take these parameters from the weighting string, i.e. something like `weighting: fastest|is_motor_vehicle=true`. Or we skip this intermediate step and replace FastestWeighting (and the others) with CustomWeighting directly.
* [ ] One problem is that we have literally thousands of usages of `GHUtility#setSpeed(..., encoder, ...)` that we will have to replace with something similar that uses the speed/access encoded values directly. I did some unfinished work in this direction some time ago here: 9715c81cef5c2eea66acfb93b72f39342dfd1ff9.

easbar:
Not all the issues we mentioned here are fixed, but flag encoders are almost gone since #2611 :)

The only place they remain now is in `config.yml`/`Graphhopper.java`. For example `graph.flag_encoders=car,bike` does two things now: 

[...]

PR #2527, opened 2022-03-02
add applyWayTags to TagParser

karussell:
When we add the applyWayTags method to the TagParser interface we could move a few EncodedValues like curvature (or even elevation based encoded values) outside of the FlagEncoder. The question is only if we should use the same TagParser interface.


easbar:
We should check if (and why) we even need `encoder.applyWayTags`. In some cases this was only because the edge geometry was unknown in `handleWayTags` before #2457. See #2469.

But otherwise, if it helps for #2463 sure go ahead and add this `applyWaytags` method.

karussell:
Good point. Will check!

> But otherwise, if it helps for https://github.com/graphhopper/graphhopper/issues/2463 sure go ahead and add this applyWaytags method.

Yes, would make it easier. But maybe not necessary as you say.

Issue #2526, opened 2022-03-01
FerrySpeedCalculator refactoring


easbar:
* [ ] remove handling of missing estimated_distance. It will never be missing and if it is it's not our fault.
* [ ] use precise distance, not some estimate, because we can
* [ ] remove maxSpeed capping (should be done elsewhere)
* [ ] remove speedFactor (should be done elsewhere)
* [ ] why do we round the result in *some* (of the too many) cases?
* [ ] clean up logic (do not set durationInHours=0 for control flow etc.)

easbar:
> remove handling of missing estimated_distance. It will never be missing and if it is it's not our fault.

Or what was the reasoning behind this?

> use precise distance, not some estimate, because we can

Currently we use the distance between the tower nodes. At some point we used the distance between the first and last *way*. I'm not sure if this was ever considered, but what if the way went beyond the map borders? It could happen that a way is only included partially, so the distance would be changed, but e.g. the duration tag would remain the same which could lead to wrong speeds.

Also why do we sometimes return speedFactor/2 or unknownSpeed depending on the `distance <= 300`? 

https://github.com/graphhopper/graphhopper/blob/c5ea4702baed57b705e757447fddf8d02532bf24/core/src/test/java/com/graphhopper/routing/util/FerrySpeedCalculatorTest.java#L42-L46

https://github.com/graphhopper/graphhopper/blob/c5ea4702baed57b705e757447fddf8d02532bf24/core/src/test/java/com/graphhopper/routing/util/FerrySpeedCalculatorTest.java#L56-L58

easbar:
> Currently we use the distance between the tower nodes. At some point we used the distance between the first and last way point. I'm not sure if this was ever considered, but what if the way went beyond the map borders? It could happen that a way is only included partially, so the distance would be changed, but e.g. the duration tag would remain the same which could lead to wrong speeds.

This is a problem actually. First of all, the `duration` tag refers to the entire way, but in some cases ferry ways are split into several edges (e.g. when different ferry ways share the same nodes, which happens sometimes close to a harbor). Using the duration tag to calculate the speed of each edge is incorrect in this case. 

The case of a missing `estimated_distance` occurred when the first/last node of a ferry way was missing: https://github.com/graphhopper/graphhopper/pull/2528#discussion_r821405075

> Also why do we sometimes return speedFactor/2 and sometimes unknownSpeed depending on distance <= 300?

[...]

PR #2528, opened 2022-03-02
Clean up FerrySpeedCalculator, use exact edge distance

easbar:
Work for #2526.

The `estimated_distance` tag is now called `beeline_distance` and represents the straight line distance between the tower nodes of the edge, not the straight line distance between the first and last nodes of the OSM way as `estimated_distance` did (at some point). We can also use this for the bendiness calculation without `applyWayTags` (see #2469). The actual length of the road/edge is now stored in an artificial tag `road_distance` and we get the full point list of the edge by `point_list`.

More comments inline.

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -394,8 +394,9 @@ public class OSMReader {
         }
 
         double speedInKmPerHour = distance / 1000 / (durationInSeconds / 60.0 / 60.0);
-        if (speedInKmPerHour < 0.01d) {
-            LOGGER.warn("Unrealistic low speed calculated from duration. Maybe duration is too long for OSM way: "
+        if (speedInKmPerHour < 0.1d) {
+            // often there are mapping errors like duration=30:00 (30h) instead of duration=00:30 (30min)
+            LOGGER.warn("Unrealistic low speed calculated from duration. Maybe the duration is too long, or it is applied to a way that only represents a part of the connection? OSM way: "
                     + way.getId() + ". duration=" + durationTag + " (= " + Math.round(durationInSeconds / 60.0) +
                     " + minutes), distance=" + Math.round(distance / 1000.0) + "km");
             return;
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -199,7 +199,7 @@ public class OSMReader {
         double firstLat = pointList.getLat(0), firstLon = pointList.getLon(0);
         double lastLat = pointList.getLat(pointList.size() - 1), lastLon = pointList.getLon(pointList.size() - 1);
         way.setTag("beeline_distance", distCalc.calcDist(firstLat, firstLon, lastLat, lastLon));
-        way.setTag("total_distance", distance);
+        way.setTag("road_distance", distance);
         way.setTag("point_list", pointList);
 
         // we have to remove existing artificial tags, because we modify the way even though there can be multiple edges
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -352,6 +352,12 @@ public class OSMReader {
                     + ", difference: " + (edgeDistance - geometryDistance));
     }
 
+    /**
+     * This method is called for each way during the second pass and before the way is split into edges.
+     * We currently use it to calculate the distance of a way and determine the speed based on the duration tag when
+     * it is present. This cannot be done on a per-edge basis, because the duration tag refers to the duration of the
+     * entire way.
+     */
     protected void preprocessWay(ReaderWay way, WaySegmentParser.CoordinateSupplier coordinateSupplier) {
         if (!isCalculateWayDistance(way))
             return;
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -395,7 +395,7 @@ public class OSMReader {
 
         double speedInKmPerHour = distance / 1000 / (durationInSeconds / 60.0 / 60.0);
         if (speedInKmPerHour < 0.1d) {
-            // often there are mapping errors like duration=30:00 (30h) instead of duration=00:30 (30min). If there are none anymore, maybe make this check more strict.
+            // often there are mapping errors like duration=30:00 (30h) instead of duration=00:30 (30min). If there are none anymore, maybe raise the limit to find more cases.
             LOGGER.warn("Unrealistic low speed calculated from duration. Maybe the duration is too long, or it is applied to a way that only represents a part of the connection? OSM way: "
                     + way.getId() + ". duration=" + durationTag + " (= " + Math.round(durationInSeconds / 60.0) +
                     " + minutes), distance=" + Math.round(distance / 1000.0) + "km");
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
```diff
@@ -8,65 +8,58 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(FerrySpeedCalculator.class);
     private final double speedFactor;
-    private final double unknownSpeed, longSpeed, shortSpeed, maxSpeed;
+    private final double unknownSpeed, maxSpeed;
 
-    public FerrySpeedCalculator(double speedFactor, double maxSpeed, double longSpeed, double shortSpeed, double unknownSpeed) {
+    public FerrySpeedCalculator(double speedFactor, double maxSpeed, double unknownSpeed) {
```
easbar, line 13: `longSpeed` and `shortSpeed` were never used, because the distance was never missing (see below). At least when this code was used with our `OSMReader`. Also note that `longSpeed` and `shortSpeed` made no sense anyway for some encoders, because they exceeded `maxSpeed` (and this was not accounted for here).

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -195,17 +195,16 @@ protected boolean isBarrierNode(ReaderNode node) {
      * This method is called during the second pass of {@link WaySegmentParser} and provides an entry point to enrich
      * the given OSM way with additional tags before it is passed on to the tag parsers.
      */
-    protected void setArtificialWayTags(PointList pointList, ReaderWay way) {
-        // Estimate length of ways containing a route tag e.g. for ferry speed calculation
+    protected void setArtificialWayTags(PointList pointList, ReaderWay way, double distance) {
         double firstLat = pointList.getLat(0), firstLon = pointList.getLon(0);
         double lastLat = pointList.getLat(pointList.size() - 1), lastLon = pointList.getLon(pointList.size() - 1);
+        way.setTag("beeline_distance", distCalc.calcDist(firstLat, firstLon, lastLat, lastLon));
+        way.setTag("road_distance", distance);
```
easbar, line 202: Or should we call it `road_length`?

core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
```diff
@@ -8,65 +8,58 @@
...
+        double distanceInKm = way.getTag("road_distance", -1.0) / 1000;
+        if (distanceInKm < 0)
+            throw new IllegalStateException("The artificial 'road_distance' tag is missing for way: " + way.getId());
         Long duration = way.getTag("duration:seconds", 0L);
-        // seconds to hours
         double durationInHours = duration / 60d / 60d;
-        // Check if our graphhopper specific artificially created estimated_distance way tag is present
-        Number estimatedLength = way.getTag("estimated_distance", null);
         if (durationInHours > 0) {
-            if (estimatedLength != null) {
```
easbar, line 32: `estimatedLength != null` can never happen.

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
karussell: road_distance is fine for me (or edge_distance?)

Do we need the separate beeline_distance only for curvature? Then we could move the calculation there and avoid it here.

karussell (approved):
Thanks for this cleanup :)

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: First I thought `edge_distance` is not a good choice, because it sounded like the distance/length of the (straight line 'edge'/connection between the tower nodes). But actually it's consistent to call it like this, because it is equal to `edge.getDistance()`.

> Do we need the separate beeline_distance only for curvature? Then we could move the calculation there and avoid it here.

:+1:

easbar:
Thanks for taking a look. I think we can merge here, but this is probably not finished yet. I still don't like how the `FerrySpeedCalculator` rounds the speed and caps the speed to min/max speed, because this is not part of the speed modelling/calculation, but only something that needs to be done because of our limited speed storage. We should have this in the same place where we do similar stuff (see the different `setSpeed` methods of the encoders. I know the reason this is here is that `FerrySpeedCalculator` was inside `AbstractFlagEncoder` at some point.

core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
```diff
@@ -7,66 +7,55 @@
...
+            return minSpeed;
+        else
             // unknown speed -> put penalty on ferry transport
             return unknownSpeed;
-        } else if (durationInHours > 1) {
-            // lengthy ferries should be faster than short trip ferry
-            return longSpeed;
-        } else {
-            return shortSpeed;
-        }
```
easbar, line 70: > lengthy ferries should be faster than short trip ferry

I still think this is something we might have to do. If the duration tag is missing we probably *should* apply a faster speed for long ferries than for short ones. Especially because the `unknownSpeed` is very slow (5km/h). There is probably no ferry that crosses the sea at 5km/h. But so far we only applied this long/shortSpeed when the distance was missing (which never happened anyway), *not* when the duration was missing, or at least we only distinguished between "very slow/`unknownSpeed` and "as slow as possible"/`speedFactor / 2` in this case. 

We probably need to figure out how often the duration is missing and if it is sometimes missing also for longer ferries. I can imagine that mappers skip the duration tag for short ferries that just cross a river, but for longer distance ferries this seems like a mapping error and we could even print a warning, report this to OSM and/or skip the ferry in this case. I opened #2532 for this.

easbar:
There is a problem still: Some ferry ways share nodes with each other, for example 'Burgeo - Ramea' and 'Grey River - Ramea' here: https://www.openstreetmap.org/way/168781977. This means that with this PR in it's current state we would calculate the ferry speed for some edges that only represent a part of the ferry way, but still using the duration tag of the entire way.

core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
easbar: I only just understood this now: Before #2448 `estimatedLength == null` **could** happen when the first/last OSM node of a ferry way was missing:

https://github.com/graphhopper/graphhopper/blob/f95397b7dbcb186de7cafb735af8275abd6122d0/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java#L206-L212

And the reason we have to consider the distance of the entire way is that 1) the duration tag refers to the duration of the entire way and 2) ferry ways sometimes intersect and are thus split into several edges for which we cannot directly use the duration tag.

core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
easbar: > We probably need to figure out how often the duration is missing and if yes if it is sometimes missing also for longer ferries.

It is in fact missing in very many cases and also for longer ferries like this one: https://www.openstreetmap.org/way/999670971. Simply falling back to the `unknownSpeed=5km/h` is thus not a good idea. Here is the [route](https://graphhopper.com/maps/?point=35.78969%2C-5.804558&point=36.352186%2C-3.393402&point=36.835668%2C-1.538086&point=37.688167%2C0.626221&point=38.225235%2C1.966553&point=38.766933%2C3.378296&point=42.747012%2C7.393799&point=43.411033%2C7.959595&point=43.886016%2C8.363342&point=44.262558%2C8.449516&locale=en-US&elevation=true&profile=car&use_miles=false&selected_detail=Elevation&layer=OpenStreetMap). The average speed is 6km/h (unknownSpeed=5, but rounded to 6 because of speedFactor=2). It takes more than eleven days when really this ferry only needs around two (35km/h).

PR #2533, opened 2022-03-06
Move handleNodeTags from EncodingManager to AbstractFlagEncoder

easbar:
Currently we do:

```
encoder.handleWayTags()
encodingManager.handleNodeTags() // sets access to false if encoder.isBarrier() == true)
encoder.applyWayTags()
```

In #2469 `applyWayTags` should go into `handleWayTags` so right now this would mean it would run before `handleNodeTags`. But first I think we should move `handleNodeTags` to the encoder (this PR), because the encoders should handle the access flags directly (not indirectly via isBarrier as we do now). I therefore moved the code that blocks barriers into a new `handleNodeTags` method in `AbstractFlagEncoder` and this method is called in `encodingManager.handleWayTags` now.

Later I'd like to go further and push this code down into the single flag encoders such that everything related to the access flags will be in one place per encoder (and we can make this just a TagParser for access) at some point.

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -195,7 +195,8 @@ public class OSMReader {
      * This method is called during the second pass of {@link WaySegmentParser} and provides an entry point to enrich
      * the given OSM way with additional tags before it is passed on to the tag parsers.
      */
-    protected void setArtificialWayTags(PointList pointList, ReaderWay way) {
+    protected void setArtificialWayTags(PointList pointList, ReaderWay way, Map<String, Object> nodeTags) {
+        way.setTag("node_tags", nodeTags);
         // Estimate length of ways containing a route tag e.g. for ferry speed calculation
         double firstLat = pointList.getLat(0), firstLon = pointList.getLon(0);
         double lastLat = pointList.getLat(pointList.size() - 1), lastLon = pointList.getLon(pointList.size() - 1);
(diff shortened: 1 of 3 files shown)


easbar:
Btw for #2469 we can probably also just change the order of `handleNodeTags` and `applyWayTags`, because (at least in our code) `applyWayTags` uses the `accessEnc`, but since the node tags are only used for barrier nodes it won't affect the elevation-related things we do in `applyWayTags`.

karussell:
:+1: 

Should we handle way and node tags in a single method (my preference too) or use e.g. a separate NodeTagParser with a separate method?

But as we use EncodedValues per edge (and have no intention anymore to introduce them for nodes) we should probably convert all tags into them using a single method only.

easbar:
> Should we handle way and node tags in a single method (my preference too) or use e.g. a separate NodeTagParser with a separate method?

I think the node tags along a way (like railway crossings) are just another type of information about a way that could be considered when writing the (edge) encoded values. So we would just pass this information to `handleWayTags`. Or, like we currently do for `barrier`, we create an artificial edge and then the node tags are again just a property of this edge again.

The only other case I can think of are node tags that shall be interpreted as turn costs (like traffic lights at junctions), but this will be treated differently and is for another issue.

> But as we use EncodedValues per edge (and have no intention anymore to introduce them for nodes) we should probably convert all tags into them using a single method only.

Yes, I don't think something like encoded values for nodes is needed, and it would not be considered by the routing algorithms anyway. Like I said, the only thing we could do is using node tags to set turn costs.

easbar:
Also at some point we should probably separate our internal/artificial way tags from the actual OSM tags, c.f. this discussion: https://github.com/graphhopper/graphhopper/pull/2457/files#r746357392

PR #2536, opened 2022-03-11
Use Weighting instead of access for turn instructions

karussell:
One more change for #2463. 

 - [x] show in test that this also improves the instructions for custom profiles (where access from the base-FlagEncoder is different)
 - [x] investigate why routingCH.mean is 20% slower

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/InstructionsOutgoingEdges.java
+++ b/core/src/main/java/com/graphhopper/routing/InstructionsOutgoingEdges.java
@@ -74,7 +74,7 @@ class InstructionsOutgoingEdges {
                                      DecimalEncodedValue maxSpeedEnc,
                                      EnumEncodedValue<RoadClass> roadClassEnc,
                                      BooleanEncodedValue roadClassLinkEnc,
-                                     EdgeExplorer crossingExplorer,
+                                     EdgeExplorer allExplorer,
                                      NodeAccess nodeAccess,
                                      int prevNode,
                                      int baseNode,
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/graphhopper/routing/InstructionsOutgoingEdges.java
```diff
@@ -129,22 +128,22 @@ public boolean outgoingEdgesAreSlowerByFactor(double factor) {
...
             tmpSpeed = getSpeed(edge);
             if (tmpSpeed > maxSurroundingSpeed) {
                 maxSurroundingSpeed = tmpSpeed;
             }
         }
 
-        // Surrounding streets need to be slower by a factor
-        return maxSurroundingSpeed * factor < pathSpeed;
+        // surrounding streets need to be slower by a factor and call round() so that tiny differences are ignored
+        return Math.round(maxSurroundingSpeed * factor) < Math.round(pathSpeed);
```
karussell, line 146: These changes are the most significant. The problem with the loop over visibleAlternativeTurns was that it partially reads the speeds from edges with access=false which is possible when reading from an EncodedValue but not when calling calcWeight has we have a special protection there (i.e. it throws an exception).

The follow up changes regarding the round() call were then necessary to avoid (more or less) instructions just because of a tiny speed difference (<0.1km/h).

easbar:
Not sure if this is comparable, but we have some other code that also needs to (only) distinguish between finite/infinite weight edges here: https://github.com/graphhopper/graphhopper/blob/75d500006f663dfc1b23f25b43d830ba8572f8ca/core/src/main/java/com/graphhopper/storage/RoutingCHEdgeIteratorStateImpl.java#L130-L147

Regarding the instructions my guess would be that the quality of the instructions is more important than the query speed, especially because faster queries can be achieved by disabling instructions.

easbar:
Thanks!

PR #2537, opened 2022-03-12
Move way name parsing from EncodingManager to OSMReader

easbar:
I moved the code that parses the `name/ref` OSM tag of the OSM ways from `EncodingManager#applyWayTags` to `OSMReader`. I also added a new setting to disable instructions on the server side called `routing.instructions`, which is now separated from `datareader.instructions`. The latter now only controls whether the way names shall be parsed or not. It is probably debatable whether we even need `datareader.instructions` and it should probably be better renamed to `datareader.parse_names` or even `osmreader.parse_names`. However renaming `datareader` to `osmreader` is for another issue.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
+++ b/core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
@@ -30,7 +30,6 @@ import com.graphhopper.util.EdgeIteratorState;
 import com.graphhopper.util.PMap;
 
 import java.util.*;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static com.graphhopper.util.Helper.toLowerCase;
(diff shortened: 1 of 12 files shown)


core/src/test/java/com/graphhopper/reader/osm/GraphHopperOSMTest.java
```diff
@@ -388,17 +388,6 @@ public void testSortedGraph_noCH() {
...
-        GHRequest req = new GHRequest(51.2492152, 9.4317166, 51.2, 9.4);
-        req.setProfile(profile);
-        boolean old = instance.getEncodingManager().isEnableInstructions();
-        req.putHint("instructions", true);
-        instance.route(req);
-        assertEquals(old, instance.getEncodingManager().isEnableInstructions());
-
-        req.putHint("instructions", false);
-        instance.route(req);
-        assertEquals(old, instance.getEncodingManager().isEnableInstructions(), "route method should not change instance field");
```
easbar, line 401: Definitely not :) But testing this here does not seem necessary and now the flag is gone anyway...

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -591,44 +539,18 @@ public boolean equals(Object o) {
...
-                    name = refName;
-                else
-                    name += ", " + refName;
-            }
-
-            edge.setName(name);
-        }
-
-        if (Double.isInfinite(edge.getDistance()))
-            throw new IllegalStateException("Infinite distance should not happen due to #435. way ID=" + way.getId());
```
easbar, line 628: We do this check else where

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -46,17 +45,14 @@
...
     private final List<TagParser> edgeTagParsers = new ArrayList<>();
     private final Map<String, TurnCostParser> turnCostParsers = new LinkedHashMap<>();
-    private boolean enableInstructions = true;
-    private String preferredLanguage = "";
-    private EncodedValue.InitializerConfig turnCostConfig;
-    private EncodedValue.InitializerConfig relationConfig;
-    private EncodedValue.InitializerConfig edgeConfig;
+    private final EncodedValue.InitializerConfig turnCostConfig;
+    private final EncodedValue.InitializerConfig relationConfig;
+    private final EncodedValue.InitializerConfig edgeConfig;
```
easbar, line 55: These can be final so I made them final

PR #2538, opened 2022-03-14
Cleanup isBarrier method, fix ford=yes and locked=yes

easbar:
It seems a lot more readable when we check the access restriction tags first, because they have a higher priority anyway. For example when there is a tag like `bicycle=no/yes` it does no longer matter what kind of barrier there is. Same with `locked=yes`. It should be a barrier even when the barrier is among the `passByDefaultBarriers`. I actually removed `passByDefaultBarriers` entirely. There was a comment saying: `// barrier which may be explicitly forbidden, otherwise it is allowed`, but that just applies to *all* barriers except those in `blockByDefaultBarriers`?!

Also I noticed that I broke the `blockFords` flag in #2430. Currently we only consider nodes to be barriers when they [actually have the barrier tag](https://github.com/graphhopper/graphhopper/blob/75d500006f663dfc1b23f25b43d830ba8572f8ca/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java#L189-L195). I fixed this here as well. Maybe we should remove the handling of `highway=ford` nodes though as it is only used 36 times in total. `ford=yes` means the same (?) and is used >800.000 times.


karussell:
Nice :)

> Maybe we should remove the handling of highway=ford nodes though as it is only used 36 times in total. ford=yes means the same (?) and is used >800.000 times.

+1 from me for removing if it is used <50 times world wide.

easbar:
`highway=ford` is also deprecated on the wiki page: https://wiki.openstreetmap.org/wiki/Tag:ford%3Dyes
It is supposed to be replaced by `ford=yes`. Actually the few occurrences also seem to be in the same area (Austria). I'll flag a few of them.

PR #2542, opened 2022-03-21
Load encoded values from disk

easbar:
I tried using Jackson to serialize/deserialize an `EnumEncodedValue`.

### DIFF
--- a/core/src/test/java/com/graphhopper/routing/ev/EnumEncodedValueTest.java
+++ b/core/src/test/java/com/graphhopper/routing/ev/EnumEncodedValueTest.java
@@ -1,5 +1,9 @@
 package com.graphhopper.routing.ev;
 
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.PropertyAccessor;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.graphhopper.storage.IntsRef;
 import org.junit.jupiter.api.Test;
 
@@ -32,4 +36,19 @@ public class EnumEncodedValueTest {
         assertEquals(4, 32 - Integer.numberOfLeadingZeros(16 - 1));
         assertEquals(5, 32 - Integer.numberOfLeadingZeros(17 - 1));
     }
+
+    @Test
+    public void serializationAndDeserialization() throws JsonProcessingException {
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.setVisibility(PropertyAccessor.ALL, JsonAutoDetect.Visibility.NONE);
+        mapper.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
+
+        EnumEncodedValue<RoadClass> roadClass = new EnumEncodedValue<>(RoadClass.KEY, RoadClass.class);
+        String roadClassString = mapper.writeValueAsString(roadClass);
+        System.out.println(roadClassString);
+        EnumEncodedValue encodedValue = mapper.readValue(roadClassString, EnumEncodedValue.class);
+        EnumEncodedValue<RoadClass> roadClass2 = (EnumEncodedValue) encodedValue;
+        System.out.println(roadClass2);
+    }
+
 }
--- a/core/src/main/java/com/graphhopper/routing/ev/EnumEncodedValue.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/EnumEncodedValue.java
@@ -17,6 +17,9 @@
  */
 package com.graphhopper.routing.ev;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.util.StdConverter;
 import com.graphhopper.storage.IntsRef;
 
 import java.util.Arrays;
@@ -24,8 +27,15 @@ import java.util.Arrays;
 /**
  * This class allows to store distinct values via an enum. I.e. it stores just the indices
  */
+@JsonDeserialize(converter = EnumEncodedValue.EnumConstantsSetter.class)
 public final class EnumEncodedValue<E extends Enum> extends IntEncodedValueImpl {
-    private final E[] arr;
+    public Class<E> enumType;
+    @JsonIgnore
+    public E[] arr;
+
+    EnumEncodedValue() {
+        // for jackson
+    }
 
     public EnumEncodedValue(String name, Class<E> enumType) {
         this(name, enumType, false);
@@ -33,6 +43,7 @@ public final class EnumEncodedValue<E extends Enum> extends IntEncodedValueImpl 
 
     public EnumEncodedValue(String name, Class<E> enumType, boolean storeTwoDirections) {
         super(name, 32 - Integer.numberOfLeadingZeros(enumType.getEnumConstants().length - 1), storeTwoDirections);
+        this.enumType = enumType;
         arr = enumType.getEnumConstants();
     }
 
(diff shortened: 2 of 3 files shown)


core/src/test/java/com/graphhopper/routing/ev/EnumEncodedValueTest.java
```diff
@@ -32,4 +40,36 @@ public void testSize() {
...
+            if (jsonEV.has("enumType"))
+                deserializedEVs.add(mapper.treeToValue(jsonEV, EnumEncodedValue.class));
+            else if (jsonEV.has("factor"))
+                deserializedEVs.add(mapper.treeToValue(jsonEV, DecimalEncodedValueImpl.class));
+            else if (jsonEV.has("type") && "simple_boolean".equals(jsonEV.get("type").asText()))
+                deserializedEVs.add(mapper.treeToValue(jsonEV, SimpleBooleanEncodedValue.class));
+            else
+                deserializedEVs.add(mapper.treeToValue(jsonEV, IntEncodedValueImpl.class));
+        }
+        System.out.println(deserializedEVs);
```
easbar, line 72: Here I serialized and deserialized a list of EncodedValues with different types. It's a rather ugly, because we need to recognize the type of the encoded value for deserialization somehow (e.g. I added an extra field to `SimpleBooleanEncodedValue` just for this...). 

Putting everything into a `List<EncodedValue>` is ok, because in `EncodedValueLookup` we get the type information from the client code like `lookup.getEnumEncodedValue(RoadClass.KEY, RoadClass.class)`.

core/src/test/java/com/graphhopper/routing/ev/EnumEncodedValueTest.java
```diff
@@ -32,4 +41,44 @@ public void testSize() {
...
+            String s = mapper.writeValueAsString(e);
+            serializedEVs.add(s);
+        }
+
+        // deserialize
+        List<EncodedValue> deserializedEVs = new ArrayList<>();
+        for (String s : serializedEVs) {
+            JsonNode jsonNode = mapper.readTree(s);
+            deserializedEVs.add(mapper.treeToValue(jsonNode, EncodedValue.class));
+        }
```
easbar, line 70: This is pretty nice and seems to work for all our `EncodedValue` subclasses automatically, due to the `@JsonTypeInfo` annotation that I added to the `EncodedValue` interface.

core/src/main/java/com/graphhopper/routing/ev/EncodedValueSerializer.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public static EncodedValue deserializeEncodedValue(String serializedEncodedValue) {
+        try {
+            JsonNode jsonNode = MAPPER.readTree(serializedEncodedValue);
+            int storedVersion = jsonNode.get("version").asInt();
+            ((ObjectNode) jsonNode).remove("version");
+            EncodedValue encodedValue = MAPPER.treeToValue(jsonNode, EncodedValue.class);
+            if (storedVersion != encodedValue.getVersion())
+                throw new IllegalStateException("Version does not match. Cannot properly read encoded value: " + encodedValue.getName() + ". " +
+                        "You need to use the same version of GraphHopper you used to import the data");
```
easbar, line 54: This is a bit self-made and probably can also be done using Jackson Annotations, but since we aren't really planning to de/serialize encoded values with something other than these two methods I think this is fine. I'm not so experienced with Jackson and open for improvements of course.

core/src/main/java/com/graphhopper/routing/ev/StringEncodedValue.java
```diff
@@ -30,7 +27,7 @@ public StringEncodedValue(String name, int expectedValueCount, boolean storeTwoD
 
         this.maxValues = roundUp(expectedValueCount);
         this.values = new ArrayList<>(maxValues);
-        this.indexMap = new ObjectIntHashMap<>(maxValues);
+        this.indexMap = new HashMap<>(maxValues);
```
otbutz, line 30: Could we avoid this with a dedicated getter instead of using field access? I'm not sure if we want to deal with auto(un)boxing in a potentially hot code path.

otbutz (changes requested):

core/src/main/java/com/graphhopper/routing/ev/StringEncodedValue.java
easbar: Is this really about field access. I thought we need a custom serializer for hppc, but I might be wrong. I thought about the auto unboxing for a moment but then thought it's probably not as bad bc the maps will be very small. But you are more than welcome to make this work with the hppc map again. I just did not feel like spending a lot more time on this, also because currently the default GraphHopper server does not use StringEncodedValue currently.

easbar: I tried changing `indexMap` to the concrete type `ObjectIntHashMap` and got another error which is due to the fact that `HashOrderMixingStrategy` is an abstract type and Jackson does not know which type to use upon de-serialization.

easbar:
I went for the all-args constructor approach where we can keep the fields final but have to add some all-args constructors for Jackson. 

@otbutz maybe this is related to #2243?

core/src/main/java/com/graphhopper/routing/ev/StringEncodedValue.java
easbar: It works when we add `jackson-datatype-hppc`: https://github.com/FasterXML/jackson-datatypes-collections
I tried this here including a little speed test: fa80be946c6620290c680a72f41748ff4de0fd3b
I could not measure a notable difference, though so I would go with java.util for now.

otbutz:
Yes this would also solve #2243 :wink:

Does Jackson require those constructors to be public? Maybe we could reduce the visibility if they're only intended to be used by Jackson.

easbar:
> Does Jackson require those constructors to be public? Maybe we could reduce the visibility if they're only intended to be used by Jackson.
 
 Yes, package private also works and is better thanks. We could also make them protected, but not sure if realistically users want to create their own subclasses of `IntEncodedValueImpl` etc.

core/src/main/java/com/graphhopper/routing/ev/EncodedValueSerializer.java
karussell: I'm also no expert. I tried to get it working and embedded your work into a common Deserializer (and using `@JsonProperty(access = JsonProperty.Access.READ_ONLY)` for getVersion to avoid a custom Serializer).

But the problem was that we don't want to manually handle the className and I wasn't able to find a way to use both: 
`@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "className")` 
and
`@JsonDeserialize(using = EncodedValueDeserializer.class)`.

The only solution was with some wrapper class but this here looks simpler IMO.

core/src/main/java/com/graphhopper/routing/ev/EncodedValueSerializer.java
```diff
@@ -0,0 +1,60 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+public class EncodedValueSerializer {
+    private final static ObjectMapper MAPPER = new ObjectMapper();
+
+    static {
+        MAPPER.setVisibility(PropertyAccessor.ALL, JsonAutoDetect.Visibility.NONE);
+        MAPPER.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
```
karussell, line 33: for consistency shall we add 
MAPPER.setPropertyNamingStrategy(PropertyNamingStrategy.SNAKE_CASE);
here?

karussell (approved):
:+1:

PR #2544, opened 2022-03-22
allow skipping the LM constraint check for A*

karussell:


core/src/main/java/com/graphhopper/routing/DefaultWeightingFactory.java
```diff
@@ -75,7 +75,7 @@ public Weighting createWeighting(Profile profile, PMap requestHints, boolean dis
                 throw new IllegalArgumentException("custom weighting requires a CustomProfile but was profile=" + profile.getName());
             CustomModel queryCustomModel = requestHints.getObject(CustomModel.KEY, null);
             CustomProfile customProfile = (CustomProfile) profile;
-            if (queryCustomModel != null)
+            if (queryCustomModel != null && !requestHints.getBool("lm.disable", false))
```
karussell, line 78: The problem is that if LM is activated on the server the client would still need the lm.disable=true parameter. Will see if I can detect this with LMProfile.

core/src/main/java/com/graphhopper/routing/Router.java
```diff
@@ -186,9 +186,9 @@ private void checkNoBlockAreaWithCustomModel(GHRequest request) {
     protected Solver createSolver(GHRequest request) {
         final boolean disableCH = getDisableCH(request.getHints());
         final boolean disableLM = getDisableLM(request.getHints());
-        if (chEnabled && !disableCH) {
+        if (chEnabled && !disableCH && chGraphs.containsKey(request.getProfile())) {
             return new CHSolver(request, profilesByName, routerConfig, encodingManager, chGraphs);
-        } else if (lmEnabled && !disableLM) {
+        } else if (lmEnabled && !disableLM && landmarks.containsKey(request.getProfile())) {
```
karussell, line 191: Maybe we can remove chEnabled now? (and lmEnabled)

core/src/main/java/com/graphhopper/routing/Router.java
```diff
@@ -561,6 +561,9 @@ protected FlexiblePathCalculator createPathCalculator(QueryGraph queryGraph) {
                 throw new IllegalArgumentException("Cannot find LM preparation for the requested profile: '" + profile.getName() + "'" +
                         "\nYou can try disabling LM using " + Parameters.Landmark.DISABLE + "=true" +
                         "\navailable LM profiles: " + landmarks.keySet());
+            if (profile instanceof CustomProfile && request.getCustomModel() != null
+                    && !request.getHints().getBool("lm.disable", false))
```
easbar, line 565: The `lm.disable` check is not needed here right? It is already done before when we select the `LMSolver`.

PR #2550, opened 2022-04-05
No longer allow changing profiles between imports and loads

easbar:
We do not allow changing profiles for which there are CH or LM preparations, because this would invalidate the index data we stored for these preparations. However, changing profiles without preparations is also potentially wrong, because of the subnetwork encoded values we calculate for each profile during the import. See: https://github.com/graphhopper/graphhopper/issues/2025#issuecomment-1088004807

Adding or removing profiles is not possible already in master, because this changes the (subnetwork) encoded value setup and we get an error in `GraphHopperStorage#loadExisting` when we try to load. With this PR also changing the profiles yields an error.


core/src/test/java/com/graphhopper/GraphHopperTest.java
```diff
@@ -2477,42 +2477,40 @@ void averageSpeedPathDetailBug() {
...
+        GraphHopper hopper = new GraphHopper()
+                .setProfiles(
+                        new Profile("p_car").setVehicle("car").setWeighting("fastest"),
+                        new Profile("p_bike").setVehicle("bike").setWeighting("fastest")
+                )
+                .setGraphHopperLocation(GH_LOCATION)
+                .setOSMFile(BAYREUTH);
+        hopper.importOrLoad();
+        int nodes = hopper.getGraphHopperStorage().getNodes();
+        hopper.close();
```
easbar, line 2489: This test was testing an API I do not even think we should support: Creating a graph using a low-level API, but then loading it using the high-level GraphHopper class. For example this way we do not even create subnetwork encoded values. I now use the GraphHopper class to create the graph.

core/src/test/java/com/graphhopper/storage/GraphHopperStorageLMTest.java
```diff
@@ -1,73 +0,0 @@
...
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-
-public class GraphHopperStorageLMTest {
-    @Test
-    public void testLoad() {
-        String defaultGraphLoc = "./target/ghstorage_lm";
-        Helper.removeDir(new File(defaultGraphLoc));
-        CarFlagEncoder carFlagEncoder = new CarFlagEncoder();
-        EncodingManager encodingManager = new EncodingManager.Builder().add(carFlagEncoder).add(Subnetwork.create("my_profile")).build();
-        GraphHopperStorage graph = GraphBuilder.start(encodingManager).setRAM(defaultGraphLoc, true).create();
```
easbar, line 27: Also here we use a combination of low- and high-level API. This test does no longer test anything useful so I deleted the entire class.

PR #2551, opened 2022-04-05
Remove EncodedValueFactory

### LINKED ISSUE #2438: Why do we need EncodedValueFactory?
We currently create `EncodedValue` objects in **two** places: The `TagParser`s and in `DefaultEncodedValueFactory`. The reason is probably that we have two scenarios: 

1) We generate new graph data from an OSM file. For this we define `TagParser`s that handle different OSM tags and write to according encoded values. The `TagParser`s also create these `EncodedValue`s, e.g. `OSMMaxSpeedParser` creates the `MaxSpeed` encoded value that it writes to.

2) We load existing graph data. In this case we don't need `TagParser`s, because we are not parsing any tags. We still have to create the `EncodedValue` objects though. This is where `EncodedValueFactory` comes in. We use it to create the `EncodedValue` objects given the strings that were written to the properties file during the import here: https://github.com/graphhopper/graphhopper/blob/14bd13a78bd494173109016b58c05759d9cc680d/core/src/main/java/com/graphhopper/GraphHopper.java#L735

The obvious downside of `EncodedValues` being created in not one, but two places is that they have to be kept in sync. And, to no surprise, they currently aren't.

This test fails:

```java
    @Test
    public void lanesEV() {
        {
            GraphHopperConfig config = new GraphHopperConfig();
            config.putObject("graph.location", GH_LOCATION);
            config.putObject("datareader.file", BAYREUTH);
            config.putObject("graph.encoded_values", "lanes");
            config.setProfiles(Collections.singletonList(
                    new Profile("car").setVehicle("car").setWeighting("fastest")
            ));
            GraphHopper hopper = new GraphHopper();
            hopper.init(config);
            hopper.importOrLoad();
        }
        {
            GraphHopper hopper = new GraphHopper()
                    .setGraphHopperLocation(GH_LOCATION)
                    .setProfiles(new Profile("car").setVehicle("car").setWeighting("fastest"));
            hopper.importOrLoad();
            // error: java.lang.IllegalArgumentException: DefaultEncodedValueFactory cannot find EncodedValue lanes
        }
    }
```

It fails, because we have an `OSMLanesParser` that is listed in `DefaultTagParserFactory`, but the `Lanes` EV is not listed in `DefaultEncodedValueFactory`. But here is the catch: The test only fails, because we use  the `GraphHopper`'s 'setter' initialization when we load the graph data. If we used `init` instead it would work, because in this case, even though there is nothing to import anymore, we **do** create the `TagParsers`, which then create the encoded values (and not the `EncodedValueFactory`). Only when we skip the `init()` call `encodingManager` will be `null` in the `load` method and we try to create the encoded values from the `graph.encoded_values` string that was written to the **properties file**. 

So we only ever use `EncodedValueFactory` when we load existing graph data **and** initialize the `GraphHopper` class using the setters (do not call `init()`). 

We have a few options:

1) Just remember that we have to keep the `EncodedValueFactory` in sync with the `TagParserFactory` in case we load `GraphHopper` without calling `init()`. This is what we do currently. Really ugly (see above).

2) Remove `EncodedValueFactory` and only create encoded values via `TagParser`s. Maybe a bit better, because `EncodedValueFactory` is already only used in one specific case and it is already broken/out of sync. Still ugly though, because this means we need to create tag parser objects also when loading an existing graph. But 'parsing tags' is something that has nothing to do with 'loading existing graph data'. The fact that `GraphHopper` (as well as `GraphHopperStorage`, `DataAccess`, `FlagEncoder`, `EncodingManager` etc.) to a large part use the same code to 'load' as they use to 'import' is a huge technical debt if you ask me. The root of many evils, so to say. However, since we already use `TagParsers` to create the encoded values when we use `load()` after `init()` anyway, it actually might be worth to remove `EncodedValueFactory` first and only introduce something similar when we remove `TagParser#createEncodedValues`.

3) Always create encoded values via `EncodedValueFactory` (sounds right at least) and stop creating encoded values in `TagParsers`. For example I wonder what if we created all encoded values first and then passed them to each tag parser. Maybe this would also allow some new use cases, like modifying an encoded value from different tag parsers, or setting encoded values based on others ('aggregated encoded values').

easbar:
Fixes #2438.

**tldr;**
Why do we need EncodedValueFactory? We don't. Should we create encoded values in two different places (TagParsers and EncodedValueFactory)? No. Should we support two different ways to initialize GraphHopper that do not even follow the same semantics? Also no.

Our current approach is to configure the GraphHopper class the same way to load graph data as we did when we imported data. This does not make much sense, and we should much rather just load the graph from disk when loading, but this is the way we do it currently. So we should do it the same way no matter if we call `init()` or not, `importOrLoad()` or `load()`. Therefore, I now also build the encoding manager from the GraphHopper class' state when loading without init() (instead of reading the properties file). And since this was the only scenario where we even used EncodedValueFactory it can be removed.

When `FlagEncoder`s are gone (#2463) and encoded values are no longer created in `TagParser`s (soon) we can revisit this and hopefully some day we will simply load everything from disk when using `load()`...


core/src/main/java/com/graphhopper/routing/weighting/custom/ExpressionVisitor.java
```diff
@@ -40,7 +37,13 @@
...
             "contains", "sqrt", "abs"));
     private String invalidMessage;
-    private DefaultEncodedValueFactory factory = new DefaultEncodedValueFactory();
+    // todo: not sure if this is worth it just to slightly improve the error message
+    private final Set<String> knownEncodedValues = new HashSet<>(Arrays.asList(
+            Roundabout.KEY, "car_access", "bike_access", GetOffBike.KEY, RoadClass.KEY, RoadClassLink.KEY, RoadEnvironment.KEY,
+            RoadAccess.KEY, MaxSpeed.KEY, MaxWeight.KEY, MaxWidth.KEY, MaxAxleLoad.KEY, MaxLength.KEY, Surface.KEY, Smoothness.KEY,
+            Toll.KEY, TrackType.KEY, BikeNetwork.KEY, FootNetwork.KEY, Hazmat.KEY, HazmatTunnel.KEY, HazmatWater.KEY,
+            Lanes.KEY, MtbRating.KEY, HikeRating.KEY, HorseRating.KEY, Country.KEY
+    ));
```
easbar, line 46: @karussell: If I got this right we used the `DefaultEncodedValueFactory` here to distinguish between 'invalid identifiers' and names of encoded values that are known to the factory, but were not added to the graph. The only purpose of this seem to be improved error messages. I replaced the hard-coded factory with a hard-coded list of encoded value names here. But actually I think we should consider removing this list. A given graphhopper instance has a list of encoded values stored in the graph. I don't think we should maintain a list of encoded values that aren't there, but could be there... At least not until we use such a list of encoded values that could be there (a registry in a way) to actually create the encoded values. Anyway, I updated this list according to: https://github.com/graphhopper/graphhopper/issues/2438#issuecomment-948746635.

karussell: Yes, sure. Let's remove this.

karussell: Yes, sure. Let's remove this.

PR #2555, opened 2022-04-08
Skip emergency access

### LINKED ISSUE #2484: Routing on highway emergency exit without access tag
Instead of going to the next official exit, GH is taking an emergency exit: 
https://graphhopper.com/maps/?point=45.400303%2C5.464877&point=45.400156%2C5.464169

The wiki suggests an access=no in the wiki and this tag is missing: https://www.openstreetmap.org/way/745661088

But as this is a bit dangerous tagging mistake and similar mistakes are already made in the same area (and other routing engines assume access=no by default it seems) we should probably also assume a "block access of this road by default".

otbutz:
Fixes #2484

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### 6.0 [not yet released]
+- don't allow cars or motorcycles to use ways tagged with service=emergency_access (#2484) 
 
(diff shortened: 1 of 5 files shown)


karussell:
Thanks, look good to me! @easbar will merge when appropriate with the current refactoring.

PR #2559, opened 2022-04-14
Remove TagParser#createEncodedValues

easbar:
With this PR encoded values are no longer created inside tag parsers. The tag parsers receive the encoded values they are working with via their constructors instead. The GraphHopper class now creates encoded values via the `EncodedValueFactory` (in every case, not just when calling load() without init()...). Next thing to do will be removing `AbstractFlagEncoder#createEncodedValues` as well.

Related: #2463,#2551

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/TagParserManager.java
+++ b/core/src/main/java/com/graphhopper/routing/util/TagParserManager.java
@@ -136,12 +136,57 @@ public class TagParserManager implements EncodedValueLookup {
             return true;
         }
 
-        public boolean addIfAbsent(TagParserFactory factory, String tagParserString) {
+        public boolean addIfAbsent(EncodedValueFactory encodedValueFactory, TagParserFactory factory, String tagParserString) {
             check();
             tagParserString = tagParserString.trim();
             if (tagParserString.isEmpty()) return false;
 
-            TagParser tagParser = parseEncodedValueString(em, factory, tagParserString);
+            if (!tagParserString.equals(toLowerCase(tagParserString)))
+                throw new IllegalArgumentException("Use lower case for TagParser: " + tagParserString);
+
+            add(encodedValueFactory.create(tagParserString));
+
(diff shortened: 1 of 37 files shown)


core/src/main/java/com/graphhopper/routing/util/TagParserManager.java
```diff
@@ -268,20 +312,31 @@ public TagParserManager build() {
...
-                        _addEdgeTagParser(new DefaultTagParserFactory().create("car_access"), false);
-                    if (!em.hasEncodedValue("bike_access"))
-                        _addEdgeTagParser(new DefaultTagParserFactory().create("bike_access"), false);
+                    if (!em.hasEncodedValue("car_access")) {
+                        em.addEncodedValue(new SimpleBooleanEncodedValue("car_access", true), false);
+                        _addEdgeTagParser(new DefaultTagParserFactory().create(em, "car_access"));
+                    }
+                    if (!em.hasEncodedValue("bike_access")) {
+                        em.addEncodedValue(new SimpleBooleanEncodedValue("bike_access", true), false);
+                        _addEdgeTagParser(new DefaultTagParserFactory().create(em, "bike_access"));
```
karussell, line 321: Above you created the parser (OSMRoadEnvironmentParser etc) without the factory. Should this be more consistent? We could e.g. create the TagParserFactory once and then use it everywhere.

karussell (commented):
Thanks LGTM!

core/src/main/java/com/graphhopper/routing/util/TagParserManager.java
easbar: Ah yes, actually we always create the parsers except for these two new ones (car_access and bike_access). I removed the usages of the DefaultTagParserFactory for these two cases for now. `TagParserManager` will be removed or transformed into something different soon anyway. Then I will check if we can use the factory consistently everywhere or something.

easbar:
Thanks for taking a look!

Issue #1765, opened 2019-10-24
Via routing without u-turns for speed mode


easbar:
Using edge-based CH it should be possible to support proper u-turn prevention for via-point routing (`pass_through=true`). 

Related: https://discuss.graphhopper.com/t/node-id-must-be-a-virtual-node/4829

michaz:
Looking forward to this, as this has applications to everything from map-matching to tour planning.

easbar:
Related:

https://discuss.graphhopper.com/t/suboptimal-route-with-via-points-and-no-u-turn/4866

michaz:
*How interesting*, thank you! To me, this looks like he should try map-matching rather than via-routing. I think the problem is solved there.

michaz:
Actually, via-routing without the greediness  _is_ map-matching. Isn't it?

NoamDev:
Hi @easbar,
I understand I could use the curbside feature as a workaround for now.
Could you please explain to me how to do it correctly?

easbar:
> I understand I could use the curbside feature as a workaround for now.

You just set `curbside=right` (or left) for all points then there will not be any u-turns, but also you are forcing a certain side of the road (that is why this is rather a workaround and we have this issue to support this use-case properly).

NoamDev:
Oh, that's horrible, to make it actually work, I'd have to check any possible combination...
And even then it might not work.

I guess I'm right not to expect any work on it anytime soon (because work is centered around 1.0 release).

I would have tried sending a PR, but I'm not good with writing unit tests..

NoamDev:
When a way point is a virtual node, it would be easy.
We should just set the outEdge as the other edge of the virtual node.

In order to make that work even on tower node, we can use the same suggestion I made in (#1258).

easbar:
A little roadmap since we keep discussing this over and over again:

* [ ] remove unfavored virtual edges in map-matching (use edge-based routing instead)
* [ ] allow setting a finite penalty when using startOut/targetIn edges for bidir algorithms
* [ ] implement via-routing using startOut/targetIn edges instead of unfavored virtual edges
* [ ] remove unfavored virtual edges from code base
* [ ] implement advanced via-routing algorithm that finds the ideal start/target edges at every via-point by taking into account the following legs of the route

easbar:
I did some related work here:

https://github.com/graphhopper/graphhopper/tree/start_target_edge_weights
[...]

PR #2562, opened 2022-04-18
Implement pass_through like curbsides

michaz:
See #1765

Means pass_through needs `turn_costs: true` now.

Let's do the same for heading and we can remove unfavored virtual edges, which are a similar but not quite the same mechanism which is harder to reason about than edge-based routing, which we undestand.


michaz:
@easbar has a better idea and will take it from here. ,-)

PR #2563, opened 2022-04-18
Remove headingPenaltyMillis

easbar:
We currently apply a penalty when heading (or pass_through) is used, but for some reason we do not only use it to increase the edge weights, but also the time. This can result in very long routing times (the default heading penalty is 5 minutes) like here (for heading):

https://graphhopper.com/maps/?point=50.78734%2C8.934395&point=50.788683%2C8.93137&locale=en-US&elevation=true&profile=car&ch.disable=true&heading=180&use_miles=false&selected_detail=Elevation&layer=Omniscale

image

and here (for pass_through):

https://graphhopper.com/maps/?point=50.789334%2C8.933516&point=50.787628%2C8.933752&point=50.788683%2C8.93137&locale=en-US&elevation=true&profile=car&ch.disable=true&pass_through=true&details=weight&details=time&use_miles=false&selected_detail=weight&layer=Omniscale

image

Another problem is that we apply the heading penalty for the route calculation but not for instructions and path details. So if the heading penalty is applied the instruction time does not match the time of the route. 

Here I removed the heading penalty from the edge time.


karussell:
> for some reason we do not only use it to increase the edge weights, but also the time

Do you mean the problem is that a penalty is added twice?

Because in general I think a time penalty is what we want, so in case the heading is incorrect the necessary U turn adds to the ETA (not only "avoiding a U turn").

> default heading penalty is 5 minutes

Ok, then this high value is probably the problem and also that this value should depend on the profile.

easbar:
> Do you mean the problem is that a penalty is added twice?

No. I thought the heading penalty is meant to affect the route, but not the time. Just like a priority. But currently the penalty is applied to both the time and the weight.

> Because in general I think a time penalty is what we want (?), so in case the heading is incorrect the necessary U turn adds to the ETA (not only "avoiding a U turn").

I agree that the heading penalty time could be interpreted as the time it takes to turn around. But in its current state it looks a lot more like a bug: The penalty is much higher than the u-turn time we normally use, there is no instruction to turn around and the instruction time (let alone the path details) do not even include the heading penalty time even though the route does.

Whether or not the 'turn around time' should be included in the ETA or not is probably up to debate, but since it has been in this broken state for so long apparently nobody cares, so we can just as well remove this?

> Ok, then this high value is probably the problem and also that this value should depend on the profile.

Yes, at the very minimum the heading penalty time should be equal to the u-turn time of the profile.

easbar:
Using the heading penalty time to model the initial 'turn around time' also seems a bit like a workaround that is needed for node-based routing: For edge-based routing we can either use the correct heading, or go the other way and do a u-turn at the next junction (which will automatically add the u-turn time to the total routing time). A more advanced implementation would probably still add an initial u-turn (including the u-turn time and a u-turn instruction) at the starting location, though. 

...but our current implementation is not advanced like this. One other problem I saw is this:

https://graphhopper.com/maps/?point=50.791538%2C8.92689&point=50.792176%2C8.928934&locale=en-US&elevation=true&profile=car&ch.disable=true&pass_through=true&details=weight&details=time&heading=270&use_miles=false&selected_detail=weight&layer=Omniscale

image

Here we ask the router to start in Western direction (270 degrees), which is perfectly possible because we can just follow the road "Am kleinen Born". Internally this means the edge going east gets the penalty. But this penalty is applied even when the router follows "Am kleinen Born", goes back to the start point and **then** goes East. So we still get the route that goes East directly including the five minute penalty.

Btw I am currently trying (once again) to implement heading/pass_through properly without unfavored edges in QueryGraph (#1765)

easbar:
If it makes it easier (and so far I think so) I'd like to remove the heading penalty time, make progress with #1765 and if we really want to add the initial 'turn around time' later we should do it afterwards and do it properly this time :)

michaz:
Yes. Yes, yes, yes.

karussell:
Sure, please go ahead to make this change :)

> But this penalty is applied even when the router follows "Am kleinen Born", goes back to the start point and then goes East. So we still get the route that goes East directly including the five minute penalty.

Ah, yes, this is not good and probably only edge-based can help in this case :)

easbar:
> Ah, yes, this is not good and probably only edge-based can help in this case :)

Yes, we only get the route that follows "Am kleinen Born" to turn around with edge-based routing. However, I think with some tweaks we can still use node-based (even with CH) for heading and it will produce the same results as edge-based in many cases (the one in this example is an exception).

michaz (approved):

Issue #2566, opened 2022-04-27
Alternatives are very slow for LM/AStar+turn costs


easbar:
map: bayern-210101.osm.pbf: 
route using maps 2 client: http://localhost:3000/?point=49.583276%2C10.991626&point=48.455783%2C12.095752&profile=car
profile: fastest/car

![image](https://user-images.githubusercontent.com/17603532/165449184-f4af0de6-54c1-4926-af20-836849e0827b.png)

|traversal mode|single route|alternatives (max=3)|
|-|-|-|
|node-based|13ms|~800ms|
|edge-based|25ms|~9000ms|

So alternatives take 80 times longer than a single route for node-based. This seems very slow, but at least it still takes less than a second. But for edge-based it is even worse where the slow down is x360 and the response takes so long that its almost not useful anymore.

This is not just special for this route, I just gave an example. I actually saw routes that took tens of seconds, and I did not even try a larger map. Disabling LM makes it even slower.

easbar:
Using the approach taken by @michaz in #2092 the query time for edge-based alternative seems to go down from ~9s to ~5s. However, for node-based the response becomes (slightly) slower, like ~950ms and also the single routes for edge-based seem to take longer (~40ms vs. ~25ms), but this was just a quick manual check.

easbar:
@otbutz in #2567  you wrote:

> I can vouch for [Java Mission Control](https://adoptium.net/de/jmc/) paired with -XX:+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints for getting a good grasp about where the bottleneck is :)

I am answering here as this is about speeding up alternative routes:

Thanks I will try. I am using the YourKit Profiler, usually. But in my experience this does not help much and the real issues are typically caused by the algorithms (i.e. normally something is called too often rather than some method is slow). In the case of AlternativeRoute I am pretty sure this takes very long, especially for large search trees:

https://github.com/graphhopper/graphhopper/blob/47febfc6278e6eebf15f2940e5bac0c77c0034b5/core/src/main/java/com/graphhopper/routing/AlternativeRoute.java#L321-L326

But I did not really look at the alternative route algorithm to figure out if this can be avoided without making the quality of alternative routes worse.

otbutz:
> normally something is called too often rather than some method is slow

The flame view is immensely helpful IMHO.

https://www.youtube.com/watch?v=vVh_z34Gtn8

easbar:
[...]

PR #2567, opened 2022-04-27
Consistent edge keys

### LINKED ISSUE #2152: Use edge keys consistently
With #2150 we can now obtain edge keys from edge states and obtain a certain edge state from a graph using an edge key.

However, we use a second definition of edge keys that uses the `baseNode<=adjNode` relation instead of the storage direction to determine the edge key. This second kind is used by the edge-based routing algorithms as traversal Ids. 

This is not a real problem, but certainly not ideal and using the same definition of the edge keys would be more consistent.

The task here would be to replace `GHUtility#createEdgeKey(base, adj, edge, reverse)` with the 'internal' edge keys we obtain via `edgeState.getEdgeKey()`. This includes removing `TraversalMode#createTraversalId(base, adj, egde, reverse)`.

I think to do this we need to add `graph.getEdgeKey(edgeId, adjNode)` and/or change the first/lastOrigEdge fields of the edge-based CH shortcuts to first/lastOrigKey fields.

Some work towards this was done in #2146.

easbar:
Fixes #2152. All our code now uses the same definition of an 'edge key'. An edge key is defined as the edge ID multiplied by two when it is parallel to the storage direction (the edge key is even in this case) and the same plus one for the opposite direction (the edge key is odd in this case):

```
EdgeIteratorState edge = graph.edge(a, b);
assertEquals(2*edge.getEdge(), edge.getEdgeKey());
assertEquals(2*edge.getEdge()+1, edge.detach(true).getEdgeKey());
```

The `<`-relation between `a` and `b` no longer affects the edge key used in the routing algorithms. For loop edges there is no way to distinguish the two directions so the edge key is always equal to twice the edge ID for loops (the 'even' direction). 

Edge keys can be reversed to get the key of the opposite direction using `GHUtility#reverseEdgeKey(edgeKey)`. However, it is important not to do this for loop edges! So an edge key must be reversed like this:

```
EdgeIteratorState edge = graph.edge(a, b);
int key = edge.getEdgeKey();
int reverseEdgeKey = edge.getBaseNode() == edge.getAdjNode() ? key : GHUtility.reverseEdgeKey(key);

// there is also:
int reverseEdgeKey = TraversalMode.EDGE_BASED.createTraversalId(edge, true);
```

Here https://github.com/graphhopper/graphhopper/pull/2150#discussion_r499776872 we decided against `edge.getEdgeKey(boolean reverse)`, but considering this special handling we need to do for loops (which we did not recognize back then it seems) maybe we should introduce this now. We could either change `edge.getEdgeKey()` to `edge.getEdgeKey(boolean reverse)` or add `edge.getReverseEdgeKey()`.

Then it would look like this:

```java
int key = edge.getEdgeKey(false);
int reverseKey = edge.getEdgeKey(true);

// .. or
int key = edge.getEdgeKey();
int reverseKey = edge.getReverseEdgeKey();
```

... or we disallow loop edges for `BaseGraph` once and for all :) @michaz didn't you separate PT from BaseGraph now, which would mean there is no longer an actual  need for loop edges?

Btw: The reverse edge key is frequently needed for bidirectional routing algorithms. For example if the forward search finds an edge base=3->adj=5 with edge key 6, the backward search will discover this edge when exploring the opposite node: base=5<-adj=3 and the edge key will be 7 not 6. Therefore we need to 'reverse' the edge key for the backward search (turn 7 into 6).

The biggest change in this PR was changing the edge-based CH shortcuts such that we now store the first/last original edge key instead of the first last original edge ID. This actually removes some code that was needed to turn edge IDs into edge keys at query time (a call to `graph.getOtherNode`). In our measurements I saw a 25% query time drop for edge-based CH alternative routes (something like 19ms instead of 25ms).

Related: #2150, #2146


easbar:
The remaining problem I see here is that this change makes alternative routes even slower than they already are (#2566) (even though it makes edge-based alternative routes with CH *faster*). This is probably because of the `graph.getEdgeIteratorState()` calls we do in `AlternativeRoute` now. I see five ways to fix this at the moment:

1) #1847 
2) store the traversal ID in the shortest path tree entries, so we do not need the `getEdgeIteratorState` calls
3) figure out why alternative routes are so slow in the first place and maybe reduce the number of `getEdgeIteratorState` calls / fix #2566
4) merge this PR despite the slow down because alternative routes are already too slow anyway
5) delete AlternativeRoute, allow alternatives only for CH

Number 3 seems like the best option to me in the long run, especially for providing alternatives with custom models (https://github.com/graphhopper/graphhopper-maps/pull/183). Numbers 4+5 are the quickest options so we can merge here and continue with e.g. #1765. Numbers 1+2 would at least prevent a further slow down of the alternative routes, but also do not really fix anything either in case we all agree that in their current state alternative routes are too slow anyway.

otbutz:
> 3\. figure out why alternative routes are so slow in the first place

I can vouch for [Java Mission Control](https://adoptium.net/de/jmc/) paired with `-XX:+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints` for getting a good grasp about where the bottleneck is :)

easbar:
@otbutz -> https://github.com/graphhopper/graphhopper/issues/2566#issuecomment-1110642088

michaz:
Clearly 4.

michaz:
> didn't you separate PT from BaseGraph now, which would mean there is no longer an actual need for loop edges?

Correct, I'm not putting weird edges into your graph anymore.

michaz (approved):

PR #2568, opened 2022-04-28
custom_model: introduce value expression

karussell:
In master we can only set a number e.g. as limit:
```json
{ "if": "road_class == MOTORWAY", "limit_to": 100 }
```

And with this branch it is possible to use one encoded value on the right-hand side:

```json
{ "if": "road_class == MOTORWAY", "limit_to": "some_numeric_enc_value" }
```

or even an expression:

```json
{ "if": "road_class == MOTORWAY", "limit_to": "2 + some_numeric_enc_value * 0.5" }
```

This is useful in combination with `multiply_by` in the `priority` entry to easily use pre-filled encoded values in the custom model instead of a chain of many if+else conditions. E.g. to apply some factor stored as EncodedValue.

Or it is useful to limit the speed to `0.9*max_speed` useful if based on the unlimited `roads` encoder.

It is also useful with the `roads` encoder where "pre-filled" speeds like the one from car could be used. E.g. an [emergency routing profile](https://discuss.graphhopper.com/t/routing-for-emergency-vehicle/7217) can be implemented using the `roads` encoder, limiting the speed to `1.2 * car_average_speed` and keep the access restrictions soft, so that oneway roads could be accessed in reverse direction (except for motorways):

```json
{
 "speed": [{
   "if": "true",
   "limit_to": "car_average_speed"
  }
 ],
 "priority": [{
   "if": "road_class == MOTORWAY && car_access == false",
   "multiply_by": 0
  },{
   "else_if": "car_access == false",
   "multiply_by": 0.1
  },
  ...
 ]
}
```

(edited after changes in this PR and master)

The challenge is that we use this right-hand side value also to identify the maximum value which is used for the A* algorithm and keeping this max speed "tight" is an important factor for quick response times. As we don't want to calculate the maxima of the "function" we do a quick test with the minimum and maximum of the encoded value and limit the value expression and do not allow a `/` operation. Plus only a single encoded value can be used.

Now three evaluations are necessary instead of just one:
 1. LM constraint check before merging of the query-custom-model into the base-custom-model
 2. to find out minimum & maximum value of `priority` and `speed` (it is not sufficient to evaluate the calcWeight method twice with max+min as all branches of the expression needs to be evaluated and we have to pick the maximum)
 3. in calcWeight to use the maximum of priority and speed from created and cached class so that no re-evaluation is necessary for weightings for CH.

TODOs:

 - [x] decide if we still allow number for value in Java (in JSON we have to allow it)
 - [x] even with a single encoded value there are cases for the operations `-` and `/` where the maximum of the function is not the minimum or maximum of the encoded value e.g. `2-1/(1-x)` -> find a solution. E.g. limit it further or allow `-` and `/` only for the server side where the author has the duty to consider these problems?
 - [x] custom model editor UI: https://github.com/graphhopper/custom-model-editor/pull/1 -> #2615


easbar:
Thanks for this PR!

>  limiting the speed to 1.2 * car$average_speed

In your example there is no factor `1.2`? But I guess you just forgot to add it there?

> The challenge is that we use this value also to identify the maximum value which is used for the A* algorithm 

Maybe we should just keep this as simple as it can be? So far you mentioned an example where we want to use `car$average_speed` on the RHS and scale it using a constant factor (`1.2`). Can you come up with an example that requires more complicated expressions? So far it seems like allowing one EV and a constant factor (or offset using +/-) might be all we really need?

> decide if we still allow number for value in Java (in JSON we have to allow it)

Why do we have to keep allowing numbers in JSON?

> custom model editor UI

Having auto-complete for the EVs on the RHS would be nice and I can add this, but first we should decide what kind of expressions shall be allowed.

easbar:
For #2463 I think the most important thing here is that it will allow removing the `vehicle` field from the profiles (if we add the `set_to` operator as well). I.e. this:

```yaml
name: my_car
vehicle: car
weighting: fastest
```

can become this:

```yaml
name: my_car
weighting: custom
custom_model: {
  "speed": [
      {
          "if": "true",
          "set_to": "car$average_speed"
      }
   ],
   "priority": [
      {
          "if": "!car$access",
          "set_to": 0
      }
    ]  
}
```

? So even without any arithmetic on the RHS this would be useful, so maybe no need to worry about stuff like `2-1/(1-x)` yet?

otbutz:
Wouldn't we still need the TransportationMode which is currently implicitly derived from the vehicle?

easbar:
> Wouldn't we still need the TransportationMode which is currently implicitly derived from the vehicle?

Yes, this is a good question. The transportation mode is used to determine the access, so I think it should be a parameter of the access parser, or maybe the access parser should receive a list of access values directly instead of the transportation mode. I mean something like this:

```yaml
graph.tag_parsers: car$access|values=motorcar,motor_vehicle,vehicle,access
```

which defines the 'transportation mode' used to populate the `car$access` encoded value which can then be used freely in any custom model. Ideally I'd like to have just parsers (that can have some configuration options like this) and custom models that define how the weight of an edge is calculated based on the resulting encoded values.

otbutz:
> Ideally, I'd like to have just parsers (that can have some configuration options like this) and custom models that define how the weight of an edge is calculated based on the resulting encoded values.

100% agree

easbar:
> can become this:

I have to admit this is quite verbose in case all one needs is the default car profile. But maybe we can simplify this by using something like `base: car` which provides a base model the (optional) custom model is merged with. Don't know. My point is that besides the 'model' there should not be anything else that defines the weighting like the encoder.

PR #2569, opened 2022-04-28
Clean up javadoc comments

karussell:
See [this discussion](https://discuss.graphhopper.com/t/7196)

Everyone in the need of javadocs can still create them (I removed the javadoc errors)

### DIFF
--- a/.github/workflows/push-packages.yml
+++ b/.github/workflows/push-packages.yml
@@ -30,9 +30,44 @@ jobs:
           key: ${{ runner.os }}-node-${{ hashFiles('**/pom.xml', '**/package.json') }}
           restore-keys: |
             ${{ runner.os}}-node_modules-
-      - name: Build and publish package
+      - name: Build and publish package ${{ matrix.java-version }}
         run: |
           mvn -B versions:set -DnewVersion=$GITHUB_SHA -DgenerateBackupPoms=false
           mvn -B -DskipTests -Pskip-shaded-web-jar -Pskip-tools-jar source:jar deploy
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  build-only:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        java-version: [ 18 ]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v2
+        with:
+          java-version: ${{ matrix.java-version }}
+          distribution: temurin
+      - name: Cache Maven artifacts
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Cache node
+        uses: actions/cache@v2
+        with:
+          path: web-bundle/node
+          key: ${{ runner.os }}-node-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os}}-node-
+      - name: Cache node_modules
+        uses: actions/cache@v2
+        with:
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/graphhopper/storage/CHStorageBuilder.java
```diff
@@ -63,9 +63,9 @@ public int addShortcutNodeBased(int a, int b, int accessFlags, double weight, in
 
     /**
      * @param origFirst The first original edge that is skipped by this shortcut. For example for the following shortcut
-     *                  edge from x to y, which itself skips the shortcuts x->v and v->y the first original edge would
-     *                  be x->u: x->u->v->w->y
-     * @param origLast  like origFirst, but the last orig edge, i.e w->y in above example
+     *                  edge from x to y, which itself skips the shortcuts x-&gt;v and v-&gt;y the first original edge would
+     *                  be x-&gt;u: x-&gt;u-&gt;v-&gt;w-&gt;y
+     * @param origLast  like origFirst, but the last orig edge, i.e w-&gt;y in above example
```
easbar, line 68: This makes no sense to me to be honest. These comments are meant to be read when browsing the source code and using &gt; instead of > makes this unreadable. The javadocs as html pages on the other hand are used by literally nobody so I don't think we should modify the comments like this.

easbar: I also don't understand why this is needed now when it was not needed for the 5.0 release?

easbar: > I think the documentation inside the IDE parses it as html too (but might not be so strict). If it is just for source code then we could put it into /* normal non-javadoc comments.

Yes, maybe, but I checked and IntelliJ deals with the `>` symbol without problems for example. 

> If it is just for source code then we could put it into /* normal non-javadoc comments.

I like the code highlighting that is used for `/**`, but really I don't care as long as it is readable in IntelliJ and maybe Eclipse and Netbeans if anyone is using these. I definitely do not feel like trying to satisfy some html restrictions of some javadoc plugin that produces html that nobody needs :D

easbar:
And do these fixes even fix the errors we saw in the 5.2 build? Here: https://app.travis-ci.com/github/graphhopper/graphhopper/jobs/568520373
there were errors about all kinds of things, even javadocs that did not include `@param` for all parameters:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:jar (attach-javadocs) on project graphhopper-core: MavenReportException: Error while generating Javadoc: 

[ERROR] Exit code: 1 - /home/travis/build/graphhopper/graphhopper/core/src/main/java/com/graphhopper/GraphHopper.java:162: warning: no @param for precision

[ERROR]     public GraphHopper setPreciseIndexResolution(int precision) {
```

easbar:
Ok, these were logged as `[ERROR]` by maven, but they were only reported as warnings by the javadoc plugin...

Anyway, I think I understood what happened: So far we were using debian based JDKs for the build, which seem to disable doclint: https://discuss.graphhopper.com/t/ch-returning-connection-not-found-between-points/7196/37?u=easbar, but when we changed to Adoptium recently we started using the Temurin build which does not disable doclint so we got the errors. The only thing I still do not get is why the build was 'green', but according to the logs it failed.

This means so far we always had doclint disabled and now we switched it on accidentally. Since we never cared about enabling it I see no reason to enable it now. Therefore the minimal change we can do to continue without the failing build is just disabling doclint explicitly.

core/src/main/java/com/graphhopper/storage/CHStorageBuilder.java
easbar: Ok thanks, yes looks good to me.

easbar:
I just pushed a new tag 5.3 with doclint disabled, let's hope it works this time :)

PR #2574, opened 2022-05-04
Replace deprecated dropwizard parameter types

otbutz:
Replaces both deprecated parameter classes with their java.util counterparts.

see https://github.com/dropwizard/dropwizard/blob/4fb8906bb761886c85891d4bbf6dcf0a15ca4e9c/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params/LongParam.java#L9-L12

### DIFF
--- a/web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
+++ b/web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
@@ -23,8 +23,6 @@ import com.graphhopper.storage.GraphEdgeIdFinder;
 import com.graphhopper.storage.index.LocationIndex;
 import com.graphhopper.storage.index.Snap;
 import com.graphhopper.util.*;
-import io.dropwizard.jersey.params.IntParam;
-import io.dropwizard.jersey.params.LongParam;
 import org.hibernate.validator.constraints.Range;
 import org.locationtech.jts.geom.*;
 import org.slf4j.Logger;
@@ -40,6 +38,8 @@ import javax.ws.rs.core.UriInfo;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.OptionalInt;
+import java.util.OptionalLong;
 import java.util.function.ToDoubleFunction;
 
 import static com.graphhopper.resources.IsochroneResource.ResponseType.geojson;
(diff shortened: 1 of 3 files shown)


web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
```diff
@@ -111,31 +111,26 @@ public Response doGet(
         ShortestPathTree shortestPathTree = new ShortestPathTree(queryGraph, queryGraph.wrapWeighting(weighting), reverseFlow, traversalMode);
 
         double limit;
-        if (weightLimit.get() > 0) {
-            limit = weightLimit.get();
+        ToDoubleFunction<ShortestPathTree.IsoLabel> fz;
+        if (weightLimit.orElseThrow(() -> new IllegalArgumentException("query param weight_limit is not a number.")) > 0) {
```
easbar, line 115: Under which condition will this exception be thrown? And what happened previously under the same condition? Can we add a test for this (just for one of the parameters maybe)'

otbutz: > Under which condition will this exception be thrown? And what happened previously under the same condition?

The handling for Optinal(Double/Int/Long) within dropwizard is a bit different.

For example:

* `weight_limit=1000` -> weightLimit.isPresent() == true && weightLimit.getAsLong() == 1000L
* no query parameter -> weightLimit.isPresent() == true && weightLimit.getAsLong() == -1L (populated via `@DefaultValue`)
* `weight_limit=garbage` -> weightLimit.isPresent() == false && weightLimit.getAsLong() throws NoSuchElementException

> Can we add a test for this (just for one of the parameters maybe)?

We already have such tests:

https://github.com/graphhopper/graphhopper/blob/cff273999af74eff25cdd888087b068de21020b1/web/src/test/java/com/graphhopper/application/resources/IsochroneResourceTest.java#L280

The exception messages are the same as the previous ones to avoid client breakage.

easbar: Ok nice, so in case `weight_limit=garbage` we even get the same error message as before. And when it is missing we get the default value as before.

:+1:

easbar:
Thanks!

PR #2576, opened 2022-05-05
update deployment guide

karussell:
The deployment guide for world wide data isn't up-to-date:

 * include a hint to separate import and server command to reduce memory usage.
 * increased memory requirements due to edge-based CH (turn_costs=true) and also refactored import process.

### DIFF
--- a/docs/core/deploy.md
+++ b/docs/core/deploy.md
@@ -24,8 +24,9 @@ routing requests. The G1 is the default GC but the other two GCs are better suit
 You enable them with `-XX:+UseZGC` or `-XX:+UseShenandoahGC`. Please note that especially ZGC and G1 require quite a
 bit memory additionally to the heap and so sometimes speed can be increased when you lower the `Xmx` value.
 
-If you want to support none-CH requests you should at least enable landmarks or limit the request to a
-certain distance via `routing.non_ch.max_waypoint_distance` (in meter, default is 1).
+If you want to support none-CH requests you should consider enabling landmarks or limit requests to a
+certain distance via `routing.non_ch.max_waypoint_distance` (in meter, default is 1) or
+to a node count via `routing.max_visited_nodes`.
 Otherwise it might require lots of RAM per request! See [#734](https://github.com/graphhopper/graphhopper/issues/734).
 
 ### API Tokens


otbutz:
I'd advice to use the `-XX:+UseNUMA` flag for multiprocessor deployments.

otbutz:
https://docs.oracle.com/javase/8/docs/technotes/guides/vm/performance-enhancements-7.html#numa

> When evaluated against the SPEC JBB 2005 benchmark on an 8-chip Opteron machine, NUMA-aware systems showed the following performance increases:
> 
>     32 bit – About 30 percent increase in performance with NUMA-aware allocator
>     64 bit – About 40 percent increase in performance with NUMA-aware allocator

The only possible downside is for JVMs <11: https://bugs.openjdk.java.net/browse/JDK-8189922

otbutz:
Note that `-XX:+UseNUMA` mandates `-XX:+UseParallelGC` for it to take affect

> -XX:+UseNUMA
>
> Enables performance optimization of an application on a machine with nonuniform memory architecture (NUMA) by increasing the application's use of lower latency memory. By default, this option is disabled and no optimization for NUMA is made. **The option is available only when the parallel garbage collector is used (-XX:+UseParallelGC).** 
>
> -XX:+UseParallelGC
> 
> Enables the use of the parallel scavenge garbage collector (also known as the throughput collector) to improve the performance of your application by leveraging multiple processors.

otbutz:
G1GC is also NUMA aware since Java 14: https://openjdk.java.net/jeps/345

otbutz:
> I meant real production experience with GraphHopper. (And running GraphHopper with ParallelGC might not be something you want in production for serving routes. Maybe for import.)

Fair point. We *had* a NUMA system running with GH at some point.

boldtrn:
Interesting [ZGC seems to be using NUMA by default](https://wiki.openjdk.java.net/display/zgc/Main#Main-EnablingNUMASupport):

> ZGC has NUMA support, which means it will try it's best to direct Java heap allocations to NUMA-local memory. This feature is enabled by default. However, it will automatically be disabled if the JVM detects that it's bound to only use memory on a single NUMA node. In general, you don't need to worry about this setting, but if you want to explicitly override the JVM's decision you can do so by using the -XX:+UseNUMA or -XX:-UseNUMA options.

README.md
```diff
@@ -75,7 +75,7 @@ To get started you can try [GraphHopper Maps](README.md#graphhopper-maps), read
 
 ## Installation
 
-To install the [GraphHopper Maps](https://graphhopper.com/maps/) UI and the web service locally you [need a JVM](https://adoptopenjdk.net/) (>= Java 8) and do:
+To install the [GraphHopper Maps](https://graphhopper.com/maps/) UI and the web service locally you [need a JVM](https://adoptium.net) (>= Java 8) and do:
 
 ```bash
 wget https://github.com/graphhopper/graphhopper/releases/download/5.1/graphhopper-web-5.1.jar https://raw.githubusercontent.com/graphhopper/graphhopper/5.x/config-example.yml http://download.geofabrik.de/europe/germany/berlin-latest.osm.pbf
```
easbar, line 81: We are at 5.3 currently

docs/core/deploy.md
```diff
@@ -5,20 +5,27 @@ This guide is written for everyone interested in deploying graphhopper on a serv
...
+routing requests. The G1 is the default GC but the other two GCs are better suited for JVMs with bigger heaps (>32GB) and low pauses.
 You enabled them with `-XX:+UseZGC` or `-XX:+UseShenandoahGC`. Please note that especially ZGC and G1 require quite a
 bit memory additionally to the heap and so sometimes speed can be increased when you lower the `Xmx` value.
 
-Notes:
-
- * If you want to support none-CH requests you should at least enable landmarks or limit the request to a certain
- * distance otherwise it might require lots of RAM per request! See [#734](https://github.com/graphhopper/graphhopper/issues/734).
+If you want to support none-CH requests you should at least enable landmarks or limit the request to a
+certain distance otherwise it might require lots of RAM per request! See [#734](https://github.com/graphhopper/graphhopper/issues/734).
```
easbar, line 28: Maybe mention the parameter that does this `way_point_max_distance` (or similar).

docs/core/deploy.md
```diff
@@ -5,20 +5,27 @@ This guide is written for everyone interested in deploying graphhopper on a serv
...
+"import" command explicitly before the "server" command:
+
+```
+java [options] -jar *.jar import config.yml
+java [options] -jar *.jar server config.yml # calls the import command implicitly, if not done before
+```
 
-If you use JDK11 you should try a different garbage collectors like G1, ZGC or Shenandoah.
-The G1 is the default but the other two GCs are better suited for JVMs with bigger heaps (>32GB) and low pauses.
+Try a different garbage collectors (GCs) like ZGC or Shenandoah for serving the
```
easbar, line 22: Try ~~a~~ different ...

docs/core/deploy.md
```diff
@@ -5,20 +5,27 @@ This guide is written for everyone interested in deploying graphhopper on a serv
...
+routing requests. The G1 is the default GC but the other two GCs are better suited for JVMs with bigger heaps (>32GB) and low pauses.
 You enabled them with `-XX:+UseZGC` or `-XX:+UseShenandoahGC`. Please note that especially ZGC and G1 require quite a
```
easbar, line 24: enable~~d~~

docs/core/deploy.md
```diff
@@ -34,11 +41,13 @@ To be able to use the autocomplete feature of the point inputs you have to:
...
-It needs approximately 25GB RAM for the import (CAR only) and ~1 hour (plus ~5h for contraction).
-If you can accept slower import times this can be reduced to 14GB RAM - you'll need to set `datareader.dataaccess=MMAP` in the config file.
+GraphHopper is able to handle coverage for the world-wide [OpenStreetMap road network](http://planet.osm.org/).
+
+Without enabled CH the import step alone requires ~60GB RAM and takes ~3h for the import. If you can accept
+much slower import times (3 days!) this can be reduced to 31GB RAM when you set `datareader.dataaccess=MMAP` in the config file.
 
-Then 'only' 15GB are necessary and without contraction hierarchy this can be further reduced to about 9GB.
+With enabled CH for the car profile (with turn cost) the import command needs ~120GB RAM and the additional CH preparation takes ~25 hours
+but heavily depends on the CPU and memory speed. Without turn cost support e.g. sufficient for bike it takes much less (~5 hours).
```
easbar, line 50: Maybe also add the requirements for LM. Something like ~5 hours.

docs/core/deploy.md
```diff
@@ -34,11 +41,13 @@ To be able to use the autocomplete feature of the point inputs you have to:
 
 ## Worldwide Setup
 
-GraphHopper is able to handle coverage for the whole [OpenStreetMap road network](http://planet.osm.org/).
-It needs approximately 25GB RAM for the import (CAR only) and ~1 hour (plus ~5h for contraction).
-If you can accept slower import times this can be reduced to 14GB RAM - you'll need to set `datareader.dataaccess=MMAP` in the config file.
+GraphHopper is able to handle coverage for the world-wide [OpenStreetMap road network](http://planet.osm.org/).
```
easbar, line 44: GraphHopper can handle the world-wide OpenStreetMap road network

docs/core/deploy.md
```diff
@@ -34,11 +41,13 @@ To be able to use the autocomplete feature of the point inputs you have to:
...
+
+Without enabled CH the import step alone requires ~60GB RAM and takes ~3h for the import. If you can accept
```
easbar, line 46: Parsing the OSM planet file and creating the GraphHopper base graph requires ~60GB RAM ...

Running the CH preparation, which is required for maximum routing speed, takes ... without turn costs and ... with turn costs.

Running the LM preparation ... takes ...

It is also possible to add CH/LM preparations for existing profiles after the initial import. Adding or modifying profiles is not possible and you need to run a new import instead.

docs/core/deploy.md
easbar: Maybe mention different map sizes, for example planet car has around 415mio edges, Europe 150mio, North America 86mio.

easbar: Maybe mention different map sizes, for example planet car has around 415mio edges, Europe 150mio, North America 86mio.

docs/core/deploy.md
```diff
@@ -5,20 +5,28 @@ This guide is written for everyone interested in deploying graphhopper on a serv
...
+
+Try different garbage collectors (GCs) like ZGC or Shenandoah for serving the
+routing requests. The G1 is the default GC but the other two GCs are better suited for JVMs with bigger heaps (>32GB) and low pauses.
+You enable them with `-XX:+UseZGC` or `-XX:+UseShenandoahGC`. Please note that especially ZGC and G1 require quite a
+bit memory additionally to the heap and so sometimes speed can be increased when you lower the `Xmx` value.
 
- * If you want to support none-CH requests you should at least enable landmarks or limit the request to a certain
- * distance otherwise it might require lots of RAM per request! See [#734](https://github.com/graphhopper/graphhopper/issues/734).
+If you want to support none-CH requests you should at least enable landmarks or limit the request to a
+certain distance via `routing.non_ch.max_waypoint_distance` (in meter, default is 1).
```
otbutz, line 28: Should we also mention `routing.max_visited_nodes` ?

easbar:
thanks!

PR #2579, opened 2022-05-11
Log available country rules

otbutz:
We already log if we are using custom areas. This makes it a bit easier for library users to tell if their own rules are getting properly applied.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -745,6 +745,12 @@ public class GraphHopper {
             customAreas.addAll(readCustomAreas());
         }
         AreaIndex<CustomArea> areaIndex = new AreaIndex<>(customAreas);
+        
+        if (countryRuleFactory == null || countryRuleFactory.getCountryToRuleMap().isEmpty()) {
+            logger.info("No country rules available");
+        } else {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -745,6 +745,12 @@ protected void importOSM() {
             customAreas.addAll(readCustomAreas());
         }
         AreaIndex<CustomArea> areaIndex = new AreaIndex<>(customAreas);
+        
+        if (countryRuleFactory == null || countryRuleFactory.getCountryToRuleMap().isEmpty()) {
+            logger.info("No country rules available");
+        } else {
+            logger.info("Applying rules for the following countries: {}", countryRuleFactory.getCountryToRuleMap().keySet());
```
easbar, line 752: Can you sort the countries alphabetically, just to make it a bit more deterministic?

otbutz: The factory uses an EnumMap so the order is determined by the Country enum which is sorted.

https://github.com/graphhopper/graphhopper/blob/c28a5123d198156f8aff9242e6f8f15fb50a3de0/core/src/main/java/com/graphhopper/routing/util/countryrules/CountryRuleFactory.java#L31

easbar: Ok, I see.

easbar:
Thanks

PR #2585, opened 2022-05-24
Remove 'shared' encoded values with $ sign in name

karussell:
Also remove OSMAccessParser. Later we will probably reintroduce a similar thing but for now we keep using car$access (with this PR it will be car_access)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ### 6.0 [not yet released]
-- replaced car$access with car_priority (and same for <vehicle>$average_speed and <vehicle>$priority)
+- replaced car$access with car_access (and same for <vehicle>$average_speed and <vehicle>$priority)
 - don't allow cars or motorcycles to use ways tagged with service=emergency_access (#2484)
 - faster flexible routing, especially in conjunction with turn costs
 


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ### 6.0 [not yet released]
+- replaced car$access with car_priority (and same for <vehicle>$average_speed and <vehicle>$priority)
```
easbar, line 2: You probably meant car_access here.

core/src/test/java/com/graphhopper/routing/util/EncodingManagerTest.java
```diff
@@ -75,12 +75,13 @@ public void testSupportFords() {
...
-        for (String str : Arrays.asList("blup_test", "test", "test12", "tes$0", "car_test_test", "small_car$average_speed")) {
+        for (String str : Arrays.asList("blup_test", "test", "test12", "car_test_test")) {
             assertTrue(EncodingManager.isValidEncodedValue(str), str);
         }
 
-        for (String str : Arrays.asList("Test", "12test", "test|3", "car__test", "blup_te.st_", "car___test", "car$$access",
-                "test{34", "truck__average_speed", "blup.test", "test,21", "täst", "blup.two.three", "blup..test")) {
+        for (String str : Arrays.asList("Test", "12test", "test|3", "car__test", "small_car$average_speed", "tes$0",
+                "blup_te.st_", "car___test", "car$$access", "test{34", "truck__average_speed", "blup.test", "test,21",
+                "täst", "blup.two.three", "blup..test")) {
```
easbar, line 84: Does this mean any $ sign is forbidden in custom models now? Was it forbidden before?

easbar (approved):
Nice, this is another step towards #2463. Now the 'shared' encoded values are a bit more like the other encoded values again :)

easbar:
Should we remove `car_access` and `bike_access` here as well?

https://github.com/graphhopper/graphhopper/blob/be7b2476aaff55c7c266fd517c8cf2c37ae91b43/core/src/main/java/com/graphhopper/util/details/PathDetailsBuilderFactory.java#L72

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -211,23 +205,19 @@ private void addEncoder(VehicleEncodedValues encoder) {
...
+        }
         list = new ArrayList<>();
         encoder.createTurnCostEncodedValues(list);
         for (EncodedValue ev : list)
             addTurnCostEncodedValue(ev);
         edgeEncoders.add(encoder);
     }
 
-    private void addEncodedValue(EncodedValue ev, boolean withNamespace) {
-        String normalizedKey = ev.getName().replaceAll(SPECIAL_SEPARATOR, "_");
```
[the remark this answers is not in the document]
easbar: And it did not prevent us from adding the same EV name either...
For example when we added an EV with `ev.getName() = "car$access" twice (from different FlagEncoders that return the same name EV in createEncodedValues) the `normalizedKey` would be "car_access" (except it was not but that does not even matter here). But since we add the EV with the key `ev.getName()` the second time we add it `hasEncodedValue(normalizedKey)` would still be false, so there was nothing that stopped us from adding multiple encoded values with the same name, each time overwriting the previous one.

easbar:
> Hm, probably we should even allow this for all transportation modes? 
> Same for <encoder>_average_speed.

Yes, maybe like we deal with (the former) `vehicle$priority` above? But the `car_access` and `bike_access` here were specifically added for the osm access parser, i.e. I think they weren't meant for the former `car/bike$access`.

> (Note, that we use just average_speed to return the average speed of the specified profile and this works already for custom profiles without a separate encoded value.)

Yes, `average_speed` here is different as it considers the chosen weighting.

easbar:
And now we are at it.. (I think I asked this before): Why do we have to explicitly list all the encoded values here in the first place? Can we not just loop over `encodingManager.getEncodedValues`?

easbar:
> (Probably we started with explicit path details like time and average_speed and added the others later step by step :))

Doing this with a for loop means that encoded values are added implicitly to the server API (in path details). This is nice because then we get all the details, but also means that the server API is changed when we change the encoded values...

easbar:
I meant the available path details, i.e. if we specify `details=abc` and later `abc` is renamed to `xyz` the old request will yield an error. And similarly for custom models, even though they are not affected by PathDetailsBuilderFactory.

easbar:
> Then we should not rename the EVs :)

Ok, but we just did :) We renamed e.g. `bike$priority` to `bike_priority`. Then again, probably nobody ever knew this path detail even existed :D

PR #2587, opened 2022-05-30
Matrix client improvements

### LINKED ISSUE #2553: Matrix client hangs and exposes confusing API
In a fresh maven/Java project:

pom.xml
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>org.example</groupId>
    <artifactId>client-hc-test</artifactId>
    <version>1.0-SNAPSHOT</version>

    <dependencies>
        <dependency>
            <artifactId>directions-api-client-hc</artifactId>
            <version>5.0</version>
        </dependency>
    </dependencies>

    <properties>
        <maven.compiler.source>8</maven.compiler.source>
        <maven.compiler.target>8</maven.compiler.target>
    </properties>

</project>
```

Main.java (change the key)

```java
import com.graphhopper.api.GHMRequest;
import com.graphhopper.api.GraphHopperMatrixWeb;
import com.graphhopper.api.MatrixResponse;

import java.util.Arrays;

public class Main {
    public static void main(String[] args) {
        GraphHopperMatrixWeb matrixClient = new GraphHopperMatrixWeb();
        matrixClient.setKey("your_key");
        GHMRequest ghmRequest = new GHMRequest();
        ghmRequest.addFromPoint(new GHPoint(49.6724, 11.3494));
        for (GHPoint to : Arrays.asList(new GHPoint(49.6724, 11.3494), new GHPoint(49.6550, 11.4180))) {
            ghmRequest.addToPoint(to);
        }
        MatrixResponse response = matrixClient.route(ghmRequest);
        if (response.hasErrors())
            throw new RuntimeException(response.getErrors().toString());
        System.out.println("it's over... but apparently it's not...");
    }
}
```

```sh
mvn compile
mvn exec:java -Dexec.mainClass="Main"
```
-> prints "it's over... but apparently it's not...", but then the process never returns.
Pausing the program with a debugger showed that it hangs in `SocketInputStream#socketRead0`. 
I tried this with Java 8, 11 and 17.

The GHMRequest API is quite confusing: `setProfile` does not set the profile. This does nothing:
```
GHMRequest ghmRequest = new GHMRequest();
ghmRequest.setProfile("small_truck");
```

`setPoints` does not set the points. This does nothing either.
```
GHMRequest ghmRequest = new GHMRequest();
ghmRequest.setPoints(points);
```

It seems very strange that `GHMRequest` extends from `GHRequest`. This is probably the reason `setProfile` and `setPoints` do not work. Apparently we have to use `addPoint,addAllPoint etc.` instead.

karussell:
Fixes #2553

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/subnetwork/EdgeBasedTarjanSCC.java
+++ b/core/src/main/java/com/graphhopper/routing/subnetwork/EdgeBasedTarjanSCC.java
@@ -455,7 +455,7 @@ public class EdgeBasedTarjanSCC {
 
         @Override
         public void minTo(int key, int value) {
-            // todonow: optimize
+            // todo: optimize with map.indexOf(key) etc
             map.put(key, Math.min(map.getOrDefault(key, -1), value));
         }
 
--- a/core/src/test/java/com/graphhopper/routing/util/BikeTagParserTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/BikeTagParserTest.java
@@ -597,7 +597,7 @@ public class BikeTagParserTest extends AbstractBikeTagParserTester {
         way = new ReaderWay(1);
         way.setTag("highway", "residential");
         way.setTag("maxspeed", "15");
-        // todonow!! speed is larger than maxspeed
+        // todo: speed is larger than maxspeed tag due to rounding and storable max speed is 30
         assertPriorityAndSpeed(PREFER.getValue(), 16, way);
     }
 


client-hc/src/main/java/com/graphhopper/api/GraphHopperMatrixWeb.java
```diff
@@ -38,9 +38,8 @@ public GraphHopperMatrixWeb setKey(String key) {
     public MatrixResponse route(GHMRequest request) {
         if (!Helper.isEmpty(key))
             request.getHints().putObject(KEY, key);
-        if (!request.getPathDetails().isEmpty())
-            throw new IllegalArgumentException("Path details are not supported for the Matrix API");
-        request.compactPointHints();
+        // todonow
+//        request.compactPointHints();
```
easbar, line 42: probably there is nothing we need to compact anymore, because the point_hint field will simply be missing when it is not used?

client-hc/src/test/java/com/graphhopper/api/Examples.java
```diff
@@ -92,23 +92,29 @@ public void matrix() {
...
+        MatrixResponse responseSymm = matrixClient.route(ghmRequest);
+        if (responseSymm.hasErrors())
+            throw new RuntimeException(responseSymm.getErrors().toString());
+        // get time from first to second point:
+        // System.out.println(response.getTime(0, 1));
 
+        // Option 2: for an asymmetric matrix do:
+        ghmRequest = new GHMRequest();
+        ghmRequest.setOutArrays(Arrays.asList("distances", "times"));
+        ghmRequest.putHint("vehicle", "car");
```
easbar, line 110: I guess we should no longer recommend using the vehicle hint/parameter, but rather `setProfile("car")`? Here and a few lines above...

client-hc/src/main/java/com/graphhopper/api/GHMRequest.java
```diff
@@ -1,193 +1,163 @@
...
 
-    @Override
-    public GHRequest setCurbsides(List<String> curbsides) {
-        fromCurbsides = curbsides;
-        toCurbsides = curbsides;
+    // a good trick to serialize unknown properties into the HintsMap
+    @JsonAnySetter
+    public GHMRequest putHint(String fieldName, Object value) {
+        if (profile != null && "profile".equals(fieldName))
+            throw new IllegalArgumentException("use 'profile' only explicitly");
```
easbar, line 136: Makes kind of sense, I guess but unfortunately this can easily be circumvented by using `getHints().putObject("profile", "car")`. So maybe rather check if "profile" is contained in the hints in `createPostRequest` later?

client-hc/src/test/java/com/graphhopper/api/AbstractGHMatrixWebTester.java
```diff
@@ -185,17 +188,19 @@ public void testReadingWeights_TimesAndDistances() throws IOException {
     @Test
     public void noVehicleWhenNotSpecified() {
```
easbar, line 189: We should probably change this test to check there is no **profile** if not specified now.

client-hc/src/main/java/com/graphhopper/api/GHMRequest.java
easbar: And if we do this maybe the same for fail_fast as it is also proper field here and should be consistent with the hints.
... and maybe even a warning if 'vehicle' is still used? No, probably this is too much.

client-hc/src/test/java/com/graphhopper/api/AbstractGHMatrixWebTester.java
```diff
@@ -185,17 +188,19 @@ public void testReadingWeights_TimesAndDistances() throws IOException {
...
+        JsonNode json = requester.createPostRequest(new GHMRequest().setOutArrays(Collections.singletonList("weights")).setPoints(Arrays.asList(new GHPoint(11,12))));
+        assertEquals("{\"points\":[[12.0,11.0]],\"out_arrays\":[\"weights\"],\"fail_fast\":true}", json.toString());
     }
 
     @Test
     public void hasHintsWhenSpecified() {
         GHMatrixAbstractRequester requester = createRequester("url");
-        GHMRequest ghmRequest = new GHMRequest(5);
+        GHMRequest ghmRequest = new GHMRequest();
         ghmRequest.putHint("vehicle", "my_car").putHint("profile", "my_profile");
```
easbar, line 199: maybe use some other hints here. profile should really be set via `setProfile`.

easbar (commented):
Nice!

easbar (approved):

PR #2589, opened 2022-06-01
use for loop in PathDetailsBuilderFactory 

karussell:
as discussed in #2585

### DIFF
--- a/core/src/main/java/com/graphhopper/util/details/PathDetailsBuilderFactory.java
+++ b/core/src/main/java/com/graphhopper/util/details/PathDetailsBuilderFactory.java
@@ -21,10 +21,8 @@ import com.graphhopper.routing.ev.*;
 import com.graphhopper.routing.weighting.Weighting;
 
 import java.util.ArrayList;
-import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


otbutz:
Could we also loop over the special ones? Bonuspoint if we keep them as a Set and properly catch missing encoded values here:

https://github.com/graphhopper/graphhopper/pull/2589/files#diff-d4b26b63682c84a31c6c475e20445d463958e27b8456a53dd5916ef8234bf2f8R60

otbutz:
We can properly distinguish between missing encoded value and extra details here:

https://github.com/graphhopper/graphhopper/blob/01bd55fa086513bffbbe5043d10257e2ff82da86/core/src/main/java/com/graphhopper/util/details/PathDetailsBuilderFactory.java#L60

and directly throw an exception with the detail we couldn't find instead of the length comparison later on:

https://github.com/graphhopper/graphhopper/blob/01bd55fa086513bffbbe5043d10257e2ff82da86/core/src/main/java/com/graphhopper/util/details/PathDetailsBuilderFactory.java#L76-L78

PR #2593, opened 2022-06-02
Replace loop with System.arraycopy

otbutz:
Shorter and probably faster as `System.arraycopy()` is an IntrinsicCandidate.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/BaseGraph.java
+++ b/core/src/main/java/com/graphhopper/storage/BaseGraph.java
@@ -745,9 +745,7 @@ public class BaseGraph implements Graph, Closeable {
         public final EdgeIteratorState setFlags(IntsRef edgeFlags) {
             assert edgeId < store.getEdges() : "must be edge but was shortcut: " + edgeId + " >= " + store.getEdges() + ". Use setFlagsAndWeight";
             store.writeFlags(edgePointer, edgeFlags);
-            for (int i = 0; i < edgeFlags.ints.length; i++) {
-                this.edgeFlags.ints[i] = edgeFlags.ints[i];
-            }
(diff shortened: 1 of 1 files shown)


easbar:
Why do we copy the edge flags here anyway? Would it not be sufficient to call `store.writeFlags` and set `freshFlags` to `false`, so the copying is done lazily later? Either way I think this is not critical at all though, because during the graph import the actual writing/copying of the edge flags plays a very minor role, at least when reading OSM data or similar.

easbar:
> Unfortunately the measurement showed no speed up.

We'd probably need a micro benchmark that just sets the flags. Our measurement only calls setFlags during the import where many other things contribute to the measured time. Then again if there are many other things speeding up setFlags is probably not so important anyway :)

I'm not sure about what kind of thread safety we are trying to guarantee. The getFlags method with the lazy copying of the edge flags certainly does not look thread safe? But otherwise yes let's just remove this.

easbar:
Don't we have to set `refreshFlags = true`? It could be that we first read some flags (call getFlags()), so `refreshFlags` will be false, then call setFlags with some new flags and when we call getFlags afterwards we need to read the new flags into the local intsref.

easbar:
I guess there is a bit of an asymmetry now: We update the cached flags in the different `set` methods, because we do not really have a choice. We need an updated `IntsRef` object to write to the underlying storage and probably do not want to create it for every `set` call. For `setFlags` we do not necessarily need to do it because the `IntsRef` is passed into the method. Now we removed it so the method behaves a bit differently compared to the other `set` methods.

easbar:
And maybe add the two tests while you're at it.

easbar:
If we revert, we can still remove `refreshFlags = false` I think. Maybe then it is a bit less confusing as we do not do this for the other setters either.

PR #2594, opened 2022-06-03
Make motor vehicle and hgv properties configurable for roads vehicle EVs

easbar:
With this PR the default is `roads|is_motor_vehicle=true|is_hgv=false` (as before), but now we can change it.

Note that currently the turn costs are hard-coded for this encoder, i.e. setting `turn_costs=false` would not have an effect.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/DefaultFlagEncoderFactory.java
+++ b/core/src/main/java/com/graphhopper/routing/util/DefaultFlagEncoderFactory.java
@@ -28,7 +28,7 @@ public class DefaultFlagEncoderFactory implements FlagEncoderFactory {
     @Override
     public FlagEncoder createFlagEncoder(String name, PMap configuration) {
         if (name.equals(ROADS))
-            return FlagEncoders.createRoads();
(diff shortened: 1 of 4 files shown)


easbar:
> Or maybe directly use transportation_mode and the enum parser?

The VehicleEncodedValues (former AbstractFlagEncoder), or let's just say the FlagEncoders, do not have a transportation mode. The TransportationMode is only used to get the list of access values for the VehicleTagParsers. The `isMotorVehicle` flag is used for things like the private/destination penalties in `FastestWeighting` and should be replaced by something else at some point, but so far we still need it.

easbar:
I mean if you need a list of access values or a special TransportationMode (currently it is always set to TransportationMode.VEHICLE) for the `Roads**TagParser` then this would be a property of the parser, not the encoded values...

easbar:
... and since parsers aren't configured explicitly (yet) you would have to add a `transportation_mode` (or `access_values`) property to 'the' access encoded value (and so on :))

easbar:
Does this make sense?

Issue #1554, opened 2019-02-25
MIssing Highway Exit Number


CodeMouse92:
When driving directions involve a highway, exit numbers really need to be included!

It is tremendously difficult in places to translate something like "Turn right onto South Helena Street" from westbound I-90 in Spokane, when the exit is actually for 2nd Avenue. Exit numbers are fairly important, especially for printed instructions.

boldtrn:
It should be noted that exit numbers are important in some parts of the world, for example in north America, but are not frequently used in other parts of the world, for example in Germany many other European countries. So this is something that could be added depending on the region?

CodeMouse92:
Quite probably.

With that said, exit numbers are generally based on mileposts, and I can't help but wonder if that would be a useful feature internationally with highways and whatnot (turn onto N. Whatever St. - milepost 15). If you're in the U.S., work on exit numbers, but if you're elsewhere, make use of mileposts?

boldtrn:
>A number of those are distance-based exit numbers (correlating with mileposts/kilometers); several are sequential instead, and would have to be mapped individually.

I haven't checked that in detail, but I would assume that all would have to be mapped individually. If you are up for it, maybe you could try and have a crack at it?

CodeMouse92:
I'm afraid I don't have any free coding bandwidth, otherwise I'd go for it. :(

theadventurer62:
Regarding this, it seems that OSRM already includes this in navigation. 

To solve this issue, I propose taking consideration of using motorway links and junction nodes with `ref` tags to add exit numbers to navigation.

Issue #1086, opened 2017-06-07
Instructions/Name: Parse destination tag


boldtrn:
A nice addition for instructions would be to use the [destination tag](https://wiki.openstreetmap.org/wiki/Key:destination). This could be especially handy for _link roads, where we could create instructions like: take exit, following signs to "Berlin". One example would be this [link](http://www.openstreetmap.org/way/154765839) that misses the name tag, but has a ref.

The question is, how to store the data. 

Probably the easiest solution would be to store this in the name field. This was discussed in the topic about adding different languages for the name tag (couldn't find the issue, sorry :)).

I could implement this, if we can decide on a way to store this.

devemux86:
@boldtrn you probably mean #259.
(where I propose the process we use for multilingual Mapsforge maps in writer / readers)

boldtrn:
@devemux86, yes exactly, thanks for linking it. I used the wrong search terms, I think :).

karussell:
> if we can decide on a way to store this

This feels wrong. Let us first find a better way of storing this and if this turns out to be too complex we can still use the name string.

boldtrn:
> This feels wrong. Let us first find a better way of storing this and if this turns out to be too complex we can still use the name string.

Ok, what do you propose?

Should we use something like NameIndex, but call it DestinationIndex? Or should we try to extend the NameIndex to store multiple values, like different languages and different purposes like destination? 

Creating a second index feels a bit like an overkill, but the implementation would be straight forward as we can reuse a lot of existing code from the NameIndex.

Creating a more flexible storing solution would be probably the best, but how would we do this?

karussell:
> Creating a more flexible storing solution would be probably the best, but how would we do this?

We would need a more flexible key/value storage to store more complicated properties per edge.

boldtrn:
> We would need a more flexible key/value storage to store more complicated properties per edge.

Yes I know, that's why I am asking. However, at this point I don't have a good proposal. Easy solutions like serializing key/value pairs as JSON come to my mind, but this might be too slow. If we only use the name for creating instructions, this might be still ok, but we shouldn't store anything that we want to use during the route calculation.

Penegal:
That would be a great improvement, especially for complex junctions in which reading the destination signs can be quicker than looking for the good road following the instructions.

PR #2597, opened 2022-06-07
EdgeKVStorage: store more than Strings

karussell:
This PR does not change the implementation but makes it clear that the StringIndex is now a storage that has more flexibility as it allows us to store ints, longs, floats, doubles, byte arrays and Strings per edge instead of just Strings. It still has some limitations (StringIndex had them too):

 * the key strings are limited to 32767 unique values 
 * the dynamic value has a maximum bytes array length of 255
 * every key can store values only of the same Class (the first store call sets the Class)
 * it cannot store null values

This mechanism is introduced parallel to the EncodedValue-mechanism to store values per edge. This KV approach has the following differences:

 * it is slower as we jump to the external storage and then loop to this value x times to get value_x (but this probably doesn't matter for path details and if the average key count is low per edge)
 * if the values are written more than once the space isn't reused (even if less space is required for the new value -> could be improved later)
 * every value uses size(type) plus 3 bytes, e.g. for an int we need 7 bytes.

This EdgeKVStorage opens up a lot of possibilities for improvements like dynamically translated road names (#259), splitting the "ref" tag from the road name (PR as follow up #2598) and issues like #1554 and #1086.


easbar:
Nice, this sounds good :) 

> as it allows us to store ints, longs, floats, doubles, byte arrays and Strings per edge instead of just Strings.

Can you give an example where it would make sense to store ints, longs, floats, doubles or byte arrays using EdgeKVStorage, rather than using Int/DoubleEncodedValue? Can we use it to store the OSM way IDs (an often requested feature)? And would it make sense to use EdgeKVStorage for this or would we better implement a LongEncodedValue?

> splitting the "ref" tag from the road name (PR as follow up https://github.com/graphhopper/graphhopper/pull/2598)

But isn't this just about storing strings? Why was this not possible using StringIndex?

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -332,7 +332,8 @@ protected void addEdge(int fromIndex, int toIndex, PointList pointList, ReaderWa
         if (edgeFlags.isEmpty())
             return;
 
-        String name = way.getTag("way_name", "");
+        // the storage does not accept too long strings
+        String name = Helper.cutString(way.getTag("way_name", ""), 255);
```
easbar, line 336: What was the limit so far?

core/src/main/java/com/graphhopper/search/StringIndex.java
```diff
@@ -1,337 +0,0 @@
...
-import com.graphhopper.util.GHUtility;
-import com.graphhopper.util.Helper;
-
-import java.util.*;
-
-/**
- * @author Peter Karich
- */
-public class StringIndex {
-    private static final long EMPTY_POINTER = 0, START_POINTER = 1;
```
easbar, line 33: Unfortunately git no longer recognizes StringIndex.java and EdgeKVStorage.java to be the same file, even though they are still quite similar, and the same for the corresponding tests. It might be worth to first do (only) the renaming in master and then merging this PR. This way there will be a commit that still allows seeing the actual changes.

easbar:
Ok thanks for your explanations. That makes sense and especially being able to store values for a few edges only is very nice.

> See also comment from @michaz https://github.com/graphhopper/graphhopper/issues/2085#issuecomment-680916873

I'm not sure what exactly was meant there, but some sort of ByteArrayEncodedValue could also be useful (and would in principle be possible, right?)

core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
```diff
@@ -197,127 +233,198 @@ public Map<String, String> getAll(final long entryPointer) {
...
+        else if (name.equals("d")) return Double.class;
+        else if (name.equals("[")) return byte[].class;
+        else throw new IllegalArgumentException("Cannot find class. Unknown short name " + name);
+    }
+
+    private static class LengthResult {
+        int len;
+    }
+
+    private Object putIntoMap(LengthResult result, long tmpPointer, int keyIndex) {
```
easbar, line 321: I can't see any map here. What is put into which map?

core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
```diff
@@ -197,127 +233,198 @@ public Map<String, String> getAll(final long entryPointer) {
...
+        else if (name.equals("l")) return Long.class;
+        else if (name.equals("f")) return Float.class;
+        else if (name.equals("d")) return Double.class;
+        else if (name.equals("[")) return byte[].class;
+        else throw new IllegalArgumentException("Cannot find class. Unknown short name " + name);
+    }
+
+    private static class LengthResult {
+        int len;
+    }
```
easbar, line 319: Why is this needed? Why no primitive int or Integer?

core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
```diff
@@ -70,57 +83,61 @@ public EdgeKVStorage create(long initBytes) {
...
+    public long add(final Map<String, Object> entryMap) {
         if (entryMap.isEmpty())
             return EMPTY_POINTER;
         else if (entryMap.size() > 200)
             throw new IllegalArgumentException("Cannot store more than 200 entries per entry");
 
-        // This is a very important compressing mechanism due to the nature of OSM. Make this faster via precalculated hastcodes?
+        // This is a very important "compression" mechanism due to the nature of OSM.
         if (entryMap.equals(lastEntryMap))
             return lastEntryPointer;
```
easbar, line 142: It looks like this compression relied on `String#equals` or the fact that Strings are immutable. But does it still work when the values are arbitrary `Object`?  It seems like the objects need to implement some kind of object equality so we recognize different objects with the same contents as 'equal' here. Of course this is fulfilled for String and also Integer, Long, Double etc., but maybe it is worth a comment/javadoc or something? Or are the Objects always expected to be a `String` or a `Number`? Then we could even enforce this here.

core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
```diff
@@ -30,37 +30,50 @@
...
-    // vals count| key_idx_0| val_length_0| val_0| key_idx_1| val_length_1| val_1| -key_idx_2             | delta_2 | key_idx_3 | val_length_3 | val_3
-    // Drawback: we need to loop through the entries to get the start of val_x.
-    // Note, that we detect duplicate values via smallCache and then use the negative key index as 'duplicate' marker.
-    // We then store only the delta (signed int) instead the absolute unsigned long value to reduce memory usage when duplicate entries.
+    // storage layout per entry, examples:
+    // vals count (1 byte)|
+    // key_idx_0  (2 byte)| val_length_0 (1 byte)| val_0 (x bytes) <- store String or byte[] with dynamic length
+    // key_idx_1  (2 byte)| int          (4 byte)                  <- store int with fixed length
+    // -key_idx_2 (2 byte)| delta_2      (4 byte)                  <- store duplicate, then key is negative
+    // key_idx_3  (2 byte)| val_length_3 (1 byte)| val_3 (x bytes)
```
easbar, line 45: I am not sure I understand this. Is this the layout of the `vals` DataAccess? Or the one of the `keys` DataAccess, or maybe a combination of both? Can you maybe add some explanation of the general idea how data is stored here?

I guess I understand the first and the fourth lines: We use key_idx_0/3 to point into the vals DataAccess and the first byte we read tells us the length of the value in bytes x which we can then read. But what about the second line? It says we just read four bytes as an int, but how do we know the first byte is not the length as for the other two? And for the third line we see a negative index that is used for duplicates. But why do we not just point to the same value for duplicates? I.e. when key_idx_0 and key_idx_2 are supposed to point to the same value why don't they just store the same key (which points to the value they are both referring to)?

Sorry I am probably missing some basics here, but then again if this documentation is supposed to explain how the index worked I feel like it isn't telling me enough :)

easbar (commented):
I probably don't know the StringIndex enough to review this properly, so do not hesitate to merge this as it is.

easbar:
> Not sure why github wasn't able to figure out that this was a renaming.

I don't think it was githubs fault, it was the same using git locally. I think git does not care much about file names and when there are too many changes in a file it no longer treats it to be the same when the file was renamed. Anyway, thanks for renaming in master, this definitely makes it easier to track changes if we ever come back here.

core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
```diff
@@ -70,57 +83,61 @@ public EdgeKVStorage create(long initBytes) {
...
     }
 
     public boolean loadExisting() {
         if (vals.loadExisting()) {
             if (!keys.loadExisting())
                 throw new IllegalStateException("Loaded values but cannot load keys");
-            int stringIndexKeysVersion = keys.getHeader(0);
-            int stringIndexValsVersion = vals.getHeader(0);
-            GHUtility.checkDAVersion(keys.getName(), Constants.VERSION_STRING_IDX, stringIndexKeysVersion);
-            GHUtility.checkDAVersion(vals.getName(), Constants.VERSION_STRING_IDX, stringIndexValsVersion);
```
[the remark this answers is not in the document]
easbar: Ah yeah do we have a similar mechanism as for encoded values here? I mean after the import it should not be possible to change the layout/keys anymore right?

core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
```diff
@@ -36,20 +39,43 @@ public class EdgeKVStorage {
     static final int MAX_UNIQUE_KEYS = (1 << 15);
     // Store string value as byte array and store the length into 1 byte
     private static final int MAX_LENGTH = (1 << 8) - 1;
+
+    // It stores the mapping of "key to index" in the keys DataAccess. E.g. if your first key is "some" then we will
+    // store the mapping "1->some" there (the 0th index is skipped on purpose). As this map is 'small' the keys
+    // DataAccess is only used for long term storage, i.e. only in loadExisting and flush. Otherwise we use keysInMem and keyList.
```
easbar, line 45: But keys stores the "index to key" mapping right? I mean what we want is "key to index", but what we actually store is "index to key". 

keysInMem and keyList? You probably mean `Map<String, Integer> keyToIndex`, i.e. we read the keys DataAccess and keep the reverse mapping ("some"->1) in memory?

core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
easbar: Ah now I get it. You need a mutable object that you can use as result as method argument. But if you need to return multiple values why do you not create a result class that contains all the result values? Especially considering that you created a result class already :) Anyway, not so important.

core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
easbar: Ok just to say it in my own words: We have a `vals` array that, for each edge, stores the number of keys used for this edge followed by triples `key/value_length/value`. For every edge we store a pointer into this array pointing to the first key/value pair of that edge in the edge flags (`E_NAME`). This way, given an edge, we can use the pointer to find the first key/value pair and traverse all the key/value pairs in order by jumping to the next using the value_length. We know when to stop, because we also stored the number of keys. Each key is stored as two-byte number (key_idx) and we store the mapping between key_idxs and the actual (String) keys in a separate `keys` array. Besides the key we also store the 'type' of each value and for some types we skip storing the value_length and derive the value_length from the type, i.e. 4 bytes for an int and so on. We prevent storing the same value multiple times by storing a pointer to the first occurrence of a value instead of that value. We recognize that a value is a pointer to the actual value by negating the key, i.e. key=-6 means key=6, but the value is a pointer to the actual value.

easbar (approved):

core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
easbar: Oh, right. But why do we have to parse anything then? Can we not just check `object instanceof String || object instanceof Number || object instanceof byte[]`?

easbar: Btw for byte[] the equals method probably does not behave as we'd like it to behave. We would have to use `Arrays.equals`:

```java
    @Test
    public void byteArrayEquals() {
        byte[] a = new byte[] { 0, 1, 2};
        byte[] b = new byte[] { 0, 1, 2};
        assertFalse(a.equals(b));
        assertTrue(Arrays.equals(a, b));
    }
```

easbar: If a method accepts a `Map<String, Object>` and `Object` can only be a `Number`, `String` or a `byte[]` I think it can improve readability when the first thing this method does is checking if that is actually the case. Especially when we run lots of code before this differentiation is done (like everything before the first time we call getBytesForValue here(?)). So yes there is no real need to do it here, but since we already distinguish the `byte[]` case it would not hurt to check for String/Number as well. I would not insist on this though :)

PR #2598, opened 2022-06-07
EdgeKVStorage: store name and ref tags separately

### LINKED ISSUE #1322: Split street_name into street_name and street_ref
I've been doing some work on the openstreetmap project, specifically with the routing, and its been useful to have the street_name (i.e. `"Mittelstrasse"`) separate from the street_ref (i.e. `"E13"`).

At the moment, its all in one string `"Mittelstrasse, E13"`, which is fine, but there are cases when its useful to have the two separate entities (i.e. in the case of openstreetmap, when giving the output some custom formatting. OSRM thankfully splits up the data so we can format it as we please)

I'd be happy to write a pull request that splits up the data, but it would create quite a change for the output which would probably effect quite a lot of of applications using graphhopper, so just wanted to check here first!

karussell:
Fixes #1322 and problems where a point_hint does not match due to the `ref` included in the street name. The instructions will now only contain the `ref` tag instead of the `name` tag if the name `tag` is empty/missing.

This creates a small backward incompatibility change as the street_name does not longer contain the ref (it was separated by a comma). Another small incompatible change for the street_name path details is that the path detail now contains `null` instead of an empty string in case no name tag exists (like for max_speed).

Additionally:

 * created separate a path detail for `ref`
 * use `ref` as fallback in instruction text to avoid lack of details for bigger roads
 * Introduced `EdgeIteratorState.setKeyValues` instead of `EdgeIteratorState.setName`
 * to avoid problems with NPE the Instruction.getName will now never return a null name (although the stored name could be null). This is a compromise to avoid other incompatible changes. Same for EdgeIteratorState.getName.

Follow up of #2597

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
+++ b/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
@@ -24,8 +24,6 @@ import com.graphhopper.storage.NodeAccess;
 import com.graphhopper.util.*;
 import com.graphhopper.util.shapes.GHPoint;
 
-import java.util.Map;
-
 /**
  * This class calculates instructions from the edges in a Path.
  *
@@ -139,13 +137,12 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
             assert Double.compare(prevLon, nodeAccess.getLon(baseNode)) == 0;
         }
 
-        Map<String, Object> keyValues = edge.getKeyValues();
-        final String name = (String) keyValues.get("name");
+        final String name = (String) edge.getValue("name");
         if ((prevInstruction == null) && (!isRoundabout)) // very first instruction (if not in Roundabout)
         {
             int sign = Instruction.CONTINUE_ON_STREET;
             prevInstruction = new Instruction(sign, name, new PointList(10, nodeAccess.is3D()));
-            prevInstruction.setExtraInfo("ref", keyValues.get("ref"));
+            prevInstruction.setExtraInfo("ref", edge.getValue("ref"));
             double startLat = nodeAccess.getLat(baseNode);
             double startLon = nodeAccess.getLon(baseNode);
             double heading = AngleCalc.ANGLE_CALC.calcAzimuth(startLat, startLon, latitude, longitude);
@@ -203,7 +200,7 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
         } else if (prevInRoundabout) //previously in roundabout but not anymore
         {
             prevInstruction.setName(name);
-            prevInstruction.setExtraInfo("ref", keyValues.get("ref"));
+            prevInstruction.setExtraInfo("ref", edge.getValue("ref"));
 
             // calc angle between roundabout entrance and exit
             double orientation = AngleCalc.ANGLE_CALC.calcOrientation(prevLat, prevLon, latitude, longitude);
@@ -277,7 +274,7 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
                     prevInstructionName = prevName;
                     ways.add(prevInstruction);
                 }
-                prevInstruction.setExtraInfo("ref", keyValues.get("ref"));
+                prevInstruction.setExtraInfo("ref", edge.getValue("ref"));
             }
             // Update the prevName, since we don't always create an instruction on name changes the previous
             // name can be an old name. This leads to incorrect turn instructions due to name changes


core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -958,17 +949,26 @@ public int getReverseEdgeKey() {
...
+            int kvEntryRef = store.getKeyValuesRef(edgePointer);
+            return baseGraph.edgeKVStorage.getAll(kvEntryRef);
+        }
+
+        @Override
+        public String getName() {
+            int kvEntryRef = store.getKeyValuesRef(edgePointer);
+            String name = (String) baseGraph.edgeKVStorage.get(kvEntryRef, "name");
+            // preserve backward compatibility (returns null if not explicitly set)
+            return name == null ? "" : name;
```
easbar, line 971: also here: empty<->null?

core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
```diff
@@ -185,20 +187,35 @@ public boolean isStoreTwoDirections() {
...
     String getName();
 
-    EdgeIteratorState setName(String name);
+    /**
+     * This stores the specified key-value pairs in the storage of the current edge. This is more flexible compared to
+     * the mechanism of flags and EncodedValue but is slower and more inefficient on retrieval and this setKeyValues method
+     * should be called only once per edge as it allocates new space everytime this method is called. I.e. for many
+     * usages like in a custom_model currently the EncodedValue-approach should be preferred.
+     */
+    EdgeIteratorState setKeyValues(Map<String, Object> map);
```
easbar, line 216: Can you add a note that this mechanism allows storing 'sparse' values? I.e. instead of reserving space for a certain property for every edge it allows storing a list of key value pairs per edge and the keys do not need to be the same for each edge. That seems key to understand this feature to me. I even thought about going as far as renaming `setKeyValues` to `setSparseProps` and maybe `EdgeKVStorage` to `SparsePropertyStore` or something, but not sure if that makes sense.

core/src/main/java/com/graphhopper/util/details/KVStringDetails.java
```diff
@@ -19,32 +19,39 @@
...
+        this.key = key;
     }
 
     @Override
     public boolean isEdgeDifferentToLastEdge(EdgeIteratorState edge) {
-        if (curStreetName == null || !curStreetName.equals(edge.getName())) {
-            curStreetName = edge.getName();
+        if (curString == null) {
+            // TODO it would be a bit more efficient if we fetch the Map only once per edge (we don't if more than one KV path detail is requested)
+            //  but then we would have to couple these path detail somehow. Or at least use the slightly more efficient EdgeKVStorage.get method instead of getAll
```
easbar, line 41: But we do not use getAll here?

easbar (commented):
Nice, this looks good to me

core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -958,17 +949,26 @@ public int getReverseEdgeKey() {
...
         @Override
-        public String getName() {
-            int nameRef = store.getNameRef(edgePointer);
-            String name = (String) baseGraph.edgeKVStorage.get(nameRef, EDGEKV_NAME_KEY);
-            // preserve backward compatibility (returns null if not explicitly set)
-            return name == null ? "" : name;
+        public EdgeIteratorState setKeyValues(Map<String, Object> map) {
+            int kvEntryRef = (int) baseGraph.edgeKVStorage.add(map);
+            if (kvEntryRef < 0)
+                throw new IllegalStateException("Too many maps are stored, currently limited to 'int' pointer");
```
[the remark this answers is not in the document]
easbar: I meant theoretically `baseGraph.edgeKVStorage` could return a long like `long x = 6_000_000_000L`. Then `(int)x` would be > 0 but it would still be an overflow. So should it not be something like:

```java
long ptr = baseGraph.edgeKVStorage.add(map);
if (ptr > Integer.MAX_VALUE)
   throw new IllegalStateException("EdgeKVStorage overflow, the number of stored key value pairs exceeds our four byte pointer limitation");
 if (ptr < 0)
    throw new IllegalStateException("Negative pointer returned from EdgeKVStorage#add");
 int kvEntryRef = (int) ptr;     
```

core/src/main/java/com/graphhopper/util/details/KVStringDetails.java
easbar: Ah I see. So if this is important maybe there should be `edge.getKeyValue(key)` instead of just `edge.getKeyValues().get(key)`?

core/src/main/java/com/graphhopper/util/EdgeIteratorState.java
easbar: Ok yes.

core/src/main/java/com/graphhopper/storage/BaseGraph.java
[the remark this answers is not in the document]
easbar: Yes, the added map cannot have more than `Integer.MAX_VALUE` entries, but the number returned from the `add` method is the byte pointer and this can be increased by more than `Integer.MAX_VALUE`, e.g. when the map values are large (and numerous), doesn't it?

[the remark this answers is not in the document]
easbar: Ah ok because the number of key value pairs and the maximum string length are limited. Ok yes this was only a theoretical issue and apparently it is not even possible to run into a problem. Don't know. When I cast to a long the obvious thing to do seems to be checking if the value is in the sane range where longs are converted to int without any overflow before doing the cast, but if you rather check if the number after the cast is negative I won't complain :)

PR #2603, opened 2022-06-13
Fix bug in non-CH AlternativeRoute

### LINKED ISSUE #2566: Alternatives are very slow for LM/AStar+turn costs

karussell:
Should fix #2566

Using the fixes of this PR the plateaus show up properly now:

grafik

See the alt_route_debug branch for the necessary changes in MiniGraphUI to see this (using berlin data). Note 1. the two red lines are the best and alternative route in the end, 2. same colors does not mean the plateaus are related and 3. the two block circles are the start and end locations.

Before it was a mess and showed hundreds of potential candidates (many of them were overlapping) that we had to loop through.

The speed improvement should also show up for node-based.

### DIFF
--- a/tools/src/main/java/com/graphhopper/tools/Measurement.java
+++ b/tools/src/main/java/com/graphhopper/tools/Measurement.java
@@ -196,13 +196,13 @@ public class Measurement {
                 boolean isLM = false;
                 measureRouting(hopper, new QuerySettings("routing", count / 20, isCH, isLM).
                         withInstructions());
-                measureRouting(hopper, new QuerySettings("routing_alt", count / 1000, isCH, isLM).
+                measureRouting(hopper, new QuerySettings("routing_alt", count / 500, isCH, isLM).
                         alternative());
                 if (encoder.supportsTurnCosts()) {
                     measureRouting(hopper, new QuerySettings("routing_edge", count / 20, isCH, isLM).
                             withInstructions().edgeBased());
                     // unfortunately alt routes are so slow that we cannot really afford many iterations
-                    measureRouting(hopper, new QuerySettings("routing_edge_alt", count / 1000, isCH, isLM).
+                    measureRouting(hopper, new QuerySettings("routing_edge_alt", count / 500, isCH, isLM).
                             edgeBased().alternative()
                     );
                 }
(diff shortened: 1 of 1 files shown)


tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -202,10 +202,9 @@ protected void importOSM() {
                     measureRouting(hopper, new QuerySettings("routing_edge", count / 20, isCH, isLM).
                             withInstructions().edgeBased());
                     // unfortunately alt routes are so slow that we cannot really afford many iterations
-                    // actually it is just too slow so we skip it for now
-//                    measureRouting(hopper, new QuerySettings("routing_edge_alt", count / 1000, isCH, isLM).
-//                            edgeBased().alternative()
-//                    );
+                    measureRouting(hopper, new QuerySettings("routing_edge_alt", count / 1000, isCH, isLM).
+                            edgeBased().alternative()
```
easbar, line 206: Currently we use `count=5000`, so `count / 1000` means only five routes will be measured. Could be still useful, or could fluctuate too much. Let's see. Otherwise we could try something like `count / 500`.

tools/src/main/java/com/graphhopper/tools/Measurement.java
```diff
@@ -225,9 +224,8 @@ protected void importOSM() {
                             if (args.getBool("measurement.lm.edge_based", encoder.supportsTurnCosts())) {
                                 measureRouting(hopper, new QuerySettings("routingLM" + activeLMCount + "_edge", count / 20, isCH, isLM).
                                         withInstructions().activeLandmarks(activeLMCount).edgeBased());
-                                // this is too slow
-//                                measureRouting(hopper, new QuerySettings("routingLM" + activeLMCount + "_alt_edge", count / 1000, isCH, isLM).
-//                                        activeLandmarks(activeLMCount).edgeBased().alternative());
+                                measureRouting(hopper, new QuerySettings("routingLM" + activeLMCount + "_alt_edge", count / 1000, isCH, isLM).
+                                        activeLandmarks(activeLMCount).edgeBased().alternative());
```
easbar, line 228: also here maybe we should change this to `count / 500` or so.

PR #2607, opened 2022-06-17
Load EncodingManager from properties instead of config

easbar:
So far we use the config file to create the encoding manager, whether we run a new import or load an existing graph. This means we have to use the exact same configuration when loading a graph that we used for the import, even though we cannot change the encoding manager after the first import anyway. A more comfortable approach would be loading the encoding manager from the stored data and not having to use the same configuration again.

Another problem with the current approach is that we'd like to store the maximum speeds used during the import so we can use them for the A* heuristics later. So far these values were taken from the config (or the hard-coded defaults). We can store the max values in the encoded values and write them to the properties file (or elsewhere), but then of course we need to read them later when loading the graph. 

Starting with this PR all encoded values, flag encoders (and the encoding manager) are created from a string representation in the properties file when we load the graph. This allows storing the maximum speeds/values along with the encoded value properties.


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -497,7 +501,6 @@ public GraphHopper init(GraphHopperConfig ghConfig) {
         flagEncodersString = ghConfig.getString("graph.flag_encoders", flagEncodersString);
         encodedValuesString = ghConfig.getString("graph.encoded_values", encodedValuesString);
         dateRangeParserString = ghConfig.getString("datareader.date_range_parser_day", dateRangeParserString);
-        buildEncodingManagerAndOSMParsers(flagEncodersString, encodedValuesString, dateRangeParserString, profilesByName.values());
```
easbar, line 500: We no longer build the encoding manager in `init()`, because when we are loading we want to create it from the stored properties instead.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -858,39 +861,50 @@ public boolean load() {
 
         if (!allowWrites && dataAccessDefaultType.isMMap())
             dataAccessDefaultType = DAType.MMAP_RO;
-        if (encodingManager == null)
-            // we did not call init(), so we build the encoding manager now.
-            // just like when calling init, users have to make sure they use the same setup for import and load
+
```
easbar: This problem no longer exists. The encoding manager is never created in `init()` so it does not matter whether we called it or not.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -858,39 +861,45 @@ public boolean load() {
 
         if (!allowWrites && dataAccessDefaultType.isMMap())
             dataAccessDefaultType = DAType.MMAP_RO;
-        if (encodingManager == null)
-            // we did not call init(), so we build the encoding manager now.
-            // just like when calling init, users have to make sure they use the same setup for import and load
-            buildEncodingManagerAndOSMParsers(flagEncodersString, encodedValuesString, dateRangeParserString, profilesByName.values());
```
easbar, line 864: This problem no longer exists as we do not build the encoding manager in `init()` anymore and it does not matter whether we called it or not.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -858,39 +861,45 @@ public boolean load() {
...
-                .withTurnCosts(encodingManager.needsTurnCostsSupport())
-                .setSegmentSize(defaultSegmentSize)
-                .build();
         properties = new StorableProperties(directory);
-        checkProfilesConsistency();
 
-        if (!new File(ghLocation).exists())
+        if (!new File(ghLocation).exists()) {
+            // there is nothing to load, so we create the EM from the config
+            buildEncodingManagerAndOSMParsers(flagEncodersString, encodedValuesString, dateRangeParserString, profilesByName.values());
```
easbar, line 871: We build the encoding manager (and the parsers) from the config file, but only when there is no -gh folder.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -858,39 +861,45 @@ public boolean load() {
...
+            if (directory.getDefaultType().isStoring() && isAllowWrites()) {
                 lockFactory.setLockDir(new File(ghLocation));
                 lock = lockFactory.create(fileLockName, false);
                 if (!lock.tryLock())
                     throw new RuntimeException("To avoid reading partial data we need to obtain the read lock but it failed. In " + ghLocation, lock.getObtainFailedReason());
             }
-
-            if (!loadBaseGraphAndProperties())
+            if (!properties.loadExisting())
                 return false;
```
easbar, line 893: I just realized that this could be problematic: When there is a -gh folder, but it is empty (no properties file) we will return here but no encoding manager and no base graph was created. Can we move the `properties.loadExisting()` call further up, before obtaining the read lock? Or can we create the lock earlier? And do we really need this lock?

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -909,32 +918,35 @@ public boolean load() {
...
-            if (!encodingManager.toFlagEncodersAsString().equalsIgnoreCase(flagEncodersStr)) {
-                throw new IllegalStateException("Flag encoders do not match:"
-                        + "\nGraphhopper config: " + encodingManager.toFlagEncodersAsString()
-                        + "\nLoaded graph:       " + flagEncodersStr
-                        + "\nChange configuration to match the graph or delete " + baseGraph.getDirectory().getLocation());
-            }
+        EncodingManager.Builder emBuilder = EncodingManager.start();
+        flagEncoders.forEach(emBuilder::addWithoutAddingEncodedValues);
+        encodedValues.forEach(emBuilder::add);
+        encodingManager = emBuilder.build();
```
easbar, line 941: We deserialized the encoded values and flag encoders and now build the encoding manager. However, the flag encoders are handled a bit differently here: When we create the encoding manager from the config file the flag encoders are used to *create* certain encoded values (like car_access etc.), but here we already deserialized all the encoded values and need to skip the creation of encoded values when adding the flag encoders. Flag encoders will hopefully be removed soon anyway :)

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -909,32 +918,35 @@ public boolean load() {
...
+    }
 
-            String encodedValueStr = properties.get("graph.encoded_values");
-            if (!encodingManager.toEncodedValuesAsString().equalsIgnoreCase(encodedValueStr)) {
-                throw new IllegalStateException("Encoded values do not match:"
-                        + "\nGraphhopper config: " + encodingManager.toEncodedValuesAsString()
-                        + "\nLoaded graph:       " + encodedValueStr
-                        + "\nChange configuration to match the graph or delete " + baseGraph.getDirectory().getLocation());
-            }
```
easbar, line 933: This is the code that was forcing us to pass the same configuration when loading the graph even though it was already determined by the import.

core/src/main/java/com/graphhopper/routing/ev/IntEncodedValueImpl.java
```diff
@@ -213,7 +213,7 @@ public final String getName() {
     @Override
     public final String toString() {
         return getName() + "|version=" + getVersion() + "|bits=" + bits + "|min_value=" + minValue
-                + "|negate_reverse_direction" + negateReverseDirection + "|index=" + fwdDataIndex
+                + "|negate_reverse_direction=" + negateReverseDirection + "|index=" + fwdDataIndex
```
easbar, line 216: unrelated to this PR

core/src/main/java/com/graphhopper/routing/ev/IntEncodedValueImpl.java
```diff
@@ -243,10 +243,11 @@ public final int hashCode() {
 
     @Override
     public int getVersion() {
+        // todonow: there isn't really any 'version' here...
```
easbar, line 246: Do we need some kind of version for EVs? For example this PR changes the format of the serialized strings. Maybe something like `Constants.VERSION_EV`?

core/src/main/java/com/graphhopper/routing/ev/IntEncodedValueImpl.java
```diff
@@ -243,10 +243,11 @@ public final int hashCode() {
...
         int val = Helper.staticHashCode(name);
         val = 31 * val + (storeTwoDirections ? 1231 : 1237);
         val = 31 * val + (negateReverseDirection ? 13 : 17);
-        return staticHashCode(val, fwdDataIndex, bwdDataIndex, bits, minValue, maxValue, fwdShift, bwdShift, fwdMask, bwdMask);
+        return staticHashCode(val, bits, minValue, maxValue);
```
easbar, line 250: I wasn't entirely sure how to treat these fields. They represent the position of the encoded values in the edge flags (I think), so they are only really useful in the context of the encoding manager. So far I excluded them from the serialization string. I check the version when deserializing the encoded values (see EncodedValueSerializer), but since the encoded values are not added to an encoding manager at this point these fields are all zero, i.e. they do not match the values that were used during serialization.

We have two options: Either include all these fields in the serialization string and in the version, or exclude them in both cases. Or maybe a third option: Skip the version check when deserializing the encoded values, which also makes kind of sense at least as long `getVersion()` does not really include a 'version' (see above comment).
For now I decided to exclude these fields in both cases, because I thought they do not really belong to the 'identity' of a single encoded value. Btw: Why do we need non-default equals/hashCode implementations here?

core/src/test/java/com/graphhopper/GraphHopperTest.java
```diff
@@ -2582,6 +2583,10 @@ void testLoadingWithAnotherSpeedFactorFails() {
             hopper.importOrLoad();
         }
         {
+            // todonow: what do we want here: setting the flag encoder string obviously has no effect, because we
+            //          load the encoding manager from disk instead. should there be an error? and how? basically this
+            //          is where we deviate from our current approach and we would probably have to adjust the GH api
+            //          to make this clear
             GraphHopper hopper = new GraphHopper()
                     .setFlagEncodersString("car|speed_factor=9")
                     .setProfiles(new Profile("car").setVehicle("car").setWeighting("fastest"))
```
easbar, line 2592: This is basically what changed in this PR: We used to get an error unless we used the same `graph.encoded_values` and `graph.flag_encoders` we used for the import. But now these settings are just ignored when loading the graph, just like `prepare.ch.threads` for example. Unfortunately this is not so obvious when looking at the config file or the GraphHopper class' API. For the config file it might help to use separate sections for 'import' and 'load' so we can distinguish those settings that only affect the import from those that still can be changed when loading a graph.
Regarding the GraphHopper class' API I am a bit less concerned as I think that class should be split into an 'Importer' and a 'Loader' at some point anyway.

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: > What would be the problem in that case?

Right now we get an `IllegalStateException` when the -gh folder exists, but the properties file is missing or the entire folder is empty. The reason is that `baseGraph` is null and we check for null in the `process` method, so actually it is rather like a `NullPointerException`. 

> We have no other protection against multiple writes into the same folder.

Ok, and that protects us in case there are multiple GH processes using the same folder or why do we need this protection? But here we also do not need to write anything, we just want to read the contents of the folder and properties file (if they exist). Would it be OK to call `properties.loadExisting()` before obtaining the lock? I still think there should simply be static methods that tell us whether a certain file exists or allow us to load it without creating any DataAccess objects (let alone without obtaining any (write?) locks :), but this is for another issue...).

easbar: I fixed the issue here: cadbca1c39a4a11c8c4fa9dd87d4ad3e078e24fc, so no when we import into an empty (but existing) folder everything works. Still, the code is quite twisted :)

core/src/main/java/com/graphhopper/routing/ev/IntEncodedValueImpl.java
easbar: Ok, I am not entirely sure why I left out these fields in https://github.com/graphhopper/graphhopper/pull/2542. I found this comment:

https://github.com/graphhopper/graphhopper/blob/acd82a80784a20a2c51626609347447fa1fb5bb5/core/src/main/java/com/graphhopper/routing/ev/IntEncodedValueImpl.java#L48-L49

So probably I ran into the same problem as now: As long as we had no way of adding encoded values to the encoding manager without having the encoding manager set these fields automatically it was easiest to leave them out during serialization. But now this should be possible by adding an extra method to `EncodingManager#Builder`.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -909,32 +922,49 @@ public boolean load() {
...
+    private void loadEncodingManagerFromProperties(StorableProperties properties) {
+        if (properties.containsVersion())
+            throw new IllegalStateException("The GraphHopper file format is not compatible with the data you are " +
+                    "trying to load. You either need to use an older version of GraphHopper or run a new import");
+
+        String versionStr = properties.get("graph.em.version");
+        if (versionStr.isEmpty() || !String.valueOf(Constants.VERSION_EM).equals(versionStr))
+            throw new IllegalStateException("Incompatible encoding version. You need to use the same GraphHopper version you used to import the graph, or run a new import. "
+                    + " Stored encoding version: " + (versionStr.isEmpty() ? "missing" : versionStr) + ", used encoding version: " + Constants.VERSION_EM);
```
easbar, line 933: We have this new `Constants.VERSION_EM` now that defines the version of the stored encoding manager. We do not need a 'version' for each encoded value separately I think, because we store all the different encoded values with all their properties.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -909,32 +922,49 @@ public boolean load() {
...
+                }))
+                .collect(Collectors.toMap(FlagEncoder::getName, f -> (VehicleEncodedValues) f,
+                        (f1, f2) -> {
+                            throw new IllegalStateException("Duplicate flag encoder: " + f1.getName() + " in: " + flagEncodersStr);
+                        },
+                        LinkedHashMap::new));
+
+        EncodedValue.InitializerConfig edgeConfig = EncodedValueSerializer.deserializeInitializerConfig(properties.get("graph.em.edge_config"));
+        EncodedValue.InitializerConfig turnCostConfig = EncodedValueSerializer.deserializeInitializerConfig(properties.get("graph.em.turn_cost_config"));
+        encodingManager = new EncodingManager(encodedValues, flagEncoders, edgeConfig, turnCostConfig);
```
easbar, line 959: I changed this now: We just store the init configs in the properties file and do not use the EM Builder at all anymore when deserializing the EM.

core/src/main/java/com/graphhopper/routing/ev/IntEncodedValueImpl.java
```diff
@@ -212,100 +211,7 @@ public final String getName() {
...
-
-    /**
-     * Produces a static hashcode for an integer arrays that is platform independent and still compatible to the default
-     * of openjdk.
-     *
-     * @see Arrays#hashCode(int[])
-     */
-    static int staticHashCode(int... vals) {
-        if (vals == null)
-            return 0;
```
easbar, line 260: no need for these methods here anymore, but if you think they are generally useful we could move them into Helper or something?

core/src/main/java/com/graphhopper/routing/util/EncodingManager.java
```diff
@@ -94,29 +97,57 @@ public static class Builder {
...
+            if (init)
+                encodedValue.init(em.edgeConfig);
+            else {
+                // we do not 'init' the encoded value, but we still need to update the edge config
+                // todonow: real ugly, bc right now it only works for IntEncodedValueImpl. at the very least there should
+                //          be EncodedValue#getBits(), but maybe we need to reconsider the way the encoding manager or the
+                //          encoded values are serialized/deserialized.
+                em.edgeConfig.next(((IntEncodedValueImpl) encodedValue).bits);
+                if (encodedValue.isStoreTwoDirections())
+                    em.edgeConfig.next(((IntEncodedValueImpl) encodedValue).bits);
```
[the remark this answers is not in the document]
easbar: I did exactly this now -> much better I think.

core/src/main/java/com/graphhopper/routing/ev/IntEncodedValueImpl.java
easbar: They are included now, so now we store the complete state of the EM in the properties file and can later load it again :+1:

core/src/main/java/com/graphhopper/routing/ev/IntEncodedValueImpl.java
easbar: Actually I renamed it to `Constants.VERSION_EM` now and it represents the format of the stored EM in general, not just for encoded values.

core/src/main/java/com/graphhopper/routing/ev/SimpleBooleanEncodedValue.java
```diff
@@ -40,10 +40,16 @@ public SimpleBooleanEncodedValue(String name, boolean storeBothDirections) {
             @JsonProperty("minValue") int minValue,
             @JsonProperty("maxValue") int maxValue,
             @JsonProperty("negateReverseDirection") boolean negateReverseDirection,
-            @JsonProperty("storeTwoDirections") boolean storeTwoDirections
+            @JsonProperty("storeTwoDirections") boolean storeTwoDirections,
```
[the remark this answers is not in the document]
easbar: We already enforce (in IntEncodedImpl) that if negate reverse direction is used the min value must be zero (and store two directions must be false also).

core/src/main/java/com/graphhopper/routing/ev/IntEncodedValueImpl.java
easbar: getVersion() is gone

core/src/main/java/com/graphhopper/routing/ev/IntEncodedValueImpl.java
easbar: They are gone.

PR #2611, opened 2022-06-25
Remove FlagEncoder

### LINKED ISSUE #2463: Remove FlagEncoder

easbar:
Fixes to a large extent #2463.

The FlagEncoder interface is gone. But what previously was an implementation of it, the now very simple VehicleEncodedValues class, still remains. All it does now is grouping a few encoded values together, like car_access+car_average_speed.

So basically there is no such thing as a 'flag encoder' anymore. We already moved everything that was related to 'vehicle-specific' OSM parsing to the VehicleTagParsers in #2561. But with this PR the EncodingManager also no longer offers methods like `encodingManager.getEncoder("car")`. It only 'manages' encoded values now. So the only thing we can fetch from the EncodingManager are encoded values. The only thing that still remains is that we can add a group of certain encoded values to the encoding manager collectively, i.e. there still is `EncodingManager#add(VehicleEncodedValues.bike())` which, e.g. adds access, speed + priority encoded values for a bike. The reason I kept this here is that it corresponds to our config option `graph.flag_encoders` which defines multiple encoded values for each entry (and enables the corresponding vehicle parsers as well). I think we should also remove this, but since this requires some discussion about the new format better in a later PR. Actually I removed this pattern from almost all tests already and and it only remains on a high level / the GraphHopper class now.

The most important change here is that the fastest weighting (and others, also see here: 15426ef4588797f3c0031722ecb64297b98bfc1) no longer receives a FlagEncoder, but the encoded values directly. So for example this:

```java
FlagEncoder encoder = encodingManager.getEncoder("car");
Weighting weighting = new FastestWeighting(encoder);
```

is now this:

```java
BooleanEncodedValue accessEnc = encodingManager.getBooleanEncodedValue(VehicleAccess.key("car"));
DecimalEncodedValue speedEnc = encodingManager.getDecimalEncodedValue(VehicleSpeed.key("car"));
Weighting weighting = new FastestWeighting(accessEnc, speedEnc);
```

Importantly, the FastestWeighting also used `encoder.getMaxSpeed()` from the encoder, which we need to replace with `speedEnc.getMaxValue()`. This still remains todo before we can merge here. @karussell already did some work in this direction here: b3b3b11e35f1c9633d5c50a33c1b37ef3456ee73. Update: This was done in #2614, #2616, #2619 and the new method is now called: `getMaxOrMaxStorableDecimal()`.

The other thing that is still missing are the private/destination penalties in FastestWeighting, which we also configured via the FlagEncoder so far. They should probably be replaced with config options for the weighting, i.e. `weighting: fastest|private_penalty_factor=10`, see the todo comments in FastestWeighting. Update: This is done in DefaultWeightingFactory now.


core/src/test/java/com/graphhopper/routing/AlternativeRouteEdgeCHTest.java
```diff
@@ -41,8 +37,10 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class AlternativeRouteEdgeCHTest {
-    private final FlagEncoder carFE = FlagEncoders.createCar(new PMap().putObject("turn_costs", true));
-    private final EncodingManager em = EncodingManager.create(carFE);
+    private final BooleanEncodedValue accessEnc = new SimpleBooleanEncodedValue("access", true);
+    private final DecimalEncodedValue speedEnc = new DecimalEncodedValueImpl("speed", 5, 5, false);
+    private final DecimalEncodedValue turnCostEnc = TurnCost.create("car", 1);
+    private final EncodingManager em = EncodingManager.start().add(accessEnc).add(speedEnc).addTurnCostEncodedValue(turnCostEnc).build();
```
easbar, line 43: This is basically the new way to add 'vehicle' encoded values to the EncodingManager. We could do this using the `VehicleEncodedValues`, but I think they should rather be removed.

core/src/main/java/com/graphhopper/routing/DefaultWeightingFactory.java
```diff
@@ -70,26 +71,46 @@ public Weighting createWeighting(Profile profile, PMap requestHints, boolean dis
...
+                : null;
         if (CustomWeighting.NAME.equalsIgnoreCase(weightingStr)) {
             if (!(profile instanceof CustomProfile))
                 throw new IllegalArgumentException("custom weighting requires a CustomProfile but was profile=" + profile.getName());
             CustomModel queryCustomModel = requestHints.getObject(CustomModel.KEY, null);
             CustomProfile customProfile = (CustomProfile) profile;
             queryCustomModel = CustomModel.merge(customProfile.getCustomModel(), queryCustomModel);
-            weighting = CustomModelParser.createWeighting(encoder, encodingManager, turnCostProvider, queryCustomModel);
+            // todonow: this used to be encoder.getMaxSpeed()!
+            double maxSpeed = speedEnc.getMaxDecimal();
```
easbar, line 86: This is one place where we need the maximum actually used speed rather than the maximum storable speed.

core/src/main/java/com/graphhopper/routing/weighting/FastestWeighting.java
```diff
@@ -46,33 +47,38 @@ public class FastestWeighting extends AbstractWeighting {
...
+        double defaultPrivateFactor = 10;
         privatePenalty = checkBounds("road_access_private_factor", map.getDouble("road_access_private_factor", defaultPrivateFactor), 1, 10);
-        roadAccessEnc = destinationPenalty > 1 || privatePenalty > 1 ? encoder.getEnumEncodedValue(RoadAccess.KEY, RoadAccess.class) : null;
+        if (destinationPenalty > 1 || privatePenalty > 1) {
+            // todonow: maybe only error when factors are given *explicitly*
+//            if (roadAccessEnc == null)
+//                throw new IllegalArgumentException("road_access must not be null when destination or private penalties are > 1");
+            this.roadAccessEnc = roadAccessEnc;
+        } else
+            this.roadAccessEnc = null;
```
easbar, line 81: We need to fix these todos before we can merge.

core/src/main/java/com/graphhopper/routing/DefaultWeightingFactory.java
```diff
@@ -55,6 +57,15 @@ public Weighting createWeighting(Profile profile, PMap requestHints, boolean dis
         hints.putAll(requestHints);
 
         final String vehicle = profile.getVehicle();
+        // todonow: we used to take this information from the 'encoder', but now it is gone. We probably need to take
+        // it from the profile now somehow.
+        final boolean isMotorVehicle = true;
+        if (isMotorVehicle) {
```
easbar, line 63: Either put it into the weighting string like `weighting: fastest|private_factor=10` or into the profile like `weighting: fastest, private_factor: 10`. Or somehow map the vehicle/profile name to isMotorVehicle...

core/src/main/java/com/graphhopper/routing/weighting/FastestWeighting.java
```diff
@@ -46,33 +49,33 @@ public class FastestWeighting extends AbstractWeighting {
...
-
-        // ensure that we do not need to change getMinWeight, i.e. road_access_factor >= 1
-        double defaultDestinationFactor = encoder.isMotorVehicle() ? 10 : 1;
-        destinationPenalty = checkBounds("road_access_destination_factor", map.getDouble("road_access_destination_factor", defaultDestinationFactor), 1, 10);
-        double defaultPrivateFactor = encoder.isMotorVehicle() ? 10 : 1.2;
-        privatePenalty = checkBounds("road_access_private_factor", map.getDouble("road_access_private_factor", defaultPrivateFactor), 1, 10);
-        roadAccessEnc = destinationPenalty > 1 || privatePenalty > 1 ? encoder.getEnumEncodedValue(RoadAccess.KEY, RoadAccess.class) : null;
+        // todonow: take max speed from speedEnc
+//        maxSpeed = encoder.getMaxSpeed() / SPEED_CONV;
+        maxSpeed = 140 / SPEED_CONV;
```
easbar, line 66: This is another place where we need to use the speed encs actual max value.

core/src/test/java/com/graphhopper/routing/AlternativeRouteEdgeCHTest.java
easbar: Fixed in #2614

core/src/main/java/com/graphhopper/routing/DefaultWeightingFactory.java
easbar: This seems to be the last open question for me here.

easbar: Maybe the least invasive thing we can do is use a list of vehicle names that we consider `!isMotorVehicle` like `foot,hike,wheelchair,bike,mtb,racingbike,bike2`? This is a bit ugly, because the list needs to be adjusted if someone adds a new 'vehicle', but this requires adjusting the flag encoder factor also, anyway and we can leave a comment there.

otbutz: Would it be possible to add a new vehicle without forking GraphHopper then? With the old FlagEncoder approach this was still doable.

easbar: Yes, absolutely. This is just about the private/destination penalty in FastestWeighting. And even that can be adjusted by overwriting the DefaultWeightingFactory. Let alone that the private/destination penalty should be a thing of the past as we have much more possibilities using CustomWeighting.

... then again if you have a certain use-case in mind it would also help if you tried this branch to make sure I did not mess this up somehow :)

easbar: You can even change the private/destination penalty using the hints in the profile definition, i.e. in `config.yml`.

easbar:
> Can you update your comment to use the VehicleAccess and maybe also mention this in the changelog?

:+1: done :)

> So finally. Really nice. When will you livestream the merge ;)

3, 2, 1... few years later, and its gone!

PR #2612, opened 2022-06-26
Increase maximum edge-based CH key for preparation to 2^30 again

easbar:
Since #2567 we store the edge key instead of the edge ID in CHPreparationGraph#OrigGraph, which is used (only) for edge-based CH preparation. Before this the maximum possible edge ID was around 536mio (`Integer.MAX_VALUE >> 2`), but due to the change now the maximum possible edge **key** is around 536mio, which means that the import will fail if the base graph has more than around 268mio edges! 
However, we did not use the unsigned bitshift operator so far which means we were applying a limit that was below the physical limit implied by the 30bits we reserve for the edge ID/key. Using the unsigned shift the maximum is around 1 billion (`Integer.MAX_VALUE >> 1`) and since we are now storing edge keys this means the maximum number of edges is around `536mio` when we use the unsigned shift.

@karussell already fixed this here: 7bb8ff7d4581e8f34baf574fb470816f3adfcc64 (thanks!), but reverted to the signed shift here again: 47c57118423b57f302ab1a021b00b913f80e527a, which I think is wrong.

Here I first extracted the methods doing the bit operations and added a test that shows that with the unsigned shift the conversion goes wrong above 536mio: c4f5e923ab7010cf704256ccd95d9e4c748b3f57. Then I used the signed shift which fixes the failing test up to 1 billion: c34fc355de81ae2802bcf1302e61e9f398395ab2, and finally I inlined everything again and removed the test (or should we keep this?) for this PR.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ch/CHPreparationGraph.java
+++ b/core/src/main/java/com/graphhopper/routing/ch/CHPreparationGraph.java
@@ -166,31 +166,6 @@ public class CHPreparationGraph {
         };
     }
 
-    static int getKeyWithFlags(int key, boolean fwd, boolean bwd) {
-        // we use only 30 bits for the key and store two access flags along with the same int
-        if (key > Integer.MAX_VALUE >> 1)
-            throw new IllegalArgumentException("Maximum edge key exceeded: " + key);
-        key <<= 1;
-        if (fwd)
-            key++;
-        key <<= 1;
-        if (bwd)
-            key++;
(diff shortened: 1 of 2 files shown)


easbar:
Merging here as this is kind of urgent, but @karussell please feel free to comment or make further changes.

easbar:
I added a test: 3c2dfd1aedf8ba6881353456117ca5ec486f20a7

PR #2618, opened 2022-06-30
Valid if PathDetail is null

rafaelstelles:
Hi all,

I was getting the error, because detail.average_speed is null.

java.lang.IllegalArgumentException: Unsupported type of PathDetail value ARRAY
 at [Source: UNKNOWN; line: -1, column: -1]

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -4,11 +4,11 @@ Most of the contributors are mentioned at Github as [Members](https://github.com
 
 Here is an overview:
 
+ * agouge, discussion and API refactoring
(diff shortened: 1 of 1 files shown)


RafaelSTelles:
The written test fails if you don't have a fix.

The null value, I believe it is coming from OSM.

easbar:
The path detail seems to be `null` indeed:

https://graphhopper.com/maps/?point=-34.650384%2C-58.465985&point=-34.648084%2C-58.476658&locale=en-US&elevation=true&profile=car&use_miles=false&layer=Omniscale&details=average_speed

image

PR #2619, opened 2022-06-30
Remove FlagEncoder#getMaxSpeed()

easbar:
Since #2614 it was not really used in production code anymore, but here it is actually removed. All these changes were also part of #2611, but I thought it makes sense to do this separately here, so there is less to deal with in #2611

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/VehicleEncodedValues.java
+++ b/core/src/main/java/com/graphhopper/routing/util/VehicleEncodedValues.java
@@ -25,18 +25,12 @@ import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import static com.graphhopper.routing.util.BikeCommonTagParser.BIKE_MAX_SPEED;
-import static com.graphhopper.routing.util.CarTagParser.CAR_MAX_SPEED;
 import static com.graphhopper.routing.util.EncodingManager.getKey;
-import static com.graphhopper.routing.util.FootTagParser.FERRY_SPEED;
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/graphhopper/routing/util/VehicleEncodedValues.java
```diff
@@ -108,11 +100,10 @@ public static VehicleEncodedValues car(PMap properties) {
         double speedFactor = properties.getDouble("speed_factor", 5);
         boolean speedTwoDirections = properties.getBool("speed_two_directions", false);
         int maxTurnCosts = properties.getInt("max_turn_costs", properties.getBool("turn_costs", false) ? 1 : 0);
-        double maxSpeed = properties.getDouble("max_speed", CAR_MAX_SPEED);
         BooleanEncodedValue accessEnc = new SimpleBooleanEncodedValue(getKey(name, "access"), true);
         DecimalEncodedValue speedEnc = new DecimalEncodedValueImpl(getKey(name, "average_speed"), speedBits, speedFactor, speedTwoDirections);
         DecimalEncodedValue turnCostEnc = maxTurnCosts > 0 ? TurnCost.create(name, maxTurnCosts) : null;
-        return new VehicleEncodedValues(name, accessEnc, speedEnc, null, null, turnCostEnc, speedEnc.getNextStorableValue(maxSpeed), true, false);
+        return new VehicleEncodedValues(name, accessEnc, speedEnc, null, null, turnCostEnc, true, false);
```
easbar, line 106: The improvement here is that the 'maximum speed' no longer needs to be set for the FlagEncoder and the VehicleTagParser, but only the parser and then the maximum is taken from the speed EV (#2614)

PR #2623, opened 2022-07-06
fix value expression with encoded value that has defaultIsInfinity==true

karussell:
The maximum value of max_speed is Infinity due to defaultIsInfinity==true.

This special case is a bit ugly but we need this for edges without a (known) speed limit.

Or maybe we should move this information into getMaxValue and getMaxStorableDecimal?

### DIFF
--- a/client-hc/src/test/java/com/graphhopper/api/Examples.java
+++ b/client-hc/src/test/java/com/graphhopper/api/Examples.java
@@ -75,6 +75,10 @@ public class Examples {
         InstructionList il = res.getInstructions();
         for (Instruction i : il) {
             // System.out.println(i.getName());
+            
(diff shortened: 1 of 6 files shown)


easbar:
> Or maybe we should move this information into getMaxValue and getMaxStorableDecimal?

As long as we do not need it I think it is ok to just throw an error in getMaxValue if defaultIsInfinity=true.
For getMaxStorableDecimal it seems to make sense to return infinity in this case? In DecimalEncodedValue we promise this:

```java
    /**
     * The maximum double value this EncodedValue accepts for setDecimal without throwing an exception.
     */
    double getMaxStorableDecimal();
```

and `set(decEnc, Double.POSITIVE_INFINITY)` is actually allowed if defaultIsInfinity=true.

Issue #2572, opened 2022-05-04
Turn costs are not included in instruction times


easbar:
https://discuss.graphhopper.com/t/responsepath-time-not-equal-sum-instruction-times/7245/4?u=easbar

PR #2626, opened 2022-07-13
make global time&distance values consistent with path details and instructions

karussell:
See discussion https://discuss.graphhopper.com/t/7378

Currently we merge path details instead of adding them which avoids duplicates at the via point. But the problem with this is that for length dependent path details like "distance" or "time" we would have to add the values. The resulting code is not complicated but a bit ugly as we need to list these path details and implement different add methods per type (double, long, ...). I'm unsure why we should do this and have to hide the fact that path details can be duplicated at via points. Removing this simplifies the code and I had to adapt only a few tests.

Furthermore this PR fixes the missing addition of turn time in the instructions (if finite u-turn costs) and adds a test for this (see changes in testTurnCostsOnOff).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### 6.0 [not yet released]
+- path details at via-points are no longer separated, see #2626
 - removed the FlagEncoder interface. for example encoder.getAccessEnc() is now encodingManager.getBooleanEncodedValue(
   VehicleAccess.key("car")), #2611
 - backward incompatible change as instructions and the street_name path detail do no longer contain the ref #2598


web-api/src/main/java/com/graphhopper/ResponsePath.java
```diff
@@ -254,37 +254,15 @@ public void addPathDetails(Map<String, List<PathDetail>> details) {
...
-        // Make sure that the PathDetail list is merged correctly at via points
-        if (!pathDetails.isEmpty() && !otherDetails.isEmpty()) {
-            PathDetail lastDetail = pathDetails.get(pathDetails.size() - 1);
-            boolean extend = lastDetail.getValue() != null
-                    ? lastDetail.getValue().equals(otherDetails.get(0).getValue())
-                    : otherDetails.get(0).getValue() != null;
-            if (extend) {
-                lastDetail.setLast(otherDetails.get(0).getLast());
-                otherDetails.remove(0);
-            }
```
boldtrn, line 282: Here we could do something like:

```
boolean extend = lastDetail.extend(otherDetails.get(0)); // extend would be a new function
if (extend) {
   lastDetail.merge(otherDetails.get(0)); // merge would be a new function
   lastDetail.setLast(otherDetails.get(0).getLast());
   otherDetails.remove(0);
}
```

boldtrn:
> which path details are length-dependent (and maintain this list or create a new attribute) and
implement the merging action for every possible value like long & double with all possibilities for failure as we only have an Object value (and to avoid creating a result with a different type).

My implementation proposal to this would be to have a "merge" method in the PathDetail implementation. The default implementation, which would work for most PathDetails would simply merge them.

And for the two (or maybe three) relevant alternatives we can implement an addition implementation.

(See my proposal in the code comment).

>  It is even more precise without the merge as we can now calculate the distance & time per "route leg" only based on the path details (which isn't possible precisely before this PR). Probably this is the main argument :)

Right now, you would have to use the instructions. With this PR you would still have to use the instructions to find out the index of the waypoint. Then you could either sum up the instructions or the PathDetails. Not sure if this would be a big improvement :)?

boldtrn:
> I still do not see a big disadvantage. Can you try a typically route you have in mind and report the increase you see for compressed response? Also please let me know the total distance, number of via points and the path details. Because when I try a route from e.g. hoyerswerda to leipzig with 3 path details and 5 stops (without elevation) is 9.47KB where master was 9.44KB.

For Kurviger, we often see 50-100+ Waypoints. I haven't compared the sizes in detail though, overall this is probably not that much of a difference.

> What effort do you mean here?

Now the client needs to merge the PathDetails. Depending on how you show them.

> Yes, but I see this as a limitation that we have to fix and e.g. create a "leg" path detail. In fact we could even think about creating an "turn_instruction" path detail to avoid the separate "turn instruction" code.

I think I even created a PR a few years ago where I tried that. I don't remember the details, but there were a few issues with this. Maybe with the newer code structure it would work easier :)? This would be a great improvement though 👍.

I would prefer to merge the PathDetails, but I also understand your opinion on this to streamline the code :).

easbar:
This will fix #2572

boldtrn:
Ok :). 

Since this is a "breaking" change, can we mention it in the Changelog, so everyone is aware of the change?

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ### 6.0 [not yet released]
+- path details at via-points are no longer separated, see #2626
```
boldtrn, line 2: are no longer merged?

PR #2631, opened 2022-07-21
improvements for bike

karussell:
I noticed that a few changes are necessary:

 * paving_stones must be faster like 16km/h instead of 12 -> it is like paved
 * if hazmat=designated is bad! 

I found these two issues due to [this incorrect route](https://graphhopper.com/maps/?point=51.419089%2C14.227322&point=51.413929%2C14.230878&profile=bike) that uses an hazmat secondary road instead of the cyclepath.

From this I found also these improvements:

 * very bad smoothness: speed must be lower than "pushing speed" as for them I think pushing the bike will be slower than walking. I introduced min_speed (2km/h).
 * setSmoothnessSpeedFactor: use one map for all bikes and reduce factor for *horrible
  * steps must be min_speed too

Maybe you can have a quick look @ratrun ?

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/MountainBikeTagParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/MountainBikeTagParser.java
@@ -60,18 +60,9 @@ public class MountainBikeTagParser extends BikeCommonTagParser {
         setTrackTypeSpeed("grade4", 8);
         setTrackTypeSpeed("grade5", 6); // like sand/grass     
 
-        setSurfaceSpeed("paved", 18);
-        setSurfaceSpeed("asphalt", 18);
-        setSurfaceSpeed("cobblestone", 10);
-        setSurfaceSpeed("cobblestone:flattened", 10);
-        setSurfaceSpeed("sett", 10);
         setSurfaceSpeed("concrete", 14);
         setSurfaceSpeed("concrete:lanes", 16);
         setSurfaceSpeed("concrete:plates", 16);
-        setSurfaceSpeed("paving_stones", 16);
-        setSurfaceSpeed("paving_stones:30", 16);
-        setSurfaceSpeed("unpaved", 14);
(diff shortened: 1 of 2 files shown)


ratrun:
Thanks for pointing me to this PR.

Speed for paving_stones is not identical to paved speed, you are really a bit slower, so I would suggest 14 km/h as a compromise.

But the major change, which you haven't explicitly mentioned above, is the increase of the pushing section speed from 4 km/h to 6km/h. I would not do that because I don't think that you will be really that fast, especially pushing a loaded touring bicycle or a MTB through the forest.

The hazmat=designated downgrade might help also in other cases, so it is a good idea to include it into the priority weighting.

ratrun:
> The raise was more meant to have a more realistic speed of bikes on footways as the maximum we can allow is the legal limit of 6km/h, but most people will still drive the bike faster than 6km/h (hopefully only a bit).

The PUSHING_SECTION_SPEED is not used only for the routing on footways. It is used at all code locations where we assume that we really need to push the bicycle. And 6km/h is quite fast for walking. For this reason I would leave the default for this constant with 4km/h. Please also note that the surface speed is only taken in case that the computed speed is greater than PUSHING_SECTION_SPEED. ( See https://github.com/graphhopper/graphhopper/blob/a72c6c82718a4ce46cc006fc7edea52c089f2a64/core/src/main/java/com/graphhopper/routing/util/BikeCommonTagParser.java#L348 )

But I would agree with changing the default speed for a footway to 6km/h under the assumption that not everybody will push the bike on a footway.

ratrun:
> But unpaved also has 14 km/h and compacted and pebblestone even have 16km/h.

Good catch. Maybe set `unpaved` to 12 like `gravel` and `compacted` and `pebbelstone` also to 14? I just recently noted that `compacted` should probably be a bit slower.

ratrun:
Looks good! 😄

PR #2634, opened 2022-07-25
new edge smoothing: "ramer"

karussell:
The big problem of the current smoothing method (#1220) is that the result is optimized towards reducing the ascend&descend, but it makes the elevation data nearly unusable for routing (as the slope increases often extremely). This was also partially discussed in #1953 (see https://github.com/graphhopper/graphhopper/pull/1953#issuecomment-624580609).

So I thought about a smoothing that does not make the slope more extreme (slope can be positive or negative) and is still able to remove some small elevation changes. A simple idea is to use the Ramer-Douglas-Peucker algorithm and instead of removing the points that are below a maximum elevation change the smoothing algorithm sets them to the elevation resulting from the average slope. The result is quite good and the decrease of the ascend&descend still comes close to the old smoothing, sometimes even a bit smaller:

For [this route](https://graphhopper.com/maps/?point=49.967067%2C11.584911&point=49.970545%2C11.567595&profile=foot) (a slightly sloping forest path, so the less changes the better)

* no smoothing ↗37m ↘43m:
grafik

* old smoothing ↗33m ↘36m:
grafik

* new smoothing (max 5m) ↗32m ↘34m:
grafik

Or see [this route](https://graphhopper.com/maps/?point=50.916252%2C14.161785&point=50.890473%2C14.231415&profile=bike) which should be very smooth with a small ascend/descend still our raw CGIAR elevation data makes it hard for us due to the hills nearby:

* no smoothing ↗86m ↘92m:
grafik

* old smoothing ↗55m ↘61m:
grafik

* new smoothing ↗46m ↘52m:
grafik

See the config option:

```
graph.elevation.edge_smoothing: ramer
graph.elevation.edge_smoothing.ramer.max_elevation: 10
# old smoothing via:
# graph.elevation.edge_smoothing: window
# an exception is thrown if the the old parameter 'graph.elevation.smoothing' is used
```

One interesting property of this new method is that if you set the `graph.elevation.edge_smoothing.ramer.max_elevation` to a very big value it will set all pillar nodes to the elevation resulting from the average slope of the tower nodes leading to the most extreme form of elevation simplification we currently can do (without increasing maximum slope and without changing the elevation for tower nodes).

It would be nice to combine this with long edge sampling, but to do that we would have to swap the order of the calls. I'm unsure why the current order was picked - see discussion about this topic [here](https://github.com/graphhopper/graphhopper/pull/2222#issuecomment-749220133).

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -300,7 +300,7 @@ public class OSMReader {
             // smooth the elevation before calculating the distance because the distance will be incorrect if calculated afterwards
             if (config.getElevationSmoothing().equals("ramer"))
                 EdgeElevationSmoothing.smoothRamer(pointList, config.getElevationSmoothingRamerMax());
-            else if (config.getElevationSmoothing().equals("window"))
+            else if (config.getElevationSmoothing().equals("moving_average"))
                 EdgeElevationSmoothing.smoothMovingAverage(pointList);
         }
 


otbutz:
Quick comparison with Google Maps:

First [route](https://www.google.com/maps/dir/'49.967067,11.584911'/'49.970545,11.567595'/@49.9690784,11.5748548,16z/data=!4m11!4m10!1m3!2m2!1d11.584911!2d49.967067!1m3!2m2!1d11.567595!2d49.970545!3e2!5i1)  :arrow_upper_right: 14m :arrow_lower_right: 27m:

grafik

Tried to replicate the second route as good as possible :arrow_upper_right: 67m :arrow_lower_right: 57m: :

grafik

boldtrn:
Visually I think the new results look quite interesting.

I also think the property that it does not increase the slope is very interesting.

I am not sure if Douglas Peucker is removing too many details? If there is actually a very steep part on an edge DP might remove this steepness. If you think about bicycle routing, you really want to avoid very steep parts. 

On the other hand, DP keeps the extremes. Often these extremes, if they only occur for a very short part of the road, are data glitches and not actual changes on the road.

otbutz:
Ever tried Visvalingam or a simple moving average?

boldtrn:
> while still keeping outstanding elevation changes

I would argue, that this is a downside of RDP. Outstanding elevation changes have a high likelyhood of being data errors (e.g. short peaks or drops).

otbutz:
> somehow define a maximum slope change

That one might make sense. It's elevation data for routes after all. An unpassable incline should be filtered out.

PR #2648, opened 2022-08-22
consider multiple from members for no_entry

### LINKED ISSUE #2647: no_entry restriction not working correctly
**Describe the bug**
No_entry restricitons don't seem to work correctly. Maybe this is only the case when there is more than one "from", I am not sure.

**To Reproduce**
Try [this route](https://graphhopper.com/maps/?point=54.308155%2C10.100743&point=54.307548%2C10.10004&locale=de&elevation=true&profile=car&use_miles=false&selected_detail=Elevation&layer=Omniscale) and [this route](https://graphhopper.com/maps/?point=54.308036%2C10.100003&point=54.307548%2C10.10004&locale=de&elevation=true&profile=car&use_miles=false&selected_detail=Elevation&layer=Omniscale) as a test for [this restriction](https://www.openstreetmap.org/relation/12441440).

**Expected behavior**
The navigation should not enter Hasseer Straße, neither from Saarbrückenstraße nor from Altenrade.

**System Information**
Online version

**Screenshots & Logs**
grafik
grafik

If applicable, add screenshots and log output.

In case of a routing problem include a link to [GraphHopper Maps](https://graphhopper.com/maps/). Also tell us your expected routing result via additional markers and screenshots.

karussell:
Fixes #2647

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -576,10 +576,8 @@ public class OSMReader {
             }
         }
         if (relation.hasTag("restriction")) {
-            OSMTurnRelation osmTurnRelation = createTurnRelation(relation, relation.getTag("restriction"), vehicleTypeRestricted, vehicleTypesExcept);
-            if (osmTurnRelation != null) {
-                osmTurnRelations.add(osmTurnRelation);
-            }
+            osmTurnRelations.addAll(createTurnRelations(relation, relation.getTag("restriction"),
+                    vehicleTypeRestricted, vehicleTypesExcept));
             return osmTurnRelations;
         }
         if (relation.hasTagWithKeyPrefix("restriction:")) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -576,53 +576,49 @@ static List<OSMTurnRelation> createTurnRelations(ReaderRelation relation) {
...
+                else if (ReaderElement.NODE == member.getType() && "via".equals(member.getRole()))
                     viaNodeID = member.getRef();
-                }
             }
-            if (fromWayID >= 0 && toWayID >= 0 && viaNodeID >= 0) {
-                OSMTurnRelation osmTurnRelation = new OSMTurnRelation(fromWayID, viaNodeID, toWayID, type);
-                osmTurnRelation.setVehicleTypeRestricted(vehicleTypeRestricted);
-                osmTurnRelation.setVehicleTypesExcept(vehicleTypesExcept);
-                return osmTurnRelation;
+            if (toWayID >= 0 && viaNodeID >= 0) {
```
easbar, line 608: Not really related to this PR probably, but we just had this discussion in the forum: https://discuss.graphhopper.com/t/custom-turn-restriction-example/7453

Why do we check if `toWayID >= 0 && viaNodeID >= 0` and why do we expect negative OSM IDs here?

easbar: Ah, I see now... But this means if an OSM ID is actually negative the turn relation will be ignored without any warning or error (see forum discussion). It's like a bad way to check if there was a relation member with the from/via/to role.

easbar: Ok, then let's say it is not as it should be. I called it 'bad', because it only works when a certain assumption is fulfilled (all OSM IDs are >=0), and we have no way to control this assumption (because the IDs come from an external file), we do not check if our assumption is true, and there is not even a comment in the code that states that it only works under this assumption. It is just a bug waiting to happen, hidden in the code, and my guess is that when the code was written this case was just forgotten about. I would also use `null` here, and additionally I think we should throw an error early if an OSM ID is negative, or are negative OSM IDs a thing? It does not seem to be:

The OSM [wiki](https://wiki.openstreetmap.org/wiki/Node) says that the id is even >= 1 and

> Editors may temporarily save node ids as negative to denote ids that haven't yet been saved to the server.

PR #2651, opened 2022-08-23
replace car4wd with roads and a custom_model

karussell:
The `car4wd` vehicle is no longer necessary and we can recreate it using the `roads` vehicle and the custom_model shown in 
web/src/test/resources/com/graphhopper/application/resources/car4wd.yml

If we replace a different vehicle (like bike2, wheelchair or hike) we should include the custom_model used for the replacement in a test, but IMO this is not necessary here.

It's a bit sad that my nicely drawn icon image is disappearing ;)


easbar:
nice!

otbutz:
I just stumbled across this change and am a little confused about how we would handle multiple transportation modes? I can only configure a single vehicle named "roads" and specify its "transportation_mode". So I can not have a car4wd(CAR) and a truck(HGV) in the same installation?

otbutz:
Our current approach feels a bit off to be honest. The configuration calls it `graph.vehicles` and the code is dealing with `VehicleTagParsers` but in fact we're talking handling the concerns of different transportation modes. The actual vehicle has a transportation mode and configuration knobs and/or a custom model.

easbar:
> We could allow to create different instances using different name properties like we allow for car. 

Yes, that might help in the short term. Maybe even something like `graph.vehicles=hgv|all_roads=true,car|all_roads=true`. So all 'flag encoders' with `all_roads=true` would be "roads vehicles" and the first string (hgv/car) would be the 'name' of the encoder.

Then again, I think what we should actually do is split the vehicle tag parsers into access and speed tag parsers. In this case the "roads" vehicle would just be split into (configurable) access and speed parsers that fill associated encoded values. For example the `transportation_mode` parameter would be a parameter of the access tag parser and of course there could be multiple such tag parsers and the corresponding encoded values could be used by the different custom profiles.

> Yes, sure, it is not yet where we want it to be. It would be nice to get rid of VehicleTagParsers completely and configure the transportation mode directly in the profile.

I think the VehicleTagParsers should just be normal tag parsers. They are tag parsers already, but they still have some special meaning that we should get rid of. But I don't think there should be some special configuration options for the transportation mode  in the profile. As I said above, the transportation mode would just be some parameter of the vehicle tag parser (or better the access tag parser we split out of the vehicle tag parser) and in the profiles we can make use of it by using the corresponding access encoded value in a custom model.

The main question for me are:

* How are we going to configure tag parsers and encoded values (with parameters) in `config.yml`? So far we have `graph.encoded_values` and `graph.vehicles`, but these do not really correspond directly to the idea of tag parsers + encoded values and `graph.vehicles` is some kind of mixture of both. For example we could use `graph.encoded_values` to configure all encoded values (with number of bits and so on) and `graph.tag_parsers` to configure all the parsers (with parameters like transportation mode etc.).

* How are we going to setup the 'base' speed and access in custom profiles. So far we have the `vehicle` key which defines the base speed and access for a custom model. But I think this vehicle thing should be replaced by something more explicit like:

```json
{
 "priority": [
   {
     "if": "car_access == false",
     "multiply_by": 0 
  },
],
"speed": [
  {
    "if": "true",
    "limit_to": "car_average_speed"     
  }   
]
}
```

And maybe we should allow specifying a 'base' profile (which prepends the custom model rules of a base model) so this becomes a bit more concise.

Issue #1222, opened 2017-11-16
Turn restrictions don't work with negative ids


mmuelder:
Greetings!

Today I created a turn restriction using JOSM but GH (0.9) just ignored it. I double and triple checked everything, but the OSM tags where correct. After spending many hours I eventually found the cause: negative IDs. I'm not sure whether it was the node, ways or relation IDs. But after I replaced all negative IDs in the OSM files with positive ones, the restiction suddenly worked like a charm. I reproduced this many times. Existing restrictions (thus positive ids) work, newly created in Josm (thus negative ids) don't.

Unfortunately I have no Java experience so I can't provide any test code or the like, but I hope someone can confirm this issue.

Best
Marius

karussell:
Ah, ok. We had problems before with this https://github.com/graphhopper/graphhopper/issues/42 and should be fixed

mmuelder:
FYI:

Removing
`if (fromWayID >= 0 && toWayID >= 0 && viaNodeID >= 0) {`
from OSMReader.java:849

and

`&& wayId >= 0`
from OSMTurnRelation.java:92

makes it work. However, I have no idea what the collateral damage is ;-)

easbar:
We fixed this by not allowing negative OSM ids in the imported OSM file anymore. Negative ids are still used internally to create artificial edges that replace barrier nodes.

PR #2652, opened 2022-08-24
validate that ids are not negative

lukasalexanderweber:
closes https://github.com/graphhopper/graphhopper/issues/1222

with the discussions [here](https://discuss.graphhopper.com/t/custom-turn-restriction-example/7453) and [here](https://github.com/graphhopper/graphhopper/pull/2648/files#r951487484) it might be a good idea to restrict ids to be >0, which would also conform to the [OSM structure](https://wiki.openstreetmap.org/wiki/Node).

As we now specify explicitly that all ids are >0, the `toWayID >= 0 && viaNodeID >= 0` assumption in [OSMReader#createTurnRelation](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java#L608) is perfectly fine.

A try to create a graph with negative OSM ids would lead to a massage as

`Caused by: com.graphhopper.reader.InvalidIdException: Invalid Id: -138082; Ids must be bigger or equal to 1
`


otbutz:
See https://github.com/graphhopper/graphhopper/pull/2648/files#r951647819

negative OSM IDs are used to add custom nodes/ways/relations to OSM data: https://wiki.openstreetmap.org/wiki/Import/Software#Negative_IDs

We should probably follow osm2pgsqls approach and switch to multiple input files instead: https://github.com/openstreetmap/osm2pgsql/issues/1097

lukasalexanderweber:
TBH I'm not that familiar with GH to be the man with the experience to tell if the Turn Restriction case is the only case which might be affected from negative IDs. If it is, it should be enough to refactor [createTurnRelation()](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java#L599) 

However, till we know there are no side effects in working with negative IDs, we should not allow them.

If you use e.g. JOSM to make your custom nodes/ways/relations it's very easy to make the IDs positive using an editor before creating a graphhopper graph. However wasting multiple hours as I did or he https://github.com/graphhopper/graphhopper/issues/1222
> After spending many hours I eventually found the cause: negative IDs.

did should not be the case.

lukasalexanderweber:
> We should probably follow osm2pgsqls approach and switch to multiple input files instead: [openstreetmap/osm2pgsql#1097](https://github.com/openstreetmap/osm2pgsql/issues/1097)

If I understand correctly, they deprecated the use of negative IDs which is what I suggest in this PR?

otbutz:
So far we have three options:

### Allow negative IDs

aka the status quo.

Pro: no change from the user perspective
Contra: their might still be bugs somewhere which need  to be patched; dealing with negative IDs requires additional logic and is potentially error prone

### Prohibit negative IDs

Importing a file with negative IDs would fail and the user has to patch it with an external program.

Pro: easy
Contra: outcry from users which rely on the feature

### Prohibit negative IDs  and allow to import multiple OSM files instead

Pro: flexible
Contra: refactoring the import is still a big TODO and would be mandatory for this IMHO

> If I understand correctly, they deprecated the use of negative IDs which is what I suggest in this PR?

osm2pgsql went with option 3. You can e.g import the default planet.osm.pbf and craft an additional pbf with your own data.

lukasalexanderweber:
In option 3, are there negative IDs allowed in any of the OSM files?

otbutz:
https://github.com/openstreetmap/osm2pgsql/pull/1339

Explains how osm2pgsql handles it.

tl;dr all files must be sorted and have only positive IDs

lukasalexanderweber:
So basically for this PR it's only option 1 or option 2, is it?

> allow to import multiple OSM files instead

IMO this is a feature not related to this issue

otbutz:
> IMO this is a feature not related to this issue

Negative IDs are currently (abused) to import non-OSM data as well. If we remove this feature, we should be able to provide users with an alternative. If external tools are good enough to do this, I think that is fine.

lukasalexanderweber:
I think it's REALY EASY for users to assure that the data contains only valid ids.

I created [an external tool](https://github.com/lukasalexanderweber/replace-invalid-osm-ids) for standard osm files.

However it's REALY hard for developers of such a large system to test the edge cases of negative ids.

easbar:
> If you use e.g. JOSM to make your custom nodes/ways/relations it's very easy to make the IDs positive using an editor before creating a graphhopper graph.

@lukasalexanderweber 
Do you mean your replace-invalid-osm-ids script, or can JOSM also do this? And what about larger (pbf?) files? 

> Negative IDs are currently (abused) to import non-OSM data as well. If we remove this feature, we should be able to provide users with an alternative. 

@otbutz 
Is this really a feature people use? And why can they not simply use positive IDs for their non-OSM data? So far the only use-case I can think of is creating a graph using some UI tool like JOSM, which (unfortunately in this case) uses negative IDs.

lukasalexanderweber:
> Do you mean your replace-invalid-osm-ids script, or can JOSM also do this?

in my case I just replaced "-" with "" within an editor. But I noticed that this is not robust. replace-invalid-osm-ids, however, handles this correctly.

> And what about larger (pbf?) files?

[osmconvert](https://wiki.openstreetmap.org/wiki/Osmconvert) invalid.pbf -o=invalid.osm
python [replace_invalid_osm_ids.py](https://github.com/lukasalexanderweber/replace-invalid-osm-ids) invalid.osm valid.osm
[osmconvert](https://wiki.openstreetmap.org/wiki/Osmconvert) valid.osm -o=valid.pbf

easbar:
> [osmconvert](https://wiki.openstreetmap.org/wiki/Osmconvert) invalid.pbf -o=invalid.osm
python [replace_invalid_osm_ids.py](https://github.com/lukasalexanderweber/replace-invalid-osm-ids) invalid.osm valid.osm
[osmconvert](https://wiki.openstreetmap.org/wiki/Osmconvert) valid.osm -o=valid.pbf

Ok yes this works for small files, but for larger files it is very inefficient, because it uses a DOM parser to read the xml. Anyway, of course it is not a terribly hard task to replace the negative IDs, I was just wondering if there are already tools that do this.

otbutz:
> Is this really a feature people use?

¯\\\_(ツ)\_/¯ 

I'm just pointing out that someone might get mad at us for breaking their [workflow](https://xkcd.com/1172/).

lukasalexanderweber:
>  but for larger files it is very inefficient, because it uses a DOM parser to read the xml

Yeah I know. With the script I rather target users of JOSM which work on a local dataset. 

I just tested it with a real life junction and it worked well.

Processing a 250MB subset of a medium sized city (Karlsruhe) already needs 90s.

@otbutz wouldn't it be fun to find out what abnormalities people do with negative ids? I love the comic xD

easbar:
It seems like so far we only suspect negative IDs to be useful for something, so I think we should reject them until we have an actual use case for negative IDs that cannot be fixed by some easy preprocessing.

core/src/main/java/com/graphhopper/reader/InvalidIdException.java
```diff
@@ -0,0 +1,11 @@
+package com.graphhopper.reader;
+
+public class InvalidIdException extends RuntimeException{
```
easbar, line 3: Let's just use `IllegalArgumentException` here? We did not really start using a whole lot of different exceptions so far.

core/src/main/java/com/graphhopper/reader/ReaderElement.java
```diff
@@ -41,11 +41,18 @@ protected ReaderElement(long id, int type) {
...
     protected ReaderElement(long id, int type, Map<String, Object> properties) {
+    	validateId(id);
         this.id = id;
         this.type = type;
         this.properties = properties;
     }
 
+    private void validateId(long id) {
+    	if (id < 1) {
+    		throw new InvalidIdException("Invalid Id: " + id + "; Ids must be bigger or equal to 1");
```
easbar, line 52: Can you change this to `OSM Id`, and maybe also include the `type` to make it easier to figure out which element had the negative ID? If you ask me also just inline this method (put the check right into the constructor).

CONTRIBUTORS.md
```diff
@@ -81,6 +81,7 @@ Here is an overview:
  * thehereward, code cleanups like #620
  * vvikas, ideas for many to many improvements and #616
  * zstadler, multiple fixes and car4wd
+ * lukasalexanderweber, fixes
```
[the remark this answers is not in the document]
easbar: and please add a line to changelog.md

easbar (changes requested):
Thanks for this. Currently there are still failing tests (e.g. run `mvn test`). Can you check what we would have to do to make them pass? There might be tests using negative OSM Ids (or just 0?). And even some xml files created by JOSM that use negative IDs.

core/src/main/java/com/graphhopper/reader/ReaderElement.java
lukasalexanderweber: adding `type` into the exeption message would only print an Integer. Do you thinks its worth adding a map from e.g. 0 to node and 1 to edge?

lukasalexanderweber:
there is a realy big bunch of test data which uses 0 as ID. We could maybe limit this PR only to negative Ids?

core/src/main/java/com/graphhopper/reader/ReaderElement.java
easbar: Can we change `type` from `int` to an enum like `ReaderElementType`? I've been wondering why it is an int anyway.

easbar:
> there is a realy big bunch of test data which uses 0 as ID. We could maybe limit this PR only to negative Ids?

Yes, I thought the same.

core/src/main/java/com/graphhopper/reader/ReaderElement.java
otbutz: A simple switch should be enough. No need for a map.

(The type should be an enum anyway)

core/src/test/java/com/graphhopper/reader/osm/OSMReaderTest.java
```diff
@@ -318,26 +318,9 @@ public void testFoot() {
...
-        assertTrue(iter.next());
-        assertEquals(n30, iter.getAdjNode());
-        assertEquals(93147, iter.getDistance(), 1);
-
-        assertTrue(iter.next());
-        assertEquals(n10, iter.getAdjNode());
-        assertEquals(88643, iter.getDistance(), 1);
+	    assertThrows(RuntimeException.class, () -> {
+	    	new GraphHopperFacade(fileNegIds).importOrLoad();
+	    });
```
easbar, line 323: Can we assert the error message here so we can be sure the exception was actually thrown because of the negative ID and not for some other reason (like file not found etc.)

core/src/main/java/com/graphhopper/reader/ReaderElement.java
lukasalexanderweber: something like this?

```
public abstract class ReaderElement {
	public static enum ReaderElementType {
		NODE,
		WAY,
		RELATION,
		FILEHEADER;
	}
    private final long id;
    private final ReaderElementType type;
    private final Map<String, Object> properties;

    protected ReaderElement(long id, ReaderElementType type) {
        this(id, type, new HashMap<>(4));
    }

    protected ReaderElement(long id, ReaderElementType type, Map<String, Object> properties) {
    	if (id < 0) {
    		throw new IllegalArgumentException("Invalid OSM Id: " + id + "; Ids must not be negative");
    	}
        this.id = id;
        this.type = type;
        this.properties = properties;
    }
```

otbutz: Please also add the type to the constructed exception message

lukasalexanderweber:
> [lukasalexanderweber](https://github.com/lukasalexanderweber) requested review from [easbar](https://github.com/easbar) and removed request for [otbutz](https://github.com/otbutz) [19 seconds ago](https://github.com/graphhopper/graphhopper/pull/2652#event-7278848255)

sorry for that!

otbutz (approved):
LGTM

core/src/main/java/com/graphhopper/reader/osm/OSMXMLHelper.java
```diff
@@ -105,7 +104,13 @@ private static void readMembers(ReaderRelation rel, XMLStreamReader parser) thro
...
     public static Member createMember(XMLStreamReader parser) {
         String typeName = parser.getAttributeValue(null, "type");
-        int type = TYPE_DECODE.indexOf(typeName.charAt(0));
+        ReaderElement.Type type = ReaderElement.Type.NODE;
+        if (typeName.startsWith("w")) {
+        	type = ReaderElement.Type.WAY;
+        }
+        else if (typeName.startsWith("r")) {
+        	type = ReaderElement.Type.RELATION;
+        }
```
easbar, line 113: This change might affect performance, but I doubt it. If it did we will notice in our performance metrics and can still adjust after merging this. Please also note that the code formatting seems off (we use the default formatting of the IntelliJ IDE). But I'll just fix this in master.

easbar:
Thanks!

core/src/main/java/com/graphhopper/reader/osm/OSMXMLHelper.java
otbutz: Anything even remotely performance critical will not use XML anyway :wink:

easbar: Ah, I see this is only used for xml.

lukasalexanderweber:
Thanks for the constructive Feedback and for maintaining Graphhopper!

PR #2654, opened 2022-08-27
Maps: Add slope detail to elevation diagram

easbar:
I added two elevation diagram details: 'Slope' shows the slope of every path segment (between pillar nodes) and 'Towerslope' shows the slope between tower nodes only. Both are calculated on the client-side from the point list. 'Slope' is always shown and 'Towerslope' is only shown when the `edge_id` detail is enabled (because this is needed to calculate 'Towerslope' and both can be considered debug features). For 'Towerslope' I also adjusted the elevations (skipped those of the tower nodes so the diagram matches the slope values):

image
image

The colors indicate the absolute slope value (e.g. -5 is the same color as +5) and the spectrum ranges from 0 (blue) to 25 (red).

### DIFF
--- a/web-bundle/src/main/resources/com/graphhopper/maps/js/map.js
+++ b/web-bundle/src/main/resources/com/graphhopper/maps/js/map.js
@@ -304,7 +304,6 @@ module.exports.addElevation = function (geoJsonFeature, details, selectedDetail,
             selectedDetailIdx = detailIdx;
         GHFeatureCollection.push(sliceFeatureCollection(details[detailKey], detailKey, geoJsonFeature));
         options.mappings[detailKey] = getColorMapping(details[detailKey]);
-
     }
 
     // always show elevation and slope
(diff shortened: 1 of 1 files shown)


easbar:
> Except that Towerslope and Slope are "100% raw" and do not contain the minor smoothing or the "small distance" rejection.
> But in the future, with better elevation, this is hopefully no longer necessary for max_slope and average_slope too. But having these raw values for now can be indeed valuable for debugging.

Yes, that is exactly the idea here. The two details I added here just show the data as it is (except the path simplification which can still change the results). And they also stay the same when the other two details (which are derived to some extent) are changed or are not even present (for example while refactoring something). 

> And the same for max_slope and Slope.

No, max_slope only shows the maximum slope of an edge's geometry, while Slope shows the slope for every path segment, a level of detail we cannot achieve by using a simple encoded value (one number) per edge.

PR #2658, opened 2022-09-14
read from GeoJSON export of boundaries.osm

### LINKED ISSUE #2394: Can we improve the country borders file?
Not sure if the current file even needs improvement, but this topic came up here:

https://github.com/graphhopper/graphhopper/issues/2392#issuecomment-925690167

karussell:
Fixes #2394. This seems to work (mostly) and the higher precision is indeed visible when routing and the file is even smaller :)

~But there is [a problem in the boundaries.osm file](https://josm.openstreetmap.de/ticket/22372). The problem might be that the type of the relation member is `'way'` but should be `'relation'` to create separate Polygons inside the MultiPolygon. This leads to an incorrect handling downstream in our AreaIndex and country lookup because it thinks that there is just a single Polygon with a shell and several holes. (I think I can exclude a bug in our deserializer because when rearranging the coordinates of the MultiPolygon it reads it as expected)~

to do:

 - [x] fix MultiPolgons
 - [x] remove the original countries_josm.geojson file from the classpath

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ev/Country.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/Country.java
@@ -244,19 +244,19 @@ public enum Country {
 
     public static final String KEY = "country", ISO_ALPHA3 = "ISO3166-1:alpha3";
 
-    private final String displayName;
+    private final String countryName;
     private final String alpha2;
 
-    Country(String displayName, String alpha2) {
-        this.displayName = displayName;
+    Country(String countryName, String alpha2) {
+        this.countryName = countryName;
         this.alpha2 = alpha2;
     }
 
     /**
-     * @return the name:en field of this country
+     * @return the name of this country. Avoids clash with name() method of this enum.
      */
-    public String getDisplayName() {
-        return displayName;
+    public String getCountryName() {
+        return countryName;
     }
 
     /**


otbutz:
A bit offtopic but the file seems to cause all kinds of weird issues in Githubs GeoJSON viewer :sweat_smile: 

https://github.com/graphhopper/graphhopper/blob/d6cb0288d68fe126267104c23ff561ba8e91cb80/core/src/main/resources/com/graphhopper/countries/countries.geojson

core/src/main/java/com/graphhopper/routing/ev/Country.java
```diff
@@ -269,4 +272,14 @@ public String getTwoLetterCode() {
     public static EnumEncodedValue<Country> create() {
         return new EnumEncodedValue<>(Country.KEY, Country.class);
     }
+
+    private static final Map<String, Country> map = new ConcurrentHashMap<>();
```
otbutz, line 276: We only have read access. A regular HashMap should suffice.

core/src/main/java/com/graphhopper/routing/ev/Country.java
```diff
@@ -269,4 +272,14 @@ public String getTwoLetterCode() {
...
+
+    static {
+        for (Country c : Country.values()) map.put(c.getTwoLetterCode(), c);
```
otbutz, line 279: We could also add case-insensitive lookups while we're at it

```suggestion
        for (Country c : Country.values()) {
            map.put(c.getTwoLetterCode(), c);
            map.put(c.getTwoLetterCode().toLowerCase(Locale.ROOT), c);
        }
```

otbutz:
@westnordost this could also be interesting for you, since you mentioned that you [contributed](https://github.com/graphhopper/graphhopper/issues/2392#issuecomment-925691299) upstream

otbutz:
@karussell should we run the converter as part of the maven build using [exec:java](https://www.mojohaus.org/exec-maven-plugin/java-mojo.html)?

westnordost:
Thanks for the ping. However, I am not sure what is reported in https://josm.openstreetmap.de/ticket/22372
Is it a bug in the JOSM export as GeoJson feature or in the actual boundaries.osm? Possibly, it is not even a bug in either of the two but the export as GeoJson feature simply doesn't work (as expected) for the boundaries.osm specifically because the latter does not consist of "normal" OSM data. I.e. JOSM doesn't know whether to export the closed ways that make up the boundary as a polyline or as a polygon because there is no `area=yes` or similar tagged on them.

westnordost:
So maybe you want to read the *.osm file directly. This should work smoother as part of the build process too, because you do not have the manual step of opening that file in JOSM and exporting it to geojson but can directly pull the file from the JOSM repo.

Here is the Java code I use this, it uses JTS too:
https://github.com/westnordost/countryboundaries/blob/master/generator/src/main/java/de/westnordost/countryboundaries/JosmCountryBoundariesReader.java

If you like to use that, you can use it under the terms of the Apache license 2.0.

core/src/main/java/com/graphhopper/routing/ev/Country.java
```diff
@@ -279,7 +279,11 @@ public static EnumEncodedValue<Country> create() {
         for (Country c : Country.values()) map.put(c.getTwoLetterCode(), c);
     }
 
+    /**
+     * This method returns a Country enum for the specified ISO3166-1:alpha2 code.
+     */
     public static Country valueOfAlpha2(String alpha2) {
-        return alpha2 == null ? null : map.get(alpha2);
+        if (alpha2 == null || alpha2.length() != 2) return null;
+        return map.get(alpha2);
```
otbutz, line 287: I don't see why we shouldn't just add them. I would craft a PR if you don't have any objections.

otbutz: I am aware of that. But it might be good to offer alpha3 codes for those who integrate GraphHopper as a library. Otherwise they need additional dependencies like [nv-i18n](https://github.com/TakahikoKawasaki/nv-i18n/blob/master/src/main/java/com/neovisionaries/i18n/CountryCode.java) to convert back and forth from two- to three-letter codes.

otbutz: My usecase would rather be something like "Which countries is route XY passing?". If I need three letter codes an additional library is required to translate the two-letter code.

otbutz: Nevermind. I'm blind. We already support both of them.

core/src/main/java/com/graphhopper/routing/ev/Country.java
```diff
@@ -242,7 +245,7 @@ public enum Country {
     ZMB("Zambia", "ZM"),
     ZWE("Zimbabwe", "ZW");
 
-    public static final String KEY = "country";
+    public static final String KEY = "country", JSON_ID = "id";
```
easbar, line 248: We now use two different names for the same thing? We have a `twoLetterCode` field in `Country` (and the corresponding `getTwoLetterCode`. And then there is `Country#valueOfAlpha2` which expects as input such a `twoLetterCode`.

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -252,12 +252,12 @@ protected void setArtificialWayTags(PointList pointList, ReaderWay way, double d
         // special handling for countries: since they are built-in with GraphHopper they are always fed to the EncodingManager
         Country country = Country.MISSING;
         for (CustomArea customArea : customAreas) {
-            Object countryCode = customArea.getProperties().get("ISO3166-1:alpha3");
-            if (countryCode == null)
+            Object id = customArea.getProperties().get(Country.JSON_ID);
```
easbar, line 255: I think this is problematic, because it means that we consider every custom area with a property key `Country.JSON_ID` to be a country. And this likely happens, because the value of `Country.JSON_ID` is just "id". So far we identified countries among the custom areas by the existence of the (rather special) `"ISO3166-1:alpha3"` property.

Is it possible to rename the `id` property in the new country file?

core/src/main/java/com/graphhopper/routing/ev/Country.java
```diff
@@ -269,4 +272,18 @@ public String getTwoLetterCode() {
...
+
+    static {
+        for (Country c : Country.values()) map.put(c.getTwoLetterCode(), c);
+    }
+
+    /**
+     * This method returns a Country enum for the specified ISO3166-1:alpha2 code.
+     */
+    public static Country valueOfAlpha2(String alpha2) {
+        if (alpha2 == null || alpha2.length() != 2) return null;
```
easbar, line 286: maybe remove both these two checks?

core/src/main/java/com/graphhopper/routing/util/CustomArea.java
```diff
@@ -35,7 +34,7 @@ public static CustomArea fromJsonFeature(JsonFeature j) {
         for (int i = 0; i < j.getGeometry().getNumGeometries(); i++) {
             Geometry geometry = j.getGeometry().getGeometryN(i);
             if (geometry instanceof Polygon) {
-                PolygonExtracter.getPolygons(geometry, borders);
+                borders.add((Polygon) geometry);
```
easbar, line 37: Just out of curiosity, can you explain why this changed?

core/src/main/java/com/graphhopper/routing/util/CustomArea.java
```diff
@@ -18,15 +18,14 @@
...
-import org.locationtech.jts.geom.Polygon;
-import org.locationtech.jts.geom.util.PolygonExtracter;
+import org.locationtech.jts.geom.*;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
 public class CustomArea implements AreaIndex.Area {
+    private final static GeometryFactory factory = new GeometryFactory();
```
easbar, line 28: Seems to be unused now.

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: .. or alternatively set the alpha2/twoLetter property in `GHUtility#readCountries` and use this here. Currently we set the `"name"` property there (but do not use it) (?)

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -102,7 +103,7 @@ public double getMaxLat() {
     public double getMaxLon() {
         return envelope.getMaxX();
     }
-    
+
```
easbar, line 106: Formatting seems off.

core/src/main/java/com/graphhopper/util/shapes/Polygon.java
```diff
@@ -21,6 +21,7 @@
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.Envelope;
 import org.locationtech.jts.geom.GeometryFactory;
+import org.locationtech.jts.geom.LineString;
```
easbar, line 24: Unused import

core/src/test/java/com/graphhopper/routing/util/AreaIndexTest.java
```diff
@@ -141,11 +135,13 @@ public void testOverlap() {
     public void testCountries() {
         AreaIndex<CustomArea> countryIndex = createCountryIndex();
         // Berlin
-        assertEquals("Germany", countryIndex.query(52.5243700, 13.4105300).get(0).getProperties().get("name:en"));
+        assertEquals(Locale.GERMANY.getDisplayCountry(), countryIndex.query(52.5243700, 13.4105300).get(0).getProperties().get("name"));
```
easbar, line 138: Why use the `Locale` constants here?

easbar (commented):
Thanks, great news you were able to improve the country border resolution! I took a look at the code only, it seems like you already tested the quality of the new country file.

core/src/main/java/com/graphhopper/routing/util/CustomArea.java
easbar: Ah yeah. For some reason I thought this was related to the no longer needed GeometryFactory. And yes, definitely more readable :)

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -252,12 +252,12 @@ protected void setArtificialWayTags(PointList pointList, ReaderWay way, double d
...
-            Object countryCode = customArea.getProperties().get("ISO3166-1:alpha3");
-            if (countryCode == null)
+            Object area3 = customArea.getProperties().get(Country.JSON_AREA3);
+            if (area3 == null)
                 continue;
             if (country != Country.MISSING)
-                LOGGER.warn("Multiple countries found for way {}: {}, {}", way.getId(), country, countryCode);
-            country = Country.valueOf(countryCode.toString());
+                LOGGER.warn("Multiple countries found for way {}: {}, {}", way.getId(), country, area3);
+            country = Country.valueOf((String) area3);
```
easbar, line 260: Now we have alpha2 and area3 :D Maybe rename the constant to `Country.ALPHA3` or `Country.ISO_ALPHA3`? Its value is no longer the name of the property that is used for the two letter country code in the country json file, so the `JSON_` prefix does not really make sense?

core/src/main/java/com/graphhopper/routing/ev/Country.java
```diff
@@ -242,28 +242,28 @@ public enum Country {
     ZMB("Zambia", "ZM"),
     ZWE("Zimbabwe", "ZW");
 
-    public static final String KEY = "country";
+    public static final String KEY = "country", JSON_AREA3 = "ISO3166-1:alpha3";
 
-    private final String name;
-    private final String twoLetterCode;
+    private final String displayName;
```
easbar, line 247: `displayName` sounds like you need to distinguish the `name` and a `displayName`? But there is no `name`? Ok, maybe this String is the displayName of the country code rather than the name of the country, but then the enum should be called `CountryCode`... Anyway, my vote goes for `Country` and `name` :)

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
easbar: Yes, thanks. See my comments inline. And it is certainly better to use a distinguished string like `ISO3166-1:alpha3` rather than something generic as `id` to identify the countries among all the custom areas .

core/src/main/java/com/graphhopper/routing/ev/Country.java
easbar: Ah, I see. Maybe `getCountryName()`?

easbar:
Nice, thanks!

PR #2661, opened 2022-09-22
use same names for the keys in the EdgeKVStorage

karussell:
The same that we use for path details.

Two decisions I had to make:

 * the internal classes are not available for client-hc code (or our Example), so we now have two classes EdgeKVStorage.KeyValue and Parameter.Details. As they have a different use case I find this acceptable.
 * the related properties in the instructions change due to this change but not the "name" as it was `street_name` already before. So it is a breaking change for the client side (but only if the in ~5.0~ 6.0 newly introduced properties like `ref` or `destination` are used) but an improvement as it makes it more consistent (like `street_ref` instead of `ref`). Note: the path details are not affected by this.

### DIFF
--- a/core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
+++ b/core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
@@ -429,15 +429,12 @@ public class EdgeKVStorage {
         return vals.getCapacity() + keys.getCapacity();
     }
 
-    // TODO move into KeyValue class?
-    public static class Details {
+    public static class KeyValue {
         public static final String STREET_NAME = "street_name";
         public static final String STREET_REF = "street_ref";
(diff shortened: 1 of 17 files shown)


core/src/main/java/com/graphhopper/search/EdgeKVStorage.java
```diff
@@ -429,6 +429,14 @@ public long getCapacity() {
         return vals.getCapacity() + keys.getCapacity();
     }
 
+    // TODO move into KeyValue class?
```
easbar, line 432: Yes, I think so, too.

easbar:
Nice, thanks! :+1:

boldtrn:
> but only if the in 5.0 newly introduced properties like ref or destination are used

These were introduced in 6.0, right?

PR #2665, opened 2022-09-29
Move curvature EncodedValue out of MotorcycleTagParser

karussell:
I kept the meaning of the curvature value but instead of `(beeline/edge_distance)²` it is now only `beeline/edge_distance` as it did not make a big difference (sure, the small deviations from 1 will be "earlier snap to 0.95", but in practise this didn't matter in my simple tests and having a simpler formula was preferable).

If you want to use `curvature` to prefer curves then do not overdo it and something like this is sufficient:
```json
{ "priority": [{
   "if": "curvature > 0.9",
   "multiply_by": "0.3"
}]}
```

Otherwise, without proper turn costs, you'll see things like:

grafik

I tried to calculate some kind of a `max_curve` value instead as the curvature was sometimes a bit too similar to 1 and used the maximum orientation change of the PointList like:
```java
for (int i = 1; i < pointList.size(); i++) {
    double orientation = ANGLE_CALC.calcOrientation(pointList.getLat(i - 1), pointList.getLon(i - 1),
            pointList.getLat(i), pointList.getLon(i), false);
    if (i > 1) {
        orientation = ANGLE_CALC.alignOrientation(prevOrientation, orientation);
        maxCurve = Math.max(maxCurve, Math.abs(orientation - prevOrientation));
    }
    prevOrientation = orientation;
}
maxCurveEnc.setDecimal(false, edgeFlags, Math.max(maxCurveEnc.getMinStorableDecimal(), Math.min(maxCurveEnc.getMaxStorableDecimal(),
       Math.toDegrees(maxCurve))));
```

But at the end it did not properly show when a road was curvy (even sometimes not for serpentines). Probably because a curvy road in OSM is often mapped smooth leading to relative small orientation changes. Of course this could happen with the curvature too if multiple ways are created but often this is not the case and of course we would need another feature like "turning angle" to complement this. A good test route for this was [here](https://graphhopper.com/maps/?point=47.566653%2C11.096492&point=47.559384%2C11.132326).

### DIFF
--- a/core/src/test/java/com/graphhopper/routing/RoutingAlgorithmWithOSMTest.java
+++ b/core/src/test/java/com/graphhopper/routing/RoutingAlgorithmWithOSMTest.java
@@ -122,7 +122,7 @@ public class RoutingAlgorithmWithOSMTest {
     @Test
     public void testMonacoMotorcycleCurvature() {
         List<Query> queries = new ArrayList<>();
-        queries.add(new Query(43.730729, 7.42135, 43.727697, 7.419199, 2681, 118));
+        queries.add(new Query(43.730729, 7.42135, 43.727697, 7.419199, 2681, 119));
         queries.add(new Query(43.727687, 7.418737, 43.74958, 7.436566, 3727, 170));
         queries.add(new Query(43.728677, 7.41016, 43.739213, 7.4277, 3157, 165));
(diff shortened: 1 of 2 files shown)


easbar:
Nice! I did not look at the exact modeling of the curvature EV, but the code changes LGTM.

PR #2667, opened 2022-10-04
deprecate graph.flag_encoders and use custom weighting in example config

karussell:
The option flag_encoders is now a bit confusing in the config-example.yml

### DIFF
--- a/config-example.yml
+++ b/config-example.yml
@@ -8,9 +8,8 @@ graphhopper:
 
   ##### Routing Profiles ####
 
-  # Routing can be done only for profiles listed below. A car profile is enabled per default and can be changed with
-  # tweaking the custom_model. For more information about profiles and custom profiles have a look into the
-  # documentation at docs/core/profiles.md or the examples under web/src/test/java/com/graphhopper/application/resources/
+  # Routing can be done only for profiles listed below. For more information about profiles and custom profiles have a
+  # look into the documentation at docs/core/profiles.md or the examples under web/src/test/java/com/graphhopper/application/resources/
   # or the CustomWeighting class for the raw details.
   #
   # In general a profile consists of the following
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -533,8 +533,11 @@ public class GraphHopper {
         setProfiles(ghConfig.getProfiles());
         vehiclesString = ghConfig.getString("graph.vehicles", vehiclesString);
         if (vehiclesString.isEmpty()) {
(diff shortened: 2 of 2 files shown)


easbar:
Yeah, it is confusing it is still called `graph.flag_encoders`, but since this is a breaking change I thought we wait with the renaming until we have something we actually want. We don't want `graph.vehicles` either... But since we are working on 7.0-SNAPSHOT now, maybe doing the breaking change now and doing it again later is ok. Especially since `graph.flag_encoders` is only needed for some advanced use-cases anymore and most configs will work by only specifying the `profiles` and skipping `graph.flag_encoders` altogether.

easbar:
So now we still have the thing we don't want but two names for it, but I agree it is less confusing with the new name.  I'm happy as long as we can remove this eventually.

config-example.yml
```diff
@@ -6,30 +6,18 @@ graphhopper:
...
-  # More are: surface,smoothness,max_width,max_height,max_weight,hgv,max_axle_load,max_length,hazmat,hazmat_tunnel,hazmat_water,toll,track_type,
-  #           mtb_rating, hike_rating,horse_rating,lanes
-  # graph.encoded_values: surface,toll,track_type
-
   ##### Routing Profiles ####
 
-  # Routing can be done for the following list of profiles. Note that it is required to specify all the profiles you
-  # would like to use here. The fields of each profile are as follows:
+  # Routing can be done only for profiles listed below. A car profile is enabled per default and can be changed with
+  # tweaking the custom_model. For more information about profiles and custom profiles have a look into the
```
easbar, line 12: This sounds like there is some special default car profile that is always enabled besides the ones in the list. I would just leave out the sentence "A car profile...". What is enabled and what is not can easily be seen by looking at the list.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -531,7 +531,12 @@ public GraphHopper init(GraphHopperConfig ghConfig) {
             throw new IllegalArgumentException("spatial_rules.max_bbox has been deprecated. There is no replacement, all custom areas will be considered.");
 
         setProfiles(ghConfig.getProfiles());
-        flagEncodersString = ghConfig.getString("graph.flag_encoders", flagEncodersString);
+        vehiclesString = ghConfig.getString("graph.vehicles", vehiclesString);
+        if (vehiclesString.isEmpty()) {
+            logger.warn("The option graph.flag_encoders is deprecated and will be removed. Replace with graph.vehicles");
```
easbar, line 536: I assume we want to print the warning when `graph.flag_encoders` is used and not if `graph.vehicles` is **not** used? Actually not using either is what we recommend and in this case of course there should be no warning. And maybe throw an error if **both** are used?

PR #2668, opened 2022-10-04
remove Bike2WeightTagParser

karussell:
This avoids the extra hassle of the parser and it is now recommended to use the bike vehicle with a custom model that uses slope_average - see the new `custom_models` folder.

Additionally this results in slightly better routing it seems as e.g. [this way](https://graphhopper.com/maps/?point=49.979322%2C11.563389&point=49.983082%2C11.559398&profile=bike) is avoided with this PR (due to horrible smoothness this is nice).

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -654,9 +654,12 @@ public class GraphHopper {
             osmParsers.addWayTagParser(new OSMMaxSpeedParser(encodingManager.getDecimalEncodedValue(MaxSpeed.KEY)));
         if (!encodedValueStrings.contains(RoadAccess.KEY))
             osmParsers.addWayTagParser(new OSMRoadAccessParser(encodingManager.getEnumEncodedValue(RoadAccess.KEY, RoadAccess.class), OSMRoadAccessParser.toOSMRestrictions(TransportationMode.CAR)));
-        if (encodingManager.hasEncodedValue(AverageSlope.KEY) || encodingManager.hasEncodedValue(MaxSlope.KEY))
-            osmParsers.addWayTagParser(new SlopeCalculator(encodingManager.hasEncodedValue(MaxSlope.KEY) ? encodingManager.getDecimalEncodedValue(MaxSlope.KEY) : null,
-                    encodingManager.hasEncodedValue(AverageSlope.KEY) ? encodingManager.getDecimalEncodedValue(AverageSlope.KEY) : null));
+        if (encodingManager.hasEncodedValue(AverageSlope.KEY) || encodingManager.hasEncodedValue(MaxSlope.KEY)) {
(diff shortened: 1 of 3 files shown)


easbar:
Oh, I was just about to remove `applyWayTags` here: https://github.com/graphhopper/graphhopper/tree/remove_apply_way_tags
It is almost complete, but if you also want to remove hike and wheelchair it would be even easier :)

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -652,7 +652,8 @@ private void buildEncodingManagerAndOSMParsers(String flagEncodersStr, String en
         if (!encodedValueStrings.contains(RoadAccess.KEY))
             osmParsers.addWayTagParser(new OSMRoadAccessParser(encodingManager.getEnumEncodedValue(RoadAccess.KEY, RoadAccess.class), OSMRoadAccessParser.toOSMRestrictions(TransportationMode.CAR)));
         if (encodingManager.hasEncodedValue(AverageSlope.KEY) || encodingManager.hasEncodedValue(MaxSlope.KEY))
-            osmParsers.addWayTagParser(new SlopeCalculator(encodingManager.getDecimalEncodedValue(MaxSlope.KEY), encodingManager.getDecimalEncodedValue(AverageSlope.KEY)));
+            osmParsers.addWayTagParser(new SlopeCalculator(encodingManager.hasEncodedValue(MaxSlope.KEY) ? encodingManager.getDecimalEncodedValue(MaxSlope.KEY) : null,
+                    encodingManager.hasEncodedValue(AverageSlope.KEY) ? encodingManager.getDecimalEncodedValue(AverageSlope.KEY) : null));
```
easbar, line 656: We are already in the if clause that says average- and max-slope are available?!

core/src/test/java/com/graphhopper/routing/util/TagParsingTest.java
```diff
@@ -128,37 +128,6 @@ public void testMixBikeTypesAndRelationCombination() {
         assertTrue(bikePriorityEnc.getDecimal(false, edgeFlags) > mtbPriorityEnc.getDecimal(false, edgeFlags));
     }
 
-    @Test
-    public void testCompatibilityBug() {
-        EncodingManager manager2 = EncodingManager.create("bike2");
-        ReaderWay osmWay = new ReaderWay(1);
```
easbar, line 134: Not sure if this test is really needed but for #2669 I just replaced bike2 with mtb...

core/src/main/java/com/graphhopper/GraphHopper.java
easbar: > (the if is an OR not an AND)

Ah yes, I read it as an AND. And I was not aware the SlopeCalculator can run only for one of the two either. That is probably never really needed (run one but not the other), so this comes down to the same problem we have for `graph.flag_encoders`: We would like to enable a parser that writes multiple EVs, and the question is what is supposed to happen when we only put some but not all of these EVs in `graph.encoded_values`.

PR #2669, opened 2022-10-04
Remove VehicleTagParser#applyWayTags

easbar:
Everything that was possible with this method is still possible using `VehicleTagParser#handleWayTags` so it is no longer needed. This brings us closer to making the VehicleTagParsers just ordinary TagParsers.


easbar:
The changes in Bike2VehicleTagParser(Test) are of course no longer needed when #2668 is merged first.

PR #2670, opened 2022-10-06
Read Via Way Restrictions

lukasalexanderweber:
This one was already reviewed by @otbutz and @karussell in https://github.com/graphhopper/graphhopper/pull/2656, so I think this is ready to merge.

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/OSMTurnRelation.java
+++ b/core/src/main/java/com/graphhopper/reader/OSMTurnRelation.java
@@ -26,21 +26,27 @@ import java.util.*;
  */
 public class OSMTurnRelation {
     private final long fromOsmWayId;
-    private final long viaOsmNodeId;
+    private final ArrayList<Long> viaOSMIds;
     private final long toOsmWayId;
     private final RestrictionType restriction;
+    private final ViaType viaType;
     // vehicleTypeRestricted contains the dedicated vehicle type
     // example: restriction:bus = no_left_turn => vehicleTypeRestricted = "bus";
     private String vehicleTypeRestricted;
     private List<String> vehicleTypesExcept;
 
-    public OSMTurnRelation(long fromWayID, long viaNodeID, long toWayID, RestrictionType restrictionType) {
+    public OSMTurnRelation(long fromWayID, ArrayList<Long> viaOSMIds, long toWayID, RestrictionType restrictionType, ViaType viaType) {
         this.fromOsmWayId = fromWayID;
-        this.viaOsmNodeId = viaNodeID;
+        this.viaOSMIds = viaOSMIds;
         this.toOsmWayId = toWayID;
         this.restriction = restrictionType;
+        this.viaType = viaType;
         this.vehicleTypeRestricted = "";
         this.vehicleTypesExcept = new ArrayList<>();
+    }
+
+    public OSMTurnRelation(long fromWayID, long viaId, long toWayID, RestrictionType restrictionType, ViaType viaType) {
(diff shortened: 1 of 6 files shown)


easbar:
Thanks. I did not take a closer look, but the renaming makes reviewing a bit harder. I'm not 100% convinced that renaming `OSMTurnRelation` to `OSMTurnRestriction` is necessary (yes it is a turn restriction but it is also an OSM relation), but if you think it is usefuld and don't mind upgrading your branch I'll rename it in master so it is easier to see what changed here.

easbar:
... or actually maybe you better revert the rename for now in your branch(es). If we do the rename I think we should also rename all the variables, so we do not have something like `OSMTurnRestriction turnRelation = new OSMTurnRestriction()` etc. I could do this in master, but then you will end up having all sorts of merge conflicts probably. And if you rename everything including the variables in your branches your changes will be hard to review. Therefore my suggestion for now: Revert this (anyway minor) renaming (easy to do with IntelliJ or Eclipse) and then we rename it afterwards.

lukasalexanderweber:
I could also edit the https://github.com/graphhopper/graphhopper/pull/2670/commits/7d474bcc712b94f4c77f54a8950c42d5972d54f5 commit and update the variable names I missed (like `OSMTurnRestriction turnRelation = new OSMTurnRestriction()`).

I will later split the restrictions into `node restrictions` (restrictions with a via node) and `way restrictions` (restrictions with one or more via ways) and there `node relation` and `way relation` would be misleading. And I wouldn't like having `turn relation` and `way restriction` mixed up.

During review you can select the commits to look at. They are independent from each other. So you could ignore https://github.com/graphhopper/graphhopper/pull/2670/commits/7d474bcc712b94f4c77f54a8950c42d5972d54f5 and only evaluate the other changes and I will look for variable names I missed. 

Would this be fine for you?

lukasalexanderweber:
optionaly I delete https://github.com/graphhopper/graphhopper/pull/2670/commits/7d474bcc712b94f4c77f54a8950c42d5972d54f5 and we'll do the renaming later

easbar:
I did the renaming with all the variables in a separate branch here: ff48e2445fbf909f5e5457087a7d6ae0396747b0

Can you check if updating your code to this code is possible?

> I will later split the restrictions into node restrictions (restrictions with a via node) and way restrictions (restrictions with one or more via ways) and there node relation and way relation would be misleading. And I wouldn't like having turn relation and way restriction mixed up.

Ok that makes sense, so we should indeed rename it now or later.

> During review you can select the commits to look at. They are independent from each other. So you could ignore https://github.com/graphhopper/graphhopper/commit/7d474bcc712b94f4c77f54a8950c42d5972d54f5 and only evaluate the other changes and I will look for variable names I missed.

Yeah, but keeping the commits separated like this doesn't work so well in my experience, especially because you will have to update to latest master, there will be back and forth changes during review and so on.

easbar:
> Can you check if updating your code to this code is possible?

I mean in theory. Don't put your time into actually doing this, because when we commit this to master the commit hash might be different again. If you think you can manage this I'll add this commit to master and then you could actually update your code.

easbar:
I also moved OSMTurnRestriction.java into the osm package as you did. Not that it makes much sense that some classes like ReaderElement are in the reader package and the others are in reader/osm, but OSMTurnRestriction was even in another package than the corresponding test (probably the reason you moved it).

lukasalexanderweber:
@easbar  I finally got a working copy: https://github.com/graphhopper/graphhopper/pull/2673

PR #2672, opened 2022-10-10
Remove YAML support for server-side custom models

### LINKED ISSUE #2671: Use JSON for custom model files
https://github.com/graphhopper/graphhopper/commit/1e2c349463e64509db59e78fb48114028b052372#r86180948

karussell:
Fixes #2671.

For server-side JSON files still comments with `//` are allowed. See the example files.

Unfortunately this does not reduce the dependencies as dropwizard still uses (and has to use) the YAMLFactory

 - [x] there is intrinsic support in jackson for JSON with comments: Feature.ALLOW_COMMENTS -> don't use it to make these files still simple to parse for other JSON tools.

### DIFF
--- a/custom_models/truck.json
+++ b/custom_models/truck.json
@@ -1,0 +1,22 @@
+{
+  "comment0": "to use this custom model you need to set the following option in the config.yml",
+  "comment1": "graph.flag_encoders: roads|transportation_mode=HGV,car",
+  "comment2": "graph.encoded_values: toll,hgv,surface,max_width,max_height",
+  "distance_influence": 1,
+  "speed": [
+    {
+      "if": "true",
+      "limit_to": "car_average_speed * 0.9"
+    },
+    {
+      "if": "true",
+      "limit_to": "95"
+    }
+  ],
+  "priority": [
+    {
+      "if": "car_access == false || hgv == NO || max_width < 3 || max_height < 4",
+      "multiply_by": "0"
+    }
+  ]
+}


otbutz:
Wouldn't it be better to stick with vanilla JSON and rather use dedicated keys?

```json
{
  "comment": "this is a little customized model used to benchmark routing with a customized weighting that is similar to the standard car profile",
  "priority": [
    {
      "if": "road_access == PRIVATE",
      "multiply_by": "0.1"
    },
    {
      "else_if": "road_access == DESTINATION",
      "multiply_by": "0.1"
    }
  ],
  "distance_influence": 0
}
```

otbutz:
> why couldn't they add comments to the spec!?

Amen.

> see when you need multiple lines:

Unique keys are not a hard requirement AFAIK. As long as our parser doesn't choke on it:

```json

{
  "comment": "to use this custom model you need to set the following option in the config.yml",
  "comment": "graph.flag_encoders: roads|transportation_mode=HGV,car",
  "comment": "graph.encoded_values: toll,hgv,surface,max_width,max_height",
  "distance_influence": 1,
  "speed": [
    {
      "if": "true",
      "limit_to": "car_average_speed * 0.9"
    },
    {
      "if": "true",
      "limit_to": "95"
    }
  ],
  "priority": [
    {
      "if": "car_access == false || hgv == NO || max_width < 3 || max_height < 4",
      "multiply_by": "0"
    }
  ]
}
```

easbar:
With the `"comment"` fields we can add some descriptions to the custom model, but we cannot comment out some of the rules. I think this might be even more important than being able to add descriptions.

Btw ideally, these comments:

```
  "comment": "to use this custom model you need to set the following option in the config.yml",
  "comment": "graph.flag_encoders: roads|transportation_mode=HGV,car",
  "comment": "graph.encoded_values: toll,hgv,surface,max_width,max_height",
```

won't be necessary anyway :)

easbar (approved):
LGTM, thanks!

PR #2673, opened 2022-10-10
Read Via Way Restrictions

lukasalexanderweber:
https://github.com/graphhopper/graphhopper/pull/2670 with a new branch.

To be able to process more complex turn restrictions (with one ore more via ways instead of one via node), the functionality to read the restrictions are updated.

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMTurnRestriction.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMTurnRestriction.java
@@ -26,21 +26,27 @@ import java.util.*;
  */
 public class OSMTurnRestriction {
     private final long fromOsmWayId;
-    private final long viaOsmNodeId;
+    private final List<Long> viaOSMIds;
     private final long toOsmWayId;
     private final RestrictionType restriction;
+    private final ViaType viaType;
     // vehicleTypeRestricted contains the dedicated vehicle type
     // example: restriction:bus = no_left_turn => vehicleTypeRestricted = "bus";
     private String vehicleTypeRestricted;
     private List<String> vehicleTypesExcept;
 
-    public OSMTurnRestriction(long fromWayID, long viaNodeID, long toWayID, RestrictionType restrictionType) {
+    public OSMTurnRestriction(long fromWayID, List<Long> viaOSMIds, long toWayID, RestrictionType restrictionType, ViaType viaType) {
         this.fromOsmWayId = fromWayID;
-        this.viaOsmNodeId = viaNodeID;
+        this.viaOSMIds = viaOSMIds;
         this.toOsmWayId = toWayID;
         this.restriction = restrictionType;
+        this.viaType = viaType;
         this.vehicleTypeRestricted = "";
         this.vehicleTypesExcept = new ArrayList<>();
+    }
+
+    public OSMTurnRestriction(long fromWayID, long viaId, long toWayID, RestrictionType restrictionType, ViaType viaType) {
+        this(fromWayID, new ArrayList<>(Arrays.asList(viaId)), toWayID, restrictionType, viaType);
     }
 
     public long getOsmIdFrom() {
@@ -51,12 +57,16 @@ public class OSMTurnRestriction {
         return toOsmWayId;
     }
 
-    public long getViaOsmNodeId() {
-        return viaOsmNodeId;
+    public List<Long> getViaOSMIds() {
+        return viaOSMIds;
     }
 
     public RestrictionType getRestriction() {
         return restriction;
+    }
+
+    public ViaType getViaType() {
+        return viaType;
     }
 
     public String getVehicleTypeRestricted() {
@@ -89,7 +99,7 @@ public class OSMTurnRestriction {
 
     @Override
     public String toString() {
-        return "*-(" + fromOsmWayId + ")->" + viaOsmNodeId + "-(" + toOsmWayId + ")->*";
+        return "*-(" + fromOsmWayId + ")->" + viaOSMIds + "-(" + toOsmWayId + ")->*";
     }
 
     public enum RestrictionType {
@@ -109,7 +119,7 @@ public class OSMTurnRestriction {
         }
 
         public static RestrictionType get(String tag) {
-        	RestrictionType result = null;
+            RestrictionType result = null;
             if (tag != null) {
                 result = tags.get(tag);
             }
@@ -118,6 +128,6 @@ public class OSMTurnRestriction {
     }
 
     public enum ViaType {
-        VIA_NODE, VIA_WAY, MULTI_VIA_WAY;
+        UNSUPPORTED, NODE, WAY, MULTI_WAY;
     }
 }
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMTurnRestriction.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMTurnRestriction.java
@@ -108,8 +108,8 @@ public class OSMTurnRestriction {
             tags.put("only_straight_on", ONLY);
         }
 
-        public static RestrictionType getRestrictionType(String tag) {
-            RestrictionType result = null;
+        public static RestrictionType get(String tag) {
+        	RestrictionType result = null;
             if (tag != null) {
                 result = tags.get(tag);
             }
@@ -117,4 +117,7 @@ public class OSMTurnRestriction {
         }
     }
 
+    public enum ViaType {
+        VIA_NODE, VIA_WAY, MULTI_VIA_WAY;
+    }
 }
--- a/core/src/test/java/com/graphhopper/reader/osm/OSMReaderTest.java
+++ b/core/src/test/java/com/graphhopper/reader/osm/OSMReaderTest.java
@@ -46,6 +46,7 @@ import org.junit.jupiter.api.Test;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -577,6 +578,41 @@ public class OSMReaderTest {
 
         assertTrue(tcStorage.get(carTCEnc, edge10_11, n11, edge11_14) == 0);
         assertTrue(tcStorage.get(bikeTCEnc, edge10_11, n11, edge11_14) > 0);
+    }
+
+    @Test
+    void testViaWayTurnRestrictions() {
+        ReaderRelation rel = new ReaderRelation(1L);
+
+        rel.setTag("restriction", "no_u_turn");
+        rel.add(new ReaderRelation.Member(ReaderElement.Type.WAY, 1L, "from"));
+        rel.add(new ReaderRelation.Member(ReaderElement.Type.WAY, 2L, "via"));
+        rel.add(new ReaderRelation.Member(ReaderElement.Type.WAY, 3L, "to"));
+
+        List<OSMTurnRestriction> osmRel = OSMReader.createTurnRestrictions(rel);
+        assertEquals(osmRel.get(0).getViaType(), OSMTurnRestriction.ViaType.WAY);
+        assertEquals(osmRel.get(0).getViaOSMIds().get(0), 2);
+    }
+
+    @Test
+    void testMultiViaWayTurnRestrictions() {
+        ReaderRelation rel = new ReaderRelation(1L);
+
+        rel.setTag("restriction", "no_u_turn");
+        rel.add(new ReaderRelation.Member(ReaderElement.Type.WAY, 1L, "from"));
+        rel.add(new ReaderRelation.Member(ReaderElement.Type.WAY, 2L, "via"));
+        rel.add(new ReaderRelation.Member(ReaderElement.Type.WAY, 3L, "via"));
+        rel.add(new ReaderRelation.Member(ReaderElement.Type.WAY, 4L, "via"));
+        rel.add(new ReaderRelation.Member(ReaderElement.Type.WAY, 5L, "to"));
+
+        ArrayList<Long> viaOSMIds = new ArrayList<>();
+        viaOSMIds.add(2L);
+        viaOSMIds.add(3L);
+        viaOSMIds.add(4L);
+
+        List<OSMTurnRestriction> osmRel = OSMReader.createTurnRestrictions(rel);
+        assertEquals(osmRel.get(0).getViaType(), OSMTurnRestriction.ViaType.MULTI_WAY);
+        assertEquals(osmRel.get(0).getViaOSMIds(), viaOSMIds);
     }
 
     @Test
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -598,8 +598,8 @@ public class OSMReader {
     }
 
     static List<OSMTurnRestriction> createTurnRestrictions(ReaderRelation relation, String restrictionType,
-                                                           String vehicleTypeRestricted, List<String> vehicleTypesExcept) {
-        OSMTurnRestriction.RestrictionType type = OSMTurnRestriction.RestrictionType.getRestrictionType(restrictionType);
+                                                     String vehicleTypeRestricted, List<String> vehicleTypesExcept) {
+    	OSMTurnRestriction.RestrictionType type = OSMTurnRestriction.RestrictionType.get(restrictionType);
         if (type != OSMTurnRestriction.RestrictionType.UNSUPPORTED) {
             // we use -1 to indicate 'missing', which is fine because we exclude negative OSM IDs (see #2652)
             long viaNodeID = -1;
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -549,10 +549,13 @@ public class OSMReader {
                 }
             };
             for (OSMTurnRestriction turnRestriction : createTurnRestrictions(relation)) {
-                int viaNode = map.getInternalNodeIdOfOsmNode(turnRestriction.getViaOsmNodeId());
-                // street with restriction was not included (access or tag limits etc)
-                if (viaNode >= 0)
-                    osmParsers.handleTurnRestrictionTags(turnRestriction, map, baseGraph);
+                // till now only Via-Node Restrictions are considered 
+                if (turnRestriction.getViaType() == OSMTurnRestriction.ViaType.NODE) {
+                    int viaNode = map.getInternalNodeIdOfOsmNode(turnRestriction.getViaOSMIds().get(0));
+                    // street with restriction was not included (access or tag limits etc)
+                    if (viaNode >= 0)
+                        osmParsers.handleTurnRestrictionTags(turnRestriction, map, baseGraph);
+                }
             }
         }
     }
@@ -598,24 +601,47 @@ public class OSMReader {
     }
 
     static List<OSMTurnRestriction> createTurnRestrictions(ReaderRelation relation, String restrictionType,
-                                                     String vehicleTypeRestricted, List<String> vehicleTypesExcept) {
-    	OSMTurnRestriction.RestrictionType type = OSMTurnRestriction.RestrictionType.get(restrictionType);
+                                                        String vehicleTypeRestricted, List<String> vehicleTypesExcept) {
+        OSMTurnRestriction.RestrictionType type = OSMTurnRestriction.RestrictionType.get(restrictionType);
         if (type != OSMTurnRestriction.RestrictionType.UNSUPPORTED) {
+            ArrayList<Long> viaIDs = new ArrayList<>();
+            OSMTurnRestriction.ViaType viaType = OSMTurnRestriction.ViaType.UNSUPPORTED;
             // we use -1 to indicate 'missing', which is fine because we exclude negative OSM IDs (see #2652)
-            long viaNodeID = -1;
             long toWayID = -1;
 
+            // A Turn Restriction (https://wiki.openstreetmap.org/wiki/Relation:restriction) can consist of...
+
             for (ReaderRelation.Member member : relation.getMembers()) {
-                if (ReaderElement.Type.WAY == member.getType() && "to".equals(member.getRole()))
+                // ONE TO Way (a "no_exit" restriction can have ONE OR MORE TO ways)
+                if (ReaderElement.Type.WAY == member.getType() && "to".equals(member.getRole())) {
                     toWayID = member.getRef();
-                else if (ReaderElement.Type.NODE == member.getType() && "via".equals(member.getRole()))
-                    viaNodeID = member.getRef();
+                }
+
+                // ONE VIA NODE or
+                else if (ReaderElement.Type.NODE == member.getType() && "via".equals(member.getRole())) {
+                    if (viaType == OSMTurnRestriction.ViaType.UNSUPPORTED) {
+                        viaIDs.add(member.getRef());
+                        viaType = OSMTurnRestriction.ViaType.NODE;
+                    }
+                }
+                // ONE OR MORE VIA WAYS
+                else if (ReaderElement.Type.WAY == member.getType() && "via".equals(member.getRole())) {
+                    if (viaType == OSMTurnRestriction.ViaType.UNSUPPORTED) {
+                        viaIDs.add(member.getRef());
+                        viaType = OSMTurnRestriction.ViaType.WAY;
+                    } else if (viaType == OSMTurnRestriction.ViaType.WAY || viaType == OSMTurnRestriction.ViaType.MULTI_WAY) {
+                        viaIDs.add(member.getRef());
+                        viaType = OSMTurnRestriction.ViaType.MULTI_WAY;
+                    }
+                }
             }
-            if (toWayID >= 0 && viaNodeID >= 0) {
+
+            // ONE FROM Way (a "no_entry" restriction can have ONE OR MORE FROM ways)
+            if (toWayID >= 0 && viaIDs.size() > 0) {
                 List<OSMTurnRestriction> res = new ArrayList<>(2);
                 for (ReaderRelation.Member member : relation.getMembers()) {
                     if (ReaderElement.Type.WAY == member.getType() && "from".equals(member.getRole())) {
-                        OSMTurnRestriction osmTurnRestriction = new OSMTurnRestriction(member.getRef(), viaNodeID, toWayID, type);
+                        OSMTurnRestriction osmTurnRestriction = new OSMTurnRestriction(member.getRef(), viaIDs, toWayID, type, viaType);
                         osmTurnRestriction.setVehicleTypeRestricted(vehicleTypeRestricted);
                         osmTurnRestriction.setVehicleTypesExcept(vehicleTypesExcept);
                         res.add(osmTurnRestriction);
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTurnRestrictionParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTurnRestrictionParser.java
@@ -66,7 +66,8 @@ public class OSMTurnRestrictionParser implements TurnCostParser {
      */
     void addRelationToTCStorage(OSMTurnRestriction osmTurnRestriction, ExternalInternalMap map, Graph graph) {
         TurnCostStorage tcs = graph.getTurnCostStorage();
-        int viaNode = map.getInternalNodeIdOfOsmNode(osmTurnRestriction.getViaOsmNodeId());
+        int viaNode = map.getInternalNodeIdOfOsmNode(osmTurnRestriction.getViaOSMIds().get(0));
+
         EdgeExplorer edgeOutExplorer = getOutExplorer(graph), edgeInExplorer = getInExplorer(graph);
 
         try {
@@ -100,7 +101,7 @@ public class OSMTurnRestrictionParser implements TurnCostParser {
                 }
             }
         } catch (Exception e) {
-            throw new IllegalStateException("Could not built turn table entry for relation of node with osmId:" + osmTurnRestriction.getViaOsmNodeId(), e);
+            throw new IllegalStateException("Could not built turn table entry for relation of node with osmId:" + viaNode, e);
         }
     }
 
--- a/core/src/test/java/com/graphhopper/reader/osm/OSMTurnRestrictionTest.java
+++ b/core/src/test/java/com/graphhopper/reader/osm/OSMTurnRestrictionTest.java
@@ -34,7 +34,7 @@ public class OSMTurnRestrictionTest {
     public void testAcceptsTurnRestriction() {
         List<String> vehicleTypes = new ArrayList<>(Arrays.asList("motorcar", "motor_vehicle", "vehicle"));
         List<String> vehicleTypesExcept = new ArrayList<>();
-        OSMTurnRestriction osmTurnRestriction = new OSMTurnRestriction(1, 1, 1, OSMTurnRestriction.RestrictionType.NOT);
+        OSMTurnRestriction osmTurnRestriction = new OSMTurnRestriction(1, 1, 1, OSMTurnRestriction.RestrictionType.NOT, OSMTurnRestriction.ViaType.WAY);
         assertTrue(osmTurnRestriction.isVehicleTypeConcernedByTurnRestriction(vehicleTypes));
 
         vehicleTypesExcept.add("bus");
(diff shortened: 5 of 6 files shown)


core/src/main/java/com/graphhopper/reader/osm/OSMTurnRestriction.java
```diff
@@ -26,23 +26,29 @@
  */
 public class OSMTurnRestriction {
     private final long fromOsmWayId;
-    private final long viaOsmNodeId;
+    private final ArrayList<Long> viaOSMIds;
```
otbutz, line 29: The ArrayList is just an implementation detail. Stick with `List` for fields and method arguments.

otbutz (changes requested):

core/src/main/java/com/graphhopper/reader/osm/OSMTurnRestriction.java
lukasalexanderweber: fixed

lukasalexanderweber:
@easbar @karussell is there anything left for me to do to merge this PR?

easbar:
I'll take a look soon.

aibear-huang:
does the via-ways restriction works?i found only reader code changed.
thanks for your answer.

core/src/test/java/com/graphhopper/reader/osm/OSMReaderTest.java
```diff
@@ -579,6 +580,41 @@ public void testTurnRestrictions() {
...
+        rel.add(new ReaderRelation.Member(ReaderElement.Type.WAY, 4L, "via"));
+        rel.add(new ReaderRelation.Member(ReaderElement.Type.WAY, 5L, "to"));
+
+        ArrayList<Long> viaOSMIds = new ArrayList<>();
+        viaOSMIds.add(2L);
+        viaOSMIds.add(3L);
+        viaOSMIds.add(4L);
+
+        List<OSMTurnRestriction> osmRel = OSMReader.createTurnRestrictions(rel);
+        assertEquals(osmRel.get(0).getViaType(), OSMTurnRestriction.ViaType.MULTI_WAY);
```
easbar, line 614: JUnit fun fact: The signature of `assertEquals` is `assertEquals(expected, actual)` not `assertEquals(actual, expected)`.

easbar (commented):
I did some changes in master today after I read the turn restriction wiki again: #2680. I also added a new branch that adds the ViaType you added here (without MULTI_WAY and UNSUPPORTED) on top of this: fddb7f6893135c0892bc528714614551c2a31d03.
In your previous branch there were lots of changes to OSMReader etc. so I find it a bit hard to see where we are going here. I'm also not sure how much time on our side will be available to work on #446 currently. Can you give a rough sketch (no code) of what you think needs to be done to implement #446 (for the single via-way case) so we can get an estimate how much time will be required to finish this?

core/src/test/java/com/graphhopper/reader/osm/OSMReaderTest.java
easbar: No they just swapped the position of the message parameter (from front to back)'

lukasalexanderweber:
Ok I'll give you a short outline about my previous and future work on this topic.

At first, I did all the development in one branch/PR. It was asked to split the development in multiple PRs: https://github.com/graphhopper/graphhopper/pull/2656#issuecomment-1269560524

I then decided to create three PRs:
- Read Via Way Restrictions (this one)
- Refactor OSM Reader (https://github.com/lukasalexanderweber/graphhopper/tree/refactor_osmreader2)
- Implement Single Via Way Restrictions (https://github.com/lukasalexanderweber/graphhopper/tree/single_via_way_restrictions)

I will now parallel work on 
- Implement Multi Via Way Restrictions 

I've got two working approaches on simple via way turn restrictions.

The first one is https://github.com/graphhopper/graphhopper/issues/446#issuecomment-507221777 
and the second one is slightly different (and I think more promising) for multiple via ways. 

Restriction ID `2468010`

image

other working examples are 

`2468010, 3567246, 5614305, 5739275, 7636822, 8409418, 11290869, 13335861, 13447603
`

Both approaches have limitations regarding "self-contained" restrictions as 

image

but these can be solved with the approach I plan for multiple via ways.

Also both approaches manipulate the navigation instructions (when passing zero distance edges).

I could also imagine a meeting where we can introduce and I can give you a bit more detail about the development if you are interested.

easbar:
> Refactor OSM Reader (https://github.com/lukasalexanderweber/graphhopper/tree/refactor_osmreader2)

Why is this needed and how is it related to this issue?

> The first one is https://github.com/graphhopper/graphhopper/issues/446#issuecomment-507221777
and the second one is slightly different (and I think more promising) for multiple via ways.

Ok, this sketch in #446 was just for one specific type of junction. Do you think there are different cases depending on the types of junctions (i.e. the number of adjacent edges at the affected nodes?) (for single via way restrictions)? Can we treat `no_xyz` in a similar way as `only_xyz` restrictions? Do they even both occur in OSM (with single/multiple via-ways)?

> Both approaches have limitations regarding "self-contained" restrictions as

Can you explain the difference between the two approaches? Or do you think they are too hard to explain via GitHub? What are the limitations you mentioned? What are 'self-contained' restrictions?

> but these can be solved with the approach I plan for multiple via ways.

Ok, so you haven't really decided what the final solution will be, also for single via-ways? Shall we postpone the discussion until it is more clear?

lukasalexanderweber:
> > Refactor OSM Reader (https://github.com/lukasalexanderweber/graphhopper/tree/refactor_osmreader2)
> 
> Why is this needed and how is it related to this issue?
> 
> >

OSMReader and WaySegmentParser are tightly coupled. I needed a long time to figure out whats happening there. I really think that the refactoring would benefit the understanding of the dataflow happening while preparing graphhopper from an osm file. I put much effort in it and my implementation of via ways wouln't work without it. The biggest reason is that I need the ability to add edges to the graph while creating the via way restrictions.

> Do you think there are different cases depending on the types of junctions (i.e. the number of adjacent edges at the affected nodes?)

Till now there are two Edge Cases I encountered: Combinations of Via Node and Via Way restrictions for one Node/Way; and the thing I called 'self-contained' restrictions.

> What are 'self-contained' restrictions?

as can be seen in the image, these share the same edges and nodes (lets call them overlapping restrictions, self-contained is a really bad term for that xD)

> Can we treat `no_xyz` in a similar way as `only_xyz` restrictions? Do they even both occur in OSM (with single/multiple via-ways)?

Yes they occour and yes both `no` and `only` are correctly managed by both approaches

> What are the limitations you mentioned?

In the "self-contained aka overlapping" image, both U-Turns should be prohibited but they are still taken (I know why but there is more to implement for these to work). And then the navigation instruction manipulation I mentioned, but I think for someone who is deeper into that this might be easy to handle.

> Ok, so you haven't really decided what the final solution will be, also for single via-ways? Shall we postpone the discussion until it is more clear?

Since my task at work is to implement multi via ways as well, I would work on with the second apprach if this is fine for you. 

> Can you explain the difference between the two approaches?

Basically I create a detour starting at the start node of the from way. This allows other restrictions to "jump on" the "restriction graph" (see node `d`) when handling multiple via way restrictions. I scanned a sketch, where the text markers are "no" restrictions and below you see the resulting graph + the via node restrictions.

graph

lukasalexanderweber:
@easbar with the information that you know have, what do you think would be the next steps?

I could think of:
- I process your change requests and fix the merge conflicts
- you merge this PR
- we move on to discuss the refactoring

parallel I will continue working on multi via ways and you can discuss internally if you think a meeting could be a good idea.

easbar:
> Till now there are two Edge Cases I encountered: Combinations of Via Node and Via Way restrictions for one Node/Way; and the thing I called 'self-contained' restrictions.

> as can be seen in the image, these share the same edges and nodes (lets call them overlapping restrictions, self-contained is a really bad term for that xD)

So to implement turn restrictions with via-ways we need to add artificial nodes and/or edges to the graph, like in the sketch in #446, right? And you mean there can be separate restriction relations that affect the same parts of the graph? This is what you call overlapping restrictions? I can see how this makes the matter more complicated, yes. Do you have examples for such relations on openstreetmap.org? And how often do these occur? Would it make sense to ignore these cases first and focus on those that aren't overlapping with each other?

> I scanned a sketch, where the text markers are "no" restrictions and below you see the resulting graph + the via node restrictions.

Ok, so this is rather an advanced example where the c-b-d-e-f restriction has two via-ways **and** there are overlapping restrictions nearby? What are the directions of the green restrictions? For example is it forbidden to turn from edge 4 to 6, or 6 to 4, or both? And why did you pick such an advanced example? And do you think it general enough to test your approach? I mean do you think that if we have a working solution for this specific example does it mean we can cover all possible restrictions?

I'm probably behind here, but the first example I'd like to understand and implement is a via-way restriction that can either be a 'no' or 'only' restriction with ways from=a, via=b, to=c where there can be arbitrary adjacent edges at nodes 0, 1, 2 and 3.

```
  a   b   c
0---1---2---3
```

At least that would be my definition of the general case we need to deal with when we want to support via-way restrictions (with a single way). Do you already know how to do this? On the conceptual level the next step could then be generalizing this to multiple via-ways and/or overlapping restrictions?

> Since my task at work is to implement multi via ways as well, I would work on with the second apprach if this is fine for you.

Yes, of course. I am by no means trying to tell you what to work on. If you are willing to share your final result it could also be an option that you just finish it the way you want and once it is complete we at GH can think about if and how to include it in our code base. Of course I also see how it could be advantageous to do this together from the beginning, if there will be enough bandwidth to discuss this as we proceed.
And regarding the second approach: Again, do you think limiting the approach to single-way restrictions and/or non-overlapping restrictions makes no sense, because the first approach cannot be generalized anyway?

>  what do you think would be the next steps?

So far I mostly want to understand what needs to be done on a conceptual level. I am much less concerned about actually implementing this once it is clear what we are trying to do. I don't think it is worth to update the present PR, because I already added the ViaType on top of my recent changes in a separate branch here: https://github.com/graphhopper/graphhopper/commit/fddb7f6893135c0892bc528714614551c2a31d03 and just adding it doesn't really help much for implementing the feature we want. Regarding the OSMReader refactoring, I cannot comment much, because I don't know what needs to be done yet. For example you said:

> The biggest reason is that I need the ability to add edges to the graph while creating the via way restrictions.

At which point of the OSM reading (we read the file twice) do you think shall we add the turn restrictions? So far I was assuming we can do this simply after the graph (without turn restrictions) is finished. So I don't really understand why you think you don't have the ability to add edges while creating the via way restrictions.

easbar:
Would the following work (for the single-via way case, without any overlapping restrictions)? And maybe this is one of the two approaches you meant?

```
Assume we have three edges: a, b, c in a 'no' restriction:

 a b c
0-1-2-3

It is not allowed to go from a via b to c, but all other turns are allowed.
The restriction means: We cannot turn from b to c if we came from a. So the first thing we can do 
is restricting the turn from b to c.
But then we have a problem: We can no longer turn from b to c when we come from another edge adjacent
to node 1 a'!=a.
So we introduce an artificial edge b' parallel to b and restrict the turn from a to b'
Now it looks like this, where all turns are allowed except a to b' and b to c.

 a b c
0-1=2-3
   b'

This means we can neither go from a via b to c, nor from a via b' to c. 
We can, however, go from any edge a'!=a at node 1 via b' to c and we 
can also go to any edge c'!=c from a via b, so we replaced the via-way
restriction with two via-node restrictions and an artificial edge. We should
probably also reject turns between b and b'.
```

And what about the same in the 'only'-case?

```
The only-restriction means: Coming from a we are forced to turn onto b and then onto c.
So all turns from a are forbidden except the one to b. We also reject the turn from b to 
anything else than c. We add an artificial edge b' just as in the previous example and 
deny turning from a to b'.
So here we need to replace the via-way restriction with via-node restrictions at node 1 
for all turns from a to a' and at node 2 for all turns from b to c', the artificial edge b'
and via-node restrictions from a to b'.
```

The new thing here are these artificial edges. For the path post-processing like instructions etc. 
I do not see a problem yet, because we can just copy all the properties of the corresponding
real edge and the artificial edges will behave just like the real ones. The location index
seems more problematic. In the two examples above we should not snap to edge b, because
then we cannot turn to c. So we would have to exclude b from the index, but include edge b'
instead.

Does this make sense?

lukasalexanderweber:
> So to implement turn restrictions with via-ways we need to add artificial nodes and/or edges to the graph, like in the sketch in #446, right?

yes

> And you mean there can be separate restriction relations that affect the same parts of the graph? This is what you call overlapping restrictions?

yes

> Do you have examples for such relations on openstreetmap.org?

`2602528 2602529 2602530`
`2821560 2821561 2821562 2821563`
`2920825 2920826 2920827 2920828`

> And how often do these occur?

In Regierungsbezirk Karlsruhe, ~2/3s of the via way restrictions are overlapping (as one junction often has 3-4 restrictions like the examples)

> Would it make sense to ignore these cases first and focus on those that aren't overlapping with each other?

Yes we can do that  (I already read "nested restrictions" as a term)

> Ok, so this is rather an advanced example where the c-b-d-e-f restriction has two via-ways **and** there are overlapping restrictions nearby? What are the directions of the green restrictions? For example is it forbidden to turn from edge 4 to 6, or 6 to 4, or both?

The directions are from left to right. The restrictions are `from 2 via [3, 4] to 5`, `from 4 via [7] to 8`, `from 4 via e to 6`

> And why did you pick such an advanced example?

At first I conceptualized and implemented a simple via way restriction. Since this worked kind of well I wanted to go further and solve the overlapping/multi way cases. I always orientate myself a bit at https://github.com/Project-OSRM/osrm-backend/pull/5907 and try to transfer the approaches to a node based graph, thats where I've got the example.

> And do you think it general enough to test your approach? I mean do you think that if we have a working solution for this specific example does it mean we can cover all possible restrictions?

Probably not, it was just a first orientation how I could approach more complex via way restrictions

> I'm probably behind here, but the first example I'd like to understand and implement is a via-way restriction that can either be a 'no' or 'only' restriction with ways from=a, via=b, to=c where there can be arbitrary adjacent edges at nodes 0, 1, 2 and 3.
> 
> ```
>   a   b   c
> 0---1---2---3
> ```
> 
> At least that would be my definition of the general case we need to deal with when we want to support via-way restrictions (with a single way). Do you already know how to do this? On the conceptual level the next step could then be generalizing this to multiple via-ways and/or overlapping restrictions?

I've made a sketch of the three approaches I've tested:

IMG_20221020_0745532

- the one on the left being my first approach which also can be seen [here](https://github.com/graphhopper/graphhopper/pull/2656#issuecomment-1246399553) (both comments are marked as outdated, sorry) 
- the one in the middle being the sketch in https://github.com/graphhopper/graphhopper/issues/446. 
- the one on the right going towards the other sketch with the more complicated example

The main strategie is the same at all approaches:
1. restrict the first turn
2. create a some artificial nodes/edges where the route can detour if trying to make this turn
3. lead back on the original graph, restricting the second turn

> > Since my task at work is to implement multi via ways as well, I would work on with the second apprach if this is fine for you.
> 
> Yes, of course. I am by no means trying to tell you what to work on. If you are willing to share your final result it could also be an option that you just finish it the way you want and once it is complete we at GH can think about if and how to include it in our code base. Of course I also see how it could be advantageous to do this together from the beginning, if there will be enough bandwidth to discuss this as we proceed. And regarding the second approach: Again, do you think limiting the approach to single-way restrictions and/or non-overlapping restrictions makes no sense, because the first approach cannot be generalized anyway?

I think that working on this together would be the more sustainable way. I'm not even sure if the approach cannot be generalized anyway. Maybe it turns out that the first of the three approaches is the best. I mostly rejected it because it made too much damage at the navigation instructions. EDIT your sollution solves this case, I'm really happy for such insights.

> Would the following work (for the single-via way case, without any overlapping restrictions)? And maybe this is one of the two approaches you meant?

Yes man, this is so clever. It removes all the struggle of artificial nodes and the navigation instruction issues coming with them. Another sketch:

IMG_20221020_083221

> For the path post-processing like instructions etc.
> I do not see a problem yet, because we can just copy all the properties of the corresponding
> real edge and the artificial edges will behave just like the real ones

THIS! I never had problems with the artificial edges only with the nodes

> The location index
> seems more problematic. In the two examples above we should not snap to edge b, because
> then we cannot turn to c. So we would have to exclude b from the index, but include edge b'
> instead.

Couln't we simply switch the restrictions to `a via 1 to b` and `b' via 2 to c`? And prevent adding artificial edges to the location index? (I don't know anything about this index just yet)

> At which point of the OSM reading (we read the file twice) do you think shall we add the turn restrictions? So far I was assuming we can do this simply after the graph (without turn restrictions) is finished. So I don't really understand why you think you don't have the ability to add edges while creating the via way restrictions.

Yes we'll do it after the graph (without turn restrictions) is finished. 

From an implementation side, however, we would need to bind the turn restriction creation to the WaySegmentParser since it's got the addEdge function etc. in it (and we cannot simply access it in another class due to its complex builder stuff). Additionally, the WaySegmenthandler hadles 1. the OSM parsing, 2. node handling, 3. way handling, 4. relation handling (by getting those functions from OSMReader!?) and on top of it now the Restriction Handling would also be placed. For me this just wasn't the best way it could be done so I worked out a way to restructure the functionality (I didn't change any function! I just restructured it for clearer reponsibilities and thats what I want to present you in another PR). So even after the new insights about the conception of Turn Restrictions I'm confinced that refactoring should be considered.

lukasalexanderweber:
Even better, with this approach multi via ways are easy to model (still with the assumption that there are no overlaps). We only need to prevent the "crossing" between artificial and original ways between two via ways. I have a real example with 2 via ways ([3082392](https://www.openstreetmap.org/relation/3082392)) and transfered it to a unit test with 3 via ways, both work as expected. The implementation is [here](https://github.com/lukasalexanderweber/graphhopper/tree/multi_via_way_restrictions)

image

easbar:
> 2602528 2602529 2602530
2821560 2821561 2821562 2821563
2920825 2920826 2920827 2920828

Thanks, this is very useful and shows how important this feature is and also that overlapping restrictions must be expected. Just for reference here are some screenshots for the first junction:

image
https://www.openstreetmap.org/relation/2602528
image
https://www.openstreetmap.org/relation/2602529
image
https://www.openstreetmap.org/relation/2602530
image

... and of course we [currently violate these restrictions](https://graphhopper.com/maps/?point=49.55679%2C8.64793_49.55679%2C+8.64793&point=49.55703%2C8.64786_49.55703%2C+8.64786&profile=car&layer=Omniscale)

image

> In Regierungsbezirk Karlsruhe, ~2/3s of the via way restrictions are overlapping (as one junction often has 3-4 restrictions like the examples)

Ok, so these aren't rare or something. I think it still helps to get simplify first and think about how to do implement the feature when there are no overlaps, but ultimately we should also handle overlapping restrictions.

> I already read "nested restrictions" as a term

You mean nested=overlapping?

> The directions are from left to right. The restrictions are from 2 via [3, 4] to 5, from 4 via [7] to 8, from 4 via e to 6

Ok, I thought so but wasn't sure.

> At first I conceptualized and implemented a simple via way restriction. Since this worked kind of well I wanted to go further and solve the overlapping/multi way cases. I always orientate myself a bit at https://github.com/Project-OSRM/osrm-backend/pull/5907 and try to transfer the approaches to a node based graph, thats where I've got the example.

:+1: Looking at that discussion it doesn't sound like everything becomes simple just when using an edge-based graph either :)

> I've made a sketch of the three approaches I've tested:

Thanks, that made it clear.

> I think that working on this together would be the more sustainable way.

:+1: Ok let's try this.

> Yes man, this is so clever. It removes all the struggle of artificial nodes and the navigation instruction issues coming with them. 

:heart:

> Couln't we simply switch the restrictions to a via 1 to b and b' via 2 to c? And prevent adding artificial edges to the location index? (I don't know anything about this index just yet)

Yes, good idea. I mean this:

```
 a b c
0-1=2-3
   b'
```

is symmetric, so we can make it so the artificial edge is always the one we do not want in the location index and make sure artificial edges are ignored there. This is easier than excluding all edges for which there is an artificial counter part or something. 

> I've implemented it [here](https://github.com/lukasalexanderweber/graphhopper/tree/single_via_way_restrictions) and it works SO good!

Nice, happy to hear!

> Even better, with this approach multi via ways are easy to model (still with the assumption that there are no overlaps). We only need to prevent the "crossing" between artificial and original ways between two via ways. I have a real example with 2 via ways ([3082392](https://www.openstreetmap.org/relation/3082392)) and transferred it to a unit test with 3 via ways, both work as expected.

Yes, that makes sense. But do we really need to restrict b via 2 to c'? But maybe it doesn't even matter. And we should prevent b via 2 to b' and c via 3 to c' I think. Anyway, shall we first focus on single via-ways, then see if there are actual problems for the overlapping cases and then either try to fix the overlaps or generalize to multi via-ways? I mean if you feel like exploring these advanced cases please go ahead, and maybe I'll think about how to integrate this with the current code. I'll have a look at yours of course.

> From an implementation side, however, we would need to bind the turn restriction creation to the WaySegmentParser since it's got the addEdge function etc. in it (and we cannot simply access it in another class due to its complex builder stuff). Additionally, the WaySegmenthandler hadles 1. the OSM parsing, 2. node handling, 3. way handling, 4. relation handling (by getting those functions from OSMReader!?) and on top of it now the Restriction Handling would also be placed. For me this just wasn't the best way it could be done so I worked out a way to restructure the functionality (I didn't change any function! I just restructured it for clearer reponsibilities and thats what I want to present you in another PR). So even after the new insights about the conception of Turn Restrictions I'm confinced that refactoring should be considered.

Ok, let me explain the 'idea' behind the OSMReader and related code:

The first code we call when we read an OSM file is `OSMInput#open` which gives us access to the different OSM elements (ReaderElements). As you know basically there are nodes, ways and relations. We then check the type of each element and call a different method for each type: `handleNode/Way/Relation/FileHeader`. A generic OSM reader might simply open an OSM file and give access to these different elements. This would be very flexible, but at the same time just as useless. I mean the complicated part here is decoding the pbf format (but this is already done in OSMInput) and calling a different `handleXYZ` depending on the element type method is trivial. This is the reason the `ReaderElementHandler` is internal to the (much more specific) `WaySegmentParser`. If somebody needs to read an OSM file and call different methods depending on the element type she can just use OSMInput and write this tiny loop with the switch statement and will never even need WaySegmentParser or even OSMReader.
Ok, but since we want to create a routeable graph from the OSM file we got a lot more work to do. Basically there are two reasons why this is not trivial: 1) OSM ways do not contain any coordinates, but only ids pointing to OSM nodes and 2) other than common node references OSM contains no information about where there are 'junctions' between ways, ways need to be split into several edges of our routing graph and for better performance we want to skip intermediate/pillar nodes during routing. Doing this in a memory-efficient way like GraphHopper does is not trivial at all. At the same time once this is done it is something that might be re-used in different ways. The details about how the GraphHopper graph is created can vary a lot, but dealing with node-references in ways and splitting ways into segments connecting actual junctions is likely always the same. For this reason I decided to put all the code that does all this heavy lifting of splitting ways into segments and assigning the right coordinates into one piece of code (WaySegmentParser): https://github.com/graphhopper/graphhopper/pull/2448.
For example in this PR I demonstrated how it can easily be used to create an entirely different graph from OSM: One that uses waterways for routing instead of routes. The baseline is: WaySegmentParser does all the work to figure out where there are junctions in OSM and how they are connected and we can write code **on top of this** to specify all the details about how the routing graph is going to look like (i.e. which OSM tags we use, what kind of data we store per edge and all this). Much of this latter stuff is in OSMReader. So I don't agree that "OSMReader and WaySegmentParser are tightly coupled", but rather WaySegmentParser is completely independent of OSMReader (it can be used without it) and OSMReader uses WaySegmentParser for what OSMReader does. The relation is the same as the one between `java.util.ArrayList` and `java.lang.Object`. The WaySegmentParser doesn't "got the addEdge function etc.", but the addEdge function is in OSMReader. The WaySegmentParser only provides a callback 'handleEdge' that tells us that it found a segment that we can/should create an edge for. The WaySegmentParser doesn't even know anything about the graph, it could be used to create an entirely different graph structure, i.e. store the graph as rows in an SQL database or something.

> Additionally, the WaySegmenthandler hadles 1. the OSM parsing, 2. node handling, 3. way handling, 4. relation handling 

Yes, it does initiate the reading of the OSM file. We could move this out and feed a stream of OSM elements to WaySegmentParser instead, but so far I saw no reason to do this. It would be needed if we wanted to build the graph from OSM that we get via http instead of reading a local file, but that would be entirely unrelated to turn restrictions of course. And yes, it also does handle nodes and ways, but there is no way around this, because its purpose is to find the junctions and segments of the road network. I did not mention the relations so far, but to create edges of the routing graph we typically want to know if the corresponding OSM ways are referenced by certain relations like bike networks etc., which is the reason we also read them in WaySegmentParser. The other reason is that we don't want to read the OSM file again after running the WaySegmentParser.

> (by getting those functions from OSMReader!?)

The WaySegmentParser provides callbacks that are used by OSMReader, yes. This is how it can be used for different purposes and at the same time provide an abstraction that does all the segment splitting stuff.

> and on top of it now the Restriction Handling would also be placed

Ok, now we are getting back to the present issue :) I absolutely see how WaySegmentParser and OSMReader can seem very complex (and in parts they are), so adding more things to them can seem like and possibly is the wrong thing to do. I hope I am not missing any details, but so far I think WaySegmentParser is/will be entirely unrelated to turn restrictions. And it should be possible to move the turn restriction handling into a separate class where it also can be tested independently from the other things that are going on during the OSM reading. Basically we should be able to build the entire graph without any turn restrictions, only save a list of turn relations and a mapping between the way IDs and graph edge Ids they use, and then add all the turn restrictions afterwards. For example we can easily add the artificial turn restriction edges after the normal graph building is complete. Especially with the overlapping restrictions this might be needed anyway, because we need to add the restrictions/artificial edges at a point in time where we know about all the restrictions already? I'm not sure yet if it would be possible to simply add them one after the other as if they were independent/non-overlapping.

All that being said, I'm not strictly against changing anything about WaySegmentParser and OSMReader and there absolutely is room for improvement. Especially OSMReader is still a somewhat inflexible huge monolith of many different things. The builder in WaySegmentParser is also not ideal, which is something I wanted to fix before, but either forgot about it or ran into some problem, I can have a look again. So feel free to open a PR for your refactorings that we can discuss. Shall we do this first? I can try to re-arrange the code in the way that I think would be appropriate for the new turn restriction handling in the meantime.

lukasalexanderweber:
As I'm on training the whole next week I just wan't to give a quick response, while not answering everything in detail just yet.

> Anyway, shall we first focus on single via-ways, then see if there are actual problems for the overlapping cases and then either try to fix the overlaps or generalize to multi via-ways?

This sounds good, I was just enthusiastic that multi via ways become rather simple :)

> Ok, let me explain the 'idea' behind the OSMReader and related code

Thanks, I've worked out most of what you said, but it probably took longer than it should

> I'll have a look at yours of course.

I want to give you a short outline on how the dataflow is regarding the implementation of single via way restrictions in https://github.com/lukasalexanderweber/graphhopper/tree/single_via_way_restrictions:

[OSMReader.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java) will trigger two parsings of [OSMParser.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/OSMParser.java). 

The first parse is to preproccess ways ([WayPreprocessor.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/WayPreprocessor.java)) and Relations ([RelationPreprocessor.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/RelationPreprocessor.java)). During Relation Preprocessing, via way turn restriction information are stored ([WayRestriction.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/WayRestriction.java) in [OSMTurnRestrictionData.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/OSMTurnRestrictionData.java)).

In the second parse, first the nodes are handled ([NodeHandler.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/NodeHandler.java)). Then the ways are handled ([WayHandler.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/WayHandler.java)), storing the [ReaderWay.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/ReaderWay.java) objects of all via ways (to create artificial edges later on). In the onStart function of the [RelationHandler.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/RelationHandler.java), the turn restrictions are build ([OSMTurnRestrictionBuilder.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/OSMTurnRestrictionBuilder.java)), creating the artificial edges and the according via node restrictions ([NodeRestriction.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/NodeRestriction.java)). Those are then processed if the respective [OSMTurnRestriction.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/OSMTurnRestriction.java) is processed again in the [RelationHandler.java](https://github.com/lukasalexanderweber/graphhopper/blob/single_via_way_restrictions/core/src/main/java/com/graphhopper/reader/osm/RelationHandler.java).

Please don't hestitate to write me on any questions.

easbar:
Thanks for the update.

> This sounds good, I was just enthusiastic that multi via ways become rather simple :)

Yes, this is really nice.

> I want to give you a short outline on how the dataflow is regarding the implementation of single via way restrictions in 

Ok thanks I'll have a look.

easbar:
I started to think a lot more about this issue, took a look at https://github.com/lukasalexanderweber/graphhopper/tree/single_via_way_restrictions and ended up trying to combine everything we know so far in another implementation: #2689

The main differences are probably the following:

* The turn restrictions are added to the graph separately from the OSM 'reading' process. I felt like you were trying to use the OSMReader#addEdge method to create the artificial edges (and probably this is what led you to the refactoring of the OSMReader?), but I think the better option is to simply copy the edges later. This way the restriction handling is very much decoupled from the OSM reading graph building and also easier to test.

* One non-trivial issue was that there is no 1:1 mapping between the OSM way members of the restriction relations and the edges we add to the routing graph. I think you simply excluded these cases so far, but I wrote code that handles these as well.

* I think I found a solution to the 'overlapping restrictions' problem. The examples you found seem to work correctly in my implementation. All we need to do apparently is making sure that there is only one artificial edge for each real edge and that we deny turns between artificial from/to edges onto the artificial via-edge. I can explain this in more detail, but maybe first see if it is actually working.

I'm sure there is still room for lots of improvements and if you also like the new implementation maybe we can transfer some stuff from yours. Also do not hesitate to open the PR for the OSMReader if you think it is an improvement in itself. For this issue I did not need to change much in the OSMReader code though (see #2689 for more details).

lukasalexanderweber:
Thanks for reaching out and all the work you've put into this! ❤️‍🔥 

> I felt like you were trying to use the OSMReader#addEdge method to create the artificial edges (and probably this is what led you to the refactoring of the OSMReader?)

Yeah since I needed to add the edges, and the functionality was already there, I tried to make use of OSMReader#addEdge. I definitely agree that this can be decoupled from the reading.

> One non-trivial issue was that there is no 1:1 mapping between the OSM way members of the restriction relations and the edges we add to the routing graph. I think you simply excluded these cases so far, but I wrote code that handles these as well.

Yeah until then I had assumed that all (via) ways have only two tower nodes (for me those are the edges that really have a 1:1 mapping to the OSM Ways). Example Restrictions with more than two tower nodes are `6315945` and `10164406`, which are both managed correctly by your implementation

> I think I found a solution to the 'overlapping restrictions' problem.

Yes, this works perfectly fine. I tested with 
- `2602528 2602529 2602530`
- `2821560 2821561 2821562 2821563`
- `2920825 2920826 2920827 2920828`
- `2994746 2994747 2994749 2994750`

> I'm sure there is still room for lots of improvements and if you also like the new implementation maybe we can transfer some stuff from yours

I'll have a closer look at your implementation, i'm curious to see how you did it :) 

> Also do not hesitate to open the PR for the OSMReader if you think it is an improvement in itself

I think I will, but probably after the work on restriction handling is finished so that there are not so much merge conflicts.

PR #2684, opened 2022-10-24
Improve area check precision

otbutz:
Fixes #2683

We should only early-out if the rectangle fully contains the bounding box of the edge. This still avoids costly way geometry checks in most cases but improves accuracy for checks near the borders of the polygons bounding box.

Issue #2683, opened 2022-10-24
Custom areas are imprecise


otbutz:
I'm trying to use a custom area to describe a simple roadblock. But apparently the small boundingbox also blocks a part of the B11 which is ~110m away: [GraphHopper Maps](https://graphhopper.com/maps/?point=49.033675%2C13.234615&point=49.021885%2C13.248094&profile=car&layer=Omniscale&custom_model=%7B%22priority%22%3A%5B%7B%22if%22%3A%22in_road_block%22%2C%22multiply_by%22%3A%220%22%7D%5D%2C%22areas%22%3A%7B%22road_block%22%3A%7B%22type%22%3A%22Feature%22%2C%22properties%22%3A%7B%7D%2C%22geometry%22%3A%7B%22type%22%3A%22Polygon%22%2C%22coordinates%22%3A%5B%5B%5B13.240481423585088%2C49.0267149660803%5D%2C%5B13.240481423585088%2C49.02662503391971%5D%2C%5B13.240618576414914%2C49.02662503391971%5D%2C%5B13.240618576414914%2C49.0267149660803%5D%2C%5B13.240481423585088%2C49.0267149660803%5D%5D%5D%7D%7D%7D%7D)

<details>
  <summary>Custom model</summary>

```json
{
  "priority": [
    {
      "if": "in_roadblock",
      "multiply_by": "0"
    }
  ],
  "areas": {
    "roadblock": {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              13.240481423585088,
              49.0267149660803
            ],
            [
              13.240481423585088,
              49.02662503391971
            ],
            [
              13.240618576414914,
              49.02662503391971
            ],
            [
              13.240618576414914,
              49.0267149660803
            ],
            [
              13.240481423585088,
              49.0267149660803
            ]
          ]
        ]
      }
    }
  }
}
```

</details>

Visualized via geojson.io:

![geojson](https://user-images.githubusercontent.com/22315436/197520337-8100e922-f1d8-4974-a52c-2e4f23a62b62.png)

Might be related to #2625

otbutz:
OSM way: https://www.openstreetmap.org/way/118994937#map=17/49.02827/13.24110

![image](https://user-images.githubusercontent.com/22315436/197522528-e5608ee2-214c-4249-81eb-bbb5ef987342.png)

[...]

PR #2685, opened 2022-10-24
Reuse polygon bbox

otbutz:
Currently we recreate the bbox of the polygon on each call because the object is mutable. This PR adapts the immutable  PointList approach for the BBox and adds caching.


easbar:
Honestly, I'm not a big fan of the PointList class and/or the 'makeImmutable' thing. Where and why do we create the bbox on each call (since it is mutable)? And do you have an example where this BBox creation actually causes a slow-down?

otbutz:
> Where and why do we create the bbox on each call (since it is mutable)?

Every call of `Polygon.getBounds()` creates a new BBox:
https://github.com/graphhopper/graphhopper/blob/37380c01ac115b2fd11f4293b964eea3fde64c79/core/src/main/java/com/graphhopper/util/shapes/Polygon.java#L86-L88

The Polygon itself is immutable though. There's no real reason that the BBox needs to be recreated on each call apart from it being potentially modified by a caller.

Funnily enough we don't do this for a Circle:

https://github.com/graphhopper/graphhopper/blob/37380c01ac115b2fd11f4293b964eea3fde64c79/core/src/main/java/com/graphhopper/util/shapes/Circle.java#L63-L65

> And do you have an example where this BBox creation actually causes a slow-down?

We needlessly create `edges * areas` BBox instances by calling `Polygon.getBounds()` here:

https://github.com/graphhopper/graphhopper/blob/37380c01ac115b2fd11f4293b964eea3fde64c79/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java#L78-L85

All these instances need to be created and later garbage collected.

> Honestly, I'm not a big fan of the PointList class and/or the 'makeImmutable' thing.

Me neither. I can also take a stab at making the BBox truly immutable with all setters and update methods returning new BBox instances. The other solution would be to keep the BBox of Polygon as we do it for the Circle and hope the caller doesn't mess with it.

easbar:
> We needlessly create edges * areas BBox instances by calling Polygon.getBounds() here:

Ok, I see. I still wonder if that really matters. GC can take time, yes, but oftencase it doesn't when objects are GC'ed immediately and never get past the young generation.

> Funnily enough we don't do this for a Circle:

Ah and you mean we could do the same for the Polygon: Create `new BBox(envelope.getMinX(), envelope.getMaxX(), envelope.getMinY(), envelope.getMaxY());` once as an instance variable and return it in `getBounds()`? 

> apart from it being potentially modified by a caller.

You mean a caller could do `p.getBounds().minLat = 20` and from this point p.getBounds() would no longer be the correct bounding box of the polygon? If you ask me there is no need to protect our code from us doing silly things with it.

>  I can also take a stab at making the BBox truly immutable with all setters and update methods returning new BBox instances.

I don't see why this would be useful, but maybe I am missing the point.

>  and hope the caller doesn't mess with it.

The other solution would be to keep the BBox of Polygon as we do it for the Circle and hope the caller doesn't mess with it.

This. I don't think we need to hope for anything. The code is what it is. If someone decides to change the bounding box of the polygon the bounding box of the polygon will be no longer the bounding box of the polygon, but whatever the caller decided it should be.

Btw for PointList it is different again, because we have an extra intersects method...

https://github.com/graphhopper/graphhopper/blob/37380c01ac115b2fd11f4293b964eea3fde64c79/core/src/main/java/com/graphhopper/util/shapes/Polygon.java#L70-L72

otbutz:
> You mean a caller could do `p.getBounds().minLat = 20` and from this point p.getBounds() would no longer be the correct bounding box of the polygon? If you ask me there is no need to protect our code from us doing silly things with it.

Sounds reasonable to me. Other projects tend to be a lot stricter in that regard.

> Btw for PointList it is different again, because we have an extra intersects method...

Yeah we still have room for improvement. e.g we still recreate the BBox of an edge for each area check:

https://github.com/graphhopper/graphhopper/blob/37380c01ac115b2fd11f4293b964eea3fde64c79/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java#L84

otbutz:
@easbar switched to the simpler approach

easbar:
> Other projects tend to be a lot stricter in that regard.

Which ones?

> we still recreate the BBox and geometry of an edge for each area check:

Where do we recreate the BBox here? For the geometry of an edge we would probably have to fetch the geometry first and then loop over all areas. But not sure how easy that is because the `in_area` expressions could occur in different custom model rules and so far we just treat them independently from each other.

easbar:
> switched to the simpler approach

Ok, thanks, but I'd like to leave this up to karussell. I haven't been involved much with this code.

otbutz:
> > Other projects tend to be a lot stricter in that regard.
> 
> Which ones?

Just to list a few articles on the matter:

https://medium.com/@rufuszh90/effective-java-item-15-minimise-mutability-2526108ac7f1
https://ncrcoe.gitbooks.io/java-for-small-teams/content/style/90_prefer_immutable_objects.html
https://blogs.oracle.com/javamagazine/post/java-immutable-objects-strings-date-time-records

The general consensus is to prefer immutable objects whenever possible.

> > we still recreate the BBox and geometry of an edge for each area check:
> 
> Where do we recreate the BBox here? For the geometry of an edge we would probably have to fetch the geometry first and then loop over all areas. But not sure how easy that is because the `in_area` expressions could occur in different custom model rules and so far we just treat them independently from each other.

Each call to `CustomWeightingHelper.in()` will at least create a BBox for the edge based on its tower nodes and might also need to load the whole way geometry. As you already pointed out, this could be optimized for multiple `in_area` expressions. See https://github.com/graphhopper/graphhopper/pull/2327

easbar:
> The general consensus is to prefer immutable objects whenever possible.

Ok, thanks for the links. I think I'm all for not mutating object state wherever possible and in our case we don't want to mutate anything either. But I often it is not necessary to write protective code that enforces this at all costs (like protecting all kinds of getters, copying stuff in setters and so on). I just like to believe we can write sensible code and think we can accept that it breaks if we do something silly as using `getBounds()` to change the internal bbox values.

easbar:
Yes, I think we already agreed the bbox (based on all points of an edge) should just be stored along with the edge, didn't we? But then we still need to fetch the geometry for the detailed `p.intersects(edge.fetchWayGeometry(FetchMode.ALL)` check, right? To avoid the latter we would have to fetch the geometry only once (per edge) and re-use it for all in-area checks when we calculate the edge weight. Maybe something like this?

```java
// in custom weighting

public double calcEdgeWeight(EdgeIteratorState edgeState, boolean reverse) {
  // ... somewhere:
  PointList pointList = edgeState.fetchGeometry(); // but maybe only in case the custom model even uses areas...
  // ... pass this pointList to the code that does the single area checks...
```

otbutz:
https://github.com/graphhopper/graphhopper/pull/2327 implements this approach but adds state to the weighting. Maybe we should move this into some kind of throwaway wrapper?

easbar:
> https://github.com/graphhopper/graphhopper/pull/2327 implements this approach but adds state to the weighting. Maybe we should move this into some kind of throwaway wrapper?

Yes, maybe. But probably also something to discuss with karussell then :)

otbutz:
We really need a proper place for this kind of intermediate results. So far we ruled out the [weighting](https://github.com/graphhopper/graphhopper/pull/2327) and the [EdgeIteratorState](https://github.com/graphhopper/graphhopper/pull/2338).

otbutz:
Maybe we should extend the `CustomWeighting.EdgeToDoubleMapping` and pass some kind of lazily populated `EdgeGeometry` along:

https://github.com/graphhopper/graphhopper/blob/c943724414c782a1520d09d1e6ea8802f51236c3/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeighting.java#L152-L155

easbar:
> We really need a proper place for this kind of intermediate results. So far we ruled out the https://github.com/graphhopper/graphhopper/pull/2327 and the https://github.com/graphhopper/graphhopper/pull/2338.

Can you take one step back and explain why we really need this? It seems like in https://github.com/graphhopper/graphhopper/pull/2327 the conclusion (on our end) was that the speedup isn't really worth complicating things?

And what if we started with what we already decided: Store the edge bbox in the graph? Would this (not) lead to a speedup?

otbutz:
> Can you take one step back and explain why we really need this?

This completely avoids the repeated creation of BBox instances and thus the interaction with NodeAccess/DataAccess per area and edge combination. The speed increase is therefore only relevant for more complex CustomModels with multiple areas. But if we want to push CustomModels as the way forward for dynamic routing, we should also address scalability for these usecases.

I am currently testing this feature to model [road closures](https://github.com/graphhopper/graphhopper/issues/2683) and high traffic areas. With 30-40 areas in a CustomModel, the overhead per visited edge is no longer negligible.

> It seems like in #2327 the conclusion (on our end) was that the speedup isn't really worth complicating things?

I agree with @karussell's assessment, as the EdgeIteratorState interface is already quite bloated (https://github.com/graphhopper/graphhopper/issues/2691) and previous attempts would have introduced too much complexity compared to the potential benefit.

> And what if we started with what we already decided: Store the edge bbox in the graph? Would this (not) lead to a speedup?

That would certainly improve the precision as noted in https://github.com/graphhopper/graphhopper/pull/2684#issuecomment-1289009595 but we would still need to create those BBox instances by reading the values from DataAccess.

otbutz:
Moving this to the server side would certainly be more efficient, but the difficulty is that [we](https://github.com/OPTITOOL/) have to deal with the data provided by the user. A road may be closed to the public due to construction, but it may also be impassable to certain types of vehicles or companies without permission. So it is intended as a simple tool for the user to influence the routing and to work around missing OSM data.

Our rather special usecase aside, complex server-side models would profit the most.

> There could be hundreds of road closures

Fun fact: the number of areas is currently limited to about ~2k. The code generation has to populate a field for each area which hits the [maximum length of a Java method](https://www.baeldung.com/java-code-too-large-error) at a certain point:

```
Caused by: org.codehaus.commons.compiler.InternalCompilerException: Compiling "init(EncodedValueLookup lookup, com.graphhopper.routing.ev.DecimalEncodedValue avgSpeedEnc, com.graphhopper.routing.ev.DecimalEncodedValue priorityEnc, Map<String, com.graphhopper.util.JsonFeature> areas)": Code grows beyond 64 KB
```

But i don't think that this is a limitation for any sane usecase :sweat_smile: 

> My conclusion was more that caching is very tricky and we should get a lot of benefits before we do it.

It is certainly difficult/ugly if we handle it in the core logic like the EdgeStateIteratorImpl. #2688 tries to avoid this by limiting the scope to `CustomWeighting.calcEdgeWeight()` and `CustomWeighting.calcEdgeMillis()`

otbutz:
I hope to get around to building a minimal reproducible example next week.

otbutz:
@karussell just noticed this PR is still open. We've gotten pretty off topic with the discussion, but I think the current changes are pretty reasonable.

easbar:
Do we now have an example that shows the speed difference in absolute terms such that we can see the improvements of this or other optimizations?

easbar:
With regard to this specific PR I don't really see a reason why we should not create the BBox in the Polygon constructor here as suggested.

otbutz:
I've finally managed to gather some numbers.

## Test setup

<details>
  <summary>Custom Model</summary>


"Simple" profile to emulate 20 roadblocks:

  ```json
{
    "distance_influence": 15,
    "speed": [],
    "priority": [
        {
            "if": "in_point_0 || in_point_1 || in_point_2 || in_point_3 || in_point_4 || in_point_5 || in_point_6 || in_point_7 || in_point_8 || in_point_9 || in_point_10 || in_point_11 || in_point_12 || in_point_13 || in_point_14 || in_point_15 || in_point_16 || in_point_17 || in_point_18 || in_point_19",
            "multiply_by": "0.0"
        }
    ],
    "areas": {
        "point_0": {
            "id": "point_0",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            5.250657236524162,
                            51.8314750339197
                        ],
                        [
                            5.250657236524162,
                            51.83156496608029
                        ],
                        [
                            5.250802763475837,
                            51.83156496608029
                        ],
                        [
                            5.250802763475837,
                            51.8314750339197
                        ],
                        [
                            5.250657236524162,
                            51.8314750339197
                        ]
                    ]
                ]
            }
        },
        "point_1": {
            "id": "point_1",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            5.434074998680752,
                            51.916581033919705
                        ],
                        [
                            5.434074998680752,
                            51.9166709660803
                        ],
                        [
                            5.434220801319248,
                            51.9166709660803
                        ],
                        [
                            5.434220801319248,
                            51.916581033919705
                        ],
                        [
                            5.434074998680752,
                            51.916581033919705
                        ]
                    ]
                ]
            }
        },
        "point_2": {
            "id": "point_2",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            6.915652465248339,
                            51.0537310339197
                        ],
                        [
                            6.915652465248339,
                            51.053820966080295
                        ],
                        [
                            6.915795534751661,
                            51.053820966080295
                        ],
                        [
                            6.915795534751661,
                            51.0537310339197
                        ],
                        [
                            6.915652465248339,
                            51.0537310339197
                        ]
                    ]
                ]
            }
        },
        "point_3": {
            "id": "point_3",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            6.917187462695051,
                            51.0553838339197
                        ],
                        [
                            6.917187462695051,
                            51.05547376608029
                        ],
                        [
                            6.917330537304948,
                            51.05547376608029
                        ],
                        [
                            6.917330537304948,
                            51.0553838339197
                        ],
                        [
                            6.917187462695051,
                            51.0553838339197
                        ]
                    ]
                ]
            }
        },
        "point_4": {
            "id": "point_4",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            6.961388795228124,
                            51.034312733919705
                        ],
                        [
                            6.961388795228124,
                            51.0344026660803
                        ],
                        [
                            6.961531804771876,
                            51.0344026660803
                        ],
                        [
                            6.961531804771876,
                            51.034312733919705
                        ],
                        [
                            6.961388795228124,
                            51.034312733919705
                        ]
                    ]
                ]
            }
        },
        "point_5": {
            "id": "point_5",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            6.961605594744377,
                            51.0346262339197
                        ],
                        [
                            6.961605594744377,
                            51.034716166080294
                        ],
                        [
                            6.961748605255622,
                            51.034716166080294
                        ],
                        [
                            6.961748605255622,
                            51.0346262339197
                        ],
                        [
                            6.961605594744377,
                            51.0346262339197
                        ]
                    ]
                ]
            }
        },
        "point_6": {
            "id": "point_6",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            6.971941653884413,
                            50.9311880339197
                        ],
                        [
                            6.971941653884413,
                            50.931277966080295
                        ],
                        [
                            6.972084346115587,
                            50.931277966080295
                        ],
                        [
                            6.972084346115587,
                            50.9311880339197
                        ],
                        [
                            6.971941653884413,
                            50.9311880339197
                        ]
                    ]
                ]
            }
        },
        "point_7": {
            "id": "point_7",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            6.972160653655853,
                            50.931337033919704
                        ],
                        [
                            6.972160653655853,
                            50.931426966080295
                        ],
                        [
                            6.972303346344147,
                            50.931426966080295
                        ],
                        [
                            6.972303346344147,
                            50.931337033919704
                        ],
                        [
                            6.972160653655853,
                            50.931337033919704
                        ]
                    ]
                ]
            }
        },
        "point_8": {
            "id": "point_8",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            6.975031617651083,
                            50.954792933919705
                        ],
                        [
                            6.975031617651083,
                            50.9548828660803
                        ],
                        [
                            6.975174382348917,
                            50.9548828660803
                        ],
                        [
                            6.975174382348917,
                            50.954792933919705
                        ],
                        [
                            6.975031617651083,
                            50.954792933919705
                        ]
                    ]
                ]
            }
        },
        "point_9": {
            "id": "point_9",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            6.9755606184358285,
                            50.954282033919704
                        ],
                        [
                            6.9755606184358285,
                            50.954371966080295
                        ],
                        [
                            6.975703381564172,
                            50.954371966080295
                        ],
                        [
                            6.975703381564172,
                            50.954282033919704
                        ],
                        [
                            6.9755606184358285,
                            50.954282033919704
                        ]
                    ]
                ]
            }
        },
        "point_10": {
            "id": "point_10",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            7.174242016382047,
                            50.693264033919704
                        ],
                        [
                            7.174242016382047,
                            50.693353966080295
                        ],
                        [
                            7.174383983617952,
                            50.693353966080295
                        ],
                        [
                            7.174383983617952,
                            50.693264033919704
                        ],
                        [
                            7.174242016382047,
                            50.693264033919704
                        ]
                    ]
                ]
            }
        },
        "point_11": {
            "id": "point_11",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            7.188597050312256,
                            50.670828033919705
                        ],
                        [
                            7.188597050312256,
                            50.670917966080296
                        ],
                        [
                            7.188738949687743,
                            50.670917966080296
                        ],
                        [
                            7.188738949687743,
                            50.670828033919705
                        ],
                        [
                            7.188597050312256,
                            50.670828033919705
                        ]
                    ]
                ]
            }
        },
        "point_12": {
            "id": "point_12",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            7.195945678825479,
                            51.5555188339197
                        ],
                        [
                            7.195945678825479,
                            51.55560876608029
                        ],
                        [
                            7.19609032117452,
                            51.55560876608029
                        ],
                        [
                            7.19609032117452,
                            51.5555188339197
                        ],
                        [
                            7.195945678825479,
                            51.5555188339197
                        ]
                    ]
                ]
            }
        },
        "point_13": {
            "id": "point_13",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            7.196025678260859,
                            51.5558739339197
                        ],
                        [
                            7.196025678260859,
                            51.55596386608029
                        ],
                        [
                            7.1961703217391415,
                            51.55596386608029
                        ],
                        [
                            7.1961703217391415,
                            51.5558739339197
                        ],
                        [
                            7.196025678260859,
                            51.5558739339197
                        ]
                    ]
                ]
            }
        },
        "point_14": {
            "id": "point_14",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            7.20881510873428,
                            50.6321298339197
                        ],
                        [
                            7.20881510873428,
                            50.63221976608029
                        ],
                        [
                            7.20895689126572,
                            50.63221976608029
                        ],
                        [
                            7.20895689126572,
                            50.6321298339197
                        ],
                        [
                            7.20881510873428,
                            50.6321298339197
                        ]
                    ]
                ]
            }
        },
        "point_15": {
            "id": "point_15",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            7.274532210335119,
                            50.564626933919705
                        ],
                        [
                            7.274532210335119,
                            50.564716866080296
                        ],
                        [
                            7.274673789664881,
                            50.564716866080296
                        ],
                        [
                            7.274673789664881,
                            50.564626933919705
                        ],
                        [
                            7.274532210335119,
                            50.564626933919705
                        ]
                    ]
                ]
            }
        },
        "point_16": {
            "id": "point_16",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            7.3879338865414965,
                            51.360899333919704
                        ],
                        [
                            7.3879338865414965,
                            51.360989266080296
                        ],
                        [
                            7.388077913458503,
                            51.360989266080296
                        ],
                        [
                            7.388077913458503,
                            51.360899333919704
                        ],
                        [
                            7.3879338865414965,
                            51.360899333919704
                        ]
                    ]
                ]
            }
        },
        "point_17": {
            "id": "point_17",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            7.408931370939574,
                            53.236138233919704
                        ],
                        [
                            7.408931370939574,
                            53.236228166080295
                        ],
                        [
                            7.409081629060425,
                            53.236228166080295
                        ],
                        [
                            7.409081629060425,
                            53.236138233919704
                        ],
                        [
                            7.408931370939574,
                            53.236138233919704
                        ]
                    ]
                ]
            }
        },
        "point_18": {
            "id": "point_18",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            7.478142993158795,
                            51.356690033919705
                        ],
                        [
                            7.478142993158795,
                            51.356779966080296
                        ],
                        [
                            7.478287006841204,
                            51.356779966080296
                        ],
                        [
                            7.478287006841204,
                            51.356690033919705
                        ],
                        [
                            7.478142993158795,
                            51.356690033919705
                        ]
                    ]
                ]
            }
        },
        "point_19": {
            "id": "point_19",
            "type": "Feature",
            "geometry": {
                "type": "Polygon",
                "coordinates": [
                    [
                        [
                            7.524703991644336,
                            51.293921933919705
                        ],
                        [
                            7.524703991644336,
                            51.2940118660803
                        ],
                        [
                            7.524847808355664,
                            51.2940118660803
                        ],
                        [
                            7.524847808355664,
                            51.293921933919705
                        ],
                        [
                            7.524703991644336,
                            51.293921933919705
                        ]
                    ]
                ]
            }
        }
    }
}
  ```
  
</details>

I've imported the latest Germany PBF using GraphHopper 6.2:

```sh
java -Xms16G -Xmx16G -Ddw.graphhopper.datareader.file=germany-latest.osm.pbf -jar *.jar server config-example.yml
```

I test this with a route from `48.123298,11.543013` to `52.363195,9.797438` with a distance of ~658km. After repeating the request several times, the execution time is ~5200ms according to the GH log.

When all area checks except the very first one are dropped, the execution time drops to about 3100ms.

An empty custom model with no priority statements takes about 2600ms.

## JFR

```sh
java -Xms16G -Xmx16G -XX:+UnlockDiagnosticVMOptions -XX:FlightRecorderOptions=stackdepth=1024 -XX:+DebugNonSafepoints -Ddw.graphhopper.datareader.file=germany-latest.osm.pbf -jar *.jar server config-example.yml
```

## Memory allocation

image

## Method profiling

image

otbutz:
I also built a patched version of GraphHopper that reuses a single BBox instance per iterator for comparison. This lowers the execution time for the original 20-area test case to ~3300ms, which is a whopping 36% speed increase.

## Memory allocation

image


## Method profiling

image

## Garbage collections

### Before

image


### After

image

otbutz:
[flamegraph.zip](https://github.com/graphhopper/graphhopper/files/10722263/flamegraph.zip)

otbutz:
@karussell @easbar what's your opinion?

Issue #1287, opened 2018-02-15
Improve wording when "continuing onto" motorways..


jguthrie100:
Just a small request, but the language seems a bit strange when driving along a long highway/motorway and being told to continually "continue onto E42" and "Keep left onto E42" for example.

I think "Continue on E42" and "Keep left on E42" would sound much better!

karussell:
Thanks for the recommendation regarding the wording. Regarding the instruction count, this is currently investigated a bit in #1284

boldtrn:
Thanks for the proposal @jguthrie100. You are right, "keep left on" and "continue on" sounds better.

What do you think about "Turn left on main street". Because I think "Turn left onto main street" sounds better?

jguthrie100:
Yep, "turn left onto Main Street" is better!

boldtrn:
Thanks. @karussell how should we approach this issue? 

Should we create a new translation "turn_on". We would then use this translation for keeps and continues? Due to that we might be able to get rid of the "continue_onto" translation. WDYT?

karussell:
For English we might be able to do this simplification but I'm not sure if this is really the same and easy for other languages too. Why not just change the English translation from "onto" into "on"?

devemux86:
Agreed with @karussell to not complicate things so much.
If en translation can be improved, then can simply change that.

michaz:
> Why not just change the English translation from "onto" into "on"?

@karussell 
Because you "turn **onto**" a road, but you "keep left **on**" a road, but the two are currently using the same template "%1$s onto %2$s".

Only "continue onto" has its own entry in the localization table.

karussell:
Ah, ok. Thanks for taking the time to explain that. Makes sense to have it.

ethanotran:
Can I have a go at this issue? Trying to make my first open source contribution.

CutestNekoAqua:
@ethanotran sure! No need to ask! If you wanna do it, just do it :)

cheehwatang:
Hi, I edited the "continue onto %1$s" to "continue on %1$s" in the excel sheet.

However, I faced some issue when I run the script ./core/files/update-translations.sh to update.
It cannot find the txt files.

Error message:
User MINGW64 ~/Documents/Git/graphhopper
$ ./core/files/update-translations.sh
tail: cannot open '' for reading: No such file or directory
tail: cannot open '' for reading: No such file or directory
tail: cannot open '' for reading: No such file or directory
...

[...]

PR #2687, opened 2022-10-26
Improve Wording in en_US.txt

cheehwatang:
From issue https://github.com/graphhopper/graphhopper/issues/1287, the working for Continue Onto is changed to Continue On for better grammar. However, Keep Left Onto or Keep Right Onto is not fixed.

It seems that the other lines are updated in the Excel sheet, but not sync. So I have included the changes below as well.


cheehwatang:
Hi, this is my first time contributing. I mainly followed the issue #1287 and the guidelines in updating translations.
Not sure what I did wrong. Thanks.

PR #2698, opened 2022-11-19
Replace com.wdtinc MVT library with no.ecc

easbar:
https://github.com/wdtinc/mapbox-vector-tile-java seems very inactive and did not accept pull requests that updated the JTS version.
Here I used the MVT generation code from https://github.com/ElectronicChartCentre/java-vector-tile which according to my measurements even runs faster (something like 44ms vs. 78ms).  Unfortunately it is [not deployed to Maven Central](https://github.com/ElectronicChartCentre/java-vector-tile/issues/31) so I copied the code. This way we can control the JTS version ourselves. But currently this isn't so critical as they use a recent JTS version anyway (1.19).
Note that there is a modified and even faster version (around 38ms) [in planetiler](https://github.com/onthegomap/planetiler/blob/5296d1772e5929ef741e3a576192978d8bfc26ce/planetiler-core/src/main/java/com/onthegomap/planetiler/VectorTile.java#L56-L69), but for our purposes the speed isn't so critical, because a lot of time is spent on querying location index and rendering the road network anyway.

Related: #1803, https://github.com/graphhopper/graphhopper/commit/0f96c2deddb24efa97109e35e0c05f1c91221f59#r90066911


web-bundle/pom.xml
```diff
@@ -64,10 +64,10 @@
             <artifactId>dropwizard-testing</artifactId>
             <scope>test</scope>
         </dependency>
-
         <dependency>
-            <groupId>com.wdtinc</groupId>
-            <artifactId>mapbox-vector-tile</artifactId>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+            <version>3.21.9</version>
```
easbar, line 70: For some reason we need this to make the protobuf generated code in VectorTile.java compile (without it there is a compiler error due to the `parseUnknownField` method in VectorTile.java). Note that previously we excluded `protobuf-java` from com.wdtinc:mapbox-vector-tile explicitly. (?)

easbar:
> something like 44ms vs. 78ms

This could be due to the fact that wdtinc does some Douglas-Peucker-like geometry simplification which is also included in no.ecc, but disabled by default.

web-bundle/src/main/java/no/ecc/vectortile/VectorTile.java
```diff
@@ -0,0 +1,5735 @@
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
```
easbar, line 1: Instead of including this large file we could also just include the much smaller vector_tile.proto file and generate this using the protoc Maven plugin. But unfortunately this does not seem to be recognized by IntelliJ and one has to run `mvn compile` manually first?!

otbutz:
> Unfortunately it is [not deployed to Maven Central](https://github.com/ElectronicChartCentre/java-vector-tile/issues/31) so I copied the code.

Since the repository is globally available, why don't we just add it to our parent pom?

easbar:
> Since the repository is globally available, why don't we just add it to our parent pom?

One thing I found very annoying when I used another repository (the one needed for geotools in this case) was that Maven was trying to download artifacts from this repository before it tried Maven Central. And since the geotools repository was down or at least very slow sometimes the build took forever. But not sure if we can force Maven to try Maven Central first and maybe the ecc repository has better availability. We can try and if it makes problems we can still copy the code, I added the external repository in my last commit.

otbutz:
You can explicitly add Maven central to enforce the order:

```xml
<repositories>
  <repository>
    <id>central</id>
    <name>Central Repository</name>
    <url>https://repo1.maven.org/maven2</url>
  </repository>
  <repository>
    <id>ECC</id>
    <url>https://ecc-mvn.ams3.digitaloceanspaces.com/releases</url>
  </repository>
</repositories>
```

easbar:
> You can explicitly add Maven central to enforce the order:

ok thanks I tried this

Issue #1858, opened 2020-01-21
Potential u-turns at junctions with inaccessible roads


easbar:
When e.g. a footpath crosses a road there will be a tower-node for this junction. If u-turn costs are set to a finite value this means that the shortest path can include a u-turn at this junction, even if its not really a junction for the current vehicle (e.g. car).

![image](https://user-images.githubusercontent.com/17603532/72796503-b4ac0500-3c3f-11ea-9834-33f6bc6d2e20.png)

Avoiding such u-turns would most likely be an improvement, but there can still be u-turns at 'real' junctions that aren't desired either.

easbar:
Another (rather severe) case where this could happen are ferries that are crossing bridges (c.f. #2139).

PR #2702, opened 2022-11-29
Use explicit ignore list for highway values

easbar:
So far we used `VehicleTagParser#getAccess` to determine whether an OSM way shall be accepted when parsing the OSM file. With this PR all highways will be accepted except those listed in a global list of ignored values ~~there will be only one global list of accepted highway values~~. It is still up to the vehicle tag parsers (and in the near future only up to the access tag parsers) whether a given way is used for the route calculations. Putting a highway value into the global list is therefore mostly a performance optimization. So far this happened automatically by only using certain vehicles. For example if we used only profiles with the car profile footways were rejected automatically. With the new global list this needs to be done manually, but at the same time the mechanism becomes more powerful and easier to understand IMO.

See more comments inline.

Related: https://github.com/graphhopper/graphhopper/issues/2463#issuecomment-1174205767

### DIFF
--- a/config-example.yml
+++ b/config-example.yml
@@ -171,7 +171,7 @@ graphhopper:
 
   # Excludes certain types of highways during the OSM import to speed up the import and reduce the size of the graph.
   # A typical case is excluding footway,cycleway,path and maybe pedestrian and track for motorized vehicles. This leads
-  # to a smaller graph, because there are less ways (obviously), but also because there are fewer junctions.
+  # to a smaller graph, because there are fewer ways (obviously), but also because there are fewer junctions.
   # Another example is excluding motorway, trunk and maybe primary highways for bicycle or pedestrian routing.
   import.osm.ignored_highways: footway,cycleway,path,pedestrian,steps
 
--- a/config-example.yml
+++ b/config-example.yml
@@ -5,6 +5,32 @@ graphhopper:
   # Local folder used by graphhopper to store its data
   graph.location: graph-cache
 
+  osm.import.accepted_highways: >
+    motorway,motorway_link,
+    trunk,trunk_link,
+    primary,primary_link,
+    secondary,secondary_link,
+    tertiary,tertiary_link,
+    unclassified,
+    residential,
+    living_street,
+    service,
+    pedestrian,
+    track,
+    bus_guideway,
+    escape,
+    raceway,
+    road,
+    busway,
+    footway,
+    bridleway,
+    steps,
+    corridor,
+    path,
+    cycleway,
+    proposed,
+    construction,
+    emergency_bay
 
   ##### Routing Profiles ####
 
--- a/config-example.yml
+++ b/config-example.yml
@@ -169,11 +169,13 @@ graphhopper:
 
   #### Storage ####
 
-  # Excludes certain types of highways during the OSM import to speed up the import and reduce the size of the graph.
-  # A typical case is excluding footway,cycleway,path and maybe pedestrian and track for motorized vehicles. This leads
-  # to a smaller graph, because there are fewer ways (obviously), but also because there are fewer junctions.
-  # Another example is excluding motorway, trunk and maybe primary highways for bicycle or pedestrian routing.
-  import.osm.ignored_highways: footway,cycleway,path,pedestrian,steps
+  # Excludes certain types of highways during the OSM import to speed up the process and reduce the size of the graph.
+  # A typical application is excluding 'footway','cycleway','path' and maybe 'pedestrian' and 'track' highways for
+  # motorized vehicles. This leads to a smaller and less dense graph, because there are fewer ways (obviously),
+  # but also because there are fewer crossings between highways (=junctions).
+  # Another typical example is excluding 'motorway', 'trunk' and maybe 'primary' highways for bicycle or pedestrian routing.
+  import.osm.ignored_highways: footway,cycleway,path,pedestrian,steps # typically useful for motorized-only routing
+  # import.osm.ignored_highways: motorway,trunk # typically useful for non-motorized routing
 
   # configure the memory access, use RAM_STORE for well equipped servers (default and recommended)
   graph.dataaccess.default_type: RAM_STORE
--- a/config-example.yml
+++ b/config-example.yml
@@ -32,8 +32,6 @@ graphhopper:
 
   profiles:
     - name: car
-      # note that as long as you only use motorized vehicles you can decrease the size of the graph
-      # using import.osm.ignored_highways (see below)
       vehicle: car
       weighting: custom
       custom_model:
@@ -42,6 +40,7 @@ graphhopper:
   #   u_turn_costs: 60
 
   # - name: bike
+      # to use the bike vehicle make sure to not ignore cycleways etc., see import.osm.ignored_highways below
   #   vehicle: bike
   #   weighting: custom
   #   # the custom model in bike2.json is defined to avoid hills
@@ -174,7 +173,7 @@ graphhopper:
   # A typical case is excluding footway,cycleway,path and maybe pedestrian and track for motorized vehicles. This leads
   # to a smaller graph, because there are less ways (obviously), but also because there are fewer junctions.
   # Another example is excluding motorway, trunk and maybe primary highways for bicycle or pedestrian routing.
-  # osm.import.ignored_highways: footway,cycleway,path
+  import.osm.ignored_highways: footway,cycleway,path,steps
 
   # configure the memory access, use RAM_STORE for well equipped servers (default and recommended)
   graph.dataaccess.default_type: RAM_STORE
--- a/config-example.yml
+++ b/config-example.yml
@@ -32,6 +32,8 @@ graphhopper:
 
   profiles:
     - name: car
+      # note that as long as you only use motorized vehicles you can decrease the size of the graph
+      # using import.osm.ignored_highways (see below)
       vehicle: car
       weighting: custom
       custom_model:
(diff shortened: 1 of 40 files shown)


config-example.yml
```diff
@@ -5,6 +5,32 @@ graphhopper:
...
+    busway,
+    footway,
+    bridleway,
+    steps,
+    corridor,
+    path,
+    cycleway,
+    proposed,
+    construction,
+    emergency_bay
```
easbar, line 33: I already put these values as defaults in the corresponding Java code. So this is mostly here for illustration purposes and we should probably remove this again before we merge the PR. Explicitly putting such a list into the configuration file is only needed if one wants to reduce the size of the graph for performance reasons and maybe because of #1858 (see my other comment). 

One thing I don't like is that we cannot easily comment out values here, because YAML does not allow commenting out something from the middle of a literal (a multi-line string here). To comment out something we need to move the commented out values to the end of the list. We could use a list instead of this string, but so far our configuration only uses simple string key-values pairs (with the exception of profiles and profiles_ch/lm).

core/src/main/java/com/graphhopper/routing/OSMReaderConfig.java
```diff
@@ -28,6 +32,54 @@ public class OSMReaderConfig {
...
+    /**
+     * Sets the values of the highway tag that shall be considered when we read the OSM file. OSM ways that do not include
+     * a highway tag with one of these values will be ignored entirely during import, but just because they are
+     * accepted does not necessarily mean they will be considered by the route calculations. This still depends on the
+     * weighting and/or the access flags for the corresponding edges. Not including some road types here is mostly a
+     * performance optimization. For example if one is only interested in routing for motorized vehicles the routing
+     * graph size can be reduced by not including tracks, footways and paths (20-50% depending on your area, ~25%
+     * for planet files). Another reason to exclude footways etc. for motorized vehicle routing could be preventing
+     * undesired u-turns (#1858). Similarly, one could exclude motorway, trunk or even primary for bicycle or pedestrian routing, but since there
+     * aren't many such roads the graph size reduction will be very small (<4%).
```
easbar, line 76: I tested this for Germany where excluding footways,tracks and paths reduced the number of edges by around 50%. I think this is simply because in Germany many such ways are actually mapped. Looking at taginfo the reduction should be more around 25% for the planet file, because in most countries not all of these smaller ways are mapped (and probably in many areas not so many such roads exist).

easbar:
I did not take a look at the failing tests yet, but will fix them of course once there is agreement for this PR @karussell

config-example.yml
easbar: I changed this and use an ignore list instead now.

easbar:
We also need to accept ways tagged as route=ferry or route=shuttle_train. There are also some tagged as route=hiking,mtb etc. but I think we can assume that these will be included because they carry a highway tag as well.

core/src/test/java/com/graphhopper/GraphHopperTest.java
```diff
@@ -98,12 +98,12 @@ public void setup() {
...
-            DIJKSTRA_BI + ",false,228",
-            ASTAR_BI + ",false,142",
-            ASTAR_BI + ",true,41",
-            DIJKSTRA_BI + ",true,48"
+            DIJKSTRA + ",false,708",
+            ASTAR + ",false,363",
+            DIJKSTRA_BI + ",false,346",
+            ASTAR_BI + ",false,192",
+            ASTAR_BI + ",true,42",
+            DIJKSTRA_BI + ",true,47"
```
easbar, line 106: I don't understand this yet. Why do we explore more nodes now that there are more (inaccessible) edges?

core/src/test/java/com/graphhopper/GraphHopperTest.java
```diff
@@ -2671,7 +2672,7 @@ void testLoadingWithAnotherSpeedFactorWorks() {
                     .setProfiles(new Profile("car").setVehicle("car").setWeighting("fastest"))
                     .setGraphHopperLocation(GH_LOCATION);
             hopper.load();
-            assertEquals(1942, hopper.getBaseGraph().getNodes());
+            assertEquals(2961, hopper.getBaseGraph().getNodes());
```
easbar, line 2675: The graph size increases, because of the additional ways (and junctions).

config-example.yml
```diff
@@ -168,6 +170,12 @@ graphhopper:
 
   #### Storage ####
 
+  # Excludes certain types of highways during the OSM import to speed up the import and reduce the size of the graph.
+  # A typical case is excluding footway,cycleway,path and maybe pedestrian and track for motorized vehicles. This leads
+  # to a smaller graph, because there are less ways (obviously), but also because there are fewer junctions.
+  # Another example is excluding motorway, trunk and maybe primary highways for bicycle or pedestrian routing.
+  # osm.import.ignored_highways: footway,cycleway,path
```
easbar, line 177: Or should we enable this by default?

example/src/main/java/com/graphhopper/example/RoutingExample.java
```diff
@@ -105,14 +105,14 @@ public static void headingAndAlternativeRoute(GraphHopper hopper) {
 
         // calculate alternative routes between two points (supported with and without CH)
         req = new GHRequest().setProfile("car").
-                addPoint(new GHPoint(42.502904, 1.514714)).addPoint(new GHPoint(42.511953, 1.535914)).
+                addPoint(new GHPoint(42.502904, 1.514714)).addPoint(new GHPoint(42.508774, 1.537094)).
```
easbar, line 108: I chose another route here, because for the one we used so far there were no alternatives anymore.

navigation/src/test/java/com/graphhopper/navigation/NavigateResponseConverterTest.java
```diff
@@ -273,7 +273,7 @@ public void intersectionTest() {
 
         step = steps.get(4);
         intersection = step.get("intersections").get(3);
-        assertEquals(1, intersection.get("in").asInt());
+        assertEquals(2, intersection.get("in").asInt());
         assertEquals(0, intersection.get("out").asInt());
```
easbar, line 277: This is the intersection:

image

Why did we expect zero outgoing roads so far? It looks like one can leave this intersection in Southern direction. The reason there used to be one incoming road and now there are two is that the road coming in from the West is tagged as `highway=pedestrian` and thus was excluded for the car profile so far. But since this is a oneway shouldn't this rather be an outgoing road?

core/src/test/java/com/graphhopper/GraphHopperTest.java
easbar: Yes, I did this check. When we add this line:

```java
 hopper.getReaderConfig().setIgnoredHighways(Arrays.asList("footway", "cycleway", "pedestrian", "path", "steps"));
```

to this test we get: 516,259,232,144,38,46 for the numbers of visited nodes. So there is still a minor difference, but the performance won't be impacted (but only if these highway types are excluded).

easbar: And not sure where the difference comes from. Is there anything else we could exclude to make it equal?

config-example.yml
easbar: I did... but still not entirely sure :)

navigation/src/test/java/com/graphhopper/navigation/NavigateResponseConverterTest.java
easbar: Ok so the indices simply changed because now the pedestrian route is accepted and nothing else needs to be done here? Basically this test only makes sure nothing changes, because we can't tell which indices are 'correct'?

easbar: Yes, probably.

core/src/main/java/com/graphhopper/routing/util/OSMParsers.java
```diff
@@ -27,28 +27,43 @@
...
+    private List<String> acceptedRoutes;
     private final List<TagParser> wayTagParsers;
     private final List<VehicleTagParser> vehicleTagParsers;
     private final List<RelationTagParser> relationTagParsers;
     private final List<RestrictionTagParser> restrictionTagParsers;
     private final EncodedValue.InitializerConfig relConfig = new EncodedValue.InitializerConfig();
 
     public OSMParsers() {
-        this(new ArrayList<>(), new ArrayList<>(), new ArrayList<>(), new ArrayList<>());
+        this(new ArrayList<>(), Arrays.asList("ferry", "shuttle_train"), new ArrayList<>(), new ArrayList<>(), new ArrayList<>(), new ArrayList<>());
```
easbar, line 44: I did another check (for Germany). I counted the number of ways that use a `route` tag, but do **not** use a `highway` tag:

```
ferry: 0.01%, 983
ski: 0.00%, 413
canoe: 0.00%, 241
piste: 0.00%, 193
boat: 0.00%, 37
historic: 0.00%, 26
shuttle_train: 0.00%, 20
bicycle: 0.00%, 9
mudflat_hiking: 0.00%, 8
fitness_trail: 0.00%, 8
hiking: 0.00%, 5
railway: 0.00%, 4
flat: 0.00%, 3
foot: 0.00%, 3
sk:nordic: 0.00%, 3
RB/RE Richtung Rosenheim: 0.00%, 3
nordic: 0.00%, 2
power: 0.00%, 2
pipeline: 0.00%, 2
bus: 0.00%, 1
Horumersiel: 0.00%, 1
water_skiing: 0.00%, 1
fitness-trail: 0.00%, 1
F1 F2 F3: 0.00%, 1
train: 0.00%, 1
19 Haigerloch - Heiligenzimmern; 7402 Sulz; 7402 Horb; 7416 Neckarhausen: 0.00%, 1
yes: 0.00%, 1
tracks: 0.00%, 1
Faulungen / Ort Struth-Steinbruch: 0.00%, 1
harbour_cruise: 0.00%, 1
cruise: 0.00%, 1
Ship: 0.00%, 1
```

So most of them are ferries and everything else occurs only less than 500 times (the total number of accepted ways was >13 million).

So I would say we do not need this list of 'acceptedRoutes' and can simply accept all ways with a 'route' tag. In most cases these will be inaccessible edges in the end, because the vehicle/access parsers do not set the access to true for unknown types of routes(?).

core/src/main/java/com/graphhopper/reader/osm/WaySegmentParser.java
```diff
@@ -137,9 +137,9 @@ public Date getTimeStamp() {
     private class Pass1Handler implements ReaderElementHandler {
         private boolean handledWays;
         private boolean handledRelations;
-        private long wayCounter = 1;
+        private long wayCounter = 0;
         private long acceptedWays = 0;
-        private long relationsCounter = -1;
+        private long relationsCounter = 0;
```
easbar, line 142: no idea why wayCounter was set to 1 and relationsCounter to -1 so far (?)

easbar:
@karussell I just realized that the RoadsTagParser did not use the getAccess method in handleWayTags. So prior to this change it did not accept ways tagged with `route` (like ferries) so they weren't accessible. Unless there was another vehicle that accepts ferries, of course.
So this PR changed the behavior in case we use only the roads vehicle: Before ferries were not used, but now they are.

PR #2706, opened 2022-12-02
add crossing EV

karussell:
Make it possible to avoid railroad and other crossings. (and use this in path details)

Requires: #2705

grafik


easbar:
thanks :)

PR #2707, opened 2022-12-05
Add footway encoded value

### LINKED ISSUE #2692: Add footway encoded value?
The key footway=* refines the tag highway=footway: https://wiki.openstreetmap.org/wiki/Key%3Afootway

This would be useful for pedestrian routing: https://discuss.graphhopper.com/t/adding-additional-footway-information-to-details-on-route-endpoint/7565

easbar:
Here I added a footway encoded value for the footway=sidewalk,crossing,... etc. tag: https://wiki.openstreetmap.org/wiki/Key%3Afootway, fixes #2692 

I briefly tried calculating some pedestrian routes that avoid crossings in Berlin. Note that almost none of these red-dashed separate footways were mapped in OSM at the beginning of this year! See also here: https://wiki.openstreetmap.org/wiki/Berlin/Verkehrswende. 

The new footway encoded value simply parses the footway=* tag without any further thoughts. Calculating good pedestrian routes is another story though. For example we can easily use a custom model to prevent footway=crossing (see screenshots), but care must be taken that as a result the route does simply use the normal roads (like tertiary) instead. Therefore I also lowered the priority for everything except footway and cycleway here, but whether this is a good idea depends on the area and mapping quality.

image

image

Then again, the data can also be used to simply report crossings, without changing the route as it was requested in the forum. 

To make the footway types visible in the elevation diagram for localhost:8989/maps/ we still need to add 'footway' to the list of details here: https://github.com/graphhopper/graphhopper/blob/4ecbd61eafaee01d1812ea5aa3ddf8e6e8d25c2d/web-bundle/src/main/resources/com/graphhopper/maps/config.js#L15-L20 
IMO it would be nice to have an option to show all the available path details here.

@natedx @MonfretAugustin can you check if this is useful for what you were trying to do? Any thoughts?

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ev/DefaultEncodedValueFactory.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/DefaultEncodedValueFactory.java
@@ -74,6 +74,8 @@ public class DefaultEncodedValueFactory implements EncodedValueFactory {
             enc = new EnumEncodedValue<>(HazmatWater.KEY, HazmatWater.class);
         } else if (Lanes.KEY.equals(name)) {
             enc = Lanes.create();
+        } else if (Footway.KEY.equals(name)) {
+            enc = new EnumEncodedValue<>(Footway.KEY, Footway.class);
         } else if (OSMWayID.KEY.equals(name)) {
             enc = OSMWayID.create();
         } else if (MtbRating.KEY.equals(name)) {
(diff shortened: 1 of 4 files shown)


easbar:
For the foot route calculations we could also consider reducing the speed or priority for crossings in FootTagParser. So far we do not use this tag at all it seems.

natedx:
Great ! This is going in the exact direction we need for routing. It's true that I was initially thinking only about reporting for in-app consumption, but the fact that we can use a custom model to limit the amount of crossings and hopefully make the routing better is a great feature. Down the line we could even imagine adding extra encoded values specific to certain handicaps : kerb (lowered or not) for wheelchairs, audio_signals at crossings (yes or no) for the visually impaired.

Issue #2659, opened 2022-09-20
Documentation for creating new routing profiles is outdated


DanielAMCON:
It seems like the documentation for creating new routing profiles is outdated (e.g. https://github.com/graphhopper/graphhopper/blob/6.x/docs/core/create-new-flagencoder.md and https://github.com/graphhopper/graphhopper/blob/6.x/docs/core/profiles.md) because the interface _FlagEncoder_ does not exist anymore.

easbar:
I just removed the mention of `FlagEncoder` from `profiles.md`: a3d6ca1c0ada941a86a0feace39f793a9120419c
Regarding `create-new-flagencoder.md`, I think we should just remove the entire file. The documentation needs a complete overhaul, especially everything that describes low level stuff. But this does not make much sense before the flag encoder refactoring is finished. For now the best available documentation is the code itself and the corresponding unit tests.

easbar:
If all you want to do is tweak some of the routing profiles, take a look at custom models. They should cover most use-cases. Otherwise you can think of `VehicleTagParser`s to be the new `FlagEncoder`s (but they will eventually be removed as well). Basically there are only encoded values and parsers that write to the encoded values. Basically the only thing that makes `VehicleTagParser`s specially is that they write speed and access values, not just one...

DanielAMCON:
Thanks for your reply. In my case, changing both speed and access values is quite important. So I guess I'll try to create a new implementation of `VehicleTagParser` until it's removed.

easbar:
Yes, either try to create a new implementation or modify one of the existing ones.

HarelM:
Hi, I'm trying to migrate to version 6 and am experiencing some problems.
I'm not sure if this is a documentation issue or a bug, so for now I'm writing here:
I'm trying to migrate the no longer exiting car4wd to use a custom model.
The following is my config.yml I use in order to create this profile:
```yml
profiles:
    - name: car4wd
      vehicle: roads
      weighting: custom
      custom_model:
        distance_influence: 1
        speed:
          - if: "track_type == GRADE4 || track_type == GRADE5"
            limit_to: 5
          - else: ""
            limit_to: "car_average_speed"
        priority:
          - if: "track_type != GRADE4 && track_type != GRADE5 && car_access == false"
            multiply_by: "0"
    - name: hike
      vehicle: hike
      weighting: short_fastest
    - name: mtb
      vehicle: mtb
      weighting: short_fastest
```

[...]

PR #2710, opened 2022-12-12
New profile for busses

DanielAMCON:
My company needed to be able to calculate routes specifically for buses. As discussed in #2659 , I created a new profile for this purpose. We would like to make this available for others.

As already explained in the issue, the VehicleTagParser will also be removed in the long run. Therefore, I can understand if the PR will not be merged. If the case I described can now be implemented by other means, I will of course not merge the PR either. I am pleased about your feedback.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -18,6 +18,7 @@ Here is an overview:
  * ChristophKaser, agrees to the project's CLA, improved Android compatibility #1207
  * chucre, add special JSON output format, see #41
  * daisy1754, fixed usage of graphhopper.sh script
+ * DanielAMCON, add bus profile
  * dardin88, instructions improved
  * dewos, web API bug fixes
  * devemux86, improvements regarding Android, GPX and more
--- a/config-example.yml
+++ b/config-example.yml
@@ -1,196 +1,20 @@
 graphhopper:
-
-  # OpenStreetMap input file PBF or XML, can be changed via command line -Ddw.graphhopper.datareader.file=some.pbf
   datareader.file: ""
-  # Local folder used by graphhopper to store its data
   graph.location: graph-cache
-
-
-  ##### Vehicles #####
-
-  # More options: foot,hike,bike,bike2,mtb,racingbike,motorcycle,wheelchair (comma separated)
-  # bike2 takes elevation data into account (like up-hill is slower than down-hill) and requires enabling graph.elevation.provider below.
-  # graph.flag_encoders: car
-
-  # Enable turn restrictions for car or motorcycle.
-  # graph.flag_encoders: car|turn_costs=true
-
-  # Add additional information to every edge. Used for path details (#1548), better instructions (#1844) and tunnel/bridge interpolation (#798).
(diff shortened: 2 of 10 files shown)


DanielAMCON:
Oh, that would be great. Thank you in advance!

DanielAMCON:
Thanks for commiting the changes. I changed the config file as you described and the bus profile was created successfully.

Unfortunately, the routing behaviour is not working as before. Here are some examples:

1. The bus should be able to take the road that is marked in red (https://www.openstreetmap.org/way/81032078) because there is a bus stop. Even if I set a waypoint there, it takes the long way.
BusError1
2. Here, the bus also cannot reach the bus stop. The road is https://www.openstreetmap.org/way/160155244.
BusError2
3. This bus stop cannot be reached either. The road is https://www.openstreetmap.org/way/30016603.
BusError3

There are no barriers in all of that cases. I assume that this is related to the _motorvehcile_, _motorcar_, and _vehicle_ tags, as they are each set to _no_ for the roads mentioned. Could you please take a closer look at this?

DanielAMCON:
Yes, I've got time for this. I'll give it a try!

easbar:
But doesn't `psv=yes` rather belong to a to-be-created transportation mode dependent access value? We aren't going to create separate encoded values for `hgv=yes`, `motorcycle=yes`, `foot=yes` etc., are we? Also adding a psv encoded value won't change turn restrictions that can also be transportation mode specific. So yes, adding a psv encoded value might improve the situation for this specific bus profile, but in the longer run we need a proper replacement (and strategy) for what is currently still done in the vehicle tag parsers.

easbar:
> Hm, you mean we rather need a boolean encoded value per transportation mode?

Yes, so far I thought we were aiming for a boolean value to represent access per transportation mode. This would be a compound of multiple different OSM tags (for example oneway, different barriers etc.) including the access tags. But of course there are many options this can be dealt with. For example would you like a separate oneway encoded value instead of putting it into the access encoded value? And if yes, will there even be an access encoded value?

> This should be handled from the roads vehicle via graph.vehicles: roads|transportation_mode=BUS,car

Yes, currently `graph.vehicles` is the place where this can be configured. The only problem here is that so far we do not support `transportation_mode=bus`, but only `transportation_mode=psv`. But even when we use `transportation_mode=psv` we will currently miss those restrictions that are tagged as `bus=yes`.

DanielAMCON:
So, I now added psv as a new EncodedValue. At first, it didn't work. After analyzing this a bit further, I realized that it works when I delete this part of the bus.json:
```
{
  if": "true",
  "limit_to": "car_average_speed * 0.9"
}
```
So it now looks like this:
```
{
  "distance_influence": 90,
  "speed": [
    {
      "if": "true",
      "limit_to": "120"
    }
  ],
  "priority": [
    {
      "if": "car_access == false && psv != YES || max_width < 3 || max_height < 4",
      "multiply_by": "0"
    }
  ]
}
```
I am very confused now. Is `car_average_speed` not working here? And if yes, why exactly doesn't it work? Do you guys have an idea?

Regarding your discussion, I implemented this using an EncodedValue. If you give me further information, I could also implement this using a boolean. But based on your dicussion I actually don't quite know how to do this.

PR #2714, opened 2023-01-03
Extend Heading Documentation

lukasalexanderweber:
I had the task to work myself into the heading functionality and I thought it might be worth sharing the documented result.

As always thanks for maintaining this great library, everything worked as expected with the detours resulting with different headings.


lukasalexanderweber:
@easbar thanks for your response in the forum. I think this should now be ready to merge (if you are interested)

lukasalexanderweber:
you just have to open [heading.md](https://github.com/lukasalexanderweber/graphhopper/blob/heading/docs/core/heading.md) to see what's new

client-hc/src/test/java/com/graphhopper/api/GraphHopperWebTest.java
```diff
@@ -57,10 +57,10 @@ public void profileIncludedAsGiven() {
     public void headings() {
         GraphHopperWeb hopper = new GraphHopperWeb("http://localhost:8080/route");
         GHRequest req = new GHRequest(new GHPoint(42.509225, 1.534728), new GHPoint(42.512602, 1.551558)).
-                setHeadings(Arrays.asList(10.0, -90.0)).
+                setHeadings(Arrays.asList(10.0, 90.0)).
```
easbar, line 60: Ah, yes, negative headings aren't even allowed.

client-hc/src/test/java/com/graphhopper/api/HeadingExamples.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.runner.Request;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Arrays;
+
+@Disabled
+public class HeadingExamples {
+    static GraphHopperWeb hopper;
```
easbar, line 19: I'd rather not add disabled tests, also because the asserted distances will depend on the exact map we are using. We could move these tests into `GraphHopperTest`. Maybe heading is already covered sufficiently there.

docs/core/heading.md
```diff
@@ -0,0 +1,76 @@
+# Heading
+
+The flexible and hybrid modes allow adding a desired heading (north based azimuth between 0 and 360 degree) to any point. Adding a heading makes it more likely that a route starts towards the provided direction, because roads going into other directions are penalized (see the [heading_penalty](#heading-penalty) parameter).
+
+A heading with the value 'NaN' won't be enforced and a heading not within [0, 360] will trigger an IllegalStateException. It is important to note that if you force the heading at via or end points the outgoing heading needs to be specified. I.e. if you want to force "coming from south" to a destination you need to specify the resulting "heading towards north" instead, which is 0.
```
easbar, line 5: It's an IllegalArgumentException in `Router#checkHeadings`, isn't it?

docs/core/heading.md
```diff
@@ -0,0 +1,76 @@
...
+3.	With common get request to server (the urls used in the following)
+
+For 2 and 3 you can self-host a graphhopper web server as described [here](https://github.com/graphhopper/graphhopper#installation), but in [config-example.yml](../../config-example.yml) we need to outcomment
+
+```
+  profiles_ch:
+    - profile: car
+```
+
+in line 56-57 since `heading` is not available in speed mode.
```
easbar, line 24: Heading still works when there are CH profiles, we just need to disable it at query time. So there is no need to comment out this section from the config, but we need to `ch.disable` to the routing request. This way one can use CH queries and flexible queries with heading on the same server.

docs/core/heading.md
```diff
@@ -0,0 +1,76 @@
...
+to point
+`42.567396, 1.597807`
+
+#### Without Heading
+
+`localhost:8989/route?profile=car&point=42.566757,1.597751&point=42.567396,1.597807&type=json&instructions=false&points_encoded=false`
+
+![without_heading](./images/without_heading.PNG)
+
+(the images are created by putting the resulting coordinates LineString in a GeoJSON and plotting them)
```
easbar, line 41: Unfortunately we cannot set the heading in GraphHopper Maps yet: https://github.com/graphhopper/graphhopper-maps/issues/297 Then we could simply create screenshots with it :)

example/src/test/java/com/graphhopper/example/HeadingExample.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.io.File;
+import java.util.Arrays;
+
+public class HeadingExample {
```
easbar, line 20: For the other examples (RoutingExample.java etc.) we took a slightly different approach where the examples are in `src/main/java`, and they are called from a separate test class in `src/test/java`. Actually, I am not sure why (@karussell?), but we should probably stick to one approach. Btw `RoutingExample.java` also has an example for heading, so not sure if we need more/both?

easbar: Ok, from src/main they can copy the entire file. But from src/test it would also be possible to copy the main method. But no real need to change this, so for this PR better move HeadingExample to src/main like the others, if we keep it at all.

easbar: Using src/main is also nice, because this way the examples are contained in the jar.

docs/core/heading.md
lukasalexanderweber: ah nice I didn't know that

docs/core/heading.md
lukasalexanderweber: at the moment we are also thinking about a good way to integrate heading in our UI (optional parameter for each point) :)

client-hc/src/test/java/com/graphhopper/api/HeadingExamples.java
lukasalexanderweber: I deleted the test in client-hc, it was very redundant. In retrospect, it was more for me to understand in which ways Graphhopper can be called.

example/src/test/java/com/graphhopper/example/HeadingExample.java
lukasalexanderweber: > Btw `RoutingExample.java` also has an example for heading, so not sure if we need more/both?

In [a18c45f](https://github.com/graphhopper/graphhopper/pull/2714/commits/a18c45fb9181401947a419524194e30def301b50) I allowed myself to delete the heading part in RoutingExample.java.

I updated HeadingExample.java to be in src/main and an HeadingExampleTest.java to ensure that it compiles and assures correctly.

lukasalexanderweber:
@easbar thanks for your review! Somehow I don't get Github's re-request button so I'll just write a comment that I incorporated your remarks

docs/core/heading.md
easbar: In the previous version of GraphHopper Maps things like this were possible by manually modifying the url, but that isn't ideal either.

easbar:
Thanks!

PR #2716, opened 2023-01-06
distance_influence should not get a default value if no value is specified

karussell:
And the default value of 70 makes this impossible, so remove this.

Users of client-hc now have to set the distanceInfluence to 70 if a lower value is defined on the server-side and can remove it if they want to match the server-side value which is highly recommended for performance reasons.

### DIFF
--- a/web-api/src/main/java/com/graphhopper/util/CustomModel.java
+++ b/web-api/src/main/java/com/graphhopper/util/CustomModel.java
@@ -120,9 +120,6 @@ public class CustomModel {
         return this;
     }
 
-    /**
-     * Check with #hasDistanceInfluence() to avoid null values
-     */
     public Double getDistanceInfluence() {
         return distanceInfluence;
     }


web-api/src/main/java/com/graphhopper/util/CustomModel.java
```diff
@@ -120,6 +120,10 @@ public CustomModel setDistanceInfluence(Double distanceFactor) {
         return this;
     }
 
+    /**
+     * @return the distance influence of this CustomModel. 0 is returned if it wasn't set (i.e. it was null)
+     * @see #hasDistanceInfluence()
+     */
```
easbar, line 126: Can we not just return `null`? `null` means it's not there, so exactly what we want?

easbar:
Is/was this a client-hc-only problem or was distance_influence set to 70 also on the server side when we deserialized, e.g. an empty model?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ### 7.0 [not yet released]
 
+- users of client-hc now have to set the distanceInfluence to 70 if a lower value is defined on the server-side and can remove it if they want to match the server-side value which is highly recommended for performance reasons, see #2716
 - there is a new, required 'import.osm.ignored_highways' configuration option that must be used to not increase the graph size and decrease performance for motorized-only routing compared to previous versions, #2702
```
easbar, line 4: Can we not put it the other way around: Normally it should be null (so the server-side value will be taken) and users don't need to do anything. And if they set it to some value, well then this value will be used.

web-api/src/main/java/com/graphhopper/util/CustomModel.java
easbar: Did you commit this somewhere? Where was there a problem? I would prefer to keep it as `null` in case it is missing. The special value `0` might even be needed for some cases.

CHANGELOG.md
easbar: Ah I see :) Maybe this:

There is no longer a default value for the distanceInfluence Parameter in custom models sent via client-hc. Previously it was 70. Not setting it explicitly now means the server-side value will be used.

core/src/test/java/com/graphhopper/routing/weighting/custom/CustomWeightingTest.java
```diff
@@ -87,26 +87,26 @@ public void withPriority() {
     @Test
     public void withDistanceInfluence() {
         EdgeIteratorState edge = graph.edge(0, 1).setDistance(10_000).set(avSpeedEnc, 50).set(accessEnc, true, true);
-        assertEquals(720, createWeighting(new CustomModel().setDistanceInfluence(0)).calcEdgeWeight(edge, false), .1);
-        assertEquals(720_000, createWeighting(new CustomModel().setDistanceInfluence(0)).calcEdgeMillis(edge, false), .1);
+        assertEquals(720, createWeighting(new CustomModel().setDistanceInfluence(0d)).calcEdgeWeight(edge, false), .1);
+        assertEquals(720_000, createWeighting(new CustomModel().setDistanceInfluence(0d)).calcEdgeMillis(edge, false), .1);
```
easbar, line 91: Why do we need to use `0d` here?

web-api/src/main/java/com/graphhopper/util/CustomModel.java
```diff
@@ -112,17 +111,20 @@ public Map<String, JsonFeature> getAreas() {
...
-        return hasDistanceInfluence() ? distanceInfluence : DEFAULT_DISTANCE_INFLUENCE;
+    public CustomModel setDistanceInfluence(Double distanceFactor) {
+        this.distanceInfluence = distanceFactor;
+        return this;
+    }
+
+    /**
+     * Check with #hasDistanceInfluence() to avoid null values
+     */
+    public Double getDistanceInfluence() {
```
easbar, line 126: ...or simply check for `getDistanceInfluence() == null`. Why shall we recommend to use `hasDistanceInfluence`?

web-api/src/main/java/com/graphhopper/util/CustomModel.java
```diff
@@ -28,8 +28,7 @@ public class CustomModel {
 
     public static final String KEY = "custom_model";
 
-    // e.g. 70 means that the time costs are 25€/hour and for the distance 0.5€/km (for trucks this is usually larger)
-    static double DEFAULT_DISTANCE_INFLUENCE = 70;
```
easbar, line 32: Where is the server-side default now? Is it still 70?

otbutz: It has been bugging me for a while now that our "fastest" routing doesn't really pick the fastest route by default.

easbar: Maybe mention this in the changelog as well: Server-side profiles with custom weighting now use distance_influence: 0 by default (previously it was 70).

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomModelParser.java
```diff
@@ -115,7 +115,7 @@ static CustomWeighting.Parameters createWeightingParameters(CustomModel customMo
             CustomWeightingHelper prio = (CustomWeightingHelper) clazz.getDeclaredConstructor().newInstance();
             prio.init(lookup, avgSpeedEnc, priorityEnc, customModel.getAreas());
             return new CustomWeighting.Parameters(prio::getSpeed, prio::getPriority, prio.getMaxSpeed(), prio.getMaxPriority(),
-                    customModel.getDistanceInfluence(), customModel.getHeadingPenalty());
+                    customModel.hasDistanceInfluence() ? customModel.getDistanceInfluence() : 0, customModel.getHeadingPenalty());
```
easbar, line 118: Maybe extract a constant as this `0` is the server-side default value and it appears also in `FindMinMax`.

core/src/main/java/com/graphhopper/routing/weighting/custom/FindMinMax.java
```diff
@@ -23,10 +23,11 @@ public class FindMinMax {
     public static void checkLMConstraints(CustomModel baseModel, CustomModel queryModel, EncodedValueLookup lookup) {
         if (queryModel.isInternal())
             throw new IllegalArgumentException("CustomModel of query cannot be internal");
-        if (queryModel.hasDistanceInfluence() && queryModel.getDistanceInfluence() < baseModel.getDistanceInfluence())
+        double qmDI = queryModel.hasDistanceInfluence() ? queryModel.getDistanceInfluence() : 0;
+        double bmDI = baseModel.hasDistanceInfluence() ? baseModel.getDistanceInfluence() : 0;
```
easbar, line 27: Here the `0` is also the server-side default? Shouldn't there really only be one place where it appears: The code where the yaml definition of the model is translated to a Java object?

web-api/src/main/java/com/graphhopper/util/CustomModel.java
otbutz: As someone who expected the fastest route when using a CustomModel, the default value for "distance_influence" caught me a bit off guard. It is not obvious at the first moment that here the length of the route is included in the weighting by default.

I think that this is not clear to most who use this feature.

easbar: Ok, so far I wasn't aware that the expectation of an empty CustomModel is that it yields the fastest route. But anyway, I agree that 0 is probably the better default since 70 is quite arbitrary.

otbutz:
If the default value of the server is 0, should the restriction on the client side also be removed?

grafik

easbar:
Why would a warning in an (otherwise empty?) response be better than an error code? Increasing the distance influence on the server-side will make requests with distance influence values higher than the old but lower than the new value impossible, yes. But we could also fall back to A* in this case.

web-api/src/main/java/com/graphhopper/util/CustomModel.java
easbar: Then we can also just remove `hasDistanceInfluence`?

core/src/main/java/com/graphhopper/routing/weighting/custom/FindMinMax.java
easbar: Yes, I thought after reading the config file the distance influence field of the server-side model should either be 0 (if nothing was specified) or whatever was specified in the server config, but never `null`. And when we merge a request custom model with the one on the server-side we either use the one from the server (in case the requesting distance influence is `null` ) or use the value from the request (in case it was not `null`). 

So basically I think that `baseModel.getDistanceInfluence() == null` would be a programming error and can be treated as such just let the NullPointerException be thrown.

And for `qmDI` we don't need a default value, because when nothing is given we will use the server-side value. Written as code:

```java
        if (queryModel.getDistanceInfluence() != null && queryModel.getDistanceInfluence() < baseModel.getDistanceInfluence())
            throw new IllegalArgumentException("CustomModel in query can only use " +
                    "distance_influence bigger or equal to " + baseModel.getDistanceInfluence() + ", but was: " + queryModel.getDistanceInfluence());
```

easbar: Ok, I see this is not the way we do it currently. We do not apply the default server-side value during the deserialization of `config.yml`, but only much later in `createWeightingParameters`. This way it is also applied when we use the Java API. And because we call FindMinMax even before that we need to pay attention that there is no  value yet :( Maybe it would be cleaner to use two separate classes for the server-side/profile CustomModel and the one used for the request.

web-api/src/main/java/com/graphhopper/util/CustomModel.java
easbar: You mean the pure existence of a `hasXYZ` method (assuming the programmer sees it) serves as a hint that the value could be null? Probably a matter of preference. For me this is just a distraction. When I see something like `hasDistanceInfluence` I have no idea what it is and need to open the method to see what it does. The most standard way to make it obvious something might be missing is probably `Optional<Double>`.

otbutz: We should either use `Optional` or `@Nullable` if we want to improve developer experience. I agree with @easbar that the method only adds noise.

PR #2725, opened 2023-01-16
make custom areas available in custom models

karussell:
To avoid that areas have to be redefined in different custom models I initially introduced a special "custom-model-areas.geojson" file but then I noticed that we could just make our existing custom areas available the custom models too.

Technically we now load the custom areas twice and I think this is fine as it is rather decoupled at the moment from custom models (and I'm unsure how we could benefit from e.g. AreaIndex).

The CustomModel also only retrieves the areas that it needs and does not e.g. create a PreparedPolygon for all areas.

### DIFF
--- a/config-example.yml
+++ b/config-example.yml
@@ -192,8 +192,8 @@ graphhopper:
   #### Custom Areas ####
 
   # GraphHopper reads GeoJSON polygon files including their properties from this directory and makes them available
-  # to all tag parsers and vehicles and custom models.
-  # Country borders (see countries.geojson) are always included automatically.
+  # to all tag parsers, vehicles and custom models. All GeoJSON Features require to have the "id" property.
+  # Country borders are included automatically (see countries.geojson).
   # custom_areas.directory: path/to/custom_areas
 
 


config-example.yml
```diff
@@ -190,7 +192,8 @@ graphhopper:
   #### Custom Areas ####
 
   # GraphHopper reads GeoJSON polygon files including their properties from this directory and makes them available
-  # to all tag parsers and vehicles. Country borders (see countries.geojson) are always included automatically.
+  # to all tag parsers and vehicles and custom models.
+  # Country borders (see countries.geojson) are always included automatically.
```
easbar, line 196: Maybe add: "Areas without a `properties` field will be ignored"? Or shall we make the properties field required, so we do not need to check for null? I looked at the GeoJSON specification and apparently it is mandatory.

easbar:
Very nice :+1:

config-example.yml
easbar: > "A Feature object has a member with the name "properties". The value of the properties member is an object (any JSON object or a JSON null value)"

Yes, that means it has to have such a member, doesn't it?

> Countries without an "ISO3166-1:alpha3" property will be ignored, yes. I will make this clear.

But also when properties is `null`, see the file diff where there is this continue statement.

easbar: Sorry, I misread the code. The code that skips areas without properties is only in the parsing. So no need to change anything IMO.

> Countries without an "ISO3166-1:alpha3" property will be ignored, yes. I will make this clear.

Since the countries are shipped with the code I don't think this is really needed.

easbar: I think the general mechanism here is that for custom areas we simply make available all the properties to the tag parsers. And the country tag parser is looking at the `properties.id` specifically. But we could also feed the id field to the tag parsers as well and use this for the country code.

Issue #1234, opened 2017-11-28
Speed is 0 for low speed edges using motorcycle


boldtrn:
If you use the motorcycle profile and the curvature weighting and read an OSM way with a low maxspeed, for example maxspeed=2, the speed of the edge will be 0. An example way can be found [here](http://www.openstreetmap.org/way/367264051).

```
java.lang.IllegalStateException: Calculating time should not require to read speed from edge in wrong direction. Reverse:false, fwd:false, bwd:false
	at com.graphhopper.routing.weighting.AbstractWeighting.calcMillis(AbstractWeighting.java:44)
	at com.graphhopper.routing.weighting.FastestWeighting.calcMillis(FastestWeighting.java:83)
	at com.graphhopper.routing.Path.processEdge(Path.java:242)

```

karussell:
When I think about this a bit more: this is kind of expected as we set access to false for speed==0 and so if speed is rounded downwards the result can be access==false. The only other solution would be a different rounding strategy if close to 0. But having speed==0 and access==true does not make sense.

karussell:
But the question is why there is a path going through if access==false for both directions.

boldtrn:
> But the question is why there is a path going through if access==false for both directions.

I think I found the issue here. The CurvatureWeighting has no clause `if(speed==0) return Double.POSITIVE_INFINITIVE`.

karussell:
This is a more general problem that we always had problems with. 

For motorcycle we could solve it consistent: we could make the access-check dependent on speed==0 and do not use separate access bits as we have one speed value for every direction. 

But for other vehicles like car or foot we only want to use `speedBits+2` instead `speedBits*2` and so we need two separate access bits.

In the latest branch #1447 we could try to solve this via creating a special "combined" speed+access EncodedValue that handles this on the lowest level and switches to `speedBits*2` without any access bits if the speed value should be stored for both directions.

PR #2726, opened 2023-01-17
improve handling of low speed

### LINKED ISSUE #2662: maxspeed:forward and maxspeed:backward evaluates to the slower speed of both
If different speeds apply for each direction on a way, OSM suggests using
`maxspeed:forward` for the speed in geometry direction and `maxspeed:backward` for the opposite direction.

When setting both forward, and backward speed graphhopper appears to be using always the lower speed for both directions. 
E.g. maxspeed:forward is 30 and maxspeed:backward is 5 -> resulting speed for both directions is 5. 

From my understanding detailed information of speeds per direction are lost that way. Would be glad if someone could validate if this is the intended behaviour or I'm misunderstanding something. :)

https://github.com/graphhopper/graphhopper/blob/7682380c64abf763163a4b9f1c57833bd6babeed/core/src/main/java/com/graphhopper/routing/util/VehicleTagParser.java#L182-L193

karussell:
This change was originally only for #1234, see [VehicleTagParser.setSpeed](https://github.com/graphhopper/graphhopper/pull/2726/files#diff-e97fad3b84a2b3bfe14f9025b391ff2bbca64b440b26792568a2d37867147fc4R192).

Will try to move the other changes into another PR... the other changes are a bit too involved and now some further problems occurred ... 

~Now it is a broader fix that fixes a second problem that was mentioned in testSetSpeed0_issue367: in case when the speed only has one value for both directions but the access values are directed which leads to inconsistent behaviour. So we now force two directions for the speed (for car) which further lead me to the changes regarding the max speed and fixes #2662 too (no tests added, existing tests were only adapted).~

~Furthermore now the better value "maximum max_speed" is picked for bike parser decisions instead of the "minimum max_speed".~

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/VehicleTagParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/VehicleTagParser.java
@@ -195,6 +195,9 @@ public abstract class VehicleTagParser implements TagParser {
     }
 
     protected void setSpeed(boolean reverse, IntsRef edgeFlags, double speed) {
+        // special case when speed is non-zero but would be "rounded down" to 0 due to the low precision of the EncodedValue
+        if (speed > 0.1 && speed < avgSpeedEnc.getSmallestNonZeroValue())
+            speed = avgSpeedEnc.getSmallestNonZeroValue();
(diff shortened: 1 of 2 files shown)


easbar:
Nice, thanks!

PR #2727, opened 2023-01-18
Maxspeed fix

### LINKED ISSUE #2662: maxspeed:forward and maxspeed:backward evaluates to the slower speed of both

karussell:
I split #2726 into two pieces.

This PR fixes the problem that was mentioned in testSetSpeed0_issue367: in case when the speed only has one value for both directions but the access values are directed it no longer leads to inconsistent behaviour. We now force two directions for the car speed, so I was also able to fix #2662 (no tests added, existing tests were only adapted).

Furthermore now the better value "maximum max_speed" is picked for bike parser decisions instead of the "minimum max_speed" which pointed me to another bug regarding picking the previous/next PriorityCode (which is now fixed).

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/AStar.java
+++ b/core/src/main/java/com/graphhopper/routing/AStar.java
@@ -89,6 +89,8 @@ public class AStar extends AbstractRoutingAlgorithm implements EdgeToEdgeRouting
         double weightToGoal = weightApprox.approximate(from);
         AStarEntry startEntry = new AStarEntry(EdgeIterator.NO_EDGE, from, 0 + weightToGoal, 0);
         fromHeap.add(startEntry);
+        if (fromOutEdge == NO_EDGE || toInEdge == NO_EDGE)
+            return extractPath();
         if (!traversalMode.isEdgeBased())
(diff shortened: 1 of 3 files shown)


otbutz:
Very nice! The last piece for the correct handling of the maximum speed in GH would then be #2266. But we need to handle upper/lower bounds for untagged roads to fix this as discussed in https://github.com/graphhopper/graphhopper/pull/2268#issuecomment-1141349955

PR #2728, opened 2023-01-20
Test access and speed instead of just flags.isEmpty() for foot and bike parsers

easbar:


core/src/test/java/com/graphhopper/routing/util/AbstractBikeTagParserTester.java
```diff
@@ -276,25 +278,29 @@ public void testTramStations() {
...
         way = new ReaderWay(1);
         way.setTag("highway", "track");
         way.setTag("railway", "platform");
-        flags = encodingManager.createEdgeFlags();
-        parser.handleWayTags(flags, way);
-        assertNotEquals(true, flags.isEmpty());
+        edgeFlags = encodingManager.createEdgeFlags();
+        parser.handleWayTags(edgeFlags, way);
+        // todonow: why is this 18 for MTB, 2 for racingbike and 12 for normal bike?
+        assertEquals(12.0, avgSpeedEnc.getDecimal(false, edgeFlags));
```
easbar, line 292: Is this a bug?

easbar: Ok, so shall I just comment out this check and we fix this later?

core/src/test/java/com/graphhopper/routing/util/TagParsingTest.java
```diff
@@ -176,9 +176,9 @@ public void testSharedEncodedValues() {
         way.setTag("junction", "circular");
         tagParsers.forEach(p -> p.handleWayTags(edgeFlags2, way, relFlags));
 
-        assertTrue(roundaboutEnc.getBool(false, edgeFlags));
+        assertTrue(roundaboutEnc.getBool(false, edgeFlags2));
         for (BooleanEncodedValue accessEnc : accessEncs)
-            assertTrue(accessEnc.getBool(false, edgeFlags));
+            assertTrue(accessEnc.getBool(false, edgeFlags2));
```
easbar, line 181: I think this was a bug in the testing code.

core/src/test/java/com/graphhopper/routing/util/AbstractBikeTagParserTester.java
ratrun: I checked this particular test. The test covers the following tag combination:
```
       way.setTag("railway", "platform");
```
This combination exists in a bounding box around Germany in only two cases, I just cheched this with Overpass. Most `railway=platform` occur in combination with `highway=footway` or `highway=platform`.

We could force the speed to PUSHING_SECTION_SPEED for the `railway=platform` , but I would not do this.

Or is the above discussion a general one about the low speed of 2 km/h for a standalone `highway=track` for the racingbike? I agree that PUSHING_SECTION_SPEED would eventually fit better.

easbar: I was just surprised that in this case mtb is supposed to be x9 faster than bike and that there are different bike speeds at all. Using pushing_section_speed for `railway=platform` would make sense, yes.

PR #2730, opened 2023-01-24
contributing no_NO and updating pl_PL while at it

OlafFlebbeBosch:
Contributing norwegian bokmål message catalog


OlafFlebbeBosch:
Added my company nickname as well .

Issue #2127, opened 2020-09-10
PrepareRoutingSubnetworks sometimes uses more memory


easbar:
When I ran the import with different Xmx limits I experienced that sometimes the limit was exceeded during subnetwork removal and sometimes it was not. Not sure why, but apparently there can be bad cases in which more/too much memory is used and maybe we can control this somehow?

PR #2733, opened 2023-01-27
concurrent subnetworks

karussell:
Proposal to speed up the subnetwork creation especially when using many profiles.

I hope this does not interfere with #2127.


core/src/main/java/com/graphhopper/storage/BaseGraph.java
```diff
@@ -493,6 +493,13 @@ public int getSegmentSize() {
         return segmentSize;
     }
 
+    /**
+     * Currently MMapDataAccess cannot be used when potentially writing to the same DataAccess from multiple threads.
+     */
+    public boolean areEdgesConcurrentWritable() {
+        return store.areEdgesConcurrentWritable();
+    }
```
easbar, line 501: Setting encoded values like `xyz_subnetwork` isn't thread-safe for RAM DataAccess, or am I missing something? Basically we would potentially be modifying the same int from different threads in different ways?

PR #2738, opened 2023-01-30
Split vehicle tag parsers into access, speed and priority

easbar:
This uses #2729 as foundation, but only moves access, speed and priority into different files, without some of the other changes.

Some noteworthy changes:

* roads vehicle turn costs are now disabled by default (just like the other vehicles)
* speeds for edges where access flags are false are no longer set to zero


core/src/test/java/com/graphhopper/routing/util/parsers/AbstractBikeTagParserTester.java
```diff
@@ -260,37 +267,36 @@ public void testTramStations() {
...
         way.setTag("highway", "track");
         way.setTag("railway", "platform");
-        edgeFlags = encodingManager.createEdgeFlags();
-        parser.handleWayTags(edgeFlags, way);
-        // we use speed=2 for racingbike, 12 for normal bike and 18 for mtb, which does not really make sense
-        // todo: fix this, see #2728
-//        assertEquals(12.0, avgSpeedEnc.getDecimal(false, edgeFlags));
+        accessParser.handleWayTags(encodingManager.createEdgeFlags(), way, null);
+        // todonow: why is it 4 now?, it used to be either 2, 12 or 18, see #2728
+        assertEquals(4, avgSpeedEnc.getDecimal(false, edgeFlags));
```
easbar, line 299: @karussell did you figure this out?

PR #2740, opened 2023-02-02
New signature for encoded values and tag parsers

easbar:
So far we use encoded values like this: (using IntEncodedValue as an example, but it is the same for all EVs)

```java
intEnc.setInt(reverse, edgeFlags, value);
```

This writes the int to the given edgeFlags (IntsRef) which serve as output type so to say.

Here, I changed this to:

```java
intEnc.setInt(reverse, edgeId, intAccess, int value);
```

where `intAccess` is a new interface that can simply set or get an int.

This way we can use the same code as before, but using the `intAccess` interface instead of the concrete `IntsRef` we can basically 'redirect' the calls to any kind of (currently int-based) storage. The `edgeId` parameter allows us to identify the edge.

The most common use-case for this type of code is in `BaseGraph` where we get/set different encoded values for an edge. So far we would always copy the corresponding bits from (or to) an `IntsRef` to read/write data from/to the graph. To improve performance we cached these 'edgeFlags' between different reading calls. With the new `IntAccess` interface we can directly read from/write to the underlying storage (currently the `edges` DataAccess of the `BaseGraph`). According to my measurements this 'edgeFlag'-caching did not yield a speed-up compared to directly reading/writing from the DataAccess. In situations where there are many ints per edge but we only need a few of them the edge flag caching even yields a slow-down, because more ints are read than needed.

When we now call `edge.set(enc)` the `BaseGraph` passes an `IntAccess` implementation that directly reads from the `edges` DataAccess to the encoded value. However, we can also use encoded value implementations that ignore the given `intAccess` and store the data another way. In other words: We can now add encoded values to the encoding manager that store data somewhere outside the graph (the main motivation for this PR).

I also changed the `TagParser` interface accordingly.


easbar:
I renamed IntAccess to EdgeIntAccess. To distinguish this from DataAccess I would probably rename DataAccess to DataStorage, but that's for another issue I would say.

Issue #2625, opened 2022-07-11
Store edge BBox in the graph


otbutz:
We currently compute the BBox on-the-fly for custom area checks and only based on the tower nodes of the edge:

https://github.com/graphhopper/graphhopper/blob/af660cdf2fc1a365b279eb6047fc5956e585cfb9/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java#L78-L85

This has a few drawbacks. On the one hand, we lose precision, since the full point list could in fact have a larger BBox than that of only the tower nodes:

![image](https://user-images.githubusercontent.com/22315436/178253713-af1f9466-ec63-4ad5-b425-4d38fe4d11c4.png)

And it adds computational overhead as discussed previously:

https://github.com/graphhopper/graphhopper/pull/2338#issuecomment-937549263
https://discuss.graphhopper.com/t/why-using-areas-is-low-efficient/7374/6

easbar:
> We currently compute the BBox on-the-fly for custom area checks and only based on the tower nodes of the edge:

Ok, but the code where we use only the tower nodes to compute the bounding box is this method:

https://github.com/graphhopper/graphhopper/blob/06b261a3650c9f9ece9b394c65bc7dfe71a3ae04/core/src/main/java/com/graphhopper/util/GHUtility.java#L939-L944

We currently use it for the block_area feature (GraphEdgeIdFinder) and, like you say, for the custom weighting areas. We use the tower node bounding box in `if (!p.getBounds().intersects(bbox)) return false;`, which is only a performance optimization that skips the more costly detailed intersection check for many edges. We then use the full geometry for the detailed check `p.intersects(edge.fetchWayGeometry(FetchMode.ALL).makeImmutable());`.

>  On the one hand, we lose precision, since the full point list could in fact have a larger BBox than that of only the tower nodes

I agree the current logic is not very intuitive. For example if p is a rectangle and the tower node bounding box intersects p it can happen that a road is affected by a custom area even though none of the road points lies within the area (false positive). On the other hand, if the roads geometry partly lies outside of the tower node bounding box, like in your picture, it can happen that a road is **not** affected even though some points **are** located within the area (false negative).

[...]

PR #2741, opened 2023-02-04
Remove block area

### LINKED ISSUE #2630: Remove BlockAreaWeighting
The GH server still supports not specifying a `profle` and using the `vehicle`,`weighting`,... parameters instead (LegacyProfileResolver).
While the GH Directions API will probably support these parameters for legacy reasons for some time, we should remove this for the open source server.

easbar:
Fixes #2630.


core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
```diff
@@ -1,240 +0,0 @@
...
-                
-                if (point || calculateArea(shape) <= useEdgeIdsUntilAreaSize) {
-                    GHIntHashSet blockedEdges = findEdgesInShape(shape, filter);
-                    if (!blockedEdges.isEmpty()) {
-                        blockArea.add(shape, blockedEdges);
-                    }
-                } else {
-                    blockArea.add(shape);
-                }
-            }
```
easbar, line 161: With this optimization we used a hash set of blocked / affected edges instead of point-in-area checks for small areas. Not sure if we could/should do the same for (small) custom model areas?

core/src/test/java/com/graphhopper/GraphHopperTest.java
```diff
@@ -652,10 +659,13 @@ public void testNorthBayreuthBlockedEdges() {
         assertFalse(rsp.hasErrors(), rsp.getErrors().toString());
         assertEquals(510, rsp.getBest().getDistance(), 10);
 
-        // first point is contained in block_area => error
+        // first point is contained in blocked area => error
+        // todonow: what if a (via-?)point is located within the blocked area? so far we got a dedicated error in this
+        // case, but now we get connection-not-found?
         req = new GHRequest(49.979, 11.516, 49.986107, 11.507202).
```
easbar, line 665: I'm not sure what we should do here. Shouldn't we snap outside the blocked area? And should there be a dedicated exception for custom model areas similar to what we did for `block_area`?

web/src/test/java/com/graphhopper/application/resources/IsochroneResourceTest.java
```diff
@@ -272,6 +272,7 @@ public void requestNotANumber() {
         assertEquals("query param time_limit is not a number.", message);
     }
 
+    @Disabled("block_area is no longer supported and to use custom models we'd need a POST endpoint for isochrones")
```
easbar, line 275: So far we only have `GET` endpoints for `/isochrone` and `/spt` so we cannot use them with custom models. Therefore we don't have an alternative to the current `block_area` feature yet.

core/src/main/java/com/graphhopper/storage/GraphEdgeIdFinder.java
otbutz: I vaguely remember @karussell mentioning that this feature doesn't work well with virtual edges, but maybe I'm completely wrong.

Nevertheless , we should aim to improve the performance as noted in #2313. Wasting 30% of execution time and 80% of memory allocation just for bbox checking is pretty nasty.

There's also #2625 which aims to improve precision. Maybe we should start with that and see if it improves performance as well?

otbutz: > And getting rid of legacy should get a higher priority than an optimization.

Absolutely. That would be out of scope for this PR.

easbar: Ok, so let's just keep in mind that there was this optimization and maybe come back to this later.

core/src/test/java/com/graphhopper/GraphHopperTest.java
easbar: It looks like for custom model areas we do snap outside the area: [link](https://graphhopper.com/maps/?point=52.468273%2C13.547468&point=52.451589%2C13.524036&profile=car&layer=Omniscale&custom_model=%7B%22priority%22%3A%5B%7B%22if%22%3A%22in_berlin_bbox%22%2C%22multiply_by%22%3A%220%22%7D%5D%2C%22areas%22%3A%7B%22berlin_bbox%22%3A%7B%22type%22%3A%22Feature%22%2C%22properties%22%3A%7B%7D%2C%22geometry%22%3A%7B%22type%22%3A%22Polygon%22%2C%22coordinates%22%3A%5B%5B%5B13.253%2C52.608%5D%2C%5B13.228%2C52.437%5D%2C%5B13.579%2C52.447%5D%2C%5B13.563%2C52.609%5D%2C%5B13.253%2C52.608%5D%5D%5D%7D%7D%7D%7D)

Unless the points are too far from the area border in which case we get 'Point not found'. So we can leave this as is here.

easbar: But for some reason in this case we get connection not found?! This also happens on [GH Maps](https://graphhopper.com/maps/?point=49.97958%2C11.520549&point=49.981443%2C11.517759&profile=car&layer=OpenStreetMap&custom_model=%7B%22distance_influence%22%3A15%2C%22speed%22%3A%5B%5D%2C%22priority%22%3A%5B%7B%22if%22%3A%22in_my_area%22%2C%22multiply_by%22%3A%220%22%7D%5D%2C%22areas%22%3A%7B%22my_area%22%3A%7B%22type%22%3A%22Feature%22%2C%22geometry%22%3A%7B%22type%22%3A%22Polygon%22%2C%22coordinates%22%3A%5B%5B%5B11.515818%2C49.981875%5D%2C%5B11.515818%2C49.979522%5D%2C%5B11.521407%2C49.979522%5D%2C%5B11.521407%2C49.981875%5D%2C%5B11.515818%2C49.981875%5D%5D%5D%7D%7D%7D%7D)

PR #2742, opened 2023-02-05
Make sure waypoints are not removed during path simplification

### LINKED ISSUE #2515: Do not remove waypoints when instructions are disabled
Currently the snapped waypoint coordinates might be removed by the path simplification when instructions are disabled.

https://discuss.graphhopper.com/t/all-snapped-waypoints-is-not-present-in-points/7092/13

easbar:
Fixes #2515

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/Router.java
+++ b/core/src/main/java/com/graphhopper/routing/Router.java
@@ -303,7 +303,7 @@ public class Router {
     }
 
     private PointList getWaypoints(List<Snap> snaps) {
-        PointList pointList = new PointList(snaps.size(), true);
+        PointList pointList = new PointList(snaps.size(), graph.getNodeAccess().is3D());
         for (Snap snap : snaps) {
             pointList.add(snap.getSnappedPoint());
         }
--- a/core/src/main/java/com/graphhopper/routing/Router.java
+++ b/core/src/main/java/com/graphhopper/routing/Router.java
@@ -207,7 +207,9 @@ public class Router {
 
         RoundTripRouting.Result result = RoundTripRouting.calcPaths(snaps, pathCalculator);
         // we merge the different legs of the roundtrip into one response path
-        ResponsePath responsePath = concatenatePaths(request, solver.weighting, queryGraph, result.paths, getWaypoints(snaps));
+        // note that the waypoints are not just the snapped points of the snaps, as usual, because we do some kind of tweak
+        // to avoid 'unnecessary tails' in the roundtrip algo
+        ResponsePath responsePath = concatenatePaths(request, solver.weighting, queryGraph, result.paths, result.wayPoints);
         ghRsp.add(responsePath);
         ghRsp.getHints().putObject("visited_nodes.sum", result.visitedNodes);
         ghRsp.getHints().putObject("visited_nodes.average", (float) result.visitedNodes / (snaps.size() - 1));
--- a/core/src/main/java/com/graphhopper/util/PathMerger.java
+++ b/core/src/main/java/com/graphhopper/util/PathMerger.java
@@ -156,12 +156,14 @@ public class PathMerger {
             responsePath.addError(new ConnectionNotFoundException("Connection between locations not found", Collections.emptyMap()));
         }
 
-        for (int i = 0; i < wayPointIntervals.size(); i++) {
-            int start = wayPointIntervals.get(i).start;
-            int end = wayPointIntervals.get(i).end;
-            if (waypoints.getLat(i) != fullPoints.getLat(start) || waypoints.getLon(i) != fullPoints.getLon(start)
-                    || waypoints.getLat(i + 1) != fullPoints.getLat(end) || waypoints.getLon(i + 1) != fullPoints.getLon(end))
-                throw new IllegalStateException("waypoints are not included in points, or waypoint intervals are wrong");
+        if (allFound) {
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/graphhopper/routing/Router.java
```diff
@@ -207,7 +207,9 @@ protected GHResponse routeRoundTrip(GHRequest request, FlexSolver solver) {
 
         RoundTripRouting.Result result = RoundTripRouting.calcPaths(snaps, pathCalculator);
         // we merge the different legs of the roundtrip into one response path
-        ResponsePath responsePath = concatenatePaths(request, solver.weighting, queryGraph, result.paths, getWaypoints(snaps));
+        // note that the waypoints are not just the snapped points of the snaps, as usual, because we do some kind of tweak
+        // to avoid 'unnecessary tails' in the roundtrip algo
+        ResponsePath responsePath = concatenatePaths(request, solver.weighting, queryGraph, result.paths, result.wayPoints);
```
easbar, line 212: I don't really understand why we do this tweak, but if we do we should adjust the `snapped_waypoints` as well. The `snapped_waypoints` are quite meaningless for round trip, though, as the via-points are generated by the algorithm.

core/src/main/java/com/graphhopper/routing/Router.java
```diff
@@ -303,7 +305,7 @@ private ResponsePath concatenatePaths(GHRequest request, Weighting weighting, Qu
     }
 
     private PointList getWaypoints(List<Snap> snaps) {
-        PointList pointList = new PointList(snaps.size(), true);
+        PointList pointList = new PointList(snaps.size(), graph.getNodeAccess().is3D());
```
easbar, line 308: This was a bit of an inconsistency, where we used a 3D pointlist even without elevation, but it does not change the server-response, as the ResponsePathSerializer already took care of this.

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -136,33 +137,42 @@ public ResponsePath doWork(PointList waypoints, List<Path> paths, EncodedValueLo
...
                 responsePath.addPathDetails(PathDetailsFromEdges.calcDetails(path, evLookup, weighting, requestedPathDetails, pathBuilderFactory, origPoints, graph));
+                wayPointIntervals.add(new ResponsePath.Interval(origPoints, pathIndex < paths.size() - 1 ? fullPoints.size() : fullPoints.size() - 1));
                 origPoints = fullPoints.size();
             }
 
             allFound = allFound && path.isFound();
         }
 
-        if (!fullPoints.isEmpty()) {
-            responsePath.addDebugInfo("simplify (" + origPoints + "->" + fullPoints.size() + ")");
```
easbar, line 146: This debug info made no sense anyway, because origPoints was always equal to fullPoints.size()...

core/src/main/java/com/graphhopper/util/PathMerger.java
```diff
@@ -136,33 +137,42 @@ public ResponsePath doWork(PointList waypoints, List<Path> paths, EncodedValueLo
...
             fullInstructions = updateInstructionsWithContext(fullInstructions);
             responsePath.setInstructions(fullInstructions);
         }
 
         if (!allFound) {
-            responsePath.addError(new ConnectionNotFoundException("Connection between locations not found", Collections.<String, Object>emptyMap()));
+            responsePath.addError(new ConnectionNotFoundException("Connection between locations not found", Collections.emptyMap()));
+        }
+
+        if (allFound) {
```
easbar, line 159: Can we not just return the responsePath with the `connection-not-found` error as soon as one of the connections isn't found, or why do we keep going afterwards?

core/src/main/java/com/graphhopper/util/PathSimplification.java
```diff
@@ -67,6 +67,27 @@ public class PathSimplification {
...
+                return wayPointIntervals.get(index).end - wayPointIntervals.get(index).start;
+            }
+
+            @Override
+            public void setInterval(int index, int start, int end) {
+                wayPointIntervals.get(index).start = start;
+                wayPointIntervals.get(index).end = end;
+            }
+        });
+
```
easbar, line 90: This is the actual fix.

web-api/src/main/java/com/graphhopper/ResponsePath.java
```diff
@@ -119,11 +120,22 @@ public PointList getWaypoints() {
...
             throw new IllegalStateException("Cannot call setWaypoints twice");
 
         this.waypointList = wpList;
+        return this;
+    }
+
+    public List<Interval> getWaypointIntervals() {
+        check("getWaypointIntervals");
+        return waypointIntervals;
+    }
```
easbar, line 134: It could be useful to add these intervals to the server response, but this needs to be done manually in `ResponsePathSerializer`.

core/src/main/java/com/graphhopper/util/PathMerger.java
easbar: The idea is that we get a full list of all errors, not just the one for the first leg that fails. Will leave this as it is since it is rather unrelated.

web-api/src/main/java/com/graphhopper/ResponsePath.java
easbar: Will do in another PR.

PR #2752, opened 2023-02-15
Fix snap issue

karussell:
Related: https://github.com/graphhopper/graphhopper/commit/b24a3d7cf5caa34ba3222cc66392a0bec66256d9 and #2742

### DIFF
--- a/core/src/test/java/com/graphhopper/routing/querygraph/QueryGraphTest.java
+++ b/core/src/test/java/com/graphhopper/routing/querygraph/QueryGraphTest.java
@@ -391,7 +391,7 @@ public class QueryGraphTest {
         assertEquals(3, res1.getClosestNode());
         assertEquals(3, res2.getClosestNode());
 
-        // force skip due to **tower** node snapping in phase 2 (QueryOverlayBuilder.buildVirtualEdges -> prevPoint.equals(currSnapped))
+        // force skip due to **tower** node snapping in phase 2 (QueryOverlayBuilder.buildVirtualEdges -> Snap.considerEqual)
         // and no virtual edges should be created for res1
         edgeState = GHUtility.getEdge(g, 0, 1);
         res1 = createLocationResult(1, 0, edgeState, 0, TOWER);


core/src/main/java/com/graphhopper/storage/index/Snap.java
```diff
@@ -145,12 +146,28 @@ public void calcSnappedPoint(DistanceCalc distCalc) {
...
-        } else
-            // outside of edge boundaries
-            snappedPoint = new GHPoint3D(tmpLat, tmpLon, tmpEle);
+
+            // We want to prevent extra virtual nodes and very short virtual edges in case the snap/crossing point is
+            // very close to a tower node. Since we delayed the calculation of the crossing point until here, we need
+            // to correct the Snap.Position in these cases. Note that it is possible that the query point is very far
+            // from the tower node, but the crossing point is still very close to it.
+            // if crossingPoint is too close to a pillar or tower node we have to fix the snappedPosition and wayIndex to make it consistent with GHPoint::equals
+            if (NumHelper.equalsEps(crossingPoint.lat, tmpLat) && NumHelper.equalsEps(crossingPoint.lon, tmpLon)) {
```
easbar, line 157: Why does this have to be consistent, or is even related to `GHPoint#equals`? Also I would rather use `Math.abs() > threshold` here so we do not rely on whatever `equalsEps` does, which could be used for something unrelated to this code.

otbutz: Can't we use a `tmpPoint` and directly invoke `crossingPoint.equals(tmpPoint)` instead?

core/src/test/java/com/graphhopper/routing/querygraph/QueryGraphTest.java
```diff
@@ -391,9 +391,10 @@ public void testAvoidDuplicateVirtualNodesIfIdentical() {
         assertEquals(3, res1.getClosestNode());
         assertEquals(3, res2.getClosestNode());
 
-        // force skip due to **tower** node snapping in phase 2, but no virtual edges should be created for res1
+        // force skip due to **tower** node snapping in phase 2 (QueryOverlayBuilder.buildVirtualEdges -> prevPoint.equals(currSnapped))
```
easbar, line 394: Now this is Snap.considerEqual?

core/src/main/java/com/graphhopper/storage/index/Snap.java
easbar: I did not really get why this needs to be consistent, but since we now use a dedicated function (considerEqual) this works for me.

easbar (approved):
Thanks

core/src/main/java/com/graphhopper/storage/index/Snap.java
```diff
@@ -145,12 +146,31 @@ public void calcSnappedPoint(DistanceCalc distCalc) {
...
+                snappedPoint = new GHPoint3D(crossingPoint.lat, crossingPoint.lon, (tmpEle + adjEle) / 2);
+            }
+        } else {
+            // outside of edge segment [wayIndex, wayIndex+1] should not happen for EDGE
+            assert false : "incorrect pos: " + snappedPosition + " for " + snappedPoint + ", " + fullPL + ", " + wayIndex;
+        }
+    }
+
+    public static boolean considerEqual(double lat, double lon, double lat2, double lon2) {
+        return Math.abs(lat - lat2) < 1e-6 && Math.abs(lon - lon2) < 1e-6;
```
easbar, line 173: Btw for example this:

```java
System.out.println(DIST_PLANE.calcDist(49.600000, 11.600000, 49.600001, 11.600001));
```

returns 13cm(!), so not sure if this threshold makes a lot of sense: We consider points to be equal when they are within 13cm from each other, but when they are 20cm from each other we think they are different? Maybe this would be appropriate for indoor routing, but if this tweak is important the threshold should rather be in the ~10m range for road networks?

core/src/main/java/com/graphhopper/storage/index/Snap.java
easbar: This code in QueryOverlayBuilder:

```java
                    // no new virtual nodes if very close ("snap" together)
                    if (Snap.considerEqual(prevPoint.lat, prevPoint.lon, currSnapped.lat, currSnapped.lon)) {
                        res.setClosestNode(prevNodeId);
                        continue;
                    }
```

is supposed to make sure no extra virtual nodes are created when two snapping points are close to each other. An extreme case would be this:

```
    p
----x-----
    q
```                    
where the two points p and q are left/right of the edge and the snapping points are very close (or even equal). I still don't see why this should be related to the code in Snap.java that makes sure that edge snaps close to tower nodes are converted to tower snaps.

The reason the test fails once we use another tolerance here is that we do not update the snapping point  `res.getSnappedPoint()`: We simply update the closest node to be the one from the previous snap, but forget to update the snapped point (and the query distance). Basically, we first do the snapping, but then realize we want the snapping to be slightly different (close snaps should snap to the same virtual node), but do not update the modified snap correctly.

core/src/main/java/com/graphhopper/storage/index/Snap.java
easbar: If the granularity (or very short edges) did not matter, then we would not need this code at all :) . If it is important to prevent extra virtual nodes / very short virtual edges from being created I think the `1.e-6` threshold is much too small: If we tolerate 20cm difference why do we think a 13cm difference is problematic? They are still 'same' points for all practical applications.

Anyway, since currently we are not aware of any real problems related to this, we can just leave it as it is as far as I am concerned.

core/src/main/java/com/graphhopper/storage/index/Snap.java
easbar: You mean we should not do it in QueryOverlayBuilder? Why not? We cannot really do it before we got all the snaps and the crossing points, because first we need to know which query points should use the same virtual node?

easbar: If we don't want to modify the snaps we need to distinguish between 'Snaps' (immutable result return types of the location index) and the things used by QueryGraph/Overlay to create virtual nodes/edges. But currently they are the same.

easbar: Or if we want to use the same objects but don't want to modify the fields that were determined by location index, we need to add separate fields for this purpose (snappedPoint + resolvedPoint or something).

easbar: Yes, in the current state I think this is the best we can do.

PR #2754, opened 2023-02-16
Upstream Resource

chansbtran:
This PR is for creating an upstream resource for getting the 'upstream' path from a point. This is forked from the Buffer-Resource branch, utilizing a lot of functionality from there. The main difference being that we only want one side of the buffer returned, not both.


chansbtran:
I made the mistake of creating this pull request too early. I meant to create this for my fork with my organization first.

PR #2763, opened 2023-02-24
Add new field `waypoint_indices` to routing response

easbar:
When we request routes with via-points from `/route` we get one list of coordinates that represents the entire path, but not single lists of points ('legs') for the path segments between the different points. Without using workarounds like enabling instructions or path details we cannot tell which part of the list belong to which 'leg' of the route.

This PR adds an array of indices pointing into the list of points that indicate which coordinates represent the via points.

Say, for example, we request a route from A via B and C to D, then we get a result like this:

```
points: `[c0, c1, c2, c3, c4, c5, c6, c7, c8, c9, c10, c11, c12]`
waypoint_indices: [0, 3, 10, 12]
```

where `cn` are coordinates. This tells us that `[c0, c1, c2, c3]` is the path segment between points A and B, `[c3, ..., c10]` the one between B and C, and `[c10, c11, c12]` the one between C and D.

Another useful thing could be the time,distance and ascend/descend for the single legs(?).


easbar:
We decided to put this information into path details instead: #2768

PR #2765, opened 2023-03-06
Add landuse encoded value based on closed-ring ways tagged with landuse=*

easbar:
I built a quick prototype, which:

* stores the way tags and OSM node Ids for all OSM ways tagged with `landuse=*` in memory during the first pass of OSMReader

* finds all the corresponding node coordinates in the second pass

* builds an area index for all these (closed-ring) polygons before we read the OSM ways again in the second pass

* finds the areas each edge is contained in and feeds them to the tag parsers

* adds a tag parser for the new landuse encoded value, which uses the smallest of these areas to set the landuse value

This way we can, e.g., use custom models that prefer forest areas, or slow down the speed in residential areas etc.
We can also just look at the landuse values using path details:

image

My first quick test resulted in an import time of 117s compared to 80s for a map of Bavaria with the master branch. 

The biggest issue I ran into so far was that many forest areas seem to be mapped as relations with such polygons as members. Parsing the relations wouldn't be much harder, but since we only get to see them at the end of the first pass (after the ways) we would have to read the OSM file a third time :(

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMLanduseParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMLanduseParser.java
@@ -1,0 +1,52 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.routing.util.parsers;
+
+import com.graphhopper.reader.ReaderWay;
+import com.graphhopper.routing.ev.EnumEncodedValue;
+import com.graphhopper.routing.ev.Landuse;
+import com.graphhopper.routing.util.CustomArea;
+import com.graphhopper.storage.IntsRef;
+
+import java.util.Comparator;
+import java.util.List;
+
+import static com.graphhopper.routing.ev.Landuse.OTHER;
+
+public class OSMLanduseParser implements TagParser {
+
+    private final EnumEncodedValue<Landuse> landuseEnc;
+
+    public OSMLanduseParser(EnumEncodedValue<Landuse> landuseEnc) {
+        this.landuseEnc = landuseEnc;
+    }
+
+    @Override
+    public void handleWayTags(IntsRef edgeFlags, ReaderWay readerWay, IntsRef relationFlags) {
+        List<CustomArea> osmAreas = readerWay.getTag("gh:osm_areas", null);
+        if (!osmAreas.isEmpty()) {
+            osmAreas.sort(Comparator.comparing(CustomArea::getArea));
+            // todonow: we simply use the smallest one for now
+            CustomArea osmArea = osmAreas.get(0);
+            String landuseStr = (String) osmArea.getProperties().get("landuse");
+            Landuse landuse = Landuse.find(landuseStr);
+            if (landuse != OTHER)
+                landuseEnc.setEnum(false, edgeFlags, landuse);
+        }
+    }
+}
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMAreaData.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMAreaData.java
@@ -1,0 +1,90 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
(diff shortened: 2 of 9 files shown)


easbar:
> The biggest issue I ran into so far was that many forest areas seem to be mapped as relations with such polygons as members. 

The [OSM wiki](https://wiki.openstreetmap.org/wiki/Key:landuse) even says that landuse should **not** be tagged on relations, but there are currently around 1.8mio such relations in OSM :(

easbar:
I added support for multipolygon landuse relations in my second commit (using a third pass of the OSM file). At least for Bavaria this improves the handling of landuse=forest a lot. Here is a bike route that tries to stay in the forest as much as possible:

image

ratrun:
This branch looks very promising, although it is quite expensive in terms of RAM and preparation time. I tested the extract for Austria with four bicycle profiles active and had to increase the RAM size from 3GB to 5GB.
I just noticed a litte draw-back and this is that in some areas the way does not overlap with the landuse polygon, it is just very very close. In such cases it would be great if it would be possible to utilize the near by landuse polygon. Do you see the possibility for an enhancment to take a lets say a 20 meters margin into account?

easbar:
> I just noticed a litte draw-back and this is that in some areas the way does not overlap with the landuse polygon, it is just very very close. In such cases it would be great if it would be possible to utilize the near by landuse polygon. Do you see the possibility for an enhancment to take a lets say a 20 meters margin into account?

Yes, maybe that is actually quite a good idea. Also because in some cases for example forest areas are separated by roads like here:

image

even though the road clears leads 'through' the forest.

otbutz:
Should be doable using jts? https://locationtech.github.io/jts/javadoc/org/locationtech/jts/operation/buffer/BufferOp.html

easbar:
> Should be doable using jts?

I did not look into it yet, but yes this sounds like what we need, thanks! Precision isn't very important for us here I think, we just need to widen the polygons a bit. The tricky part might be dealing with overlapping landuse polygons then.

otbutz:
JTS should be able to properly merge them.

easbar:
> JTS should be able to properly merge them.

I meant landuse polygons with different tag values. For example when a road is located near the border between an forest and a farmland area and we extend both polygons we need to decide which landuse value we like to use for this road.

Btw here is a nice visualization of the landuse tag: https://osmlanduse.org

otbutz:
I thought you meant overlapping parts of multipolygons. Aren't different overlapping landuse polygons already quite common? It would be nice if they weren't and everyone made proper multipolygons with holes in them, but in reality there's going to be quite a lot of overlap.

You'll most likely need to define some kind of sorting behaviour. e.g preferring non-buffered matches and polygons with a smaller area.

otbutz:
It's probably easier to leave the polygons as they are and instead make the search a little less precise.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMLanduseParser.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public OSMLanduseParser(EnumEncodedValue<Landuse> landuseEnc) {
+        this.landuseEnc = landuseEnc;
+    }
+
+    @Override
+    public void handleWayTags(int edgeId, EdgeIntAccess edgeIntAccess, ReaderWay readerWay, IntsRef relationFlags) {
+        List<CustomArea> osmAreas = readerWay.getTag("gh:osm_areas", null);
+        if (!osmAreas.isEmpty()) {
+            osmAreas.sort(Comparator.comparing(CustomArea::getArea));
```
otbutz, line 44: Maybe create a buffer around the edge and search for the OSM area with biggest "hit ratio" ? e.g sorting based on the size of intersecting area.

otbutz:
I'd love to see a follow up PR that uses the landuse information to augment the urban density EV.

ratrun:
I created PR #2812 which implements the mentioned search buffer from above. @easbar Can you please check?

easbar:
>  although it is quite expensive in terms of RAM and preparation time. I tested the extract for Austria with four bicycle profiles active and had to increase the RAM size from 3GB to 5GB.

The RAM requirement should be much lower now, can you please try again? Note that I also removed the relation handling for now, and maybe do this in a follow-up.

easbar:
> Also because in some cases for example forest areas are separated by roads like here

The same problem exists for residential areas where roads might be excluded explicitly from the residential landuse areas, like here (extract from Berlin):

image

Then again, in these cases the roads are probably often tagged as `highway=residential`, too.

ratrun:
> Querying with a buffer like you proposed should fix this.

Yes, it does:

BerlinLanduse

karololszacki:
Hi, what's the status of this PR? I'd love to see this merged in (and with the buffer improvement), is there anything I could maybe help with?

otbutz:
The use case, at least from my point of view, is quite clear. We have UrbanDensity as a **heuristic** approach to detecting built-up areas. Paired with the landuse EV, we have a reliable solution to detect roads in and out of the city.

And then there's the use case mentioned by @ratrun: "I'd like to prefer cycling routes through green areas.

ratrun:
> And then there's the use case mentioned by @ratrun who'd like to prefer cycling routes through green areas.

In the meantime I think that merging of #3039, which became possible with the implementation in here, is more beneficial compared to the landuse encoded value. The reason is that the computation of the landuse encoded value is  more costly in terms of RAM and time during graph preparation time compared to the benefit it has.

karololszacki:
Yes, I'd love to use it for preferring pretty landscapes and forestry areas for "outdoor" use cases. 
[Landcover](https://wiki.openstreetmap.org/wiki/Proposal:Landcover) I think would be even better for this, but it does not have widespread usage yet.
There is also the [natural key](https://wiki.openstreetmap.org/wiki/Key%3Anatural) with eg. coastline value (or wood, or water), but parsing this would be even more tricky.

What I like about this PR most, however, is the extra pass in WaySegmentParser and the new pre- and post- hooks introduced. This would make extending custom solutions for Graphhopper much easier.
Regarding the concerns on import time, I'd say it should be an optional processing step that is off by default. Then, it could be turned on explicitly if needed.

Issue #2139, opened 2020-10-01
Suggested route includes illegal u-turn


camelCaseSucks:
This route includes an illegal u-turn in the middle of the bridge

https://graphhopper.com/maps/?point=45.501467%2C-122.661254&point=45.501237%2C-122.66142

OSRM returns the expected route

https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=45.5015%2C-122.6613%3B45.5012%2C-122.6613

easbar:
Thanks for reporting!

This seems to be a u-turn, because if I set u_turn_costs to infinity I get the same route as on OSRM:

https://graphhopper.com/maps/?point=45.501467%2C-122.661254&point=45.501237%2C-122.66142&ch.disable=true&u_turn_costs=-1

Apparently there is a tower node in the middle of the bridge, even though there is no junction. There is a region border though:

![image](https://user-images.githubusercontent.com/17603532/94778936-5f2b1900-03c6-11eb-8c4f-527be8f2bcb0.png)

@karussell any idea why we get a tower node here?

Btw: Another I ran into another problem: When I click on the above link (in @camelCaseSucks comment) and then change the vehicle to foot and back to car again I get another route once again, because for some reason `turn_costs=false` is appended to the url...

karussell:
> @karussell any idea why we get a tower node here?

No, this looks strange. As this is the main issue for me here - should I rename the issue title?

> because for some reason turn_costs=false is appended to the url...

outch, this is likely yet another bug in the UI :/

otbutz:
> @karussell any idea why we get a tower node here?

https://www.openstreetmap.org/way/587574521

easbar:
>  https://www.openstreetmap.org/way/587574521

What do you think is suspicious here?

otbutz:
It might be the line crossing the bridge

easbar:
> It might be the line crossing the bridge

Yes, it really looks like it is :) But this is just a region border, so it should not even be considered by our OSM parser? If it was a ferry I would totally understand what happens because this would create a 'ghost junction' : #1858

boldtrn:
Isn't it because the way in OSM is cut at the region border, so we get a tower node by definition? If one wants to avoid that, you would have to merge the features in OSM.

easbar:
[...]

PR #2767, opened 2023-03-07
Show u-turn instruction for plain u-turns

easbar:
Currently there are no u-turn instructions for 'simple' u-turns where we simply turn around on the same road/edge. We only show instructions for such u-turns when the occur at via-points, and we even show instructions for the (more complicated) u-turns that use a via-way:

image

The image shows a u-turn at the via-point 1, blue marker. We get an instruction for this, but not the u-turn at the end of the dead-end road near the red marker. The reason we get this u-turn is a curbside constraint at the red marker.

Here is an example for a u-turn that includes a via-way, and for which we already show a u-turn instruction:

image

In this PR I added a u-turn instruction for the simple kind of u-turns, like the one near the red marker in the first screenshot.

I also wonder if we should distinguish between these simple (but often dangerous) u-turns and the ones with via-ways, that might be perfectly safe (or completely forbidden when there is a no-u-turn sign...).

Related: #2139 where a user suggested that the sharp/simple/dangerous u-turns should be avoided as much as possible anyway. I think I agree, but at the very least we should make these turns more visible, hence this PR. For example, a navigation application could inform the driver that the route leads to a u-turn and instead of having to turn around all of a sudden the driver could start looking for a safe option to turn around.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
+++ b/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
@@ -332,6 +332,9 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
     }
 
     private int getTurn(EdgeIteratorState edge, int baseNode, int prevNode, int adjNode, String name, String destinationAndRef) {
+        if (edge.getEdge() == prevEdge.getEdge())
+            // this is the simplest turn to recognize, a plain u-turn.
+            return Instruction.U_TURN_UNKNOWN;
         GHPoint point = InstructionsHelper.getPointForOrientationCalculation(edge, nodeAccess);
         double lat = point.getLat();
         double lon = point.getLon();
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
```diff
@@ -332,6 +332,9 @@ public void finish() {
     }
 
     private int getTurn(EdgeIteratorState edge, int baseNode, int prevNode, int adjNode, String name, String destinationAndRef) {
+        if (edge.getEdge() == prevEdge.getEdge())
+            // this is the simplest turn to recognize, a plain u-turn.
+            return Instruction.U_TURN_UNKNOWN;
```
easbar, line 337: Basically, this is my entire fix. But since I'm not very familiar with the instruction calculation this might not be the best place to handle the u-turn.

easbar: @boldtrn, maybe you are more familiar with this?

boldtrn: Sorry for the late reply here, this actually seems quite good, no complaints from my side 👍

PR #2768, opened 2023-03-07
Add path details for the time, distance and weight of the single legs

easbar:
When we query `/route` with via-points it returns the time, distance, weight and coordinates of the entire route. But so far there is no direct information about the time/distance/weight of the single sections between the (via) points. Also it is not very clear how the global point list can be separated for the single sections ('legs'). We can work around this using instructions or path details like 'time', but this is rather cumbersome and enabling instructions or certain path details just for this purpose can decrease performance. 

Therefore, I added three new path details, e.g. with a single via-point we get two entries (because there are two legs) for each detail:

```json
{
	"details": {
		"leg_time": [
			[
				0,
				6,
				48559
			],
			[
				6,
				10,
				60719
			]
		],
		"leg_distance": [
			[
				0,
				6,
				501.46548797983564
			],
			[
				6,
				10,
				505.98093195638285
			]
		],
		"leg_weight": [
			[
				0,
				6,
				48.55733718679518
			],
			[
				6,
				10,
				60.717711834765936
			]
		],
	}
}
```

`leg_time/distance/weight` simply represent the time/distance/weight for the corresponding leg. 

Also note that the first/last indices are equal for all details. They can be used to obtain the coordinates for a certain leg. In this example the points 0-6 represent the leg between the start and the via-point, and points 6-10 represent the leg between the via-point and the destination point.

### DIFF
--- a/web/src/test/java/com/graphhopper/application/resources/RouteResourceClientHCTest.java
+++ b/web/src/test/java/com/graphhopper/application/resources/RouteResourceClientHCTest.java
@@ -46,6 +46,8 @@ import org.junit.jupiter.params.provider.EnumSource;
 import java.io.File;
 import java.util.*;
 
+import static com.graphhopper.util.Instruction.FINISH;
+import static com.graphhopper.util.Instruction.REACHED_VIA;
 import static org.junit.jupiter.api.Assertions.*;
 
 /**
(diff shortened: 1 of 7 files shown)


web/src/test/java/com/graphhopper/application/resources/RouteResourceClientHCTest.java
```diff
@@ -379,4 +381,86 @@ public void testCustomModel() {
...
+        List<PointList> legs = new ArrayList<>();
+        List<PathDetail> legDetails = path.getPathDetails().get(detail);
+        for (PathDetail legDetail : legDetails) {
+            PointList leg = new PointList(legDetail.getLast() - legDetail.getFirst() + 1, path.getPoints().is3D());
+            for (int j = legDetail.getFirst(); j <= legDetail.getLast(); j++)
+                leg.add(path.getPoints(), j);
+            legs.add(leg);
+        }
+        return legs;
+    }
```
easbar, line 456: This is how we can use the new path details to subdivide the global point list into multiple point lists for each leg.

Sawa-i:
Hi, Thanks for the commit, just what I was looking for but wanted to inquire what importance does 'leg' field has? I was expecting route sections to be in leg field ( ArrayList ) where each leg would have their information including the coordinates, legTime and legDistance.

image

Kindly guide me as I'm still understanding.

michaz:
@Sawa-i That one is used for public transit, where there's an alternation of access / egress / transfer by walking, and taking transit connections.

The legs in the sense used in this ticket (legs between via points) could well go in there, too, but we're not there yet.

Sawa-i:
> @Sawa-i That one is used for public transit, where there's an alternation of access / egress / transfer by walking, and taking transit connections.
> 
> The legs in the sense used in this ticket (legs between via points) could well go in there, too, but we're not there yet.

I see, Thanks for quick response. 

It'd be great though, getting legwise information (current issue context) though we'd have to think up some other name. And when done so, wouldn't need the leg_distance and leg_time in pathDetails.

michaz:
@Sawa-i Exactly. I'm still hoping we can do without finding another name, and put it under the same umbrella.

PR #2770, opened 2023-03-08
Allow skipping the creation of ReaderNodes/Ways/Relations for pbfs, skip nodes in first pass of OSM import

easbar:
Currently we read the OSM file twice during import and there even were some suggestions to read it three times (e.g. for #2765 or #82). With the simple tweak in this PR we can reduce the time it takes to read the file if we are only interested in certain OSM elements.

I did a quick check using a map of Germany with ~356mio nodes, ~58mio ways and ~0.7mio relations, with two worker threads. Reading the file (just counting the elements, and doing nothing else) took around 113s using master, around 73s using this branch when I skipped the nodes and 56s when I skipped both nodes and ways.

We can already use this for our current import where we can skip the reading of nodes in the first pass, which speeds up the import.

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/pbf/PbfBlobDecoder.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/pbf/PbfBlobDecoder.java
@@ -8,6 +8,7 @@ import com.graphhopper.reader.ReaderNode;
 import com.graphhopper.reader.ReaderRelation;
 import com.graphhopper.reader.ReaderWay;
 import com.graphhopper.reader.osm.OSMFileHeader;
+import com.graphhopper.reader.osm.SkipOptions;
 import com.graphhopper.util.Helper;
 import org.openstreetmap.osmosis.osmbinary.Fileformat;
 import org.openstreetmap.osmosis.osmbinary.Osmformat;
@@ -33,6 +34,7 @@ public class PbfBlobDecoder implements Runnable {
     private final byte[] rawBlob;
     private final PbfBlobDecoderListener listener;
     private List<ReaderElement> decodedEntities;
+    private final SkipOptions skipOptions;
 
     /**
      * Creates a new instance.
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/graphhopper/reader/osm/pbf/PbfBlobDecoder.java
```diff
@@ -326,10 +329,14 @@ private void processOsmPrimitives(byte[] data) throws InvalidProtocolBufferExcep
         PbfFieldDecoder fieldDecoder = new PbfFieldDecoder(block);
```
easbar, line 329: I briefly tried creating the PbfFieldDecoder lazily, only when the current block contains some elements we want to keep, but it made no real difference. So I think unless we do something on a lower level (pbf decoding) this is the earliest we can skip some work here.

otbutz: I wouldn't want to invest a lot of time in this task. The consensus is that the entire input pipeline will need to be rebuilt at some point. At least that's what I've taken away from previous discussions.

easbar: Yes, this kind of was my point: Skipping this step is easy enough and already yields a nice speed-up.

otbutz: I took another look and apparently the codebase has improved quite a bit since then. Most of my previous pain points have been fixed by https://github.com/graphhopper/graphhopper/pull/2448.

It would still be nice to ditch our forked osmosis parts and use the upstream version as a dependency instead:

https://github.com/graphhopper/graphhopper/blob/53d87aead93ee6753bbb157add63235196bab5e2/core/src/main/java/com/graphhopper/reader/osm/pbf/PbfReader.java#L11-L21

https://github.com/openstreetmap/osmosis/blob/447e5df474584ee2d9ff97b3c26e677968d5331e/osmosis-pbf2/src/main/java/org/openstreetmap/osmosis/pbf2/v0_6/PbfReader.java#L27-L36

PR #2782, opened 2023-03-31
non functional code but ready to get conceptual feedback

christophlingg:
Hello again,

this draft PR describes an idea on how to extend the OSMReader to be more flexible in terms of custom implementations for smoothing and line simplification. It allows to experiment, develop and adjust the algorithms more easily. We are using graphhopper as a library and it was hard - almost impossible - to tune and evaluate new ideas that ultimately led to https://github.com/graphhopper/graphhopper/pull/2772

This is meant to get early feedback and confirmation for this approach. Afterwards I would wrap it up and make it production ready.

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -27,9 +27,9 @@ import com.graphhopper.reader.ReaderElement;
 import com.graphhopper.reader.ReaderNode;
 import com.graphhopper.reader.ReaderRelation;
 import com.graphhopper.reader.ReaderWay;
-import com.graphhopper.reader.dem.EdgeElevationSmoothing;
 import com.graphhopper.reader.dem.EdgeSampling;
 import com.graphhopper.reader.dem.ElevationProvider;
+import com.graphhopper.reader.osm.pointlist.*;
 import com.graphhopper.routing.OSMReaderConfig;
 import com.graphhopper.routing.ev.Country;
 import com.graphhopper.routing.util.AreaIndex;
@@ -83,6 +83,8 @@ public class OSMReader {
     private final OSMParsers osmParsers;
     private final DistanceCalc distCalc = DistanceCalcEarth.DIST_EARTH;
     private final RestrictionSetter restrictionSetter;
+    private PointListProcessor elevationSmoother;
+    private PointListProcessor edgeSimplifier;
     private ElevationProvider eleProvider = ElevationProvider.NOOP;
     private AreaIndex<CustomArea> areaIndex;
     private CountryRuleFactory countryRuleFactory = null;
@@ -105,14 +107,28 @@ public class OSMReader {
         this.osmParsers = osmParsers;
         this.restrictionSetter = new RestrictionSetter(baseGraph);
 
-        simplifyAlgo.setMaxDistance(config.getMaxWayPointDistance());
-        simplifyAlgo.setElevationMaxDistance(config.getElevationMaxWayPointDistance());
         turnCostStorage = baseGraph.getTurnCostStorage();
 
         tempRelFlags = osmParsers.createRelationFlags();
         if (tempRelFlags.length != 2)
             // we use a long to store relation flags currently, so the relation flags ints ref must have length 2
             throw new IllegalArgumentException("OSMReader cannot use relation flags with != 2 integers");
+
+        this.edgeSimplifier = new RamerDouglasPeuckerProcessor(
+                config.getMaxWayPointDistance(),
+                config.getElevationMaxWayPointDistance()
+        );
+
+
+        if (config.getElevationSmoothing().equals("ramer"))
+            this.elevationSmoother = new ElevationSmootherRamer();
+        else if (config.getElevationSmoothing().equals("moving_average"))
+            this.elevationSmoother = new ElevationSmootherAverage();
+        else if (config.getElevationSmoothing().equals(""))
+            this.elevationSmoother = new PointListProcessorNoOp();
+        else {
+            throw new AssertionError("Unsupported elevation smoothing algorithm: '" + config.getElevationSmoothing() + "'");
+        }
     }
 
     /**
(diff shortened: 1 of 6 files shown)


christophlingg:
not sure if draft pr trigger any notifications @karussell

christophlingg:
I assume you see this as a way to allow adjustments or enhancements without the need to feed it back to this project. That is not our intention and does not reflect our stance towards open source.

We want this flexibility to make it easier for us (and any other contributors) to experiment with improvements. Currently there is no way to test and benchmark improvements when graphhopper is used as a library. We spent around 3 days to find some workaround to get #2772 tested. We would like to spend this time rather on improving graphhopper.

christophlingg:
> I don't understand what you mean. Why does it take 3 days and why can't you use your branch and test it out immediately?

Because for an evaluation we need to run it in our code base and with graphhopper as a library. This allows it to evaluate the change in our komoot-specific test suits and with our evaluation tools which goes against our routing API. I needed to implement the algorithm first in some hacky way (via overwritting `OSMReader.processWay` and some heuristics to protect elevation values of tower nodes from being changed). Once evaluation was finished I needed to implement another time including tests to feed it back to graphhopper. I guess forking and making maven releases in some other repository (I am not the java/maven expert here) would be an alternative. Still not joyful or quick to change anything.

I know about the challenge to strike a good balance between flexibility and simplicity/maintainability. It's your call here. I can just tell you it was no fun and I would have had an easier life with this flexibility.

otbutz:
Wouldn't it be easier for komoot to upstream some of these tests? This work will easily pay off as the GraphHopper project will keep an eye on the test results as the code base evolves.

christophlingg:
The test suit i was referring to are not unit tests. It is more like for gravel routing between from point A to point B, we don't expect to take a certain way.

easbar:
> Probably you will now tell me you can publish an artifact to some internal maven repo, right?

Yes. Anyone who likes to modify GraphHopper can do this very easily by forking the code. The problem many people seem to struggle with is then deploying the modified code into their production systems, i.e. making available the modified jar. We like to share our code, but we don't feel responsible for providing infrastructure for such deployments. There are many ways to do this, for example [Github Packages](https://github.com/features/packages).

Then again, there certainly are cases where many users of the library can benefit from such changes without creating a lot of maintenance overhead for us in the future.  

In this specific example: Would it be possible to just move the code snippet that does the point list modification in OSMReader into a separate method and make it protected?

easbar:
> We would like to spend this time rather on improving graphhopper.

That is very nice to hear and highly appreciated. If you will do more such experiments in the future it is probably very much worth it to implement some kind of workflow on your side for trying out your changes quickly without having to wait for us. Like I said, all you should really need is a fork of this repository and some kind of way to get the modified jar into your system.

otbutz:
> Probably you will now tell me you can publish an artifact to some internal maven repo, right?

Just my two cents but running your own fork isn't that bad. As a company you'll always have a few changes which are simply not suitable for upstreaming. We currently rebase our patches after a new GH release and try to contribute upstream whenever it's possible to do so :slightly_smiling_face:

mwolf-net:
Hi , I am a colleague of Christoph in the Komoot routing team.

@easbar:
> If this was only useful for experiments and evaluation of new algorithms I'd say this is nothing we should add to this repository, because experiments should be done by just forking the code. But if we find that it is useful to switch between different smoothing algorithms depending on the use-case, the elevation data, area etc. I wouldn't be against making this configurable.

Indeed that's the idea, to make the library more customisable in order to support a wide variety of use cases. Just as the vehicle profiles and weightings are customisable, because the out-of-the-box defaults are not optimal for every project, we think it would be valuable to make the graph import process a bit more customisable too. E.g. a product aimed at mountain bikers may have different elevation smoothing requirements than a product meant for car navigation. Tweaking the formula for different DEMs is another good example indeed.

So the basic idea is: a generic interface with a default implementation; if you are happy with the default implementation then you don't even need to know that the interface is there; if you want to do more fine-grained tweaking you have the ability to override it. Graphhopper already uses such generic interfaces in a lot of places, but the elevation smoothing is an example where there currently isn't really a clean way to override it with a custom implementation.

Indeed for experimental purposes it is fine to create a local fork. That's how we created the other PR. But we think in general it would be good if, for the majority of common use cases, it is possible to customise the library to one's needs by overriding specific interfaces rather than by maintaining a long-lived fork of the entire library.

easbar:
> But we think in general it would be good if, for the majority of common use cases, it is possible to customise the library to one's needs by overriding specific interfaces rather than by maintaining a long-lived fork of the entire library.

Yes, the tricky part is identifying the 'common' use cases. I agree that it can be helpful when the library code is flexible and customizable, but at the other end of the spectrum we would have an overly generic library with more maintenance burden that allows customizing everything even though such customization is only needed for a minority of users. Our general heuristic to figure out if something needs to be customizable is asking ourselves if we need to customize it.

That said, the current code is certainly not perfect and it is always possible that we can even make things both more customizable and simpler, or at least more flexible without making it more complicated.

>  E.g. a product aimed at mountain bikers may have different elevation smoothing requirements than a product meant for car navigation. Tweaking the formula for different DEMs is another good example indeed.

I'd really like to see an example where it is clear that one kind of smoothing algorithm is superior to another for two different vehicles or DEMs. I'll also take a look and see if we can simplify the OSM reader code to make it easier to switch the smoothing code.

mwolf-net:
@easbar Thanks for the thoughtful response!

I can certainly see the point of "Our general heuristic to figure out if something needs to be customizable is asking ourselves if we need to customize it." OTOH since this is a library meant to be usable by a wide variety of different parties and projects, maybe it also makes sense to err on the side of using overridable interfaces in the design, as long as that can be done without introducing extra complexity for the default usage. (IOW if you're happy with the default then you don't need to do anything special.)

> I'd really like to see an example where it is clear that one kind of smoothing algorithm is superior to another for two different vehicles or DEMs.

For different DEMs, it would probably be not so much a completely different algorithm, but more a tweaking of the parameters. Depending on the resolution of the DEM you may want to use a wider or narrower window size. Of course making the window size configurable can be done without making the entire algorithm replaceable, but then you need to expose the algorithm parameters on the OSMReader class.

For different vehicles, it would perhaps be more a matter of speed/accuracy tradeoff. If you really care about elevation you may want to do some kind of fancy interpolation which is slow and overkill for most projects, while for the majority of projects the current sliding-window approach is sufficient. For car navigation, the main reason why smoothing is desirable at all is to avoid the scenario where a road winds around a cliff and the DEM thinks some parts of the road are on the bottom of the cliff while others are at the top.

Anyway, it's a not a major issue for us right now -- we're using the new version of the algo which we submitted in the other PR, and we're happy with that. And if we want to experiment with more changes we can always do a local fork indeed. But if you'd like to take a further look into whether you think introducing some overridable interfaces is worth it, we'd be very interested in your findings and we'd be happy to help out!

Cheers,

Martin

PR #2785, opened 2023-03-31
Improve bicycle handling: Now prioritise cycleways with SLIGHT_PREFER

### LINKED ISSUE #2784: Bike routing prefers detour without cycleway
https://graphhopper.com/maps/?point=48.726001%2C12.134477_Ettenkofen%2C+84088+Neufahrn+in+Niederbayern%2C+Deutschland&point=48.705687%2C12.200848&profile=bike&layer=Omniscale

The main road is tagged with `cycleway=lane` but GH favors a detour using a road without any bike friendly features.

Main road: https://www.openstreetmap.org/way/552856505

Side road: https://www.openstreetmap.org/way/86263970

ratrun:
Now that we have 4 bits for the priority available and not only 3 like in the past we can increase the priority for `cycleway=lane` by one step. Fixes #2784:

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
@@ -177,7 +177,7 @@ public abstract class BikeCommonPriorityParser implements TagParser {
 
         String cycleway = way.getFirstPriorityTag(Arrays.asList("cycleway", "cycleway:left", "cycleway:right"));
         if (Arrays.asList("lane", "shared_lane", "share_busway", "shoulder").contains(cycleway)) {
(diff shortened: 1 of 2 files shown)


bt90:
@ratrun i took a different approach: https://github.com/graphhopper/graphhopper/pull/2786

bt90:
My interpretation:

* `shared_lane`, `share_busway` and `shoulder` should be graded with `UNCHANGED`, as they offer low protection to the cyclist
* `lane` is a bit better in that regard and IMHO deserves `PREFER`
* `track` is obviously the best value. I'd argue that we should rather change this one to `VERY_NICE`

bt90:
The missing bits in your PR would be `opposite_lane` and `opposite_track` handling.

I also added support for `cycleway:*:bicycle=designated` which is a local tagging scheme used in Germany and Poland.

PR #2786, opened 2023-03-31
Improve cycleway priority handling

bt90:


core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
```diff
@@ -180,7 +180,8 @@ void collect(ReaderWay way, double wayTypeSpeed, TreeMap<Double, Integer> weight
         String cycleway = way.getFirstPriorityTag(Arrays.asList("cycleway", "cycleway:both", "cycleway:left", "cycleway:right"));
         if (SHARED_CYCLEWAYS.contains(cycleway)) {
             weightToPrioMap.put(100d, UNCHANGED.getValue());
-        } else if ("lane".equals(cycleway) || "track".equals(cycleway)) {
+        } else if ("lane".equals(cycleway) || "track".equals(cycleway)
+                || "opposite_lane".equals(cycleway) || "opposite_track".equals(cycleway)) {
```
ratrun, line 187: Here I would keep the `opposite_lane` as unchanged. I read a report that accident numbers here are pretty high and this sounded reasonable to me. But slightlig boosing `opposite_track` sound reasonable, I would give this a SLIGHT_PREFER.

core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
```diff
@@ -149,7 +151,8 @@ private PriorityCode convertClassValueToPriority(String tagvalue) {
      */
     void collect(ReaderWay way, double wayTypeSpeed, TreeMap<Double, Integer> weightToPrioMap) {
         String highway = way.getTag("highway");
-        if (way.hasTag("bicycle", "designated") || way.hasTag("bicycle", "official")) {
+        if (way.hasTag("bicycle", "designated") || way.hasTag(CYCLEWAY_ACCESS_KEYS, Arrays.asList("designated"))
+                || way.hasTag("bicycle", "official")) {
```
ratrun, line 155: This change looks good to me.

ratrun (changes requested):
I would suggest that you revert your three first commits and I include the slight boosting of the opposite `track`into #2785. If you additionally provide a unit test for your commit number 4 then we could suggest @karussell to review both both PRs.

bt90:
Sounds good :+1:

ratrun (approved):
The only thing I would change is the title this PR. Access tags is something different. I would call it something like : Priority handling for tag "cycleway:bicycle"

bt90:
In the context of OSM these are indeed access restrictions.

https://wiki.openstreetmap.org/wiki/Key:bicycle

The `cycleway:*:` prefix just lowers the scope.

bt90:
My bad. I picked a highway type which is already preferred.

PR #2788, opened 2023-04-03
certain conditional tags allow to ignore max_weight

karussell:
See e.g. [here](https://www.openstreetmap.org/way/4188422) that has the tag `hgv:conditional=destination @ (maxweight > 5)` which specifies an exception to the `maxweight=5` restriction, allowing heavier HGVs access if they are destined for a specific location. In such a case, regardless of the maxweight tag then it would follow `max_weight_except=DESTINATION`.

Make ignoring maxweight possible in a custom model too. The usage would be:

```json
{ 
  "priority": [{
    "if": "max_weight < 6 && max_weight_except != DESTINATION",
    "multiply_by": "0"
  }]
}
```

The encoded value is named max_weight_except (although "exception" would be probably better but then this is confusing as a Java class)

Also this PR uses conditional tagging to feed hgv and max_weight. And finally adds `maxweightrating:hgv` parsing for max_weight as this seems to be the latest tag for `maxgcweight`.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/MaxWeightExceptParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/MaxWeightExceptParser.java
@@ -9,13 +9,14 @@ import com.graphhopper.routing.util.parsers.helpers.OSMValueExtractor;
 import com.graphhopper.storage.IntsRef;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 import static com.graphhopper.routing.util.parsers.helpers.OSMValueExtractor.stringToTons;
 
 public class MaxWeightExceptParser implements TagParser {
     private final EnumEncodedValue<MaxWeightExcept> mweEnc;
     private static final List<String> HGV_RESTRICTIONS = OSMRoadAccessParser.toOSMRestrictions(TransportationMode.HGV).stream()
-                .map(e -> e + ":conditional").collect(Collectors.toList());
+            .map(e -> e + ":conditional").collect(Collectors.toList());
 
     public MaxWeightExceptParser(EnumEncodedValue<MaxWeightExcept> mweEnc) {
         this.mweEnc = mweEnc;
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxWeightParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMMaxWeightParser.java
@@ -26,13 +26,14 @@ import com.graphhopper.storage.IntsRef;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class OSMMaxWeightParser implements TagParser {
 
     // do not include OSM tag "height" here as it has completely different meaning (height of peak)
     private static final List<String> MAX_WEIGHT_TAGS = Arrays.asList("maxweight", "maxgcweight"/*abandoned*/, "maxweightrating:hgv");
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/MaxWeightExceptParser.java
```diff
@@ -0,0 +1,57 @@
...
+    public void handleWayTags(int edgeId, EdgeIntAccess edgeIntAccess, ReaderWay way, IntsRef relationFlags) {
+        // tagging like maxweight:conditional=no/none @ destination/delivery/forestry/service
+        String condValue = way.getTag("maxweight:conditional", "");
+        if (!condValue.isEmpty()) {
+            String[] values = condValue.split("@");
+            if (values.length == 2) {
+                String key = values[0].trim();
+                String value = values[1].trim();
+                if ("no".equals(key) || "none".equals(key)) {
+                    if (value.contains("(") && value.contains(")")) value = value.substring(1, value.length() - 1);
```
otbutz, line 32:
```suggestion
                    if (value.startsWith("(") && value.endsWith(")")) value = value.substring(1, value.length() - 1);
```

core/src/main/java/com/graphhopper/routing/util/parsers/MaxWeightExceptParser.java
```diff
@@ -0,0 +1,57 @@
...
+import com.graphhopper.routing.util.parsers.helpers.OSMValueExtractor;
+import com.graphhopper.storage.IntsRef;
+
+import java.util.List;
+
+import static com.graphhopper.routing.util.parsers.helpers.OSMValueExtractor.stringToTons;
+
+public class MaxWeightExceptParser implements TagParser {
+    private final EnumEncodedValue<MaxWeightExcept> mweEnc;
+    private static final List<String> HGV_RESTRICTIONS = OSMRoadAccessParser.toOSMRestrictions(TransportationMode.HGV);
```
otbutz, line 17:
```suggestion
    private static final List<String> HGV_RESTRICTIONS = OSMRoadAccessParser.toOSMRestrictions(TransportationMode.HGV).stream()
                .map(e -> e + ":conditional").collect(Collectors.toList());
```

core/src/main/java/com/graphhopper/routing/util/parsers/helpers/OSMValueExtractor.java
```diff
@@ -41,6 +41,29 @@ public static void extractTons(int edgeId, EdgeIntAccess edgeIntAccess, ReaderWa
...
+        try {
+            int index = value.indexOf("weight>"); // maxweight or weight
+            if (index < 0) {
+                index = value.indexOf("weight >");
+                if (index > 0) index += "weight >".length();
+            } else {
+                index += "weight>".length();
+            }
+            if (index > 0) {
+                int lastIndex = value.indexOf(')', index); // (value) or value
```
otbutz, line 57:
```suggestion
                int lastIndex = value.lastIndexOf(')', index); // (value) or value
```

I'm not sure if more complex statements with multiple parentheses are allowed, but using the last parenthesis is correct either way.

otbutz (changes requested):
Very nice addition! :rocket:

core/src/main/java/com/graphhopper/routing/util/parsers/MaxWeightExceptParser.java
otbutz: Exactly

Issue #2349, opened 2021-06-23
Cancel requests on the server-side when client aborts


karussell:
If client is no longer interested in the response we should cancel it on the server side. Maybe there is some jetty or dropwizard function to do this.

michaz:
I remember the following (broad strokes):

When you stream your response (i.e. write to a `StreamingResponse` rather than building a large object and returning it), you get this automatically. The `StreamingResponse` throws an exception when you write to it while the connection is closed, and you're done.

Otherwise you have to periodically check `Thread.isInterrupted()` or something in your computation code, and _possibly_ also tell the server somehow that it _should_ interrupt (not sure).

boldtrn:
AFAIK the main problem is that when a client cancels the request, the client doesn't notify the server. The client will just reject the reply from the server. So the server doesn't know about cancelled requests.

michaz:
Oh interesting! My response up there was because I remembered the Shortest Path Tree endpoint, where we do notice when the connection is closed -- but there of course we write all the time because we write nodes while we are still routing!

michaz:
(So what now, do we have to reply with some kind of progress bar in JSON? :-D

{
  "1": "Still calculating...",
  "2": "Still calculating...",

.
.

so we notice when the connection drops?

(Just kidding)

PR #2795, opened 2023-04-13
Add timeout_ms parameter for routing requests

easbar:
Using something like `timeout_ms=300` we can limit routing requests to `300ms`. When the timeout is exceeded an exception is thrown and 'connection not found' is returned from the server.


otbutz:
Loosely related to #2349

easbar:
> Thread.interrupted() isn't that helpful on its own

Ok maybe not helpful on its own, but if we checked for `Thread.interrupted()` we would definitely get an exception when we interrupt the thread, no?

> We could think about running the timeout/limit tests only every x cycles

Even without this I did not notice a measurable slow down of the algorithm, but good point doing this only for every nth visited node.

otbutz:
Do we want to introduce this also as a configurable knob in the server configuration to enforce a global cap?

easbar:
Ok, yes, we do the same for the maximum visited nodes already.

core/src/main/java/com/graphhopper/routing/Router.java
```diff
@@ -373,6 +374,11 @@ private void checkMaxVisitedNodes() {
                 throw new IllegalArgumentException("The max_visited_nodes parameter has to be below or equal to:" + routerConfig.getMaxVisitedNodes());
         }
 
+        private void checkTimeout() {
+            if (getTimeoutMillis(request.getHints()) > routerConfig.getTimeoutMillis())
```
otbutz, line 378: I would prefer to take the `Math.min()` approach instead. Lowering the timeout on the serverside would cause all requests to fail which probably isn't expected.

easbar: Not sure tbh, we could even make this configurable :)

easbar: Ok, you mean there should not be an immediate error when we request a longer timeout than what is specified on the server-side, but the server-side timeout should be used silently?

easbar: Ok, guessing and trial and error would only be needed if the server limit was not documented. But I see your point that such errors aren't very helpful. 

Actually I am wondering why a 'client' would setup a timeout in the first place, especially if it is not allowed to request a **higher** limit. When the GH server is used without a proxy only the server limit seems useful, because clients do not benefit from setting a limit at all. When the GH server is used with a proxy the best place to set the limit would be the proxy in which case the server limit is not needed. I don't really see a scenario where a server limit will be used and clients set the timeout as well?

otbutz: > When the GH server is used with a proxy the best place to set the limit would be the proxy in which case the server limit is not needed.

Is it? The gist of #2349 was that we're unable to tell if the client or proxy cancelled the request.

otbutz: For me, the client limit is more of a hint, and it is up to the server to enforce a reasonable limit.

easbar: > Is it? The gist of https://github.com/graphhopper/graphhopper/issues/2349 was that we're unable to tell if the client or proxy cancelled the request.

Yes, I mean the value of the timeout parameter would be specified by the proxy, but of course the timeout is implemented in the GH server (this PR).

easbar: > For me, the client limit is more of a hint, and it is up to the server to enforce a reasonable limit.

Under which circumstances would a client like to hint for a **shorter** timeout?

otbutz: > Under which circumstances would a client like to hint for a **shorter** timeout?

I can see the more theoretical use case of a request that needs to be completed within a certain timeframe from the perspective of the client, but I agree that this limit is more suited to server administrators who want to restrict resource usage.

"I don't care about the result if the calculation takes longer than x milliseconds" vs "I don't want client requests to take longer than x milliseconds".

> Yes, I mean the value of the timeout parameter would be specified by the proxy, but of course the timeout is implemented by the GH server (this PR).

It just seems a bit odd to introduce a parameter that's only intended for proxies or other middleware. The worst case scenario would be for a user to mistakenly use this parameter instead of a simple HTTP timeout. Things would work at first glance, until the admin decides to introduce a server-side limit via configuration.

My only intention here is to avoid a potential footgun. Using the lower of both limits would work for both cases.

easbar: > I can see the more theoretical use case of a request that needs to be completed within a certain timeframe from the perspective of the client

But then the client simply uses a timeout on the client side, no?

> but I agree that this limit is more suited to server administrators who want to restrict resource usage.

I think that's the only use-case. Nobody forces any client to wait until the server responds or times out. Any client can move on whenever it wants to.

> My only intention here is to avoid a potential footgun. Using the lower limit would work for both cases.

Same here :) To me doing something silently that is contrary to what was requested is normally a no-go. Like using a timeout of 1000ms, when a request came in requesting `timeout_ms=5000`. I'd much rather return an error in this case and handle the error where it occurred (at the place the timeout is set). I agree that if that place is the **end**-user such an error can be annoying, but like I said, I don't see a reason why an end-user should setup a timeout in the first place.

Ok, but using the minimum we can support both scenarios: Standalone servers can setup the limit on the server-side only and clients will probably never use the timeout parameter, but even if they do they cannot really do any harm.
In case a proxy is used the server limit can be set to the maximum (or left out since this is the default) and the timeout can be controlled via the request parameter.

This comes at the cost of the (to me unexpected) behavior, that the server can ignore the requested timeout and use a lower limit silently.

easbar: >  Standalone servers can setup the limit on the server-side only and clients will probably never use the timeout parameter, but even if they do they cannot really do any harm.

Hm, but if they never use it there is no real problem to thrown an exception when it is above the limit? ;)

> Things would work at first glance, until the admin decides to introduce a server-side limit via configuration.

I think my point is that end-users will never set a timeout (why would they?), so changing the value on the server-side won't cause any problems.

otbutz: I think it's fine either way. But the parameter should be properly documented to reflect that.

easbar: Ok, I do not care very much, either. I'll use the minimum for now.

PR #2796, opened 2023-04-17
bike: differentiate between bad highways

karussell:
Avoid picking dangerous `trunk` for bike:

 * https://graphhopper.com/maps/?point=50.441948%2C-4.703575&point=50.474151%2C-4.517054&profile=bike
  grafik
 * https://graphhopper.com/maps/?point=53.25632%2C-2.584999&point=53.28341%2C-2.600749&profile=bike
  grafik
 * [whole route](https://graphhopper.com/maps/?point=50.066263%2C-5.714822_Land%E2%80%99s+End%2C+TR19+7AA+Sennen%2C+Vereinigtes+K%C3%B6nigreich&point=58.643214%2C-3.08372_John+o%27+Groats%2C+KW1+4YX+John+o%27+Groats%2C+Vereinigtes+K%C3%B6nigreich&profile=bike&layer=Omniscale) contains >50km trunk roads

Can you have a look @ratrun if the approach makes sense? At some point we should probably try to move this code to a custom model too.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
+++ b/core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
@@ -17,6 +17,9 @@
  */
 package com.graphhopper.routing.util;
 
+import java.util.Collections;
+import java.util.TreeSet;
+
 /**
  * Used to store a priority value in the way flags of an edge. Used in combination with
  * PriorityWeighting
@@ -36,7 +39,13 @@ public enum PriorityCode {
     PREFER(12),
     VERY_NICE(13),
     BEST(15);
+
     private final int value;
+    public static final TreeSet<PriorityCode> VALUES = new TreeSet<>();
+
+    static {
+        Collections.addAll(VALUES, values());
+    }
 
     PriorityCode(int value) {
         this.value = value;
@@ -55,10 +64,12 @@ public enum PriorityCode {
     }
 
     public PriorityCode worse() {
-        return this == EXCLUDE ? EXCLUDE : PriorityCode.values()[ordinal() - 1];
+        PriorityCode ret = VALUES.lower(this);
+        return ret == null ? EXCLUDE : ret;
     }
 
     public PriorityCode better() {
-        return this == BEST ? BEST : PriorityCode.values()[ordinal() + 1];
+        PriorityCode ret = VALUES.higher(this);
+        return ret == null ? BEST : ret;
     }
 }
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
@@ -228,7 +228,8 @@ public abstract class BikeCommonPriorityParser implements TagParser {
         if (way.hasTag("scenic", "yes") || maxSpeed > 0 && maxSpeed <= wayTypeSpeed) {
             int lastEntryValue = weightToPrioMap.lastEntry().getValue();
             if (lastEntryValue < BEST.getValue()) {
-                PriorityCode priorityCode = Arrays.stream(PriorityCode.values()).filter(pc -> pc.getValue() == lastEntryValue).findFirst().orElse(UNCHANGED);
+                // TODO migrate weightToPrioMap to Map<Integer, PriorityCode>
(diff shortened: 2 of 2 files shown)


otbutz:
Wouldn't this also avoid roads with a dedicated cycleway?  e.g https://graphhopper.com/maps/?point=48.736881%2C12.189247&point=48.704716%2C12.197717_Jellenkofen%2C+84061+Ergoldsbach%2C+Deutschland&profile=car&layer=Cyclosm

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
```diff
@@ -53,4 +53,12 @@ public static double getFactor(int value) {
     public static double getValue(int value) {
         return getFactor(value);
     }
+
+    public PriorityCode worse() {
```
otbutz, line 57: By using a NavigableSet like a TreeSet, this would be more readable and avoid allocating an array on each call:

```suggestion
    public PriorityCode worse() {
        VALUES.floor(this);
    }
```

otbutz: Arrays are mutable. Therefore any call to `values()` has to create a new array.

https://stackoverflow.com/a/32799324

otbutz:
> > Wouldn't this also avoid roads with a dedicated cycleway? e.g https://graphhopper.com/maps/?point=48.736881%2C12.189247&point=48.704716%2C12.197717_Jellenkofen%2C+84061+Ergoldsbach%2C+Deutschland&profile=car&layer=Cyclosm
> 
> I don't think so because we put a higher priority (100d) to those dedicated cycleways e.g. [here](https://github.com/graphhopper/graphhopper/blob/bdc7a3bc9f6d358640b00fadc0cf92a7d12d46e5/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java#L153) but will ensure this via a test.
> 
> btw: The example does not have a cycleway on the secondary I think: https://www.openstreetmap.org/way/1074529412 vs https://www.openstreetmap.org/way/65414266

image


https://www.openstreetmap.org/way/25798194

otbutz:
> I would be interested in what your trigger for this change was, because until now there were no user complaints here on github about this as far as I remember.

Context: https://www.reddit.com/r/openstreetmap/comments/12p3nee/comment/jglx2lo/

tl;dr: the developer of https://cycle.travel/ was a bit unhappy about us routing bikes over motorways and communicated this with some unfortunate wording :)

ratrun:
@otbutz Thank you for the background. 

I analysed the following [example](http://localhost:8989/maps/?point=49.645693%2C-121.002232&point=49.750024%2C-121.011607&profile=racingbike&layer=OpenStreetMap) before and after the change. I used a race-bike example in Canada in order to illustrate what I meant with that the downgrade in the priority for motorways is too drastic.

Here is the route before the change: 
ScreenshotCurrent

And here the some route after the change:
ScreenshotAfterChanges

I'm sure that with a racing bike you will not choose such long unpaved `highway=path` segments and prefer the [motorway](https://www.openstreetmap.org/way/41376957) instead.

ratrun:
> But for bike we currently prefer the motorway which seems to be incorrect

I agree.

> without complaints from people with ground truth knowledge we should IMO not change this.

I cannot tell for this exact spot, but I was close and therefore I know that where bicycles are allowed on motorways there are nearly always shoulders and usually all of them are paved.

> Even on primary roads in Germany I prefer sometimes to safely get off the bike on a side road...

Side roads are of course always better as long as they are paved. But I assume that you are not riding a race bike then, where kilometers of unpaved roads get really annoying.

Would you accept a PR which prioritises motorways and trunks for the race bike profile as `AVOID_MORE` or maybe `BAD` instead of `REACH_DESTINATION`?

ratrun:
The PR [5080599](https://github.com/ratrun/biketourplanner/commit/50805997d336f9a76ca4aaa2beafce85d31e0998) implements the priority increase for the highway classes as suggested above. The routing result of [my example](https://graphhopper.com/maps/?point=49.645693%2C-121.002232&point=49.750024%2C-121.011607&profile=racingbike&layer=OpenStreetMap) uses the motorway.

otbutz:
As this has quite a big impact on bike routing, could you add this to the changelog @karussell?

PR #2797, opened 2023-04-18
Extend railway OSM Parsers

kanephamphu:
In the early OSM reading process, 7.0.0 GrapHopper is filtering almost railway values except for the platform value. So, I would like to extend the accepted railway list.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/OSMParsers.java
+++ b/core/src/main/java/com/graphhopper/routing/util/OSMParsers.java
@@ -36,6 +36,7 @@ public class OSMParsers {
     private final List<TagParser> wayTagParsers;
     private final List<RelationTagParser> relationTagParsers;
     private final List<RestrictionTagParser> restrictionTagParsers;
+    private final List<String> acceptedRailways = new ArrayList<>();
     private final EncodedValue.InitializerConfig relConfig = new EncodedValue.InitializerConfig();
 
     public OSMParsers() {
@@ -48,6 +49,7 @@ public class OSMParsers {
         this.wayTagParsers = wayTagParsers;
         this.relationTagParsers = relationTagParsers;
         this.restrictionTagParsers = restrictionTagParsers;
+        this.initAcceptedRailways();
     }
 
     public OSMParsers addIgnoredHighway(String highway) {
@@ -81,7 +83,7 @@ public class OSMParsers {
             return true;
         else if ("pier".equals(way.getTag("man_made")))
             return true;
-        else if ("platform".equals(way.getTag("railway")))
+        else if (acceptedRailways.contains(way.getTag("railway")))
             return true;
         else
             return false;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/graphhopper/routing/util/OSMParsers.java
```diff
@@ -123,4 +125,14 @@ public List<RelationTagParser> getRelationTagParsers() {
...
+
+    private void initAcceptedRailways() {
+        this.acceptedRailways.add("platform");
+        this.acceptedRailways.add("narrow_gauge");
+        this.acceptedRailways.add("tram");
+        this.acceptedRailways.add("subway");
+        this.acceptedRailways.add("light_rail");
+        this.acceptedRailways.add("monorail");
+        this.acceptedRailways.add("funicular");
+    }
```
easbar, line 137: We cannot add this, because it would mean that by default we accept all these railways even though they are only needed for the very special use-case of railway routing, which in turn requires a lot more adjustments (I guess).

Like I said in the forum, I don't even think we should make this configurable, because changes in Java code are necessary either way.

I think the only thing that would be acceptable is a minor restructuring of the code to make it easier to adjust to accept railways (but I did not try this myself yet, so honestly I am not sure if we like to go this direction at all).

kanephamphu: To be honest, I don't understand why we have to force the OSMParsers values. I'm stuck between breaking the structure to allow user self customize the OSMParsers and some temporary solution. 

To quickly fix this issue(I mean my railway issue). Create the option to turn on and off allowing all railway values. But, we still face RoadsAccessParser broke if turn on this option. Easy to handle for only RoadAccessparser, but other people who use graphHopper are different. :D . Really thanks for your advice, I have to do some report about the upgrading version PBI result. Ha ha.

easbar: > To be honest, I don't understand why we have to force the OSMParsers values. 

You mean you don't know why we even need to setup a list of ways we accept? The reason is that in OSM ways are not directly associated with their coordinates, but rather they only contain references to OSM nodes which in turn contain the coordinates. When we read the OSM file we need to do this association ourselves and the more ways we accept the more coordinates we need to keep in memory. Also we create graph edges for every accepted way, so the more ways we accept the bigger the graph will be.

> I'm stuck between breaking the structure to allow user self customize the OSMParsers and some temporary solution.

For advanced customizations of GraphHopper we recommend maintaining a separate fork that allows you to adjust it the way you need for your application. When the code can easily be adjusted in a way that less such changes are necessary for interesting use cases we accept a PR, but otherwise we don't like to add such changes, because then we need to do the maintenance in the future.

> But, we still face RoadsAccessParser broke if turn on this option. Easy to handle for only RoadAccessparser, but other people who use graphHopper are different. :D . Really thanks for your advice, I have to do some report about the upgrading version PBI result. Ha ha.

Ok, I cannot tell from your description which problems you are running into here, but if you can report your experience with the upgrade it will certainly be useful for us when developing GraphHopper.

PR #2800, opened 2023-04-19
cleanup of EnumEncodedValue 

karussell:
and avoid calling Enum::values()

I'm a bit unsure why we needed this in the first place. Probably to better match tag names, but we use toUpperCase (or before equalsIgnoreCase) there anyway.


otbutz:
Hm...might still be a breaking change. The countrycodes are currently all uppercase.

PR #2805, opened 2023-04-24
Connect tower nodes of highways tagged as area=yes

easbar:
This PR shall add support for some simple cases of routing through pedestrian areas. 

In OSM there are currently around [390k ways tagged as area=yes that carry a highway tag](https://taginfo.openstreetmap.org/keys/area?filter=ways#combinations). There are another around [19k relations tagged as area=yes that carry a highway tag](https://taginfo.openstreetmap.org/keys/area?filter=relations#combinations).

The textbook example for such an area we'd like to route through would be something like Glücksburger Platz in Hamburg, Germany:

[image](https://www.openstreetmap.org/way/30852872)

This is a closed-ring way tagged as `area=yes;highway=pedestrian` that shares some nodes with adjacent roads/footways. As long as we treat it as a normal way tagged as `highway=pedestrian`, ignoring the `area=yes` tag, the route will simply go along the edge of the area, like this:

[image
](https://graphhopper.com/maps/?point=53.569292%2C9.949045&point=53.56993%2C9.949429&profile=foot&layer=OpenStreetMap)

... which is of course not ideal. Note that the dashed line crossing the area is an administration boundary not a footway in this example.

Because GraphHopper is not the only router that doesn't offer support for `area=yes`, mappers often add artificial ways, even though some consider mapping for routers bad practice. Take a look at [Gerhart-Hauptmann-Platz in Hamburg](https://www.openstreetmap.org/way/43176536):

|Satellite|OSM|
|-|-|
|image|image|

In this case GraphHopper uses these extra ways to route across the plaza of course:

[image](https://graphhopper.com/maps/?point=53.550811%2C9.999&point=53.552088%2C9.99927&profile=foot&layer=OpenStreetMap)

But obviously for many pedestrian areas such ways are missing and as far as I know it is not even the consensus among mappers that they should be mapped, or some might be even against mapping them.

So the question here is what we, as a routing engine, can do to support routing through such areas. 

The examples I showed above are very simple, because (a) the corresponding areas are convex, i.e. all points of the area boundary can be connected with a straight line without leaving the area (at least this is true for all points on the boundary that connect the area with adjacent roads), and (b) there are only a few such points that are clearly separated from each other. In this case we can simply connect the points to add extra edges to the routing graph and the area will become 'routeable'. This is done in this PR. In case of the first example (Glücksburger Platz) it looks like this:

image

The grey lines represent routing graph edges, while the red arrows point to the six artificial edges we added to connect the four points of the area that are connected to adjacent roads/footways.

This results in a much more realistic route that no longer goes along the edge of the area:

image

For the second example (Gerhart-Hauptmann-Platz) the situation is already more complicated:

image

Here we added edges to connect the entry nodes of the area again, but since OSM already contained some extra ways that mappers added to make the area routeable the routing network becomes rather messy. Due to the way we snap coordinates to the route network this can lead to strange routes like this:

image

To fix this we would have to create extra nodes at points where these edges intersect, recognize areas for which there are such extra ways already and either ignore them or do not add more edges, or change our snapping approach. Unfortunately none of these solutions would be easy or even feasible for us.

Another problem is that many areas aren't convex, like [Harald-Stender Platz](https://www.openstreetmap.org/way/228258175):

[image](https://graphhopper.com/maps/?point=53.55391%2C9.965395&point=53.553412%2C9.967905&profile=foot&layer=OpenStreetMap)

In such cases the present simple approach might add too few extra edges. I think it would be feasible to apply some more sophisticated logic and create edges that follow shortest paths without leaving the area, even when the area is not convex. However, the bigger issue seems to be that often too many edges would be added and lead to a similar problem as the one mentioned above.

Even the simple approach might add too many edges. Take for example this pedestrian area, which does not really represent a 'plaza', but rather some kind of pedestrian 'lane':

[image](https://www.openstreetmap.org/way/44778749)

In this case the routing graph will look like this (with this PR):

image

and quite obviously this dense net of edges leads to strange routes when we snap inside the area (for the same reasons mentioned above):

[image](http://localhost:3000/?point=53.550936%2C10.000339&point=53.551044%2C10.00108&profile=foot&layer=OpenStreetMap)

The situation becomes even worse when there are not only many tower nodes along the area border, but when areas share a common border. In this case basically every node becomes a tower node (because it belongs to two or more ways).

For example here there are many little pedestrian areas with complicated shapes that share common ways:

[image](https://www.openstreetmap.org/way/161916852#map=19/53.54230/9.99280)

When we apply the simple logic in this PR we get a super dense routing graph with very many crossing edges:

image

In summary there seem to be:

* areas that could easily be made routeable using the simple approach represented here
* areas where adding more edges is unnecessary and even can result in too many crossed edges because mappers already added extra ways to make the area routeable
* areas that are not convex where a more sophisticated approach would be needed to connect the different roads adjacent to the area
* areas where even with the simple/conservative approach here we would add too many edges

The good news is that with this PR the total edge count only increases by around 0.5% (tested for Germany). As it stands it can lead to strange routes like the ones shown above, though. Overall, I am not really convinced this is a real improvement yet. 

To reduce the number of extra edges and avoid the strange routes I mentioned we would probably need a (much) more sophisticated approach that only adds a small number of 'useful' edges and/or another snapping algorithm?


easbar:
I think an important observation is that in OSM `area=yes&highway=pedestrian/footway` can both mean we are dealing with some kind of 'pedestrian lane/area' or a 'plaza':

image

In the screenshot we see area=yes ways highlighted in blue. Some encircle blocks of houses, while others (the one in the top right) represent a 'plaza'. We probably only want to add edges for the latter kind.

easbar:
> Do you still have the links for those areas on GH Maps?

The screenshots should be clickable/provide links to the areas.

> Would this really happen that often in real world?

I think it happens whenever we start or end a route within such an area (or both). We simply snap to the next edge and this might lead us into the wrong direction (make a detour that seems not plausible).

> Maybe this is not really properly mapped and instead a single enclosing area should be used?

It looks like they used separate areas to map different surfaces or something?

otbutz:
Could we try to compute some kind of minimum network of support points for an area? At least for convex area polygons, routing through the centre of mass seems like a good place to start.

grafik

otbutz:
https://stackoverflow.com/questions/62645727/how-can-i-get-center-line-of-a-complex-shaped-polygon#comment110828520_62651638

Vectorial1024:
Hi there! I have been redirected to this issue.

So far there are these cases:
- convex pedestrian areas
- concave pedestrian areas

Large area or not would not be a problem; hypothetically, we would just solve them the same way regardless of the size of the area.

The main problem is to enable routing through the areas (i.e., outside the area -> cross the area -> outside the area) . It seems to me routing within these areas is less of a concern (i.e. inside the area -> inside the area) .

I have no opinion about the exact methods on how to generate the extra edges (e.g. every node to every other node? find minimal support nodes? or other methods?)

However, I was thinking whether it is possible to tell GraphHopper to not snap to those generated edges. If that's possible, then we have the following benefits:
- foot profile can still go through pedestrian areas
- no more "sharp angle" paths inside pedestrian areas as found in this PR
- preserves explicitly-mapped foot paths, in case OSM mappers believe those are somehow important to the map

easbar:
I think it would be useful if we had a whole list of area examples where routing along artificial edges would be important (not just nice to have). Maybe like 10-100 examples. There is very likely some way to improve on the current situation, but honestly I am not convinced it is really worth it, yet.

1ec5:
> I think an important observation is that in OSM area=yes&highway=pedestrian/footway can both mean we are dealing with some kind of 'pedestrian lane/area' or a 'plaza':

For what it’s worth, [this straw poll](https://community.openstreetmap.org/t/clarification-of-footway-area-mapping/130974/44) back in September found a very clear consensus (among 30 participants) for a distinction between `area:highway=*` and `highway=*` `area=yes` that is based on whether travel within the area is expected to be mostly linear or non-linear, respectively. In other words, this example should be retagged as `area:highway=footway` `footway=sidewalk`, and `highway=footway` `footway=sidewalk` ways should be drawn through the remaining sides of these blocks. Not everyone agrees, but this is one of the more lopsided polls I’ve ever seen in the forum’s global tagging category.

The discussion also largely accepted the idea that manually mapped virtual ways are only for pathological cases, or besides that only a temporary crutch while we wait for the sort of improvement you’re making here.

Vectorial1024:
I just thought of something.

It seems a big outstanding problem is that pedestrian areas might already contain explicitly-mapped paths within them. Fabricating paths while these explicit paths exist would inflate the graph too much.

What if, we just detect whether there are explicitly-mapped paths inside the pedestrian areas. If these paths exist, then we just skip processing this pedestrian area? If the paths are explicitly mapped, then it might be the (hopefully local) OSM mapper knows something about the pedestrian area which caused them to draw out the explicit paths instead of deferring it to the router (e.g. GraphHopper) to generate/fabricate "default paths".

PR #2807, opened 2023-04-26
Add ways from highway relations to road network

easbar:
Currently [there are around 53k relations tagged as `highway=*`](https://taginfo.openstreetmap.org/keys/highway?filter=relations#values). Many (~18k, c.f. #2805) of them carry the `area=yes` tag, but even without considering that they might represent (routeable) areas, it seems useful to allow routing along the corresponding member ways.

In this PR I added a third (0th) pass for reading the OSM file to create a list of these relations, which is then used to create the corresponding edges in pass1/2. Relation member ways  inherit the tags from their parent relation, but they are only considered if they would not have been accepted based on their (way) tags already.

This screenshot shows highway relations in Paris with (blue) and without (red) an `area=yes` tag:

image

where highway relations are often used to map sidewalks:

image

Adding them allows routing along the sidewalk of course:

|before|after|
|-|-|
|image|image|

An important gotcha here is that we should exclude the 'inner' polygons of the highway relations, because sometimes they are connected to the outer polygons and/or the remaining road network only at one or a few points. This means that when we snap onto the inner polygon we might calculate a route with a long detour that goes to these connecting points and back like this:

image

This problem is not completely new, though and exists already in master (for normal highways with a similar shape), like here:

[image](https://graphhopper.com/maps/?point=48.815958%2C2.320979&point=48.815845%2C2.32083&profile=foot&layer=Omniscale)

To fix this we would probably either need to add artificial extra edges as in #2805 and/or extend our greedy snapping algorithm.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/OSMParsers.java
+++ b/core/src/main/java/com/graphhopper/routing/util/OSMParsers.java
@@ -21,8 +21,8 @@ package com.graphhopper.routing.util;
 import com.graphhopper.reader.ReaderRelation;
 import com.graphhopper.reader.ReaderWay;
 import com.graphhopper.reader.osm.RestrictionTagParser;
-import com.graphhopper.routing.ev.EncodedValue;
 import com.graphhopper.routing.ev.EdgeIntAccess;
+import com.graphhopper.routing.ev.EncodedValue;
 import com.graphhopper.routing.util.parsers.RelationTagParser;
 import com.graphhopper.routing.util.parsers.TagParser;
 import com.graphhopper.storage.IntsRef;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/graphhopper/routing/util/OSMParsers.java
```diff
@@ -87,6 +88,17 @@ else if ("platform".equals(way.getTag("railway")))
             return false;
     }
 
+    public Predicate<ReaderRelation> getRelationFilterForWays() {
+        if (ignoredHighways.contains("pedestrian") && ignoredHighways.contains("footway"))
+            // if are both ignored we can skip pass0 altogether
+            return null;
```
easbar, line 94: The 0th pass is skipped entirely if `pedestrian` and `footway` are both added to the list of ignored highways, because >80% of these highway relations use one of these two highway values. Then again, there are also around [10k relations tagged as `service,tertiary,platform,bus_stop...`](https://taginfo.openstreetmap.org/keys/highway?filter=relations#values). I'm not sure if we really need them for motorized-only routing (i.e. when pedestrian+footway are excluded). In case they are not excluded we would accept all highway relations anyway with this PR.

PR #2810, opened 2023-04-27
default speed limits

karussell:
This PR integrates the "legal speed limit" library from @westnordost via a new MaxSpeedCalculator (`max_speed_calculator.enabled: true`). See #2392.

The library increases our jar file size a bit from 47MB to 49MB due to the kotlin std lib dependency. (The added json file is probably nearly neglectable I think)

We store a bit if max_speed was picked from OSM or the maxspeed library ([discussion](https://github.com/westnordost/osm-legal-default-speeds/pull/12#issuecomment-1571667784)) -> stored in `max_speed_estimated`.

The only real drawback of this library is to get the kotlin dependency right:

 - client-hc uses okhttp:jar:4.9.3 uses okio:jar:2.8.0 which uses kotlin-stdlib:jar:1.4.10 
 - vs speed limit lib uses kotlin 1.7.10. Problem is that even [latest okhttp 4.10.0 uses only kotlin 1.6.20 and relies on okio that uses yet another kotlin version](https://square.github.io/okhttp/changelogs/changelog_4x/).

external links:

 * library: https://github.com/westnordost/osm-legal-default-speeds
 * testing site: https://westnordost.github.io/osm-legal-default-speeds/
 * wiki: https://wiki.openstreetmap.org/wiki/Default_speed_limits

TODO / Limitations:

 * https://github.com/westnordost/osm-legal-default-speeds/discussions/17


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -897,6 +897,21 @@ protected void postImport() {
             BooleanEncodedValue roadClassLinkEnc = encodingManager.getBooleanEncodedValue(RoadClassLink.KEY);
             UrbanDensityCalculator.calcUrbanDensity(baseGraph, urbanDensityEnc, roadClassEnc, roadClassLinkEnc, residentialAreaRadius, residentialAreaSensitivity, cityAreaRadius, cityAreaSensitivity, urbanDensityCalculationThreads);
         }
+
+        if (encodingManager.hasEncodedValue(Country.KEY)
+                && encodingManager.hasEncodedValue(UrbanDensity.KEY)
+                && encodingManager.hasEncodedValue(MaxSpeed.KEY)) {
```
otbutz, line 903: This check would also be more specific with a dedicated EV. Chances are that the user will inadvertently activate the new maxspeed logic.

core/src/main/java/com/graphhopper/routing/util/MaxSpeedCalculator.java
```diff
@@ -0,0 +1,153 @@
...
+        // You can replace the result of any number of placeholders in a tag filter here (e.g. for name = "urban"),
+        // for example if you have another data source to acquire whether a road is in a built-up area or not.
+        // For those you do not want to replace, simply pass on the result of evaluate as result
+        Function2<String, Function0<Boolean>, Boolean> replacerFunction = (name, eval) -> eval.invoke();
+        AllEdgesIterator iter = graph.getAllEdges();
+        while (iter.next()) {
+            // ISO 3166-1 alpha-2 code optionally concatenated with a ISO 3166-2 code, e.g. "DE", "US" or "BE-VLG"
+            String countryCode = iter.get(countryEnumEncodedValue).getAlpha2();
+            Map<String, String> tags = new HashMap<>();
+            tags.put("rural", iter.get(urbanDensityEnc) == UrbanDensity.RURAL ? "yes" : "no");
```
westnordost, line 49: You shouldn't need to do this. This is what the replacerFunction is for:

See https://github.com/westnordost/osm-legal-default-speeds#replacing-placeholders

So, something like...

```
Function2<String, Function0<Boolean>, Boolean> replacerFunction = (name, eval) -> {
    if ("urban".equals(name)) {
        return iter.get(urbanDensityEnc) == UrbanDensity.URBAN;
    }
    if ("rural".equals(name)) {
        return iter.get(urbanDensityEnc) == UrbanDensity.RURAL;
    }
    return eval.invoke();
}
```

(Or, first evaluate the filter normally and if it returns false, then fall back to your data source)

Your solution works too, though. But only because `rural=yes/no` is used in the source data (i.e. in the table in the wiki).

core/src/main/java/com/graphhopper/routing/util/MaxSpeedCalculator.java
```diff
@@ -0,0 +1,153 @@
...
+            tags.put("highway", iter.get(roadClassEnc).toString() + (iter.get(roadClassLinkEnc) ? "_link" : ""));
```
westnordost, line 50: You shouldn't need to do that (if I understand that piece of code correctly):

See e.g.
https://westnordost.github.io/osm-legal-default-speeds/#tags=highway%3Dmotorway_link&cc=DE

westnordost: What I mean to say is that the filters already include `<something>_link`. E.g. Search for `motorway` in this table:

https://wiki.openstreetmap.org/wiki/Default_speed_limits#Road_types_to_tag_filters

The filter is `highway~motorway|motorway_link`

If for any country, different speed limits exist for motorway ramps / links than for motorways, it would be noted in the table. If e.g. in South Sudan, motorways had a speed limit of 100 and motorway ramps a speed limit of 80, the table would look like this:

<table>
<tr><th rowspan="2">South Sudan</th><td>motorway</td><td>100</td></tr>
<tr><td>motorway link</td><td>80</td></tr>
</table>

and in the road types table, there would be an entry like

<table>
<tr><td>motorway link</td><td><code>highway=motorway_link</code></td></tr>
</table>

westnordost:
Note that the library generally doesn't overwrite maxspeeds. But it can complement them Examples:

https://westnordost.github.io/osm-legal-default-speeds/#tags=maxspeed%3D100&cc=DE

Note how the result does not include the `maxspeed` itself, because it was already given. Only those that are added.
Notice how it changes when also `maxspeed:bus=60` is specified:

https://westnordost.github.io/osm-legal-default-speeds/#tags=maxspeed%3D100%0Amaxspeed%3Abus%3D60&cc=DE

core/src/main/java/com/graphhopper/routing/util/MaxSpeedCalculator.java
```diff
@@ -0,0 +1,153 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class MaxSpeedCalculator {
```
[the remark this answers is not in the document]
otbutz: I see. But we could try to generalize them. e.g introducing a TagPostProcessor interface?

core/src/main/java/com/graphhopper/routing/util/countryrules/CountryRule.java
```diff
@@ -27,10 +27,6 @@
  * GraphHopper uses country rules to adjust the routing behavior based on the country an edge is located in
  */
 public interface CountryRule {
-    default double getMaxSpeed(ReaderWay readerWay, TransportationMode transportationMode, double currentMaxSpeed) {
-        return currentMaxSpeed;
-    }
-
```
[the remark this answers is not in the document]
otbutz: I would support dropping this as the library will do a better job :slightly_smiling_face:

core/src/main/java/com/graphhopper/routing/util/MaxSpeedCalculator.java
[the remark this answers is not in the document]
easbar: We could use the library (with all tags) in the normal MaxSpeed tag parser and calculate **three** maxspeed values: one for urban, one for residential and one for rural urban density. We'd need to store these somewhere outside the graph and then we could set the final value for the maxspeed encoded value after the urban density calculation (once we know the urban density).

westnordost:
Perhaps you want to use the parser written in python to generate the default speed limits json yourself as a (pre-)build task so that you do not have to manually download that file from the default speed limits repo everytime I feel like updating it. I do update it only very irregularly from that wiki page as it is used only for the demo page anyway.

bt90:
If the file is in a repository, we should rather intrgrate it as a submodule.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ### 8.0 [not yet released]
 
+- Country rules no longer contain maxspeed handling, enable a much better alternative via `max_speed_calculator.enabled: true`. On the client side use `max_seed_estimated` to determine if max_speed is from OSM or an estimation. See #2810
```
westnordost, line 3: > `max_seed_estimated`

whoops, typo

PR #2812, opened 2023-04-30
Landuse buffer

ratrun:
Here I implemented a 10 meters search buffer and fixed a typo.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ev/Landuse.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/Landuse.java
@@ -23,7 +23,7 @@ import com.graphhopper.util.Helper;
 public enum Landuse {
     OTHER("other"), FARMLAND("farmland"), RESIDENTIAL("residential"),
     GRASS("grass"), FOREST("forest"), MEADOW("meadow"), ORCHARD("orchard"), FARMYARD("farmyard"),
-    INDUSTRIAL("industrial"), VINEYARD("vineyard"), CEMETRY("cemetry"), COMMERCIAL("commercial"), ALLOTMENTS("allotments"),
+    INDUSTRIAL("industrial"), VINEYARD("vineyard"), CEMETERY("cemetery"), COMMERCIAL("commercial"), ALLOTMENTS("allotments"),
     RETAIL("retail"), BASIN("basin"), RESERVOIR("reservoir"), CONSTRUCTION("construction"), QUARRY("quarry");
 
     public static final String KEY = "landuse";
--- a/core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
+++ b/core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
@@ -18,6 +18,7 @@
 
 package com.graphhopper.routing.util;
 
+import com.graphhopper.util.DistanceCalcEarth;
 import org.locationtech.jts.geom.*;
 import org.locationtech.jts.geom.prep.PreparedGeometry;
 import org.locationtech.jts.geom.prep.PreparedGeometryFactory;
@@ -56,18 +57,28 @@ public class AreaIndex<T extends AreaIndex.Area> {
     }
 
     private void addBorder(PreparedGeometryFactory pgf, T area, Polygon border) {
-        // todonow: do we really need the prepared geometry for osm areas?
         IndexedCustomArea<T> indexedCustomArea = new IndexedCustomArea<>(area, pgf.create(border));
         index.insert(border.getEnvelopeInternal(), indexedCustomArea);
     }
 
     public List<T> query(double lat, double lon) {
         Envelope searchEnv = new Envelope(lon, lon, lat, lat);
+        double bufferDistance = 10; // meters
+        double rLon = (bufferDistance * 360.0 / DistanceCalcEarth.DIST_EARTH.calcCircumference(lat));
+        double rLat = bufferDistance / DistanceCalcEarth.METERS_PER_DEGREE;
+        searchEnv.expandBy(rLon, rLat);
         @SuppressWarnings("unchecked")
         List<IndexedCustomArea<T>> result = index.query(searchEnv);
-        Point point = gf.createPoint(new Coordinate(lon, lat));
+        Coordinate[] coordinates = {
+            new Coordinate(lon-rLon,lat-rLat),
+            new Coordinate(lon-rLon,lat+rLat),
+            new Coordinate(lon+rLon,lat+rLat),
+            new Coordinate(lon+rLon,lat-rLat),
+            new Coordinate(lon-rLon,lat-rLat)
+        };
+        Polygon poly = gf.createPolygon(coordinates);
         return result.stream()
-                .filter(c -> c.intersects(point))
+                .filter(c -> c.intersects(poly))
                 .map(c -> c.area)
                 .collect(Collectors.toList());
     }
@@ -81,9 +92,10 @@ public class AreaIndex<T extends AreaIndex.Area> {
             this.preparedGeometry = preparedGeometry;
         }
 
-        boolean intersects(Point point) {
-            return preparedGeometry.intersects(point);
+        boolean intersects(Polygon poly) {
+            return preparedGeometry.intersects(poly);
         }
+
     }
 
 }


core/src/main/java/com/graphhopper/routing/ev/Landuse.java
```diff
@@ -23,7 +23,7 @@
 public enum Landuse {
     OTHER("other"), FARMLAND("farmland"), RESIDENTIAL("residential"),
     GRASS("grass"), FOREST("forest"), MEADOW("meadow"), ORCHARD("orchard"), FARMYARD("farmyard"),
-    INDUSTRIAL("industrial"), VINEYARD("vineyard"), CEMETRY("cemetry"), COMMERCIAL("commercial"), ALLOTMENTS("allotments"),
+    INDUSTRIAL("industrial"), VINEYARD("vineyard"), CEMETERY("cemetery"), COMMERCIAL("commercial"), ALLOTMENTS("allotments"),
```
easbar, line 26: Nice catch! I fixed this directly in the main branch

core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
```diff
@@ -56,18 +57,28 @@ public AreaIndex(List<T> areas) {
     }
 
     private void addBorder(PreparedGeometryFactory pgf, T area, Polygon border) {
-        // todonow: do we really need the prepared geometry for osm areas?
         IndexedCustomArea<T> indexedCustomArea = new IndexedCustomArea<>(area, pgf.create(border));
         index.insert(border.getEnvelopeInternal(), indexedCustomArea);
```
easbar, line 61: Did you try anything in this direction?

easbar:
I think I agree we should do something along this line, but did you try it and check the results? How does it affect the quality of the landuse encoded value and the performance of the import? Did you look for some examples where the landuse association is improved and maybe some where this leads to problems? Also I'm not sure if we should change it here, because it would affect the other usages of the area index as well.

core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
ratrun: No, I didn't. I'm also not sure what you mean by that comment.

ratrun:
> but did you try it and check the results? How does it affect the quality of the landuse encoded value and the performance of the import? 

> Did you look for some examples where the landuse association is improved and maybe some where this leads to problems?

So far I tested with the display feature of the graph in my home area only. From what I could see the problematic cases all work as they should.

> Also I'm not sure if we should change it here, because it would affect the other usages of the area index as well.
OK, then we need another method.

ratrun:
I ran a benchmark with the austria latest OSM file and compared your implementation against mine:
Your version: `[main] INFO  org.eclipse.jetty.server.Server - Started @330185ms`
My version: `[main] INFO  org.eclipse.jetty.server.Server - Started @526005ms`

I was able to reduce `-Xmx5000m -Xms5000m` into `-Xmx4500m -Xms4500m`, with `4000` I get an OOM exception.

So my version is unfortunately significantly slower 😞 . Anyway, I still think that it is an improvement which is worth the increased time. Do you have an idea how this can be optimised?

otbutz:
@ratrun `PreparedPolygon.intersects(Geometry)` has an optimization if the polygon itself is a rectangle. So you would need to create a prepared version of your bbox polygon and switch the order in the intersection check method.

You could also use `DistanceCalcEarth.createBBox()`.

@easbar i fear that these calculations might over- or underflow near the lat/lon extrema. There are no bound checks or logic to shift something like `-180.0004` to `0.0004`

https://github.com/graphhopper/graphhopper/blob/948294f3ed3304cfaf05a25627f9278b7cf095a6/core/src/main/java/com/graphhopper/util/DistanceCalcEarth.java#L105-L117

otbutz:
It might also be worth to apply the `TopologyPreservingSimplifier` to the landuse polygons.

core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
```diff
@@ -56,18 +57,28 @@ public AreaIndex(List<T> areas) {
...
         List<IndexedCustomArea<T>> result = index.query(searchEnv);
-        Point point = gf.createPoint(new Coordinate(lon, lat));
+        Coordinate[] coordinates = {
+            new Coordinate(lon-rLon,lat-rLat),
+            new Coordinate(lon-rLon,lat+rLat),
+            new Coordinate(lon+rLon,lat+rLat),
+            new Coordinate(lon+rLon,lat-rLat),
+            new Coordinate(lon-rLon,lat-rLat)
+        };
+        Polygon poly = gf.createPolygon(coordinates);
```
otbutz, line 79:
```suggestion
        Polygon poly = (Polygon) gf.toGeometry(searchEnv);
```

core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
ratrun: Sorry, I didn't want to delete your comment. I restored it now.

easbar: Ok, I tried this in the meantime. The SRTRee only allows querying for rectangular shapes, so use it for the envelopes of the areas. But once we query the index and get the envelopes we need a more precise intersection check and this would be too slow if we did not create a prepared geometry for each area (I tested this and it was about 10x slower). So we need to prepare the geometries beforehand, even though it does not really matter what we put into the SRTree

easbar:
> @ratrun PreparedPolygon.intersects(Geometry) has an optimization if the polygon itself is a rectangle. So you would need to create a prepared version of your bbox polygon and switch the order in the intersection check method.

But then we would have to create a prepared rectangle for every query (=for every edge), don't we?. As we only do a few (often only one) lookup for each rectangle preparation this could be even slower? Or maybe I did not understand what you mean? Also right now we first select the 'center' point for each edge and then use it as the center of a rectangle we query with. I'd rather use the entire edge and draw a buffer around it. And actually I'd rather buffer the landuse polygons, and then query with the edge, because we can easily parallelize the landuse buffering. I'm working on this currently.

> @easbar i fear that these calculations might over- or underflow near the lat/lon extrema. There are no bound checks or logic to shift something like -180.0004 to 179.9996

I'm also not  sure if we need this, since we use the JTS calculations on the unprojected coordinates anyway. Of course this is not accurate, but then we can just as well skip these calculations. And like I said I do not see much value in first converting an edge into a single point and then extending it to become a rectangle.

> It might also be worth to apply the TopologyPreservingSimplifier to the landuse polygons.

Good idea, but when I tried this unfortunately the queries did not become much faster while running the simplification took a significant amount of time.

ratrun:
> PreparedPolygon.intersects(Geometry) has an optimization if the polygon itself is a rectangle. So you would need to create a prepared version of your bbox polygon and switch the order in the intersection check method.

If I understood it correctly I gave this a try in commit [a4d34e4](https://github.com/graphhopper/graphhopper/pull/2812/commits/a4d34e453911a6e9b9d11aa7a43cc0695cd8c4f0). The result of my benchmark was `[main] INFO  org.eclipse.jetty.server.Server - Started @645787ms`, so it was even more slow.

>  I'd rather use the entire edge and draw a buffer around it. And actually I'd rather buffer the landuse polygons, and then query with the edge, because we can easily parallelize the landuse buffering.

I`m courious about this outcome here. I currently do not believe that such an approach could be faster, but we'll see.

otbutz:
> But then we would have to create a prepared rectangle for every query (=for every edge), don't we?. As we only do a few (often only one) lookup for each rectangle preparation this could be even slower?

Good point. Buffering the landuse polygons is the best way to scale this.

easbar:
> So my version is unfortunately significantly slower. Anyway, I still think that it is an improvement which is worth the increased time. 

Can we take a step back and think about what we want to use landuse for? @ratrun which of the landuse values do you think are even relevant for routing? According to [taginfo](https://taginfo.openstreetmap.org/tags/landuse) the most relevant values are:

farmland 21.92%
residential 21.10%
grass 13.94%
forest 12.98%
meadow 10.14%
orchard 3.26%
farmyard 3.08%
industrial 2.65%
vineyard 1.60%
cemetery 1.20%
commercial 1.03%
allotments 0.84%
retail 0.76%

otbutz:
I would think so, yes. The heuristic approach is based on the density of the graph. A [linear settlement](https://en.wikipedia.org/wiki/Linear_settlement) with no driveways might be enough to cause a false-negative.

otbutz:
If a route avoids residential landuse, wouldn't it be considered greenery?

otbutz:
I would combine them to be honest. Residential is a bit misleading because the classification should rather be called "built-up" which includes residential, industrial, etc

easbar:
> But I expect problems with landuse=grass, 

Why is `landuse=grass` not 'greenery'? Do you just want to stay out of populated areas (villages, cities and such)? Did you try urban_density to achieve this? Or do you want to distinguish also between different landuse values outside of populated areas (like staying in the forest rather than on farmland or something)?

core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
```diff
@@ -42,48 +45,54 @@ default Polygon getBorder() {
...
-        IndexedCustomArea<T> indexedCustomArea = new IndexedCustomArea<>(area, pgf.create(border));
-        index.insert(border.getEnvelopeInternal(), indexedCustomArea);
+    private void addBorder( T area, Polygon border) {
+        IndexedArea<T> indexedOsmArea = new IndexedArea<>(area, border);
+        index.insert(border.getEnvelopeInternal(), indexedOsmArea);
     }
 
     public List<T> query(double lat, double lon) {
-        Envelope searchEnv = new Envelope(lon, lon, lat, lat);
+        BBox bbox = dc.createBBox(lon, lat, 10);
```
easbar, line 72: This should be `lat, lon, 10` not `lon, lat, 10`, no?

easbar:
> Or this [one](https://graphhopper.com/maps/?point=48.14101%2C16.129813&point=48.339426%2C16.390738&profile=bike&layer=Omniscale) from my area. With greenry routing I hope that the city of Vienna can be avoided.

That should work fairly well using urban_density

easbar:
> A [linear settlement](https://en.wikipedia.org/wiki/Linear_settlement) with no driveways might be enough to cause a false-negative.

Yes, this is a case where urban_density might fail to identify a settlement. The landuse=residential tag would work if it encircles the whole settlement as it is often done for small villages.

core/src/main/java/com/graphhopper/routing/util/AreaIndex.java
ratrun: Thanks for detecting this bug! I ran a new benchmark with the corrected order of the parameters and got following result: `[main] INFO  org.eclipse.jetty.server.Server - Started @596397ms`, which is just a bit slower.

ratrun:
> Why is landuse=grass not 'greenery'?

According to the [wiki](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dgrass) it is to be used for small areas. From what I have seen so far it is also used most often in urban areas. There people seem to map every small green spot.

ratrun:
> That should work fairly well using urban_density

You are probably right. But with the landuse encoded value it would also be possible to prefer `forest` with a higher priority compared to the other `greenery` values. This would be useful during the hot summer periods.

easbar:
> But with the landuse encoded value it would also be possible to prefer forest with a higher priority compared to the other greenery values

Ok so besides urban_density we would only need `landuse=forest`?

ratrun:
> Ok so besides urban_density we would only need landuse=forest?

I think that this would be sufficient, but other people might have other needs.

otbutz:
> According to the [wiki](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dgrass) it is to be used for small areas. From what I have seen so far it is also used most often in urban areas. There people seem to map every small green spot.

That would be an argument for me to ignore it.

PR #2814, opened 2023-04-30
B-tree: Make space for values configurable

karussell:
The pillar node count is close to 2^31 and we would have to either switch to long values (where memory increase would be significant) or we would have to move the split of tower vs pillar nodes from 0 to -500million (nice idea from @easbar).

But having this possibility to increase (or decrease) the space for the values makes things easier and might be interesting for other use cases.

One question is if the introduction of the bytes array slows down the import speed for the planet -> no.

When we set the bytes array size to 5 by default the memory usage for the map is ~8% higher, but this means only a ~2% increase of the peak memory used for the import, so this is IMO acceptable. (tested for Germany with `-Xmx4500m`)

**Update**:

Have tested this manually for larger pillar nodes on a bigger server (with MMAP setting as otherwise RAM requirement would be huge). So I set "5 bytes max minus a few items":

```
private long nextPillarId = (1L << (40 - 1)) - 10;
```

And waited for a proper exception after the 8th pillar node:

```
Caused by: java.lang.IllegalStateException: id for pillar node cannot be bigger than 549 755 813 887
```

Also tested a values slightly below Integer.Max so that the 4->5 byte border was tested and it completed fine

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMNodeData.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMNodeData.java
@@ -83,7 +83,7 @@ class OSMNodeData {
         // We use a b-tree that can store as many entries as there are longs. A tree is also more
         // memory efficient, because there is no waste for empty entries, and it also avoids
         // allocating big arrays when growing the size.
-        idsByOsmNodeIds = new GHLongLongBTree(200, 5, -1);
+        idsByOsmNodeIds = new GHLongLongBTree(200, 5, EMPTY_NODE);
         towerNodes = nodeAccess;
         pillarNodes = new PillarInfo(towerNodes.is3D(), directory);
 
@@ -250,7 +250,7 @@ class OSMNodeData {
     }
 
     public void setTags(ReaderNode node) {
-        int tagIndex = (int) nodeTagIndicesByOsmNodeIds.get(node.getId());
+        int tagIndex = Math.toIntExact(nodeTagIndicesByOsmNodeIds.get(node.getId()));
         if (tagIndex == -1) {
             long pointer = nodeKVStorage.add(node.getTags().entrySet().stream().map(m -> new KVStorage.KeyValue(m.getKey(),
                             m.getValue() instanceof String ? KVStorage.cutString((String) m.getValue()) : m.getValue())).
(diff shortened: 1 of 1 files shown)


easbar:
>  But shouldn't the edges file be exactly the same? 

This was only because I ran the import on two different dates, so unrelated to this PR.

core/src/main/java/com/graphhopper/reader/osm/OSMNodeData.java
```diff
@@ -83,11 +83,11 @@ public OSMNodeData(PointAccess nodeAccess, Directory directory) {
         // We use a b-tree that can store as many entries as there are longs. A tree is also more
         // memory efficient, because there is no waste for empty entries, and it also avoids
         // allocating big arrays when growing the size.
-        idsByOsmNodeIds = new GHLongLongBTree(200, 5);
+        idsByOsmNodeIds = new GHLongLongBTree(200, 5, -1);
```
easbar, line 86: We could use `EMPTY_NODE` here now

core/src/main/java/com/graphhopper/reader/osm/OSMNodeData.java
```diff
@@ -248,7 +250,7 @@ public void addCoordinatesToPointList(int id, PointList pointList) {
     }
 
     public void setTags(ReaderNode node) {
-        int tagIndex = nodeTagIndicesByOsmNodeIds.get(node.getId());
+        int tagIndex = (int) nodeTagIndicesByOsmNodeIds.get(node.getId());
```
easbar, line 253: Maybe use `Math.toIntExact` so we get an error if the tag index is outside the int range?

core/src/main/java/com/graphhopper/reader/osm/OSMNodeData.java
```diff
@@ -262,7 +264,7 @@ public void setTags(ReaderNode node) {
     }
 
     public Map<String, Object> getTags(long osmNodeId) {
-        int tagIndex = nodeTagIndicesByOsmNodeIds.get(osmNodeId);
+        int tagIndex = (int) nodeTagIndicesByOsmNodeIds.get(osmNodeId);
```
easbar, line 267: same here

Issue #2806, opened 2023-04-24
add osm source date to route response


karussell:
Currently it is tricky to know the age of the used OSM data from just the UI. We have it in the /info endpoint but if a cluster of GraphHopper servers is used it is not clear for the client how to pick the dates as the cluster depends on the profile and algorithm. See here: https://github.com/graphhopper/graphhopper-maps/pull/284

One simple solution would be to directly add this to the response:

```
{
  "info": {
     copyrights: ["GraphHopper", "OpenStreetMap contributors"],
     sources: ["GraphHopper release date or version", "date of OSM data"],
     // or
     data_source_date: "date of OSM data",
     took: 123
  },
  // continue /route response
}
```

otbutz:
The only other solution would be to add a reference to the executing backend server and allow direct access to it. As this is obviously a bad idea, adding metadata to the response is the best solution and feels natural IMHO.

PR #2818, opened 2023-05-12
add osm source and import date to route response

ratrun:
Here is an implementation for #2806.


ratrun:
I opened this PR quite a while ago, seems it got overlooked as there is no feedback so far. May I ask to take a look?

PR #2821, opened 2023-05-17
Set empty custom model in custom profile constructor

easbar:
In the current code base there are three places where we call `CustomProfile#getCustomModel` and none of them expects the custom model to be null. Therefore we almost always need to create a custom profile like this: `new CustomProfile("profile_name").setCustomModel(new CustomModel())`. I moved this into the constructor.

In the rare cases where we use the `custom_model_files` hint (in Java code, for tests etc.) we now need to set the custom model to `null`.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomProfile.java
+++ b/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomProfile.java
@@ -34,6 +34,7 @@ public class CustomProfile extends Profile {
     public CustomProfile(String name) {
         super(name);
         setWeighting(CustomWeighting.NAME);
+        setCustomModel(new CustomModel());
     }
 
     public CustomProfile setCustomModel(CustomModel customModel) {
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/graphhopper/routing/weighting/custom/CustomProfile.java
```diff
@@ -34,12 +34,13 @@ public CustomProfile(Profile profile) {
...
         setWeighting(CustomWeighting.NAME);
+        setCustomModel(new CustomModel());
     }
 
     public CustomProfile setCustomModel(CustomModel customModel) {
-        customModel.internal();
+        if (customModel != null)
+            customModel.internal();
         getHints().putObject(CustomModel.KEY, customModel);
-        getHints().putObject("custom_model_files", Collections.emptyList());
```
easbar, line 42: I think there is no need for this as CustomProfile is basically independent from the `custom_model_files` hint. It is only used in `config.yml` to specify the custom model indirectly using files (a single method in GraphHopperManaged).

PR #2827, opened 2023-05-26
move to JDK17

karussell:
Currently we support Java down to JDK8. With this PR we move to the latest LTS JDK and make new APIs like [this](https://bugs.openjdk.org/browse/JDK-5029431) possible - [see branch](https://github.com/graphhopper/graphhopper/commit/127b6fb8dc6a42b89928b79ddff5a8467f2c88ca). (projects like lucene are already on JDK17)

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/MMapDataAccess.java
+++ b/core/src/main/java/com/graphhopper/storage/MMapDataAccess.java
@@ -17,7 +17,6 @@
  */
 package com.graphhopper.storage;
 
-import com.graphhopper.util.Constants;
 import com.graphhopper.util.Helper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -183,8 +183,8 @@
                         <compilerArgument>-Xlint:deprecation</compilerArgument>
                         -->
                         <fork>true</fork>
-                        <source>1.8</source>
-                        <target>1.8</target>
+                        <source>17</source>
+                        <target>17</target>
```
otbutz, line 187:
```suggestion
                        <release>17</release>
```

https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html

otbutz:
This could also be simplified if Java 17 is required:

https://github.com/graphhopper/graphhopper/blob/347003188a208d66c4c9068a10da5d873100279e/core/src/main/java/com/graphhopper/storage/MMapDataAccess.java#L82-L143

otbutz:
@karussell i also made some preparations in #2835 which are also viable while we still build using Java 8.

PR #2828, opened 2023-05-27
Process callables GHUtility.runConcurrently in batches to save memory

easbar:
For the urban density calculation we submit millions of callables (one for each edge) to the ExecutorService which results in millions of Future objects (112byte each) and thus a very high memory consumption. Here we process the callables in batches so we can free the memory.

### DIFF
--- a/core/src/main/java/com/graphhopper/util/GHUtility.java
+++ b/core/src/main/java/com/graphhopper/util/GHUtility.java
@@ -659,8 +659,23 @@ public class GHUtility {
     public static void runConcurrently(Stream<Callable<String>> callables, int threads) {
         ExecutorService executorService = Executors.newFixedThreadPool(threads);
         ExecutorCompletionService<String> completionService = new ExecutorCompletionService<>(executorService);
-        AtomicInteger count = new AtomicInteger();
+        // process callables in batches, because otherwise we require lots of memory for all the Future objects
+        // that will be returned
+        final int batchSize = 100_000;
+        List<Callable<String>> batch = new ArrayList<>(batchSize);
         callables.forEach(c -> {
+            batch.add(c);
+            if (batch.size() == batchSize) {
+                processBatch(batch, executorService, completionService);
+                batch.clear();
+            }
+        });
+        processBatch(batch, executorService, completionService);
+    }
+
+    private static void processBatch(List<Callable<String>> batch, ExecutorService executorService, ExecutorCompletionService<String> completionService) {
+        AtomicInteger count = new AtomicInteger();
+        batch.forEach(c -> {
             count.incrementAndGet();
             completionService.submit(c);
         });


bt90:
Couldn't we split the stream into one chunk per thread instead?

easbar:
How exactly would you split it and what kind of benefit do you expect? We tried a few different things but ended up doing it this way.

bt90:
Let's say you want to execute 4 million Callables using 8 threads. Wouldn't it be more efficient to split them up into chucks of 500k each and execute those sequentially per executor thread?

You'd still end up with all cores being used but with a lot less task scheduling overhead. The chunk size could also be lowered if you fear that some thread could end up with more work.

A parallel stream with a dedicated ForkJoinPool might also handle this a bit better as it's designed to handle small tasks. But no idea if this would have a better memory footprint. We might just replace the memory overhead of the Futures with some Stream internal datastructure.

easbar:
> Let's say you want to execute 4 million Callables using 8 threads. Wouldn't it be more efficient to split them up into chucks of 500k each and execute those sequentially per executor thread?

I don't know, but feel free to try it. 

> You'd still end up with all cores being used but with a lot less task scheduling overhead.

If this really yields a measurable performance improvement we should change it, yes, but I did not make this comparison.

> A parallel stream with a dedicated ForkJoinPool might also handle this a bit better as it's designed to work with small tasks. But no idea if this would have a better memory footprint.

Using a parallel stream there is no way to control the number of threads being used(?).

>  We might just replace the overhead of the Futures with some Stream internal datastructure.

The overhead of the Futures is no longer a problem since this PR processes the jobs in batches.

bt90:
> Using a parallel stream there is no way to control the number of threads being used(?).

You can instantiate your own ForkJoinPool with the desired level of parallelism and invoke the parallel Stream in it.

https://blog.krecan.net/2014/03/18/how-to-specify-thread-pool-for-java-8-parallel-streams/

otbutz:
(to avoid confusion: this is my work related account)

Small test importing germany-latest.pbf with the urban density calculator capped at 8 threads:

## Batch (master)

```
2023-05-30 10:16:53.700 [main] INFO  c.g.r.util.UrbanDensityCalculator - Calculating residential areas ..., radius=300.0, sensitivity=60.0, threads=8
2023-05-30 10:17:06.435 [main] INFO  c.g.r.util.UrbanDensityCalculator - Finished calculating residential areas, took: 12.735071s
2023-05-30 10:17:06.435 [main] INFO  c.g.r.util.UrbanDensityCalculator - Calculating city areas ..., radius=2000.0, sensitivity=30.0, threads=8
2023-05-30 10:22:25.934 [main] INFO  c.g.r.util.UrbanDensityCalculator - Finished calculating city areas, took: 319.4993s
```

## ForkJoinPool non-async

A little bit faster than our current master and the code is IMHO more readable.

<details>
  <summary>Implementation</summary>
  
  ```java
    public static void runConcurrently(Stream<Callable<String>> callables, int threads) {
        ForkJoinPool pool = new ForkJoinPool(threads);
        try {
            pool.submit(() -> callables.parallel().forEach(c -> {
                try {
                    c.call();
                } catch (Exception e) {
                }
            })).get();
        } catch (InterruptedException | ExecutionException e) {
        } finally {
            pool.shutdown();
        }
    }
  ```
</details>

```
2023-05-30 10:05:02.254 [main] INFO  c.g.r.util.UrbanDensityCalculator - Calculating residential areas ..., radius=300.0, sensitivity=60.0, threads=8
2023-05-30 10:05:14.443 [main] INFO  c.g.r.util.UrbanDensityCalculator - Finished calculating residential areas, took: 12.188523s
2023-05-30 10:05:14.443 [main] INFO  c.g.r.util.UrbanDensityCalculator - Calculating city areas ..., radius=2000.0, sensitivity=30.0, threads=8
2023-05-30 10:10:27.089 [main] INFO  c.g.r.util.UrbanDensityCalculator - Finished calculating city areas, took: 312.6464s
```

## ForkJoinPool async

As we don't depend on the order of execution, the async mode might offer better performance. Contrary to my expectations it was a little bit slower in my tests.

<details>
  <summary>Implementation</summary>
  
  ```java
    public static void runConcurrently(Stream<Callable<String>> callables, int threads) {
        ForkJoinWorkerThreadFactory threadFactory = ForkJoinPool.defaultForkJoinWorkerThreadFactory;
        ForkJoinPool pool = new ForkJoinPool(threads, threadFactory, null, true);
        try {
            pool.submit(() -> callables.parallel().forEach(c -> {
                try {
                    c.call();
                } catch (Exception e) {
                }
            })).get();
        } catch (InterruptedException | ExecutionException e) {
        } finally {
            pool.shutdown();
        }
    }
  ```
</details>

```
2023-05-30 10:47:20.661 [main] INFO  c.g.r.util.UrbanDensityCalculator - Calculating residential areas ..., radius=300.0, sensitivity=60.0, threads=8
2023-05-30 10:47:32.178 [main] INFO  c.g.r.util.UrbanDensityCalculator - Finished calculating residential areas, took: 11.517575s
2023-05-30 10:47:32.178 [main] INFO  c.g.r.util.UrbanDensityCalculator - Calculating city areas ..., radius=2000.0, sensitivity=30.0, threads=8
2023-05-30 10:52:48.625 [main] INFO  c.g.r.util.UrbanDensityCalculator - Finished calculating city areas, took: 316.44693s
```

otbutz:
Memory consumption is also looking good:

image

Issue #2663, opened 2022-09-27
Many warnings 'Multiple countries found for way'


karussell:
I think since #2658 we get these errors. It seems to be mainly a Great Britain issue - could it be an issue of an overlapping border of GBR for the JOSM boundaries.osm data @westnordost?

Omitted many duplicates:

Multiple countries found for way [2783749](https://www.openstreetmap.org/way/2783749): JEY, GBR
Multiple countries found for way [4066366](https://www.openstreetmap.org/way/4066366): GGY, GBR
Multiple countries found for way [4965566](https://www.openstreetmap.org/way/4965566): DNK, FRO
Multiple countries found for way [10229296](https://www.openstreetmap.org/way/10229296): GIB, GBR
Multiple countries found for way [14143304](https://www.openstreetmap.org/way/14143304): FLK, GBR
Multiple countries found for way 17782588: NIU, NZL
Multiple countries found for way 22058615: CYM, GBR
Multiple countries found for way 22333039: AIA, GBR
Multiple countries found for way 23087382: SHN, GBR
Multiple countries found for way 23376204: GGY, GBR
Multiple countries found for way 23565589: GBR, VGB
Multiple countries found for way 23893198: NZL, COK
Multiple countries found for way 25969150: TCA, GBR
Multiple countries found for way 25986080: IMN, GBR
Multiple countries found for way 58659323: IOT, GBR
Multiple countries found for way 59638034: BMU, GBR
Multiple countries found for way 86379854: GRL, DNK

westnordost:
I wouldn't call it an issue. Jersey ( `JE`) is simply a part of the United Kingdom. The ISO Code for the United Kingdom is `GB`

westnordost:
And to take into account country subdivisions, whether they have their own ISO-3166-1 alpha-2 (or 3) code (like `GB` or `GBR`, respectively) or "just" a ISO-3166-2 code (like `GB-SCT`) is something that you would likely want to do anyway because speed limits (and other rules) may vary even in country subdivisions.

E.g. each state in the United States, Canada and Australia has completely disjunct speed limit restrictions. Scotland has slightly different speed limit restrictions. All three regions of Belgium have different speed limit restrictions. Until very recently, each state in India had own speed limit laws. Et cetera.

westnordost:
Look out for `FX`, then

otbutz:
Judging from the different shades of blue, there are quite a few overlapping boundaries:

[...]

PR #2830, opened 2023-05-29
Country subdivision (state)

karussell:
To make the Country enum more useful and for #2810.

The country path details stays the same and we only return alpha3 country names there. (We could in the future create a country2 path detail that returns the 2 letter country name plus the subdivision.)

Without this change the default maxspeeds won't be possible e.g. in the USA because every state has different limits and no assignment is possible.

**Update:** this PR introduces a new `state` encoded value. Usage to select a single state is e.g. 

`state == US_CA`

Or to exclude a state from a country:

`country == USA && state != US_CA`

Currently only a subset of country subdivisions are stored: only those that are relevant for speed limit estimation.

### DIFF
--- a/core/src/main/java/com/graphhopper/util/GHUtility.java
+++ b/core/src/main/java/com/graphhopper/util/GHUtility.java
@@ -621,8 +621,7 @@ public class GHUtility {
         for (Country c : Country.values()) {
             if (c == Country.MISSING) continue;
             if (c.getStates().isEmpty()) enumSet.add(c.getAlpha2());
-            else
-                enumSet.addAll(c.getStates().stream().map(State::getStateCode).collect(Collectors.toList()));
+            else for (State s : c.getStates()) enumSet.add(s.getStateCode());
         }
 
         try (Reader reader = new InputStreamReader(GHUtility.class.getResourceAsStream("/com/graphhopper/countries/countries.geojson"), StandardCharsets.UTF_8)) {


otbutz:
`country == USA && subdivision == IN` ?

otbutz:
Adding the country to the subdivision wouldn't hurt and isn't such an eyesore :sweat_smile: 
For consistency, we could stick to a three letter prefix here.

```
country == USA && subdivision == USA_IN
```

easbar:
Using a second encoded value sounds reasonable? It would allow do something like:

```
if: country == USA && state == INDIANA
multiply_by: 0.8
else if: country == USA
multiply_by: 0.9
```

According to this table: https://wiki.openstreetmap.org/wiki/Key:maxspeed
There are only a few countries that use a different legislation for the different 'states':
Australia, Belgium, Canada, United States, Micronesia, UK, Netherlands

> If we store everything we could shorten the above statement further to code2 == US_IN 

But maybe we will introduce some kind of nested if statements which would also be useful for other encoded values then this won't be so important anymore.

easbar:
> I'm unsure about subdivisions that starts with a number [like AD-07](https://en.wikipedia.org/wiki/Parishes_of_Andorra).

But AD-07 does not start with a number? And why do you think would it be necessary that the 'state' encoded value uses an iso code? It could be anything really, no?

> Not sure, as they can get very lengthy like AUSTRALIAN_CAPITAL_TERRITORY.

Don't know, I think we could just use something practical and document the list of possible values.

> But this can change over time as it was https://github.com/graphhopper/graphhopper/issues/2663#issuecomment-1259436517.

But that wouldn't be a problem, would it? Say Mexico did not have subdivisions, then rules using `state == UNSPECIFIED` (or similar) would apply to all of Mexico, and if later on there were specific states like `MXC_1`, `MXC_2`, `MXC_3` using `state == UNSPECIFIED` would still apply to all of Mexico, but `state == MXC_1` could be used to be more specific.

otbutz:
I'd go with the prefix and use the shorter subdivision code.

> There are only a few countries that use a different legislation for the different 'states':

We should focus on a subset of ISO 3166-2 which and extend that if the need arises. The list is otherwise quite big:

https://service.unece.org/trade/locode/2022-2%20SubdivisionCodes.htm

easbar:
> And e.g. CA_NT vs AU_NT and US_WA vs AU_WA do already clash.

You mean using `state == NT` would be ambigiuous and could refer to the state in Canada or the one in Australia? Then users could still do `country == CANADA && state == NT` to be clear. But if we used a longer name for this state this could be avoided, yes.

core/src/main/java/com/graphhopper/routing/ev/Country.java
```diff
@@ -260,15 +285,41 @@ public String getCountryName() {
...
+
+    @Override
+    public String toString() {
+        return alpha3; // for background compatibility
+    }
+
+    /**
+     * @return all ISO 3166-2 i.e. including subdivisions (states) if any.
+     */
+    public List<String> createISO31662Codes() {
```
otbutz, line 321: I would create this list once in the constructor and change this to a getter.

PR #2833, opened 2023-05-30
Use ForkJoinPool for concurrent processing

otbutz:
see https://github.com/graphhopper/graphhopper/pull/2828#issuecomment-1566139538


otbutz:
@easbar what do you think?

easbar:
I like it because the code is simple and in my (limited) tests it even seemed to be a bit faster. Can you update this to recent master, please?

otbutz:
Done :)

easbar:
Thanks

PR #2835, opened 2023-05-31
Improve shading

otbutz:
* properly handles [Multi-Release](https://openjdk.org/jeps/238) dependencies
* removes any `module-info.class` during shading

### DIFF
--- a/web/pom.xml
+++ b/web/pom.xml
@@ -72,6 +72,8 @@
                         <filter>
                             <artifact>*:*</artifact>
                             <excludes>
+                                <exclude>module-info.class</exclude>
+                                <exclude>META-INF/versions/*/module-info.class</exclude>
                                 <exclude>META-INF/*.SF</exclude>
                                 <exclude>META-INF/*.DSA</exclude>
                                 <exclude>META-INF/*.RSA</exclude>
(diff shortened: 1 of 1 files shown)


otbutz:
That's a different story as we're mixing javax and jakarta in our dependencies. Both provide the same API, so we end up with overlapping classes. (Background: https://www.baeldung.com/java-enterprise-evolution)

My changes enable Multi-Release functionality which allows dependencies to make use of newer Java APIs at runtime. You can extract the shaded jar and check the `META-INF/versions` directory to see which classes are using this mechanism:

image

I'm also removing any `module-info.class` files. The strong encapsulation doesn't make sense in a shaded jar hence the previous warning:

```
[WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
```

otbutz:
Any objections?

otbutz:
> > My changes enable Multi-Release functionality which allows dependencies to make use of newer Java APIs at runtime.
> 
> What does this mean?

See https://openjdk.org/jeps/238. The JVM is capable since version 9 to load alternative class implementations targeting different Java versions. The is all transparently handled by the JVM.

e.g. you could ship a jar that still works with Java 8 but provides an alternative implementation for a class that uses Java 17 features.

While GH might not use this mechanism itself, we break it for our shaded dependencies that might require it.

otbutz:
An example where this caused problems in the past: https://github.com/komoot/photon/discussions/731

PR #2837, opened 2023-06-05
Fix error logging for RouteResource

### LINKED ISSUE #2660: Error stack trace becomes visible when using the XML accept header
Depending on the accept header we use with the /route endpoint we can see the error stack trace on the client side:

* Using 'Accept: application/xml' we get to see the error stack trace as XML. For example Firefox and Chrome seem to do this by default, so we see this when we open a /route url in the browser.
* Using 'Accept: application/json' or no accept header at all we get to see the error message as JSON

Not sure if this is a real problem and I just put this as a reminder here. For users who do not use GraphHopper behind a proxy it might be important, but probably this can simply be disabled by some DropWizard configuration parameter in `config.yml`(?).

karussell:
This seems to fix #2660 and for bad requests it avoids a strange stacktrace in the logs.

Update: Ok, something is wrong with this.

### DIFF
--- a/web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
+++ b/web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
@@ -192,7 +192,6 @@ public class RouteResource {
                 + ", custom_model: " + request.getCustomModel();
 
         if (ghResponse.hasErrors()) {
-            logger.info(logStr + " " + ghResponse);
             throw new MultiException(ghResponse.getErrors());
         } else {
             logger.info(logStr + ", alternatives: " + ghResponse.getAll().size()


web-api/src/main/java/com/graphhopper/jackson/MultiException.java
```diff
@@ -26,6 +26,7 @@ public class MultiException extends RuntimeException {
     private final List<Throwable> errors;
 
     public MultiException(List<Throwable> errors) {
+        super(errors.toString());
```
easbar, line 29: Since we did not set the `message` field for the super exception we got this ugly `MultiException: null` in our logs.

web-bundle/src/main/java/com/graphhopper/resources/RouteResource.java
```diff
@@ -190,9 +188,7 @@ public Response doPost(@NotNull GHRequest request, @Context HttpServletRequest h
                 + ", custom_model: " + request.getCustomModel();
 
         if (ghResponse.hasErrors()) {
-            MultiException ex = new MultiException(ghResponse.getErrors());
-            logger.error(logStr, ex);
```
easbar, line 194: Strangely we logged this as an error for POST , but as info for GET. Either way we already log the error in MultiExceptionMapper so I removed this log.

easbar:
Slightly off-topic, but maybe related:

 For this request (note the initial backslash before points)
 
 `curl -X POST 'http://localhost:8989/route' -d '{\"points": [[100000,100000],[100000,100000]], "profile": "car"}' -H "Content-Type: application/json"`

we get an internal server error (500) (and an XML response): `com.fasterxml.jackson.core.JsonParseException: Unexpected character ('\' (code 92)): was expecting double-quote to start field name`. Probably because we do not convert JsonParseException to a MultiException or at least something for which we have setup our custom serialization.

PR #2843, opened 2023-07-05
Make car speed parser bad surfaces consistent with surface encoded value

easbar:
In OSMSurfaceParser we map `sett` and `unhewn_cobblestone` to `COBBLESTONE`, `earth` to `DIRT`, `pebblestone` to `GRAVEL` (!) and `grass_paver` to `GRASS`. 

I think we need to extend the `badSurfaceMap` in CarAverageSpeedParser accordingly to make this consistent.


easbar:
There is another inconsistency: In `OSMSurfaceParser` we ignore everything that is written after a colon in the surface value:

https://github.com/graphhopper/graphhopper/blob/881a04c761a141860a72aaf4fe9c56081062e8bd/core/src/main/java/com/graphhopper/routing/ev/Surface.java#L68-L71

But in `CarAverageSpeedParser` we don't. For example `cobblestone:flattened` will be treated like `cobblestone` by `OSMSurfaceParser`, but `CarAverageSpeedParser` only considers `cobblestone` a bad surface, and not `cobblestone:flattened` (because we only add `cobblestone` to the `badSurfaceMap`).

PR #2846, opened 2023-07-07
Make RoadEnvironment.FERRY overrule RoadEnvironment.FORD also for node tags

easbar:
[This](https://www.openstreetmap.org/node/1306290896) is a fairly exotic example of a barrier node (`ford=yes`) within a ferry route. In this case the question arises which road environment shall be assigned. I think it should be `ferry`, because also for ways `ferry` overrules `ford`.

Btw: It seems a bit odd that there is a special road environment for `ford` but not for the other types of barriers like `gate` etc.?

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/CarAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/CarAverageSpeedParser.java
@@ -105,7 +105,7 @@ public class CarAverageSpeedParser extends AbstractAverageSpeedParser implements
     }
 
     protected double getSpeed(ReaderWay way) {
-        String highwayValue = way.getTag("highway");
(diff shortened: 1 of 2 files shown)


easbar:
> Is this a different road environment as a barrier is still "on the road"?

Not sure, but is ford really a different 'road environment'? And currently we have no other encoded value to identify gates etc.?

easbar:
Maybe they are all 'obstacles' or something?

easbar:
> So even if a barrier is passable it might be interesting that there is one

And even more so when it is **not** passable (when we use the roads vehicle and try to block these edges?).

> could there be barriers in tunnels?

Maybe, yes. But there are also tunnels on roads where we also just set the environment to 'tunnel'. It is just the more specific value in this case and we could do the same for gates in tunnels?

PR #2847, opened 2023-07-12
throw error for too small speed values

karussell:
This also removes the extra setSpeed from foot and there is no need for a separate maxPossibleSpeed parameter in the constructor.

Furthermore this will potentially use a bigger speed value for ferries as it is now limited from max storable value and not from a static value, but this should not make ETAs worse.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/AbstractAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/AbstractAverageSpeedParser.java
@@ -45,7 +45,7 @@ public abstract class AbstractAverageSpeedParser implements TagParser {
 
     protected void setSpeed(boolean reverse, int edgeId, EdgeIntAccess edgeIntAccess, double speed) {
         if (speed < avgSpeedEnc.getSmallestNonZeroValue() / 2) {
-            throw new IllegalArgumentException("Speed cannot be lower than " + avgSpeedEnc.getSmallestNonZeroValue() / 2);
+            throw new IllegalArgumentException("Speed was " + speed + " but cannot be lower than " + avgSpeedEnc.getSmallestNonZeroValue() / 2);
         } else {
             avgSpeedEnc.setDecimal(reverse, edgeId, edgeIntAccess, speed);
         }


core/src/main/java/com/graphhopper/routing/util/parsers/AbstractAverageSpeedParser.java
```diff
@@ -53,13 +44,10 @@ public final DecimalEncodedValue getAverageSpeedEnc() {
     }
 
     protected void setSpeed(boolean reverse, int edgeId, EdgeIntAccess edgeIntAccess, double speed) {
-        // special case when speed is non-zero but would be "rounded down" to 0 due to the low precision of the EncodedValue
-        if (speed > 0.1 && speed < avgSpeedEnc.getSmallestNonZeroValue())
-            speed = avgSpeedEnc.getSmallestNonZeroValue();
-        if (speed < avgSpeedEnc.getSmallestNonZeroValue()) {
-            avgSpeedEnc.setDecimal(reverse, edgeId, edgeIntAccess, 0);
+        if (speed < avgSpeedEnc.getSmallestNonZeroValue() / 2) {
+            throw new IllegalArgumentException("Speed cannot be lower than " + avgSpeedEnc.getSmallestNonZeroValue());
```
easbar, line 48: ... cannot be lower than value ** / 2 **

easbar (approved):

idegeus:
Not sure if this is the right place to comment, but it seems like this causes some issues with some OSM extracts, such as some routes in [Brussels](https://drive.google.com/file/d/111J7liPdZ9nuL9H0EjKWmu5a2krwEn1Z/view?usp=sharing). Seems like it's caused by the factor 0.9 in [AbstractAverageSpeedParser.java:41](core/src/main/java/com/graphhopper/routing/util/parsers/AbstractAverageSpeedParser.java), which is not taken into account when comparing for the smallestNonZeroValue. Adding it there fixes the problem.

easbar:
Do you have a recent OSM file / way ID for which an error is thrown?

idegeus:
Yes, it happens for me with files in Paris and Brussels, see an example file created from planet.osm.pbf using osmium here: https://drive.google.com/file/d/111J7liPdZ9nuL9H0EjKWmu5a2krwEn1Z/view

easbar:
Did you use the default configuration (config-example.yml), or anything special?

idegeus:
Just the default configuration, built in IntelliJ using the default build settings. It builds fine when checking out repo before #fcb95492 but after it, it throws the error.

otbutz:
Could you link the way via openstreetmap.org?

easbar:
I can reproduce it for this way: https://www.openstreetmap.org/way/26172571
... which is tagged as `maxspeed=0.9`.

idegeus:
The function in Brussels gets stuck on EdgeID 52497, I don't know how to make it into a way-id.

easbar:
You would have to set a breakpoint where the error is thrown to find the way ID. But it might be useless if Osmium re-numbered the ways for example. I used the extract from here: https://download.geofabrik.de/europe/belgium.html (which does not use such renumbering).

otbutz:
> I can reproduce it for this way: https://www.openstreetmap.org/way/26172571 ... which is tagged as `maxspeed=1`.

This looks like a tagging error. The tagger has modified `maxspeed` and `lanes` for a few ways and probably mixed them up for this one.

But our import should still be able to cope.

idegeus:
The way I manged to get my docker to work is by incorporating the 0.9 in the if statement too. 
image

easbar:
The minimum, non-zero speed we can store depends on the speed_factor we use for the speed encoded value of the corresponding vehicle. By default it is 2km/h. In this case there will be an error whenever the maxspeed is <= 1km/h. For now I added this simple fix: https://github.com/graphhopper/graphhopper/commit/7c530c2809b93fec8bcf7fb0ee52fc325eb2816a

Hopefully we can soon remove the speed encoded values altogether.

Issue #2530, opened 2022-03-04
Ferry is way too slow for the foot encoder


easbar:
For example this ferry takes 3h44min for the car profile, more than seven hours for bikes and more than nine hours for the foot profile. This is because the maximum speed we can store for the bike/foot profile is less than the speed of the ferry. According to the duration tag it takes only 2h15min. For the car profile the difference arises mostly because we add some waiting time as well. One simple improvement would be using smaller (or no) waiting penalty for foot and bike when the calculated ferry speed is this fast.

https://graphhopper.com/maps/?point=57.592079%2C9.96769&point=58.142808%2C7.990837&locale=en-US&elevation=true&profile=foot&use_miles=false&selected_detail=Elevation&layer=Omniscale

karussell:
Fixed via #3267 and #3266.

PR #2849, opened 2023-07-20
ferry_speed encoded value

karussell:
This adds a new default encoded value `ferry_speed` (5 bits) and solves the different speed problem for ferries (e.g. #2530).

It is also another step to move vehicle parsers to a custom model.


easbar:
Nice, looks good!

PR #2853, opened 2023-07-27
Use same definition for ferries for road environment and speed parsers

easbar:
OSMRoadEnvironmentParser and the speed parsers did not use the same definition of when an OSM way is a ferry route (the road environment parser considered `ferry=no`, but the speed parsers did not). So now they use the same static method for this. 

Note that the access parsers still consider ways tagged with both `route=ferry` and `highway` as highways, while the speed parsers and the road environment parser considers these as ferries. Will fix this in a follow-up PR.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
+++ b/core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
@@ -6,15 +6,17 @@ import com.graphhopper.routing.ev.EdgeIntAccess;
 import com.graphhopper.routing.util.parsers.TagParser;
 import com.graphhopper.storage.IntsRef;
 
-import java.util.Arrays;
-import java.util.Collection;
-
 public class FerrySpeedCalculator implements TagParser {
-    public static final Collection<String> FERRIES = Arrays.asList("shuttle_train", "ferry");
-    private DecimalEncodedValue ferrySpeedEnc;
+    private final DecimalEncodedValue ferrySpeedEnc;
 
     public FerrySpeedCalculator(DecimalEncodedValue ferrySpeedEnc) {
         this.ferrySpeedEnc = ferrySpeedEnc;
+    }
+
+    public static boolean isFerry(ReaderWay way) {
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
```diff
@@ -54,7 +56,7 @@ public static double minmax(double speed, DecimalEncodedValue avgSpeedEnc) {
 
     @Override
     public void handleWayTags(int edgeId, EdgeIntAccess edgeIntAccess, ReaderWay way, IntsRef relationFlags) {
-        if (!way.hasTag("highway") && way.hasTag("route", FERRIES)) {
+        if (isFerry(way)) {
```
easbar, line 59: Since #2849 the speed parsers considered ways tagged with highway as highways, even when there was a `route=ferry` tag. Now they are considered to be ferries again.

core/src/test/java/com/graphhopper/routing/util/parsers/OSMRoadEnvironmentParserTest.java
```diff
@@ -39,6 +39,13 @@ void ferry() {
         parser.handleWayTags(edgeId, edgeIntAccess, way, new IntsRef(2));
         RoadEnvironment roadEnvironment = roadEnvironmentEnc.getEnum(false, edgeId, edgeIntAccess);
         assertEquals(RoadEnvironment.FERRY, roadEnvironment);
+
+        way = new ReaderWay(1);
+        way.setTag("highway", "footway");
+        way.setTag("route", "ferry");
+        parser.handleWayTags(edgeId, edgeIntAccess = new ArrayEdgeIntAccess(1), way, new IntsRef(2));
+        roadEnvironment = roadEnvironmentEnc.getEnum(false, edgeId, edgeIntAccess);
+        assertEquals(RoadEnvironment.FERRY, roadEnvironment);
```
easbar, line 48: We consider route=ferry+highway to be a ferry. Probably not very important, but it should be consistent with the OSMRoadEnvironment parser (and access).

PR #2855, opened 2023-08-02
PT-Isochrones with different access-egress modes. 

idegeus:


web-bundle/src/main/java/com/graphhopper/resources/PtIsochroneResource.java
```diff
@@ -93,14 +99,22 @@ public Response doGet(
...
+        DecimalEncodedValue accessSpeedEnc = encodingManager.getDecimalEncodedValue(VehicleSpeed.key(accessProfile));
+        final Weighting accessWeighting = new FastestWeighting(accessProfileEnc, accessSpeedEnc);
+        DefaultSnapFilter accessSnapFilter = new DefaultSnapFilter(accessWeighting, encodingManager.getBooleanEncodedValue(Subnetwork.key(accessProfile)));
+        BooleanEncodedValue egressProfileEnc = encodingManager.getBooleanEncodedValue(VehicleAccess.key(egressProfile));
+        DecimalEncodedValue egressSpeedEnc = encodingManager.getDecimalEncodedValue(VehicleSpeed.key(egressProfile));
+        final Weighting egressWeighting = new FastestWeighting(egressProfileEnc, egressSpeedEnc);
+        DefaultSnapFilter egressSnapFilter = new DefaultSnapFilter(egressWeighting, encodingManager.getBooleanEncodedValue(Subnetwork.key(egressProfile)));
+        PtLocationSnapper.Result snapResult = new PtLocationSnapper(baseGraph, locationIndex, gtfsStorage).snapAll(Arrays.asList(location), Arrays.asList(accessSnapFilter, egressSnapFilter));
+
+        // TODO: Should also incorporate egressWeighting in some way? Maybe?
```
michaz, line 113: Ah. The issue is that when we exit a station while expanding our circle, we don't know yet if we are going to take another transit leg, or we are already on our egress leg. The router class doesn't support creating labels for both at once.

I think what you chose to do is fine -- only consider one street mode, and call it access or egress depending on the "direction". It does mean that when we say "car", we will take our transfers by "car" as well, so we basically take our car on the bus.

To be resolved another time.

michaz (approved):

PR #2862, opened 2023-08-14
No more loop edges

### LINKED ISSUE #1631: Setting different speeds for both directions can cause problems with loops

easbar:
This shall fix #1631. I think we should just deny loop edges in BaseGraph altogether. For PT we use a separate graph now.


easbar:
> I cannot really comment on the impact of this change as you are more experienced for loops, but I like that the code gets simpler :)

Thanks for taking a look. We discussed this already in #1631 where this issue is explained quite well, I think.

easbar:
If we ever need edge-based CH with base-graph loops we should def. take a look here, because the implementation was working with loops, just not with loops that have different edge properties per direction (which is ill-defined, and cannot work with normal Dijkstra either). To get this right took some serious effort, but since we just don't need it I think it is sufficient to preserve it only in the git history and move on with a simpler graph model without loops.

Issue #1260, opened 2018-01-05
List of restrictions in path details


thomfre:
It would be nice if we could request a list of restrictions (seasonal closure, controlled border crossings etc) in the path details, to warn the user about possible road closures or expected delays.

boldtrn:
I agree this would be interesting. Border crossings could be done by checking for [`barrier=border_control`](https://wiki.openstreetmap.org/wiki/Tag:barrier=border_control). In Europe most border_control tags some border control tags are still used, even though they are not active anymore, which could lead to unexpected results. Seasonal closure with conditional access values.

Issue #1326, opened 2018-04-06
Consider via_ferrata for hike mode


karussell:
See [this route](https://graphhopper.com/maps/?point=46.45199%2C11.65928&point=46.451859%2C11.659058&locale=en&vehicle=hike&layer=TF%20Outdoors). Or is this too dangerous or with [<1000 occasions](https://taginfo.openstreetmap.org/tags/highway=via_ferrata) not worth?

hungerburg:
There is a nice mountain hiking path to Bettelwurfhütte from Absam - https://graphhopper.com/maps/?point=47.307968%2C11.519214&point=47.331019%2C11.510749&point=47.338586%2C11.507562&profile=hike&layer=OpenStreetMap

Remove the via node, graphhopper selects the shorter path over a via ferrata, in both the foot and hike profiles. Is this deliberate?

easbar:
No, GraphHopper should not use the via-ferrata here, especially not for the foot profile.

This is the via-ferrata: https://www.openstreetmap.org/way/128322791
```
access:conditional=no @ (Nov 1-Apr 30)
foot=yes
highway=via_ferrata
name=Absamer Klettersteig (KS4-B bzw. C/D)
source=survey
trail_visibility=good
via_ferrata_scale=3
wikidata=Q30350091
wikipedia=de:Absamer Klettersteig
```

There are multiple reasons GraphHopper still uses it currently:

1) For access tags the most specific tag counts. We have:
```
access:conditional=no @ (Nov 1-Apr 30)
foot=yes
```
So `foot=yes` overrules the (conditional) `access=no` tag. The correct tagging would probably be:
```
access=no
foot:conditional=yes @ (May 1-Oct 31)
```
which means that access is forbidden for all transportation modes, except for pedestrians during summer.

2) GraphHopper normally would not consider `highway=via_ferrata` (hence this issue), but it still does, because of the explicit `foot=yes` tag. Since the current idea is to exclude via-ferratas this is a bug in GraphHopper. 

3) GraphHopper not only sets the access flag for the via-ferrata to 'true', but sets the speed to 5km/h, which would probably be way too fast, even if the via-ferrata was allowed. For this reason it even prefers the via-ferrata over the other route (the one you showed with the via-node).

[...]

PR #2863, opened 2023-08-14
Make conditional access restrictions configurable

### LINKED ISSUE #2477: Allow ignoring seasonal roads
Right now I'm planning for a roadtrip next summer, Graphhopper so far has been by far the best tool for this that I found. This route will take us through Norway which has lots of roads only open in the summer. Since I'm planning for the summer, this isn't a problem but I'm making this route in a season in which those roads are closed, and thus Graphhopper refuses to calculate routes on those roads. It makes it hard to plan months in advance as the route as I make it now will look different when we are actually driving it.

I'd like an option to ignore seasonal restrictions and just route over such a road, no matter what season I'm in.

karussell:
This PR removes the conditional blocking access from the vehicle parsers. I.e. after this change you need to add e.g. car_road_access_conditional == NO into a custom model to still consider temporary restrictions due to constructions.

This PR also:

 - removes the ConditionalOSMTagInspector
 - fixes a bug in KVStringDetails
 - fixes #2477 -> [for foot](https://graphhopper.com/maps/?point=50.909136%2C14.213924&point=50.90918%2C14.213549&profile=foot) but keep via custom model [for car](https://graphhopper.com/maps/?point=51.19414%2C12.957052&point=51.178754%2C12.997822&profile=car)
 - introduces path details for conditional tags like `access_conditional`. (see also #2860)
 - related to #1260
 - minor increase of graph storage, but it is neglectable: edgekv_vals increases 0.5% (4MB) for planet import.

I added some more path details to the Measurement so the response time is expected to be a bit slower for this single test.

[Here](https://www.openstreetmap.org/way/206636669) is an OSM way with `vehicle:conditional=no @ (Nov 1-Mar 31)` for testing purposes. [GH Maps link](https://graphhopper.com/maps/?point=50.806881%2C13.784196&point=50.80638%2C13.779143&profile=car).

On GraphHopper Maps we now show a route hint for outdoor profiles and in the details widget the exact values are shown:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,13 @@
 
 ![Build Status](https://github.com/graphhopper/graphhopper/actions/workflows/build.yml/badge.svg?branch=master)
 
-GraphHopper is a fast and memory-efficient routing engine released under Apache License 2.0. It can be used as a Java library or standalone web server to calculate the distance, time, turn-by-turn instructions and many road attributes for a route between two or more points. Beyond this "A-to-B" routing it supports ["snap to road"](README.md#Map-Matching), [Isochrone calculation](README.md#Analysis), [mobile navigation](README.md#mobile-apps) and [more](README.md#Features). GraphHopper uses OpenStreetMap and GTFS data by default and it can import [other data sources too](README.md#OpenStreetMap-Support).
+GraphHopper is a fast and memory-efficient routing engine released under Apache License 2.0.
+It can be used as a Java library or standalone web server to calculate the distance, time,
+turn-by-turn instructions and many road attributes for a route between two or more points.
+Beyond this "A-to-B" routing it supports ["snap to road"](README.md#Map-Matching),
+[Isochrone calculation](README.md#Analysis), [mobile navigation](README.md#mobile-apps) and
+[more](README.md#Features). GraphHopper uses OpenStreetMap and GTFS data by default and it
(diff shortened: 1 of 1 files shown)


mihai-craciun-ursu:
Hello, is there any timeline when this will be released? I supposed this will be included in 9.0, I just hoped to know when will 9.0 be available.

hungerburg:
I do not see the conditional here, all the way "Type: null" - https://graphhopper.com/maps/?point=47.307968%2C11.519214&point=47.338586%2C11.507562&profile=hike&layer=TF+Outdoors - It routes over a [via ferrata](https://www.openstreetmap.org/way/128322791) that is closed in winter. C.f. https://github.com/graphhopper/graphhopper/issues/1326

hungerburg:
I read "refresh necessary" and did not see a change. Then I read "click on the icon" and now I see. It would have never appeared to me that the icons can be clicked.

From A user PoV I would be fine with being told, a bit more immediately visible than now - perhaps the section pre-clicked? I would not expect a magic button to solve the issue. Some routers offer alternatives in all cases, maybe suggest to tune by placing a via close by - What do I know?

PS - In this very case: I think, graphhopper should just drop any "highway=via_ferrata" when ingesting OSM data. Much as OSM-Carto does or thunderforest does. These highways are of no use neither for the foot not the hike profile.

PR #2864, opened 2023-08-16
Remove Weighting#calcWeightWithAccess

### LINKED ISSUE #1835: Move access flag checks into Weighting

easbar:
This is related to #1835. The access encoded values should be treated like all the other encoded values within the weightings.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
+++ b/core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
@@ -569,8 +569,9 @@ public class LandmarkStorage {
         return ((int) landmarkWeightDA.getShort(pointer) & 0x0000FFFF) == SHORT_INFINITY;
     }
 
+    // todonow: apparently this method is only used in one test. can we remove it?
     int calcWeight(EdgeIteratorState edge, boolean reverse) {
-        return (int) (weighting.calcEdgeWeight(edge, reverse) / factor);
+        return (int) (weighting.calcEdgeWeightWithAccess(edge, reverse) / factor);
     }
 
     // From all available landmarks pick just a few active ones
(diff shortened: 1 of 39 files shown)


core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
```diff
@@ -569,8 +569,9 @@ boolean isInfinity(long pointer) {
         return ((int) landmarkWeightDA.getShort(pointer) & 0x0000FFFF) == SHORT_INFINITY;
     }
 
+    // todonow: apparently this method is only used in one test. can we remove it?
```
easbar, line 572: @karussell ?

easbar:
As a first step I tried to use `calcEdgeWeightWithAccess` everywhere to make sure the access flags are checked whenever we are calculating edge weights. Next we could rename `calcEdgeWeight` to `calcEdgeWeightWithoutAccess` and `calcEdgeWeightWithAccess` to `calcEdgeWeight`? And then we're done already?!

core/src/main/java/com/graphhopper/routing/lm/LandmarkStorage.java
easbar: thanks for the cleanup :)

easbar:
> to filter edges based on the access before we even calculated the edge weight, but this does not even seem to be the case anymore?

We still do this for CH in a way (using `weighting#edgeHasNoAccess`), but with this PR we at least get rid of the `...withAccess` methods.

PR #2865, opened 2023-08-16
Remove shortest weighting from DefaultWeightingFactory class

karussell:
Vehicle parsers are still something we want to remove and for shortest (and fastest) weighting this is still necessary.

The ShortestWeighting can still be used internally and it is still possible to consider the distance cost via the distance_influence of the custom model.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/weighting/ShortestWeighting.java
+++ b/core/src/main/java/com/graphhopper/routing/weighting/ShortestWeighting.java
@@ -49,13 +49,8 @@ public class ShortestWeighting implements Weighting {
     }
 
     @Override
-    public boolean edgeHasNoAccess(EdgeIteratorState edgeState, boolean reverse) {
-        return reverse ? !edgeState.getReverse(accessEnc) : !edgeState.get(accessEnc);
-    }
-
-    @Override
     public double calcEdgeWeight(EdgeIteratorState edgeState, boolean reverse) {
-        if (edgeHasNoAccess(edgeState, reverse))
+        if (reverse ? !edgeState.getReverse(accessEnc) : !edgeState.get(accessEnc))
             return Double.POSITIVE_INFINITY;
         return edgeState.getDistance();
     }
--- a/core/src/main/java/com/graphhopper/routing/weighting/AbstractAdjustedWeighting.java
+++ b/core/src/main/java/com/graphhopper/routing/weighting/AbstractAdjustedWeighting.java
@@ -39,11 +39,6 @@ public abstract class AbstractAdjustedWeighting implements Weighting {
     }
 
     @Override
-    public boolean edgeHasNoAccess(EdgeIteratorState edgeState, boolean reverse) {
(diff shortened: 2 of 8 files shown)


core/src/main/java/com/graphhopper/routing/weighting/ShortestWeighting.java
```diff
@@ -51,6 +60,28 @@ public double calcEdgeWeight(EdgeIteratorState edgeState, boolean reverse) {
         return edgeState.getDistance();
     }
 
+    @Override
+    public long calcEdgeMillis(EdgeIteratorState edgeState, boolean reverse) {
+        if (edgeState.getBaseNode() == edgeState.getAdjNode()) reverse = false;
```
easbar, line 65: This should be removed or an error should be thrown, there should be no loops.

core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -248,7 +248,7 @@ public GraphHopper setStoreOnFlush(boolean storeOnFlush) {
      * <pre>
      * {@code
      *   hopper.setProfiles(
-     *     new Profile("my_car").setVehicle("car").setWeighting("shortest"),
+     *     new CustomProfile("my_car").setVehicle("car"),
```
easbar, line 251: maybe the same for fastest below. even though it is still available it's not recommended

core/src/test/java/com/graphhopper/routing/ch/NodeBasedNodeContractorTest.java
```diff
@@ -216,7 +216,8 @@ public void testContractNode_directed_withWitness() {
 
     @Test
     public void testNodeContraction_shortcutDistanceRounding() {
-        assertTrue(weighting instanceof ShortestWeighting, "this test was constructed assuming we are using the ShortestWeighting");
+        // TODO NOW assertTrue(weighting instanceof ShortestWeighting, "this test was constructed assuming we are using the ShortestWeighting");
```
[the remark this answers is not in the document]
easbar: yes, can we put this back? the test was constructed in a certain way that only failed for the shortest weighting

web-bundle/src/main/java/com/graphhopper/resources/IsochroneResource.java
```diff
@@ -104,7 +104,7 @@ public Response doGet(
         ToDoubleFunction<ShortestPathTree.IsoLabel> fz;
         if (weightLimit.orElseThrow(() -> new IllegalArgumentException("query param weight_limit is not a number.")) > 0) {
             limit = weightLimit.getAsLong();
-            shortestPathTree.setWeightLimit(limit + Math.max(limit * 0.14, 2_000));
+            shortestPathTree.setWeightLimit(limit + Math.max(limit * 0.14, 200));
```
easbar, line 107: just to be sure: this was intentional?

easbar (commented):
lgtm

PR #2869, opened 2023-08-22
Remove optimization in CH iterator

easbar:


core/src/main/java/com/graphhopper/storage/RoutingCHEdgeIteratorStateImpl.java
```diff
@@ -119,24 +119,12 @@ public double getWeight(boolean reverse) {
...
-     */
-    double getOrigEdgeWeight(boolean reverse, boolean needWeight) {
-        // todo: for #1835 move the access check into the weighting
-        final EdgeIteratorState baseEdge = getBaseGraphEdgeState();
-        if (baseEdge.getBaseNode() == baseEdge.getAdjNode())
-            throw new IllegalStateException("Unexpected loop-edge at node: " + baseEdge.getBaseNode());
-        if (weighting.edgeHasNoAccess(baseEdge, reverse))
-            return Double.POSITIVE_INFINITY;
-        if (!needWeight)
-            return 0;
```
easbar, line 138: I'm pretty sure this made a difference at *some* point, but now it looks like the speed difference is negligible if present at all. Implementing the method with custom weighting will be hard in the future anyway, because we want to get rid of the explicit access check in CustomWeighting. One option might be to implement an early access in CustomWeighting when speed/priority are multiplied by zero, and thus the weight will certainly be infinite.

PR #2872, opened 2023-08-29
Remove edgeHasNoAccess from Weighting interface

easbar:


reader-gtfs/src/main/java/com/graphhopper/gtfs/GraphExplorer.java
```diff
@@ -160,7 +160,7 @@ private Iterable<MultiModalEdge> streetEdgeStream(int streetNode) {
             @Override
             public boolean tryAdvance(Consumer<? super MultiModalEdge> action) {
                 while (e.next()) {
-                    if (!accessEgressWeighting.edgeHasNoAccess(e, reverse)) {
```
easbar, line 163: @michaz Is this ok? `!edgeHasNoAccess` did not necessarily mean that the edge is accessible, but I guess you were not interested in edges with no 'access' and infinite weight here? So it should be ok to just keep those for which the weight is finite?

PR #2877, opened 2023-09-07
include osm data date as road_data_timestamp in response

### LINKED ISSUE #2806: add osm source date to route response

karussell:
Fixes #2806 and replaces #2818.

Here I return only one property (as the import date is not a meaningful information). 

(without modifying GHResponse; no call to the `synchronized` StorableProperties.get method in every routing request)

### DIFF
--- a/web-api/src/main/java/com/graphhopper/jackson/ResponsePathSerializer.java
+++ b/web-api/src/main/java/com/graphhopper/jackson/ResponsePathSerializer.java
@@ -86,12 +86,16 @@ public class ResponsePathSerializer {
         sb.append((char) (num));
     }
 
-    public static ObjectNode jsonObject(GHResponse ghRsp, boolean enableInstructions, boolean calcPoints, boolean enableElevation, boolean pointsEncoded, double took) {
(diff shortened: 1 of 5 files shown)


ratrun:
As far as I know the osm data date may be missing in an OSM data extract. Did you check that it is present in the planet data you are using in your regular update process for the hosted version? I know that it is present in geofabric data extracts, but I and many more would like to also see the data date in the public graphhopper web site. In case that you checked that I'm happy to leave out the import date.

PR #2883, opened 2023-09-13
Allow u-turns only at dead ends

easbar:
This is based on #2882.

So far we allowed u-turns at every junction, which in some cases yields suboptimal routes. The number of suggested u-turns can be reduced by increasing the u-turn costs, but it is still debatable if suggesting a u-turn at any junction is a useful. The better default might be to only allow u-turns at dead-ends, where they are required when curbside or heading parameters are used. In this PR I added a new turn cost encoded value that marks dead ends and adjusted the turn weighting such that u-turns are only allowed at these dead ends.


core/src/main/java/com/graphhopper/routing/ev/DeadEnd.java
```diff
@@ -0,0 +1,32 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.graphhopper.routing.ev;
+
+import static com.graphhopper.routing.util.EncodingManager.getKey;
+
+public class DeadEnd {
```
easbar, line 23: In case there will be u-turns other than dead-ends that we want to explicitly allow in the future, maybe we should use a more general name(?).

0709wiwiwi: Sorry as (non dev) GPS user I have to describe it less technically.

U-turns are sure very useful.
In the Kurviger and Cruiser app 180° u-turns are very usefully used.
See example in the image and attached gpx navigation track.
By Notepad++ or gpx editor, find turn instructions in trackpoints.

Cruiser_GraphHopper planning.
The 180° u-turn is helpfully placed to turn here. A 180° U-turn so NO Left nor Right u-turn command. All U-turns are prominently marked as a red Icon in the Cruiser turnlist. Warn thus that a shaping or via point may have been falsely placed on a side road in the planner.

Cruiser_Brouter planning.
Same as above but with an extra.
A "correct misplaced via points" tool can presumably thus automatically prevent false out and return track glitches with 180° u-turn commands. The correction distance is adjustable between 0 and 100m (default 40m).
Provided additional shaping point, a 180° u-turn can still be enforced even in this correction range.

Cruiser_Brouter navigation.
If you ignore the u-turn command during navigation, the system does not insist on a 180° u-turn but suggests a forward alternative (large diversions if necessary). As a person, you can of course still turn back.

RouteYou website planner. (osrm router).
Sure allows 180° U-turns. In addition, there is a "correct misplaced via points" tool. The correction distance at which the tool intervenes is 100m. Track glitches shorter than 100m are so automatically prevented.
Usefull u-turn example
[Usefull u-turn navtrack.gpx.txt](https://github.com/graphhopper/graphhopper/files/13390588/Usefull.u-turn.navtrack.gpx.txt)

PR #2889, opened 2023-10-18
Unsychronized MMapDataAccess

karussell:
Since #2827 we can now use getBytes with absolute indices and so the MMapDataAccess should be read-thread-safe even without `synchronized`. Note, that writing is still not thread-safe (but this is the same for RAMDataAccess) as e.g. a value change for an EncodedValue might still involve several method calls, i.e. the synchronization for write-thread-safety would be  required on a level above the DataAccess.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -19,3 +19,11 @@ ij_formatter_tags_enabled = true
 ij_smart_tabs = false
 ij_visual_guides = none
 ij_wrap_on_typing = false
+
+[*.json]
+indent_size = 2
+indent_style = space
+
+[*.{yml,yaml}]
+indent_size = 2
+indent_style = space
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,11 +2,11 @@ name: Build and Test
 on: push
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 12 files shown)


otbutz:
We could also switch to a MethodHandle while we're at it:

```java
    static final MethodHandle CLEANER;
    static {
        MethodHandle handle;
        try {
            Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
            MethodHandles.Lookup lookup = MethodHandles.privateLookupIn(unsafeClass, MethodHandles.lookup());
            MethodType type = MethodType.methodType(void.class, ByteBuffer.class);
            handle = lookup.findVirtual(unsafeClass, "invokeCleaner", type);
            VarHandle unsafeField = lookup.findStaticVarHandle(unsafeClass, "theUnsafe", unsafeClass);
            Object unsafe = unsafeField.get();
            handle = handle.bindTo(unsafe);
        } catch (Exception e) {
            LOGGER.error("Failed to obtain method handle for ByteBuffer cleaner.", e);
            handle = null;
        }
        CLEANER = handle;
    }

    // ....

    public static void cleanMappedByteBuffer(final ByteBuffer buffer) {
        if (CLEANER == null) {
            throw new RuntimeException("Cleaner is not available");
        }

        try {
            CLEANER.invoke(buffer);
        } catch (Throwable e) {
        }
    }
```

otbutz:
https://github.com/graphhopper/graphhopper/pull/2893

michaz:
Nice!

PR #2891, opened 2023-10-19
Remove hidden evs

easbar:


easbar:
Yes, also the turn cost EV would no longer need to be hidden since #2884.

PR #2893, opened 2023-10-19
Switch to static MethodHandle

otbutz:
Avoids the reflection overhead on each call.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/MMapDataAccess.java
+++ b/core/src/main/java/com/graphhopper/storage/MMapDataAccess.java
@@ -24,18 +24,16 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
-import java.lang.reflect.Field;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/graphhopper/storage/MMapDataAccess.java
```diff
@@ -55,6 +53,9 @@
  */
 public final class MMapDataAccess extends AbstractDataAccess {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(MMapDataAccess.class);
+    private static final MethodHandle BYTE_BUFFER_CLEANER = getByteBufferCleaner();
```
otbutz, line 57: We can also switch to an anonymous class implementing `Supplier<MethodHandle>` if we want the exception on each call.

otbutz:
I can throw in a dedicated unit test if you prefer.

otbutz:
@karussell done :wink:

otbutz:
This may also benefit GraalVM compilation. Static method handles are supported.

PR #2901, opened 2023-11-01
barrier parser

karussell:
Add barrier to encoded_values:

```
  graph.encoded_values: barrier
```

Then you can use it in a custom_model and as path detail

See [discussion](https://discuss.graphhopper.com/t/routing-to-avoid-stiles/8275).

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMBarrierParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMBarrierParser.java
@@ -1,0 +1,43 @@
+package com.graphhopper.routing.util.parsers;
+
+import com.graphhopper.reader.ReaderWay;
+import com.graphhopper.routing.ev.Barrier;
+import com.graphhopper.routing.ev.EdgeIntAccess;
+import com.graphhopper.routing.ev.EnumEncodedValue;
+import com.graphhopper.storage.IntsRef;
+import com.graphhopper.util.Helper;
+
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/OSMBarrierParser.java
```diff
@@ -0,0 +1,43 @@
...
+public class OSMBarrierParser implements TagParser {
+    protected final EnumEncodedValue<Barrier> barrierEnc;
+
+    public OSMBarrierParser(EnumEncodedValue<Barrier> barrierEnc) {
+        this.barrierEnc = barrierEnc;
+    }
+
+    @Override
+    public void handleWayTags(int edgeId, EdgeIntAccess edgeIntAccess, ReaderWay way, IntsRef relationFlags) {
+        if (way.hasTag("gh:barrier_edge")) return;
```
easbar, line 22: Shouldn't this be `!way.hasTag(...)`? The ways with the `gh:barrier_edge` are the ones we are interested it, no?

otbutz (changes requested):

otbutz:
`barrier=cycle_barrier` might be useful for bike routing.

GeoffreyFardeau:
Up 

This looks great! It would be very useful to have this merged into master so we can start making use of the barrier encoded value in our custom models and routing logic. Thanks for the work on this!

PR #2913, opened 2023-12-09
bus access

karussell:
This introduces a compound boolean EncodedValue `bus_access` including the `oneway` handling but without `road_class` handling ("bus" in the sense of a psv = public service vehicle). 

A separate `bus_oneway_access` is unnecessary as the only known use case atm (for ambulances) would be to allow opposite travel of oneways which could be achieved with [`backward_bus_access`](https://github.com/graphhopper/graphhopper/pull/2760) instead.

This is a replacement of #2710. The new ModeAccessParser can be used with any of the values in the TransportationMode enum e.g. also `motorcycle`. The idea is that is can be (later) used as a replacement for CarAccessParser too. (Furthermore the oneway code in CarAccessParser should be simplified)

Unfortunately the enum in #2710 cannot handle oneway roads properly. For #2710 there is a problem: if bus=yes on oneway roads and motor_vehicle=no then the bus profile would allow access in the opposite direction of the oneway road too. See [this route](http://localhost:8989/?point=52.508764%252C13.370041&point=52.508813%252C13.368072) as an example. (We could make the enum direction-dependent to handle oneway roads. Then we could also consider tags like "bus=designated" in the profile to prefer these ways. But this would be too confusing as the enum would no longer represent the OSM data and bus==YES in the custom_model would not necessarily mean bus=yes in OSM.)

The usage is outlined in the bus.json.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/ModeAccessParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/ModeAccessParser.java
@@ -57,7 +57,11 @@ public class ModeAccessParser implements TagParser {
     }
 
     private static String getFirstPriorityNodeTag(Map<String, Object> nodeTags, List<String> restrictionKeys) {
-        return restrictionKeys.stream().map(key -> (String) nodeTags.get(key)).findFirst().orElse("");
+        for (String key : restrictionKeys) {
+            String val = (String) nodeTags.get(key);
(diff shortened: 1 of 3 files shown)


tolot27:
Will this PR also result in a new transportation profile at GraphHopper Maps? Currently, I'm using custom profiles but they do not work for distances further than 500 km apart.

PR #2917, opened 2023-12-19
Simplify import hooks

karussell:
This PR avoids calling calculateUrbanDensity from importOSM and hides writeEncodingManagerToProperties.


easbar:
Thanks, this looks good to me.

Issue #1773, opened 2019-10-29
Graph sorting should be internal to the graph


easbar:
See discussion in #1757. For example this is needed when we include turn costs, because we also need to sort the `TurnCostExtension` (related: #1398). This is a bit tricky right now, because `BaseGraph` only 'sees' the `GraphExtension` interface, see #1772.

karussell:
What exactly is necessary here? Implementing something like graph.sort()? The problem I see with this is that currently sorting requires a new graph and cannot be done inplace (which would be very tricky).

easbar:
Well first of all the method that does the sorting needs to have access to everything that needs to be sorted. Right now this is not the case because it is missing the turn cost storage for example. Second, this will not work the same way for the BaseGraph and the CHGraph, so first these should be decoupled (#1780). Creating a (sorted) copy of the graph I think should be ok, because it will not take more memory than for example CH preparation.

PR #2919, opened 2024-01-03
remove graph sorting

### LINKED ISSUE #2170: Graph sorting inflates StringIndex size
**Describe the bug**

The size of the StringIndex increases if `grap.do_sort` is enabled.

Size comparison of `string_index_vals` with `austria-latest.osm.pbf`:

* unsorted: 6.1MB
* sorted: 8.1MB

**To Reproduce**

Run an import with `graph.do_sort: false` and one with `graph.do_sort: true` and compare the size of the resulting `string_index_vals`.

**Expected behavior**

Sorting the graph shouldn't impact the StringIndex size.

karussell:
It would be nice to have this small performance boost but the graph sorting does not work with (complex) turn restrictions (yet), is mostly unused and so for now it is better to remove it. 

Fixes #2170 and #1773.

Was reintroduced in #3177.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -899,6 +899,12 @@ public class GraphHopper {
     }
 
     protected void postImportOSM() {
+        // Important note: To deal with via-way turn restrictions we introduce artificial edges in OSMReader (#2689).
+        // These are simply copies of real edges. Any further modifications of the graph edges must take care of keeping
+        // the artificial edges in sync with their real counterparts. So if an edge attribute shall be changed this change
+        // must also be applied to the corresponding artificial edge.
+
+
         calculateUrbanDensity();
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/graphhopper/GraphHopper.java
```diff
@@ -917,17 +906,6 @@ protected void postImportOSM() {
             maxSpeedCalculator.close();
         }
 
-        // Important note: To deal with via-way turn restrictions we introduce artificial edges in OSMReader (#2689).
-        // These are simply copies of real edges. Any further modifications of the graph edges must take care of keeping
```
easbar, line 921: This comment is still true and not only related to graph sorting so I would keep it.

easbar: Maybe at the beginning of postImportOSM? It does not really matter. If you want we can also delete it and just keep it in mind :)

Issue #2783, opened 2023-03-31
Dropwizard 3/4


otbutz:
The stable releases of Dropwizard 3.x and 4.x just dropped: https://jvm.social/@dropwizard/110114501892964091

PR #2920, opened 2024-01-04
Upgrade to dropwizard 3.0.8

karussell:
See https://github.com/dropwizard/dropwizard/discussions/7880

The Nullable import now looks a bit strange. This is due to https://github.com/dropwizard/dropwizard/pull/5176


otbutz:
Fixes #2783

PR #2923, opened 2024-01-09
accept external EncodedValues in custom model

### LINKED ISSUE #2894: Setting a EncodedValueFactory fails if the EnumEncodedValue class is not within `com.graphhopper.routing.ev` package
**Describe the bug**
I needed some time to understand that setting a EncodedValueFactory fails if the EnumEncodedValue class is not within `com.graphhopper.routing.ev` package.
https://github.com/graphhopper/graphhopper/blob/7e0d17a79165146df1b6f09d4ccf1d356e2fced2/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomModelParser.java#L283

**To Reproduce**

Create a package `com.test.publictransport` with an EncodedValueFactory:

```
public class BusEncodedValueFactory extends DefaultEncodedValueFactory {

	public BusEncodedValueFactory() {
		super();
	}

	@Override
	public EncodedValue create(String name, PMap properties) {
		if (Bus.KEY.equals(name)) {
			return new EnumEncodedValue<>(Bus.KEY, Bus.class);
		}  else {
			return super.create(name, properties);
		}

	}
}
```

and a `Bus` Enum:

```
public enum Bus {
	MISSING, YES, DESIGNATED, DESTINATION, NO;

	public static final String KEY = "bus";

	public static EnumEncodedValue<Bus> create() {
		return new EnumEncodedValue<>(Bus.KEY, Bus.class);
	}

	@Override
	public String toString() {
		return Helper.toLowerCase(super.toString());
	}

	public static Bus find(String name) {
		if (name == null)
			return MISSING;
		try {
			return Bus.valueOf(Helper.toUpperCase(name));
		} catch (IllegalArgumentException ex) {
			return MISSING;
		}
	}
}
```

If you set a `BusEncodedValueFactory` instance via `hopper.setEncodedValueFactory` you should get the following error:

`java.lang.IllegalStateException: Could not deserialize encoded value: {"className":"com.graphhopper.routing.ev.EnumEncodedValue","name":"bus","bits":3,"min_storable_value":0,"max_storable_value":7,"max_value":4,"negate_reverse_direction":false,"store_two_directions":false,"fwd_data_index":2,"bwd_data_index":0,"fwd_shift":8,"bwd_shift":-1,"fwd_mask":1792,"bwd_mask":0,"enum_type":"com.graphhopper.routing.ev.Bus"}, error: Cannot construct instance of `java.lang.Class`, problem: com.graphhopper.routing.ev.Bus`

**Expected behavior**

- the factory should be loaded correctly OR 
- it should be documented e.g. in `setEncodedValueFactory` that you have to place the enums into `com.graphhopper.routing.ev` OR
- give the user the possibility to specify another location for Janino imports

karussell:
fixes #2894.

@lukasalexanderweber can you have a look into this if this works for you too?

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/weighting/custom/ClassHelper.java
+++ b/core/src/main/java/com/graphhopper/routing/weighting/custom/ClassHelper.java
@@ -1,5 +1,5 @@
 package com.graphhopper.routing.weighting.custom;
 
 interface ClassHelper {
-    String getPackageName(String encVal);
+    String getClassName(String encVal);
 }
--- a/core/src/main/java/com/graphhopper/routing/weighting/custom/ConditionalExpressionVisitor.java
+++ b/core/src/main/java/com/graphhopper/routing/weighting/custom/ConditionalExpressionVisitor.java
@@ -124,7 +124,7 @@ class ConditionalExpressionVisitor implements Visitor.AtomVisitor<Boolean, Excep
                     if (variableValidator.isValid(lhVarAsString) && Helper.toUpperCase(rhValueAsString).equals(rhValueAsString)) {
                         if (!eqOps)
                             throw new IllegalArgumentException("Operator " + binOp.operator + " not allowed for enum");
-                        String value = classHelper.getPackageName(binOp.lhs.toString());
+                        String value = classHelper.getClassName(binOp.lhs.toString());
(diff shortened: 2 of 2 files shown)


lukasalexanderweber:
@karussell i just tried it and it works. thanks for the fix :)

PR #2940, opened 2024-02-12
Allow routing cars in pedestrian streets

michaelkrog:
This PR add support for importing pedestrian streets in the car profile - allowing routing cars in pedestrian streets that is tagged with vehicles allowed in pedestrian streets.

Because https://github.com/graphhopper/graphhopper/issues/733 and https://github.com/graphhopper/graphhopper/issues/374 are open issues, routes through pedestrian streets might occur when not legal. When they are fixed this should work fine.

Adding specific/correct legal_default_speeds for pedestrian streets will ensure that routing will most often be routed using other ways.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,4 +36,3 @@ jobs:
             ${{ runner.os}}-node_modules-
       - name: Build ${{ matrix.java-version }}
         run: mvn -B clean test
-
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,3 +36,4 @@ jobs:
             ${{ runner.os}}-node_modules-
       - name: Build ${{ matrix.java-version }}
         run: mvn -B clean test
+
--- a/.github/workflows/remove-old-artifacts.yml
+++ b/.github/workflows/remove-old-artifacts.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
 jobs:
   remove-old-artifacts:
-    if: github.repository_owner == 'graphhopper'
+    if: github.repository_owner == 'Apaq'
     name: Remove old artifacts
     runs-on: ubuntu-latest
     timeout-minutes: 10 # stop the task if it takes longer
@@ -19,7 +19,7 @@ jobs:
         skip-recent: 300
 
   remove-old-artifacts-2:
-    if: github.repository_owner == 'graphhopper'
+    if: github.repository_owner == 'Apaq'
     name: Remove old artifacts
     runs-on: ubuntu-latest
     timeout-minutes: 10 # stop the task if it takes longer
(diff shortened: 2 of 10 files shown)


michaelkrog:
@karussell Yeah. It seems to give me the same results as I had with my PR (but this is obviously improved only allow those explicit tagged)

Issue that remains as stated above is that routing through pedestrian streets in city centers that is tagged correctly (fx. motor_vehicle=destination; motor_vehicle:conditional=no @ (11:00-04:00)) may be routed via shortcuts in the pedestrian streets instead of around the pedestrian streets.


## Example of wrongful short cut route across city center using the changes in this PR
Skærmbillede 2024-03-08 kl  11 46 34

## Example of correct route across city center using production version
Skærmbillede 2024-03-08 kl  11 54 44


## Example of correct routing into pedestrian streets using the changes in this PR
Skærmbillede 2024-03-08 kl  11 55 42

## Example of wrongful routing into pedestrian streets using the changes in this PR
Skærmbillede 2024-03-08 kl  11 56 02

otbutz:
@michaelkrog not sure if I can follow. So in the branch created by @karussell, the accessibility of the ways is evaluated correctly, but the routing incorrectly favors them instead of the faster detour?

Maybe the average speed estimation is a bit off? https://github.com/graphhopper/graphhopper/commit/5ae8297257a117542aeef8c541967d4954ee0687 only touches the access bits but doesn't introduce a default speed value. The fallback of 10km/h might be to high? Should we treat it like a `living_street` and assume 5km/h aka walking speed?

michaelkrog:
@otbutz Yes and no. 

As I understand it, [graphhopper does not support motor_vehicle=destination for car](https://github.com/graphhopper/graphhopper/issues/733). In the shown examples, the pedestrian streets are tagged with motor_vehicle=destination; maxspeed=15, and the branch by @karussell still uses the pedestrian streets to cross the city center. The optimal would be to only route using the pedestrian streets when the destination is in a pedestrian street.

Changing the default speed for pedestrian streets (as you mention) may help it to use normal highways when destination is not in the pedestrian streets. I did that in my own PR(this PR), but as I recall it still wasn't perfect.  But maybe that is because the streets are tagged with maxspeed=15 and then overrides the default walking speed?

otbutz:
> But maybe that is because the pedestrians streets in this example are tagged with maxspeed=15 and then overrides the default walking speed?

An explicitly tagged value would override the default. But I'd still aim for a rather pessimistic assumption if no max_speed tag is present, just to be sure.

> As I understand it, [graphhopper does not support motor_vehicle=destination for car](https://github.com/graphhopper/graphhopper/issues/733).

We take the road access into account as far as I'm aware:

https://github.com/graphhopper/graphhopper/blob/6905ff4575726714ee2ec04f63ea326897f1b9bf/core/src/main/java/com/graphhopper/routing/weighting/FastestWeighting.java#L90-L97

Could you check if the parser yields `DESTINATION` for the relevant sections?

karussell:
This was implemented in #3203 (i.e. only accessible for cars where explicitly allowed)

PR #2950, opened 2024-03-14
higher default precision for way geometry

karussell:
I noticed a difference between our routes and the map tiles on several occasions especially for certain curves which could also result in unexpected instructions. ~Decreasing the default maxWayPointDistance fixes this.~

I'm unsure but is a change to the visited_nodes expected?


easbar:
Did you notice a difference between our routes and the rendered OSM tiles, or between the routes and our /mvt layer?

> I'm unsure but is a change to the visited_nodes expected?

Not really?! Du you mean the change in GraphHopperTest?

PR #2954, opened 2024-03-22
Initial simple turn lane support

### LINKED ISSUE #1131: Instructions: parse turn:lanes to improve instructions

karussell:
Fixes #1131 and replaces #1269. These turn lane details help with the instructions to e.g. announce turns earlier.

Every instruction can contain multiple lane lists. For example one lane list 500m before the turn, then the lane list changes because the right turn was done (or the "right" and "through" lane separated) and then 200m before the turn there is another lane list.

To enable turn lane support for a certain profile like `car` the following configuration is required:

```
    datareader.turn_lanes_profiles: car
```

Screenshot from a [prototype client](https://github.com/graphhopper/graphhopper-maps/pull/279) that can consume this data (`turn_lanes` branch).

image

Work to do:

 - [x] GraphHopper does not pick any of the lanes as `valid` although it should pick the right lane (`continue`). Lane info is `left|continue`. GraphHopper detects a "slight right" turn although it should use a "continue", but turn angle is too big for a "continue" due to the lane merge. So also related to #2953. Example link: https://graphhopper.com/maps/?point=52.528168%2C13.402536&point=52.52994%2C13.401086&profile=car -> this is likely unrelated to the turn lane support and the instructions should be fixed instead (i.e. no instructions). But this is tricky as different road names and road classes.
 - [ ]

<details>

minor improvement for OSMReader regarding "none":

```java
    static String replaceLanes(String entireStr) {
        if (entireStr == null || entireStr.isEmpty()) return "";
        String[] strs = entireStr.split("\\|");
        StringBuilder out = new StringBuilder(entireStr.length());
        for (int i = 0; i < strs.length; i++) {
            String str = strs[i].trim();
            if (str.isEmpty()) {
                // empty lane like ||right into none|none|right
                out.append("none");
            } else {
                // naming must be consistent with our instructions
                out.append(str.replace("through", "continue"));
            }
            if (i < strs.length - 1)
                out.append('|');
        }
        return out.toString();
    }
```

</details>

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
+++ b/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
@@ -258,7 +258,7 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
 
         } else {
             InstructionsOutgoingEdges outdoingEdges = new InstructionsOutgoingEdges(prevEdge, edge, weighting, maxSpeedEnc,
-                    roadClassEnc, roadClassLinkEnc, allExplorer, nodeAccess, prevNode, baseNode, adjNode);
+                    roadClassEnc, roadClassLinkEnc, lanesEnc, allExplorer, nodeAccess, prevNode, baseNode, adjNode);
             prevOrientation = AngleCalc.ANGLE_CALC.calcOrientation(doublePrevLat, doublePrevLon, prevLat, prevLon);
             int sign = edge.getEdge() == prevEdge.getEdge()
                     ? Instruction.U_TURN_UNKNOWN // this is the simplest turn to recognize, a plain u-turn.
@@ -459,7 +459,7 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
 
         // there is no other turn possible
         if (nrOfPossibleTurns <= 1) {
-            if (Math.abs(sign) > 1 && outgoingEdges.getVisibleTurns() > 1 && !outgoingEdges.mergedOrSplitWay(lanesEnc)) {
+            if (Math.abs(sign) > 1 && outgoingEdges.getVisibleTurns() > 1 && !outgoingEdges.mergedOrSplitWay()) {
                 // This is an actual turn because |sign| > 1
                 // There could be some confusion, if we would not create a turn instruction, even though it is the only
                 // possible turn, also see #1048
@@ -474,7 +474,7 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
             // Don't show an instruction if the user is following a street, even though the street is
             // bending. We should only do this, if following the street is the obvious choice.
             if (InstructionsHelper.isSameName(name, prevName) && outgoingEdges.outgoingEdgesAreSlowerByFactor(2)
-                    || outgoingEdges.mergedOrSplitWay(lanesEnc)) {
+                    || outgoingEdges.mergedOrSplitWay()) {
                 return Instruction.IGNORE;
             }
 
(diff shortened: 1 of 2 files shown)


hactar:
@karussell nice! I don't see any changes to https://github.com/graphhopper/graphhopper/tree/c64febe4522984a61b2d80dcf0d78135c19bfea5/navigation - am I correct to assume this will not improve the navigation module? If not, any plans to update that too? ☺️

hactar:
Thanks for the info - I had assumed that the navigate endpoint would need some "conversion" so that it matches that format, glad to hear its in. I wish I could help review but alas Java has been a while - looking forward to seeing this in Graphhopper, and happy to test this against the Maplibre Navigation SDKs once its in and we can deploy it. Thanks for the contribution!

hactar:
I'd need to set up the whole toolchain, best someone else tackles this 😅. I'll reach out to our backend team and see if they can help. For now at least linking some documentation. While this is documentation for the newest API version, I quickly checked and it appears to still match the older format used by MapLibre Navigation SDK:

https://docs.mapbox.com/api/navigation/directions/#lane-object

Some sample code of a route step containing lane info:

```json
{
  "intersections": [
    {
      "out": 1,
      "location": [13.424671, 52.508812],
      "bearings": [120, 210, 300],
      "entry": [false, true, true],
      "in": 0,
      "stop_sign": true,
      "lanes": [
        {
          "valid": true,
          "active": true,
          "valid_indication": "straight",
          "indications": ["left", "straight"]
        },
        {
          "valid": true,
          "active": false,
          "valid_indication": "straight",
          "indications": ["straight", "right"]
        },
        {
          "valid": false,
          "active": false,
          "indications": ["right"]
        }
      ]
    }
  ],
  "geometry": "asn_Ie_}pAdKxG",
  "maneuver": {
    "bearing_after": 202,
    "type": "turn",
    "modifier": "left",
    "bearing_before": 299,
    "location": [13.424671, 52.508812],
    "instruction": "Turn left onto Adalbertstraße"
  },
  "duration": 59.1,
  "distance": 236.9,
  "driving_side": "right",
  "weight": 59.1,
  "name": "Adalbertstraße",
  "mode": "driving"
}
```

hactar:
You're right, there's more to it. I went through the iOS navigation code, and the mentioned lane info for banners comes from where you say it does (https://github.com/flitsmeister/mapbox-directions-swift/blob/6c19ecc4e1324887ae3250802b8d13d8d8b3ff2d/MapboxDirections/MBVisualInstruction.swift#L73)

Here's a test file from the MapLibre Navigation iOS SDK,  it contains lane info on both the level that I described but also at the level you described - should show how it should look like.
 
https://github.com/maplibre/maplibre-navigation-ios/blob/main/MapboxCoreNavigationTests/A12-To-Veenendaal-Bigger-Detour.json

IAbderrahmane:
hi @karussell, I want to add lane support to the /navigate response. I'll start with adding it to the bannerInstructions but I have some questions:

1. Instruction.instructionDetails is an array, if the size is one I can just add the lane information to the current bannerInstruction but what if it contains more than one element ? let's say we have to InstructionDetails and both of them contain lane information. Do I create 2 bannerInstructions ? they both contain the same primary attribute but different sub attributes. the first one is the closest lane information (based on beforeTurn maybe ?) and the second one is the farthest.  
2. in the MapBox spec there's an additional field active_direction here's the definition: _When components.active is true, this property shows which of the lane's components.directions applies to the current route, when there is more than one_  . Where should I put this information ? I found to possible locations one is LaneInfo class the other one is NavigateResponseConverter.putSingleBannerInstruction. I'm leaning towards the second one since it's needed only in the navigate endpoint.

Issue #2937, opened 2024-02-07
max_slope should be able to get negative


karussell:
Currently we only store the absolute value but this reduces information and it is unknown if the slope is an include or decline.

(to restore the old behaviour we could think to support `abs(max_slope)` in the custom model)

jessLryan:
@karussell are you working on this? If not I'd like to work on it

PR #2955, opened 2024-03-23
Issue #2937 max slope should be able to get negative

jessLryan:
max slope is now stored using DecimalEncodedValueImpl and stores both positive and negative values as appropriate
test added and all passing
name added to contributors file


jessLryan:
r.e. the custom model, no I haven't tried that yet will do
r.e. the POM, I couldn't get the required depencies without these changes

jessLryan:
@karussell can you clarify what you mean with the custom model?

PR #2958, opened 2024-03-25
fix Arrival Polyline

OlafFlebbeBosch:
The polyline for the arrival part should include the geometry up to the arrival point.


boldtrn (approved):
Thanks for looking into this. You are right, the point should be included for the last step before the arrival step and not in the arrival step itself.

PR #2967, opened 2024-04-11
Avoid String concatenation in loops

otbutz:


otbutz:
We might as well switch the whole debugInfo fields to `List<String>` in GHResponse/ResponsePath and delay the concatenation until `GHResponse.getDebugInfo()`

PR #2968, opened 2024-04-11
Add missing setup call

otbutz:
While DijkstraOneToMany evaluates the timeout, it's missing the setup.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/DijkstraOneToMany.java
+++ b/core/src/main/java/com/graphhopper/routing/DijkstraOneToMany.java
@@ -67,6 +67,7 @@ public class DijkstraOneToMany extends AbstractRoutingAlgorithm {
 
     @Override
     public Path calcPath(int from, int to) {
+        setupFinishTime();
         fromNode = from;
         endNode = findEndNode(from, to);
         if (endNode < 0 || isWeightLimitExceeded()) {


otbutz:
Currently we're sort of doing both? The visited nodes are reset on each call:

https://github.com/graphhopper/graphhopper/blob/6f56fc37dd73417c3c788bbf0cab0f6892f5b928/core/src/main/java/com/graphhopper/routing/DijkstraOneToMany.java#L150

IMHO we should reset all limits per run. Enforcing them over multiple passes should be up to the caller.

easbar:
I agree that the time limit should rather be used per call, because there might be unrelated things running in between the `calcPath` call that likely should be excluded from the time measurement. For the visited nodes and the maximum weight I think the global limits make sense, because the current weight and visited nodes can only increase from call to call anyway. Or do you have an actual use case for per-call limited weights and nodes?

otbutz:
I'm currently using DijkstraOneToMany to compute a distance matrix. While CH-based routing would be an order of magnitude faster, it's unfortunately not an option for my use case, which involves more dynamic custom models.

Global limits would have to scale with the number of destinations for which the algorithm instance is to be reused. The problem is that some destinations might simply not be reachable, and I don't want GraphHopper touching the whole graph to come to that conclusion. In the event that a limit is reached while calculating the path for a destination, I would still want to continue with the remaining destinations, which is not possible with a global limit in place.

easbar:
> In the event that a limit is reached while calculating the path for a destination, I would still want to continue with the remaining destinations, which is not possible with a global limit in place.

You mean if you are trying to calculate a path to node A which is only found at, say, weight=1000, and it isn't found because the limit is 100 for example you aren't able to find node B at weight 50? That shouldn't be of course. I think the way it should work is:

1) start the search to some node, which grows the Dijkstra tree until it is found or the weight limit is reached
2) search for another node. If it was discovered already return it, otherwise keep growing the tree (up to the weight limit if necessary)

This algorithm is basically left over from a previous CH preparation algorithm, so it is quite possible that it doesn't fit more general usages.

otbutz:
> search for another node. If it was discovered already return it, otherwise keep growing the tree (up to the weight limit if necessary)

As long as paths for already discovered nodes are still returned even when the algorithm has reached one of its limits, it would be fine.

easbar:
Ok, but you are saying this is currently not the case? Anyway as far as I am concerned we can merge this PR.

otbutz:
I think the current state is fine.

PR #2975, opened 2024-04-16
fix: support Barrier Nodes for NavigateResponse

OlafFlebbeBosch:
Supplying a route with Intersections located on the same spot, connected with a zero length edge seems to confuse Mapbox like APIs. This can happen if there are barriers on the route, which are split into two intersections connected by a zero length edge. The idea is to find these intersections and merge all the edges of both intersections and remove the out outgoing edge of the connecting edge from the first and the in edge of the incoming edge for the connection.

Patch is rather large unfortunately because of reformatting based on editorconfig.


OlafFlebbeBosch:
Converted to DRAFT. There seem to be errors due to rebasing

OlafFlebbeBosch:
Fixed rebaseing issue

OlafFlebbeBosch:
The newly introduced unittest barrierTest in NavigateResponseConverterTest is triggering the issue while
commenting out lines 197 to end of (excluding return) of filterIntersectionDetails  NavigateResponseConverter .

However I have to admit that I have issues to reproduce the issue ad hoc on routes like https://graphhopper.com/maps/?point=55.720661%2C13.213068&point=55.720375%2C13.213033&profile=bike&layer=Omniscale
where mapbox previously tripped over the kerbs https://www.openstreetmap.org/node/10233057887 and https://www.openstreetmap.org/node/10233057888  .  Please let me investigate this again .

OlafFlebbeBosch:
Was tripping over my own feet :) 

The issue can be seen in sweden:

https://graphhopper.com/api/1/navigate/directions/v5/gh/bike/13.2130268,55.7206509;13.2130345,55.7204650?geometries=polyline6&steps=true&roundabout_exits=true&voice_instructions=true&banner_instructions=true&key=xxx

The 2nd/3rd  and 4th/5th are located on the same point . The PR fixes this.

boldtrn:
Thanks a lot for providing this PR @OlafFlebbeBosch. I can also reproduce the issue at different places on GraphHopper.com 👍.

IMHO we should not have two intersections at the same coordinate with incorrect bearings, so this fix makes sense to me. While I also think that clients should not crash due to that, it is good if we can fix this issue here as well :).

core/src/main/java/com/graphhopper/util/details/IntersectionValues.java
```diff
@@ -0,0 +1,83 @@
...
+    public static List<IntersectionValues> createList(Map<String, Object> intersectionMap) {
+        List<IntersectionValues> list = new ArrayList<>();
+
+        List<Integer> bearings = (List<Integer>) intersectionMap.get("bearings");
+        Integer in = (Integer) intersectionMap.get("in");
+        Integer out = (Integer) intersectionMap.get("out");
+        List<Boolean> entry = (List<Boolean>) intersectionMap.get("entries");
+
+        if (bearings.size() != entry.size()) {
+
```
boldtrn, line 44: empty line

```suggestion
```

core/src/main/java/com/graphhopper/util/details/IntersectionValues.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public static List<IntersectionValues> createList(Map<String, Object> intersectionMap) {
+        List<IntersectionValues> list = new ArrayList<>();
+
+        List<Integer> bearings = (List<Integer>) intersectionMap.get("bearings");
+        Integer in = (Integer) intersectionMap.get("in");
+        Integer out = (Integer) intersectionMap.get("out");
+        List<Boolean> entry = (List<Boolean>) intersectionMap.get("entries");
+
+        if (bearings.size() != entry.size()) {
```
boldtrn, line 43: I am not sure about this case, so this is an error case, right? Should we return silent? Or should we at least log the error or even throw an IllegalStateException? This should never happen right?

boldtrn (commented):
Could you clarify the error case where the sizes don't match? Otherwise this change looks good to me 👍

core/src/main/java/com/graphhopper/util/details/IntersectionValues.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * create a PathDetail from a List of IntersectionValues
+     */
+    public static Map<String, Object> createIntersection(List<IntersectionValues> list) {
+        Map<String, Object> intersection = new HashMap<>();
+
+        intersection.put("bearings",
+                list.stream().map(x -> x.bearing).collect(Collectors.toList()));
+        intersection.put("entries",
+                list.stream().map(x -> x.entry).collect(Collectors.toList()));
```
otbutz, line 70: Not that it matters that much, but the old implementation is more efficient. This adds two more list iterations.

core/src/main/java/com/graphhopper/util/details/IntersectionValues.java
OlafFlebbeBosch: AFAIK this should never happen. Not sure what is an appropriate reaction on that. I wasn't aware that IllegalStateException is a RuntimeException, so we do not need to attribute each throw , which is good.
But I can create an Logger instance and Log it as well... Please advise me here.

boldtrn: Yes, this should result in a HTTP request with response code 400, as Dropwizard catches this. Since we believe this should never happen, I'd say let's throw an exception at least (and maybe even log the result).

OlafFlebbeBosch:
> Could you clarify the error case where the sizes don't match? Otherwise this change looks good to me 👍

Throwing now an IllegalstateException for this "should not happen" case

boldtrn (approved):
Thanks the PR looks good to me 👍

PR #2977, opened 2024-04-17
New Highway-Tag based Encoded Values

TheDropZone:
Provide a first pass at adding in `highway` tag based osm data to the options of encoded data. The goal is to provide traffic signal, stop signs, and traffic crossings back in the routing requests.

The encoded data (tags) being targeted are references in the highway-tag wiki page: [https://wiki.openstreetmap.org/wiki/Key:highway](highway-tag) under the "Other highway features".
Currently, the tags that are being included are
- TRAFFIC_SIGNALS
- STOP
- CROSSING

Additionally, this provides a channel to add on additional tags under the `highway` key using this encoded_data tag.
```
graph.encoded_values: highway_feature
```
This also provides the ability to request those highway features (currently traffic signals, stop signs, and crossing) from the web api via the `details` parameter

---

I attempted to model this after other encoded_value tags and parsers, but am very happy to refactor or adjust as needed. If this data is better loaded using other methods, I am glad to refactor as well. Open to any critiques as this is my first contribution here. Thanks for the time!

---
**Note:**
- added username and blurb to CONTRIBUTORS.md file

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -86,6 +86,7 @@ Here is an overview:
  * thehereward, code cleanups like #620
  * vvikas, ideas for many to many improvements and #616
  * zstadler, multiple fixes and car4wd
+ * TheDropZone, added new optional encoded values
 
 ## Translations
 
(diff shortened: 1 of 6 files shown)


TheDropZone:
@karussell Hey! From playing around with the `Crossing` value, I found it got close but 2 issues
1. It didn't seem to include Stop Signs, though I could be totally wrong there
2. In true encoded-value fashion, it grouped values together, seemingly removing a lot of information about where the actual crossing was on the edge.

Currently, I'm trying to set up better tests to really exercise this issue. The ultimate goal is to provide route consumers with the knowledge of where vehicle stop points are on the returned route.

PR #2983, opened 2024-04-22
Approximate with DistancePlaneProjection

### LINKED ISSUE #2973: LMApproximator overhead
The overhead of the LMApproximator using DistanceCalcEarth for its BeelineWeightApproximator seems to be quite substantial:

grafik

Would it be possible to use DistancePlaneProjection instead?

otbutz:
Fixes #2973


otbutz:
We could switch to a two-pass calculation based on the result of the plane distance. But I'm not sure we want that complexity.

Issue #1688, opened 2019-08-18
Make EdgeAccess#INT_DIST_FACTOR confirgurabel


boldtrn:
Currently, the max distance of an edge is roughly 1000km. There are cases where this is not enough, especially when using a different datasource than OSM. 

Would it make sense to make the INT_DIST_FACTOR configurable? 

Related to #1663.

easbar:
Would it not suffice to set `INT_DIST_FACTOR` to  `10` ? In this case the minimum distance that is not considered to be zero would be increased from `0.5mm` to `5cm` (should not be relevant ?) and the maximum distance would be increased from about `2147km` to `214748km`. The problem with changing the value lies mostly in adjusting the failing tests (I guess 99% rounding issues and/or hard coded expected distances that need to be adjusted), but this would be rather annoying if the conversion factor would not be fixed.

boldtrn:
This would fix at least my scenario perfectly @easbar :+1:

boldtrn:
Should I provide a PR? Could there be other downsides? Probably there was a reason why the factor is so big?

easbar:
> Could there be other downsides?

Likely there will be some more rounding issues, and we probably have to identify the places where the (rounded) distances are compared to the precise ones ? For example for AStar we might be comparing the (rounded) edge-distance with the precise result of `calcDist(lat, lon)`. For CH we are storing the sum of several (rounded) edge-distances as (rounded) distance of the shortcut. For QueryGraph there can be a mismatch between the edge distance and the sum of the distances of the virtual edges. These issues already exist, but might get worse for a lower distance precision ? I am not sure to be honest. Maybe we need to try fixing the tests and see what complications arise.

Ultimately the best solution might be to use `0.1m` as distance unit and integer values all across GH (already during data import, `calcDist(lat,lon)` etc.) and only convert it to meters where the result is shown to the user (?). I guess it is the 'do not use floating point numbers where exact answers are required' type of problem.

karussell:
> Ultimately the best solution might be to use 0.1m as distance unit and integer values all across GH 

This one I do not understand as rounding issues can sum up even with integers (?)

boldtrn:
> For example for AStar we might be comparing the (rounded) edge-distance with the precise result of calcDist(lat, lon)

Could we round the results from calcDist? So that we are comparing two rounded distances?

easbar:
> This one I do not understand as rounding issues can sum up even with integers (?)

[...]

Issue #1663, opened 2019-07-14
Store distances as integers


easbar:
Using 0.1m as distance unit should be sufficiently precise also for indoor routing and the maximum distance would still be sufficiently large: 214748km so more than five times the circumference of the earth. IMO using integers is more comfortable and prevents rounding / comparison issues. 

Related https://github.com/graphhopper/graphhopper/issues/435

boldtrn:
I think that's a good idea! If I understand this correct, this would mean changing the INT_DISTANCE_FACTOR in EdgeAccess to 10?

easbar:
Yes changing the distance factor, but also I think distances should be stored as ints and any int/double conversion should happen somewhere in the top layer (osm import / user api), so internally all distances are ints.

boldtrn:
Ah, now I understand it. Now I am not so sure if I think this is a good idea :). This change could require significant changes for everyone using GraphHopper as a dependency and would make using it feel a bit cumbersome.

For example many existing FlagEncoders oder custom weightings could break due to that.

What improvements do you expect from this change? Better memory efficiency of 4 vs 8 bytes and better performance by avoiding the conversion from int to double?

If that's the case, it might make sense to have an internal method like getIntDistance and the regular getDistance? So we could use the int on the low level where performance is of the essence and use the convenient version where performance is not that critical?

easbar:
I think my vision would be to interpret the 4 bytes we are using for distance in `EdgeAccess` at the moment as integer number that tells us the distance as multiple of `0.1m`. I think the main advantage of 
using int over double/float is that floating point numbers are never exact, and we have to include error margins etc. when comparing them. 

At the same time I fully agree that having `EdgeIteratorState#getDistance()` return the distance in multiples of `0.1m` would be very ugly, because it would result in x10 errors in existing code without giving a compiler error or similar.

So maybe having an internal method `int getIntDistancel()` for the edge iterator as you suggest could be a good solution, yes. 

> Better memory efficiency of 4 vs 8 bytes

Yes, in a few cases this might make a difference, not with storage, because we are already using only 4 bytes, but sometimes we keep many distances in memory.

> better performance by avoiding the conversion from int to double?

Hmm, I do not think the conversion is critical for performance, but haven't really tried either.

> and would make using it feel a bit cumbersome.

[...]

PR #2986, opened 2024-04-24
Replace int-based access for edges with bytes-based access

karussell:
Currently we are only able to read integers from `edges`, which is a bit limiting as everything has to fit in one integer and if not, it will require two integers (instead of expanding to just 5 bytes).

So, with this change we can use only 5 bytes for the georef (see #2978) and we do not have to use int-based edge flags too, i.e. potential less space for this either. I.e. this change can reduce the storage requirement by 6 bytes (e.g. if flags use 5 or 9 bytes).

There is only a minor slowdown for routingCH_full, probably because of the change of the georef to bytes. The other minor slowdown for routingCH was compensated via removing the new instance of EdgeIntAccess for every EdgeIteratorState (independent of this PR).

Later this change will help if other edge properties like the edge IDs or link ID grow beyond 4 bytes.

TODO:

 - [x] getByte/Short additionally to getInt to speed up retrieval for values with only a few bits -> tricky. Do later as it requires rewrite of InitializerConfig and IntEncodedValueImpl to a byte-based approach.
 - [x] avoid padding via implementing padding in DataAccess.set/getInt/Short
 - [x] later: use bytes for `nodes` too, then we could reduce 4 to probably just 1 byte for the turn restriction information
 - [x] later: distance could be 5 bytes now, see #1688 and #1663.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/BaseGraphNodesAndEdges.java
+++ b/core/src/main/java/com/graphhopper/storage/BaseGraphNodesAndEdges.java
@@ -19,7 +19,10 @@
 package com.graphhopper.storage;
 
 import com.graphhopper.routing.ev.EdgeBytesAccess;
-import com.graphhopper.util.*;
+import com.graphhopper.util.Constants;
+import com.graphhopper.util.EdgeIterator;
+import com.graphhopper.util.GHUtility;
+import com.graphhopper.util.Helper;
 import com.graphhopper.util.shapes.BBox;
 
 import java.util.Locale;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/graphhopper/storage/TurnCostStorage.java
```diff
@@ -224,14 +238,14 @@ public int getToEdge() {
...
-            intsRef.ints[0] = turnCosts.getInt(turnCostPtr() + TC_FLAGS);
-            return booleanEncodedValue.getBool(false, -1, edgeIntAccess);
+            edgeAccess.setInt(0, 0, turnCosts.getInt(turnCostPtr() + TC_FLAGS));
+            return booleanEncodedValue.getBool(false, 0, edgeAccess);
         }
 
         @Override
         public double getCost(DecimalEncodedValue encodedValue) {
-            intsRef.ints[0] = turnCosts.getInt(turnCostPtr() + TC_FLAGS);
-            return encodedValue.getDecimal(false, -1, edgeIntAccess);
```
easbar, line 234: I thought using -1 made it a bit more clear that we are passing an unused value? In the code above we still use -1

core/src/main/java/com/graphhopper/storage/BytesRef.java
```diff
@@ -108,16 +109,16 @@ public boolean intsEquals(IntsRef other) {
      * Signed int order comparison
      */
     @Override
-    public int compareTo(IntsRef other) {
+    public int compareTo(BytesRef other) {
```
easbar, line 112: (Why) do we need to implement Comparable here?

easbar (commented):
Thanks, looks good!

PR #2988, opened 2024-04-25
Use 9 bits to store max_weight

otbutz:
Fixes #2986

* raises the size of MaxWeight EV to 9 bits, which is enough to encode `maxweight` values up to 51t.
* ~~re-enable logging of capped values which should (hopefully) occur less often~~ still the case


otbutz:
sigh...referenced the wrong issue

Issue #2987, opened 2024-04-25
Default MaxWeight resolution is insufficient for truck routing


otbutz:
**Describe the bug**

This way has a `maxweight` tag with a value of `44`: https://www.openstreetmap.org/way/4281359

Our current default values limit the maximum storeable weight to `0.1 * 2^8` which is around ~25,6t:

https://github.com/graphhopper/graphhopper/blob/ff8f2b23a478bc17eecd409a46a35e4677e75172/core/src/main/java/com/graphhopper/routing/ev/MaxWeight.java#L30-L34

The problem is that we are limiting every value we find to this maximum. This causes the marked limit of 44 to be silently changed to `25.6` in the graph.

The result is quite surprising:

![image](https://github.com/graphhopper/graphhopper/assets/22315436/c5be2064-2292-4100-b86c-f2faab067881)

![image](https://github.com/graphhopper/graphhopper/assets/22315436/d02a45db-1269-4083-99bf-de255b86454c)

**To Reproduce**

* Route limited to 26t: https://graphhopper.com/maps/?point=52.907683%2C-1.853742&point=52.907408%2C-1.851745&profile=car&layer=Omniscale&custom_model=%7B%22distance_influence%22%3A15%2C%22priority%22%3A%5B%7B%22if%22%3A%22max_weight%3C%3D26%22%2C%22multiply_by%22%3A%220.0%22%7D%5D%2C%22speed%22%3A%5B%5D%2C%22areas%22%3A%7B%22type%22%3A%22FeatureCollection%22%2C%22features%22%3A%5B%5D%7D%7D
* Route limited to 25t: https://graphhopper.com/maps/?point=52.907683%2C-1.853742&point=52.907408%2C-1.851745&profile=car&layer=Omniscale&custom_model=%7B%22distance_influence%22%3A15%2C%22priority%22%3A%5B%7B%22if%22%3A%22max_weight%3C%3D25%22%2C%22multiply_by%22%3A%220.0%22%7D%5D%2C%22speed%22%3A%5B%5D%2C%22areas%22%3A%7B%22type%22%3A%22FeatureCollection%22%2C%22features%22%3A%5B%5D%7D%7D

Instead of a manual reproduction it would be even better to have a unit or integration test reproducing the issue.

**Expected behavior**

Either we should raise the default values to cover existing limits or we skip values between maximum and infinity.

otbutz:
[...]

PR #2990, opened 2024-04-25
Encode values larger than maxStorableValue as infinity

otbutz:


otbutz:
It's only conservative as long as the maximum value we allow to be encoded exceeds what's relevant for routing decisions. #2987 is an example where this strategy introduces artificial weight limits that are far lower than what is actually tagged in OSM, while the encoded value is still in the range of values relevant to truck routing. Searching for a route with a 26t truck in my case.

Ideally, the EV should be able to cover all "sane" values (see #2989), so that we can ignore values that are either tagging errors like kilograms instead of tonnes, or are far too high for our intended use case.

For example, this [way](https://www.openstreetmap.org/way/27152424) has a maximum weight limit of 120t. In practice, this is so high that it's irrelevant for any normal vehicle. But by limiting it to the maximum storage value of the EV, we create an artificial limit within the range of values relevant to large trucks.

PR #2993, opened 2024-04-29
Use byte-based storage for edges

karussell:
Replaces #2986 but with less changes.

Currently we store all properties like the distance or nodeA/B ID to the `edges` DataAccess as integer, which is a bit limiting as every property has to fit in one integer and if not, it will require two integers (instead of expanding to e.g. 5 bytes).

With this change the `edges` is byte-based and we can use only 5 bytes for the georef (see #2978) and the edge flags are stored with byte-precision too i.e. this change can reduce the storage requirement by up to 6 bytes (e.g. up to 3 bytes for the edge flags plus 3 bytes less for georef).

There is only a minor slowdown of 2-5% for routing and CH/LM preparation. As there was a [minor speed-up](https://github.com/graphhopper/graphhopper/commit/57408524ad9ffe68cf827e9c55f835516a0537eb) found in this process the overall slowdown regarding the latest stable for the routing is neglectable.

Later this change will help if other edge properties like the edge IDs or link ID grow beyond 4 bytes.

### DIFF
--- a/core/src/main/java/com/graphhopper/util/BitUtil.java
+++ b/core/src/main/java/com/graphhopper/util/BitUtil.java
@@ -125,7 +125,7 @@ public class BitUtil {
     }
 
     /**
-     * Note, value can be negative and no exceptions is thrown as it'll be interpreted as an unsigned number.
+     * Note, currently value with higher bits set (like for a negative value) won't throw an exception at this level.
      */
     public final void fromUInt3(byte[] bytes, int value, int offset) {
         bytes[offset + 2] = (byte) (value >>> 16);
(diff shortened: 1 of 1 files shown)


core/src/test/java/com/graphhopper/routing/util/parsers/FootTagParserTest.java
```diff
@@ -63,7 +61,7 @@ public FootTagParserTest() {
 
     @Test
     public void testGetSpeed() {
-        EdgeIntAccess edgeIntAccess = new ArrayEdgeIntAccess(encodingManager.getIntsForFlags());
+        EdgeIntAccess edgeIntAccess = new ArrayEdgeIntAccess(encodingManager.getBytesForFlags() / 4 + 1);
```
easbar, line 64: Given how frequently we seem to use this we could just keep `getIntsForFlags` next to `getBytesForFlags` for such cases where need to count full ints?

easbar: Ok, I just thought it wouldn't be wrong to ask the encoding manager how many ints it needs even though it uses bytes internally. 

> We could create a TestArrayEdgeIntAccess 

Or a static creator using bytes for ArrayEdgeIntAccess.

PR #2994, opened 2024-04-29
use Map instead of List for KVStorage API

karussell:


core/src/main/java/com/graphhopper/search/KVStorage.java
```diff
@@ -541,4 +554,8 @@ public static String cutString(String value) {
         // See #2609 and test why we use a value < 255
         return bytes.length > 250 ? new String(bytes, 0, 250, Helper.UTF_CS) : value;
     }
+
+    public static Map<String, KValue> createKV(String key, Object value) {
+        return Collections.singletonMap(key, new KValue(value));
```
otbutz, line 559: Since we're on Java 17, we might as well use the new static factory methods:

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Map.html#unmodifiable

PR #3005, opened 2024-05-12
Minor speed up using MethodHandles in DataAccess

karussell:
Speedup is not more than 5% for e.g. routingCH.

This lower level approach is also used in DataInput/OutputStream so it should be production ready.

Tried also with "theUnsafe" directly, but it wasn't faster.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/RAMDataAccess.java
+++ b/core/src/main/java/com/graphhopper/storage/RAMDataAccess.java
@@ -36,8 +36,8 @@ public class RAMDataAccess extends AbstractDataAccess {
     private byte[][] segments = new byte[0][];
     private boolean store;
     // we could also use UNSAFE but it is not really faster (see #3005)
-    private static final VarHandle INT = MethodHandles.byteArrayViewVarHandle(int[].class, ByteOrder.LITTLE_ENDIAN);
-    private static final VarHandle SHORT = MethodHandles.byteArrayViewVarHandle(short[].class, ByteOrder.LITTLE_ENDIAN);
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/graphhopper/storage/RAMDataAccess.java
```diff
@@ -32,6 +35,9 @@
 public class RAMDataAccess extends AbstractDataAccess {
     private byte[][] segments = new byte[0][];
     private boolean store;
+    // we could also use UNSAFE but it is not really faster (see #3005)
+    private static final VarHandle INT = MethodHandles.byteArrayViewVarHandle(int[].class, ByteOrder.LITTLE_ENDIAN);
+    private static final VarHandle SHORT = MethodHandles.byteArrayViewVarHandle(short[].class, ByteOrder.LITTLE_ENDIAN);
```
otbutz, line 40:
```suggestion
    private static final VarHandle INT = MethodHandles.byteArrayViewVarHandle(int[].class, ByteOrder.LITTLE_ENDIAN).withInvokeExactBehavior();
    private static final VarHandle SHORT = MethodHandles.byteArrayViewVarHandle(short[].class, ByteOrder.LITTLE_ENDIAN).withInvokeExactBehavior();
```

https://bugs.openjdk.org/browse/JDK-8255375

otbutz (changes requested):

core/src/main/java/com/graphhopper/storage/RAMDataAccess.java
otbutz: The VarHandle has some built-in logic to handle casts. This has some overhead and tends to be a silent performance problem. The `withInvokeExactBehavior()` disables the casts and instead throws an error if there's a type mismatch.

e.g. without the exact behavior this would work, but with degraded performance:

```java
short value = 5;
INT.set(segments[bufferIndex], index, value);
```

PR #3007, opened 2024-05-14
Fix constant path details like leg_distance for very short routes with duplicate points

easbar:
We get this error: https://github.com/graphhopper/graphhopper/blob/34b333aadf20f104f6bc1d6add3900f75805eb77/core/src/main/java/com/graphhopper/util/PathSimplification.java#L315 for routes with three points where the last two are identical. But only if the points are so close to each other that the resulting path has <=2 points. The root cause is that we do not set the interval boundaries for empty paths in ConstantDetailsBuilder correctly, but this bug is 'fixed' by the path simplification **unless** the simplification is disabled and **if** the route has more than two points (otherwise we return early here: https://github.com/graphhopper/graphhopper/blob/34b333aadf20f104f6bc1d6add3900f75805eb77/core/src/main/java/com/graphhopper/util/PathSimplification.java#L183-L185 so the 'fix' does not happen).

Here I fixed this by correctly setting the interval boundaries in ConstantDetailsBuilder for empty paths. c.f. the special handling we introduced for empty paths here: #2924

### DIFF
--- a/core/src/main/java/com/graphhopper/util/details/PathDetailsFromEdges.java
+++ b/core/src/main/java/com/graphhopper/util/details/PathDetailsFromEdges.java
@@ -72,11 +72,6 @@ public class PathDetailsFromEdges implements Path.EdgeVisitor {
         Map<String, List<PathDetail>> pathDetails = new HashMap<>(pathBuilders.size());
         for (PathDetailsBuilder builder : pathBuilders) {
             Map.Entry<String, List<PathDetail>> entry = builder.build();
-//            if (entry.getValue().get(0).getFirst() != previousIndex)
-//                throw new IllegalStateException("Invalid path detail entry: " + entry + ", previousIndex: " + previousIndex);
-//            for (int i = 1; i < entry.getValue().size(); i++)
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/graphhopper/util/details/AbstractPathDetailsBuilder.java
```diff
@@ -77,7 +77,7 @@ public void endInterval(int lastIndex) {
     }
 
     public Map.Entry<String, List<PathDetail>> build() {
-        return new MapEntry(getName(), pathDetails);
+        return new MapEntry<>(getName(), pathDetails);
```
otbutz, line 80: We might also switch to using `Map.entry(getName(), pathDetails)` if we don't need mutability.

otbutz: It's using a value-based class behind the scenes with all the advantages that come with it: https://www.baeldung.com/java-value-based-classes

> this is unrelated to the PR.

Yep. I have a bit of a bad habit of pointing at Java 9+ goodies when I see them.

Issue #3014, opened 2024-06-09
bike should preffer more footways with bicycle=yes


caspg:
I noticed that bike profile is preffering `tertiary` highway instead of footway where bicycle traffic is allowed (`bicycle=yes`). In my case, it leads to strange route. 

Should we lower speed that much if there is `bicycle=yes` tag in the `footway`?

```java
         } else if (way.hasTag("bicycle", "designated") || way.hasTag("bicycle", "official") ||
                    way.hasTag("segregated", "yes") || way.hasTag("bicycle", "yes")) {
                // Here we handle the cases where the OSM tagging results in something similar to "highway=cycleway"
                if (way.hasTag("segregated", "yes"))
                    speed = highwaySpeeds.get("cycleway");
                else
                    speed = way.hasTag("bicycle", "yes") ? 10 : highwaySpeeds.get("cycleway");
```

https://graphhopper.com/maps/?point=54.404198%2C18.624522&point=54.406793%2C18.626314&profile=bike&layer=OpenStreetMap

![CleanShot 2024-06-09 at 12 17 29@2x](https://github.com/graphhopper/graphhopper/assets/10419439/7829f6d7-a2d3-4e6d-b86a-b4b3ee476e92)

After changing speed from 10 to "cycleway" speed (18) it routed better:
![CleanShot 2024-06-09 at 12 17 09@2x](https://github.com/graphhopper/graphhopper/assets/10419439/72d36e29-f391-4621-bdd3-570600a83a9b)

ratrun:
As long as there is no `segregated` tag I think that the speed penalty of  8km/h is reasonable compared to the speed you can go on the parallel `tertiary` road. We need to assume pedestrians and therefore one needs to slow down.

caspg:
I get that but routing in this case is clearly not ideal. Valhalla and Komoot are using footway in my example.

Can I somehow prioritize that footway with bicycle=yes?

ratrun:
> Can I somehow prioritize that footway with bicycle=yes?

We are already prioritizing `bicycle=yes` with the value PREFER, see `BikeCommonPriorityParser`

https://github.com/graphhopper/graphhopper/blob/5eb09e78a7216b7a436f6f592e1c6ee1bc4c9d6f/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java?plain=1#L200-L201

But we are prioritizing `tertiary` also as PREFER, see
[...]

PR #3015, opened 2024-06-10
[DRAFT] remove tertiary highways from preferred and increase speed on footways

caspg:
Right now, GH preffers `tertiary` roads over footways with `bicycle=yes` tags for bike profile. 
This leads to not opimial routing, especially when those footways are connecting cycleways.

@ratrun  suggested that we should remove `tertiary` from preffered highways and bump speed in the `footway` (only with `bicycle=yes`) https://github.com/graphhopper/graphhopper/issues/3014

Here are two examples before and after the change:

[GH map](https://graphhopper.com/maps/?point=54.404478%2C18.62471&point=54.406794%2C18.626314&profile=bike&layer=OpenStreetMap)

Before:
CleanShot 2024-06-10 at 09 35 28@2x

After:
CleanShot 2024-06-10 at 09 35 08@2x

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
@@ -201,7 +201,7 @@ public abstract class BikeCommonAverageSpeedParser extends AbstractAverageSpeedP
                 if (way.hasTag("segregated", "yes"))
                     speed = highwaySpeeds.get("cycleway");
                 else
-                    speed = way.hasTag("bicycle", "yes") ? 10 : highwaySpeeds.get("cycleway");
+                    speed = way.hasTag("bicycle", "yes") ? 12 : highwaySpeeds.get("cycleway");
 
                 // valid surface speed?
                 if (surfaceSpeed > 0)
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikePriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikePriorityParser.java
@@ -18,8 +18,6 @@ public class BikePriorityParser extends BikeCommonPriorityParser {
         addPushingSection("path");
 
         preferHighwayTags.add("service");
-        preferHighwayTags.add("tertiary");
-        preferHighwayTags.add("tertiary_link");
         preferHighwayTags.add("residential");
         preferHighwayTags.add("unclassified");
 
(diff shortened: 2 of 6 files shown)


caspg:
I updated tests but not sure what to do with this one: https://github.com/graphhopper/graphhopper/blob/master/core/src/test/java/com/graphhopper/GraphHopperTest.java#L464

We probably need to find new example location?

caspg:
@karussell They look exactly the same. I run it with:

```yaml
profiles:
    - name: bike
      custom_model_files: [bike.json]
```


Master branch:
master

My branch:
zmiany

caspg:
Yup, removing `tertiary` only from MTB and leaving in RacingBike seems reasonable. Will update PR.

caspg:
@ratrun @karussell How should I change test? Also, in test it returns 1 route (probably without any alternative).
Earlier screenshots comes from running default `bike.json` custom model.

https://github.com/graphhopper/graphhopper/blob/master/core/src/test/java/com/graphhopper/GraphHopperTest.java#L464

```
[INFO]
[ERROR] Failures:
[ERROR]   GraphHopperTest.testAlternativeRoutesBike:464 expected: <3> but was: <1>
```

caspg:
@karussell How should my `bike.json` custom model look to match the test configuration? I run it with default `bike.json` before.

```yaml
profiles:
    - name: bike
      custom_model_files: [bike.json]
```

ratrun:
@caspg I could not find out what you are doing differently, but I cannot reproduce your screenshots about the alternatives. Here is my screenshot for the original code. It matches the test with 3 alternatives:
Bildschirmfoto vom 2024-06-16 07-32-59

I'm using the checked in `bike.json` from git.

With your branch there is no alternative found as indicated by the test, but by slightly changing the destination coordinates into `49.982089,11.599224` the three alternatives are found again. 

Therefore my suggestion is  to change the test into:
```
        GHRequest req = new GHRequest(50.028917, 11.496506, 49.982089,11.599224).
                setAlgorithm(ALT_ROUTE).setProfile(profile);
        req.putHint("alternative_route.max_paths", 3);
        GHResponse rsp = hopper.route(req);
        assertFalse(rsp.hasErrors(), rsp.getErrors().toString());

        assertEquals(3, rsp.getAll().size());
        // via ramsenthal
        assertEquals(2697, rsp.getAll().get(0).getTime() / 1000);
        // via unterwaiz
        assertEquals(2985, rsp.getAll().get(1).getTime() / 1000);
        // via eselslohe -> theta; BTW: here smaller time as 2nd alternative due to priority influences time order
        assertEquals(2783, rsp.getAll().get(2).getTime() / 1000);

```

caspg:
@ratrun Thanks! I've just updated the test and PR is ready for review.

PR #3028, opened 2024-06-21
clean up bike speed parser

karussell:
Currently the bike speed parsing is too complicated, especially the handling of the "pushing speed". I adapted the code while keeping all unit tests. This made it clear that e.g. the default speed for path and footway was incorrectly set IMO which I fixed.

I stumbled over this unclear code while I wanted to fix [this bug](https://graphhopper.com/maps/?point=37.771237%2C-122.454767&point=37.770946%2C-122.456489&profile=bike) (see new test case).

image

@ratrun @caspg please have a look if this is an improvement for you too.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/AbstractAccessParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/AbstractAccessParser.java
@@ -17,8 +17,7 @@ public abstract class AbstractAccessParser implements TagParser {
     protected final List<String> restrictionKeys = new ArrayList<>(5);
     protected final Set<String> restrictedValues = new HashSet<>(5);
 
-    protected final Set<String> intendedValues = new HashSet<>(INTENDED);
-    protected final Set<String> oneways = new HashSet<>(ONEWAYS);
+    protected final Set<String> intendedValues = new HashSet<>(INTENDED); // possible to add "private" later
     // http://wiki.openstreetmap.org/wiki/Mapfeatures#Barrier
     protected final Set<String> barriers = new HashSet<>(5);
     protected final BooleanEncodedValue accessEnc;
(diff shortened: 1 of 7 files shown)


core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -677,4 +678,23 @@ public void temporalAccess() {
...
+
+        way.setTag("bicycle", "yes");
+        way.setTag("surface", "asphalt");
+        way.setTag("highway", "pedestrian");
+        assertPriorityAndSpeed(PREFER, 12, way);
+
+        way.setTag("cycleway:right", "track");
+        way.setTag("highway", "pedestrian");
+        assertPriorityAndSpeed(PREFER, 18, way);
+    }
```
ratrun, line 699: I find following easier to read:
```
    @Test
    public void testPedestrian() {
        way.setTag("highway", "pedestrian");
        assertPriorityAndSpeed(SLIGHT_AVOID, 4, way);
        way.setTag("bicycle", "yes");
        assertPriorityAndSpeed(PREFER, 12, way);
        way.setTag("surface", "asphalt");
        assertPriorityAndSpeed(PREFER, 12, way);

        way.clearTags();
        way.setTag("highway", "pedestrian");
        way.setTag("cycleway:right", "track");
        way.setTag("bicycle", "yes");
        assertPriorityAndSpeed(PREFER, 18, way);
    }
```
The last check currently requires the `bicycle=yes` tag to be present, otherwise the priority would result in SLIGHT_AVOID. It would be nice if we could classify this also as PREFER also without `bicycle=yes` as this tag is kind of redundant for this edge case in combination with `cycleway:right`

ratrun (approved):
The implementation code looks indeed much cleaner now!

caspg:
I just tested this. Didn't see any improvements (because I think didn't encounter bugs like this)  but works as expected 👍

ratrun (approved):

PR #3030, opened 2024-06-25
Handle turn restrictions with multiple via ways and overlapping restrictions

easbar:
This PR extends the turn restriction support to restrictions with multiple and/or overlapping via-edges. These include OSM turn relations with multiple via-ways, turn relations with a single via-way that are crossed by other ways (such that they must be split into multiple edges for routing) and turn relations that could not be handled so far because they share a common via-way.

Fixes ~~#446~~ #3155
Closes #2695

Here are some examples:

* overlapping via-way restrictions, with via-ways pointing in the same restriction (ignored so far): [rel1](https://www.openstreetmap.org/relation/8634880), [rel2](https://www.openstreetmap.org/relation/8634881), [route](https://graphhopper.com/maps/?point=52.171998%2C9.105377&point=52.171386%2C9.104561&profile=car&layer=OpenStreetMap)

* overlapping no-straight-on and no-u-turn, via-ways pointing in the same direction (ignored so far): [rel1](https://www.openstreetmap.org/relation/3662288), [rel2](https://www.openstreetmap.org/relation/3662296), [route](https://graphhopper.com/maps/?point=53.47316%2C10.176372&point=53.473167%2C10.17617&profile=car&layer=OpenStreetMap)

* two via-ways (because via-way is split into two for some reason): [rel](https://www.openstreetmap.org/relation/16103380) [route](https://graphhopper.com/maps/?point=50.997619%2C11.026714&point=50.997677%2C11.026261&profile=car&layer=OpenStreetMap)

* straight-ahead-only with two via-ways: [rel](https://www.openstreetmap.org/relation/14932796) [route](https://graphhopper.com/maps/?point=53.558515%2C10.037019&point=53.558152%2C10.039284&profile=car&layer=OpenStreetMap)
* straight-ahead-only with three via-ways: [rel](https://www.openstreetmap.org/relation/13716158) [route](https://graphhopper.com/maps/?point=53.563401%2C10.017506&point=53.563138%2C10.01553&profile=car&layer=OpenStreetMap)

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/FootPriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/FootPriorityParser.java
@@ -104,6 +104,10 @@ public class FootPriorityParser implements TagParser {
         if (way.hasTag("foot", "designated"))
             weightToPrioMap.put(100d, PREFER);
 
+        if (way.hasTag("foot", "use_sidepath")) {
+            weightToPrioMap.put(100d, VERY_BAD); // see #3035
+        }
+
         double maxSpeed = Math.max(getMaxSpeed(way, false), getMaxSpeed(way, true));
         if (safeHighwayTags.contains(highway) || (isValidSpeed(maxSpeed) && maxSpeed <= 20)) {
             weightToPrioMap.put(40d, PREFER);
--- a/core/src/main/java/com/graphhopper/storage/TurnCostStorage.java
+++ b/core/src/main/java/com/graphhopper/storage/TurnCostStorage.java
@@ -141,7 +141,7 @@ public class TurnCostStorage {
     }
 
     private void ensureTurnCostIndex(int nodeIndex) {
-        turnCosts.ensureCapacity(((long) nodeIndex + 4) * BYTES_PER_ENTRY);
+        turnCosts.ensureCapacity(((long) nodeIndex + 1) * BYTES_PER_ENTRY);
     }
 
     private long findPointer(int fromEdge, int viaNode, int toEdge) {
--- a/core/src/test/java/com/graphhopper/routing/util/parsers/FootTagParserTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/parsers/FootTagParserTest.java
@@ -333,7 +333,11 @@ public class FootTagParserTest {
         way.clearTags();
         way.setTag("highway", "tertiary");
         assertEquals(PriorityCode.UNCHANGED.getValue(), prioParser.handlePriority(way, null));
+        way.setTag("foot","use_sidepath");
+        assertEquals(PriorityCode.VERY_BAD.getValue(), prioParser.handlePriority(way, null));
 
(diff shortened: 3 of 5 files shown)


matkoniecz:
> By testing I mean checking out this branch, running an import

is there somewhere step-to-step guide? (AFAIK I have never compiled graphhopper)

Looked at https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md and it seems to have no such steps nor link to them

https://github.com/graphhopper/graphhopper?tab=readme-ov-file#installation has for installation of released code

(sorry if I missed it, I done quick check so far)

lukasalexanderweber:
I had a quick look over the changes as notices that `RestrictionSetter#addArtificialEdges` was removed. It seems you have a new strategy to cover via way turn restrictions.

> The concept is pretty simple

could you outline the concept in a few words or a sketch? In https://github.com/graphhopper/graphhopper/pull/2695 we thought about multiple artificial edges to support overlapping via way restrictions and multi via way restrictions. Is the approach now something completely different or close to _multiple artificial edges_?

[Here](https://github.com/graphhopper/graphhopper/pull/2695/commits/73c62b1b3832aa6fc9fc75b8dd10ba46fe2ded29) is the failing test I included to demonstrate the problem with multi via way restriction. Have you used this to test your new implementation?

> I've added another test to show when this approach does not work:
> 
> ```
>       a   b   c   d
>     0---1---2---3---4
>         |e  |f
>         5   6
> ```
> 
> with the restrictions `e via b to f` and `a via b via c to d`.
> 
> when turning right from e, we'll be detoured on b'. However, we cannot reach `d` due to `b' via 2 to c` and `c' via 3 to d` of the second restriction. We **should** actually be able to reach `d` from `e`.
> 
> On the other hand, we can not reach `f` from `a` since `b' via 2 to f` is restricted by the first restriction.
> 
> **SUMMARY**
> 
> The presented approach will fail if a via way of a multi-via-way-restriction is already _in use_ by another via-way-restriction. I tested that this is the case in `115 of 150` multi-via-way-restrictions in germany.

lukasalexanderweber:
> More precisely, there is now one artificial edge for every different incoming edge that is part of a restriction. This means there can be several paths leading to an edge that encode the different possible origins of the path.

```
      a   b   c   d
    0---1---2---3---4
        |e  |f
        5   6
```

so previously there was only one artificial `b'` which was used by `e via b to f ` and `a via b via c to d`. What exactly is now the situation?

> I added a few tests, but not this one exactly. But if you don't mind I will pick some of your tests and explanations to this branch.

Of cause feel free to take anything you need from the previous PR :)

lukasalexanderweber:
okay great and if there would be another restriction

```
      a   b   c   d
    0---1---2---3---4
        |e  |f
        5   6
        |g  
        7   
```

`g via e via b to c` then for this restriction there would be an artificial `e'` and then there would be three artificial b's (coming from `a`, from `e` and `e'`)?

PR #3035, opened 2024-07-06
De-prioritize ways tagged with foot=use_sidepath

ratrun:
Triggered by the discussion in the [OSM community forum](https://community.openstreetmap.org/t/foot-use-sidepath-sinnvoll-bei-separat-kartierten-gehwegen/115654) I discovered that graphhopper does not yet have support for `foot=use_sidepath`. This PR adds it.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/FootPriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/FootPriorityParser.java
@@ -105,6 +105,8 @@ public class FootPriorityParser implements TagParser {
             weightToPrioMap.put(100d, PREFER);
 
         if (way.hasTag("foot", "use_sidepath")) {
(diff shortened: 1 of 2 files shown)


ratrun:
> Thanks! Can we just use REACH_DESTINATION as we do for bikes?

I expected this question :smile: ! Of course it would be possible. But when you read the discussion on the OSM community forum as graphhopper currently does not have any support for this tag/value combination, graphhopper does not produce the discussed long detours yet. It will return the detours with the change.  I believe that REACH_DESTINATION is too restrictive for foot. When the detour is too long you jump on the street. It is subjective. If you really want me to change it to REACH_DESTINATION I will.

ratrun:
I just added the requested reference in the code in [40d970d](https://github.com/graphhopper/graphhopper/pull/3035/commits/40d970d1957a4753cdaaec67ac66917af2d0969b)

PR #3037, opened 2024-08-02
Update Dropwizard version to 4.0.7

ratrun:
As we now depend on Java 17 I updated the used dropwizard dependency to the current version 4.0.7.

The upgrade documentation used for the upgrade can be found here:

https://www.dropwizard.io/en/release-3.0.x/manual/upgrade-notes/upgrade-notes-3_0_x.html
https://www.dropwizard.io/en/stable/manual/upgrade-notes/upgrade-notes-4_0_x.html

The upgrade to version 3.0.7 was pretty straightforward. 
In the 4.0.x upgrade documentation there is no description about the need for implementing the MessageBodyReader class. I found this out by studying the debug logs when using the mapmatching endpoint.


ratrun:
I didn't see that you already upgraded to version 3.0 ! But I also didn't notice any strange behavior with my version. Is it possible to trigger a build action in order to see if the same problem occurs also on this branch?

ratrun:
I was able to reproduce the hang problem with the Dropwizard3Test.java test on version 4.0.7. As there is no response from the dropwizard developers a solution would be to add the line 
```
app.client().property(Apache5ClientProperties.REUSE_STRATEGY, (ConnectionReuseStrategy) (request, response, context) -> false);
```
to our TestUtils.java. Shouldn't we do simply that instead of waiting for a response?

joschi:
@ratrun @karussell Could you please retry with the latest release (Dropwizard 4.0.8) which contains Jersey 3.0.15?

That hopefully will address the issue.

ratrun:
> We updated Dropwizard to version 3 here: #2920 Is there any good reason we should use version 4?

According the [upgrade notes]( https://www.dropwizard.io/en/stable/manual/upgrade-notes/upgrade-notes-4_0_x.html) I expect dropwizard 4.0.x will not include additional features that are not included in Dropwizard 3.0.x, but other then that currently there is no difference for us.

PR #3042, opened 2024-08-21
feat: Treat separate sidewalks like foot=use_sidepath

aritozen:
Separately mapping sidewalks becomes more and more common. In many jurisdictions, available sidewalks must be used, so route via the separate sidewalks by default. See also [this post on a related OSM discussion](https://community.openstreetmap.org/t/foot-use-sidepath-sinnvoll-bei-separat-kartierten-gehwegen/115654/9).

Previously, `sidewalk=use_sidepath` was implemented in #3035. This PR is similar in that it adds the same behaviour when encountering `sidewalk*=separate`. Since there are many variations, I've added a bunch of test cases.
In particular, if one side is separately mapped, but there is still a sidewalk included in the way, use the usual priority.

Several tags are in use related to sidewalks (sidewalk:both, sidewalk:left, sidewalk:right) with yes/no/separate as options.
If only one side is specified (like sidewalk:left without sidewalk:right), default behaviour is maintained.
Best practice is to define both or sidewalk:both.
The tag sidewalk=* can additionally define left/right/both indicating which side has or hasn't a sidewalk and that it is part of the way.

If either side is separate and the other not part of the way, don't route via the way unless the alternatives are impractical.


core/src/main/java/com/graphhopper/routing/util/parsers/FootPriorityParser.java
```diff
@@ -104,8 +104,20 @@ void collect(ReaderWay way, TreeMap<Double, PriorityCode> weightToPrioMap) {
         if (way.hasTag("foot", "designated"))
             weightToPrioMap.put(100d, PREFER);
 
-        if (way.hasTag("foot", "use_sidepath")) {
-            weightToPrioMap.put(100d, VERY_BAD); // see #3035
+        if (way.hasTag("foot", "use_sidepath") || way.hasTag("sidewalk", "separate") || way.hasTag("sidewalk:both", "separate")) {
```
ratrun, line 107: As all the added if statements below lead to the same result I would prefer to have a singe if statement with a bunch of logical operations instead.

ratrun (commented):
I think that the proposed changes make sense

aritozen:
@ratrun Thanks for your review. I've added a protected method `use_sidepath`, similar to [isOneway in CarAccessParser](https://github.com/aritozen/graphhopper/blob/add_sidewalk_separate_handling/core/src/main/java/com/graphhopper/routing/util/parsers/CarAccessParser.java#L173) to aid readability.

core/src/main/java/com/graphhopper/routing/util/parsers/FootPriorityParser.java
```diff
@@ -129,4 +129,12 @@ void collect(ReaderWay way, TreeMap<Double, PriorityCode> weightToPrioMap) {
         if (way.hasTag("bicycle", "official") || way.hasTag("bicycle", "designated"))
             weightToPrioMap.put(44d, SLIGHT_AVOID);
     }
+
+    protected boolean use_sidepath(ReaderWay way) {
+        return way.hasTag("foot", "use_sidepath")
+                || way.hasTag("sidewalk", "separate")
+                || way.hasTag("sidewalk:both", "separate")
+                || way.hasTag("sidewalk:left", "separate") && way.hasTag("sidewalk:right", sidewalksNoValues)
+                || way.hasTag("sidewalk:right", "separate") && way.hasTag("sidewalk:left", sidewalksNoValues);
```
otbutz, line 138:
```suggestion
                || (way.hasTag("sidewalk:left", "separate") && way.hasTag("sidewalk:right", sidewalksNoValues))
                || (way.hasTag("sidewalk:right", "separate") && way.hasTag("sidewalk:left", sidewalksNoValues));
```

otbutz: The `&&` has precedence over `||`, but I'd still add the parenthesis because it makes it easier to spot.

aritozen:
@otbutz Thanks for reviewing, I've added the parenthesis.

aritozen:
I've added the changelog entry. Is something blocking this PR from getting merged?

aritozen:
> One question regarding the case if only one side has a sidewalk available. Should we force pedestrians across a street in that case? Do you have real world cases for this situation?

At least in Germany, one must use a sidewalk if available, even if it is on the other side. Usually, it's a good idea anyway.
Take this way: https://www.openstreetmap.org/way/44130504. It has a sidewalk on the right, while the left consists of perpendicular parking (not shown in OSM, but visible on aerial imagery; I passed by there recently too). There is no path of any kind on the left side, so it's only sensible to use the right sidewalk.
I'd expect the result to look something like this: https://graphhopper.com/maps/?point=50.099099%2C8.760847&point=50.098919%2C8.761913&point=50.099192%2C8.763605&point=50.09933%2C8.763978&point=50.099302%2C8.764308&point=50.09977%2C8.765526&profile=foot&layer=OpenStreetMap

> Would it make sense to make this encoded value dependent on the direction (currently forward and backward values are the same) 

It would make some sense for bicycle access, mostly the situation with `USE_SIDEPATH`. If say the right cycleway is mapped separately and the left one isn't (or doesn't exist), then bicycles going forward along the way have `USE_SIDEPATH`, while bicycles going in the other direction have `YES` or `DESIGNATED`. Though we'd probably need to figure out whether that separate cycleway is two-way or oneway...?

Searching on Overpass, I found `access:forward` ~300 times, `bicycle:forward` ~6500 times, and `foot:forward` ~330 times. `bicycle:forward` seems useful, the other two are probably mapping mistakes/inconsistencies that are better handled with `oneway`.

> [...] and consider sidewalk*=separate in the parser there?

Would make things more consistent I think? It could use the same code from me here. Though access `YES` isn't a problem if the priority still considers the sidewalk as better. Maybe that's something you can better decide.

ratrun:
@karussell To me it looks complete now and can be merged. Or is something missing?

aritozen:
> Currently we have foot=use_sidepath but also several places where we use way.hasTag("sidewalk", sidewalksNoValues) in FootPriorityParser. Shouldn't we try to make this more consistent and avoid foot=use_sidepath where we avoid sidewalk=separate?

Yes, I think that would be beneficial. Both are effectively the same, only difference being that in one case the sidepath isn't considered a sidewalk (due to distance, quality, or other factors), which isn't relevant for access.

zdila:
Some data in case it helps unstick this. We hit exactly this on a production instance (GH 11.0) and traced it here.

**The ignored spelling is now the more common one.** From taginfo's Europe instance:

| tag | ways |
| --- | --- |
| `sidewalk=separate` | 242,042 |
| `sidewalk:both=separate` | **272,512** |
| `sidewalk:left=*` | 495,158 |
| `sidewalk:right=*` | 511,705 |

`FootPriorityParser` reads only the plain `sidewalk` key, so most ways that say "the sidewalks are mapped separately" are not merely ignored — they are promoted. Because the tag isn't found in `sidewalksNoValues`, the way takes the `else` branch and gets `priorityCode.better().better()`, i.e. two steps *up* the ladder.

**A measured case.** Repašského in Bratislava, `highway=tertiary` with `sidewalk:both=separate`:

- `foot_priority` measures **1.0** (AVOID 0.8 → SLIGHT_AVOID → UNCHANGED). With `sidewalk=separate` it would be 0.8.
- A `foot` route from `48.189755/17.032993` to `48.190370/17.036051` leaves the footway network and runs along the carriageway, in a junction that has nine crossings and twenty-nine footways mapped. Forcing the same route onto footways costs 9 m.

So the effect is the opposite of the intent: the better-tagged the road is about its separate sidewalks, the more attractive it becomes to walk on.

**One thing that changed since this PR was opened:** #3305 merged a `Sidewalk` encoded value that already parses `sidewalk`, `sidewalk:left`, `sidewalk:right` and `sidewalk:both`, and models left/right/both separately. That may be a natural home for the distinction discussed above, and could let the change here shrink to consuming an existing value rather than re-parsing the tags.

Still reproducible on current master (`d9506cd7d`): lines 89, 96 and 100 of `FootPriorityParser` are unchanged since the 11.0 tag.

PR #3046, opened 2024-08-23
Use surface speed at bridgleways for bicycle profiles

ratrun:
This should improve the bicycle routing on `highway=bridleway`, see the discussion in the [graphhopper forum](https://discuss.graphhopper.com/t/bike-routing-ignoring-designated-cycle-routes/8987/2).

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
@@ -160,7 +160,8 @@ public abstract class BikeCommonAverageSpeedParser extends AbstractAverageSpeedP
             speed = PUSHING_SECTION_SPEED; // unknown surface
         } else if (way.hasTag("service")) {
             speed = highwaySpeeds.get("living_street");
-        } else if ("track".equals(highwayValue)) {
+        } else if ("track".equals(highwayValue) ||
+                   "bridleway".equals(highwayValue) ) {
             if (surfaceSpeed != null)
                 speed = surfaceSpeed;
             else if (trackTypeSpeeds.containsKey(trackTypeValue))
--- a/core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
@@ -305,6 +305,14 @@ public class BikeTagParserTest extends AbstractBikeTagParserTester {
         assertPriorityAndSpeed(VERY_NICE, PUSHING_SECTION_SPEED, way);
         way.setTag("bicycle", "yes");
         assertPriorityAndSpeed(VERY_NICE, 18, way);
+
+        way.clearTags();
+        way.setTag("highway", "bridleway");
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
```diff
@@ -160,7 +160,8 @@ else if (way.hasTag("bicycle", "yes"))
             speed = PUSHING_SECTION_SPEED; // unknown surface
         } else if (way.hasTag("service")) {
             speed = highwaySpeeds.get("living_street");
-        } else if ("track".equals(highwayValue)) {
+        } else if ("track".equals(highwayValue) ||
+                   "bridleway".equals(highwayValue) ) {
```
otbutz, line 164:
```suggestion
        } else if ("bridleway".equals(highwayValue) && surfaceSpeed != null ) {
              speed = surfaceSpeed;
        } else if ("track".equals(highwayValue) {
```

Otherwise, we might take into account a track type that was mistakenly assigned to the bridleway.

otbutz (changes requested):

core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
ratrun: Good catch! Thank you!

ratrun: I was wrong, your suggested changes do not compile. And this test already passes :
```
        way.clearTags();
        way.setTag("highway", "bridleway");
        way.setTag("surface", "gravel");
        way.setTag("bicycle", "designated");
        way.setTag("tracktype", "grade5");
        assertPriorityAndSpeed(PREFER, 12, way);
```
@otbutz What did you mean?

otbutz: Could you test the following?

```java
        way.clearTags();
        way.setTag("highway", "bridleway");
        way.setTag("bicycle", "designated");
        way.setTag("tracktype", "grade5");
        assertPriorityAndSpeed(PREFER, 12, way);
```

ratrun: > Could you test the following?
> 
> ```java
>         way.clearTags();
>         way.setTag("highway", "bridleway");
>         way.setTag("bicycle", "designated");
>         way.setTag("tracktype", "grade5");
>         assertPriorityAndSpeed(PREFER, 12, way);
> ```

This test fails as the speed result is 4km/h, which is the assigned trackTypeSpeed for 'grade5'. This result is fine from my perspective.

otbutz: I was under the impression that this tag shouldn't be paired with a bridleway, so I went the extra mile and fired up Overpass Turbo. As it turns out, tagging the tracktype for a bridleway is legit and actually quite common:

grafik

PR #3055, opened 2024-10-05
Add German footway limits for `highway=footway` and `highway=pedestrian`

Nadyita:
In Germany, when a pedestrian street, or a footway is open to vehicles, conditionally or not, these must not drive faster than walking speed. This applies to bicycles and cars alike.

There doesn't seem to be a test I could possibly add for this, as it is purely a declaration.


Nadyita:
I changed it in the wiki now. Didn't notice it's auto created, sorry.

Issue #2460, opened 2021-11-17
Encode road access per transportation mode


otbutz:
Our current implementation only stores the applicable RoadAccess for cars:

https://github.com/graphhopper/graphhopper/blob/91afeb76674710d28a3744c332e98e8861b1e7b1/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java#L35-L37

There is already an open issue regarding access handling(#1835) but it seems to be more focused on the FlagEncoder/Weighting perspective. Before we can tackle this, we'd need to figure out how we want to store parsed access restrictions for multiple transportation modes in the graph.

Without this we can't properly express truck routing via a CustomModel.

Related: #223

easbar:
>  we'd need to figure out how we want to store parsed access restrictions for multiple transportation modes in the graph.

We already store access flags for each vehicle. Do you need other 'access' values than `true/false` (like destination, private, etc.), or what do you mean with `parsed access restrictions`?

otbutz:
> We already store access flags for each vehicle.

The more fine-grained logic is only used within the FlagEncoders:

https://github.com/graphhopper/graphhopper/blob/1e5f4e74f8365bf322bd3ab001afb78e638198e1/core/src/main/java/com/graphhopper/routing/util/AbstractFlagEncoder.java#L86

The RoadAccess EV only represents the access for `TransportationMode.CAR`:

https://github.com/graphhopper/graphhopper/blob/1e5f4e74f8365bf322bd3ab001afb78e638198e1/core/src/main/java/com/graphhopper/routing/util/EncodingManager.java#L312-L315

easbar:
> The more fine-grained logic is only used within the FlagEncoders:
> The RoadAccess EV only represents the access for TransportationMode.CAR:

Yes, but which 'RoadAccess' values do you think should exist for which transportation mode? Currently we have `YES,NO,DESTINATION,CUSTOMERS,DELIVERY,FORESTRY,AGRICULTURAL,PRIVATE,OTHER`.

Do we need all of them for e.g. bike and foot? And should something like `CarRoadAccess=YES/NO` be equivalent to `access$car=true/false`? Or should we use an encoded value that just represents the tagged OSM value `DESTINATION,CUSTOMERS,DELIVERY,FORESTRY,AGRICULTURAL,PRIVATE` and `YES/NO` should only go into the current access flags? And do we need things like `access=delivery, but only for truck`?

otbutz:
[...]

PR #3056, opened 2024-10-07
introduced separate foot_road_access and bike_road_access

### LINKED ISSUE #2819: Enum for foot, bike and psv access tags
Create `foot_road_access` and `bike_road_access` encoded value enums for access tags like we already have for motor vehicles (road_access or hgv). Otherwise we cannot properly move from parsers to custom models. See #2820.

- fix cargo_bike.json
- fix hike.json
- psv_road_access enum too: https://github.com/graphhopper/graphhopper/pull/2710
- finally remove block private options #2780 (had to be reverted due to a similar problem)
- rename hgv to hgv_road_access? (discussion #2744)
- duplicate https://github.com/graphhopper/graphhopper/issues/2460 ?

karussell:
This should fix #2819 / #2460.

An additional MISSING value is required to support the difference between YES and "no mapping". Necessary if access defaults are different like for #3067.
 
Follow up:

 * remove the `block_private` parser option, #3076.
 * backward compatible rename of `road_access` to `car_road_access`, #3066.
 * avoid bridleway for foot only in Germany possible, #3067.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ev/DefaultImportRegistry.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/DefaultImportRegistry.java
@@ -19,11 +19,8 @@
 package com.graphhopper.routing.ev;
 
 import com.graphhopper.routing.util.*;
-import com.graphhopper.routing.util.countryrules.CountryRule;
 import com.graphhopper.routing.util.parsers.*;
 import com.graphhopper.util.PMap;
-
-import java.util.Arrays;
 
 public class DefaultImportRegistry implements ImportRegistry {
     @Override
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/graphhopper/routing/ev/DefaultImportRegistry.java
```diff
@@ -51,11 +54,28 @@ else if (RoadEnvironment.KEY.equals(name))
                     (lookup, props) -> new OSMRoadEnvironmentParser(
                             lookup.getEnumEncodedValue(RoadEnvironment.KEY, RoadEnvironment.class))
             );
+        else if (FootRoadAccess.KEY.equals(name))
+            return ImportUnit.create(name, props -> FootRoadAccess.create(),
+                    (lookup, props) -> new OSMRoadAccessParser<>(
+                            lookup.getEnumEncodedValue(FootRoadAccess.KEY, FootRoadAccess.class),
+                            OSMRoadAccessParser.toOSMRestrictions(TransportationMode.FOOT),
```
otbutz, line 61: Ideally we should be able to pass the TransportationMode directly as explained in the CountryRule related comment.

otbutz: An OSMRoadAccessParser is bound to a particular TransportationMode. Currently this is done indirectly as we pass the resolved restrictions and a country hook which is also TransportationMode dependent (or at least it should/will be).

If we pass the TransportationMode, we can do the resolution inside the parser and call the countryrule directly without a hook.

Issue #2932, opened 2024-01-18
Add crossing encoded values.


natedx:
Crossings have additional keys in OpenStreetMap which indicate a number of features from appearance to accessibility : https://wiki.openstreetmap.org/wiki/Key:crossing

In the [SonarVision](https://www.sonarvision.fr/) app, we have been using the results of PR #2707 for a bit over a year now, allowing us to automatically display crossings to our users ! It's quite amazing ! However, crossings vary a lot in their level of accessibility, making additional crossing keys invaluable for accessible routing.

For example, our blind and visually impaired users typically prefer to be routed through crossings tagged with 
```
crossing=traffic_signals
traffic_signals:sound=yes
tactile_paving=yes
```

I imagine adding these values to Graphhopper implies creating a number of new encoded values. The main ones I believe are important for a wide range of accessible routing use cases are : 

- crossing=traffic_signals/uncontrolled/unmarked/no/informal
- [kerb](https://wiki.openstreetmap.org/wiki/Key:kerb)=yes/no/raised/regular/rolled/lowered/flush/...
- [tactile_paving](https://wiki.openstreetmap.org/wiki/Key:tactile_paving)=yes/(no)/incorrect
- [traffic_signals:sound](https://wiki.openstreetmap.org/wiki/Key:traffic_signals:sound)=yes/no

Finally, adding these encoded values might also help solve the problems raised in issue #2809. Pedestrians have wildly different expectations when it comes to routing. A hurried sighted person might want the shortest path in terms of length, as crossings aren't a problem for them (they might even cross without stopping at red lights...), however a blind pedestrian will want to minimize the overall number of crossings to a certain extent while maximizing the number of crossings equipped with audio traffic signals for obvious safety reasons. A wheelchair user might want to avoid all crossings where kerbs are not lowered or flush.

Thank you for this amazing piece of software !

PR #3058, opened 2024-10-09
Add crossing encoded values for accessible pedestrian routing

CClerger:
See https://github.com/graphhopper/graphhopper/issues/2932

Adding the following `encoded_values`:
- `tactile_paving`: https://wiki.openstreetmap.org/wiki/Key:tactile_paving
- `kerb`: https://wiki.openstreetmap.org/wiki/Key:kerb
- `traffic_signals_sound`: https://wiki.openstreetmap.org/wiki/Key:traffic_signals:sound

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -13,6 +13,7 @@ Here is an overview:
  * baumboi, path detail and landmark improvements
  * boldtrn, one of the core developers with motorcycle knowledge :)
  * bt90, fixes like #2786
+ * CClerger, added encoded values for accessibility
  * cgarreau, increase of routing success rate via subnetwork cleanup
  * chollemans, fixes like #1482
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/graphhopper/routing/ev/DefaultImportRegistry.java
natedx, line 1: Seems to have been accidentally auto-formatted.

natedx (commented):
Great work !
New encoded values behave as expected on a pedestrian graph such as Paris, which has broad coverage for kerbs, traffic_signals_sound and tactile_paving.

Everything looks good to me apart maybe from the DefaultImportRegistry.java file which seems to have been auto-formatted and therefore has a large diff. Perhaps it could be changed back ?

Looking forward to this being merged !

PR #3059, opened 2024-10-10
Allow decimal speeds to be parsed

pantsleftinwash:
Add support for decimal values in speed tags, such as `maxspeed:forward 100.0` found in OS Multi Modal Routing Network

https://docs.os.uk/osngd/os-premium-download-products/premium-products-overview/networks/os-multi-modal-routing-network/technical-specification/features/transport-link#maxspeed-forward


pantsleftinwash:
According to the OSM spec here, then decimals should be valid

https://wiki.openstreetmap.org/wiki/Key:maxspeed#Parser

PR #3062, opened 2024-10-11
fix: Mapbox Arrival and Destinations assumptions

OlafFlebbeBosch:
We learned that there are some implicit assumptions about the structure of the Directions Reponse.
The Start has to have exactly one outgoing "entry" with a corresponding bearing, while the destination needs to have exactly one incoming "entry". 

All test case setup should be done with "intersections" on, since bearings are mandatory for directions.


navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
```diff
@@ -158,6 +159,54 @@ private static void putLegInformation(ObjectNode legJson, ResponsePath path, int
...
+        newFirstIntersection.put( "entries", firstEntries);
+
+        // new first Pathdetail
+        PathDetail firstPathDetail =  new PathDetail( newFirstIntersection);
+        firstPathDetail.setFirst( intersectionDetails.get(0).getFirst());
+        firstPathDetail.setLast( intersectionDetails.get(0).getLast());
+        list.add( firstPathDetail);
+
+        // copy over all elements but the first and last
+        for (int i = 1; i < intersectionDetails.size() - 1; i++) {
```
boldtrn, line 204: I don't fully understand that. Why don't we copy the last element? The method is called `fixFirstIntersectionDetail` after all. We don't add a fixed version of the lat intersectionDetail here as well.

navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
```diff
@@ -260,6 +309,37 @@ private static ObjectNode putInstruction(PointList points, InstructionList instr
...
+            ArrayNode entryArray = intersection.putArray("entry");
+            entryArray.add(true);
+
+            // copy the bearing from the previous instruction
+            ArrayNode bearingsrray = intersection.putArray("bearings");
+            bearingsrray.add( 0);
+
+            // add the in tag
+            intersection.put("in", 0);
+            putLocation(pointList.getLat(0), pointList.getLon(0), intersection);
```
boldtrn, line 340: Shouldn't this also be part of the method `fixFirstIntersectionDetail` and then the method should be renamed? This feels quite unexpected to me.

navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
```diff
@@ -260,6 +309,37 @@ private static ObjectNode putInstruction(PointList points, InstructionList instr
...
+        PointList pointList = instruction.getPoints().clone(false);
+
+        if (instructionIndex == 0) {
+             // Fix first (departure) instruction
+             // to only have one "out" intersection
+            intersectionDetails = fixFirstIntersectionDetail(intersectionDetails);
+        }  else if (instructionIndex < instructions.size() - 1) {
+            // modify pointlist to include the first point of the next instruction
+            // for all instructions but the arrival
+            PointList nextPoints = instructions.get(instructionIndex + 1).getPoints();
```
boldtrn, line 322: We ran this code also for the first instruction before, now this isn't run for the first instruction anymore. Is this desired, and if yes, why?

boldtrn (changes requested):
Thanks for working on this and I agree, we should keep the results as close to the OSRM compatible equivalents as possible. I added a few remarks in the code that I think we should look at before merging this.

navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
OlafFlebbeBosch: moved the fixFirstIntersectionDetail one level up and it is now really doing only the first Intersection

navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
```diff
@@ -158,6 +159,54 @@ private static void putLegInformation(ObjectNode legJson, ResponsePath path, int
...
+    /**
+     * fix the first IntersectionDetail.
+     *
+     * The first Intersection of the first step should only have 1 bearing and one "out" entry
+     *
+     */
+    private static List<PathDetail> fixFirstIntersectionDetail( List<PathDetail> intersectionDetails) {
+        List<PathDetail> list = new ArrayList<>();
+
+        // job1: find out the interesting part of the intersectionDetails
```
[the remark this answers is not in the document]
OlafFlebbeBosch: removed outdfated comment

navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
```diff
@@ -158,6 +159,54 @@ private static void putLegInformation(ObjectNode legJson, ResponsePath path, int
         legJson.put("distance", Helper.round(distance, 1));
     }
 
+
+    /**
+     * fix the first IntersectionDetail.
+     *
+     * The first Intersection of the first step should only have 1 bearing and one "out" entry
```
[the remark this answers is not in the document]
OlafFlebbeBosch: thanks done

OlafFlebbeBosch:
I think I adressed the comments, but the git fork got out of sync and I had to rewrite the PR. Sorry
@boldtrn

boldtrn:
> Are these changes compatible with the Maplibre SDK too?

Yes they are, the change to the response format is pretty small as it only changes parts of the first and last intersection.

PR #3066, opened 2024-10-15
car_road_access instead road_access

karussell:
Use `car_road_access` instead `road_access` everywhere.
Based on #3056. 

Note that for consistency the `car_road_access` must contain `MISSING` but it should not be added to `road_access` as the meaning of YES would be changed, which is important for custom_models.

TODO: maybe better get rid of MISSING (like partially done for `toll`) and set the default access like dismount/yes/no for the different highways (e.g. highway=path/pedestrian/footway is `dismount` or for motorway it is `no` etc).

### DIFF
--- a/core/src/test/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParserTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParserTest.java
@@ -81,4 +81,19 @@ class OSMRoadAccessParserTest {
         parser.handleWayTags(edgeId, edgeIntAccess, way, new IntsRef(1));
         assertEquals(CarRoadAccess.PRIVATE, roadAccessEnc.getEnum(false, edgeId, edgeIntAccess));
     }
+
+    @Test
+    public void testCar() {
+        ArrayEdgeIntAccess edgeIntAccess = new ArrayEdgeIntAccess(1);
+        int edgeId = 0;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/graphhopper/routing/util/countryrules/CountryRule.java
```diff
@@ -19,15 +19,15 @@
...
+import com.graphhopper.routing.ev.CarRoadAccess;
 import com.graphhopper.routing.ev.Toll;
 import com.graphhopper.routing.util.TransportationMode;
 
 /**
  * GraphHopper uses country rules to adjust the routing behavior based on the country an edge is located in
  */
 public interface CountryRule {
-    default RoadAccess getAccess(ReaderWay readerWay, TransportationMode transportationMode, RoadAccess currentRoadAccess) {
+    default CarRoadAccess getAccess(ReaderWay readerWay, TransportationMode transportationMode, CarRoadAccess currentRoadAccess) {
```
otbutz, line 30:
```java
getCarAccess(ReaderWay, CarRoadAccess)
```

Otherwise we would need to come up with a generic approach for the various RoadAccess enums in order to make use of the passed TransportationMode.

otbutz: btw why do we want to split the RoadAccess enum? I understand that some values only make sense for certain transportation modes but keeping them in a single enum keeps interfaces like this one a lot cleaner.

Issue #3070, opened 2024-10-17
maxspeed=0.5 on a way triggers exception when using bicycle profile


ratrun:
**Describe the bug**
I found a spot where OSM contains ways tagged with `maxspeed=0.5`. This bad data triggers the following exception when using one of the bicycle profiles

```
java.lang.IllegalArgumentException: Speed was 0.5 but cannot be lower than 1.0

	at com.graphhopper.routing.util.parsers.AbstractAverageSpeedParser.setSpeed(AbstractAverageSpeedParser.java:42)
	at com.graphhopper.routing.util.parsers.BikeCommonAverageSpeedParser.handleWayTags(BikeCommonAverageSpeedParser.java:176)
	at com.graphhopper.routing.util.parsers.AbstractAverageSpeedParser.handleWayTags(AbstractAverageSpeedParser.java:54)
	at com.graphhopper.routing.util.OSMParsers.handleWayTags(OSMParsers.java:101)
	at com.graphhopper.routing.util.parsers.AbstractBikeTagParserTester.getSpeedFromFlags(AbstractBikeTagParserTester.java:103)
	at com.graphhopper.routing.util.parsers.BikeTagParserTest.testSmoothness(BikeTagParserTest.java:366)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)

```
**To Reproduce**
Tag a way with tag  `maxspeed=0.5` and use a bicycle profile

**Expected behavior**
No exception

**System Information**

Latest master

ratrun:
I'm not sure if this change is the way to go:

```
    protected void setSpeed(boolean reverse, int edgeId, EdgeIntAccess edgeIntAccess, double speed) {
        if (speed < avgSpeedEnc.getSmallestNonZeroValue() / 2) {
            avgSpeedEnc.setDecimal(reverse, edgeId, edgeIntAccess, avgSpeedEnc.getSmallestNonZeroValue());
        } else {
            avgSpeedEnc.setDecimal(reverse, edgeId, edgeIntAccess, speed);
        }
    }
```

PR #3074, opened 2024-10-27
Avoid parsing exception for very low max_speed values

ratrun:
This is a change for the issue described in #3070 about very low max_speed values, which lead to a parsing exception.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/AbstractAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/AbstractAverageSpeedParser.java
@@ -39,7 +39,7 @@ public abstract class AbstractAverageSpeedParser implements TagParser {
 
     protected void setSpeed(boolean reverse, int edgeId, EdgeIntAccess edgeIntAccess, double speed) {
         if (speed < avgSpeedEnc.getSmallestNonZeroValue() / 2) {
-            throw new IllegalArgumentException("Speed was " + speed + " but cannot be lower than " + avgSpeedEnc.getSmallestNonZeroValue() / 2);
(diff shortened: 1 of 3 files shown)


otbutz:
Is there a real-world example of enforcing a limit slower than walking speed? I'd rather filter this cases in the parser to be honest.

ratrun:
> Is there a real-world example of enforcing a limit slower than walking speed? I'd rather filter this cases in the parser to be honest.

There was at least one spot. In the meantime I fixed it in the data, see https://www.openstreetmap.org/way/750883115/history.

I consider maxspeed=0.5 (km/h) as bad data. Anyway, bad that shouldn't trigger our OSM reader to throw an exception and consequently stop the import.

core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
```diff
@@ -119,6 +119,9 @@ protected BikeCommonAverageSpeedParser(DecimalEncodedValue speedEnc, EnumEncoded
      */
     double applyMaxSpeed(ReaderWay way, double speed, boolean bwd) {
         double maxSpeed = getMaxSpeed(way, bwd);
+        if (maxSpeed < avgSpeedEnc.getSmallestNonZeroValue() / 2) {
+            maxSpeed = avgSpeedEnc.getSmallestNonZeroValue() / 2;
+        }
```
otbutz, line 124:
```suggestion
        maxSpeed = Math.max(maxSpeed, avgSpeedEnc.getSmallestNonZeroValue() / 2);
```

ratrun:
> I think I accidentally fixed this issue in #3077

But if I see it right both pull requests didn't make it into 10.10. Was this intentional and if so, why?

PR #3075, opened 2024-10-28
Snap preventions default

karussell:
For route planning the current empty default for "snap_preventions" is suboptimal. This PR makes it possible to provide a default snap_preventions list on the server-side which can be overwritten on the client-side.


AndreB99:
What if i want to do it in the local version of the open source routing engine? I'm not being able to alter the snap preventions, it's always going for the default even if i do List<String> snappreventions = new ArrayList<>(); and then GHRequest req = new GHRequest(start.valid_lat, start.valid_lon, end.valid_lat, end.valid_lon)
                .setProfile("car")
                .setSnapPreventions(snappreventions);

PR #3084, opened 2024-11-29
refactor(gtfs-reader): replace javacsv with opencsv

kschrab:
This PR exchanges the library `net.sourceforge.javacsv:javacsv` with `com.opencsv:opencsv` which is used to parse CSV files from GTFS. 

The motivation for this change is, that `javacsv` was released 16 years ago, its source code cannot be found anymore, and it was distributed under LGPL, whereas opencsv is Apache 2.0.

For the start I just wrote a wrapper to preserve the old API of the CsvReader from javacsv, and then used the new CSVReader from opencsv within that wrapper. There's also code for writing CSV files, which is not used anywhere though.


reader-gtfs/src/main/java/com/conveyal/gtfs/model/Entity.java
```diff
@@ -100,7 +100,7 @@ protected boolean checkRangeInclusive(double min, double max, double actual) {
         private String getFieldCheckRequired(String column, boolean required) throws IOException {
             String str = reader.get(column);
             if (str == null) {
-                if (!missingRequiredColumns.contains(column)) {
+                if (required && !missingRequiredColumns.contains(column)) {
```
kschrab, line 103: There's one difference between both implementations. The `reader.get(..)` method in `javacsv` returns an empty string if the column is not existing, whereas in `opencsv` it returns `null`. This now leads to this `MissingColumnError` for required fields instead of an `EmptyFieldError`. I also added the `required` check here, as otherwise the `MissingColumError` was raised even for required fields.

kschrab:
I double checked for several GTFS feeds, such as VBB, London, Munich, Japan, and Cairo. The only difference I found was in the number of feed errors being raised, due to a behavioral change in the get method (see other comment). 

I adjusted the `NOTICE.md`.

kschrab:
I furthermore removed the unused code for writing GTFS feeds.

kschrab:
@karussell I noticed that `opencsv` comes with several transient dependencies from Apache Commons: `commons-beanutils`, `commons-text`, `commons-collections4`, and `commons-lang3`. 

However, the CSV parsing requires only `commons-lang3`, the others are just for mapping CSV rows directly on Java objects, which we do not need here. Therefore, I would exclude `commons-beanutils`, `commons-text`, and `commons-collections4` via `pom.xml`.

pom.xml
```diff
@@ -113,9 +113,29 @@
...
+                    <exclusion>
+                        <groupId>org.apache.commons</groupId>
+                        <artifactId>commons-collections4</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-lang3</artifactId>
+                <version>3.13.0</version>
```
otbutz, line 138: We don't use it explicitly, so I'd keep it as a transitive dependency.

CONTRIBUTORS.md
```diff
@@ -48,8 +48,8 @@ Here is an overview:
  * joe-akeem, improvements like #2158
  * JohannesPelzer, improved GPX information and various other things
  * karussell, one of the core developers
- * khuebner, initial turn costs support
```
otbutz, line 51: Looks like an accidental removal?

reader-gtfs/src/main/java/com/conveyal/gtfs/util/CsvReader.java
```diff
@@ -0,0 +1,90 @@
...
+            return null;
+        }
+        return index >= currenRecord.length ? null : currenRecord[index];
+    }
+
+    public boolean readRecord() {
+        try {
+            currenRecord = this.reader.readNext();
+            return currenRecord != null;
+        } catch (Exception e) {
```
otbutz, line 72: Wouldn't it be better to check using `hasNext()` instead of waiting for an exception?

reader-gtfs/src/main/java/com/conveyal/gtfs/util/CsvReader.java
```diff
@@ -0,0 +1,90 @@
...
+            return false;
+        }
+    }
+
+    static class SkipEmptyLines implements LineValidator {
```
otbutz, line 77: Does this really skip? Looks like it would simply throw an exception on the fist occurrence of an empty line.

otbutz (changes requested):

reader-gtfs/src/main/java/com/conveyal/gtfs/util/CsvReader.java
kschrab: From my understanding the CSVReader would skip all lines in which the LineValidator raises an exception, but I will double check this!

CONTRIBUTORS.md
kschrab: No, that was my old handle ;)

reader-gtfs/src/main/java/com/conveyal/gtfs/util/CsvReader.java
kschrab: There is no `hasNext()`, but I could use `peek()` instead. I rewrote the whole method and it now takes care of skipping empty lines by itself now, as the LineValidator does not work as expected (see other comment).

reader-gtfs/src/main/java/com/conveyal/gtfs/util/CsvReader.java
kschrab: Your intention was right. There is no nice way to skip empty lines in opencsv though.

otbutz: Do we even encounter such files?

kschrab: In the most examples there's an empty line at the end of the file which produced that one empty record.

reader-gtfs/src/main/java/com/conveyal/gtfs/util/CsvReader.java
```diff
@@ -0,0 +1,93 @@
...
+                if (this.reader.peek() == null) {
+                    currenRecord = null;
+                    return false;
+                }
+                nextRecord = this.reader.readNextSilently();
+            } while (isEmptyLine(nextRecord));
+
+            currenRecord = nextRecord;
+            return true;
+        } catch (IOException e) {
```
otbutz, line 79: This error should be propagated, not quietly discarded.

otbutz (changes requested):

kschrab:
I will check that asap! I somehow forgot about this PR 🫣

PR #3089, opened 2024-12-13
Switch to osmpbf

### LINKED ISSUE #3083: Reading PBF based maps depends on old Protobuf library
**Describe the bug**

Osmosis dependency used by Graphhopper depends on protobuf-java version 3.25.5 max.

My project depends on protobuf-java: 4.28.2 and overwrites dependency version of the one defined by osmosis. 

Project fails because `makeExtensionsImmutable` method is not present anymore in GeneratedMessage in any version of ProtoBuf higher than 3.25.5.

**To Reproduce**

I have a higher protobuf-java dependency in my project. 

I read PBF formatted OSM maps from
http://download.geofabrik.de/europe/netherlands-latest.osm.pbf

I import it with GraphHopper. 

It fails with java.lang.NoSuchMethodError for `makeExtensionsImmutable`.

```
Caused by: java.lang.RuntimeException: Unable to read PBF file.
	at com.graphhopper.reader.osm.pbf.PbfReader.close(PbfReader.java:66)
	at com.graphhopper.reader.osm.OSMInputFile.close(OSMInputFile.java:241)
	at com.graphhopper.reader.osm.WaySegmentParser.readOSM(WaySegmentParser.java:402)
	... 14 more
Caused by: java.lang.NoSuchMethodError: 'void org.openstreetmap.osmosis.osmbinary.Fileformat$BlobHeader.makeExtensionsImmutable()'
	at org.openstreetmap.osmosis.osmbinary.Fileformat$BlobHeader.<init>(Fileformat.java:1235)
	at org.openstreetmap.osmosis.osmbinary.Fileformat$BlobHeader.<init>(Fileformat.java:1158)
	at org.openstreetmap.osmosis.osmbinary.Fileformat$BlobHeader$1.parsePartialFrom(Fileformat.java:1914)
	at org.openstreetmap.osmosis.osmbinary.Fileformat$BlobHeader$1.parsePartialFrom(Fileformat.java:1908
```

**Expected behavior**
No errors with latest protobuf version in project.

**System Information**
JVM 17 on macOS

otbutz:
Fixes #3083

* switched the dependency as outlined in https://github.com/openstreetmap/osmosis/pull/97
* transitive protobuf-java dependency changed to 4.28.3
* regenerated VectorTile from [vector_tile.proto](https://github.com/ElectronicChartCentre/java-vector-tile/blob/master/src/main/protobuf/vector_tile.proto) using protoc 29.1 (compatible according to https://protobuf.dev/support/version-support/#java)


otbutz:
@easbar could you verify that the VectorTile support still works as intended? I simply recreated that file based on the information in #2698

otbutz:
I've added a unit test just to be sure. At least the decoding seems to work correctly.

otbutz:
Blocked by https://github.com/MobilityData/gtfs-realtime-bindings/issues/138 :disappointed:

otbutz:
@karussell we could simply generate the protobuf classes for `GtfsRealtime` and `VectorTile` as part of our build. We can use `protobuf-maven-plugin` to automate this. As a small benefit, we get rid of generated code in our repository and don't have to keep the protobuf version of three dependencies in sync. WDYT?

otbutz:
@karussell The current PR now generates both classes from their protobuf source.

otbutz:
> I would avoid creating the pbf as part of the build as I made not too good experiences with different IDEs when this is done.

Fair point. I've tested the integration in Intellij which worked flawlessly. I've added the necessary m2e annotations for Eclipse as well.

> Probably the better way is to fork and create com.graphhopper.external::gtfs-realtime-bindings?

We'd need to create an additional external project if we take that approach. [VectorTile](https://github.com/graphhopper/graphhopper/blob/master/web-bundle/src/main/java/vector_tile/VectorTile.java) was manually generated from [vector_tile.proto](https://github.com/ElectronicChartCentre/java-vector-tile/blob/master/src/main/protobuf/vector_tile.proto) and also needs to be kept in sync with the protobuf version we're using internally.

otbutz:
@karussell how should we proceed here? I can create two repositories which encapsulate the pre-built protobuf classes.

karussell:
With two projects do you mean the VectorTile and realtime GTFS dependencies?

Currently we are using the VectorTile.java which was created once. Couldn't we just update it to use the more recent 4.x version without the need for an external dependency?

btw: Does the PBF version update have other advantages, such as faster import?

otbutz:
> Currently we are using the VectorTile.java which was created once. Couldn't we just update it to use the more recent 4.x version without the need for an external dependency?

We could do this. However, if I remember correctly, newer versions of protobuf are more restrictive in terms of the compatibility of generated classes with the library. We might need to update it every time we update the protobuf dependency.

We could also mix both strategies. The generated files stay checked into the repository, but we provide the manual Maven goal to easily regenerate them if we update the dependency.

> btw: Does the PBF version update have other advantages, such as faster import?

Maybe? I haven't done any benchmarks as the focus was purely on dependency management. Our current dependency won't be updated any longer, so we'll need to switch either way :shrug:

otbutz:
@karussell the generated classes are now in the repository and can be recreated with the `protobuf` maven profile:

```sh
mvn compile -P protobuf
```

otbutz:
@karussell any chance to get this merged for the v11 release?

karussell:
Unfortunately no. This is too tricky and also not fully tested (I think) for the parts that were replaced.

otbutz:
@karussell feel free to give it a spin :wink:

otbutz:
> This is too tricky

I'd agree if it was my code, but the protobuf files were simply vendored into our repo and the resulting code was generated.

> and also not fully tested (I think) for the parts that were replaced.

I've tested the following:

* OSM import :heavy_check_mark:
* Vector tiles (unit test + map) :heavy_check_mark: 
* GTFS :negative_squared_cross_mark: (the [example](https://github.com/graphhopper/graphhopper/blob/master/reader-gtfs/README.md#quick-start) also fails on the current master though :slightly_frowning_face: )

otbutz:
Superseded by https://github.com/graphhopper/graphhopper/commit/d6093d65434cd497933400670adbc7dc7a316789

PR #3096, opened 2025-01-01
fix missing road_access option in config-example.yml

OsamaAlmaani:
Add missing option `road_access` to the graph.encoded_values in config-example.yml file 

image


config-example.yml
```diff
@@ -92,7 +92,7 @@ graphhopper:
   # More are: surface,smoothness,max_width,max_height,max_weight,max_weight_except,hgv,max_axle_load,max_length,
   #           hazmat,hazmat_tunnel,hazmat_water,lanes,osm_way_id,toll,track_type,mtb_rating,hike_rating,horse_rating,
   #           country,curvature,average_slope,max_slope,car_temporal_access,bike_temporal_access,foot_temporal_access
-  graph.encoded_values: car_access, car_average_speed
+  graph.encoded_values: car_access, car_average_speed, road_access
```
ratrun, line 95: Please also remove the value "road_access" in line 91 as it is no default any longer

OsamaAlmaani: done

ratrun (approved):
Looks good to me! Now it needs to be checked by the maintainers.

karussell:
Thanks. Just changed some minor things.

Issue #3098, opened 2025-01-02
Navigation leads through low-quality roads


Wertok07:
Describe the bug
It is related to this topic https://github.com/graphhopper/graphhopper/issues/3038
Navigation leads through low-quality roads

To Reproduce
https://www.openstreetmap.org/directions?engine=graphhopper_car&route=50.06995%2C19.85669%3B50.07361%2C19.84839#map=17/50.07020/19.85284

Expected behavior
https://graphhopper.com/maps/?point=50.070113%2C19.856362&point=50.067964%2C19.84795&point=50.073487%2C19.848294&profile=car&layer=Omniscale

The “Pylna” road is a local road, made of gravel such status also has in OpenStreetMap and it takes significantly more time to drive it.

System Information
Versions of the map available at the links above

Wertok07:
It's hard to say, the regulations don't explicitly say whether it's 50 or 40km/h. In fact, you can drive about 15 km/h on this road but this is not the limit.

otbutz:
While not the culprit here, I think it's weird that we completely ignore the badSurfaceSpeed once we find a `maxspeed` tag:

https://github.com/graphhopper/graphhopper/blob/4e232ac93019c6eb140e436542fff7a983c21123/core/src/main/java/com/graphhopper/routing/util/parsers/CarAverageSpeedParser.java#L133-L137

https://github.com/graphhopper/graphhopper/blob/4e232ac93019c6eb140e436542fff7a983c21123/core/src/main/java/com/graphhopper/routing/util/parsers/CarAverageSpeedParser.java#L145-L147

You can't drive faster on gravel just because the (legal) speed limit allows it.

otbutz:
@easbar I have trouble understanding how we compute an average speed of 36km/h for this road?

![grafik](https://github.com/user-attachments/assets/03384417-800e-4cc5-af18-5c60f5198299)

It's a residential road without a maxspeed tag, so the result should be derived from the defaultSpeedMap.

DujaOSM:
I was about to open an issue, but this one looks like the correct venue, so I'll just give an even more drastic use case:
[...]

PR #3099, opened 2025-01-03
Always limit speed on bad surfaces

otbutz:
As outlined in https://github.com/graphhopper/graphhopper/issues/3098#issuecomment-2569037058

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/CarAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/CarAverageSpeedParser.java
@@ -131,7 +131,6 @@ public class CarAverageSpeedParser extends AbstractAverageSpeedParser implements
 
         // get assumed speed from highway type
         double speed = getSpeed(way);
-        speed = applyBadSurfaceSpeed(way, speed);
 
         setSpeed(false, edgeId, edgeIntAccess, applyMaxSpeed(way, speed, false));
         setSpeed(true, edgeId, edgeIntAccess, applyMaxSpeed(way, speed, true));
@@ -144,7 +143,8 @@ public class CarAverageSpeedParser extends AbstractAverageSpeedParser implements
      */
     protected double applyMaxSpeed(ReaderWay way, double speed, boolean bwd) {
         double maxSpeed = OSMMaxSpeedParser.parseMaxSpeed(way, bwd);
-        return maxSpeed != MaxSpeed.MAXSPEED_MISSING ? Math.max(1, maxSpeed * 0.9) : speed;
+        maxSpeed = maxSpeed != MaxSpeed.MAXSPEED_MISSING ? Math.max(1, maxSpeed * 0.9) : speed;
+        return applyBadSurfaceSpeed(way, maxSpeed);
     }
 
     /**
--- a/web/src/test/java/com/graphhopper/application/resources/RouteResourceLeipzigTest.java
+++ b/web/src/test/java/com/graphhopper/application/resources/RouteResourceLeipzigTest.java
@@ -80,12 +80,12 @@ public class RouteResourceLeipzigTest {
 
     @ParameterizedTest
     @CsvSource(value = {
-            "80,-1,algorithm=" + DIJKSTRA_BI,
-            "108,-1,algorithm=" + ASTAR_BI,
-            "30743,1,ch.disable=true&algorithm=" + DIJKSTRA,
-            "21133,1,ch.disable=true&algorithm=" + ASTAR,
-            "14866,1,ch.disable=true&algorithm=" + DIJKSTRA_BI,
-            "10542,1,ch.disable=true&algorithm=" + ASTAR_BI
+            "86,-1,algorithm=" + DIJKSTRA_BI,
+            "109,-1,algorithm=" + ASTAR_BI,
+            "30736,1,ch.disable=true&algorithm=" + DIJKSTRA,
+            "21127,1,ch.disable=true&algorithm=" + ASTAR,
+            "14828,1,ch.disable=true&algorithm=" + DIJKSTRA_BI,
+            "10500,1,ch.disable=true&algorithm=" + ASTAR_BI
     })
     void testTimeout(int expectedVisitedNodes, int timeout, String args) {
         {


otbutz:
@karussell WDYT?

otbutz:
@karussell rebased the PR on the current master

karussell:
Do you have an example where this improves the routing? I know it was originally for #3098 but this was fixed and IMO more a data issue.

And the other route mentioned in this issue is not fixed [when a lower speed is used for bad surface values](https://graphhopper.com/maps/?point=43.23047%2C19.718412&point=43.272873%2C19.999173&profile=car&layer=Omniscale&custom_model=%7B%22speed%22%3A%5B%7B%22if%22%3A%22surface%3D%3DWOOD%7C%7Csurface%3D%3DDIRT%7C%7Csurface%3D%3DGRAVEL%7C%7Csurface%3D%3DUNPAVED%22%2C%22limit_to%22%3A%2210%22%7D%5D%7D) - even if I use 15km/h. Or is this an appropriate alternative @DujaOSM? On screen this looks fine (regarding highway and surface tag) and much better already than our current choice. However, the 30km/h cap in this PR won't do it.

Also we should probably move this code out into a custom model so that one can easier modify it if necessary. Especially when we use lower speeds it is questionable if surface=compact is really that bad.

DujaOSM:
That's a _reasonable_ route, at least with the data available. There's a [long-distance cycling route](https://www.openstreetmap.org/relation/11753312#map=11/43.4213/19.8420&layers=C) running over it. The catch is that some 10-15 km of the central part (Mileševo - Vodena poljana) is unpaved and of rather poor quality (last time I was there, around 2022) but that is not explicitly recorded in the OSM data -- those roads are just tagged as `unclassified` without surface or smoothness data. But 15 km/h over presumably unpaved roads looks like a reasonable assumption.

otbutz:
@karussell I can look for an example, but, conceptually speaking, it's the right thing to do. Physical limits trump legal ones :wink:

karussell:
Theoretically yes, but in practise surface is not 100% mapped which could lead to strange detours if we add a too strong penalty.

[I found a route in real world](https://graphhopper.com/maps/?point=56.677692%2C25.983519&point=57.152496%2C26.798803&profile=car&layer=OpenStreetMap) where reducing speed for bad surfaces would lead to a better route. However, it is unclear to me if this is not more a tagging mistake and if the road is really a secondary roads. 

Or [here](https://graphhopper.com/maps/?point=56.677692%2C25.983519&point=56.766063%2C26.401153&profile=car) is another example where even a primary road has surface=gravel. Of course it could be like this in real world, but it could be also a tagging mistake. 

But at least this situation confuses other routing engines too, which does not only pick surface=gravel but even prefers unclassified+secondary over primary:

image

Where [it should be](https://graphhopper.com/maps/?point=56.677692%2C25.983519&point=56.744304%2C27.116033&profile=car&layer=OpenStreetMap&custom_model=%7B%22speed%22%3A%5B%7B%22if%22%3A%22surface%3D%3DCOMPACTED%22%2C%22limit_to%22%3A%2250%22%7D%2C%7B%22if%22%3A%22surface%3D%3DGRAVEL%7C%7Csurface%3D%3DUNPAVED%22%2C%22limit_to%22%3A%2230%22%7D%5D%7D):

image

otbutz:
> Or [here](https://graphhopper.com/maps/?point=56.677692%2C25.983519&point=56.766063%2C26.401153&profile=car) is another example where even a primary road has surface=gravel. Of course it could be like this in real world, but it could be also a tagging mistake.

Seems like a good example:

grafik

https://www.google.com/maps/@56.6848202,26.0116024,3a,73y,77.96h,76.62t/data=!3m7!1e1!3m5!1sIEc9dqe1zQ_fcWH2ErlSWg!2e0!6shttps:%2F%2Fstreetviewpixels-pa.googleapis.com%2Fv1%2Fthumbnail%3Fcb_client%3Dmaps_sv.tactile%26w%3D900%26h%3D600%26pitch%3D13.375616229138302%26panoid%3DIEc9dqe1zQ_fcWH2ErlSWg%26yaw%3D77.95964175831533!7i16384!8i8192!5m1!1e1?entry=ttu&g_ep=EgoyMDI1MTIwMi4wIKXMDSoASAFQAw%3D%3D

DujaOSM:
> Or [here](https://graphhopper.com/maps/?point=56.677692%2C25.983519&point=56.766063%2C26.401153&profile=car) is another example where even a primary road has surface=gravel. Of course it could be like this in real world, but it could be also a tagging mistake.

According to [Google StreetView](https://maps.app.goo.gl/BBtsmZbSSQdzfJBv6), that's indeed (good-quality, well-compacted) gravel, but then, it should not be tagged `primary` in a developed country (whatever the official classification). So I'm not sure whether to consider those edge cases.

otbutz:
> Theoretically yes, but in practise surface is not 100% mapped which could lead to strange detours if we add a too strong penalty.

@karussell the `applyBadSurfaceSpeed` only lowers the speed if the way is tagged with a `surface` that's part of our `badSurfaceSpeedMap`. That means that it's not a problem if the surface tag is complexly missing for a way. If parts of the highway are tagged and others are not, your point is valid. But that kind of inconsistencies are somewhat expected and should be treated as a data issue IMHO.

karussell:
See latest commit

karussell:
This approach is too generic and e.g. compacted and paving_stones are now a too heavy penalty.
Will try a more differentiated approach:

```java
public static final Map<Surface, Integer> BAD_SURFACES = Map.ofEntries(
            entry(PAVING_STONES, 60),
            entry(COMPACTED, 50), entry(FINE_GRAVEL, 50),
            entry(GRAVEL, 40),
            entry(UNPAVED, 30), entry(COBBLESTONE, 30), entry(DIRT, 30), entry(GROUND, 30), entry(WOOD, 30), entry(OTHER, 30),
            entry(SAND, 20), entry(GRASS, 20));
```

It based on the order we use for bike, except for wood where I think it is fine to be lower in case of car. 
(btw: entry is necessary in Java because the count is >10)

This also improves [this route](https://graphhopper.com/maps/?point=56.677692%2C25.983519&point=56.744304%2C27.116033&profile=car).

PR #3106, opened 2025-01-16
fix: starting at a barrier for navigation converter

OlafFlebbeBosch:
The previous fix handled two issues: 
* Deduplicating barriers (intersections) for NavigationConverter
* Fixing Arrival and Departures (intersections) for NavigationConverter to have only an in or out link. 
What I missed is an arrival or departure intersection, deduplicating an Arrival or Departure intersection with only an in or out link.

@boldtrn can you check please ?

### DIFF
--- a/core/src/main/java/com/graphhopper/util/details/IntersectionValues.java
+++ b/core/src/main/java/com/graphhopper/util/details/IntersectionValues.java
@@ -36,8 +36,8 @@ public class IntersectionValues {
         List<IntersectionValues> list = new ArrayList<>();
 
         List<Integer> bearings = (List<Integer>) intersectionMap.get("bearings");
-        Integer in = (Integer) intersectionMap.get("in");
-        Integer out = (Integer) intersectionMap.get("out");
+        Integer in = (Integer) intersectionMap.getOrDefault("in", -1);
+        Integer out = (Integer) intersectionMap.getOrDefault("out", -1);
         List<Boolean> entry = (List<Boolean>) intersectionMap.get("entries");
 
         if (bearings.size() != entry.size()) {
--- a/navigation/src/test/java/com/graphhopper/navigation/NavigateResponseConverterTest.java
+++ b/navigation/src/test/java/com/graphhopper/navigation/NavigateResponseConverterTest.java
@@ -55,7 +55,8 @@ public class NavigateResponseConverterTest {
     @Test
     public void basicTest() {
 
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/graphhopper/util/details/IntersectionValues.java
```diff
@@ -49,8 +49,8 @@ public static List<IntersectionValues> createList(Map<String, Object> intersecti
             IntersectionValues iv = new IntersectionValues();
             iv.bearing = bearings.get(i);
             iv.entry = entry.get(i);
-            iv.in = (in == i);
-            iv.out = (out == i);
+            iv.in = (in == null ? false : in == i);
+            iv.out = (out == null ? false : out == i);
```
boldtrn, line 53: Good catch. I am just wondering if it would be better to use `getOrDefault` in line 39/40 with a default value of `-1`,  that way we are sure the value is not null?

navigation/src/test/java/com/graphhopper/navigation/NavigateResponseConverterTest.java
```diff
@@ -349,6 +349,58 @@ public void barrierTest() {
         assertEquals(step.get("intersections").size(), 2);
     }
 
+
+    @Test
+    // start the route exactly the barrier 
+    // The barrier should be deduplicated and have only one out. 42,6017641 
+    public void startAtBarrierTest() {
+        // There is a barrier https://www.openstreetmap.org/node/2206610569 on the route
+        GHResponse rsp = hopper.route(new GHRequest(  42.6017641, 1.6878903, 42.601616, 1.687888).setProfile(profile)
```
boldtrn, line 358: code formatting :)

boldtrn (approved):
From my point of view this looks good. I would only ask @OlafFlebbeBosch to reformat the code. I am fine without the getOrDefault.

core/src/main/java/com/graphhopper/util/details/IntersectionValues.java
OlafFlebbeBosch: Isn't getOrDefault a Map method? Where would be an Map around to use getOrDefault ?

boldtrn: In line 39 and 40. `intersectionMap.get("in");`

OlafFlebbeBosch:
I force pushed the changes proposed by @boldtrn

boldtrn:
Thanks, your changes look good to me.

PR #3124, opened 2025-02-19
Lookup state from Map

otbutz:
Takes #3123 one step further by getting rid of any String manipulation.


karussell:
It seems that this small change improved our import time (for Germany+car) over 15% :+1: 

Or, more specific, the previous change #3123 was the real problem, I think.

Thanks again!

PR #3126, opened 2025-02-20
Handle implicit HGV access

otbutz:
The OSMHgvParser in its current form only handles two types of tags: `hgv` and `hgv:conditional`. Other access EVs like `road_access`,  `bike_road_access` or `foot_road_access` are populated by the `OSMRoadAccessParser` which can deal with the hierarchical nature of access tagging and also covers barriers.

This PR changes the `OSMHgvParser` to be a simple extension of the `OSMRoadAccessParser` which only handles the conditional parsing logic.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMHgvParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMHgvParser.java
@@ -5,22 +5,29 @@ import com.graphhopper.reader.ReaderWay;
 import com.graphhopper.routing.ev.EdgeIntAccess;
 import com.graphhopper.routing.ev.EnumEncodedValue;
 import com.graphhopper.routing.ev.Hgv;
+import com.graphhopper.routing.util.TransportationMode;
 import com.graphhopper.storage.IntsRef;
 
 import static com.graphhopper.routing.util.parsers.helpers.OSMValueExtractor.conditionalWeightToTons;
 
-public class OSMHgvParser implements TagParser {
-    EnumEncodedValue<Hgv> hgvEnc;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/OSMHgvParser.java
```diff
@@ -5,22 +5,29 @@
...
+                (readerWay, accessValue) -> accessValue, Hgv::find);
     }
 
     @Override
     public void handleWayTags(int edgeId, EdgeIntAccess edgeIntAccess, ReaderWay way, IntsRef relationFlags) {
         String value = way.getTag("hgv:conditional", "");
         int index = value.indexOf("@");
-        Hgv hgvValue = index > 0 && conditionalWeightToTons(value) == 3.5 ? Hgv.find(value.substring(0, index).trim()) : Hgv.find(way.getTag("hgv"));
-        hgvEnc.setEnum(false, edgeId, edgeIntAccess, hgvValue);
+        if (index > 0 && conditionalWeightToTons(value) == 3.5) {
```
otbutz, line 24: Why do we check for *exactly* 3.5t?

The following examples wouldn't work while they seem perfectly fine IMHO:

* `private @ (weight>12)`
* `delivery @ (weight>5.5)`

```suggestion
        if (index > 0 && conditionalWeightToTons(value) >= 3.5) {
```

otbutz:
Paired with #3066 this would solve #2460

otbutz:
@karussell if you don't have any objections, I'd add the missing pieces as outlined in https://github.com/graphhopper/graphhopper/pull/3126#discussion_r1963948352

PR #3127, opened 2025-02-21
Enchantments to the routing output with GTFS data.

elvedin-hamzagic:
Added route_type, agency_id and agency_name to a route leg for 'pt' legs.

route_type - type of the vehicle used in public transport, e.g. 3 for a bus, 0 for a tram and 11 for a trolley
agency_id - the ID of the operator as stated in the GTFS table agencies.txt
agency_name - the name of the operator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - road_access now contains value of highest transportation mode for CAR, i.e. access=private, motorcar=yes will now return YES and not PRIVATE
 - car.json by default avoids private roads
 - maxspeed<5 is ignored, maxspeed=none is ignored with some exceptions, maxspeed parsing and related constants were renamed #3077
-- Added route_type in Trip.PtLeg to describe the vehicle type used in public transport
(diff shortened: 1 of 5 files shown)


elvedin-hamzagic:
Further enhancements in output data of "/route" endpoint:

- Fixed "distance" for public transport legs. The current implementation of routing didn't take into consideration the distance of PT legs, because it relied solely on timestamps from "stop_times" GTFS table, so the distance was always 0. In this implementation the distance has been calculated from GTFS table "shapes", if the table exists. If the table does not exist, the distance is still 0.

- Added "shapes" field to the output for pricise route drawing. This field contains LineString geometry, which are coordinates taken from "shapes" table from GTFS feed. In case there is no table in the feed, it will be empty. Current implementation includes "geometry" field, but it contains only coordinates of stations.

In the future, we could add an input parameter in the request to control whether "shape" field should be included in the output, because the output size can increase significantly, approximately around 20-25%.

PR #3128, opened 2025-02-25
In JTSTriangulator, use DelaunayTriangulationBuilder instead of ConformingDelaunayTriangulator

james-willis:
This PR is based on this discussion with Martin (Dr. JTS): https://github.com/locationtech/jts/issues/1121

He suggested that DelaunayTriangulationBuilder is likely the most appropriate way to build this triangulation.

i'm unsure if there are other simplifications that can be made to this class based on the changes in this PR.

I apologize for failing to add a new test, I am struggling to build a test case that captures the bug in the JTS issue above.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -41,6 +41,7 @@ Here is an overview:
  * hoofstephan, bug fix   
  * IsNull, improvements like #708
  * IldarKhayrutdinov, use cache on different machines
+ * james-willis, improved isochrones triangulation robustness
  * Janekdererste, GUI for public transit
  * jansoe, many improvements regarding A* algorithm, forcing direction, roundabouts etc
  * jansonhanson, general host config
--- a/core/src/main/java/com/graphhopper/isochrone/algorithm/JTSTriangulator.java
+++ b/core/src/main/java/com/graphhopper/isochrone/algorithm/JTSTriangulator.java
@@ -27,16 +27,15 @@ import com.graphhopper.util.FetchMode;
 import com.graphhopper.util.PointList;
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.Polygon;
-import org.locationtech.jts.triangulate.ConformingDelaunayTriangulator;
-import org.locationtech.jts.triangulate.ConstraintVertex;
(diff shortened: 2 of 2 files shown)


james-willis:
> why is triangulationBuilder.getSubdivision() called twice?

Fixed, it was not intentional.

> Did you test if there are any speed differences? 

The new version seems to be substantially faster. 

I have an isochrone generation test that generates a 4 hour isochrone from some point in modesto with `california-latest.osm.pbf` from a few weeks ago. With the old implementation, it takes 5m 15s. With the new implementation it takes 1m 54s. I ran the test a few times to confirm.

> Unfortunately without this it will be harder to accept this PR.

I will think about how I can write a good test to reproduce the stability issue.

james-willis:
I wrote a test but it doesn't trigger the triangulation error in master.

Unfortunately I don't have more time to dedicate to this PR. 

Test can be found in this branch: https://github.com/james-willis/graphhopper/tree/triangulator-builder-with-test.

I can add to this PR if you would like

james-willis:
> What do you mean the test doesn't trigger the triangulation error in master. Does it trigger the error somewhere else?

I mean the test doesn't test the bug I've fixed with this PR. Fairly worthless.

james-willis:
Yes, I'm not sure if there is some characteristic of the graph (not OSM derived) I'm generating Isochrones from in that issue report that causes the bug to trigger.

I think the test code I wrote could be used if we find a good case to repro with. I might be able to come back to this next week and see if I can fit the case from the issue above into the unit test in the linked branch.

michaz (approved):
Extremely good catch, many thanks!

Background: I ultimately do want to use  `ConformingDelaunayTriangulator`, because I want to set the entire road network as constraints. Because that solves some problems that still exist. I verified that with a different, non-Java, implementation. However, I couldn't do that with this class, because I ran into issues of stability and performance.

What I missed was that some issues are still there if I **don't** set any constraints. Apparently I just took for granted that the class would behave the same as the regular one if the constraint set is empty.

If your feedback to Dr. JTS now means that someday we will be able to actually do the constrained triangulation, I will be up in the clouds.

michaz:
I mean...

michaz:
There is also `ConformingDelaunayTriangulationBuilder`. I'm not sure if I used that back then.

Probably not.

By analogy, it might magically work if I use it.

james-willis:
> There is also ConformingDelaunayTriangulationBuilder. I'm not sure if I ever tried that.

Yeah Dr JTS recommended this until he noticed we had no constraints.

PR #3131, opened 2025-02-28
Normalize and check custom area Ids

otbutz:
See https://discuss.graphhopper.com/t/custom-areas-area-null-already-exists/9387/2

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -1492,7 +1492,16 @@ public class GraphHopper {
                         .sorted(Comparator.comparing(Path::toString))
                         .forEach(customAreaFile -> {
                             try (BufferedReader reader = Files.newBufferedReader(customAreaFile, StandardCharsets.UTF_8)) {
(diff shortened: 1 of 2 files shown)


sephib:
Hi,
possibly update the [config-example](https://github.com/graphhopper/graphhopper/blob/2ca85c1dd4a0af92ad23ab533c7a0f87f734ba58/config-example.yml#L216) with:
> All GeoJSON Features require to have the "id" property in the root of the feature. e.g.:
{    "type": "FeatureCollection",    "features": [{"type": "Feature","id": "my_id", "geometry": {"type": "Polygon", "coordinates": [ [ [...],] ] } }

otbutz:
+1 from me. The property fallback always felt a bit weird.

@karussell that would require a rebuild of the current `countries.geojson` :

```json
{"type":"Feature","properties":{"id":"TT"},"geometry":{"type":"Polygon","coordinates":[[[-62.08693,10.04435],[-60.993,9.88751],[-60.54419,10.87433],[-60.21728,11.59108],[-61.08554,11.23803],[-61.70744,10.98489],[-62.08693,10.04435]]]}}
```

We might as well add newlines as @easbar already [suggested](https://github.com/graphhopper/graphhopper/pull/3130#issuecomment-2685716752)

PR #3132, opened 2025-02-28
Allow to request path details without path simplification

otbutz:
Currently it's a bit cumbersome to request path details without path simplification.

This doesn't return any path details, as either `calc_points` or `instructions` are implicitly required:

```java
GHRequest request = new GHRequest(start, destination)
				.setPathDetails(List.of("distance", "country", "toll"))
				.putHint(Parameters.Routing.INSTRUCTIONS, false)
				.putHint(Parameters.Routing.CALC_POINTS, false);
```

Returns path details but also triggers path simplification by default:

```java
GHRequest request = new GHRequest(start, destination)
				.setPathDetails(List.of("distance", "country", "toll"))
				.putHint(Parameters.Routing.INSTRUCTIONS, false);
				.putHint(Parameters.Routing.CALC_POINTS, true);
```

Works, but is arguably kinda ugly:

```java
GHRequest request = new GHRequest(start, destination)
				.setPathDetails(List.of("distance", "country", "toll"))
				.putHint(Parameters.Routing.INSTRUCTIONS, false)
				.putHint(Parameters.Routing.CALC_POINTS, true)
				.putHint(Parameters.Routing.WAY_POINT_MAX_DISTANCE, 0);
```

### DIFF
--- a/core/src/main/java/com/graphhopper/util/PathMerger.java
+++ b/core/src/main/java/com/graphhopper/util/PathMerger.java
@@ -125,7 +125,7 @@ public class PathMerger {
                 }
 
             }
-            if (calcPoints || enableInstructions) {
+            if (calcPoints || enableInstructions || !requestedPathDetails.isEmpty()) {
                 PointList tmpPoints = path.calcPoints();
                 if (fullPoints.isEmpty())
                     fullPoints = new PointList(tmpPoints.size(), tmpPoints.is3D());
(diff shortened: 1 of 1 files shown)


otbutz:
> How do you use the path details without the `points` array?

I'm only interested in distance statistics. In this particular case km per country and toll mileage. I don't need the full path for that, so the simplification doesn't do anything useful for me.

> Can you outline how it looks after this PR? I.e. what has improved.

The first example returns the requested path details with the unsimplified path now.

PR #3141, opened 2025-03-11
fix: remove outdated dependency

MichaelsJP:
The test jar dependency for Graphhopper is not needed and prevents the modules being built when not running the tests.


MichaelsJP:
Please ignore.

PR #3142, opened 2025-03-12
Improve compressed stream handling

otbutz:
The default buffer size for GZIPInputStream and InflaterInputStream is only 512 bytes, resulting in many tiny OS read calls.

I also replaced all custom InputStream to OutputStream piping with [InputStream.transferTo(OutputStream)](https://docs.oracle.com/javase/9/docs/api/java/io/InputStream.html#transferTo-java.io.OutputStream-) calls.


otbutz:
Local testing with a large gzipped file shows a performance improvement of around 25%

ratrun:
> Local testing with a large gzipped file shows a performance improvement of around 25%

Seems to apply only for one specific elevation provider, but such high improvements are always nice.

otbutz:
> Seems to apply only for one specific elevation provider

SRTM uncompression should also be around 10% faster now.

PR #3143, opened 2025-03-13
Modernize Downloader

otbutz:
* Reimplement Downloader based on java.net HttpClient
* allow callers to skip transport compression
* use a proper User-Agent e.g `graphhopper/11.0`
* handle partial downloads

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/dem/SonnyProvider.java
+++ b/core/src/main/java/com/graphhopper/reader/dem/SonnyProvider.java
@@ -19,6 +19,8 @@ package com.graphhopper.reader.dem;
 
 import java.io.*;
 
+import com.graphhopper.util.Downloader;
+
 import static com.graphhopper.util.Helper.close;
 
 /**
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/graphhopper/reader/dem/SonnyProvider.java
```diff
@@ -19,36 +19,47 @@
...
                 -56,
                 90,
                 3601
         );
+		this.downloader = new Downloader() {
+			@Override
+			public void downloadFile(String url, File toFile) {
+				throw new RuntimeException("Sonny elevation data cannot be downloaded automatically. " +
+						"Please download the data manually from " + SONNY_DOWNLOAD_URL +
+						", unzip it and place the .hgt files in the cache directory: " + getCacheDir());
```
otbutz, line 60: @ratrun This should make things a bit more user-friendly

otbutz:
@karussell successfully tested with all our providers. This should also make elevation handling much more robust. GraphHopper can now be safely stopped during the import phase without corrupting the cache :slightly_smiling_face:

core/src/main/java/com/graphhopper/reader/dem/SonnyProvider.java
ratrun: Thank you, this looks good! :smile:

karussell:
Hm, I do not see the benefits here. All the advantages you listed could as well be implemented for HttpURLConnection. Or am I misunderstanding something?

> GraphHopper can now be safely stopped during the import phase without corrupting the cache 🙂

Because of storing the data into a .part file first? Wouldn't this work for the old HttpURLConnection too?

otbutz:
> Hm, I do not see the benefits here. All the advantages you listed could as well be implemented for HttpURLConnection. Or am I misunderstanding something?

The API is an ugly relic from the Java 1.1 era. The main benefit though is that HttpClient supports HTTP/2 (and even HTTP/3 with Java 26+), which saves bandwidth because they use a binary protocol with header compression. The new client also supports asynchronous calls, if we want to add some kind of prefetching :slightly_smiling_face: 

> > GraphHopper can now be safely stopped during the import phase without corrupting the cache 🙂
> 
> Because of storing the data into a .part file first? Wouldn't this work for the old HttpURLConnection too?

Sure. I finally managed to pinpoint this bug while testing my HttpClient changes and simply added it as part of this PR :shrug:

karussell:
Sorry, I forgot about this PR. Does this mean this is not a change to a new library and just makes the downloader more robus?

otbutz:
#3317 made the download itself more robust. The remaining bits of this PR replace the legacy HttpURLConnection based downloader with one based on java.net.http.HttpClient

PR #3148, opened 2025-03-19
Increaese speed for compacted surface for MTB

caspg:
It feels that [compacted](https://wiki.openstreetmap.org/wiki/Tag:surface=compacted?uselang=en-GB) surface type should have similar speed to gravel and ground.


ratrun:
Makes sense!

PR #3151, opened 2025-03-31
Extend conditional tags supported by MaxWeightExceptParser

otbutz:
I recently updated the [Achenseestraße](https://www.openstreetmap.org/way/71890079) in OSM, which has a conditional weight restriction with an exemption for destination traffic:

```
maxweightrating:hgv=7.5
maxweightrating:hgv:conditional=none @ destination
```

My local instance kept blocking this road even after whitelisting `max_weight_except == DESTINATION` in my custom model. The problem was quite simple, the parser doesn't handle `maxweightrating:hgv:conditional` yet.

With these changes it should cover all conditional variants of the tags supported by the OSMMaxWeightParser.


otbutz:
In addition, conditional tags are quite difficult to understand. In this case, a maxweight tag indirectly changes the access of a road. So +1 from me for lifting this a level higher up into the custom model.

I'd still like this to be merged if only for the selfish purpose of keeping my fork a little closer to upstream :)

PR #3165, opened 2025-05-08
mapbox profile specific voice instructions

OlafFlebbeBosch:
For cycling and hiking use different distances for announcing turn instructions:
For Biking at 150 m (i.e. roughly 500ft) and for hiking at 50m (roughly 150ft) .


OlafFlebbeBosch:
Something is wrong here, there is an unrelated change included, have to check branches :)

navigation/src/main/java/com/graphhopper/navigation/DistanceConfig.java
```diff
@@ -30,22 +30,47 @@ public class DistanceConfig {
...
+            break;
+            case "hiking":
+            if (unit == DistanceUtils.Unit.METRIC) {
+                voiceInstructions =  Arrays.asList(
+                    new ConditionalDistanceVoiceInstructionConfig(IN_LOWER_DISTANCE_PLURAL.metric, translationMap, locale, new int[]{50},
+                        new int[]{50}));
+            } else {
+                voiceInstructions =  Arrays.asList(
+                    new ConditionalDistanceVoiceInstructionConfig(IN_LOWER_DISTANCE_PLURAL.imperial, translationMap, locale, new int[]{50},
+                        new int[]{150}));
```
boldtrn, line 55: Is the 150 a copy/paste issue?

boldtrn (approved):
Having different distance configs depending on the profile makes a lot of sense 👍

OlafFlebbeBosch:
I discovered I was mixing "hiking" and "walking" (correct one is "walking", so I need to update the PR anyway. 
@karussell I get your point. Yes, seems like a good suggestion, however how can the code find out that a given graphhopper Profile name is a "walking" or "cycling" profile when an user chose to use a homegrown profile? I can implement it for the standard profile names "mtb", "ecargobike" et al. , for sure ...

OlafFlebbeBosch:
@karussell  ping ?

karussell:
> however how can the code find out that a given graphhopper Profile name is a "walking" or "cycling" profile when an user chose to use a homegrown profile?

But how is this case supported at the moment? Let's say the name is "super_duper_mtb" for the mapboxProfile then currently it would end up in the wrong `default` branch (?)

To fix this we could try to make the mapping to mapbox profiles configurable and maybe specify an instruction type like bike, car and foot.

> (there is no reverse mapping for resolverMap)

Why would this be necessary?

OlafFlebbeBosch:
Ok, I implemented what I didn't wanted to do.

Assume we have "racingbike", "mtb", "bike" GH profiles : There is no way to specify that they should use the mapboxProfile "biking" for all of them since there is no way to add the key/value pairs 
biking -> racingbike
biking -> mtb
biking -> bike
to resolverMap , at least I dont see how.

IMO we need a map of ghProfile -> mapbox on top or an additional attribute / property on the ghProfile . But how do you want to have it?

karussell:
I think we should use the ghProfile for the DistanceConfig class:

```
public DistanceConfig(DistanceUtils.Unit unit, TranslationMap translationMap, Locale locale, String ghProfile)
```

Then the translation `gh -> mapbox` is not necessary. Just `mapbox -> gh` for doGet and this mapping does already exist.

> at least I dont see how.

Yes, inside DistanceConfig we need to detect if a profile is foot-alike, bike-alike or car-alike (and later maybe truck-alike). But this is independent of the `gh -> mapbox` mapping and would be also necessary for mapbox e.g. `driving-traffic` would be mapped to `driving`.

For now we can make it really simple like we have [in GraphHopper Maps](https://github.com/graphhopper/graphhopper-maps/blob/3d1ff27068663ea638f7f05e5521944e1fa17151/src/api/Api.ts#L438) (basically checks if it contains a certain string):

```js
    public static isFootLike(profile: string) {
        return profile.includes('hike') || profile.includes('foot')
    }

    public static isBikeLike(profile: string) {
        return profile.includes('mtb') || profile.includes('bike')
    }

    public static isMotorVehicle(profile: string) {
        return (
            (profile.includes('car') && !profile.includes('cargobike')) ||
            profile.includes('truck') ||
            profile.includes('scooter') ||
            profile.includes('bus') ||
            profile.includes('motorcycle')
        )
    }

    public static isTruck(profile: string) {
        return profile.includes('truck')
    }
```

> Yes, seems like a good suggestion, however how can the code find out that a given graphhopper Profile name is a "walking" or "cycling" profile when an user chose to use a homegrown profile? 

Probably this is the part I still do not understand. How are you using homegrown profiles with the current approach? Do you have them named like mapbox? (BTW: I would have no problem if we include `profileName.contains("cycling")` for `isBikeLike` and `profileName.contains("walk")` for `isFootLike`.)

Another possibility could be to use a new `navigation-transportation-mode` option that we would specify in the config.yml without string stuff. This would also avoid that we have to introduce a config like `datareader.turn_lanes_profiles: car` for [turn lanes](https://github.com/graphhopper/graphhopper/pull/2954).

OlafFlebbeBosch:
First: It works for us, because we introduced a new endpoint for us since we needed more information than what mapbox provides in the directions response. Than we can use the Response Converter with the Bike DistanceConfig easily. 

Overall consensus internally is that we do not like to rely on profile names. We would like to explore navigation-transport-mode suggestion option, however this is a bit out of scope for this PR and a collegue wants to work on it since he is more experienced in this area. 

Will the work for you, if I revert back to the second commit and we submit a seperate one for extending profiles ?

karussell:
Ok

OlafFlebbeBosch:
Reverted

karussell:
> we submit a seperate one for extending profiles ?

Could you already submit this (based on this PR)?

OlafFlebbeBosch:
Thanks @karololszacki for providing the missing functionality. Since it was easier to handle it as a single PR because of the interdependencies, I merged his contribution in this branch.

karussell:
Thanks a lot @OlafFlebbeBosch @karololszacki !

Now the only thing is the naming. 

Could you change the "mapbox"-alike names like walking, cycling and driving into "GraphHopper TransportationMode"-alike names: foot, bike and car for a more consistent solution?

And also the mapboxProfile parameter should be the graphHopperProfile (in the DistanceConfig constructor) as I suggested earlier.

karololszacki:
But how would that work in the `NavigateResource::doGet` exactly?
With the current approach, it passes `mapboxProfile` (String) directly from the URL param.

Applying your suggestion @karussell I guess it would need to resolve somewhat like:
`mapboxProfile`(the URL param) -> through `resolverMap` to get a `ghProfile` -> determine which NavigationTransportationMode is suitable (by reading the Profile, and through it, its hints).

Seems more convoluted, but should still work for both endpoints (GET and POST) - I'll try to do it today. 
Let me know if I understood the intention right (or wrong!) in the meantime :)

OlafFlebbeBosch:
Merged propsal from @karololszacki for a fix

karussell:
Thanks a lot for the change @OlafFlebbeBosch @karololszacki !

> Seems more convoluted

I prefer to have the naming per default everywhere "GraphHopper"-alike and only where really necessary follow the mapbox naming. And e.g. for the POST request it should use "GraphHopper"-names and therefor utility classes like DistanceConfig should use GraphHopper names. Following this reasoning the names in the config-example.yml should use GraphHopper names too.

I'm not sure the separate NavigationTransportationMode enum is necessary (maybe you mean this with "more convoluted"?). Why can't we use the TransportationMode and do the case-switch statement in the DistanceConfig?

I think even your own custom profile names should be possible with this approach as you could either use the GraphHopper names in the config.yml or use a different mapbox->graphhopper mapping (`profiles_mapbox` entry) in the config (?)

karololszacki:
OK I think this should do it :)

https://github.com/OlafFlebbeBosch/graphhopper/pull/6

OlafFlebbeBosch:
Seem's like @karololszacki  understood the request of @karussell  way better than me :) Added his new proposal to this PR . I guess we should do a squash merge into graphhopper , in order not to show all of our tries to get everything right.

karussell:
@karololszacki nice, thanks a lot!

> in order not to show all of our tries to get everything right

As we squash merge pull requests into master anyway we prefer to keep the pull request commits even if there are many :) (but you can squash it here too, however you prefer)

OlafFlebbeBosch:
Wasn't aware graphhopper project always squashes in master merge. Will not do any changes any more :)

OlafFlebbeBosch:
Oh wait. We found out that we have to fix the email adress in Karols contribution.

OlafFlebbeBosch:
I was able to rewrite the history of this patch only. Force uploading it with correct email adresses

Issue #3157, opened 2025-04-10
vehicle=destination should be preferred over access=private


spixi:
**Describe the bug**
A bicyle road in Germany is usually tagged with "highway=unclassified", "bicycle=designated" and "vehicle=destination". In that case, the algorithm tries to avoid that bicycle road and wants to use private parking side-lanes ("highway=service", "access=private").

**To Reproduce**
https://graphhopper.com/maps/?point=48.312076%2C11.913495_Am+Stadion%2C+85435+Erding%2C+Bayern%2C+Deutschland&point=48.314453%2C11.912604_Volksfestplatz%2C+Sch%C3%BCtzenstra%C3%9Fe%2C+85435+Erding%2C+Bayern%2C+Deutschland&profile=car&layer=Omniscale
https://www.openstreetmap.org/directions?engine=graphhopper_car&route=48.312285%2C11.913536%3B48.315004%2C11.911594#map=18/48.313707/11.914201

**Expected behavior**
The detour over the parking side-lanes should be avoided. Instead the direct way should be preferred.

**Screenshots & Logs**
![Image](https://github.com/user-attachments/assets/ff86bb81-362c-44a4-a3b5-7dca960a8420)
![Image](https://github.com/user-attachments/assets/38d42998-75e9-47ed-b03e-79d057514775)
![Image](https://github.com/user-attachments/assets/3ca306ec-f1d7-4a24-8613-b17e3487f66b)

otbutz:
Please, don't do this:

![Image](https://github.com/user-attachments/assets/6ac45353-8f0d-4a76-a35a-462daeb8c70d)

[Tagging for the renderer](https://wiki.openstreetmap.org/wiki/Tagging_for_the_renderer) is frowned upon in the OSM community.

spixi:
> Please, don't do this:
> 
> ![Image](https://github.com/user-attachments/assets/6ac45353-8f0d-4a76-a35a-462daeb8c70d)
> 
> [Tagging for the renderer](https://wiki.openstreetmap.org/wiki/Tagging_for_the_renderer) is frowned upon in the OSM community.

You propably mean [Tagging for the router"](https://wiki.openstreetmap.org/wiki/Tagging_for_the_router), but how can I avoid this?

otbutz:
This is something that should be fixed in GraphHopper and not in the OSM dataset. The previous tagging was correct as far as I can see. You should revert your changeset.

ratrun:
[...]

PR #3174, opened 2025-06-26
scriptable turn statements: "turn_penalty"

karussell:
This PR implements scriptable statements for turns (currently the turn *penalty*, not turn *time*). I.e. one can now model advanced turn costs and there is no need for the [several turn costs parameters in the config.yml](https://github.com/graphhopper/graphhopper/pull/2957) (see `avoid_turns.json`) and instead you can model much more complex situations like a turn cost for bigger roads or different costs for left hand traffic etc. Furthmore you can try all changes out at query time directly in the custom model. However, it needs to be enabled in the config.yml via `allow_turn_weight_in_request: true` in the `turn_costs` section.

Compared to the turn cost parameters that we currently have there are no downsides, except maybe the initial setup is a bit more complicated, but even query and import speed is comparable or slightly faster for the new scriptable `turn_penalty` statements. Here is the example `avoid_turns.json`:

```
{
  "turn_penalty": [
    // straight:
    { "if": "change_angle > -25 && change_angle < 25", "add":  "0" },
    // right:
    { "else_if": "change_angle >= 25 && change_angle < 80", "add":  "3" },
    // sharp right:
    { "else_if": "change_angle >= 80 && change_angle <= 180", "add":  "3" },
    // left:
    { "else_if": "change_angle <= -25 && change_angle > -80", "add":  "3" },
    // sharp left:
    { "else_if": "change_angle <= -80 && change_angle >= -180", "add":  "3" },
    // uTurnCosts:
    { "else": "", "add":  "Infinity" }
  ]
}
```

🎉 Work is sponsored by [OPTITOOL](https://www.optitool.de/en/) (cc @otbutz)

TODOs:

 - [x] document this in custom-models.md
 - [x] [add to custom model editor](https://github.com/graphhopper/custom-model-editor/pull/9)
 - [x] ~decide on naming `turn_weight` and `add`~ use `turn_penalty` instead of `turn_weight`
 - [x] change CustomModelParser.createWeighting to work with `turn_penalty` too (for tests)
 - [x] the config option in profile `turn_costs` is named differently because the costs consists of the penalty and the time. So this should be fine. Different names with different meaning.
 - [x] move config `allow_turn_penalty_in_request` into TurnCostsConfig

The main driving force behind this feature was to fix the longstanding issue #733 via [the idea mentioned here](https://github.com/graphhopper/graphhopper/issues/733#issuecomment-580927983):

 * it properly avoids destination even for short ways and large detours like [this](https://graphhopper.com/maps/?point=47.41529%2C8.342974&point=47.419353%2C8.368015) but it still allows routing through it if detour would be too large. E.g. [this](https://graphhopper.com/maps/?point=47.414803%2C8.343365&point=47.415427%2C8.34712). (examples from #733). With e.g. `"add": "2500"` it seems to be the right balance for this case (not sure yet, if we should use it for everything else too).
 * of course it still works for normal detours like [this](https://graphhopper.com/maps/?point=60.63293%2C6.41896&point=60.62968%2C6.41025).
 * no detour is taken [here](https://graphhopper.com/maps/?point=51.528177%2C4.404659&point=51.523235%2C4.4135&profile=car) (example from #733) and similarly it avoids strange zigzag tour [like this](https://graphhopper.com/maps/?point=48.312047%2C11.913452&point=48.314942%2C11.912508&profile=car) (example from #3157).

Although there are some downsides with this change the destination-only (or other avoidance) cases behave finally exactly how they should: they are avoided except if a waypoint is on an edge (or in an area) with destination-only access. The acceptable detour can be controlled with the value for `add`. Furthermore this change enables this "advanced avoidance" for other use cases, like `hazmat_water` and `highway=track`. Also it can be used to avoid country border crossings via the custom model without specially crafted polygons.

And in the future we can use the same approach to fix a few problems for short ferries, but we would need to modify the ETA too not only the weight. I.e. instead of `turn_penalty` we would need `turn_time`. With this more realistic ferry times are possible.

Downsides of this approach compared to just a low priority for destination-only cases:

 * enabling turn restriction is required and if CH is used this means a much slower preparation (compared to without turn restrictions). But turn restriction and turn cost support is so essential for car (where this feature is most important) this is acceptable as `turn_costs` support is already enabled.
 * it slows down the (edge-based) CH preparation a bit (compared to just turn restrictions without this feature). But on the other hand there might be less shortcuts and so the query speed might improve
 * minor: a missing `private` (or `destination`) tagging inside such an private (or destination-only) area will become very obvious. E.g. for `point=51.11853%2C13.61062&point=51.12035%2C13.60681` but it is easy to fix this IMO tagging mistake.
    
 * a similar minor problem (but no tagging mistake!) is in cases like for this barrier where the ways are private but the [barrier itself is not private](https://www.openstreetmap.org/node/3356938357) and leads to a wrong detour nevertheless:
    image
    This is likely detectable and fixable via some pre-processing.

Instead of the `turn_penalty` we could add barrier edges whenever there is a change e.g. from road access "yes" to "destination". But we would have to pre-process the data and create a boolean encoded value that is true whenever there is such a change and use `add` in `priority`. This would avoid that we need to enable turn costs, but then these cases need to be known before the import (and a different solution would be required for "per-request") and also the implementation of it sounds trickier, also regarding multiple edges which needs to be joined or junctions with different values.

Note: I slightly adjusted the maxPollFactorHeuristic in EdgeBasedNodeContractor for ~10% faster preparation (compared to maxPollFactorHeuristic=5). With this the CH preparation compared to just turn restrictions is still 17% slower (95k sec vs 111k sec with private/destination avoidance; planet; car)


otbutz:
I'm on vacation next week but I'm looking forward to providing feedback once I'm back. :slightly_smiling_face:

michaz:
I like this.

michaz:
My opinions on the wording question:

- Cost and weight are synonyms. For edges and routes (aka paths), we've always called it weight. For turns, we've mostly called it cost, and sometimes weight, mostly when it appears right next to an edge weight. I think I always call it [cost](https://en.wikipedia.org/wiki/Generalised_cost) when it's about transportation, and weight when I'm inside a routing algorithm or something. If, in the config file and similar places, we've only ever called it cost so far (not sure), we should perhaps keep it that way.
- We need to be a bit careful now that we give direct access to the cost: It must be clear to me what happens to the cost when I change the travel time. Do I also have to add it there? Or does the cost automatically contain the travel time component, including for any extra travel time I add here?
- I think it may be convenient to say that the cost that I add here is an **extra** cost, **on top** of the travel time component, which it automatically contains. So in the model specification, e.g. for your turn angles, you would perhaps just add time (and it would still make the route more costly). And for the "destination" penalty, you would just add cost (and it would not make the route take longer).
- Wait, should we say that the user can "specify a penalty" instead of "add to the cost"?
- Then what we would have is: For turns: `cost = time + penalty`, and for links: `cost = time + distance_influence * distance + whatever_priority_does_exactly * time (or was it distance)`
- It's yet another word, but it's something we don't have so far. And which sooner or later we may also want to have for links.
- Unless for links I would **also** get direct access to the cost, so that I can say `cost: if bla add 500` (adding to whatever travel time and priority are already doing), or if I insist `cost: if blubb set_to 10000` (overwriting anything travel time and priority are already doing), but maybe that would be unnecessarily powerful and really an additive component, penalty, is all I should get?
- But we could have the same without a new word by allowing only `add`. So it's probably all good.

michaz:
Though I still maintain that, at least for road transportation, a 90° turn is a turn of regular sharpness, not a sharp turn. 🤪 But now that they aren't prescribed categories anymore, I shouldn't care that much.

michaz:
Let's do a quick survey where we say "cost" and where we say "weight" at the moment.

michaz:
And I still think it's important to be clear that turn costs in the narrower sense of e.g. "discourage left turns" are not about angles.

When some road goes left, even sharply, that's not a turn, that's just a bend.

A left turn is (in many countries) when there's conflict with oncoming traffic, and you don't see that from the angle.

Just like or even more than with estimating link travel time, I'm almost sure this **ultimately** isn't scriptable by looking just at local attributes. You will quickly want to look at the entire node with **all** its edges, and, for complicated intersections, at the entire complicated intersection.

michaz:
As a solution to the destination-restriction problem, it just sounds really, really good.

config-example.yml
```diff
@@ -74,7 +82,7 @@ graphhopper:
   # more RAM/disk space for holding the prepared graph but also means less memory usage per request. Using the following
   # list you can define for which of the above routing profiles such preparation shall be performed. Note that to support
   # profiles with `turn_costs` a more elaborate preparation is required (longer preparation time and more memory
-  # usage) and the routing will also be slower than with `turn_costs: false`.
+  # usage) and the routing will also be slower than with `turn_costs`.
```
otbutz:
```suggestion
  # usage) and the routing will also be slower than without `turn_costs`.
```

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java
```diff
@@ -100,4 +99,22 @@ public static boolean in(Polygon p, EdgeIteratorState edge) {
...
+
+    public static double calcChangeAngle(EdgeIntAccess edgeIntAccess, DecimalEncodedValue orientationEnc,
+                                         int inEdge, boolean inEdgeReverse, int outEdge, boolean outEdgeReverse) {
+
+        double prevAzimuth = orientationEnc.getDecimal(inEdgeReverse, inEdge, edgeIntAccess);
+        double azimuth = orientationEnc.getDecimal(outEdgeReverse, outEdge, edgeIntAccess);
+
+        // bring parallel to prevOrientation
+        if (azimuth >= 180) azimuth -= 180;
+        else azimuth += 180;
```
otbutz:
```suggestion
        azimuth = azimuth % 180;
```

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java
```diff
@@ -100,4 +99,22 @@ public static boolean in(Polygon p, EdgeIteratorState edge) {
...
+
+        double changeAngle = azimuth - prevAzimuth;
+
+        // keep in [-180, 180]
+        if (changeAngle > 180) changeAngle -= 360;
+        else if (changeAngle < -180) changeAngle += 360;
+        return changeAngle;
```
otbutz:
```suggestion
        return (changeAngle + 180) % 360 - 180;
```

core/src/main/java/com/graphhopper/routing/weighting/custom/ValueExpressionVisitor.java
```diff
@@ -37,6 +37,7 @@
  */
 public class ValueExpressionVisitor implements Visitor.AtomVisitor<Boolean, Exception> {
 
+    private static final String INFINITY = "Infinity"; // can be parsed via Double.parseDouble
```
otbutz:
```suggestion
    private static final String INFINITY = Double.toString(Double.POSITIVE_INFINITY);
```

core/src/main/resources/com/graphhopper/custom_models/truck.json
```diff
@@ -11,6 +11,7 @@
   "priority": [
     { "if": "hgv == NO",  "multiply_by": "0" },
     { "if": "!car_access && road_access != PRIVATE && hgv != DELIVERY && hgv != DESTINATION",  "multiply_by": "0" },
+    // also have a look into avoid_private_etc.json
```
otbutz, line 14: Should we follow the car custom model and provide a `avoid_delivery.json` ?

otbutz (changes requested):

otbutz:
I haven't tested the angle calculation changes, but this should be doable without conditions and only using modulo operations.

core/src/main/java/com/graphhopper/routing/weighting/custom/CustomWeightingHelper.java
karussell: Thanks! Unfortunately the proposals won't work. The azimuth won't work for values <180 (it won't add 180 obviously) and the changeAngle won't work for negative values it seems (you can run CustomWeightingHelperTest to see what I mean).

But the following branchless code seems to work:

```java
// bring parallel to prevOrientation
azimuth = (azimuth + 180) % 360.0;

double changeAngle = azimuth - prevAzimuth;

// keep in [-180, 180]
return (changeAngle + 540.0) % 360.0 - 180.0;
```

So thanks for the suggestion :) !

otbutz:
Works :slightly_smiling_face: 

## master (multiply with 0.1 for private)

grafik

## PR (add 300 for private)

grafik


https://graphhopper.com/maps/?point=48.58223%2C13.396052&point=48.583599%2C13.394713&profile=truck&layer=OpenStreetMap

otbutz:
And if you think the previous screenshot from the current master doesn't look that bad:

grafik

The effect of virtually inflating the edge sizes is quite apparent. :sweat_smile: 

https://graphhopper.com/maps/?point=48.582409%2C13.395431&point=48.583599%2C13.394713&profile=truck&layer=OpenStreetMap

PR #3176, opened 2025-07-04
fix: enable last intersection of last step of via points

OlafFlebbeBosch:
We noticed that intersections are not populated on the arrival intersection on both OSRM and mapbox directions creating issues to pass VIA points in navigation. 

-    rename instructionJson to stepJson, since it is about the step Json object
-    Adding an ManeuverType enum to better document the Maneuver we are going to create
-    Handle ManeuverTypes and TurnTypes differently, since they are different concepts, sharing some values but not all
-    Fix typo bearingsArray vs bearingsrray
-    rename isFirstInstructionOfLeg to isDepartInstruction and use maneuverType in the subroutines instead
-    rename fixFirstIntersectionDetail to fixDepartIntersectionDetail since intersections have to be fixed on every leg
-    fix one test setup and one test

### DIFF
--- a/navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
+++ b/navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
@@ -38,6 +38,13 @@ import java.util.stream.Stream;
 
 import static com.graphhopper.util.Parameters.Details.INTERSECTION;
 
+enum ManeuverType {
+    ARRIVE,
+    DEPART,
+    TURN,
+    ROUNDABOUT
+} ;
+
 public class NavigateResponseConverter {
     private static final Logger LOGGER = LoggerFactory.getLogger(NavigateResponseConverter.class);
     private static final int VOICE_INSTRUCTION_MERGE_TRESHHOLD = 100;
@@ -92,12 +99,11 @@ public class NavigateResponseConverter {
 
         long time = 0;
         double distance = 0;
-        boolean isFirstInstructionOfLeg = true;
+        boolean isDepartInstruction = true;
         int pointIndexFrom = 0;
 
         Map<String, List<PathDetail>> pathDetails = path.getPathDetails();
         List<PathDetail> intersectionDetails = pathDetails.getOrDefault(INTERSECTION, Collections.emptyList());
-        fixFirstIntersectionDetail(intersectionDetails);
 
         ObjectNode annotation = null;
         ArrayNode maxSpeedArray = null;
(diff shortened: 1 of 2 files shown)


navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
```diff
@@ -205,29 +226,29 @@ private static void putLegInformation(ObjectNode legJson, ResponsePath path, int
      * The first Intersection of the first step should only have one "bearings" and one
```
boldtrn: The comments should be updated as well :)

navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
```diff
@@ -205,29 +226,29 @@ private static void putLegInformation(ObjectNode legJson, ResponsePath path, int
...
      * "out" entry
      */
-    private static void fixFirstIntersectionDetail(List<PathDetail> intersectionDetails) {
+    private static void fixDepartIntersectionDetail(List<PathDetail> intersectionDetails, int pos) {
```
boldtrn: I don't like the name pos, can you rename this to something that explains what pos is?

navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java
```diff
@@ -576,50 +596,51 @@ private static void putSingleBannerInstruction(Instruction instruction, Locale l
...
         ObjectNode maneuver = instructionJson.putObject("maneuver");
         maneuver.put("bearing_after", 0);
         maneuver.put("bearing_before", 0);
 
         PointList points = instruction.getPoints();
         putLocation(points.getLat(0), points.getLon(0), maneuver);
 
-        String modifier = getModifier(instruction);
-        if (modifier != null)
-            maneuver.put("modifier", modifier);
```
boldtrn, line 587: Why did you remove the modifier here? This is now missing.

For example:
```
"maneuver": {
  "bearing_after": 0,
  "bearing_before": 0,
  "location": [
    13.354213,
    52.499177
  ],
  "type": "turn",
  "instruction": "turn left onto Bülowstraße"
},
```

How to know if we should turn left or right if the modifier is missing?

Here is the corresponding json response from Mapbox/OSRM:
```
"maneuver": {
  "type": "turn",
  "instruction": "Turn left onto Bülowstraße.",
  "modifier": "left",
  "bearing_after": 94,
  "bearing_before": 154,
  "location": [
    13.354213,
    52.499176
  ]
```

OlafFlebbeBosch: Oops, thank you. Something went wrong when upgrading to latest graphhoppper

boldtrn (approved):
LGTM, just gave this another try with Maplibre after the latest change.

OlafFlebbeBosch:
Rebased and fixed an edge case

navigation/src/test/java/com/graphhopper/navigation/NavigateResponseConverterTest.java
```diff
@@ -548,7 +558,7 @@ public void testMultipleWaypoints() {
         request.addPoint(new GHPoint(42.504776, 1.527209));
         request.addPoint(new GHPoint(42.505144, 1.526113));
         request.addPoint(new GHPoint(42.50529, 1.527218));
-        request.setProfile(profile);
+        request.setProfile(profile).setPathDetails(Collections.singletonList("intersection"));
```
karussell, line 559: minor: can you check the formatting here and elsewhere in this file?

OlafFlebbeBosch: done

karussell:
There is a now test failure in master. And if I change `null` to "straight":

```
assertEquals("straight", maneuver.get("modifier").asText());
```

it runs fine. Is this expected or did I something wrong when merging?

boldtrn:
> There is a now test failure in master. And if I change null to "straight":

Actually depart usually has no modifier IIRC. I am not sure this will break anything, if there is a modifier - I would not expect it to (at least for Maplibre I would expect the impact to be rather low)

OlafFlebbeBosch:
Sorry, I am not sure why this slipped through. And agree to @boldtrn , it shouldn't have a modifier. Looking into why that happend

karussell:
Hm, this is now just as it was before the merge of this PR:

https://github.com/graphhopper/graphhopper/blob/a148d512442e293c354553b37f0c9fb866ef296b/navigation/src/test/java/com/graphhopper/navigation/NavigateResponseConverterTest.java#L85

boldtrn:
> Hm, this is now just as it was before the merge of this PR

Ah sorry, I must have missed this, OK, if we have always added a modifier here, then this is probably not an issue, so all good from my side then.

oflebbe:
Looks like i introduced the change with a previous attempt fixing and somehow the test didnt trigger w/o a mvn clean on my side

PR #3177, opened 2025-07-07
Sort graph for better cache locality

easbar:
This PR sorts the graph along a Hilbert curve during import. This speeds up routing, but also the node-based CH preparation. The sorting is done in-place but still requires some extra memory during import: eight bytes per node, so around 4GB for planet OSM. However, the overall memory requirement might be unchanged since other steps of the import like location index building and CH/LM preparation require more memory anyway. In any case the sorting can be disabled by setting `graph.sort: false` in the config file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - road_access now contains value of highest transportation mode for CAR, i.e. access=private, motorcar=yes will now return YES and not PRIVATE
 - car.json by default avoids private roads
 - maxspeed<5 is ignored, maxspeed=none is ignored with some exceptions, maxspeed parsing and related constants were renamed #3077
+- improved performance by sorting graph during import, #3177
 
 ### 10.0 [5 Nov 2024]
 


otbutz:
https://discuss.graphhopper.com/t/improving-memory-locality/9412

Nice!

easbar:
Here I measured the routing speed for the car profile (without turn costs, no instructions and `calc_points=false`) on a planet import using random queries (for 0.5km and 20km beeline distance between start and target points) for increasing Hilbert order (=increasing sorting precision):

image

The y-axis shows the routing time relative to the time I measured for the unsorted graph in % (100=no change). The connection between routing speed and Hilbert order can be seen very clearly. The more we sort the graph the faster the routing becomes, and at some point we hit a plateau where further sorting has no advantage. This depends on the routing distance, but since we use Hilbert order 31 we use a sufficiently high precision anyway. The dashed horizontal lines show the relative routing time for a randomized sorting order (much slower) and DFS sorting that we used previously (#2919) (slower, at least in this experiment).

otbutz:
So, if I've interpreted your graph correctly, we're talking about a ~20% improvement for routes of more than 20 km?

easbar:
For the measurement I described, yes. For the car profile with turn costs it was more like a 13% improvement. I did not do further experiments with instructions or calc_points=true, but since there is basically no downside to the sorting I think it is worth enabling it by default anyway. I'd be happy to hear how this sorting affects performance in everyone's GraphHopper setup :)

PR #3178, opened 2025-07-14
Allow freezing TurnCostStorage to remove TC_NEXT

easbar:
When we are done adding turn cost entries we can sort the entries such that no TC_NEXT pointer is required anymore. This saves four bytes per turn cost entry and turns the linked list iteration into a simple loop.

However, since we use this storage only for OSM turn restrictions currently this is no real improvement: The overall memory requirement for turn restrictions is small, so saving memory is not important. Also for the majority of nodes there are no turn cost entries at all such that the linked list iteration is not an issue, either. For nodes without turn cost entries the TC_NEXT pointer is even more efficient since it requires only a single read, while the loop method requires two reads. I could not measure any difference in routing speed between the two methods for an OSM planet import.

Keeping this here only for later reference as it introduces some complexity and might only be an improvement if a significant number of turn cost entries were needed.

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/TurnCostStorage.java
+++ b/core/src/main/java/com/graphhopper/storage/TurnCostStorage.java
@@ -38,16 +38,20 @@ import java.util.function.IntUnaryOperator;
  */
 public class TurnCostStorage {
     static final int NO_TURN_ENTRY = -1;
-    // we store each turn cost entry in the format |from_edge|to_edge|flags|next|. each entry has 4 bytes -> 16 bytes total
+    // turn cost entry format:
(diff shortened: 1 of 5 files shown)


easbar:
Related: #3177 

sort graph, freeze turn cost storage, remove turn cost pointer, turn cost storage memory

otbutz:
> Keeping this here only for later reference as it introduces some complexity and might only be an improvement if a significant number of turn cost entries were needed.

#3174 might introduce more entries?

easbar:
> https://github.com/graphhopper/graphhopper/pull/3174 might introduce more entries?

Probably not, because the turn weights can be derived from the edges' encoded values and do not need to be stored explicitly.

Issue #2823, opened 2023-05-20
Sonny's LiDAR Digital Terrain Models of Europe


SonnyLidarDTM:
As far as I know Graphhopper is supporting elevation data used from external elevation providers.

So maybe you wanna try **_"Sonny's LiDAR Digital Terrain Models"_** at https://sonny.4lima.de/ as an alternative within Europe?

The accuracy and resolution of its models' elevations compared to satellite-based surveys like ASTER, ALOS, COPERNICUS or SRTM is significantly better. Especially in wooded areas, steep rocky terrain or narrow valleys. File format (.hgt) and file size stays the same as used in SRTM.

otbutz:
@karussell this dataset looks really good to be honest.

https://github.com/abrensch/brouter/issues/539#issuecomment-1540621266

boldtrn:
One of the sources for Sonny was viewfinderpanoramas? 

IIRC, there is some potential licensing issue with viewfinderpanorams?

> [Viewfinder Panoramas](http://www.viewfinderpanoramas.org/dem3.html). The data is not under an open license but can be freely used in open source projects according to [a note here](http://igorbrejc.net/openstreetmap/viewfinders-dem-data-comparison-with-srtm) based from the author's response. However, some of the data was obtained from Soviet topographical maps that is not available for licensing, according to a [blog post comment](http://igorbrejc.net/openstreetmap/viewfinders-dem-data-comparison-with-srtm#comment-36413). Please use with caution.

https://wiki.openstreetmap.org/wiki/SRTM

I am not sure if this would be relevant, but wanted to mention it. Great work on creating a great DTM @SonnyLidarDTM 👍

ratrun:
This issue can be closed, support was added in #3183.

PR #3183, opened 2025-08-02
Adding support for Sonny's LiDAR 1" elevation provider

ratrun:
For the use-case of a not public graphhopper instance, I added support for [Sonny's LiDAR Digital Terrain Models of Europe](https://sonny.4lima.de/) as elevation provider. The advantage is that for the covered area in Europe the DTM data is much better compared to CGIAR data. The disadvantage is that the data is [**not free** to use](https://github.com/graphhopper/graphhopper/issues/2823)!

Although I expect that this won't get merged due to the licensing issues and because the data does not get automatically fetched, I decided to create this public pull request such that this work does not get lost. It still could be useful for certain scenarios.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -723,6 +723,10 @@ public class GraphHopper {
             elevationProvider = new MultiSourceElevationProvider(cacheDirStr);
         } else if (eleProviderStr.equalsIgnoreCase("skadi")) {
             elevationProvider = new SkadiProvider(cacheDirStr);
+        } else if (eleProviderStr.equalsIgnoreCase("sonny")) {
+            elevationProvider = new SonnyProvider(cacheDirStr);
+        } else if (eleProviderStr.equalsIgnoreCase("multi3")) {
+            elevationProvider = new MultiSource3ElevationProvider(cacheDirStr);
         }
 
         if (elevationProvider instanceof TileBasedElevationProvider) {
--- a/core/src/main/java/com/graphhopper/reader/dem/MultiSource3ElevationProvider.java
+++ b/core/src/main/java/com/graphhopper/reader/dem/MultiSource3ElevationProvider.java
@@ -1,0 +1,127 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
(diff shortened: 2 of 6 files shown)


mprins:
> to the licensing issues and because the data does not get automatically fetched

Why would that be a problem? Isn't that up to the user that downloads the extra data

karussell:
> Why would that be a problem? Isn't that up to the user that downloads the extra data

IMO it would be similar to using a closed source library (?) But maybe not that bad when it is not used by default and properly documented?

Or, what if we could define a list of polygons to exclude these areas (as part of this elevation provider?) and use some fallback?

core/src/test/java/com/graphhopper/reader/dem/MultiSourceElevationProviderTest.java
```diff
@@ -87,6 +87,6 @@ public void testGetEle() {
         assertEquals(108, instance.getEle(60.0000001, 16), precision);
         assertEquals(0, instance.getEle(60.0000001, 19), precision);
         // Stor Roten
-        assertEquals(14, instance.getEle(60.251, 18.805), precision);
+        assertEquals(4, instance.getEle(60.251, 18.805), precision);
```
karussell: Can you move this change out into a separate PR? I think it is just the fix in core/MultiSourceElevationProvider.java which is unrelated to this PR?

ratrun: >> I think it is just the fix in core/MultiSourceElevationProvider.java which is unrelated to this PR?

This is true. It was not detected previously because this test is disabled. I'll create a separate PR after the merge if I don't forget :smiling_face_with_tear:

ratrun:
> Did you test it already in production?

Well, as graphhopper is just a hobby of mine I won't call it "production", but I did plan a 10 days bicycle tour in covered parts of Europe in hilly terrain and didn't notice problems.

karussell:
Sorry, meant more "real world" than "production" :)

ratrun:
@karussell I don't know how you managed to create commit 6168f2c, but thank you. Now there is no need any longer for fixing the disabled test.

karussell:
> @karussell I don't know how you managed to create commit https://github.com/graphhopper/graphhopper/commit/6168f2c257827aaa6a8f6c14f2379a6a6e361fe8, but thank you.

It is an **unsigned** commit with your changes and you as the author :) (`git commit -m "..." --author "ratrun <your@email.com>`)

PR #3185, opened 2025-08-04
Increase used JTS version to 1.20.0

ratrun:
See https://github.com/locationtech/jts/re…leases/tag/1.20.0
In my testing this improves the graph preparation performance by about 6%.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,7 +98,7 @@
             <dependency>
                 <groupId>org.locationtech.jts</groupId>
                 <artifactId>jts-core</artifactId>
-                <version>1.19.0</version>
+                <version>1.20.0</version>
             </dependency>
             <dependency>
                 <groupId>org.apache.commons</groupId>


karussell:
Thanks!

The update needs to be done for our fork [jackson-datatype-jts](https://github.com/graphhopper/jackson-datatype-jts) too (see #1803). Have now released version 2.19.2 of this - should be (soonish) available.

And to make it consistent we have to update jackson to 2.19.2, i.e. update [dropwizard to 3.0.15](https://github.com/dropwizard/dropwizard/blob/8a52fe4aea45f84571b33eb173735802e20bbc64/dropwizard-dependencies/pom.xml).

ratrun:
Thank you for providing the jackson-datatype-jts version 2.19.2, now we are using it. The local tests are all green, but I haven't run another benchmark yet.

karussell:
Thanks!

I have merged it and also included the dropwizard dep.

PR #3186, opened 2025-08-06
Allow bike_road_access==PRIVATE for foot and bicycle profiles

ratrun:
The result of the [discussion in the forum](https://discuss.graphhopper.com/t/correct-ferry-terminal-access-tagging/9655/1) was that we should not block ways tagged as `access=customers` by default as this may block access to ferries for example. 

As `access=customers` currently gets turned into a `bike_road_access=PRIVATE` encoded value we need to prevent assigning priority=0 in the custom models as this blocks such ways completely. This PR changes the priority for `bike_road_access=PRIVATE` into the lowest possible value together with assigning walking speed instead.

Additionally, this PR improves the documentation about the currently undocumented  `block_private=false` encoding value option.

By explicitly configuring `block_private=false` in the config.yml it is possible to allow access additionally also for private roads. If this parameter is missing in the config, access is only provided to ways with `access` tags set to `delivery`, `customer` or `permit`.

### DIFF
--- a/config-example.yml
+++ b/config-example.yml
@@ -64,7 +64,7 @@ graphhopper:
 #
 #   Other custom models not listed here are: car4wd.json, motorcycle.json, truck.json or cargo-bike.json. You might need to modify and test them before production usage.
 #   See ./core/src/main/resources/com/graphhopper/custom_models and let us know if you customize them, improve them or create new onces!
-#   Also there is the curvature.json custom model which might be useful for a motorcyle profile or the opposite for a truck profile.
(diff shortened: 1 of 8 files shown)


karussell:
Please have a look into the commit where I included your changes.

ratrun:
> Please have a look into the commit where I included your changes.

Looks good, thank you. I just detected one issue in foot.json which probably was not intentional: 
```
{ "if": "foot_road_access == PRIVATE", "multiply_by": "0" },
{ "if": "foot_road_access == PRIVATE", "multiply_by": "0.1" }
```

The first line should be deleted, shouldn't it?

PR #3191, opened 2025-08-25
Trunk roads are not generally toll roads in Austria

### LINKED ISSUE #3190: Not all trunk roads in Austria are tool roads
**Describe the bug**

In Austria, a decent amount of toll-free motorroads are tagged as `highway=trunk`. Examples:

* https://www.openstreetmap.org/way/555295671
* https://www.openstreetmap.org/way/200569645
* https://www.openstreetmap.org/way/26796738

As a general rule, [in Austria](https://www.asfinag.at/maut-vignette/vignette/), only the roads with a `ref` that starts with `S` (or `A`) require toll. Those are tagged with `highway=trunk + toll=yes` (or `highway=motorway`). Here are some examples:

* https://www.openstreetmap.org/way/148348424
* https://www.openstreetmap.org/way/208434582
* https://www.openstreetmap.org/way/26803745
* https://www.openstreetmap.org/way/203977975

**To Reproduce**

* https://graphhopper.com/maps/?point=47.559236%2C10.65746&point=47.233482%2C10.744503&profile=car&layer=Omniscale
* note that graphhopper falsely reports that the _“route has tolls”_, while in reality it does not

**Expected behavior**

Graphhopper should only consider `highway=trunk` roads in Austria as toll roads if they are also tagged with `toll=yes`.

tyrasd:
fixes #3190

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTollParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTollParser.java
@@ -99,7 +99,7 @@ public class OSMTollParser implements TagParser {
                 else
                     return Toll.NO;
             }
-            case AUT, BGR, CZE, FRA, GRC, HRV, ITA, PRT, SRB, ESP -> {
+            case BGR, CZE, FRA, GRC, HRV, ITA, PRT, SRB, ESP -> {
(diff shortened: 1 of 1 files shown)


easbar:
Thanks for the analysis and fix! I added you to the list of contributors here: 6aea94251e7bdf162efaf08c73003fef97a7042f

PR #3197, opened 2025-09-02
Add support for profile parameter to NearestResource

Nakaner:
This pull request adds support for the parameters `profile` and `snap_preventions` to the `/nearest` endpoint.

The endpoint is currently not official. One of our clients faces the issue of isochrones not including the requested start location because their request point is a highway rest area (island for pedestrian, eliminated during subnetwork removal). I would like to point them to the `/nearest` endpoint which returns the closest snap on the graph. However, that endpoint does not take profiles into account.

This pull request ignores custom models because they require HTTP POST requests but `/nearest` accepts GET requests only.

### DIFF
--- a/web-bundle/src/main/java/com/graphhopper/resources/NearestResource.java
+++ b/web-bundle/src/main/java/com/graphhopper/resources/NearestResource.java
@@ -20,7 +20,18 @@ package com.graphhopper.resources;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.graphhopper.jackson.MultiException;
+import com.graphhopper.GHRequest;
+import com.graphhopper.GraphHopper;
+import com.graphhopper.GraphHopperConfig;
+import com.graphhopper.config.Profile;
+import com.graphhopper.routing.ev.EnumEncodedValue;
+import com.graphhopper.routing.ev.RoadClass;
+import com.graphhopper.routing.ev.RoadEnvironment;
+import com.graphhopper.routing.ev.Subnetwork;
+import com.graphhopper.routing.util.DefaultSnapFilter;
 import com.graphhopper.routing.util.EdgeFilter;
+import com.graphhopper.routing.util.SnapPreventionEdgeFilter;
+import com.graphhopper.routing.weighting.Weighting;
 import com.graphhopper.storage.index.LocationIndex;
 import com.graphhopper.storage.index.Snap;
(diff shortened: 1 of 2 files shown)


karussell:
I'm not that happy with that endpoint to be honest and we should remove it, I think.

Why not use the /route endpoint with two identical coordinates? It already supports snap preventions and custom models.

CeallachKN:
I've been thinking about a Use Case for this endpoint lately and there has been one thing on my mind that would be partially solved by this PR.

One thing that always bugged me for the starting point of a route (and thats true for all routing engines i used / tested so far) has always been that it is basically impossible to check if the snapped starting point is even reachable from the user provided point. basically a situation like this
https://graphhopper.com/maps/?point=48.520622%2C12.114217&point=48.493227%2C12.024309&profile=bike&layer=Omniscale
where the snapped point is placed across an impassable object like a river (cliffs, fences, railways etc.).

Since the route endpoint should always provide only one snapped point, this endpoint could be extended to provide a list of candidates to open up the possibility for consumers to start some sort of spatial analysis (since the routinggraph can't contain obstacles like the ones mentioned above) to alter their provided endpoint for the route calculation.

That's far outside the scope of this PR (but would make even less sense for me without the possibility to use profiles) and maybe a solution without a customer (since at least a real user can alter his starting point afterwards), but i wanted to at least give another perspective for this kind of endpoint instead of just removing it.

karussell:
@Nakaner Can you clarify what the disadvantages of using the /route endpoint would be?

> has always been that it is basically impossible to check if the snapped starting point is even reachable from the user provided point. 

It is not possible without adding all the data for costlines, rivers, trains, lakes, houses, ... and even if you have all the data: you can never be sure if there is no 100% match for the coordinates.

Nakaner:
@karussell wrote:

> @Nakaner Can you clarify what the disadvantages of using the /route endpoint would be?

The client uses the /isochrone endpoint for their use case (finding POIs in walking distance of a charging stations). Sometimes, their starting point is not reachable by foot and therefore, the /isochorone endpoint snaps onto a completely different edge.

karussell:
> Sometimes, their starting point is not reachable by foot and therefore, the /isochorone endpoint snaps onto a completely different edge.

but why can't you use the /route endpoint with two identical points in such a case?

Nakaner:
> but why can't you use the /route endpoint with two identical points in such a case?

Thank you for the hint.

karussell:
And do you agree that we could even remove the /nearest endpoint in a later release and point to using /route with two identical points?

PR #3207, opened 2025-10-12
fix wrong mode in navigation steps

### LINKED ISSUE #3206: /navigate reports incorrect mode
"mode" should not be "driving" for bike or foot profiles:

https://github.com/graphhopper/graphhopper/blob/master/navigation/src/main/java/com/graphhopper/navigation/NavigateResponseConverter.java#L422

binora:
Fixes #3206 . The idea here is to add `mode` to `DistanceConfig` class so as to leverage existing logic for mode determination. Please let me know if you have a different approach in mind.


karussell:
Thanks for this! Could you reduce the testing code a bit? E.g. to just a single test that the "driving" is fixed to "walking" and "cycling"? (usually I'm for many tests but in this case you'd only test the switch statement, I think :) )

binora:
sorry I missed this. thanks for helping out!

karussell:
No worries. A new release is coming out and I wanted to have this PR in there :)

PR #3209, opened 2025-10-14
allow null values in headings array in POST /route

### LINKED ISSUE #3192: headings should accept `null` for POST
Currently to specify a heading for just a single point requires to specify NaN für the other points, i.e. for the POST request this is something like `"headings":[10, "NaN"]` which is too ugly (but required as JSON does not accept NaN without quotes) and we should allow `null` which makes more sense: `"headings":[10, null]` but currently triggers an exception.

binora:
Fixes #3192 .  Wrote a custom deserializer for parsing headings in POST /route. Open to other ideas.


karussell:
For simplicity I have used a version without a deserializer. See https://github.com/graphhopper/graphhopper/commit/8fe2ba0002c07ad0433338f15f67bd4638af1729

PR #3214, opened 2025-10-17
Improve bike speed handling

karussell:
This change makes the bike speed more consistent across the different bike types. [Here a link](https://github.com/graphhopper/graphhopper/compare/3e016ecdad...0f5387b3808) for easier comparison as the tests were restructured a bit (i.e. the link avoids the first two commits and the master commits).

Furthermore it tries to make the decision when to pick which speed more consistent. And the theory is that there are 4 cases bicycles=yes/no & surface=good/bad using different default assumptions:

1. there are highways like tertiary, primary, ... where we assume by default: a) bicycles allowed and b) good surface
2. for `track` we assume by default a) bicycles allowed and b) **bad surface**
3. for `footway`, `pedestrian` and `platform` we assume by default a) bicycles **not allowed** and b) good surface
4. for `path` and `bridleway` we assume by default a) bicycles **not allowed** and b) **bad surface**

And for cases 2. to 4. the speed can increase, because of allowed/designated bike access or a good surface.

Furthermore "bike designated" is also true if the way is part of an official bike network (taken from the relationFlags).

See previous discussion [here](https://discuss.graphhopper.com/t/racingbike-vs-bike/6554/5?u=karussell).

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/MountainBikeAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/MountainBikeAverageSpeedParser.java
@@ -20,7 +20,7 @@ public class MountainBikeAverageSpeedParser extends BikeCommonAverageSpeedParser
         setTrackTypeSpeed("grade2", 16); // now unpaved ...
         setTrackTypeSpeed("grade3", 12);
         setTrackTypeSpeed("grade4", 8);
-        setTrackTypeSpeed("grade5", 6); // like sand/grass
+        setTrackTypeSpeed("grade5", PUSHING_SECTION_SPEED); // like sand
 
         // +4km/h on certain surfaces (max 16km/h) due to wide MTB tires
         setSurfaceSpeed("dirt", 14);
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
@@ -37,7 +37,7 @@ public abstract class BikeCommonAverageSpeedParser extends AbstractAverageSpeedP
         setTrackTypeSpeed("grade2", 12); // now unpaved ...
         setTrackTypeSpeed("grade3", 8);
         setTrackTypeSpeed("grade4", 6);
-        setTrackTypeSpeed("grade5", 4); // like sand/grass
+        setTrackTypeSpeed("grade5", PUSHING_SECTION_SPEED); // like sand
 
         setSurfaceSpeed("paved", 18);
         setSurfaceSpeed("asphalt", 18);
--- a/core/src/test/java/com/graphhopper/routing/util/parsers/MountainBikeTagParserTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/parsers/MountainBikeTagParserTest.java
@@ -121,11 +121,11 @@ public class MountainBikeTagParserTest extends AbstractBikeTagParserTester {
 
         way.clearTags();
         way.setTag("highway", "track");
-        way.setTag("tracktype", "grade5");
-        assertEquals(6, getSpeedFromFlags(way), 0.01);
+        way.setTag("tracktype", "grade4");
+        assertEquals(8, getSpeedFromFlags(way), 0.01);
 
         way.setTag("smoothness", "bad");
-        assertEquals(4, getSpeedFromFlags(way), 0.01);
+        assertEquals(6, getSpeedFromFlags(way), 0.01);
 
         way.setTag("smoothness", "impassable");
         assertEquals(MIN_SPEED, getSpeedFromFlags(way), 0.01);


core/src/main/java/com/graphhopper/routing/util/parsers/MountainBikeAverageSpeedParser.java
```diff
@@ -7,38 +7,28 @@ public class MountainBikeAverageSpeedParser extends BikeCommonAverageSpeedParser
...
-        setSurfaceSpeed("salt", 12);
-        setSurfaceSpeed("sand", 10);
-        setSurfaceSpeed("wood", 10);
-
-        setHighwaySpeed("path", 18);
-        setHighwaySpeed("footway", PUSHING_SECTION_SPEED);
-        setHighwaySpeed("track", 18);
-        setHighwaySpeed("residential", 16);
+        setSurfaceSpeed("grass", 12);
+        setSurfaceSpeed("grass_paver", 12);
```
ratrun, line 34: I would add 
```
setSurfaceSpeed("gravel", 14); `
setSurfaceSpeed("pebblestone", 16);
```
and change the comment 
`// +4km/h for certain surfaces (max 16km/h)` 
into
`// increase speed for the advantage of wide MTB tires on certain surfaces (up to max 16km/h)`

web/src/test/java/com/graphhopper/application/resources/RouteResourceClientHCTest.java
```diff
@@ -406,7 +406,7 @@ public void testWaypointIndicesAndLegDetails(TestParam p) {
 
         GHResponse response = gh.route(req);
         ResponsePath path = response.getBest();
-        assertEquals(5436, path.getDistance(), 5);
+        assertEquals(5158, path.getDistance(), 5);
```
ratrun, line 409: Question: Did you cross check the reason for this pretty big difference?

core/src/main/java/com/graphhopper/routing/util/parsers/MountainBikeAverageSpeedParser.java
karussell: Regarding:

> setSurfaceSpeed("gravel", 14); `

I would prefer 16 here. Then we would have the same approach as for the other surface types (x+4km/h and max 16).

Or should I lower the value for `gravel` to 10 in BikeCommonAverageSpeedParser?

web/src/test/java/com/graphhopper/application/resources/RouteResourceClientHCTest.java
karussell: Will check.

core/src/main/java/com/graphhopper/routing/util/parsers/MountainBikeAverageSpeedParser.java
ratrun: In this case I choose 

`setSurfaceSpeed("gravel", 16); `

although `gravel` is objectively slower compared to `fine_gravel` and `compacted`.

web/src/test/java/com/graphhopper/application/resources/RouteResourceClientHCTest.java
ratrun: Thank you!

karussell: It is the result of increasing the speed of highway=pedestrian from 4 to 6km/h resulting in this IMO more realistic route:

image

instead of this:

image

(Note that the OSM data in the repo is heavily outdated and [the route in reality](https://graphhopper.com/maps/?point=42.509141%2C1.546063_42.509141%2C+1.546063&point=42.507173%2C1.531902_42.507173%2C+1.531902&point=42.505435%2C1.515943_42.505435%2C+1.515943&point=42.495611%2C1.498368_42.495611%2C+1.498368&profile=bike&layer=OpenStreetMap) is something completely different. Also note that this test does not use bike.json and instead a simplified bike profile for testing purposes which is without e.g. soft reverse oneway.)

core/src/main/java/com/graphhopper/routing/util/parsers/MountainBikeAverageSpeedParser.java
```diff
@@ -7,38 +7,28 @@ public class MountainBikeAverageSpeedParser extends BikeCommonAverageSpeedParser
...
+    protected MountainBikeAverageSpeedParser(DecimalEncodedValue speedEnc,
+                                             EnumEncodedValue<Smoothness> smoothnessEnc,
+                                             DecimalEncodedValue ferrySpeedEnc,
+                                             EnumEncodedValue<RouteNetwork> bikeRouteEnc) {
+        super(speedEnc, smoothnessEnc, ferrySpeedEnc, bikeRouteEnc);
         setTrackTypeSpeed("grade1", 18); // paved
         setTrackTypeSpeed("grade2", 16); // now unpaved ...
         setTrackTypeSpeed("grade3", 12);
         setTrackTypeSpeed("grade4", 8);
         setTrackTypeSpeed("grade5", 6); // like sand/grass
```
karussell: I just found that this speed is still inconsistent. We see "like sand/grass" here but for grass we are actually using 12 and sand has yet another speed (4)

I propose to remove "grass" from the comment and use 4km/h here too.

ratrun: I agree with your proposal!

karussell:
Thanks for the discussion @ratrun!

PR #3216, opened 2025-10-22
Fix #3211: Ignore car 'oneway' tag in FootAccessParser

### LINKED ISSUE #3211: avoid changing side of road
Somehow "foot" does not like this side of the road:

https://graphhopper.com/maps/?point=33.868179%2C-117.619521&point=33.864658%2C-117.616996&profile=foot&layer=Omniscale

Image

Not sure if it is a data issue as OSRM routing is fine.

modhtom:
Hi, this PR fixes issue #3211.

I found an asymmetrical behavior (the route from A->B failed, but B->A worked). This was because the FootAccessParser was not ignoring the generic oneway tag meant for cars, blocking pedestrian access in one direction.

My fix changes the logic in handleWayTags to only apply one-way rules for specific pedestrian tags (like oneway:foot), and to treat all other roads as bidirectional for pedestrians.

I have tested this on my local server and it now produces the correct route. I've also added a new unit test to FootTagParserTest to confirm this fix and prevent future regressions.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -8,6 +8,7 @@ Here is an overview:
  * AnahitaS, docs for Android, Android, Tomcat
  * andreaswolf, flag encoder versioning and more
  * andreylh, polygon for blocked area #1306
+ * modhtom, fixes like #3211
  * Anvoker, fixes like #1614 and helped with JUnit 5 migration #1632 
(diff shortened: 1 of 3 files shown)


karussell:
There is no oneway tag involved. See the original issue where I think I found the problem.

PR #3218, opened 2025-10-24
racingbike: increase from 20km/h to 24km/h

karussell:
The average racing biker is faster than 20 km/h.


ratrun (approved):
Looks good! :+1:

PR #3221, opened 2025-10-26
Don't adjust bicycle speed for highway=service when highway is designated

### LINKED ISSUE #3220: Bike routing: ways with service=* tag get pushing speed (4 km/h) even with bicycle=designated
GraphHopper's bike profile assigns pushing speed (4 km/h) to any way that has a service=* tag (e.g., `service=parking_aisle`), even when `bicycle=designated` is set. This causes major detours.

Sample routes are mentioned in this OSM note: https://www.openstreetmap.org/note/4854192

Possible cause:

In BikeCommonAverageSpeedParser.java (line ~131), the condition `way.hasTag("service")` triggers `PUSHING_SECTION_SPEED` for any service tag value, ignoring bike permissions:

```
if (way.hasTag("surface") && surfaceSpeed == null
    || way.hasTag("bicycle", "dismount")
    || way.hasTag("railway", "platform")
    || pushingRestriction && !way.hasTag("bicycle", INTENDED)
    || way.hasTag("service")) {  // <-- This line
  speed = PUSHING_SECTION_SPEED;
}
```

(Analysis was partially generated with AI help and might be wrong)

ratrun:
This changes fixes #3220.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
@@ -151,7 +151,7 @@ public abstract class BikeCommonAverageSpeedParser extends AbstractAverageSpeedP
                 || way.hasTag("bicycle", "dismount")
                 || way.hasTag("railway", "platform")
                 || pushingRestriction && !way.hasTag("bicycle", INTENDED)
-                || way.hasTag("service")) {
+                || way.hasTag("service") && !isDesignated(way)) {
             speed = PUSHING_SECTION_SPEED;
 
         } else {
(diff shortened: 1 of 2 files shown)


karussell:
Thanks!

karussell:
@ratrun do you know why we have this in the first place? Maybe we should remove this special condition with `service` in general? 

Currently we assume speed=12km/h for highway=service, but as soon as a way is tagged with `service=alley` [like this](https://www.openstreetmap.org/way/30955228) we set speed to 4km/h. I understand why we would avoid it, but the speed is unrealistic low and inconsistent with `highway=service` and leads to [too long detours](https://graphhopper.com/maps-dev/route_stats/?point=51.144924%2C12.643669&point=51.145586%2C12.643792&point=51.146508%2C12.645776&profile=bike&layer=TF+Cycle).

Or maybe check explicitly for `service=parking_aisle` only?

ratrun:
> @ratrun do you know why we have this in the first place? Maybe we should remove this special condition with `service` in general?

I agree that the speed decrease is too drastic. If I remember correctly it was an attempt to consider the danger of [dooring](https://en.wikipedia.org/wiki/Dooring). But the check for only the presence of the service tag  is not good and should really be changed to explicitly check for 'service=parking' . I think 8 or 10km/h would be realistic for such a situation.

PR #3225, opened 2025-11-07
fix class:bicycle handling in parsers

karussell:
If we classify a route better than the value from `class:bicycle`, the smaller PriorityCode from class:bicycle should not overwrite ours, I think.

This PR also makes the mapping for the negative `class:bicycle` values more align with our PriorityCode.


ratrun (approved):
This looks very good to me.

PR #3226, opened 2025-11-07
move bike route handling into custom models

karussell:
This moves the bike route handling (`bike_network`) out of the priority parser into the custom models, like we already do for `hike_network` and `hike.json`.

The advantage is a bit less code and also that we can apply higher values necessary for a situation like [this route](https://graphhopper.com/maps/?point=48.055902%2C11.532919&point=48.042006,11.520103&profile=bike&layer=TF+Cycle). In general this area is a nice real world case to tune "bike route" and "bike designated" relevance. E.g. [this route](https://graphhopper.com/maps/?point=48.064363%2C11.543991&point=48.037781%2C11.521546&profile=bike&layer=TF+Cycle&custom_model=%7B%22priority%22%3A%5B%7B%22if%22%3A%22bike_network%21%3DLOCAL%22%2C%22multiply_by%22%3A%220.4%22%7D%5D%7D) uses the local bike network and [this route](https://graphhopper.com/maps/?point=48.064363%2C11.543991&point=48.037781%2C11.521546&profile=bike&layer=TF+Cycle&custom_model=%7B%22priority%22%3A%5B%7B%22if%22%3A%22bike_network%21%3DINTERNATIONAL%22%2C%22multiply_by%22%3A%220.4%22%7D%5D%7D) the international and [this route](https://graphhopper.com/maps/?point=48.064363%2C11.543991&point=48.037781%2C11.521546&profile=bike&layer=TF+Cycle&custom_model=%7B%22distance_influence%22%3A150%7D) the "middle" way that has a good & separated & designated highway=path.

But not too heavy boost is required to [avoid local routes with rather big incline](https://graphhopper.com/maps/?point=51.108047%2C13.676413&point=51.108573%2C13.65899&profile=bike).


core/src/main/resources/com/graphhopper/custom_models/bike.json
```diff
@@ -1,6 +1,6 @@
 // to use this custom model you need to set the following option in the config.yml
 // graph.elevation.provider: srtm   # enables elevation
-// graph.encoded_values: bike_priority, bike_access, roundabout, bike_average_speed, bike_road_access, foot_road_access, average_slope, mtb_rating, hike_rating, country, road_class
+// graph.encoded_values: bike_priority, bike_access, bike_network, roundabout, bike_average_speed, bike_road_access, foot_road_access, average_slope, mtb_rating, hike_rating, country, road_class
```
ratrun, line 3: I just tested this branch with my existing config and am a little bit confused now: adding of the  `bike_network` to `graph.encoded_values` in config.yml is actually not needed. You also didn't add it to config-example.yml. Can you please explain why this is not needed? As it is not needed, would it be possible to just don't add `bike_network` here?

core/src/main/resources/com/graphhopper/custom_models/mtb.json
```diff
@@ -1,15 +1,17 @@
 // to use this custom model you need to set the following option in the config.yml
 // graph.elevation.provider: srtm   # enables elevation
-// graph.encoded_values: mtb_priority, mtb_access, roundabout, mtb_average_speed, bike_road_access, average_slope, mtb_rating, hike_rating, country, road_class
+// graph.encoded_values: mtb_priority, mtb_access, bike_network, roundabout, mtb_average_speed, bike_road_access, average_slope, mtb_rating, hike_rating, country, road_class
```
ratrun, line 3: @karussell Depending on the answer on my comment to the changes in the file bike.json the same changes might be needed here as well.

core/src/main/resources/com/graphhopper/custom_models/racingbike.json
```diff
@@ -1,15 +1,17 @@
 // to use this custom model you need to set the following option in the config.yml
 // graph.elevation.provider: srtm   # enables elevation
-// graph.encoded_values: racingbike_priority, racingbike_access, roundabout, racingbike_average_speed, bike_road_access, average_slope, mtb_rating, hike_rating, sac_scale, country, road_class
+// graph.encoded_values: racingbike_priority, racingbike_access, bike_network, roundabout, racingbike_average_speed, bike_road_access, average_slope, mtb_rating, hike_rating, sac_scale, country, road_class
```
ratrun, line 3: @karussell Depending on the answer on my comment to the changes in the file bike.json the same changes might be needed here as well.

core/src/test/java/com/graphhopper/routing/RoutingAlgorithmWithOSMTest.java
```diff
@@ -698,8 +706,8 @@ private GraphHopper createHopper(String osmFile, Profile... profiles) {
                 setOSMFile(osmFile).
                 setProfiles(profiles).
                 setEncodedValuesString("average_slope, max_slope, hike_rating, car_access, car_average_speed, " +
-                        "foot_access, foot_priority, foot_average_speed, " +
-                        "bike_access, bike_priority, bike_average_speed, foot_network, roundabout, " +
+                        "foot_access, foot_priority, foot_average_speed, foot_network, " +
+                        "bike_access, bike_priority, bike_average_speed, bike_network, roundabout, " +
```
ratrun, line 710: @karussell Depending on the answer on my comment to the changes in the file bike.json the same changes might be needed here as well.

web/src/test/java/com/graphhopper/application/resources/RouteResourceCustomModelTest.java
```diff
@@ -66,7 +66,7 @@ private static GraphHopperServerConfiguration createConfig() {
                 putObject("custom_areas.directory", "./src/test/resources/com/graphhopper/application/resources/areas").
                 putObject("import.osm.ignored_highways", "").
                 putObject("graph.encoded_values", "car_access, car_average_speed, road_access, " +
-                        "bike_access, bike_priority, bike_average_speed, bike_road_access, " +
+                        "bike_access, bike_priority, bike_average_speed, bike_road_access, bike_network, " +
```
ratrun, line 69: @karussell Depending on the answer on my comment to the changes in the file bike.json the same changes might be needed here as well.

ratrun (commented):
Except for the questions  in the file bike.json I don't have other comments

core/src/main/resources/com/graphhopper/custom_models/bike.json
karussell: This is strange, when you remove it you should get "Encoded values missing: bike_network." as we do not add it by default [here](https://github.com/graphhopper/graphhopper/blob/25903cd0c6cfd23e1e72da71900b26dc2cfc362f/core/src/main/java/com/graphhopper/GraphHopper.java#L884). Are you sure you didn't add it via java by default in your code and you are using the latest bike.json from this branch?

For config-example.yml, we have enabled only car by default, and therefore I did not add it there and only in the documentation of bike.json

ratrun: >Are you sure you didn't add it via java by default in your code and you are using the latest bike.json from this branch?

I seem to have tested with a wrong branch. Now I need to add `bike_network` to the `graph.encoded_values`. :relieved: 

>For config-example.yml, we have enabled only car by default, and therefore I did not add it there and only in the documentation of bike.json

I would prefer to list all available encoded value strings in the example file as it is nowhere documented that the example strings only contain those which are applicable for car. It actually says 

`# Possible values: road_class,road_class_link,road_environment,max_speed,road_access `

...

karussell: Ok, sounds indeed simpler. Should we include all necessary encoded values for car.json, bike.json and foot.json in config-example.yml by default? (I would do this in a separate code change)

ratrun: >Should we include all necessary encoded values for car.json, bike.json and foot.json in config-example.yml by default? 

Yes please!

Reading your answer carefully I get the impression that you intend to remove the documentation in the various custom profile files. I see the need for both as there are two use-cases:

1. If one just wants to use one of the example default profiles: Then I need to look up the needed encoded values  in the respective json file and make sure to configure these in my config.yml
2. If one wants to learn what is possible with encoded values I need a list of the available ones. The best location  for this list is in the config-example.yml

>I would do this in a separate code change

Definitely! Maybe you can somehow structure the available encoded values somehow as the list is becoming longer and longer and there is currently no obvious order. One order method would be alphabetically, but maybe you have better ideas.

Issue #3219, opened 2025-10-24
highway=residential without bicycle infrastructure preferred over one with infrastructure


ratrun:
See bug description and discussion in [forum](https://discuss.graphhopper.com/t/issues-with-cycling-routing/9767).

ratrun:
@karussell: I would like to get started with this. Do we need to apply the quickfix or do you have a better idea?

karussell:
Hm, tricky to say. This looks to be again due to the complexity of the priority parser, which I would prefer to reduce. We could use the same approach I took for #3226 and slightly move out a few things into the custom model. A cycleway encoded value (or just `boolean bike_designated`?) would be helpful anyway.

wltrr:
I think having a separate encoded value + parser to parse [cycleway](https://wiki.openstreetmap.org/wiki/Key:cycleway) would be a great addition. I can also work on it if desired. 

Generally I also agree that the priority parser is too complex. Moving more logic to `bike.json` would be more transparent and accessible in my opinion.

karussell:
A new encoded value would be nice, yes. Although it might be a bit tricky, but haven't looked into it yet.

> Generally I also agree that the priority parser is too complex. 

A few steps are already done and future change I will try is inside #3227. Further steps will hopefully also fix a few of the open issues for bike and e.g. remove at least a few bike priority parsers so that maybe just one BikePriorityParser is left (currently we have 4)

karussell:
Shoudn't we handle `highway=residential` more like we handle `highway=tertiary`? (see #3015)

The tag `highway=residential` often ([not always!](https://graphhopper.com/maps/?point=51.435879%2C14.244483&point=51.435571%2C14.229569&profile=bike)) means less traffic, but at least in Germany you often have lots of parked cars on the side which speaks more for UNCHANGED=1.0 than 1.2. For example I don't think that it is the same like `highway=path`, `bicycle=designated` and `foot=yes` where we currently use PREFER.

(And as we are at it we should also use UNCHANGED=1.0 for `highway=unclassified`, `highway=service` and `highway=road`. At least there should be no such difference with 1.2 for `unclassified` and 1.0 for `tertiary`. Changing highway=residential to UNCHANGED will also remove the difference to racingbike, which I'd prefer. All the mentioned highway values will be automatically "preferred" as the other highway values get a penalty and we could probably remove `preferHighwayTags`.)

ratrun:
[...]

PR #3227, opened 2025-11-13
Simplify bike priority parsers

karussell:
This PR does the following changes (via separate commits):

 * move ferry avoidance into custom model
 * replaces PriorityCode with just a double (allows us to use all the 16 values and IMO also makes reading easier when details are required, because it is very likely that you need to look up the exact value)
 * removes the weightToPrioMap (this required some logic changes but I was able to keep most tests identical)
 * the following 3 cases should all give the same prio of 1.3: 
     * `highway=*&cycleway=track`
     * `highway=cycleway`
     * `highway=path&surface=asphalt&bicycle=designated`
 * further simplifies the code to get a single if-else-if block for a simpler reasoning
 * minor alignment of racingbike code required to get same test response

Probably for separate PR: one question I stumbled over is: why do we always avoid the avoidHighwayTags? Shouldn't we decrease the avoidance a bit in case bicycle=yes or designated? Similar question for bike=designated for preferHighwayTags and pushingSectionsHighways. And e.g. avoiding tunnel should be done always (except if bike=designated). And at the same time we could integrate the default avoidance of track for bad surfaces for bike and rcbike.

### DIFF
--- a/core/src/main/java/com/graphhopper/GraphHopper.java
+++ b/core/src/main/java/com/graphhopper/GraphHopper.java
@@ -733,6 +733,10 @@ public class GraphHopper {
             elevationProvider = new MultiSourceElevationProvider(cacheDirStr);
         } else if (eleProviderStr.equalsIgnoreCase("skadi")) {
             elevationProvider = new SkadiProvider(cacheDirStr);
+        } else if (eleProviderStr.equalsIgnoreCase("sonny")) {
+            elevationProvider = new SonnyProvider(cacheDirStr);
+        } else if (eleProviderStr.equalsIgnoreCase("multi3")) {
+            elevationProvider = new MultiSource3ElevationProvider(cacheDirStr);
         }
 
         if (elevationProvider instanceof TileBasedElevationProvider) {
--- a/core/src/main/java/com/graphhopper/reader/dem/MultiSource3ElevationProvider.java
+++ b/core/src/main/java/com/graphhopper/reader/dem/MultiSource3ElevationProvider.java
@@ -1,0 +1,127 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.reader.dem;
+
+import com.graphhopper.storage.DAType;
+
+/**
+ * The MultiSource3ElevationProvider mixes different elevation providers to provide the best available elevation data
+ * for the whole world.
+ *
+ * @author ratrun
+ */
+public class MultiSource3ElevationProvider extends TileBasedElevationProvider {
+
+    // The  provider that provides elevation data for Europe
+    private final TileBasedElevationProvider sonnyProvider;
+    // Usually a high resolution provider in the SRTM area
+    private final TileBasedElevationProvider srtmProvider;
+    // The fallback provider that provides elevation data globally
+    private final TileBasedElevationProvider globalProvider;
+
+    public MultiSource3ElevationProvider(TileBasedElevationProvider srtmProvider, TileBasedElevationProvider globalProvider, TileBasedElevationProvider sonnyProvider) {
+        super(srtmProvider.cacheDir.getAbsolutePath());
+        this.srtmProvider = srtmProvider;
+        this.globalProvider = globalProvider;
+        this.sonnyProvider = sonnyProvider;
+    }
+
+    public MultiSource3ElevationProvider() {
+        this(new CGIARProvider(), new GMTEDProvider(), new SonnyProvider());
+    }
+
+    public MultiSource3ElevationProvider(String cacheDir) {
+        this(new CGIARProvider(cacheDir), new GMTEDProvider(cacheDir), new SonnyProvider(cacheDir));
+    }
+
+    @Override
+    public double getEle(double lat, double lon) {
+        try {
+            return sonnyProvider.getEle(lat, lon);
+        } catch ( Exception ex) {
+            // Sometimes the cgiar data north of 59.999 equals 0
+            if (lat < 59.999 && lat > -56) {
+                double ele = srtmProvider.getEle(lat, lon);
+                if (Double.isNaN(ele))  {
+                    // If the SRTM data is not available, use the global provider
+                    ele = globalProvider.getEle(lat, lon);
+                }
+                return ele;
+            }
+            return globalProvider.getEle(lat, lon);
+        }
+    }
+
+    /**
+     * For the MultiSource3ElevationProvider you have to specify the base URL separated by a ';'.
+     * The first for cgiar, the second for gmted, the third for sonny
+     */
+    @Override
+    public MultiSource3ElevationProvider setBaseURL(String baseURL) {
+        String[] urls = baseURL.split(";");
+        if (urls.length != 3) {
+            throw new IllegalArgumentException("The base url must consist of three urls separated by a ';'. The first for cgiar, the second for gmted");
+        }
+        srtmProvider.setBaseURL(urls[0]);
+        globalProvider.setBaseURL(urls[1]);
+        sonnyProvider.setBaseURL(urls[2]);
+        return this;
+    }
+
+    @Override
+    public MultiSource3ElevationProvider setDAType(DAType daType) {
+        srtmProvider.setDAType(daType);
+        globalProvider.setDAType(daType);
+        sonnyProvider.setDAType(daType);
+        return this;
+    }
+
+    @Override
+    public MultiSource3ElevationProvider setInterpolate(boolean interpolate) {
+        srtmProvider.setInterpolate(interpolate);
+        globalProvider.setInterpolate(interpolate);
+        sonnyProvider.setInterpolate(interpolate);
+        return this;
+    }
+
+    @Override
+    public boolean canInterpolate() {
+        return srtmProvider.canInterpolate() && globalProvider.canInterpolate() && sonnyProvider.canInterpolate();
+    }
+
+    @Override
+    public void release() {
+        srtmProvider.release();
+        globalProvider.release();
+        sonnyProvider.release();
+    }
+
+    @Override
+    public MultiSource3ElevationProvider setAutoRemoveTemporaryFiles(boolean autoRemoveTemporary) {
+        srtmProvider.setAutoRemoveTemporaryFiles(autoRemoveTemporary);
+        globalProvider.setAutoRemoveTemporaryFiles(autoRemoveTemporary);
+        sonnyProvider.setAutoRemoveTemporaryFiles(autoRemoveTemporary);
+        return this;
+    }
+
+    @Override
+    public String toString() {
+        return "multi3";
+    }
+
+}
--- a/core/src/main/java/com/graphhopper/reader/dem/SonnyProvider.java
+++ b/core/src/main/java/com/graphhopper/reader/dem/SonnyProvider.java
@@ -1,0 +1,133 @@
+/*
+ *  Licensed to GraphHopper GmbH under one or more contributor
+ *  license agreements. See the NOTICE file distributed with this work for
+ *  additional information regarding copyright ownership.
+ *
+ *  GraphHopper GmbH licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except in
+ *  compliance with the License. You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.graphhopper.reader.dem;
+
+import java.io.*;
+
+import static com.graphhopper.util.Helper.close;
+
+/**
+ * Sonny's LiDAR Digital Terrain Models contains elevation data for Europe with 1 arc second (~30m) accuracy.
+ * The description is available at https://sonny.4lima.de/. Unfortunately the data is provided on a Google Drive
+ * https://drive.google.com/drive/folders/0BxphPoRgwhnoWkRoTFhMbTM3RDA?resourcekey=0-wRe5bWl96pwvQ9tAfI9cQg
+ * Therefore, the data is not available via a direct URL and you have to download it manually. After downloading,
+ * the data has to be unzipped and placed in the cache directory. The cache directory is expected to contain DTM
+ * data files with the naming convention like "N49E011.hgt" for the area around 49°N and 11°E.
+ * <p>
+ * Please note that the data cannot be used for public hosting or redistribution due to the terms of use of the data. See
+ * https://github.com/graphhopper/graphhopper/issues/2823
+ * <p>
+ *
+ * @author ratrun
+ */
+public class SonnyProvider extends AbstractSRTMElevationProvider {
+
+    public SonnyProvider() {
+        this("");
+    }
+
+    public SonnyProvider(String cacheDir) {
+        super("https://drive.google.com/drive/folders/0BxphPoRgwhnoWkRoTFhMbTM3RDA?resourcekey=0-wRe5bWl96pwvQ9tAfI9cQg/", // This base URL cannot be used, as the data is not available via a direct URL
+                cacheDir.isEmpty() ? "/tmp/sonny" : cacheDir,
+                "GraphHopper SonnyReader",
+                -56,
+                90,
+                3601
+        );
+    }
+
+    public static void main(String[] args) throws IOException {
+        SonnyProvider provider = new SonnyProvider();
+        // 338
+        System.out.println(provider.getEle(49.949784, 11.57517));
+        // 462
+        System.out.println(provider.getEle(49.968668, 11.575127));
+        // 462
+        System.out.println(provider.getEle(49.968682, 11.574842));
+        // 982
+        System.out.println(provider.getEle(47.468668, 14.575127));
+        // 1094
+        System.out.println(provider.getEle(47.467753, 14.573911));
+        // 1925
+        System.out.println(provider.getEle(46.468835, 12.578777));
+        // 834
+        System.out.println(provider.getEle(48.469123, 9.576393));
+        // Out of area
+        try {
+            System.out.println(provider.getEle(37.5969196, 23.0706507));
+        } catch (Exception e) {
+            System.out.println("Error: Out of area! " + e.getMessage());
+        }
+
+    }
+
+    @Override
+    byte[] readFile(File file) throws IOException {
+        InputStream is = new FileInputStream(file);
+        BufferedInputStream buff = new BufferedInputStream(is);
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
+        byte[] buffer = new byte[0xFFFF];
+        int len;
+        while ((len = buff.read(buffer)) > 0) {
+            os.write(buffer, 0, len);
+        }
+        os.flush();
+        close(buff);
+        return os.toByteArray();
+    }
+
+    @Override
+    String getFileName(double lat, double lon) {
+        String str = "";
+
+        int minLat = Math.abs(down(lat));
+        int minLon = Math.abs(down(lon));
+
+        if (lat >= 0)
+            str += "N";
+        else
+            str += "S";
+
+        if (minLat < 10)
+            str += "0";
+        str += minLat;
+
+        if (lon >= 0)
+            str += "E";
+        else
+            str += "W";
+
+        if (minLon < 10)
+            str += "0";
+        if (minLon < 100)
+            str += "0";
+        str += minLon;
+        return str;
+    }
+
+    @Override
+    String getDownloadURL(double lat, double lon) {
+        return getFileName(lat, lon) + ".hgt";
+    }
+
+    @Override
+    public String toString() {
+        return "sonny";
+    }
+
+}
--- a/core/src/main/resources/com/graphhopper/custom_models/bike.json
+++ b/core/src/main/resources/com/graphhopper/custom_models/bike.json
@@ -1,6 +1,6 @@
 // to use this custom model you need to set the following option in the config.yml
 // graph.elevation.provider: srtm   # enables elevation
-// graph.encoded_values: bike_priority, bike_access, bike_network, roundabout, bike_average_speed, bike_road_access, foot_road_access, average_slope, mtb_rating, hike_rating, country, road_class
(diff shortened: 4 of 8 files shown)


core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
ratrun, line 1: I don't like loosing the possibility of searching for certain priority values. I believe that we will need this soon for the first time at fixing #3219. 

Therefore I suggest to define constants like this:

```
package com.graphhopper.routing.util;

public final class PriorityConstants {

    // Private constructor to prevent instantiation
    private PriorityConstants() {
    }

    public static final double PRIO_01 = 0.1;
    public static final double PRIO_02 = 0.2;
    public static final double PRIO_03 = 0.3;
    public static final double PRIO_04 = 0.4;
    public static final double PRIO_05 = 0.5;
    public static final double PRIO_06 = 0.6;
    public static final double PRIO_07 = 0.7;
    public static final double PRIO_08 = 0.8;
    public static final double PRIO_09 = 0.9;
    public static final double PRIO_10 = 1.0;
    public static final double PRIO_11 = 1.1;
    public static final double PRIO_12 = 1.2;
    public static final double PRIO_13 = 1.3;
    public static final double PRIO_14 = 1.4;
    public static final double PRIO_15 = 1.5;
}
```

and utilize these constants in the code instead of hard coding the double values.

karussell: Why is searching for a certain priority value an advantage? Also I do not understand what this has to do with #3219.

IMO the missing step for fixing #3219 is that there is still no clear concept of what should get which priority and when to boost or penalize (similar to the average speed parser). But at least we now have a single if-else block which makes it easier to reason about. 

But this if-else block should be further simplified and this simplification will then hopefully fix issues like #3219 (as problems like these get more or less obvious from the code).

ratrun: > Why is searching for a certain priority value an advantage

This is needed to analyze how other situations are prioritized in comparison to a new one. In #3219 I think that we will need to shift some existing priority deltas by a step of 0.1 in order to be able to make use of the currently left out priority values. 

And hard coding these values is ugly IMHO.

karussell: > This is needed to analyze how other situations are prioritized in comparison to a new one.

But comparing numbers is much simpler than comparing words which you need to look up (!?)

> In https://github.com/graphhopper/graphhopper/issues/3219 I think that we will need to shift some existing priority deltas by a step of 0.1 in order to be able to make use of the currently left out priority values.

We could try to use 1.1 instead of 1.2 and add 0.1 if it has a lane. (Or we could even try to use `residential` without any boost by default.) 

IMO speaking of numbers instead of SLIGHT_PREFER and PREFER is much simpler. E.g. do you know without looking how much worse BAD is compared to AVOID?

One further argument for the numbers is that we just have numbers for the `priority` section of custom models too.

btw: it is unclear to me why we handle tertiary and unclassified differently. Is there a reason? And shouldn't service be 0.9 instead of getting a boost by default? (but this is yet another PR I think :) )

ratrun: >IMO speaking of numbers instead of SLIGHT_PREFER and PREFER is much simpler. E.g. do you know without looking how much worse BAD is compared to AVOID?

I'm not proposing to keep the class with those names, but instead define constants for the double priority values: 
```
public static final double PRIO_01 = 0.1; 
..
```
see above. This way you can easily search for all source code which results in a specific prio value and from the naming of the constants you can easily derive its value.

>One further argument for the numbers is that we just have numbers for the priority section of custom models too.

But in Java we have the means for better readable code, which we should utilize.

Let's discuss #3219 once this PR is finished.

karussell: I'm still not convinced that an enum gives us any advantages. If we simplify the code then all the logic will be in one place with a single & simple if-else-block or whatever simple logic. Nothing we need to search for.

> But in Java we have the means for better readable code, which we should utilize.

It is definitely not better readable because a "number" as concept for priority is immediately clear and such an enum needs to be understood first, even if it just mirrors numbers.

The only advantage would be that we can find the usage which I do not see as an advantage as this shouldn't be necessary. E.g. if we know there is just a single class and a single test class then we can as well search those two files for e.g. `0.1` too.

But I doubt that this is a real problem that we have to solve.

For example when I tried to further simplify the code I found one more inconsistency which you cannot find from a "usage search" of an enum.

The inconsistency I found is: for path, track and bridleway with a good surface and when bicycle=designated we use 1.3 (VERY_NICE), but for other highways the designated will be just 1.2 (PREFER). On the other hand a highway included in pushing section set (like `path`) with just bicycle=yes we give a 1.2 even if a bad surface! Here is the test which makes the inconsistency obvious:

```java

    @Test
    public void test() {
        way.setTag("bicycle", "yes");
        way.setTag("surface", "ground");
        assertPriorityAndSpeed(1.2, 10, way);

        way = new ReaderWay(1);
        way.setTag("highway", "residential");
        way.setTag("bicycle", "designated");
        assertPriorityAndSpeed(1.2, 18, way);
    }
```

ratrun: > The only advantage would be that we can find the usage which I do not see as an advantage as this shouldn't be necessary. E.g. if we know there is just a single class and a single test class then we can as well search those two files for e.g. 0.1 too.

Currently there are more files involved, even if one limits a search to just a single profile. 
I counted five: BikeCommonPriorityParser, BikePriorityParser, AbstractBikeTagParserTester, BikeTagParserTest, BikeCustomModelTest. But if you really insist on these numbers then there is nothing I can do about it, except putting it into my fork, which I don't want in order to avoid merge conflicts.

karussell: > Currently there are more files involved

Well, this will change. I plan to do this similarly to how we have it for foot and hike. Just a single parser (or later maybe even none). 

> But if you really insist on these numbers

I do not insist. I'm just not convinced that such an enum is a good solution in this case. 

But for a moment, let's focus on the other changes: are you fine with them?

core/src/main/java/com/graphhopper/routing/util/parsers/MountainBikePriorityParser.java
```diff
@@ -32,18 +29,12 @@ protected MountainBikePriorityParser(DecimalEncodedValue speedEnc, DecimalEncode
...
-                weightToPrioMap.put(50d, SLIGHT_PREFER);
-            else if (trackType == null)
-                weightToPrioMap.put(90d, PREFER);
-            else if (trackType.startsWith("grade"))
-                weightToPrioMap.put(100d, VERY_NICE);
-        }
+    double collect(ReaderWay way, double wayTypeSpeed, boolean bikeDesignated) {
+        double prio = super.collect(way, wayTypeSpeed, bikeDesignated);
+
+        // worse grade should not mean we prefer it because it is also less fun
```
ratrun, line 35: We still should prefer tracktype=grade2, grade3, grade4 and eventually also grade5 more compared to grade1. Imagine a tracktype grade1 and parallel a highway=track with one of the other grades. From a MTB profile I would expect that the tracktype=grade1 is avoided in such a case.

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
```diff
@@ -27,31 +25,33 @@ protected RacingBikePriorityParser(DecimalEncodedValue priorityEnc, DecimalEncod
...
     @Override
-    void collect(ReaderWay way, double wayTypeSpeed, boolean bikeDesignated, TreeMap<Double, PriorityCode> weightToPrioMap) {
-        super.collect(way, wayTypeSpeed, bikeDesignated, weightToPrioMap);
+    double collect(ReaderWay way, double wayTypeSpeed, boolean bikeDesignated) {
+        double prio = super.collect(way, wayTypeSpeed, bikeDesignated);
 
         String highway = way.getTag("highway");
         if ("service".equals(highway) || "residential".equals(highway)) {
-            weightToPrioMap.put(40d, SLIGHT_AVOID);
+            prio *= 0.7;
```
ratrun, line 46: I think the factor 0.7 is too small, I would set it to 0.9. In the old code we fixed set it to SLIGHT_AVOID (0.9). Can you please explain your reasoning behind?

core/src/test/java/com/graphhopper/routing/util/parsers/AbstractBikeTagParserTester.java
```diff
@@ -341,33 +339,33 @@ public void testTram() {
...
         way.setTag("highway", "service");
-        assertPriorityAndSpeed(PREFER, 12, way);
+        assertPriorityAndSpeed(1.2, 12, way);
 
         way.setTag("service", "parking_aisle");
-        assertPriorityAndSpeed(SLIGHT_AVOID, 4, way);
+        assertPriorityAndSpeed(0.9, 4, way);
         way.setTag("bicycle", "designated");
-        assertPriorityAndSpeed(VERY_NICE, 12, way);
+        assertPriorityAndSpeed(1.2, 12, way);
```
ratrun, line 358: This was VERY_NICE (1.3) before. Was this change intended? What is the reasoning behind?

core/src/test/java/com/graphhopper/routing/util/parsers/FootTagParserTest.java
```diff
@@ -322,74 +321,74 @@ public void testMixSpeedAndSafe() {
...
         way.setTag("foot", "use_sidepath");
-        assertPriority(PriorityCode.VERY_BAD, way);
+        assertPriority(0.3, way);
 
         way.clearTags();
         way.setTag("highway", "tertiary");
         // tertiary without sidewalk is roughly like primary with sidewalk
         way.setTag("sidewalk", "no");
-        assertPriority(PriorityCode.AVOID, way);
+        assertPriority(0.7, way);
```
ratrun, line 345: Note: This was AVOID (0.8) before, but 0.7 is also OK.

core/src/test/java/com/graphhopper/routing/util/parsers/FootTagParserTest.java
```diff
@@ -322,74 +321,74 @@ public void testMixSpeedAndSafe() {
...
         way.setTag("highway", "secondary");
-        assertPriority(PriorityCode.AVOID, way);
+        assertPriority(0.8, way);
         way.setTag("highway", "trunk"); // secondary should be better to mostly avoid trunk e.g. here 46.9889,10.5664->47.0172,10.6059
-        assertPriority(PriorityCode.BAD, way);
+        assertPriority(0.5, way);
 
         way.setTag("sidewalk", "no");
-        assertPriority(PriorityCode.REACH_DESTINATION, way);
+        assertPriority(0.2, way);
```
ratrun, line 378: Note: Was REACH_DESTINATION (0.1) before. Was this change intended?

core/src/test/java/com/graphhopper/routing/util/parsers/MountainBikeTagParserTest.java
```diff
@@ -64,35 +61,35 @@ protected BikeCommonPriorityParser createPriorityParser(EncodedValueLookup looku
...
 
         way.setTag("bicycle", "yes");
-        assertPriorityAndSpeed(PREFER, 12, way);
+        assertPriorityAndSpeed(1.2, 12, way);
 
         way.setTag("highway", "track");
         way.setTag("bicycle", "yes");
         way.setTag("tracktype", "grade3");
-        assertPriorityAndSpeed(VERY_NICE, 12, way);
+        assertPriorityAndSpeed(1.2, 12, way);
```
ratrun, line 82: See my comment for the respective code change. I would prefer to keep the old classification here.

core/src/test/java/com/graphhopper/routing/util/parsers/RacingBikeTagParserTest.java
```diff
@@ -73,45 +71,45 @@ public void testAvoidTunnel() {
...
+        assertPriorityAndSpeed(1.2, 24, osmWay);
     }
 
     @Test
     @Override
     public void testService() {
         ReaderWay way = new ReaderWay(1);
         way.setTag("highway", "service");
-        assertPriorityAndSpeed(SLIGHT_AVOID, 12, way);
+        assertPriorityAndSpeed(0.7, 12, way);
```
ratrun, line 89: Note: Was SLIGHT_AVOID (0.9) before. 0.7 is probably also OK, but I would prefer 0.8.

core/src/test/java/com/graphhopper/routing/util/parsers/RacingBikeTagParserTest.java
```diff
@@ -246,56 +244,56 @@ public void testPriority_avoidanceOfHighMaxSpeed() {
...
-        assertPriorityAndSpeed(encodingManager, priorityEnc, speedEnc, parsers, UNCHANGED, 24, osmWay);
+        assertPriorityAndSpeed(encodingManager, priorityEnc, speedEnc, parsers, 1.0, 24, osmWay);
 
         osmWay.setTag("highway", "motorway");
-        assertPriorityAndSpeed(encodingManager, priorityEnc, speedEnc, parsers, BAD, 18, osmWay);
+        assertPriorityAndSpeed(encodingManager, priorityEnc, speedEnc, parsers, 0.5, 18, osmWay);
 
         osmWay.setTag("tunnel", "yes");
-        assertPriorityAndSpeed(encodingManager, priorityEnc, speedEnc, parsers, REACH_DESTINATION, 18, osmWay);
+        assertPriorityAndSpeed(encodingManager, priorityEnc, speedEnc, parsers, 0.2, 18, osmWay);
```
ratrun, line 265: Was REACH_DESTINATION (0.1) before, 0.2 i also OK.

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -71,208 +68,210 @@ protected BikeCommonPriorityParser createPriorityParser(EncodedValueLookup looku
...
         way.setTag("segregated", "yes");
-        assertPriorityAndSpeed(PREFER, 18, way);
+        assertPriorityAndSpeed(1.2, 18, way);
 
         way.clearTags();
         way.setTag("highway", "platform");
         way.setTag("surface", "paved");
         way.setTag("bicycle", "designated");
-        assertPriorityAndSpeed(VERY_NICE, 18, way);
+        assertPriorityAndSpeed(1.2, 18, way);
```
ratrun, line 144: Was VERY_NICE (1.3) before. 1.2 is better.

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -71,208 +68,210 @@ protected BikeCommonPriorityParser createPriorityParser(EncodedValueLookup looku
...
         way.setTag("foot", "no");
-        assertPriorityAndSpeed(VERY_NICE, 18, way);
+        assertPriorityAndSpeed(1.2, 18, way);
         way.setTag("foot", "yes");
-        assertPriorityAndSpeed(PREFER, 18, way);
+        assertPriorityAndSpeed(1.2, 18, way);
 
         way.setTag("segregated", "yes");
-        assertPriorityAndSpeed(VERY_NICE, 18, way);
+        assertPriorityAndSpeed(1.2, 18, way);
```
ratrun, line 216: Now there is no difference between segregated=yes and no any longer. There should be a difference here. Please change!

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -71,208 +68,210 @@ protected BikeCommonPriorityParser createPriorityParser(EncodedValueLookup looku
...
         way.clearTags();
         way.setTag("highway", "cycleway");
-        assertPriorityAndSpeed(VERY_NICE, 18, way);
+        assertPriorityAndSpeed(1.3, 18, way);
         way.setTag("foot", "yes");
-        assertPriorityAndSpeed(PREFER, 18, way);
+        assertPriorityAndSpeed(1.2, 18, way);
         way.setTag("segregated", "yes");
-        assertPriorityAndSpeed(VERY_NICE, 18, way);
+        assertPriorityAndSpeed(1.3, 18, way);
```
ratrun, line 265: Note: here the differentiation between segregated=yes and no is correct :+1:

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -350,7 +349,7 @@ public void testSmoothness() {
         way.setTag("smoothness", "excellent");
         assertEquals(20, getSpeedFromFlags(way), 0.01);
 
-        way.setTag("smoothness", "bad");
+        way.setTag("smoothness", "0.5");
```
ratrun, line 352: This is obviously a replace all BAD->0.5 error :smile: , please fix it back!

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -364,15 +363,15 @@ public void testSmoothness() {
         way.setTag("surface", "ground");
         assertEquals(10, getSpeedFromFlags(way), 0.01);
 
-        way.setTag("smoothness", "bad");
+        way.setTag("smoothness", "0.5");
```
ratrun, line 366: This is obviously a replace all BAD->0.5 error :smile: , please fix it back!

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -364,15 +363,15 @@ public void testSmoothness() {
...
         assertEquals(8, getSpeedFromFlags(way), 0.01);
 
         way.clearTags();
         way.setTag("highway", "track");
         way.setTag("tracktype", "grade5");
         assertEquals(4, getSpeedFromFlags(way), 0.01);
 
-        way.setTag("smoothness", "bad");
+        way.setTag("smoothness", "0.5");
```
ratrun, line 374: This is obviously a replace all BAD->0.5 error :smile: , please fix it back!

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -388,63 +387,68 @@ public void testLowMaxSpeedIsIgnored() {
...
+        assertPriorityAndSpeed(1.3, 18, way);
+
+        way.clearTags();
         way.setTag("highway", "primary");
         way.setTag("surface", "paved");
-        assertPriority(BAD, way);
+        assertPriority(0.5, way);
         way.setTag("cycleway", "track");
-        assertPriority(PREFER, way);
+        assertPriority(1.3, way);
```
ratrun, line 401: Was PREFER (1.2) before. 1.3 is also OK

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -566,53 +570,53 @@ public void testClassBicycle() {
...
+        assertPriority(1.5, way);
 
         // test overriding by a specific class subtype
         way.setTag("class:bicycle:touring", "2");
-        assertPriority(VERY_NICE, way);
+        assertPriority(1.3, way);
 
         way.setTag("maxspeed", "15");
-        assertPriority(BEST, way);
+        assertPriority(1.4, way);
```
ratrun, line 603: Was 1.5 (BEST) before. Question: Does it become 1.5 for class:bicycle:touring = 3? Please add this test and make sure the result is 1.5!

core/src/main/java/com/graphhopper/routing/util/PriorityCode.java
ratrun: > Well, this will change. I plan to do this similarly to how we have it for foot and hike. Just a single parser (or later maybe even none).

If this is possible I agree with the change to the numeric factors in this PR. But please then assign this change some priority in your planning and create a separate PR for it.

> But for a moment, let's focus on the other changes: are you fine with them?

I agree. I'm through now with all my comments. Don't get shocked by the big amount, it is not as bad as it might look at the first glance :smiley:

ratrun (changes requested):

core/src/main/java/com/graphhopper/routing/util/parsers/MountainBikePriorityParser.java
karussell: We probably need to clarify what 'mountainbike routing' means. IMO it means that

1. we want to go from A to B in a short time, and
2. accept more ways, ways that an mtb can access, and
3. reduce the penalty on certain ways with a bad surface, and
4. prefer "mtb ways". 

I don't think highway=track&tracktype=grade5 should fall under requirement 4 but more 3. And at least we should **not** prefer a way just because the surface is worse like we currently do. This does not make sense to me because of requirement 1. **But more importantly**: a worse surface doesn't mean it makes more fun with a mtb (especially uphill) and we do not know the skill/preference of the person on the mtb.

So requirement 4 is tricky. IMO preferring a road because of its "mountainbikability" can be defined via `mtb:scale` and slightly boosting `highway=path` and `highway=track` (or not adding a penalty if there is no surface tag). And even a big value of mtb:scale does not mean we should prefer it more than e.g. mtb:scale==0 because of the same reason of why a worse surface shouldn't be preferred.

ratrun: For me it is crystal clear that 4. is the essence of a good mountainbike profile and 1. is not really a requirement. If 1. is a requirement, then I'd choose the trekking bike profile instead. But I must say that I am not a hard core mountainbike freak myself, but friends of mine are and for them this is their expectation. 

I'm skeptical about AI, but as I'm not a reader of Mountainbike Forums I thought it would be interesting what the output would be. This is what I [got](https://copilot.microsoft.com/shares/kkU9DzNjCfAciFYrfxhjt)

otbutz: Mountain biking is more of a leisure activity than a means of transportation. It's similar to planning a motorcycle ride with [Kurviger](https://kurviger.com/de/plan).

karussell: > is the essence of a good mountainbike profile and 1. is not really a requirement

Hm, ok. Our disagreement once again shows that we really need to make more things easier to configure via a custom model and so we can have both :)

Because I want a mtb profile where the purpose is "just" to go from A to B with a mtb, so very similar to the hike profile. And then finding nice mtb spots or detours preferable for mtb is out of scope of the profile because we mostly do not have this data in OSM (and IMO only properly done with external/user data like you find in komoot).

karussell: > Mountain biking is more of a leisure activity than a means of transportation

We need to differentiate between the A-to-B routing and the planning where we find the spots and then create a nice tour with multiple via points (or a round tour) to get there. IMO the scope of the GraphHopper routing engine here can be surely be "prefer tracks and mtb-alike roads" (or for kurviger "prefer curvy roads and avoid motorways") but our aim should not be to find "nice mtb spots" on the way to our goal and do big detours because of that and accept really ugly surface (it is not fun!), because there is just no good data for this in OSM (except maybe mtb:scale ~and scenic~).

karussell: actually `mtb:scale` **and** `route=mtb` :grinning: (#3250) 

Also I propose to remove the `scenic=yes` preference from all bikes (#3252).

ratrun:
What is the status of this PR? Did you forget about it or did I add too many comments :question: The discussion stalled although it is a pre-requisite for fixing the annoying bug #3219.

karussell:
I was actually waiting for your feedback regarding the first step here: #3251 :) (but no rush required from my side)

ratrun:
> I was actually waiting for your feedback regarding the first step here: #3251 :) (but no rush required from my side)

Good that I asked. But I already did comment on #3251 and I don't know what I missed there.

Issue #1904, opened 2020-02-12
Make algorithm fallback configurable


easbar:
If we request a routing profile for which no prepared algorithm exists we throw an error and one has to set `disable.ch/lm=true` to get around this error (and fallback to another algorithm). This is potentially not desired in any case. Maybe it would be more useful to automatically fall back to a (slower) algorithm. Ultimately it would be best if we had a configuration parameter to control this behavior.

karussell:
Fixed via #3228.

PR #3228, opened 2025-11-14
more profiles in config-example.yml

karussell:
Include foot and bike in default config. This does not use much more RAM as only a few more encoded values are added and no CH, but this shows how easy it is to work with many profiles. (see [discussion](https://github.com/graphhopper/graphhopper/pull/3226))

To make GraphHopper Maps demo working there is a change required on how we handle ch.disable=true for profiles without a CH preparation. See also #1904.

### DIFF
--- a/config-example.yml
+++ b/config-example.yml
@@ -219,8 +219,9 @@ graphhopper:
   # motorized vehicles. This leads to a smaller and less dense graph, because there are fewer ways (obviously),
   # but also because there are fewer crossings between highways (=junctions).
   # Another typical example is excluding 'motorway', 'trunk' and maybe 'primary' highways for bicycle or pedestrian routing.
-  import.osm.ignored_highways: footway,construction,cycleway,path,steps # typically useful for motorized-only routing
-  # import.osm.ignored_highways: motorway,trunk # typically useful for non-motorized routing
+  import.osm.ignored_highways: ''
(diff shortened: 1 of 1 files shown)


config-example.yml
```diff
@@ -101,13 +102,16 @@ graphhopper:
...
-  #   car_access|block_private=false
-  graph.encoded_values: car_access, car_average_speed, road_access
+  # Example values:
+  #   average_slope,country,curvature,hazmat,hgv,hike_rating,lanes,max_height,max_length, ...
+  # For more details and a full list see the documentation: https://github.com/graphhopper/graphhopper/blob/master/docs/core/custom-models.md
+  # Private roads are blocked by default. To disable this you can specify block_private=false as option:
+  #   car_access|block_private=false, bike_access|block_private=false
+  # By default we add the necessary encoded values for car.json, bike.json and foot.json (remove them if you do not need them)
+  graph.encoded_values: |
+    car_access, car_average_speed, country, road_class, roundabout
```
ratrun: It obviously is not detected by the yml parser as error, but for consistency please add a comma at the end of the string after `roundabout` in this line

ratrun (approved):
Thank you! This looks even better than expected as it makes the foot and bike an as important profile as compared to car. :tada: 

Please note that I only reviewed the changes in config-example.yml

Issue #3232, opened 2025-11-30
Fix signed short overflow and operator precedence bug in BaseGraph geometry storage


wltrr:
**Describe the bug**
I encountered a crash when importing an OSM file containing a very long edge (a 1,300km ferry route) with high-resolution elevation sampling. This generated \> 32,767 geometry points for a single edge.

I believe that the crash is caused by `BaseGraph.java` inside the `getPillarCount(long geoRef)` method. There are two issues in the logic used to read the 24-bit length:

1.  **Signed Short Overflow:** `wayGeometry.getShort(geoRef)` reads a signed Java short. If the lower 16 bits of the count represent a value \> 32,767 (e.g., 32,768), it is interpreted as a negative number (e.g., -32,768).
2.  **Operator Precedence:** The expression `wayGeometry.getByte(geoRef + 2) & 0xFF << 16` is evaluated as `byte & (0xFF << 16)`. Since `<<` has higher precedence than `&`, this masks the high byte entirely, effectively ignoring it.

This results in a negative pillar count, causing `NegativeArraySizeException` during buffer allocation or corrupted coordinate reading (leading to `Infinity` distances).

**To Reproduce**
I have written a reproduction test case for `BaseGraphTest.java`. This creates a 3D graph and adds a geometry list with 32,768 points to trigger the overflow.

```java
@Test
public void testWayGeometryOverflow_ReproducesInfinityCrash() {
    // 1. Setup a 3D graph
    BaseGraph graph = newGHStorage(new RAMDirectory(), true).create(1000);
    NodeAccess na = graph.getNodeAccess();

    // 2. Create nodes and an edge
    na.setNode(0, 50.0, 10.0, 0.0);
    na.setNode(1, 51.0, 10.0, 0.0);
    EdgeIteratorState edge = graph.edge(0, 1);

    // 3. Create a massive geometry that overflows a signed short (max 32767)
    int overflowCount = 32768; 
    PointList massiveList = new PointList(overflowCount, true);
    for (int i = 0; i < overflowCount; i++) {
        massiveList.add(50.0 + (i * 0.000001), 10.0, 0.0);
    }

    // 4. Store the geometry
    edge.setWayGeometry(massiveList);

    // 5. Fetch it back
    try {
        PointList fetched = edge.fetchWayGeometry(FetchMode.ALL);
        assertEquals(overflowCount + 2, fetched.size());
    } catch (Exception e) {
        // Fails here with NegativeArraySizeException or similar corruption
[...]

PR #3233, opened 2025-11-30
Fix signed short overflow in BaseGraph way geometry

wltrr:
**Description:**
Fixes a bug where pillar counts > 32,767 were interpreted as negative numbers due to incorrect operator precedence and signed `short` handling. This caused `NegativeArraySizeException` or corrupted coordinate reading on edges with massive geometries.

**Changes:**
* Corrected bitwise logic in `BaseGraph.getPillarCount` to properly read the 24-bit unsigned integer.
* Added `testWayGeometryOverflow_ReproducesInfinityCrash` to `BaseGraphTest` to prevent regression.

**Fixes:** #3232

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -95,6 +95,7 @@ Here is an overview:
  * vvikas, ideas for many to many improvements and #616
  * zstadler, multiple fixes and car4wd
  * binora, fix mode in navigation response converter
+ * jwltrr, fix for geometry storage overflow
 
 ## Translations
 


karussell:
Thanks! 

This is interesting. Maybe we should limit the number of allowed pillar nodes instead? Because basically it was limited until now and we didn't noticed it and probably won't happen again as we now do no sampling for ferries anyway (your PR #3235)?

Then we could reduce the number of bytes for this count to 2 and reduce the memory usage for `geometry` slightly (430M edges => -0.43GB for planet).

wltrr:
Yes I agree that 2 bytes is probably more than enough. On the [OSM wiki](https://wiki.openstreetmap.org/wiki/Way) it says that a way is limited to having 2000 nodes anyways.

karussell:
We could use 2 bytes and still allow a count up to ~65k if we interpret it as positive integer (and only if more is necessary throw an exception). Would you be able to prepare a PR for this?

wltrr:
Great, yes simply treating it as an unsigned Int sounds like a good solution. I can prepare a PR. I probably can do it in the next week or so.

karussell:
Nice, thank you and no hurry :)

Issue #3234, opened 2025-11-30
Disable elevation sampling for ferry routes to prevent geometry explosion


wltrr:
**Describe the bug**
Currently, OSMReader applies elevation sampling to all edges that exceed the configured `long_edge_sampling_distance`. This includes long-distance ferry routes. This caused a related bug in #3232, where you can also find the bug reports. 

For example, a 1,300 km ferry route (e.g., Denmark to Iceland) combined with a sampling distance of 40m generates over 32,000 geometry points. This creates excessive memory pressure and storage usage for data that is effectively flat (sea level). Even shorter ferry routes naturally do not benefit from additional points for elevation. 

We should thus simply exclude ferries from `EdgeSampling` in `OSMReader`:

```
if (config.getLongEdgeSamplingDistance() < Double.MAX_VALUE && !isFerry(way))
        pointList = EdgeSampling.sample(pointList, config.getLongEdgeSamplingDistance(), distCalc, eleProvider);
```

PR #3235, opened 2025-11-30
Disable elevation sampling for ferries

wltrr:
Here is the PR title and description for your second pull request.

**Title:** Disable elevation sampling for ferries

**Description:**
This PR disables elevation sampling for edges identified as ferries.

Long-distance ferry routes (e.g., 1,300 km) generate massive amounts of artificial geometry points when `graph.elevation.long_edge_sampling_distance` is set to a low value (e.g., 40m). This significantly increases graph size and memory usage during import while adding no value, as ferry routes are effectively flat.

**Changes:**
* Added a check for `!isFerry(way)` in `OSMReader.addEdge` before triggering `EdgeSampling`.

**Fixes:** #3234

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -331,7 +331,7 @@ public class OSMReader {
 
         if (pointList.is3D()) {
             // sample points along long edges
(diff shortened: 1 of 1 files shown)


karussell:
Thanks - this is actually a good idea to do!

Maybe also for bridges but probably not that critical as they are much shorter.

PR #3237, opened 2025-12-02
fix: OSRM response when a legs'has no intermediate steps

OlafFlebbeBosch:
Fix the edge case that part of the necessary fields (like "summary") is missing in case a leg has no intermediate routing instructions.


karussell:
Thank you!

PR #3240, opened 2025-12-07
Fix #1287: Improve wording for 'continue onto' instruction to 'contin…

### LINKED ISSUE #1287: Improve wording when "continuing onto" motorways..

AMEY36:
Closes #1287. This PR updates the US English translation string for continue_onto from 'continue onto %1$s' to 'continue on %1$s' and updates the necessary unit tests to ensure validation.


karussell:
This PR should get rid of the added temporary files and to fix the issue it should implement both things:

> "continue onto" when the road name changes, and "continue on" otherwise.

PR #3243, opened 2025-12-09
Fix bugs in HeightTile and HeightTileTest. 

### LINKED ISSUE #3242: Bugs in HeightTile and HeightTileTest
**Describe the bug**
When implementing a new `DemProvider`, I found some bugs in `HeightTile` and `HeightTileTest`:
- In `HeightTile`, the `precision` field is in fact for the inverse precision. In the `HeightTileTest`, 1e-6 was used as precision, leading to an ineffective test. 
- In the `getHeight` method of `HeightTile`, the relative coordinates of the requested coordinates are computed using `Math.abs()`. This means that the `HeightTile` that is meant for (0,10) x (0,10) would return a height for (-10, 10) x (-10,10). I don't think this was desired behaviour. 

On the other hand, the HeightTileTest had some tests that in my opinion should expect an Exception. I am not sure what the original intention of these tests were so maybe it would be good to double check that.

wltrr:
Fixes #3242.

**1. Fix Invalid Test Precision**
Corrected the precision in `HeightTileTest` from `1e-6` to `1e6`. The previous value (`1e-6`) resulted in a   `lowerBound` of -1,000,000, which effectively disabled boundary checks and allowed invalid tests to pass. Added new assertions for precision clamping and OOB exceptions.

**2. Remove risky `Math.abs()` in `HeightTile`**
Removed `Math.abs()` from the `deltaLat` and `deltaLon` calculations.
* **Unnecessary for valid cases:** Valid requests are always "above/right" of the lower corner, so the delta is naturally positive (e.g., `-8 - (-10) = 2`).
* **Bug Fix:** `Math.abs()` masked errors by "mirroring" invalid coordinates into the tile. For example, requesting latitude `8` on a tile starting at `10` resulted in a delta of `|-2| = 2`, silently returning data for latitude `12` instead of throwing an exception.

### DIFF
--- a/core/src/test/java/com/graphhopper/reader/dem/HeightTileTest.java
+++ b/core/src/test/java/com/graphhopper/reader/dem/HeightTileTest.java
@@ -19,7 +19,6 @@ package com.graphhopper.reader.dem;
 
 import com.graphhopper.storage.DAType;
 import com.graphhopper.storage.DataAccess;
-import com.graphhopper.storage.Directory;
 import com.graphhopper.storage.GHDirectory;
 import org.junit.jupiter.api.Test;
 


karussell:
Thanks for this! Did you test it for a larger region like Germany or Europe? Maybe it uncovers other or related bugs?

PR #3246, opened 2025-12-10
Classify Hong Kong/Macao roads correctly

### LINKED ISSUE #3244: Roads in Hong Kong wrongly classified as "CHN"
**Describe the bug**
Roads in Hong Kong SAR (often known simply as "Hong Kong") are wrongly classified as being "People's Republic of China", which is a problem and a disruptive inaccuracy. (see "additional context" below)

**To Reproduce**

1. checkout recent and stable version of GraphHopper
2. use the Hong Kong OSM extract https://download.geofabrik.de/asia/china/hong-kong.html
3. enable country-specific speed limits, default settings
4. start the GraphHopper server, default settings
5. locate a major road in Hong Kong with no provided `maxspeed` (e.g. Tai Po Road; see screenshots below)
6. enable "Show Routing Graph"
7. look at one of the `highway=primary` in the area
8. `country=?` shows `CHN` (?????)

**Expected behavior**
`country=?` should show `HKG`.

**System Information**

Using the `master` branch. Supposedly this should also be seen in the live web version, but we can't look at the Routing Graph there.

**Screenshots & Logs**

Locate the relevant major road:

Image

After enabling "Show Routing Graph":

Image

**Additional Context**

The Hong Kong Special Administrative Region ("Hong Kong SAR", or just "Hong Kong") is currently a territory under the government of People's Republic of China ("China").

I would risk it and claim that Hong Kong is a non-sovereign country under People's Republic of China. This is because it has its own ISO 3166 country code ("hk" or "HKG"), has its own phone prefix ("+852"), and most importantly, almost every aspect of Hong Kong's policies and practices are incompatible with the suzerain Chinese counterparts other than "they share the same constitution". (Actual degree of local self-rule and other political commentaries are out of scope of this discussion.)

Classifying Hong Kong OSM features as being "CHN" is politically correct, but is disruptively inaccurate and counterproductive because Hong Kong features have their own set of rules, such as their own max speed rules as seen in https://wiki.openstreetmap.org/wiki/Default_speed_limits . Notice how the Hong Kong rules (default is 50km/h everywhere) are different from the Chinese rules (default is 30km/h - 120km/h depending on highway class).

Vectorial1024:
This fixes #3244 by adding in a country entry for Hong Kong SAR.

As per Hong Kong - Macao companionship traditions, an additional entry is also added for Macao SAR.

Values are taken from ISO 3166.


core/src/main/java/com/graphhopper/routing/ev/Country.java
```diff
@@ -112,6 +112,7 @@ public enum Country {
     GTM("Guatemala", "GTM", "GT", true),
     GUY("Guyana", "GUY", "GY", false),
     HND("Honduras", "HND", "HN", true),
+    HKG("Hong Kong", "HKG", "HK", false),
```
otbutz, line 114:
```suggestion
    HKG("Hong Kong", "HKG", "HK", false),
    HND("Honduras", "HND", "HN", true),
```

karussell:
Thanks!

Issue #3248, opened 2025-12-11
Encoded values: `max_speed`, `average_speed`, and country default speed limits?


Vectorial1024:
**Describe the bug**
Consider this to be somewhere between a question and a bug report.

This spins off from #3244.

While reviewing GraphHopper behavior, and after fixing the country classification problem, I noticed that roads in Hong Kong contain seemingly inconsistent max speed / average speed, as demonstrated below. 

Perhaps I don't understand GraphHopper well enough to see what is the intended/actual behavior.

**To Reproduce**

1. checkout `master` branch of GraphHopper
2. use the Hong Kong OSM extract https://download.geofabrik.de/asia/china/hong-kong.html
3. enable country-specific speed limits, default settings
4. start the GraphHopper server, default settings
5. locate a major road in Hong Kong with no provided maxspeed (e.g. Tai Po Road; see screenshots below)
6. enable "Show Routing Graph"
7. look at one of the highway=primary in the area
8. notice that `max_speed=50.0` and also `car_average_speed=66.0`
9. hmmm...?

**Expected behavior**
It seems this should be `max_speed=50.0` and also `car_average_speed=45.0` (see Additional Context for justification).

**System Information**

Using the `master` branch. Supposedly this should also be seen in the live web version, but we can't look at the Routing Graph there.

**Screenshots & Logs**

Locate the relevant major road:

<img width="786" height="615" alt="Image" src="https://github.com/user-attachments/assets/ee1ab483-c381-494c-96fd-67a0cb0caaa7" />

After enabling "Show Routing Graph":

<img width="275" height="685" alt="Image" src="https://github.com/user-attachments/assets/60393b65-fc0d-47d9-9e18-794d7bdea95e" />

**Additional Context**

As per Hong Kong regulations documented on the [Default Speed Limits page](https://wiki.openstreetmap.org/wiki/Default_speed_limits), all roads shall have an implicit 50 km/h speed limit unless otherwise specified.

The relevant roads are correctly classified as "HKG" thanks to #3246. In addition, the relevant roads do not have any OSM-provided `maxspeed=*` value. Therefore, they should have the implicit 50 km/h speed limit.

In this case, the GraphHopper encoded value `max_speed=50.0` is correct.

As far as I have experimented with GraphHopper before and have followed the major GraphHopper update news, currently GraphHopper will use the `*_average_speed` to compare between different paths during CH processing.

[...]

PR #3249, opened 2025-12-12
Limit motor vehicle speeds by country default speed limits

Vectorial1024:
Discussion on #3248 has revealed that GraphHopper did not respect local default speed limits when determining/estimating path journey times, which may result in underestimated journey times and hypothetical unlawful speeding.

Respecting local default speed limits should be well-accepted by everyone. Therefore, this PR modifies the default profiles of select vehicles so that their travel times do not exceed local default speed limits. Biking and walking speeds are not modified for now.

----------

Regarding https://github.com/graphhopper/graphhopper/issues/3248#issuecomment-3645958515 , it turns out that this PR now requires adding the `max_speed` encoded value to the list of encoded values. Missing this value will cause GraphHopper to fail to start with the error of `java.lang.IllegalArgumentException: Encoded values missing: max_speed.`.

This may need to be noted in the changelogs.

### DIFF
--- a/core/src/main/resources/com/graphhopper/custom_models/bus.json
+++ b/core/src/main/resources/com/graphhopper/custom_models/bus.json
@@ -1,5 +1,5 @@
 // to use this custom model you need to set the following option in the config.yml
-// graph.encoded_values: max_weight, max_width, max_height, bus_access, road_class, car_average_speed
+// graph.encoded_values: max_weight, max_width, max_height, bus_access, road_class, car_average_speed, max_speed
 // profiles:
 //    - name: bus
 //      custom_model_files: [bus.json]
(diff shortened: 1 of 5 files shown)


karussell:
> it turns out that this PR now requires adding the max_speed encoded value to the list of encoded values

Could you please add this to the documentation for each of the custom models (first lines -> encoded_values)? And also add a short notice about this in the change log (via linking this PR `#3249`).

Vectorial1024:
Hopefully I understood your requirements correctly and completed the requested changes.

karussell:
Thanks! Additionally I meant directly in e.g. the car.json (and the others too). See the 2nd line:

https://github.com/graphhopper/graphhopper/blob/master/core/src/main/resources/com/graphhopper/custom_models/car.json#L2

Vectorial1024:
@karussell see https://github.com/graphhopper/graphhopper/pull/3249/changes#diff-84c0e87629ad337383db02ce29a72c8525ef045caff67200a028aeddc6ae8a63R2 , etc.; I think I got it right.

karussell:
Ups, sorry. Somehow I viewed only the first commit (the one of the notification).

PR #3251, opened 2025-12-13
bike profiles: more consistent highway tag classification

karussell:
To reduce complexity for #3227 I moved one of the inconsistencies into a separate PR. /cc @ratrun.

In #3015 we reduced priority for tertiary. We should do this for unclassified too as it is very similar to tertiary.

Furthermore I suggest to remove tertiary&unclassified from preferred highways. They are just normal highways if they come without bike lane or "bike infrastructure".  (IMO we should do the same for service&residential but let's do this later.)

For racingbike the state for residential&service is confusing in master: residential was in the "preferred tag" but also was avoided later on. I cleaned this up and put it just into avoidHighwayTags.

I also made the penalty for "tunnel" less severe.


core/src/test/java/com/graphhopper/routing/util/parsers/BikeCustomModelTest.java
```diff
@@ -362,7 +362,7 @@ public void testHandleWayTagsInfluencedByBikeAndMtbRelation() {
         rel.setTag("route", "mtb");
         rel.setTag("network", "lcn");
         edge = createEdge(way, rel);
-        assertEquals(1.2, p.getEdgeToPriorityMapping().get(edge, false), 0.01);
+        assertEquals(1.0, p.getEdgeToPriorityMapping().get(edge, false), 0.01);
```
karussell: This is actually I how uncovered #3250 ... so this will be fixed once both are merged.

core/src/main/java/com/graphhopper/routing/util/parsers/BikePriorityParser.java
```diff
@@ -17,9 +17,13 @@ public BikePriorityParser(DecimalEncodedValue priorityEnc, DecimalEncodedValue s
 
         addPushingSection("path");
 
+//        preferHighwayTags.add("road");
```
ratrun, line 20: This looks like a temporary change. It should be removed.

core/src/main/java/com/graphhopper/routing/util/parsers/MountainBikePriorityParser.java
```diff
@@ -19,14 +21,16 @@ public MountainBikePriorityParser(EncodedValueLookup lookup) {
     protected MountainBikePriorityParser(DecimalEncodedValue speedEnc, DecimalEncodedValue priorityEnc) {
         super(priorityEnc, speedEnc);
 
-        preferHighwayTags.add("road");
         preferHighwayTags.add("track");
         preferHighwayTags.add("path");
+
```
ratrun, line 26: This looks like a temporary change. It should be removed.

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
```diff
@@ -20,12 +20,16 @@ protected RacingBikePriorityParser(DecimalEncodedValue priorityEnc, DecimalEncod
...
         addPushingSection("path");
 
-        preferHighwayTags.add("road");
         preferHighwayTags.add("secondary");
         preferHighwayTags.add("secondary_link");
-        preferHighwayTags.add("tertiary");
-        preferHighwayTags.add("tertiary_link");
-        preferHighwayTags.add("residential");
+
+//        preferHighwayTags.add("road");
```
ratrun, line 26: Removal of road is fine

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
```diff
@@ -20,12 +20,16 @@ protected RacingBikePriorityParser(DecimalEncodedValue priorityEnc, DecimalEncod
...
+//        preferHighwayTags.add("unclassified");
```
ratrun, line 27: I think we should keep this preference for tertiary, tertiary_link and unclassified the racingbike profile.

core/src/test/java/com/graphhopper/routing/util/parsers/RacingBikeTagParserTest.java
```diff
@@ -246,37 +246,37 @@ public void testPriority_avoidanceOfHighMaxSpeed() {
         ReaderWay osmWay = new ReaderWay(1);
         osmWay.setTag("highway", "tertiary");
         osmWay.setTag("maxspeed", "50");
-        assertPriorityAndSpeed(encodingManager, priorityEnc, speedEnc, parsers, PREFER, 24, osmWay);
+        assertPriorityAndSpeed(encodingManager, priorityEnc, speedEnc, parsers, UNCHANGED, 24, osmWay);
```
ratrun, line 249: See my comment in the code section. This should remain PREFER. The same applies for the other changes in this file as well. But the slight improvement for tunnel is OK.

core/src/main/java/com/graphhopper/routing/util/parsers/BikePriorityParser.java
karussell: Do you mean the code should be removed or the change should be reverted? 

As I wrote in the PR:

> I suggest to remove tertiary&unclassified from preferred highways.

Any reason why you would keep them preferred?  (if you comment was about this)

The idea was to keep them neutral to reduce artifacts like in #3015. And preferring roads more based on bicycle infrastructure. Note that if there is a secondary vs. tertiary choice then still the tertiary will be picked because of the penalty for secondary.

ratrun: I'm fine with the changes. Please remove the currently commented code completely.

core/src/main/java/com/graphhopper/routing/util/parsers/MountainBikePriorityParser.java
ratrun: Same here, please remove the commented code.

PR #3252, opened 2025-12-13
remove scenic=yes

karussell:
`scenic=yes` might be "nice" but not necessarily "nice" for bikes. See the [wiki page](https://wiki.openstreetmap.org/wiki/Key:scenic)? 

E.g. see [this OSM way](https://www.openstreetmap.org/way/169798802) which explicitly lists `class:bicycle=-2` and [we choose the secondary](https://graphhopper.com/maps/?point=48.059781%2C11.376165&point=48.016933%2C11.351414&profile=bike) because of that. With this fix everything is fine (at least locally)

image


karussell:
Ok, yes. And thanks for the review.

PR #3256, opened 2025-12-19
move country rules into parser

karussell:
Similar to #3111 move country handling for road_access into parser. And as a result we can remove country rules.

Furthermore access handling for bike and foot will work automatically once country-specific handling is added.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
@@ -140,6 +140,6 @@ public class OSMRoadAccessParser<T extends Enum> implements TagParser {
     }
 
     public static OSMRoadAccessParser<FootRoadAccess> forFoot(EnumEncodedValue<FootRoadAccess> roadAccessEnc) {
-        return new OSMRoadAccessParser<>(roadAccessEnc, toOSMRestrictions(TransportationMode.BIKE), FOOT_HANDLER, FootRoadAccess::find);
+        return new OSMRoadAccessParser<>(roadAccessEnc, toOSMRestrictions(TransportationMode.FOOT), FOOT_HANDLER, FootRoadAccess::find);
     }
 }


core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
```diff
@@ -80,28 +85,49 @@ private T getRoadAccess(String tagValue, T accessValue) {
...
-                return Arrays.asList("psv", "motor_vehicle", "vehicle", "access");
-            case BUS:
-                return Arrays.asList("bus", "psv", "motor_vehicle", "vehicle", "access");
-            case HOV:
-                return Arrays.asList("hov", "motor_vehicle", "vehicle", "access");
-            default:
-                throw new IllegalArgumentException("Cannot convert TransportationMode " + mode + " to list of restrictions");
-        }
+        return switch (mode) {
+            case FOOT -> Arrays.asList("foot", "access");
```
otbutz: I'd switch to `List.of(...)` while we're at it.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
```diff
@@ -80,28 +85,49 @@ private T getRoadAccess(String tagValue, T accessValue) {
         return accessValue;
     }
 
+    public interface RoadAccessDefaultHandler<T> {
```
otbutz, line 88:
```suggestion
    @FunctionalInterface
    public interface RoadAccessDefaultHandler<T> {
```

core/src/test/java/com/graphhopper/reader/osm/OSMReaderTest.java
```diff
@@ -910,10 +909,9 @@ public void testCountries() throws IOException {
         EncodingManager em = new EncodingManager.Builder().add(roadAccessEnc).build();
         OSMParsers osmParsers = new OSMParsers();
         osmParsers.addWayTagParser(new OSMRoadAccessParser<>(roadAccessEnc,
-                OSMRoadAccessParser.toOSMRestrictions(CAR), RoadAccess::countryHook, RoadAccess::find));
+                OSMRoadAccessParser.toOSMRestrictions(CAR), OSMRoadAccessParser.CAR_HANDLER, RoadAccess::find));
```
otbutz: Should we use dedicated subclasses or static factory methods instead of having to pass multiple transportation dependent parameters that need to match?

e.g

```java
osmParsers.addWayTagParser(new OSMCarAccessParser(roadAccesEnc));
```

or 

```java
osmParsers.addWayTagParser(OSMRoadAccessParser.forCar(roadAccesEnc));
```

otbutz (changes requested):

core/src/test/java/com/graphhopper/reader/osm/OSMReaderTest.java
karussell: > OSMRoadAccessParser.forCar

good idea :+1:

PR #3257, opened 2025-12-29
Use unsigned short for pillar node count & reduce geometry header size

wltrr:
As mentioned in #3232, `BaseGraph.java` contained a bug when a `wayGeometry` consisted of more than 32,767 pillar nodes. This triggered overflows. 

In #3233 I provided a fix, however @karussell suggested to instead store the number of pillar nodes in an *unsigned short*. (I deleted my old fork accidentally closing that PR). 

As the old implementation did not cause problems in the past, we know that ~32,000 pillar nodes is enough. Even though moving to an unsigned short increases this limit to 65,535 nodes.

I added two tests:
- `testMaxPillarNodes` and `testPillarNodesOverflow` in `BaseGraphTest` to verify the new capacity (65,535 nodes) and ensure `IllegalArgumentException` is correctly thrown for overflows (>65,535).

- Furthermore, I updated `AbstractGraphStorageTester` to now have 2-byte geometry headers instead of 3-byte geometry headers.

I ran an import of the `brandenburg-251228.osm.pbf` from [Geofabrik](https://download.geofabrik.de/europe/germany/brandenburg.html).
The resulting graph-caches show a small reduction in size of the geometry file from 18.9mb to 17.8mb.

Finally, I increased `VERSION_GEOMETRY` in `Constants.java` to handle the backward compatibility break.

### DIFF
--- a/core/src/main/java/com/graphhopper/util/Constants.java
+++ b/core/src/main/java/com/graphhopper/util/Constants.java
@@ -64,7 +64,7 @@ public class Constants {
     public static final int VERSION_EM = 4;
     public static final int VERSION_SHORTCUT = 10;
     public static final int VERSION_NODE_CH = 0;
-    public static final int VERSION_GEOMETRY = 7;
+    public static final int VERSION_GEOMETRY = 8;
     public static final int VERSION_TURN_COSTS = 0;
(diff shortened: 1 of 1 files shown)


karussell:
Nice, thanks a lot. This looks good!

PR #3259, opened 2026-01-05
country rules for bike and foot

### LINKED ISSUE #2051: Don't walk on a trunk
You shouldn't walk on a trunk road as a pedestrian.

https://graphhopper.com/maps/?point=47.390219%2C0.671404&point=47.399373%2C0.636318&locale=fr-fr&vehicle=foot&weighting=fastest&elevation=true&turn_costs=false&use_miles=false&layer=Omniscale

Capture d’écran 2020-06-02 à 14 45 23

karussell:
This PR is based on the changes in #3256. Fixes #2051.

I used [the wiki](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions) for the access rules for bike & foot. This avoids problems [like this](https://graphhopper.com/maps/?point=50.45259%2C4.57512&point=50.441932%2C4.551656&profile=foot) where despite avoiding trunk we still use it.

(Using a roadClass based switch would result in less code lines but having this country-based is easier to maintain I think.)

Note: [motorroad=yes skipping was later removed](https://github.com/graphhopper/graphhopper/commit/efa8cd444a73fa04977a67fe6db197158698445e) from the RoadAccessDefaultHandlers as [according to the wiki](https://wiki.openstreetmap.org/wiki/Tag:motorroad%3Dyes) it done for all countries and we already do it in FootAccessParser and BikeCommonAccessParser anyway. For countries that do not have a motorroad tag mapped the statement will do nothing.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
@@ -85,6 +85,7 @@ public class OSMRoadAccessParser<T extends Enum> implements TagParser {
         return accessValue;
     }
 
+    @FunctionalInterface
     public interface RoadAccessDefaultHandler<T> {
         T getAccess(ReaderWay readerWay, Country country);
     }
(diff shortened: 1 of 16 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
```diff
@@ -89,8 +89,13 @@ public interface RoadAccessDefaultHandler<T> {
         T getAccess(ReaderWay readerWay, Country country);
     }
 
+    public static RoadClass getRoadClass(ReaderWay readerWay) {
+        String hw = readerWay.getTag("highway", "");
+        return RoadClass.find(hw.endsWith("_link") ? hw.substring(0, hw.length() - 5) : hw);
```
otbutz, line 95: Is there a reason why we don't move this into `RoadClass.find()` ? We could also get rid of the whole substring handling if we use a simple map internally:

```java
public enum RoadClass {
    OTHER, MOTORWAY, TRUNK, PRIMARY, SECONDARY, TERTIARY, RESIDENTIAL, UNCLASSIFIED,
    SERVICE, ROAD, TRACK, BRIDLEWAY, STEPS, CYCLEWAY, PATH, LIVING_STREET, FOOTWAY,
    PEDESTRIAN, PLATFORM, CORRIDOR, CONSTRUCTION, BUSWAY;

    public static final String KEY = "road_class";

	private static final Set<RoadClass> LINK_CLASSES = Set.of(MOTORWAY, TRUNK, PRIMARY, SECONDARY, TERTIARY);
	private static final Map<String, RoadClass> TAG_MAP;
	static {
		Map<String, RoadClass> map = new HashMap<>();
		for (RoadClass rc : values()) {
			if (rc == OTHER) {
				continue;
			}
			String tag = Helper.toLowerCase(rc.name());
			map.put(tag, rc);
			if (LINK_CLASSES.contains(rc)) {
				map.put(tag + "_link", rc);
			}
		}
		TAG_MAP = Map.copyOf(map);
	}

    public static EnumEncodedValue<RoadClass> create() {
        return new EnumEncodedValue<>(RoadClass.KEY, RoadClass.class);
    }

    @Override
    public String toString() {
        return Helper.toLowerCase(super.toString());
    }

    public static RoadClass find(String name) {
        if (name == null || name.isEmpty())
            return OTHER;
        return TAG_MAP.getOrDefault(Helper.toLowerCase(name), OTHER);
    }
}
```

karussell: The IMO cleaner method would be to use the roadClassEnc (that contains the already parsed highway tag) in the OSMRoadAccessParser but for simplicity I decided against it and parse it again. I see it as an internal thing and RoadClass.find should not do this logic and stay simple.

Issue #3138, opened 2025-03-06
FerrySpeedCalculator underestimates shuttle trains


otbutz:
**Describe the bug**

GraphHopper uses the same fallback speed assumptions for both regular ferries and shuttle trains. This results in a huge difference between the actual duration and the one calculated by GH.

**To Reproduce**

https://graphhopper.com/maps/?point=54.906554%2C8.306643&point=54.781226%2C8.829625&profile=car&layer=Omniscale

GraphHopper favors a detour over Denmark with a duration of 2h37m and 100km distance:

![Image](https://github.com/user-attachments/assets/99b9d012-c52d-4a63-b46d-a55f03a86673)

**Expected behavior**

The shuttle train should be preferred:

![Image](https://github.com/user-attachments/assets/fa723433-93e4-423e-8bf2-ddc5f5a83657)

[GraphHopperMaps](https://graphhopper.com/maps/?point=54.906554%2C8.306643&point=54.781226%2C8.829625&profile=car&layer=Omniscale&custom_model=%7B%22distance_influence%22%3A15%2C%22priority%22%3A%5B%7B%22if%22%3A%22in_area1%22%2C%22multiply_by%22%3A%220%22%7D%5D%2C%22speed%22%3A%5B%5D%2C%22areas%22%3A%7B%22type%22%3A%22FeatureCollection%22%2C%22features%22%3A%5B%7B%22type%22%3A%22Feature%22%2C%22geometry%22%3A%7B%22type%22%3A%22Polygon%22%2C%22coordinates%22%3A%5B%5B%5B8.417337%2C55.096981%5D%2C%5B8.611658%2C55.102089%5D%2C%5B8.616464%2C55.021081%5D%2C%5B8.41253%2C55.01872%5D%2C%5B8.417337%2C55.096981%5D%5D%5D%7D%2C%22properties%22%3A%7B%7D%2C%22id%22%3A%22area1%22%7D%5D%7D%7D) (forced via custom model)

otbutz:
OSRM and Valhalla prefer the shuttle train route and estimate the duration with about 45 minutes:

https://valhalla.openstreetmap.de/directions?profile=car&wps=8.30653775%2C54.906496950000005%2C8.82962380855796%2C54.781222579319426

https://routing.openstreetmap.de/?z=12&center=54.842806%2C8.576850&loc=54.906497%2C8.306538&loc=54.781223%2C8.829624&hl=en&alt=0&srv=0

[...]

PR #3265, opened 2026-01-14
Increase fallback speed for shuttle trains

otbutz:
Fixes #3138

Issue #3264, opened 2026-01-14
Use turn penalty for ferry wait times


otbutz:
Using turn penalties, this could be replaced with an initial penalty for boarding the ferry:

https://github.com/graphhopper/graphhopper/blob/88335ad789af3f54f920b537baac8868445073ef/core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java#L28-L33

PR #3266, opened 2026-01-14
Ferry speed fix

### LINKED ISSUE #2532: Better handling for missing ferry duration tags
If there is no duration tag on a ferry we currently use `speed=5km/h`, or even smaller if the distance is below 300m. But for longer distance ferries this seems very likely too slow. Also I think not tagging duration for such longer ferries is a mapping error and if this happens frequently we should report it to OSM.

Also see here: https://github.com/graphhopper/graphhopper/pull/2528/files#r819521430

karussell:
This should fix #2532 and #3138.

An [intermediate commit](https://github.com/graphhopper/graphhopper/pull/3266/changes/8f1b7974166376906b46c324461a9a1c34df7e08) contains a "proper" ferry relation handling which picks the `duration` tag of the relation if the way does not have it but it requires to have the ways also tagged with `route=ferry/shuttle_train` and also often the `duration` tag is missing there too so this does not solve the issue and is rather complicated and so I picked @otbutz suggestion to fix it according to #2532.

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
+++ b/core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
@@ -515,7 +515,7 @@ public class OSMReader {
         // tag will be present if 1) isCalculateWayDistance was true for this way, 2) no OSM nodes were missing
         // such that the distance could actually be calculated, 3) there was a duration tag we could parse, and 4) the
         // derived speed was not unrealistically slow.
-        way.setTag("duration:seconds", durationInSeconds);
+        way.setTag("duration_in_seconds", durationInSeconds);
     }
 
     static String fixWayName(String str) {
--- a/core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
+++ b/core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
@@ -22,9 +22,9 @@ public class FerrySpeedCalculator implements TagParser {
     static double getSpeed(ReaderWay way) {
         // todo: We cannot account for waiting times for short ferries as speed is slower than the slowest we can store
 
-        // OSMReader adds the artificial 'duration:seconds' and 'way_distance' tags that we can
+        // OSMReader adds the artificial 'duration_in_seconds' and 'way_distance' tags that we can
         // use to set the ferry speed. Otherwise we need to use fallback values.
-        long durationInSeconds = way.getTag("duration:seconds", 0L);
+        long durationInSeconds = way.getTag("duration_in_seconds", 0L);
         if (durationInSeconds > 0) {
             // a way can consist of multiple edges like https://www.openstreetmap.org/way/61215714 => use way_distance
             double waitTime = 30 * 60;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### 12.0 [not yet released]
 
-- OSMReader no longer sets the artificial speed_from_duration tag but instead uses duration:seconds, when the duration tag is present (#3266)
+- OSMReader no longer sets the artificial speed_from_duration tag but instead uses duration_in_seconds, when the duration tag is present (#3266)
 - country rules were moved into parsers and are now enabled by default
 - speeds generated from highway class now respects country-specific default speed limits, but the max_speed encoded value is now required; see #3249
 
--- a/core/src/test/java/com/graphhopper/routing/util/FerrySpeedCalculatorTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/FerrySpeedCalculatorTest.java
@@ -42,7 +42,7 @@ class FerrySpeedCalculatorTest {
         ReaderWay way = new ReaderWay(1);
         way.setTag("route", "ferry");
         way.setTag("way_distance", 30_000.0);
-        way.setTag("duration:seconds", 1800L);
+        way.setTag("duration_in_seconds", 1800L);
 
         EdgeIntAccess edgeIntAccess = new ArrayEdgeIntAccess(1);
         int edgeId = 0;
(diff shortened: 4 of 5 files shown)


otbutz:
I'd go for a simpler speed heuristic (#3265) and solve the waiting time using turn penalties as proposed in #3264

karussell:
The issue #3264 wouldn't work for bike and foot as we have them deployed via node-based (and turn_penalty only works for edge-based). And also it is not that simple as `ferry_speed` is profile-independent.

core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
```diff
@@ -32,20 +32,22 @@ static double getSpeed(ReaderWay way) {
             // we reduce the speed to account for waiting time (we increase the duration by 40%)
             return Math.round(speedInKmPerHour / 1.4);
```
otbutz: Shouldn't this factor also change based on the length of the way? Increasing the duration by 40% for a long ferry route seems excessive.

karussell: Good idea. (This was introduced nearly [13 years ago](https://github.com/graphhopper/graphhopper/commit/7d629b30dd7129a455ac55fe3c17226c5b2a21a1) :))

otbutz: Nothing is more permanent than a temporary solution, I guess :smile:

karussell: Have a look into the latest two commits. I prefer the last one.

otbutz: I like the delay accounting, but having this directly in the OSMReader for all duration tags seems wrong to me. We know that we need to account for boarding in the case of ferries and shuttle trains, but it might be inaccurate for other modes of transportation, such as cable cars.

karussell: Can you elaborate what you mean e.g. an OSM example? Note that currently only ferry and shuttle_train are supported (FerrySpeedCalculator.isFerry==true called via isCalculateWayDistance)

otbutz: While the FerrySpeedCalculator is currently the only consumer of the virtual tag `speed_from_duration`, it gets populated in the neutral `preprocessWay` function of the OSMReader.

Principle of least surprise: I wouldn't expect ferry related time corrections while handling a tag that might also be used for non-ferry routing.

Can we push the delay calculation down to the FerrySpeedCalculator?

karussell: We could remove the artificial tag `speed_from_duration` and use `way_distance` and a new artificial tag like `duration_in_s(ec)` or something.

karussell: Have committed a possible solution.

core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
```diff
@@ -20,32 +20,29 @@ public static boolean isFerry(ReaderWay way) {
...
+        // OSMReader adds the artificial 'duration_in_seconds' and 'way_distance' tags that we can
         // use to set the ferry speed. Otherwise we need to use fallback values.
-        double speedInKmPerHour = way.getTag("speed_from_duration", Double.NaN);
-        if (!Double.isNaN(speedInKmPerHour)) {
-            // we reduce the speed to account for waiting time (we increase the duration by 40%)
-            return Math.round(speedInKmPerHour / 1.4);
+        long durationInSeconds = way.getTag("duration_in_seconds", 0L);
+        if (durationInSeconds > 0) {
+            double waitTime = 30 * 60;
+            double wayDistance = way.getTag("way_distance", Double.NaN);
```
otbutz, line 31:
```suggestion
            double wayDistance = way.getTag("way_distance", Double.NaN);
            if (Double.isNaN(wayDistance))
                 throw new IllegalStateException("No 'way_distance' set for edge created for way: " + way.getId());
```

otbutz: btw what's the difference between `edge_distance` and `way_distance` ?

otbutz:
Hijacking the current topic, but would you mind taking a look at #3143 ?

core/src/main/java/com/graphhopper/routing/util/FerrySpeedCalculator.java
karussell: An OSM way might be split into several GraphHopper edges, hence the edge_distance might be only a smaller part and not the distance that should be used with the `duration` tag, which is for the entire OSM way.

core/src/main/java/com/graphhopper/reader/osm/OSMReader.java
```diff
@@ -516,7 +515,7 @@ protected void preprocessWay(ReaderWay way, WaySegmentParser.CoordinateSupplier
         // tag will be present if 1) isCalculateWayDistance was true for this way, 2) no OSM nodes were missing
         // such that the distance could actually be calculated, 3) there was a duration tag we could parse, and 4) the
         // derived speed was not unrealistically slow.
-        way.setTag("speed_from_duration", speedInKmPerHour);
+        way.setTag("duration:seconds", durationInSeconds);
```
otbutz: Should we add the `gh:` prefix like we do it for `gh:barrier_edge`? That namespaces make it much clearer that it's some viirtual GraphHopper populated tag IMHO.

karussell: Unfortunately we were not that consistent in the past and would have to change way_distance, edge_distance, ... too

Issue #3135, opened 2025-03-02
missing turn instruction


karussell:
There should be a "slight turn right" (or "keep right") and then "keep left":

https://graphhopper.com/maps/?point=50.59943%2C7.03619&point=50.59081%2C7.02795&profile=car

limes11:
Could anyone take a look at this please ?

https://www.openstreetmap.org/way/536612040

destination:lanes and turn:lanes are correct. Clearly there needs to be an indication to the driver to "keep right" in order to follow straight on. 

Thank you

modhtom:
Hi @karussell, I checked this issue and found the root cause.

The missing instruction comes from InstructionsFromEdges. There is logic that hides turn instructions when the Road Class stays the same. The idea is that if the class does not change, the driver is just following the same road.

I updated this logic to also check how ambiguous the other path is (otherDelta).

If the other path turns away clearly, we treat it as an obvious exit and keep the instruction hidden.
If the other path is also mostly straight, we treat it as a confusing fork and allow the "Keep Left/Right" instruction.
```Java
final double OBVIOUS_EXIT = 0.22;
if ((roadClass == prevRoadClass && link == prevLink) 
&& (otherRoadClass != prevRoadClass || otherLink != prevLink) 
&& effectiveOtherDelta > OBVIOUS_EXIT) {
```
The detection uses different thresholds based on road type.

Motorways use a stricter threshold because shallow forks are common and important.

Tertiary roads use a looser threshold to reduce noise.
```Java
// Motorways/Trunks: High speed means even shallow splits (12-15) are significant forks.
// We lower the threshold (0.85) so we don't accidentally suppress valid instructions.
// Tertiary: Slower roads often have "wobbly" intersections.
// We increase the threshold (1.15) to suppress noise/false positives.
double roadClassFactor = 1.0;
if (roadClass == RoadClass.MOTORWAY || roadClass == RoadClass.TRUNK) {
      roadClassFactor = 0.85;
} else if (roadClass == RoadClass.TERTIARY) {
      roadClassFactor = 1.15;
}
// How "sharp" the alternative turn looks given the road context.
double effectiveOtherDelta = Math.abs(otherDelta) * roadClassFactor;
```
I tested this locally, The reported route now correctly shows "Keep Right" then "Keep Left".
Normal Autobahn exits are still silent as expected.

[...]

PR #3271, opened 2026-01-22
Fix/issue 3135 ambiguous forks

modhtom:
Fixes issue #3135 where necessary "Keep Left" or "Keep Right" instructions were ignored on motorways when the road class and name remained constant.
Previously `InstructionsFromEdges` would aggressively Ignore instructions if roadClass == prevRoadClass.

I modified the suppression logic in getTurn():
The new check is applied only to MOTORWAY and TRUNK. All other road classes retain the original behavior.
For motorways/trunks, we now check the angle of the alternative path.
If the link peels off sharply (effectiveOtherDelta > 0.22), we assume it is visually obvious and return Instruction.IGNORE.
If the link angle is shallow (<= 0.22), we skip the ignore command, this allows the subsequent logic to generate the necessary "Keep Left/Right" instruction.

Verified at the reported location (A 565) and the A9 example. Both now correctly generate "Keep Right" / "Keep Left" instructions.
Image
Image


modhtom:
I took a look into the routing logic and found a few hidden traps in my initial approach. I've patched them and added 19 tests to cover different scenarios.

1. The engine was aggressively ignoring valid highway forks if `roadClass == prevRoadClass`, causing missed Keep Left / Keep Right instructions. I updated the logic to evaluate the fork angles instead for Motorways and Trunks.

2. There was a legacy `0.15` rad threshold designed to output `CONTINUE_ON_STREET` to prevent Keep instructions at off-ramps. However, this clashed with the new 0.22 threshold, creating a dead zone (0.16 to 0.21) where valid forks were being hijacked and turned into "Continue" instructions. Added `continueThreshold` for highways to 0.22 to close this gap and 0.15 for anything else.

3. By opening up the angle math for Motorways, the engine bypassed the original `return Instruction.IGNORE;` safety net and started evaluating all intersecting roads. This accidentally generated Keep instructions if a lower-class road (like a `PRIMARY` country road) intersected a Motorway at a shallow angle. I restored the safety net to explicitly ignore non-link roads of a different class.

PR #3273, opened 2026-01-23
feat: mapmatching - accept GeoJSON, WKT, WKB and Polyline formats

### LINKED ISSUE #2773: Map Matching with list of Coordinates (not gpx file)
Hello everyone,

The `match` endpoint requires the coordinates to be in a `.gpx` file. It would be creat to have the possibility to directly send all the coordinates in the request body. Similar to the mapbox api. This was also suggested in the forum https://discuss.graphhopper.com/t/is-it-possible-to-bring-a-sequence-of-matching-dots-using-the-graphhopper-map-matching-application/5862/5

This should not be to difficult to integrate and would just require a few changes to this file:
 https://github.com/graphhopper/graphhopper/blob/d7817079fb1f6938c043d64f66add63cc5816d80/web-bundle/src/main/java/com/graphhopper/resources/MapMatchingResource.java

What do you think? Should this functionality be implemented? With a little bit of guidance I could get started and create a PR.

Thank you in advance,
Sören

karololszacki:
This PR adds support for MapMatching in formats:
- GeoJSON
- Well-Known Text (WKT) for LineString
- Well-Known Binary (WKB) also for LineString
- Polyline (5 and 6)

This should (partly) resolve #2773, which asked for non-gpx MapMatching.

GeoJSON parsing needed one additional import from JTS (Java Topology Suite), namely their reader.

I tried to separate changes into logical commits / steps.
I am not primarily a Java dev, but I have tried my best to stick to your conventions and to best practices in general.

### DIFF
--- a/web-bundle/src/main/java/com/graphhopper/resources/MapMatchingResource.java
+++ b/web-bundle/src/main/java/com/graphhopper/resources/MapMatchingResource.java
@@ -38,6 +38,7 @@ import com.graphhopper.storage.index.LocationIndexTree;
 import com.graphhopper.util.*;
 import com.graphhopper.util.details.PathDetailsBuilderFactory;
 import com.graphhopper.util.shapes.GHPoint;
+import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.io.WKBReader;
 import org.locationtech.jts.io.WKTReader;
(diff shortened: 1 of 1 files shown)


web/src/test/java/com/graphhopper/application/resources/MapMatchingResourceTest.java
```diff
@@ -122,15 +215,5 @@ public void testEmptyGPX() {
...
-    private LineString readWktLineString(String wkt) {
-        WKTReader wktReader = new WKTReader();
-        LineString expectedGeometry = null;
-        try {
-            expectedGeometry = (LineString) wktReader.read(wkt);
-        } catch (ParseException e) {
-            e.printStackTrace();
-        }
-        return expectedGeometry;
-    }
```
karololszacki, line 134: This func was actually moved from the test file into `MapMatchingResource` in ee36913 
Recently it got replaced with the generic version

Issue #3029, opened 2024-06-24
`living_street` for bike is avoided 


caspg:
**Describe the bug**

It looks that `living_street` is avoided for bike and mtb profiles but racing bike seems routing fine. Where the difference come from?

Is [PUSHING_SECTION_SPEED](https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java#L71C42-L71C63) assigned to `living_street` affecting routing that much? But how racing bike profile is overriding it?

Also, speed for `living_street` should vary country by country and is greater than "walking speed" [osm wiki](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dliving_street). Should there be a way to define that as a country rule? 
We could also use `maxspeed` tag if preset.

**To Reproduce**

[Example route](https://graphhopper.com/maps/?point=53.728202%2C17.508648&point=53.726755%2C17.511033&profile=bike&layer=Omniscale)

![CleanShot 2024-06-24 at 09 09 16@2x](https://github.com/graphhopper/graphhopper/assets/10419439/90c9b7cb-14b4-4ccf-a9c3-f1e61b885d2a)

ratrun:
>  It looks that living_street is avoided for bike and mtb profiles but racing bike seems routing fine. Where the difference come from?

We are de-prioritizing `highway=residential` to SLIGHT_AVOID for race bikes in order to make sure that we stick on  higher classified highways within villages and do not create de-tours over residentials. This we don't do for the bike and MTB profile. So this is where the difference comes from.

Please note that I'm from Austria and we have the rule that the maximum speed for any traffic for living_street is walking speed. I do not remember that such a detailed country dependent table existed when we implemented this about 10 years ago.  I think that we just need to increase the assumed maximum speed to 18km/h. In case that the maxspeed is explicitly tagged to a lower speed we should already now take this as our assumed speed then.

caspg:
@ratrun Thanks for the explanation.

Not sure that setting it as 18km/h would be correct either. Even if we take `maxspeed` into consideration because not every way have it. So we will create incorrect routes for Austria and other countries where `living_speed` have walking speed.

To make it correct there should be a way to set speed based on country.

caspg:
@karussell Changing speed to 6km/h helped. Do you think it should be new default?

[...]

PR #3280, opened 2026-02-10
bike: allow higher speed for living_street if explicitly allowed

karussell:
We had a similar issue here #3029.

But still currently we make unnecessary detours for living_street like [here](https://graphhopper.com/maps/?point=51.099724%2C13.679337&point=51.102439%2C13.681118&profile=bike) (420 vs 820m). Increasing priority while keeping the speed is not a good solution as in other cases the living_street should not be overly preferred.


karussell:
@ratrun do you think this is reasonable in real world?

ratrun:
> @ratrun do you think this is reasonable in real world?

I ran an overpass query and the combination of `highway=living_street` with `bicycle=*` appears rather frequently. And I think it is realistic that bicycles reach more than 6 km/h. Therefore, I think this change will improve those cases. I would merge it.

PR #3286, opened 2026-02-17
Use exact millimeter arithmetic for edge distance summation

easbar:
New edge.getDistance_mm() / setDistance_mm(long) methods allow reading and writing edge distances as exact integer millimeters. Distance accumulation in Path and all path extractors  now uses long arithmetic instead of lossy double summation. QueryGraph adjusts virtual edge distances to exactly match the original edge distance.

Route distances can now be compared exactly between CH and Dijkstra, because virtual node distances no longer have floating-point summation errors.

We could also use unsigned four byte int to store the distance which would allow a maximum edge length of ~4294km instead of ~2147km, but this is probably not needed. Accumulating distances with four bytes would not be compatible with routes >2147km, so if we wanted to reduce memory for these we would have to decrease the distance precision to mm or cm even, but there does not seem to be much benefit doing this.

We can now also reduce the tolerance of our distance comparisons in many tests, but I'll save this for later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### 12.0 [not yet released]
 
-- new edge.getDistance_mm/setDistance_mm methods for exact millimeter distances; distance accumulation now uses exact long arithmetic instead of lossy double summation (#xxx)
+- new edge.getDistance_mm/setDistance_mm methods for exact millimeter distances; distance accumulation now uses exact long arithmetic instead of lossy double summation (#3286)
 - OSMReader no longer sets the artificial speed_from_duration tag but instead uses duration_in_seconds, when the duration tag is present (#3266)
 - country rules were moved into parsers and are now enabled by default
 - speeds generated from highway class now respects country-specific default speed limits, but the max_speed encoded value is now required; see #3249


core/src/main/java/com/graphhopper/routing/querygraph/QueryOverlayBuilder.java
```diff
@@ -203,11 +208,38 @@ private double distanceOfSnappedPointToPillarNode(Snap o) {
                             fullPL.get(fullPL.size() - 1), fullPL.size() - 2,
                             fullPL, closestEdge, virtNodeId - 1, adjNode);
 
+                // todonow: should we really adjust distances even when addedEdges=false? then again addedEdges is never false in any of our tests?!
```
karussell: there is this todonow and 2 others. but does not look critical :)

easbar: Thanks for taking a look. I forgot to push the latest changes. I think adedEdges is never false, but if the virtual edge arrays were empty we would divide by zero (see latest commits after https://github.com/graphhopper/graphhopper/pull/3286/commits/99a0a2162d9248b9de970e242b50cad978c79363).

PR #3287, opened 2026-02-18
pmtiles elevation data

karussell:
This PR allows us to read pmtiles elevation data (direct file access, not via web API). From the locally available pmtiles file it creates several temporary files which it memory maps for fast access. Similar to what happens for the other elevation providers (except it is not using DataAccess as for these small files it is not really necessary and a small overhead). 

In a future version it might download the portions via http range queries from a hosted pmtiles file, but wasn't goal of this PR. Still you can use the official pmtiles tool to create extracts:

```
# germany
pmtiles extract https://download.mapterhorn.com/planet.pmtiles de.pmtiles  --bbox=5.8663,47.2701,15.0419,55.0584 --minzoom=11 --maxzoom=12
```

Unfortunately I had to pick the library `com.github.usefulness:webp-imageio` which is a wrapper of `libwebp`. The 100% Java library `com.twelvemonkeys.imageio:imageio-webp` has bugs in its webp implementation that gives wrong (or slightly off?) elevation data for certain cases (although the tests pass and certain regions like saxony are fine! but e.g. germany does not pass at coast.) And when I tried to export those coastal tiles to png I got a different strange error, [which seems to be already known](https://github.com/haraldk/TwelveMonkeys/issues/978). I tested version `3.13.1` released a few days ago. Btw: instead of the legacy native stuff we could use [NightMonkeys](https://github.com/gotson/NightMonkeys) which uses FFM, but this would require JDK22.

trying to get it faster/smaller:

 - [x] tile cache: super tiles that combine 8x8 tiles in one file. It is even slower!? (e765c8e ... I run it twice where only the cache was used and still nearly 2x slower for pass2. tested for planet)
 - [x] tile cache: compress small 4x4 block via base elevation and storing the differences in just 1 byte and if not possible store normal raw data. This made the cache 2x smaller on average and also ~10% faster. Speed on 2nd run similar to cgiar but fast NVMe required (with zoom=11). (045232fc1c122c). For zoom=12 a 3GB/s NVMe gave approx. 20% slower times compared to cgiar.
 - [x] use this [go script](https://gist.github.com/karussell/e6ad9918b6cd9913ddba998af43860a2) to extra a single zoom out of the mapterhorn elevation pmtiles to reduce size of it.

Disk space needed for planet and zoom=11 is approx 260GB:

 * 150GB if zoom=11 is extracted from the 660GB planet.pmtiles file (use the go script above)
 * 110GB for the compressed cache files

Disk space needed for planet and zoom=12 is approx 770GB:

 * 350GB if zoom=12 is extracted from the 660GB planet.pmtiles file (use the go script above)
 * 420GB for the compressed cache files

Get the data from the [mapterhorn.com](https://mapterhorn.com/) project.


wipfli:
Great to see this pull request. Thanks @karussell for working on this!

ratrun:
> Get the data from [mapterhorn.com](https://mapterhorn.com/).

Could you please add or replace the link with this better [link](https://mapterhorn.com/data-access/)  for the mapterhorn.com download? It took me a bit to find where this download is actually described.

karussell:
I'd prefer to keep it this way. It is not that hard really :) ... and then we do not have to maintain the URL if they change the URL structure and people also are more likely to read about their attribution requirements etc.

PR #3288, opened 2026-02-19
Add scenic encoded value

ratrun:


karussell:
As scenic is rather rare (<12k) it might make more sense (memory-wise) to use the KVStorage which is very easy via #3284. Did you have a look into it?

A boolean in encoded values could mean 0 more memory but if there is no bit free in the flags (byte-wise aligned) it will increase memory usage by approx 0.5GB (for planet) vs under 100k for the KVStorage.

ratrun:
> As scenic is rather rare (<12k) it might make more sense (memory-wise) to use the KVStorage which is very easy via #3284. Did you have a look into it?

I had this change lying around before you published #3284.  When I looked at #3284 and read about the list of downsides I got reminded that I should publish my PR. But I didn't try out #3284 yet, I'll wait with this until it gets merged.

> A boolean in encoded values could mean 0 more memory but if there is no bit free in the flags (byte-wise aligned) it will increase memory usage by approx 0.5GB (for planet) vs under 100k for the KVStorage.

That is interesting. As I understand your response my PR now does not make sense any longer. Right?

In this case I just would ask to document this fact in #3284 and describe there on a high level the use-cases when to take which method such that further unnecessary effort for implementing new encoded values can be saved.

karussell:
> As I understand your response my PR now does not make sense any longer. Right?

I wouldn't be that fast :) ... the PR is not yet tried under different circumstances and also performance is unclear. Having that said I would be surprised if there are major downsides in this case for very rare tags.

PR #3292, opened 2026-02-23
Use proper encoded value map for exception message in getTurnEncodedValue 

baybatu:
Use proper encoded value map in the error log at `getTurnEncodedValue` to make debugging easy


karussell:
Thanks! Merged without the tests.

PR #3295, opened 2026-02-27
bike_network: reduce boost further

karussell:
Currently [this route](https://graphhopper.com/maps/?point=50.998211%2C11.955872&point=51.06729%2C13.697204&profile=bike) shows that we do not make a good job balancing the bike network preferences compared to the other properties. The boost is too strong and so a much longer and very hilly route is picked as **first** suggestion. See [some statistics](https://gist.github.com/karussell/c3b03df2a3f1fdc179a963a404cd5545).

With this change the route is 20km shorter (2h20 less time), 1.3km less incline, has 6% less unpaved surface and 3.5% less big roads. However the bike network percentage decreased by 14% (this branch 67% vs. master 81.3%).

Other routes to consider when balancing bike networks and incline:

 * [This route](https://graphhopper.com/maps/?point=51.108047%2C13.676413&point=51.108573%2C13.65899&profile=bike) where the more hilly version shouldn't be picked although it is a regional bike network. (This was already working due to [this commit](https://github.com/graphhopper/graphhopper/commit/92799bcc5e315160344ba3196d061e2d1fbaa72a) where the boost was reduced from the initially too strong boost in #3226)
 * international vs regional boost difference becomes obvious in [this example route](https://graphhopper.com/maps/?point=48.064363%2C11.543991&point=48.037781%2C11.521546&profile=bike&layer=TF+Cycle). Note that for one direction the bike network is already chosen and that currently [this secondary](https://www.openstreetmap.org/way/40641663) is incorrectly mapped as bike route (although it is tagged as use_sidepath)
 * [this route](https://graphhopper.com/maps/?point=50.845805%2C13.11945&point=50.846409%2C13.069372&profile=bike&layer=Cyclosm) should not go over the hill and follow more the bike network along the river (in this example following the bike network and avoiding the hill are not contrary but the hilly route is shorter and might be picked)

### DIFF
--- a/core/src/test/java/com/graphhopper/routing/util/parsers/BikeCustomModelTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/parsers/BikeCustomModelTest.java
@@ -240,7 +240,7 @@ public class BikeCustomModelTest {
         osmRel.setTag("network", "icn");
 
         edge = createEdge(way, osmRel);
-        assertEquals(1.7, p.getEdgeToPriorityMapping().get(edge, false), 0.01);
+        assertEquals(1.4, p.getEdgeToPriorityMapping().get(edge, false), 0.01);
 
         // unknown highway tags will be excluded
         way = new ReaderWay(1);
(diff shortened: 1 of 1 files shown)


ratrun:
It seems that the bike routes in Germany prioritize crossing larger cities, which results in this observation. I believe that this is a bit different in Austria, where because of the topology the international bike routes often follow rivers and naturally also cross larger cities.

But, I would not consider the current tuning as a bug. For me the national route D4 is a bit questionable given the observed better alternatives.

Anyway, I tested a bit with your tuned branch in Austria and most of the time the well equipped international/national/regional routes are still taken. Therefore, I'm fine with the changes made in this PR.

PR #3301, opened 2026-03-05
bike dismount should ignore surface+smoothness

karussell:
[Our current route choice here](https://graphhopper.com/maps/?point=51.193154%2C13.362408&point=51.195171%2C13.374167&profile=bike&layer=TF+Cycle) is a detour and more dangerous:

image

than how the bike network was intended.

The problem is the double penalty we give to [this way](https://www.openstreetmap.org/way/229899458) because of the dismount and bad smoothness etc. If we are already pushing the surface shouldn't have a (big) impact how slow it is.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
@@ -170,14 +170,15 @@ public abstract class BikeCommonAverageSpeedParser extends AbstractAverageSpeedP
                         // if explicitly allowed then allow speeds above limit to get more realistic routes and ETAs
                         speed = bikeDesignated ? Math.max(speed, 12) : Math.max(speed, 10);
             }
+
+            // speed reduction if bad surface
+            if (surfaceSpeed != null)
+                speed = Math.min(surfaceSpeed, speed);
+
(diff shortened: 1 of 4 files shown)


karussell:
> I don't have the impression, because at least on very_horrible you won't make 6km/

I did, but I have not yet found a real world example were this makes a difference in a routing choice because the low pushing speed is already a big penalty. Furthermore, probably more important (?), the pushing speed is already 4kmh, not 6km/h. Maybe we should use 6kmh?
One more example, with this change, we are still avoiding `smoothness=very_horrible` [here](https://graphhopper.com/maps/?point=47.739612%2C9.041491&point=47.738934%2C9.04619&profile=bike).

btw: why don't we prefer mtb_rating for the mtb profile? Even with this fix the mtb profile is also avoiding the path. But this is exactly the problem that we have for mtb: the surface often puts a heavy penalty on the way and the mtb_rating boost is not sufficient.

> Did you already test https://github.com/graphhopper/graphhopper/pull/3302 with this example?

Indeed, this already works! (Still 4km/h is rather low in this case.)

ratrun:
> Furthermore, probably more important (?), the pushing speed is already 4km/h, not 6km/h. 

I forgot that it is 4km/h.

> Maybe we should use 6kmh?

No, 4km/h is fine. And probably you are right that in real world 2km/h delta does not make a difference. You convinced me, please merge it!

karussell:
This change was already present in #3302 and is merged now.

PR #3302, opened 2026-03-05
bike speed: consider either surface or smoothness

karussell:
Smoothness should not add to the penalty that surface already did. Instead pick the worst speed of surface and smoothness.

Also the tracktype speed is IMO too small if you compare the description [in the wiki](https://wiki.openstreetmap.org/wiki/Key:tracktype).

Here is the [real world route](https://graphhopper.com/maps/?point=51.152741%2C13.16885&point=51.148501%2C13.14902&profile=bike&layer=TF+Cycle) that fails because of the too low speed we give to [this way](https://www.openstreetmap.org/way/39092421).

Similar problems:

 * for [this way](https://www.openstreetmap.org/way/25057643) which is also part of an international bike network and should not be that much avoided.
 * for [this way](https://www.openstreetmap.org/way/31722931) things gets very interesting as we accept a ~2km detour (for a ~1km way) but this migtht be even fine due to smoothness=horrible? I'm unsure.
 * for [this route](https://graphhopper.com/maps/?point=51.560178%2C13.009727&point=51.584956%2C12.760015&point=51.593463%2C12.58847&profile=bike&layer=TF+Cycle) where the bike network is also avoided. See e.g. [this way](https://www.openstreetmap.org/way/47359191).
 * [this way](https://www.openstreetmap.org/way/307146793) is also part of a bike network and avoided, but probably more because of tracktype=grade3 (smoothness=good). Still related to this PR as we also slightly improved tracktype speeds

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
@@ -171,11 +171,11 @@ public abstract class BikeCommonAverageSpeedParser extends AbstractAverageSpeedP
                         speed = bikeDesignated ? Math.max(speed, 12) : Math.max(speed, 10);
             }
 
-            // pick the smallest of smoothness<->surface, if both are present
             double smoothSpeed = smoothnessFactor.get(smoothnessEnc.getEnum(false, edgeId, edgeIntAccess)) * speed;
 
             // speed reduction if bad surface
             if (surfaceSpeed != null) {
+                // pick the smallest of smoothness<->surface, if both are present
                 speed = Math.max(MIN_SPEED, Math.min(Math.min(surfaceSpeed, speed), smoothSpeed));
             } else {
                 speed = Math.max(MIN_SPEED, smoothSpeed);
--- a/core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
@@ -379,13 +379,11 @@ public class BikeTagParserTest extends AbstractBikeTagParserTester {
         way.setTag("tracktype", "grade4");
         assertEquals(10, getSpeedFromFlags(way), 0.01);
 
-        // pick smallest
+        // pick smallest of highway, tracktype, and applied smoothness speed
         way.setTag("smoothness", "intermediate");
         assertEquals(10, getSpeedFromFlags(way), 0.01);
-
         way.setTag("smoothness", "bad");
         assertEquals(8, getSpeedFromFlags(way), 0.01);
-
         way.setTag("smoothness", "impassable");
         assertEquals(MIN_SPEED, getSpeedFromFlags(way), 0.01);
     }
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
```diff
@@ -166,18 +166,22 @@ else if (isRacingBike)
...
+                    if (bikeAllowed)
                         // if explicitly allowed then allow speeds above limit to get more realistic routes and ETAs
                         speed = bikeDesignated ? Math.max(speed, 12) : Math.max(speed, 10);
             }
-        }
 
-        // speed reduction if bad surface
-        if (surfaceSpeed != null)
-            speed = Math.min(surfaceSpeed, speed);
+            // pick the smallest of smoothness<->surface, if both are present
```
ratrun: Please move this comment down to the if clause before the line and change it a bit into:

`// pick the smallest of smoothness, surface, and highway speed if all are present
speed = Math.max(MIN_SPEED, Math.min(Math.min(surfaceSpeed, speed), smoothSpeed));`

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -372,15 +372,19 @@ public void testSmoothness() {
...
+        assertEquals(10, getSpeedFromFlags(way), 0.01);
 
         way.clearTags();
         way.setTag("highway", "track");
-        way.setTag("tracktype", "grade5");
-        assertEquals(4, getSpeedFromFlags(way), 0.01);
+        way.setTag("tracktype", "grade4");
+        assertEquals(10, getSpeedFromFlags(way), 0.01);
+
+        // pick smallest
```
ratrun: A reader who hasn't checked the code will not understand this comment. Therefore, please change into
`// pick smallest of highway, tracktype, and applied smoothness speed`

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -372,15 +372,19 @@ public void testSmoothness() {
...
+        way.setTag("smoothness", "intermediate");
+        assertEquals(10, getSpeedFromFlags(way), 0.01);
```
ratrun: Please remove this empty line such that it becomes clearer that the comment above also applies here

core/src/test/java/com/graphhopper/routing/util/parsers/BikeTagParserTest.java
```diff
@@ -372,15 +372,19 @@ public void testSmoothness() {
...
 
         way.setTag("smoothness", "bad");
-        assertEquals(2, getSpeedFromFlags(way), 0.01);
+        assertEquals(8, getSpeedFromFlags(way), 0.01);
```
ratrun: Please remove this empty line such that it becomes clearer that the comment above also applies here

core/src/test/java/com/graphhopper/routing/util/parsers/MountainBikeTagParserTest.java
```diff
@@ -116,17 +116,16 @@ public void testSmoothness() {
         way.setTag("surface", "ground");
         assertEquals(14, getSpeedFromFlags(way), 0.01);
 
+        // pick smallest
```
ratrun: A reader who hasn't checked the code will not understand this comment. Therefore, please change into
`// pick smallest of highway, tracktype, and applied smoothness speed`

ratrun (approved):
Good catch, I only have minor suggestion for improving the comments.

PR #3307, opened 2026-03-10
bike: consider implicit access via highway=cycleway

karussell:
[Here](https://www.openstreetmap.org/way/89482973) is a real world example which we currently exclude.

Or maybe we are a bit too strict regarding access=agricultural? See e.g. [this way](https://www.openstreetmap.org/way/34540645) which we currently exclude but e.g. komoot allows.


karussell:
Fixed via #3313 instead.

PR #3309, opened 2026-03-11
Allow skipping instructions for via points (#3299)

### LINKED ISSUE #3299: allow to skip instructions for via points
[Some use cases](https://discuss.graphhopper.com/t/9912) require that the turn-by-turn instructions we create for via points needs to be suppressed.

modhtom:
Implements a new routing parameter `ignore_via_points_for_instructions` to suppress turn-by-turn instructions for via points, Closes #3299.

Added IGNORE_VIA_POINTS_FOR_INSTRUCTIONS to Parameters.Routing.
Updated createPathMerger to read this parameter from the request hints.
Modified PathMerger.doWork() to drop the FinishInstruction between path segments instead of converting it to a ViaInstruction when the flag is enabled.

Tested locally via the /route endpoint with a multi-leg route.
Without parameter: "Waypoint X" instructions are generated normally.
With `ignore_via_points_for_instructions=true`:The via point instructions are completely removed, and the route flows continuously without erroneous "Arrive at destination" markers mid-route.


VictorYanson:
For traceability, would you mind using a closing keyword in the PR description (e.g., 'closes #3299' instead of 'addressing')? Like this github can link the PR in the issue’s development section and show the PR icon si--pull-request-fill (1). Thanks!

karussell:
Thanks for your contribution.

However it will be fixed via #3310.

PR #3312, opened 2026-03-14
Define HK bike rules separately from UK rules

### LINKED ISSUE #3289: We prohibit pushing your bike in Hong Kong specifically
**Describe the bug**

When using the `bike` profile, currently GraphHopper cannot correctly route through bollards.

**To Reproduce**

GraphHopper Web; see https://graphhopper.com/maps/?point=22.410168%2C114.212777&point=22.410274%2C114.212895&profile=bike&layer=OpenStreetMap

Currently, GraphHopper routes this through an excessive detour totaling about 5.3km in mileage.

Image

**Expected behavior**

This should be a 10-meter path across a pedestrian crossing:

Image

**System Information**

GraphHopper Web

**Additional Context**

The nodes in question:
- [South Node](https://www.openstreetmap.org/node/7018781667)
- [North Node](https://www.openstreetmap.org/node/4741009747)

They both have the following:
- `barrier=bollard`
- `access=yes`

Per the OSM Wiki page ([link](https://wiki.openstreetmap.org/wiki/Tag:barrier=bollard?uselang=en)), by default bikes can pass through these bollards. For good measure, we did add `access=yes` to both nodes, but GraphHopper currently doesn't seem to understand any of this.

Vectorial1024:
This resolves #3289.

This is done by splitting off the Hong Kong bike rules from the UK bike rules. As Hong Kong currently does not have comprehensive signage and legislation around bikes, bikes are allowed to go through regular foot paths on the condition that the cyclists must dismount when going through such foot paths (e.g. dismount when using pedestrian crossings).

This essentially means cyclists may use regular footpaths as low-priority roads, and will still strongly prefer using existing cycling infrastructure.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
@@ -227,9 +227,13 @@ public class OSMRoadAccessParser<T extends Enum> implements TagParser {
                         || roadClass == RoadClass.BRIDLEWAY) return BikeRoadAccess.NO;
                 else if (roadClass == RoadClass.PEDESTRIAN) return BikeRoadAccess.YES;
             }
-            case GRC, GBR, HKG, IRL -> {
+            case GRC, GBR, IRL -> {
                 if (roadClass == RoadClass.PEDESTRIAN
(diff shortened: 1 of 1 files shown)


karussell:
Please see #3326. I think it will fix the issue too.

Issue #2980, opened 2024-04-21
highway=cycleway + access=no + bicycle=NULL is used for routing


zidel:
**Describe the bug**
Cycleways that are tagged as closed with access=no (and no conflicting bicycle=*) are still used for routing, e.g. https://www.openstreetmap.org/way/424854123

**Expected behavior**
Cycleways with access=no are not used for routing unless there is a bicycle=* or vehicle=* with a positive value

**System Information**
Tested via openstreetmap.org and graphhopper.com

**Screenshots & Logs**

https://graphhopper.com/maps/?point=61.459975%2C5.814842&point=61.459033%2C5.827119&profile=bike&layer=Omniscale
https://graphhopper.com/maps/?point=59.911557%2C10.627221&point=59.909744%2C10.625891&profile=bike&layer=Omniscale

bt90:
Access parsing aside, the road isn't tagged properly IMHO. This should be `highway=construction` + `construction=cycleway` as explained in the OSM wiki:

https://wiki.openstreetmap.org/wiki/Tag:highway%3Dconstruction#How_to_map_the_start_of_a_road_repair_or_reconstruction

ratrun:
Currently we are treating `access=no` and `vehicle=no` the same way with allowing this segment for bicycles. But according to https://wiki.openstreetmap.org/wiki/Tag:access%3Dno the intention of `access=no` is stronger. While on a segment tagged as `vehicle=no` it is still possible and legal to dismount and push for `access=no` this is not the case. So I think that this problem report is valid and we should change it. I'm going to prepare a PR if there aren't any objections.

ratrun:
>  And should we remove the special case way.hasTag("highway", "cycleway")) 
I extended the check there in my first version of the PR, see #2981 

> And then consider the "dismount" for vehicle=no in OSMGetOffBikeParser and BikeCommonAverageSpeedParser?
Yes that is a good idea. I'll try to extend the PR.

bt90:
> And then consider the \"dismount\" for vehicle=no

Broadly speaking, we should do this for all cases where foot access is allowed but biking isn't.

ratrun:
This issue can be closed, the provided examples are working.

PR #3313, opened 2026-03-16
bike: fix problem with agricultural

karussell:
See [this inaccessible OSM way](https://www.openstreetmap.org/way/89482973) or [this route](https://graphhopper.com/maps/?point=51.19367%2C13.388138&point=51.195114%2C13.390896&point=51.19464%2C13.395275&profile=bike&layer=Omniscale).

Now access restrictions for highway=cycleway have a bit of a history: #870 and #2980 with PR #2981.
From that I can read that highway=cycleway is nothing special regarding access/vehicle=no and so we then assume dismount for all highway=*.
But now for access=agricultural it also restricts access which it should not because `highway=cycleway` should be something special as bicycle=yes by default (even if not tagged) and so access should be allowed. Or WDYT @ratrun ?


karussell:
Thank you for the review!

PR #3317, opened 2026-03-21
Handle partial downloads

otbutz:
This is a subset of the changes proposed in #3143:

* handle partial downloads
* use a proper User-Agent e.g `graphhopper/12.0`


karussell:
Thanks!

PR #3318, opened 2026-03-21
Move heading validation into GHRequest

cezarysanecki:
`Router` was reaching into `GHRequest` internals to iterate over headings and validate each value. This is a case of Feature Envy — the validation logic belongs in `GHRequest` itself.

This PR moves the full heading validation (azimuth range check and count vs points mismatch) into `GHRequest.checkHeadings()`. `Router` now delegates with a single call.

`isAzimuthValue(double)` is preserved as `@Deprecated` for backwards compatibility.

Tests have been added to `GHRequestTest` to cover the validation logic directly on `GHRequest`, which was previously untested at this level.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -12,7 +12,7 @@ Here is an overview:
  * b3nn0, Android improvements
  * baumboi, path detail and landmark improvements
  * baybatu, improved error for EncodingManager
- * cezarysanecki, refactoring to decouple core from web-api
+ * cezarysanecki, refactoring for better maintainability through tests and clearer responsibilities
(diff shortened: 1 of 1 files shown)


karussell:
Thanks for your contribution. A refactoring-only contribution makes only sense if you have already some experience with the code base and we do not want to review refactorings from anyone just because they asked Claude or ChatGPT to make some code cleaner. If you want to contribute a small improvement like a bug fix is appreciated - read more about it [here](https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md). Often times the best contributions are the ones that improve the usage of GraphHopper for your own use case.

PR #3320, opened 2026-03-21
Introduce graphhopper-api module to decouple core from web-api

cezarysanecki:
## Problem

`graphhopper-core` currently depends on `graphhopper-web-api`. This is an inverted dependency — the name `web-api` implies a higher-level concern, yet the lower-level routing engine depends on it.

The practical consequence: anyone embedding `graphhopper-core` (e.g. in a microservice that only needs routing, without an HTTP layer) is forced to pull in web-related dependencies (Jackson serializers, GPX support, HTTP-specific exceptions) that they do not need.

## Approach

This PR introduces a new `graphhopper-api` module (folder: `api`) as a home for shared domain objects that are currently in `graphhopper-web-api` but are needed by `graphhopper-core`.

The target dependency structure:
```
graphhopper-api     <- shared domain: DTOs, value objects, configuration
      ^
      |
graphhopper-core    <- routing engine, depends on graphhopper-api
      ^
      |
graphhopper-web-api <- Jackson serializers, HTTP concerns, depends on graphhopper-api
```

## This PR

- Adds the empty `graphhopper-api` module registered in the parent pom
- Moves `MinMax` as the first class — a plain value object used by the custom weighting engine in core, with no reason to live in web-api
- `web-api` gains a dependency on `graphhopper-api` to maintain full backwards compatibility (same package `com.graphhopper.json`, no import changes for existing users)

## Benefit

Embedded users of `graphhopper-core` will eventually be able to use the routing engine without pulling in web-related dependencies. This PR is the first small step in that direction.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>graphhopper-api</artifactId>
+    <packaging>jar</packaging>
+    <version>12.0-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


cezarysanecki:
Withdrawing this PR — the direction turned out to be wrong. The graphhopper-api module would duplicate what web-api is meant to achieve.

PR #3321, opened 2026-03-21
Extract CustomAreasProvider to decouple area loading from GraphHopper

cezarysanecki:
## Problem

`GraphHopper.readCustomAreas()` was responsible for reading GeoJSON files from disk, deserializing them via Jackson and converting them to `CustomArea` objects. This mixed I/O and deserialization concerns directly into the routing engine.

Two concrete issues:
- **Untestable**: there was no way to test the custom areas loading logic in isolation — it required a real directory on disk wired through `GraphHopper`
- **Inflexible**: embedded users had no way to provide custom areas from any other source (database, classpath, in-memory) without subclassing `GraphHopper`

## Solution

Introduce a `CustomAreasProvider` interface:

```java
public interface CustomAreasProvider {
    List<CustomArea> loadAreas();
}
```

`DefaultCustomAreasProvider` encapsulates the original file-based logic and is used automatically when `custom_areas.directory` is configured — so **existing behaviour is fully preserved**.

`GraphHopper` gains a `setCustomAreasProvider(CustomAreasProvider)` setter, allowing embedded users to supply areas from any source without touching the engine internals.

## Benefits

- **Testable**: `DefaultCustomAreasProvider` can now be tested in isolation (tests included)
- **Extensible**: embedded users can provide areas from any source
- **Foundation**: separates I/O from the routing engine, a step towards reducing `core`'s responsibility for infrastructure concerns

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -14,6 +14,7 @@ Here is an overview:
  * baybatu, improved error for EncodingManager
  * boldtrn, one of the core developers with motorcycle knowledge :)
  * bt90, fixes like #2786
+ * cezarysanecki, refactoring for better maintainability through tests and clearer responsibilities
  * cgarreau, increase of routing success rate via subnetwork cleanup
  * chollemans, fixes like #1482
  * ChristianSeitzer, motorcycle improvements


karussell:
Thanks for your contribution. Unfortunately I do not see a reason for this change.

PR #3322, opened 2026-03-22
Extract CustomModelResolver from GraphHopper

cezarysanecki:
## Motivation

`GraphHopper.resolveCustomModelFiles` is a static method that mixes I/O, JSON deserialization, and merging logic directly inside the routing engine class. This violates the single responsibility principle — `GraphHopper` should orchestrate routing, not handle file loading and model resolution.

Additionally, the method had no dedicated unit tests, making it hard to verify individual behaviours (inline model, files from jar, files from external directory, merging, error cases) without running a full integration test.

## Changes

- Introduced `CustomModelResolver` — a dedicated class responsible for resolving `CustomModel` instances for profiles that use `custom` weighting. Supports:
  - inline `custom_model` from config (handles both raw JSON trees and object instances via `ObjectMapper.convertValue`)
  - built-in models from jar resources (`/com/graphhopper/custom_models/`)
  - external model files from a configurable directory
  - merging multiple files into a single model
  - injecting global custom areas into all resolved models
- Added `Profile.hasCustomModel()` to avoid leaking hints internals into `CustomModelResolver`
- Marked `GraphHopper.resolveCustomModelFiles` as `@Deprecated` with a Javadoc pointer to `CustomModelResolver#resolveAll`; the deprecated method now delegates to the new class
- `GraphHopper.init()` uses `CustomModelResolver` directly
- Added unit tests covering all resolution paths and error cases

### DIFF
--- a/core/src/main/java/com/graphhopper/CustomModelResolver.java
+++ b/core/src/main/java/com/graphhopper/CustomModelResolver.java
@@ -42,19 +42,19 @@ import static com.graphhopper.util.Helper.readJSONFileWithoutComments;
  * Resolves {@link CustomModel} instances for profiles that use {@link CustomWeighting}.
  * Supports loading custom models inline from config or from files (built-in jar resources or external directory).
  */
-class CustomModelResolver {
+public class CustomModelResolver {
 
     private final String customModelFolder;
     private final JsonFeatureCollection globalAreas;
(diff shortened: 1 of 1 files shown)


karussell:
Thanks for your contribution. Unfortunately I do not see a reason for this change.

PR #3323, opened 2026-03-25
Add RAMInt1SegmentDataAccess

easbar:
It's like RAMIntDataAccess, but it is backed by int[], rather than int[][]. So it does not offer cheap resizing and is limited to 2B ints, but the reading speed can be slightly faster, since we do not need the outer index access.

Even though there are no real segments the implementation still uses the segment size:

* to keep file compatibility with the other DataAccess implementations (ensureCapacity and trimTo make sure the array length is a multiple of the segment size)
* to fit into the DataAccessTest hierarchy 
* as buffer size in loadExisting() and flush(), just like RAMIntDataAccess

### DIFF
--- a/core/src/main/java/com/graphhopper/storage/DAType.java
+++ b/core/src/main/java/com/graphhopper/storage/DAType.java
@@ -44,6 +44,16 @@ public class DAType {
      */
     public static final DAType RAM_INT_STORE = new DAType(MemRef.HEAP, true, true, true);
     /**
+     * Like RAM_INT_STORE, but backed by a single contiguous int[] for maximum read speed.
+     * Not a good fit if the array needs to be resized frequently. Limited to Integer.MAX_VALUE ints
+     * No support for short,byte and bytes.
+     */
+    public static final DAType RAM_INT_1SEG = new DAType(MemRef.HEAP, false, true, true, true);
+    /**
+     * See RAM_INT_1SEG
+     */
+    public static final DAType RAM_INT_1SEG_STORE = new DAType(MemRef.HEAP, true, true, true, true);
+    /**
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/graphhopper/storage/DAType.java
```diff
@@ -43,6 +43,16 @@ public class DAType {
...
+    /**
+     * Like RAM_INT, but backed by a single contiguous int[] for maximum read speed.
+     * Not a good fit if the array needs to be resized frequently. Limited to Integer.MAX_VALUE ints
+     * No support for short,byte and bytes.
+     */
+    public static final DAType RAM_INT_1SEG = new DAType(MemRef.HEAP, false, true, true, true);
+    /**
+     * See RAM_INT_1SEG
+     */
+    public static final DAType RAM_INT_1SEG_STORE = new DAType(MemRef.HEAP, true, true, true, true);
```
karussell, line 55: will have to refactor this into an enum. but will do this after we merge this here.

core/src/main/java/com/graphhopper/storage/RAMInt1SegmentDataAccess.java
```diff
@@ -0,0 +1,243 @@
...
+
+    @Override
+    public long getCapacity() {
+        return (long) data.length * 4;
+    }
+
+    @Override
+    public int getSegments() {
+        return data.length / (segmentSizeInBytes / 4);
+    }
```
karussell, line 225: this is a bit inconsistent as I would have expected 1 here (hence the 1Segment name). But we can keep it as it is to be more compatible with the other DataAccess classes

karussell (approved):
LGTM

PR #3324, opened 2026-03-25
fix: prevent conditional tags from overriding implicit pedestrian roa…

### LINKED ISSUE #3308: GraphHopper routes cars through pedestrian roads
**Describe the bug**
highway=pedestrian roads are intended for pedestrians only; exceptions for vehicles should be tagged explicitly.

GraphHopper currently routes vehicles through pedestrian roads if motor_vehicle:conditional (e.g., delivery) is defined. I would expect motor_vehicle=no to remain the default for this road type, even when a conditional access tag is present, unless the routing profile explicitly validates that the condition is met.

Example:
https://www.openstreetmap.org/way/124259425
highway = pedestrian
motor_vehicle:conditional = delivery @ (Mo-Fr 06:00-11:00, 19:00-21:00; Sa 06:00-09:30; PH off)
bicycle = yes

**To Reproduce**
https://graphhopper.com/maps/?point=52.374555%2C9.733054&point=52.375861%2C9.737142&profile=car&layer=Omniscale

**Expected Behavior**
The routing engine should prioritize a detour around the pedestrian area.
https://www.openstreetmap.org/directions?engine=fossgis_osrm_car&route=52.374807%2C9.734209%3B52.375789%2C9.737197#map=16/52.37804/9.73678

pgrigolli:
Fixes #3308 
highway=pedestrian roads were being routed for cars when a motor_vehicle:conditional tag was present, even without any explicit access tag. For example, a road tagged with:
```
highway = pedestrian
motor_vehicle:conditional = delivery @ (Mo-Fr 06:00-11:00, 19:00-21:00; Sa 06:00-09:30; PH off)
bicycle = yes
```
was being treated as accessible for cars, causing the router to send vehicles through pedestrian areas instead of calculating a detour.
### Root cause
The if block responsible for blocking cars on pedestrian roads calls hasPermissiveTemporalRestriction as its third condition. This method was returning true whenever a conditional tag contained a value present in allowedValues (such as delivery) and a time range — regardless of whether the restriction on the road was explicit (access=no) or implicit (derived from highway=pedestrian). This caused the CAN_SKIP return to be skipped, leaving the road accessible for cars.
### Fix
Added a private method hasPermissiveTemporalRestrictionForPedestrian in CarAccessParser that adds a guard before calling hasPermissiveTemporalRestriction — only considering conditional tags if an explicit restriction (access=no or motor_vehicle=no) is present on the way. The original hasPermissiveTemporalRestriction method was left untouched to avoid breaking other parsers that depend on it (bike, temporal access).

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -75,6 +75,7 @@ Here is an overview:
  * otbutz, added multiple EncodedValues
  * PabloaRuiz, pt_BR 1i8n improvements
  * pantsleftinwash, speed parsing improvements
+ * pgrigolli, pedestrian roads conditional accesses fix
  * PGWelch, shapefile reader #874
  * rafaelstelles, fix deserializer web-api
  * rajanski, script to do routing via PostGIS
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/CarAccessParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/CarAccessParser.java
@@ -27,6 +27,7 @@ import com.graphhopper.util.PMap;
 import java.util.*;
 
 import static com.graphhopper.routing.util.parsers.OSMTemporalAccessParser.hasPermissiveTemporalRestriction;
+import static com.graphhopper.routing.util.parsers.OSMTemporalAccessParser.hasPermissiveTemporalRestrictionForPedestrian;
 
 public class CarAccessParser extends AbstractAccessParser implements TagParser {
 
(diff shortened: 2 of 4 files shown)


pgrigolli:
@karussell could you please check if this is ok?

michaz:
Hi! Thanks, but I think this isn't it. Technically speaking, your second test already succeeds before your patch (we are *not* in fact opening anything for "delivery" exemptions here, despite what it looks like in the ticket), and your second test only tests one side (that we aren't opening "part-time" pedestrian zones anymore), but it doesn't show what we *do* open.

Less technically: What's the difference between a pedestrian zone with "motor_vehicle=no" and one without? Shouldn't they be treated the same when a conditional like this hits them?

The question is more like should a pedestrian zone that is only open for cars for parts of the day be open or closed if we can only pick one.

PR #3325, opened 2026-03-27
remove npm from build

karussell:
This works and does not require npm (and updating it). Not sure yet what the downsides are.


easbar:
Nice one. And honestly I don't remember why we used npm pack rather than the simple http get.

PR #3326, opened 2026-03-30
remove 'no' and 'dismout' from bike country handler

### LINKED ISSUE #3289: We prohibit pushing your bike in Hong Kong specifically

karussell:
Fixes #3289. Replaces #3312. /cc @Vectorial1024 @ratrun.

In master branch we return DISMOUNT or NO but in practice in OpenStreetMap the tag `bicycle=no` does not mean that bikes are not allowed, but rather it means `bicycle=dismount` and so it is accessible for bike. As this is the default value for `footway` and `pedestrian` highways I removed this.

For `trunk` this is different as foot is then excluded too. (The same for `bridleway`.)

### DIFF
--- a/core/src/test/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParserTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParserTest.java
@@ -141,7 +141,12 @@ class OSMRoadAccessParserTest {
     @Test
     void testFoot() {
         assertEquals(FootRoadAccess.YES, OSMRoadAccessParser.FOOT_HANDLER.getAccess(createReaderWay("cycleway"), Country.BEL));
-
+    }
+ 
+    @Test
+    void germany() {
+        assertEquals(RoadAccess.DESTINATION, OSMRoadAccessParser.CAR_HANDLER.getAccess(createReaderWay("track"), Country.DEU));
+        assertEquals(RoadAccess.YES, OSMRoadAccessParser.CAR_HANDLER.getAccess(createReaderWay("primary"), Country.DEU));
     }
 
     @Test


Vectorial1024:
Concern:

https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Dno

> By itself, it does not indicate whether pushing a bicycle is allowed or prohibited.

I do not know GraphHopper well enough, but will there be further processing to determine whether bike-pushing is allowed or not? Some places may have explicit `bicycle=no` to indicate "really no bikes are allowed, not even pushing".

karussell:
> but will there be further processing to determine whether bike-pushing

Yes, see the BikeCommonAccessParser (->bike_access). I need to merge both soonish (xy_access and xy_road_access).

> Some places may have explicit bicycle=no to indicate "really no bikes are allowed, not even pushing".

Yes. And one would expect that this is the default, but unfortunately the opposite is true and so bicycle=no and bicycle=dismount basically mean the same. The tag bicycle=no almost always means "pushing allowed except when foot is not allowed". But of course there are also cases where foot is allowed and bike cannot be even pushed (or in a ferry bike is not allowed). For that case the wiki says one should use bicycle:pushing but this tag has not even 100 usages and can be ignored on the global scale...

michaz:
Just for clarity: We're not returning `DISMOUNT` at all, not previously and not now, `DISMOUNT` is unused. We just stop returning `NO` because `NO` should mean `DISMOUNT` and `DISMOUNT` isn't implemented yet but `NO` is wrong.

Vectorial1024:
This was closed faster than I expected; I am not commenting on the coding part of this PR.

-----------

@karussell I can see the wiki page for `bicycle=no` says that `bicycle:pushing=*` is a very new tag that was established in late 2025: https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Dno

> For clarity, in late 2025, this tagging scheme is recommended:
> 
> - Pushing prohibited: `bicycle=no` + `bicycle:pushing=no`
> - [...]

This is not a neglected ancient tag, and could qualify for future work. Future people can look at this paragraph, and open a new issue when the need arises.

PR #3327, opened 2026-03-30
Fix bike priority: cycleway tags no longer downgrade preferred highways

### LINKED ISSUE #3219: highway=residential without bicycle infrastructure preferred over one with infrastructure

michaz:
## Summary
Fixes #3219.

- **Remove residential/unclassified/service from bike `preferHighwayTags`**, making them UNCHANGED (1.0) baseline instead of PREFER (1.2). The `maxspeed<=30` rule still prefers calm streets with explicit low speed limits.
- **Bump cycleway priority values**: `cycleway=track` from PREFER (1.2) to VERY_NICE (1.3), `cycleway=lane` etc. from SLIGHT_PREFER (1.1) to PREFER (1.2). `cycleway=track` now matches a standalone `highway=cycleway`, and `cycleway=lane` no longer downgrades roads that were previously preferred.

Before and after:

| Scenario | Before | After |
|---|---|---|
| residential alone | 1.2 | 1.0 |
| residential + `cycleway=lane` | 1.1 (!) | 1.2 |
| residential + `cycleway=track` | 1.2 | 1.3 |
| secondary alone | 0.8 | 0.8 |
| secondary + `cycleway=lane` | 1.1 | 1.2 |
| secondary + `cycleway=track` | 1.2 | 1.3 |
| standalone `highway=cycleway` | 1.3 | 1.3 |

Note: the cycleway bump is in `BikeCommonPriorityParser` and affects all bike profiles (bike, mtb, racingbike). The `preferHighwayTags` removal is bike-specific only.

## Test plan
- [x] All existing bike parser tests updated and passing (BikeTagParserTest, MountainBikeTagParserTest, RacingBikeTagParserTest, BikeCustomModelTest)

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
@@ -135,9 +135,9 @@ public abstract class BikeCommonPriorityParser implements TagParser {
 
         Set<String> cyclewayValues = Stream.of("cycleway", "cycleway:left", "cycleway:both", "cycleway:right").map(key -> way.getTag(key, "")).collect(Collectors.toSet());
         if (cyclewayValues.contains("track")) {
-            weightToPrioMap.put(100d, PREFER);
+            weightToPrioMap.put(100d, VERY_NICE);
         } else if (Stream.of("lane", "opposite_track", "shared_lane", "share_busway", "shoulder").anyMatch(cyclewayValues::contains)) {
-            weightToPrioMap.put(100d, SLIGHT_PREFER);
+            weightToPrioMap.put(100d, PREFER);
         } else if (pushingSectionsHighways.contains(highway) || "parking_aisle".equals(way.getTag("service"))) {
             PriorityCode pushingSectionPrio = SLIGHT_AVOID;
             if (way.hasTag("highway", "steps"))
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikePriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikePriorityParser.java
@@ -17,10 +17,6 @@ public class BikePriorityParser extends BikeCommonPriorityParser {
 
         addPushingSection("path");
 
-        preferHighwayTags.add("service");
-        preferHighwayTags.add("residential");
-        preferHighwayTags.add("unclassified");
-
         setSpecificClassBicycle("touring");
     }
 }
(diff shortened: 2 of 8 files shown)


karussell:
Thanks! This fixes two inconsistencies and is a tiny change, which I like a lot :) (As I understand it: the cycleway=track should get the same boost regardless of the highway and the original issue #3219)

However I would prefer that the lane boost stays stays at 1.1. IMO especially important for secondary roads. This should be easily possible without the boost change regarding `cycleway=lane` (?)

(Also related to #3251)

michaz:
Yes I agree, this was a close call anyway. Either should be fine, and overall I think our parameters are probably too big rather than too small.

Note, I didn't actually look at many routes yet. 😅 Will try that change and then do that.

michaz:
Sorry, have to take back that promise. This change solves #3219. (Route quality to be determined.)

But if I would leave `cycleway=lane` at 1.1, it would not solve it because roads with `maxspeed=30` still get 1.2 which is more than 1.1. 

This change only works if every `cycleway=*` boost is `>=` every `highway=*` boost, and the thing where we look at `maxspeed` is effectively a `highway=*` boost.

But setting the `maxspeed=30` boost *and* the `cycleway=lane` boost down to 1.1 would work.

Or simply deleting one of the two and doing without it.

But I must admit that this is as far as I can help here right now. I found out (looking at more code and tickets) that I would want to restart several things from first principles, but I don't have the capacity to execute that and it's also maybe a bit too contested for me to succeed even if I had it.

michaz:
Just one more comment dump, isn't it a problem that before the custom model change you were *picking* the bike network ratings as priority, but now you are *multiplying* by them instead?

And also I found a thing where we *raise* the priority *because* we are slowed down by `maxspeed`. You can probably guess what I'm going to say -- I would really really advise against such things, and if I had a say I would just delete it.

Because if we believe bikes are routed by (preference-weighted) speed, then we shouldn't "compensate" for lost speed. (And if we would, we would compensate for the *exact* lost speed, not bump it by a category.)

If we don't believe bikes are routed by speed, then we simply shouldn't route them by speed but by (preference-weighted) **distance**, or by a *perceived* speed that isn't influenced by speed restrictions. 

If we *do* believe bikes are routed by speed, but don't believe their speed is determined much by speed restrictions, then we simply shouldn't cap their speed at speed restrictions.

If we want to route by distance but can't route by distance because we are using custom_model and custom_model can't do that, then we shouldn't use custom_model but something which can do that or make custom_model able to do that.

ratrun:
This PR  can be closed

PR #3330, opened 2026-03-31
Do not call graph.wrapWeighting excessively in MapMatching

easbar:
#3297 introduced a regression for MapMatching, slowing it down extremely, especially when map-matching many observations. The problem is the normally relatively cheap call of `queryGraph.wrapWeighting()` that happens in the most inner loop of the map matching algorithm (Router#calcOnePath). 

As quick fix I added a cache in QueryGraph here: https://github.com/graphhopper/graphhopper/commit/64cbdbfa0647fbaa454508a81c1e0a4ed885ccab, but this PR removes the wrapWeighting calls explicitly.

Basically only one such call is needed for each match, since the QueryGraph and the weighting never change. We just need a way to set the queryGraph and the wrapped weighting once for the router. Therefore I added a setQueryGraph method to the MapMatching#Router interface.

But we also need to remove the wrapWeighting call in LMApproximator#findClosestRealNode. Since the LMApproximator is re-created in calcOnePath (and it's reverse is also recreated in AStarBidirection via the BalancedWeightApproximator constructor) we need to pass the wrapped weighting to LMApproximator. In turn we don't really need to pass the unwrapped lmWeighting to LMApproximator as we did so far.

See a few more comments inline.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/lm/LMApproximator.java
+++ b/core/src/main/java/com/graphhopper/routing/lm/LMApproximator.java
@@ -18,8 +18,10 @@
 package com.graphhopper.routing.lm;
 
 import com.graphhopper.routing.Dijkstra;
+import com.graphhopper.routing.querygraph.QueryGraph;
 import com.graphhopper.routing.util.TraversalMode;
 import com.graphhopper.routing.weighting.BeelineWeightApproximator;
+import com.graphhopper.routing.weighting.QueryGraphWeighting;
 import com.graphhopper.routing.weighting.WeightApproximator;
 import com.graphhopper.routing.weighting.Weighting;
 import com.graphhopper.storage.Graph;
@@ -63,6 +65,8 @@ public class LMApproximator implements WeightApproximator {
      */
     public LMApproximator(Graph graph, Weighting lmWeighting, Weighting routingWeighting, LandmarkStorage lms, int activeCount,
                           boolean reverse) {
+        if (graph instanceof QueryGraph && (!(lmWeighting instanceof QueryGraphWeighting) || (!(routingWeighting instanceof QueryGraphWeighting))))
+            throw new IllegalStateException("Weighting must use QueryGraphWeighting");
         this.reverse = reverse;
         this.lms = lms;
         this.factor = lms.getFactor();


core/src/main/java/com/graphhopper/routing/lm/LMApproximator.java
```diff
@@ -54,18 +54,18 @@ public class LMApproximator implements WeightApproximator {
...
-    }
-
-    public LMApproximator(Graph graph, Weighting lmWeighting, Weighting routingWeighting, int maxBaseNodes, LandmarkStorage lms, int activeCount,
-                          double factor, boolean reverse) {
+    public LMApproximator(Graph graph, Weighting lmWeighting, Weighting routingWeighting, LandmarkStorage lms, int activeCount,
+                          boolean reverse) {
         this.reverse = reverse;
         this.lms = lms;
-        this.factor = factor;
+        this.factor = lms.getFactor();
```
easbar, line 72: Not really related to this PR, but the factor argument seemed redundant since we already have lms here.

core/src/main/java/com/graphhopper/routing/lm/LMApproximator.java
```diff
@@ -78,9 +78,9 @@ public LMApproximator(Graph graph, Weighting lmWeighting, Weighting routingWeigh
         this.graph = graph;
         this.lmWeighting = lmWeighting;
         this.routingWeighting = routingWeighting;
-        this.fallBackApproximation = new BeelineWeightApproximator(graph.getNodeAccess(), lmWeighting);
+        this.fallBackApproximation = new BeelineWeightApproximator(graph.getNodeAccess(), routingWeighting);
         this.beelineApproximation = new BeelineWeightApproximator(graph.getNodeAccess(), routingWeighting);
-        this.maxBaseNodes = maxBaseNodes;
+        this.maxBaseNodes = lms.getBaseNodes();
```
easbar, line 87: Not really related to this PR, but the maxBaseNodes argument seemed redundant since we already have lms here.

core/src/main/java/com/graphhopper/routing/lm/LMApproximator.java
```diff
@@ -78,9 +78,9 @@ public LMApproximator(Graph graph, Weighting lmWeighting, Weighting routingWeigh
         this.graph = graph;
         this.lmWeighting = lmWeighting;
         this.routingWeighting = routingWeighting;
-        this.fallBackApproximation = new BeelineWeightApproximator(graph.getNodeAccess(), lmWeighting);
+        this.fallBackApproximation = new BeelineWeightApproximator(graph.getNodeAccess(), routingWeighting);
```
easbar, line 85: I think this was a bug. The fall back approximator should use the routingWeighting, not lmWeighting.

core/src/main/java/com/graphhopper/routing/lm/LMApproximator.java
```diff
@@ -54,18 +54,18 @@ public class LMApproximator implements WeightApproximator {
...
+     *                         the same we used for the LM preparation. All edge weights must be larger
+     *                         or equal compared to those used for the preparation.
      */
-    public static LMApproximator forLandmarks(Graph g, Weighting weighting, LandmarkStorage lms, int activeLM) {
-        return new LMApproximator(g, lms.getWeighting(), weighting, lms.getBaseNodes(), lms, activeLM, lms.getFactor(), false);
-    }
-
-    public LMApproximator(Graph graph, Weighting lmWeighting, Weighting routingWeighting, int maxBaseNodes, LandmarkStorage lms, int activeCount,
-                          double factor, boolean reverse) {
+    public LMApproximator(Graph graph, Weighting lmWeighting, Weighting routingWeighting, LandmarkStorage lms, int activeCount,
```
easbar, line 66: The lmWeighting argument changed its meaning here: It no needs to be wrapped with the graph if graph is the QueryGraph. Otherwise we'll get an error from the Dijkstra constructor in findClosestRealNode (the only place where it is needed at all).

PR #3331, opened 2026-03-31
Support directional toll tags (toll:forward, toll:backward)

### LINKED ISSUE #3272: One-sided toll roads are not recognized
**Describe the bug**
GraphHopper currently recognizes and highlights toll roads. However, it fails to recognize "one-sided tolls" (elaborated below).

**To Reproduce**

See https://graphhopper.com/maps/?point=22.310608%2C113.991405&point=22.308657%2C114.012428&profile=car&layer=OpenStreetMap

It currently doesn't recognize any toll roads. (The small red toll section is likely a data error, and will be separately investigated.)

Image

**Expected behavior**

The section circled in red should be recognized as a toll road.

Image

**System Information**

Using GraphHopper Web Maps

**Additional Context**

Toll roads may have the following "sides":
- Zero-sided (`toll=yes+oneway=yes`, already handled)
- One-sided (`toll:forward=yes` or `toll:backward=yes`; **👈 we are here**)
- Two-sided (normal `toll=yes`, already handled)

The road sections in question have `toll:backward=yes`.

Notice this road is a tolled tunnel connecting the neighbourhood on the east side to the railway/highway combo on the west side. However, while there is a nominal "toll per passage", the irl implementation is that, vehicles pay double when entering the neighbourhood, but payment is waived during their way out. This therefore is "one-sided toll".

The idea is that, GraphHopper should highlight toll sections with respect to how irl works. If a routing request brings people away from the neighbourhood, it shouldn't highlight any toll sections, but conversely when entering the neighbourhood, it should highlight the toll sections.

michaz:
## Summary

Fixes #3272: One-sided toll roads (e.g. `toll:backward=yes`) were not recognized because toll was stored as a non-directional edge property.

### Changes

- **`Toll.create()`**: Changed to `storeTwoDirections=true` so toll is stored per direction on each edge (like `Cycleway` and `Sidewalk`).
- **`OSMTollParser`**: Now parses `toll:forward` and `toll:backward` OSM tags. Directional tags override the base `toll` value for their respective direction — including explicit `no` overriding a base `toll=yes`. Supports all `Toll` enum values (`yes`→ALL, `no`→NO, `hgv`→HGV).

### Data analysis: real-world usage of `toll:forward` / `toll:backward`

Checked all ~305 ways worldwide that use these tags (via Overpass/Taginfo). Zero ways have both `toll:forward=yes` AND `toll:backward=yes` — nobody is redundantly splitting a symmetric toll into two tags. All cases are genuine one-sided tolls.

| Pattern | Count |
|---------|-------|
| `toll:forward=yes` + `toll:backward=no` | 134 |
| `toll:backward=yes` + `toll:forward=no` | 41 |
| `toll:forward=yes` alone | 44 |
| `toll:backward=yes` alone | 86 |

Real-world scenarios:
1. **US/Canadian toll bridges/tunnels** (~130 ways) — toll collected in one direction only (Confederation Bridge, Lincoln Tunnel, Bayonne Bridge, etc.)
2. **Norwegian toll rings** (~84 ways) — "bomringer" where you pay entering a city but not leaving
3. **Greek motorway ramps** (~22 ways) — directional toll status on interchange ramps
4. **Hong Kong tunnels** (~9 ways) — like the Discovery Bay Tunnel from the original issue

About ~80 ways carry `toll=yes` alongside `toll:forward=yes` + `toll:backward=no`. This is a tagging conflict (`toll=yes` implies both directions), but the directional tags are clearly more specific. Our parser now handles this correctly: `toll:backward=no` overrides the `toll=yes` base for the backward direction.

## Test plan
- [x] `testDirectionalToll`: forward-only, backward-only, both-directions, explicit `no` overriding base `toll=yes`, `toll:forward=hgv`
- [x] Extended existing symmetric toll tests with reverse-direction assertions
- [x] All `OSMTollParserTest` tests pass

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/ev/Toll.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/Toll.java
@@ -29,7 +29,7 @@ public enum Toll {
     public static final String KEY = "toll";
 
     public static EnumEncodedValue<Toll> create() {
-        return new EnumEncodedValue<>(KEY, Toll.class);
+        return new EnumEncodedValue<>(KEY, Toll.class, true);
     }
 
     @Override
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTollParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTollParser.java
@@ -20,6 +20,7 @@ package com.graphhopper.routing.util.parsers;
 import com.graphhopper.reader.ReaderWay;
 import com.graphhopper.routing.ev.*;
 import com.graphhopper.storage.IntsRef;
+import com.graphhopper.util.Helper;
 
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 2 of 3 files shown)


michaz:
yay

PR #3332, opened 2026-03-31
replace PriorityQueue with cache-friendly binary heap

karussell:
I gave the AI the task to find performance improvements and this was one of the results which gives a >10% query speed improvement but not for CH routing queries.

This should use slightly more memory as it is one array more.

(Here is [a similar change](https://github.com/graphhopper/graphhopper/tree/try_pq_replace_4ary)).

### DIFF
--- a/core/src/main/java/com/graphhopper/coll/BinaryHeapWithDuplicates.java
+++ b/core/src/main/java/com/graphhopper/coll/BinaryHeapWithDuplicates.java
@@ -1,0 +1,83 @@
+package com.graphhopper.coll;
+
+import java.util.Arrays;
+
+/**
+ * A binary min-heap that stores items with parallel double[] weights for cache-friendly comparisons.
+ * Unlike java.util.PriorityQueue which dereferences Object pointers during sift operations,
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/graphhopper/coll/BinaryHeapWithDuplicates.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Arrays;
+
+/**
+ * A binary min-heap that stores items with parallel double[] weights for cache-friendly comparisons.
+ * Unlike java.util.PriorityQueue which dereferences Object pointers during sift operations,
+ * this heap compares weights from a contiguous double[] array, dramatically reducing cache misses.
+ */
+public class BinaryHeapWithDuplicates<E> {
+    private double[] weights;
+    private Object[] items;
```
OlafFlebbeBosch, line 12: Shouldn't it be `E` rather Object

OlafFlebbeBosch: Ah I see .. We need an Array of E

PR #3333, opened 2026-04-02
Add OSM implied access defaults to ModeAccessParser

michaz:
## Summary
- ModeAccessParser now considers implied access defaults from OSM highway types (e.g. `footway` → `motor_vehicle=no`, `busway` → `bus=designated`) and barrier types (e.g. `bus_trap` → `bus=yes`), integrated into the restriction key hierarchy so explicit tags always take priority
- Simplifies `bus.json` custom model by removing the `road_class` filter — access control is now handled correctly in the parser
- Expands `BusCustomModelTest` with realistic tag combinations: normal roads, non-motor-vehicle highways, explicit overrides (`bus=yes` on footway, `psv=designated` on pedestrian zone), barriers (bollard, bus_trap, sump_buster)

## Test plan
- [x] `BusCustomModelTest` (13 tests) — covers normal roads, blocked highways, busway, explicit overrides, barriers
- [x] `ModeAccessParserTest` (14 tests) — existing tests still pass

### DIFF
--- a/core/src/main/resources/com/graphhopper/custom_models/bus.json
+++ b/core/src/main/resources/com/graphhopper/custom_models/bus.json
@@ -10,6 +10,7 @@
   "distance_influence": 90,
   "priority": [
     { "if": "max_weight < 5 || max_width < 3 || max_height < 4", "multiply_by": "0" },
+    { "if": "road_class == TRACK", "multiply_by": "0" },
     { "if": "bus_access",
       "multiply_by": "1"
     },
--- a/core/src/test/java/com/graphhopper/routing/util/parsers/BusCustomModelTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/parsers/BusCustomModelTest.java
@@ -80,6 +80,13 @@ public class BusCustomModelTest {
     }
 
     @Test
+    public void testTrack() {
+        ReaderWay way = new ReaderWay(0L);
+        way.setTag("highway", "track");
+        assertEquals(0, priority(way), 0.01);
+    }
+
+    @Test
     public void testBusway() {
         ReaderWay way = new ReaderWay(0L);
         way.setTag("highway", "busway");
--- a/web/src/test/java/com/graphhopper/application/resources/RouteResourceCustomModelTest.java
+++ b/web/src/test/java/com/graphhopper/application/resources/RouteResourceCustomModelTest.java
@@ -125,8 +125,8 @@ public class RouteResourceCustomModelTest {
         // the bus profile is a custom profile and we can use its CH preparation as long as we do not add a custom model
         String body = "{\"points\": [[11.58199, 50.0141], [11.5865, 50.0095]], \"profile\": \"bus\"}";
         JsonNode path = getPath(body);
-        assertEquals(1501, path.get("distance").asDouble(), 10);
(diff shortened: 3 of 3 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/ModeAccessParser.java
```diff
@@ -18,6 +19,25 @@ public class ModeAccessParser implements TagParser {
...
     private static final Set<String> INTENDED = Set.of("yes", "designated", "official", "permissive", "private", "permit");
+    static final Map<String, Map<String, String>> HIGHWAY_TYPE_DEFAULTS;
+    static final Map<String, Map<String, String>> BARRIER_TYPE_DEFAULTS;
+
+    static {
+        Map<String, Map<String, String>> m = new HashMap<>();
+        m.put("steps", Map.of("access", "no"));
+        m.put("footway", Map.of("motor_vehicle", "no"));
+        m.put("cycleway", Map.of("motor_vehicle", "no"));
+        m.put("pedestrian", Map.of("motor_vehicle", "no"));
```
easbar, line 30: Not sure about pedestrian highways, since they are often used by delivery vehicles outside of shopping hours for example.

michaz: True. And when you tag `motor_vehicle=yes` or (for better or worse) `motor_vehicle=delivery` on a `highway=pedestrian`, you get access from this class.

It doesn't know about conditionals (yet).

michaz: > you get access from this class

Though that may be unintended, because you also get it from `motor_vehicle=pupsaffe`

michaz: Unlike in `CarAccessParser`

michaz: > It doesn't know about conditionals (yet)

Though it looks like it's trying to

michaz: Ok now I *think* it behaves more or less like the main `CarAccessParser` except that `Car` is just a parameter, it's more or less fully data-driven (I think), `pedestrian` isn't a special case, it knows the meaning of `highway` tags as they relate to `access`, and perhaps with slightly different (and to be determined) defaults.

You can say what you are (`delivery`, `military`, ..), or what you are *not* (meant for previously *unknown* nots, i.e. if we found a tag `access=yes,bicycle=hippo` and we want to say we are not being a hippo, then `bike_access|block=hippo` accomplishes that, otherwise it would fall through as an *unknown* tag, i.e. ignored in either direction. But that's more for completeness I think.

And any time means every time (speaking of conditionals), and that's what still only half makes sense as a default, but it's what we currently need.

michaz: Up to details such as whether a car is supposed to drive on a completely unknown highway tag or not.

michaz: Not for production use.

PR #3338, opened 2026-04-07
fix #3337 (instructions): propagate major road names for unnamed link instructions 

### LINKED ISSUE #3337: Merged instruction loses main road name when unnamed *_link continues onto named major road
**Describe the bug**
When GraphHopper merges a short unnamed `*_link` edge (for example `motorway_link`) with the following named major-road segment into a single instruction, the merged instruction can lose the actual main road name/ref and fall back to a generic instruction such as `Turn left`.

This is not a routing geometry issue. The route itself is correct. The problem happens during instruction generation when an unnamed link immediately continues onto a named major road, but the name/ref of the following road is not propagated into the merged instruction.

**To Reproduce**

Real-world example:

* start: `35.192312,-114.06485`
* end: `34.476839,-114.318983`

Observed route behavior:

* unnamed `motorway_link` [way `172747746`](https://www.openstreetmap.org/way/172747746#map=17/35.188747/-114.067811)
* followed by named motorway [`Purple Heart Trail` / `I-40` (way `1416245111`)](https://www.openstreetmap.org/way/1416245111#map=15/35.18461/-114.06654)

Current instruction result:

* `Turn left`

Expected instruction result:

* `Turn left onto Purple Heart Trail (I-40)`

I also confirmed the same behavior in official GraphHopper routing, not only in my own deployment.

Additionally, I created a local regression test that reproduces the same class of issue in a minimal synthetic graph:

* instruction starts on an unnamed link edge
* same instruction continues onto a named non-link major road
* expected result is that the major road name/ref is used in the instruction

**Expected behavior**
When an instruction starts on an unnamed link road and continues on the same routed instruction onto a named major road, GraphHopper should carry over the following road name/ref into the instruction text instead of leaving it generic.

For example, instead of:

* `Turn left`

it should produce:

* `Turn left onto Purple Heart Trail (I-40)`

Image

The image on the left shows the current behavior.
The image on the right shows the behavior after the fix.

I already implemented a fix and added a regression test, and I will submit a pull request.

**System Information**

* OS: Ubuntu Linux
* JVM: Java 17
* GraphHopper: `12.0-SNAPSHOT` (local fork for testing)
* Reproduced both in my local fork and in official GraphHopper Maps routing behavior

**Screenshots & Logs**

GraphHopper Maps reproduction link:
`https://graphhopper.com/maps/?point=35.192312%2C-114.06485&point=34.476839%2C-114.318983&profile=car&layer=Omniscale`

Equivalent route coordinates:

* `35.192312,-114.06485`
* `34.476839,-114.318983`

gulbalasalamov:
Fixes [#3337](https://github.com/graphhopper/graphhopper/issues/3337)

## Summary

This PR fixes missing road names for instructions that start on unnamed link roads and continue onto named major roads.

Previously, GraphHopper could generate instructions like:

* `Turn left`

even though the merged instruction actually continued onto a named road such as:

* `Purple Heart Trail (I-40)`

## Root cause

The unnamed link and following major road were merged into a single instruction, but the major road name/ref was not propagated into the merged instruction.

## Fix

Instead of guessing from neighboring graph edges, this change performs deferred name fallback on the actual routed continuation:

* if an instruction starts on an unnamed link road,
* and the next edges remain part of the same instruction,
* and a named non-link major road is encountered,
* the instruction name is updated using that road's name/ref.

To avoid regressions for motorway destination-style instructions, the fallback is not applied when the instruction already carries destination-related information.

## Result

The instruction now changes from:

* `Turn left`

to:

* `Turn left onto Purple Heart Trail (I-40)`

for this class of cases.

## Testing

* Added a regression test in `InstructionListTest` for unnamed link -> named major road instruction generation.
* Verified the change with local test runs, including a full successful Maven build/test run.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -36,6 +36,7 @@ Here is an overview:
  * fredao, translations 
  * gberaudo, improvements regarding elevation
  * GProbo, fixes like #2241
+ * gulbalasalamov, fix for instruction generation when unnamed link roads continue onto named major roads (#3337) 
  * HarelM, improvements regarding elevation
  * HelgeKrueger, modularization of javascript, #590
  * henningvs, doc improvements
--- a/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
+++ b/core/src/main/java/com/graphhopper/routing/InstructionsFromEdges.java
@@ -48,6 +48,13 @@ public class InstructionsFromEdges implements Path.EdgeVisitor {
     private final IntEncodedValue lanesEnc;
     private final DecimalEncodedValue maxSpeedEnc;
 
+    /**
+     * True when the current instruction started on an unnamed link road and we want to
+     * replace its blank name later with the first suitable named non-link major road
+     * encountered on the actual routed continuation of the same instruction.
+     */
+    private boolean prevInstructionNeedsNameFallback;
+
     /*
      * We need three points to make directions
      *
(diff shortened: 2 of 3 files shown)


michaz (approved):
Thanks!

PR #3342, opened 2026-04-20
Limit weight increase caused by priority factors

easbar:
Custom model allows setting arbitrarily small (but non-zero) priority factors. And even multiple ones can be applied to the same edge via different rules, making them even smaller. This increases weights to arbitrarily large values, even ones exceeding our (large limit) (#3297). At the same time these factors have no practical meaning: For example it makes no sense to accept a three day detour to avoid a bridge, but still use the bridge in case the detour is three days an half an hour.

Here I have limited the maximum weight penalty that can be applied by the priority factor to one million. That is the equivalent to ~28h (100k seconds). I cant think of any useful scenario where it would be required to increase the weight of a single edge more than that.

Performance seems unaffected by this change.


karussell (approved):
LGTM

Issue #2242, opened 2021-02-22
Custom weighting, area improvements


easbar:
See this forum discussion: https://discuss.graphhopper.com/t/custom-routing-with-areas-to-block-road/6328/19

- [ ] allow 'directed' areas, e.g. to block roads in only one direction, maybe using something like a heading parameter
- [ ] allow circles instead of just polygons

PR #3343, opened 2026-04-21
Custom modes: Add property for direction of edge

Nakaner:
I added this feature in a GraphHopper fork I maintain for a client (they use GraphHopper with HERE data). I thought that it might be useful for the public.

This pull request adds support for `is_forward` in custom models. This property allows you to block an edge in one direction only.

For example, if you import OSM data with the encoded value `osm_way_id`, you can send route requests like this:

```json
{
  "points": [
    [8.483888, 48.944448],
    [8.480776, 48.944547]
  ],
  "profile": "car",
  "custom_model": {
    "distance_influence": 15,
    "priority": [
      {
        "if": "osm_way_id == 1091253952 && is_forward",
        "multiply_by": "0.0"
      }
    ]
  }
}

```

This will make OSM way 1091253952 unpassable in the direction of the edge from base to adjacent node (which is the direction of the OSM way). Note: OSM way IDs are not stable. If a way is split, parts of the way get a new ID.

It can be also used to block a road in one direction only with a tiny no-go area (example see GraphHopperTest) in order to model a temporary traffic restriction.

I am happy to add more tests if this feature is appreciated by you.

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomModelParser.java
+++ b/core/src/main/java/com/graphhopper/routing/weighting/custom/CustomModelParser.java
@@ -48,6 +48,7 @@ public class CustomModelParser {
     static final String PREV_PREFIX = "prev_";
     static final String CHANGE_ANGLE = "change_angle";
     static final String STREET_NAME = "street_name";
+    static final String IS_FORWARD = "is_forward";
     private static final boolean JANINO_DEBUG = Boolean.getBoolean(Scanner.SYSTEM_PROPERTY_SOURCE_DEBUGGING_ENABLE);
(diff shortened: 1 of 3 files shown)


karussell:
Regarding the implementation: is this basically returning the stored direction like `!edge.get(EdgeIteratorState.REVERSE_STATE)` ?

If yes, I'm unsure what the use case would be with OSM data. Also if you had a custom data import, why would you need this in the custom model and not put the direct-dependent value in the data directly?

> It can be also used to block a road in one direction only with a tiny no-go area (example see GraphHopperTest) in order to model a temporary traffic restriction.

But only if you know the direction of the OSM way, which is a bit of a drawback unlike the proposed "heading" in #2242.

PR #3350, opened 2026-05-05
RAM DataAccess: fix length field

easbar:


easbar:
Claude's review:


  AbstractDataAccess.writeHeader(file, length, segmentSize) stores length so the loader can recover the file size. The convention — set by MMapDataAccess.flush writing raFile.length(), and assumed by RAMDataAccess.loadExisting / RAMIntDataAccess.loadExisting doing readHeader(raFile) - HEADER_OFFSET — is that the stored value is HEADER_OFFSET + data.

  The RAM writers were storing only getCapacity() (data, no HEADER_OFFSET).

  - Broken invariant. The header field no longer matched the file length, so any reader relying on "stored length == file length" got wrong data — silently off by HEADER_OFFSET.
  - Latent landmine. Same-format RAM→RAM and cross-format RAM↔MMap reload happened to round-trip correctly only because HEADER_OFFSET (100) < SEGMENT_SIZE_MIN (128), so the loader's +1-on-non-zero-remainder rounding compensated for the missing offset. Any future change that raised HEADER_OFFSET to ≥ SEGMENT_SIZE_MIN would have silently dropped the
  last segment on reload.
  - Empty-data divergence. After trimTo(0); flush(), the buggy writer stores 0; the loader computes byteCount = -HEADER_OFFSET, hits the < 0 branch, and reports the file as missing — whereas MMap correctly reloads it as empty.

karussell (approved):
LGTM

PR #3353, opened 2026-05-16
Add custom model for ambulances

ge65cer:
This PR proposes a navigation profile for emergency ambulances, based on data from south-western Germany.

https://im.baden-wuerttemberg.de/fileadmin/redaktion/m-im/intern/dateien/pdf/20260324_Parametrierung_f%C3%BCr_die_georeferenzierte_Disposition.pdf

### DIFF
--- a/core/src/main/resources/com/graphhopper/custom_models/ambulance.json
+++ b/core/src/main/resources/com/graphhopper/custom_models/ambulance.json
@@ -1,0 +1,30 @@
+// to use this custom model you need to set the following option in the config.yml
+// graph.encoded_values: car_access, car_average_speed, max_speed, ferry_speed
+// profiles:
+//    - name: car_ambulance
+//      turn_costs:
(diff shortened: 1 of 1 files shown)


karussell:
Thank you! Also for providing the reference documentation. However, we would like to integrate a new profile only if it is used in real world. Is this already the case?

ge65cer:
@karussell Thanks for your question. The proposed model is based upon the documentation, the parameters described therein are at this time mandatory for all dispatch centers in Baden-Wuerttemberg. 

The profile does currently not need to be exposed to the front-end, but I would still like to commit it to the source make it findable with the goal of sparking future discussions and refinements with the expert community (as an [area of active research](https://ojs.iscram.org/index.php/Proceedings/article/view/238)).

karussell:
Unfortunately I would not want to have it in the core if no real users (yet). You could link e.g. to your repo for your colleagues for the testing period.

PR #3364, opened 2026-06-12
upgrade java to 25

OlafFlebbeBosch:
JDK 17 is not included as default in recent linux distros any more.
JDK 21 should be considered battle tested and includes some upstream changes for memory mapping.

~~Pinning the jetty version to resolve CVE-2026-1605 since there is no new dropwizard version for now.~~

### DIFF
--- a/.github/workflows/publish-github-packages.yml
+++ b/.github/workflows/publish-github-packages.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-java@v5
         with:
-          java-version: 21
+          java-version: 25
           distribution: temurin
       - name: Cache Maven artifacts
         uses: actions/cache@v5
--- a/.github/workflows/publish-github-packages.yml
+++ b/.github/workflows/publish-github-packages.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-java@v5
         with:
-          java-version: 17
+          java-version: 21
           distribution: temurin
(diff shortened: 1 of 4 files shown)


otbutz:
If you adjust the POM, you need to update the workflows as well.

OlafFlebbeBosch:
Thanks, looking into, done

karussell:
We were thinking to upgrade to java 25 anyway (because of the foreign memory API).

Is the jetty pinning related to the java upgrade? So we could do this separately and later change everything to java 25.

OlafFlebbeBosch:
I create a seperate for the jetty pinning.

OlafFlebbeBosch:
Should I update the PR for 25, I only thought you might want to stay on the "slow" side .

karussell:
Lets move to JDK25. (a few bigger projects like lucene are also already on it)

otbutz:
Honestly, from version 9 onwards, Java upgrades are pretty much hassle-free.

karussell:
Documentation would also need an update + readme, I think.

OlafFlebbeBosch:
Found the version number in the Installation section. 

I cannot change the [Dockerfile](https://github.com/IsraelHikingMap/graphhopper-docker-image-push) , though I am about to create an [issue](https://github.com/IsraelHikingMap/graphhopper-docker-image-push/issues/48) there. 

What else is missing, controlled by this repository ?

karussell:
Thanks! I think there is appveyor config, but this was broken anyway and will fix it separately.
Let's see what is missing later on :)

PR #3366, opened 2026-06-17
Allow custom scoring of transition route length in map-matching

JerzyJerzu:
## Summary

Allows callers to customize how the route length is computed for comparison with linear distance in `MapMatching`'s transition probability scoring. Users can plug in a function via `setTransitionRouteLengthFunction(...)` which can respect the route weight and time; the default keeps the existing geometric scoring (raw `path.getDistance()`), so behaviour is unchanged byte-for-byte for users who don't opt in. 

  ## Motivation

Today, the transition probability for each candidate-pair is computed purely from geometric route distance:
``` java
  double transitionLogProbability = probabilities.transitionLogProbability(path.getDistance(), linearDistance);
```
This works well when candidate edges differ geometrically - but it breaks down when two candidates are geometrically similar, or when we want to add custom priorites.
For example:
Parallel tracks in a rail/tram network. In many rail networks tracks may hold `preferred_direction` tag rather than strict directional prohibitions (see https://wiki.openstreetmap.org/wiki/Key:railway:preferred_direction). This preferences could be recognized by a custom_model's priority multiplier, but a current map matching code would not respect that and in this case may very often return routes which use a wrong opposite track

Letting a caller plug in a function that incorporates routing metadata (weight, time) into the effective length passed to the HMM solves this, biasing the matcher toward preferred candidates without changing the underlying HMM machinery.

  ## Usage example
``` java
  mapMatching.setTransitionRouteLengthFunction((distance, timeMillis, weight) -> {
      if (timeMillis <= 0) return distance;
      return distance * weight / (timeMillis / 1000.0);
  });
```
For a custom_model with `distance_influence = 0` and `multiply_by: 0.5` on against-direction edges (rail's `preferred_direction` pattern), `weight/time = 2.0`, so the effective length is doubled and Viterbi prefers the preferred direction.

A similar gap exists for road profiles with prioritised road classes — e.g., a bus profile down-weighting service roads relative to tertiary ones, or a cyclist profile preferring cycle paths over shared lanes. The router uses these preferences to find paths, but the matcher discards them, so when geometrically similar candidates exist, it may snap to a deprioritised class even though the router would have avoided it. The same strategy hook applies; the formula shown below is rail-specific (assumes distance_influence = 0), but other formulas can be plugged in to fit each

License
This contribution is made under the Apache License 2.0, as outlined in CONTRIBUTING.md (https://github.com/graphhopper/graphhopper/blob/master/CONTRIBUTING.md). I've added myself to CONTRIBUTORS.md as required for first-time contributors.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -48,6 +48,7 @@ Here is an overview:
  * Janekdererste, GUI for public transit
  * jansoe, many improvements regarding A* algorithm, forcing direction, roundabouts etc
  * jansonhanson, general host config
+ * JerzyJerzu, TransitionRouteLengthFunction strategy hook for map-matching transition scoring
  * jessLryan, max elevation can now be negative
  * joe-akeem, improvements like #2158
  * JohannesPelzer, improved GPX information and various other things
--- a/map-matching/src/main/java/com/graphhopper/matching/MapMatching.java
+++ b/map-matching/src/main/java/com/graphhopper/matching/MapMatching.java
@@ -72,6 +72,8 @@ public class MapMatching {
     private double measurementErrorSigma = 10.0;
     private double transitionProbabilityBeta = 2.0;
     private double uTurnCost = 40.0;
+    private static final TransitionRouteLengthFunction GEOMETRIC_ROUTE_LENGTH = (distance, timeMillis, weight) -> distance;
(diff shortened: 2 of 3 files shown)


JerzyJerzu:
@karussell, could you please take a look at this PR when you have time? 🥺
I believe it complements #3284: that PR allows OSM tags to influence custom-model weighting, while this PR optionally carries the resulting route weight into map-matching transition scoring. This matters for geometrically similar candidates such as parallel tracks with railway:preferred_direction.

PR #3368, opened 2026-06-24
Netherlands HGV toll

otbutz:
The Netherlands is rolling out an HGV toll, which will come into effect on 01.07. With this change, by default, motorways will be classified as `Toll.HGV` if no tagging is present in OSM. While some provincial and municipal roads are affected too, the proportion is too low to include them by default, IMHO. This should rather be handled by explicit tagging.

[image](https://www.vrachtwagenheffing.nl/-/media/trucktol/website/hier-gaat-u-vrachtwagenheffing-betalen/nieuwe-wegenkaarten/truck-toll-road-map.pdf?rev=45d521bfce0143aba12f41a55de66f11&hash=0F5F2417BD636E9B34A516DF59A982F7)

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTollParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMTollParser.java
@@ -112,7 +112,7 @@ public class OSMTollParser implements TagParser {
                 else
                     return Toll.NO;
             }
-            case BEL, BLR, LUX, POL, SWE -> {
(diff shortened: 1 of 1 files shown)


otbutz:
OverpassTurbo query for validation purposes:

```
[out:json][timeout:60];
{{geocodeArea:Netherlands}}->.searchArea;
(
  way["highway"="motorway"](area.searchArea);
);
// print results
out geom;
```

image

karussell:
Thanks!

PR #3369, opened 2026-06-25
Avoid unnecessary detours for racingbike

karussell:
This is based on feedback from [racingbike enthusiasts](https://github.com/gpxstudio/gpx.studio/issues/325).

 * removes the penalty for primary roads
 * removes the preference for cycleway
 * removes inconsistencies (not adding residential to preferHighwayTags and no longer different speeds for unclassified+residential)
 * avoids roads with foot=yes (even if segregated)

/cc @ratrun 

Do later:

 * create separate collect method for racingbike to make things easier to understand and avoid still existing inconsistencies like with `if(way.hasTag("bicycle", "optional_sidepath"))` or pushing bike handling and also generic `class:bicycle` handling should be removed and the avoidSpeedLimit is no longer used too for racingbike.
 * lower penalty for reverse one-way roads (especially for France, Belgium and Netherlands where it is allowed by default for slow roads)
 * increase penalty for cycleways on the wrong side of the road

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
@@ -26,8 +26,7 @@ public abstract class BikeCommonPriorityParser implements TagParser {
     protected final Set<String> preferHighwayTags = new HashSet<>();
     protected final Map<String, PriorityCode> avoidHighwayTags = new HashMap<>();
     protected final DecimalEncodedValue priorityEnc;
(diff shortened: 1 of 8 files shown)


ratrun:
I'm fine with these changes! Anyhow, as this PR modifies the behavior substantially, I would like to point out that it is possible to get a similar result  compared to prior. Just apply the following custom rule on to a selected race-bike profile:
```
{
  "priority": [
    { "if": "true",  "multiply_by": "bike_priority" }
  ]
}
```
If one can afford the increased graph computation time, it might make sense to switch the behavior and activate the preference of bike paths only in city areas:
```
{
  "priority": [
    { "if": "urban_density == CITY",  "multiply_by": "bike_priority" }
  ]
}
```

Issue #3372, opened 2026-06-29
BikeCommonPriorityParser && operator precedence


OlafFlebbeBosch:
While reading the `BikeCommonPriorityParser` I noticed this line
[BikeCommonPriorityParser](https://github.com/graphhopper/graphhopper/blob/5d2137ef2b747b79171bee78b31904e92ea8311d/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java#L91)

```
if ("path".equals(highway) || "track".equals(highway) && isGoodSurface) {
  weightToPrioMap.put(100d, VERY_NICE);
...
```

Is that really intended? Since `&&` has higher precedence than `||`  , the Priority is `VERY_NICE` for any `path` regardless of its surface.

My guess it should have been 
```
if (("path".equals(highway) || "track".equals(highway)) && isGoodSurface) {
  weightToPrioMap.put(100d, VERY_NICE);
```

There seems no Unittest breaking?!

karussell:
I think it was intentionally changed for just the tracks: #3022 (/cc @ratrun, found via git blame)

Do you have a case where this should be changed?

OlafFlebbeBosch:
I was not able to find a spot in my surrounding where this really matters. 

So if it is intended, any objections for submitting a PR clarifying it with additional braces that it is intentionally ?

ratrun:
@OlafFlebbeBosch Good catch! Can you please provide the PR?

PR #3373, opened 2026-06-30
testing surface properties for designated bicycle ways

OlafFlebbeBosch:
While checking https://github.com/graphhopper/graphhopper/issues/3372 I created tests for various bike types and testing went  sideways: 

The order of the parenthesis doesn't make any difference for Normal and Racing bikes, because a different rule jumps in: Pushing Sections for bikeDesignated are  VERY_NICE 
https://github.com/graphhopper/graphhopper/blob/75e8e8290d493686a9ccd1c7a6e3e112b4dec696/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java#L139

Normal bikes declare "paths" to be pushing sections, which feels wrong to me, at least for germany. 
https://github.com/graphhopper/graphhopper/blob/75e8e8290d493686a9ccd1c7a6e3e112b4dec696/core/src/main/java/com/graphhopper/routing/util/parsers/BikePriorityParser.java#L18

For racing bikes there is a penalty for tracks not having good surface, while paths are pushing sections (but preferred) -- which may  be ok since the speed of a pushing section is very low.

Sorry for the patch being rather bulky, due to editorconfig wrapping/reformatting a lot of lines.

Not sure if you like that outcome @ratrun

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
@@ -40,6 +40,10 @@ public abstract class BikeCommonPriorityParser implements TagParser {
         addPushingSection("steps");
         addPushingSection("platform");
 
+        if (this instanceof RacingBikePriorityParser || this instanceof BikePriorityParser) {
+            addPushingSection("path");
+        }
+
         avoidHighwayTags.put("motorway", REACH_DESTINATION);
         avoidHighwayTags.put("motorway_link", REACH_DESTINATION);
         avoidHighwayTags.put("trunk", REACH_DESTINATION);
@@ -88,7 +92,7 @@ public abstract class BikeCommonPriorityParser implements TagParser {
         String highway = way.getTag("highway");
         if (bikeDesignated) {
             boolean isGoodSurface = way.getTag("tracktype", "").equals("grade1") || goodSurface.contains(way.getTag("surface", ""));
-            if ("path".equals(highway) || "track".equals(highway) && isGoodSurface)
+            if (("path".equals(highway) || "track".equals(highway)) && isGoodSurface)
                 weightToPrioMap.put(100d, VERY_NICE);
             else
                 weightToPrioMap.put(100d, PREFER);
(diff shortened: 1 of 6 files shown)


OlafFlebbeBosch:
Management Summary: the order of the expression is almost meaningless since there are other rules overriding the result anyway. I am ok with dropping the whole topic. 

But what is concerning me is "path" being pushings sections for regular bikes, if I am not mistaken,

karussell:
> Normal bikes declare "paths" to be pushing sections, which feels wrong to me, at least for germany.

highway=path is tricky without any tags and so this happens only if no access tags are present. [The wiki](https://wiki.openstreetmap.org/wiki/Tag:highway=path) already says:

> Without [access](https://wiki.openstreetmap.org/wiki/Key:access) and [surface](https://wiki.openstreetmap.org/wiki/Key:surface) tags, [routing](https://wiki.openstreetmap.org/wiki/Routing) engines may penalise the use of a path.

Maybe instead of two separate places where we handle path (pushingSections and this IF-statement) we should create a single place [similar to what we do for the speed parser](https://github.com/graphhopper/graphhopper/blob/75e8e8290d493686a9ccd1c7a6e3e112b4dec696/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java#L149).

OlafFlebbeBosch:
will work on moving the pushing sections code ... and yes "path" is a mixed bag.

OlafFlebbeBosch:
I moved the pushingSections to the Abstract Common class. 

I reverted most of the reformatting of unrelated lines since IMO since editorconfig in VS Code seems to fight over the maximum line length with the java formatter. 

Since I had to rebase I squashed all the commits. Makes it better readable IMO

karussell:
> I moved the pushingSections to the Abstract Common class.

Sorry, I meant a different "two places" and we shouldn't do it this way.

What I meant is the first place `if ("path".equals(highway) ...) weightToPrioMap.put(100d, VERY_NICE)` and the second place `else if (pushingSectionsHighways.contains(highway)...)` where path is handled.

I'm unsure how to fix that and probably it shouldn't be part of this PR?

OlafFlebbeBosch:
Ok, since the only change is for MountainBikes (because path is not a pushing sections) and I am not really sure if we want to increase priority for "concrete" surfaces ... I am now refraining from proposing to change anything. 

Only adding parenthesis to document the expression works as intended, removing the Tests since the Test are rather pointless, since they test many other things. 

So only cosmetic changes are included. Feel free to commit or cancel.

Issue #3298, opened 2026-03-02
FFM based DataAccess


otbutz:
Java 22 added a new API for fast on- and off-heap memory access. 

https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/foreign/MemorySegment.html

karussell:
When I last tested it it was even slightly slower. But I'm open to such changes if it gives us some advantages :)

otbutz:
I'm more interested in the off-heap scenario. FFM also adds better mmap access 😉   https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/nio/channels/FileChannel.html#map(java.nio.channels.FileChannel.MapMode,long,long,java.lang.foreign.Arena)

karussell:
Yes, I also meant mmap. But what does "better" mean in this context? Faster?

otbutz:
The old 2GB limit inherited by the MappedByteBuffer is gone. MemorySegment is apparently also more JIT friendly. I'd also expect that an FFM based off-heap DataAccess implementation lowers GC overhead.

karussell:
We need multiple segments even without the 2GB limit (to incrementally grow if needed on import). So staying under the 2GB limit is not really a disadvantage. However I can imagine that we could use multiple segments for import and a single one for optimized read speed when routing. Feel free to play around with this ;)

> I'd also expect that an FFM based off-heap DataAccess implementation lowers GC overhead.

Sure, would be nice to have.

But again, it isn't guaranteed to give us a speed up at the end. So without numbers this is not convincing as FFM has a minor but still existing overheap compared to simple arrays.

The only advantage I see is that we can avoid the unmapping hack for MMAP. And maybe the JDK removal of this unsafe stuff will at some point force us to migrate to FFM, but for now I do not see any advantages but I'm open to be convinced otherwise :) (but only perf tests can convince me atm ;) )

karussell:
Btw: see this branch here: https://github.com/graphhopper/graphhopper/tree/large_bytes

It seems that ForeignMemoryDataAccess might be slightly faster than existing DataAccesses (for our ordinary Measurement class, feel free to test it on your own). But the implementation uses only one segment and the speed up comes just from this and the segmented version is much slower. (but we need a segmented solution only for the import as the required memory grows and is not known before)

[...]

PR #3374, opened 2026-07-01
More DataAccess implementations

karussell:
E.g. based on the new foreign memory API: the off-heap based one is called NATIVE_STORE and the memory mapped one is now the default for MMAP. The old ByteBuffer based is MMAP_OLD.

This is the first PR that really requires JDK 25. (#3364)

Since commit e56acf45 ForeignMemoryDataAccess has the same speed (or faster) than RAM(Int)DataAccess for all cases except for the CH preparation, which is 5% slower. Also the import is only as fast as RAM(Int)DataAccess on Linux as we can grow the memory without copying (i.e. basically we only let the OS remap the pages).

Otherwise ForeignMemoryDataAccess would be nice as a default. And for an instance that only serves routing requests (no CH/LM prep, no OSM import) we can already recommend it as it is a bit faster and underlying code is much simpler. We'll probably use it soonish in production too.

to do later: check performance of MMapForeignMemoryDataAccess+readonly=true vs. MMapForeignReadOnlyDataAccess


ratrun (commented):
I'm just wondering what this is all about, but I do not see documentation or default config comment changes. Does this introduce new config possibilities to the `graph.dataaccess.default_type` config setting? If so, what are the advantages and disadvantages of these?
Thank you for answering this stupid question in advance.

karussell:
Sorry for the lack of info in the pull request - not a stupid question at all. And yes, this is about graph.dataaccess.default_type (or also used for elevation). See issue also #3298.

The different DataAccess implementations have all certain advantages and limitations:

Let's start with the RAMDataAccess which is implemented via byte[] (exists already in master, configured via RAM_STORE). 
 * heap-based memory
 * This class has to use multiple segments as one byte array can only have max INT_MAX entries and would not be sufficient for medium to large graphs. The segments also have a nice advantage that you can grow them step by step without knowing the end size. However, multiple segments lead to a (rather tiny) speed penalty because every access needs to determine the segment number and the index into it.
 * Also we use integers in many places and then we have to convert it forth/back to bytes

The MMapDataAccess is implemented using MappedByteBuffer (exists already in master, now configured via MMAP_OLD).
 * off-heap
 * This class needs multiple segments as well because unfortunately the ByteBuffer API is also limited to INT_MAX. 
 * The advantage of MMAP is that it uses off-heap memory and works with tiny heaps e.g. on smaller servers or even mobile devices.

The RAMIntDataAccess is like RAMDataAccess but implemented via int[]
 * heap
 * It is also segmented for the same reasons as RAMDataAccess
 * It it faster compared to RAMDataAccess as no forth/back conversation needs to happen to integer. However currently set/getBytes is not implemented and so it cannot be used in all circumstances. (And even if it would be implemented it would be a bit ugly and also slower in these cases compared to RAMDataAccess)

The new RAM1SegmentDataAccess is implemented via int[] but compared to RAMDataAccess has only a single array and so no overhead for the segment calculation exists. But of course is only suited for smaller objects, i.e. limited use and might get removed later.

The new RAMLongDataAccess is only implemented for testing purposes via a long[] array and probably is always worse than RAMIntDataAccess and RAMDataAccess, i.e. very limited use and might get removed later.

**Now to the two most interesting new classes.**

The new **ForeignMemoryDataAccess** is implemented via a single MemorySegment (the new foreign memory API) but is **off-heap memory**. It is rather similar to the [old Unsafe implementation](https://github.com/graphhopper/graphhopper/pull/1620). In many tests this seems to be the fastest for two reasons:
 * no overhead due to segments but can still grow step by step
 * it is not required to convert from/to bytes as a native cast is used under the hood

(There is also a new ForeignMemorySegmentedDataAccess which is used for testing purposes only and uses multiple segments to see the overhead of the segmentation)

The new **MMapForeignMemoryDataAccess** is also off-heap like ForeignMemoryDataAccess but with a file-backed storage. Under linux you can even use `/dev/shm´ as a tmp file-based storage which would make it as fast as the memory-only ForeignMemoryDataAccess. In theory this would be the only class which is required, but in practice other implementations like ForeignMemoryDataAccess are still faster on different systems. Still something we need to explore. So MMapForeignMemoryDataAccess should fully replace MMapDataAccess.

(There is also a new MMapForeignReadOnlyDataAccess which is suited if data exists already and might get slightly faster than MMapForeignMemoryDataAccess.)

otbutz:
I've yet to dive deeper into the topic, but can the FFM off-heap memory be allocated on (transparent) huge pages? For the regular Java heap and all the DataAccess implementations based on it, it is a simple as starting the JVM with ` -XX:+UseTransparentHugePages` or `-XX:UseLargePages`.

Most sources claim a performance boost between 5% and 15% due to lower TLB misses.

karussell:
I have no knowledge in this area tbh. Did you try these options and run some benchmarks? I tried but without success (or improvement).

I also asked the AI and it says 

> one could call mmap with MAP_ANONYMOUS | MAP_HUGETLB and pass the pointer to MemorySegment.ofAddress(addr).reinterpret(size, arena, cleanup) and register munmap as the cleanup action. 

> But first check if the workload is even TLB-bound: 
> 1. perf stat -e dTLB-load-misses,dTLB-loads,page-faults java ... during import and during a query benchmark.
> 2. Toggle echo always | sudo tee /sys/kernel/mm/transparent_hugepage/enabled vs never, rerun, and confirm the segments actually got huge pages via grep AnonHugePages /proc/<pid>/smaps_rollup.
>  3. Only if always shows a clear win and you can't rely on THP being enabled on target machines, ship the small madvise-only variant (~30 lines) — drop the MAP_HUGETLB path entirely.

So it seems still possible but requires code (?)
And you can always still use RAMDataAccess.

otbutz:
Looks like you need to invoke `madvise` using FFM. At least that's what Elastic does:

https://github.com/elastic/elasticsearch/blob/f1703ac7ee6e8116fc7067687e52e489c4be8452/libs/native/src/main/java/org/elasticsearch/nativeaccess/jdk/JdkPosixCLibrary.java#L61-L64

otbutz:
> Did you try these options and run some benchmarks?

We're currently running our GH instances using `-XX:+AlwaysPreTouch -XX:+UseTransparentHugePages` with the following sysfs knobs:

```
kernel/mm/transparent_hugepage/enabled = madvise
kernel/mm/transparent_hugepage/shmem_enabled = advise
kernel/mm/transparent_hugepage/defrag = defer
kernel/mm/transparent_hugepage/khugepaged/defrag = 1
```

We switched a about a year ago, and if memory serves me right, the performance gains were roughly in the region of 5–8%.

core/src/main/java/com/graphhopper/storage/ForeignMemoryDataAccess.java
```diff
@@ -50,6 +86,54 @@ public ForeignMemoryDataAccess(String name, String location, boolean store, int
...
+            // are released together with the rest of the mapping when the arena is closed:
+            //
+            // raw (from mmap, 4K-aligned)                                  raw + mapLength
+            //  v                                                            v
+            //  |-- head, up to ~2MB --|========== advisedLength ==========|-- tail --|
+            //                         ^
+            //                         aligned (first 2MB boundary >= raw)
+            long advisedLength = Math.ceilDiv(byteCount, HUGE_PAGE_SIZE) * HUGE_PAGE_SIZE;
+            long mapLength = advisedLength + HUGE_PAGE_SIZE;
+            MemorySegment raw = (MemorySegment) MMAP.invokeExact(MemorySegment.NULL, mapLength,
```
otbutz: Do we even need mmap here or couldn't we simply invoke madvise on the MemorySegment returned by `allocate(long byteSize, long byteAlignment)` (page aligned) ?

core/src/main/java/com/graphhopper/storage/ForeignMemoryDataAccess.java
```diff
@@ -40,6 +49,33 @@ public final class ForeignMemoryDataAccess extends AbstractDataAccess {
...
+    private static final MethodHandle MUNMAP = link("munmap", FunctionDescriptor.of(JAVA_INT, ADDRESS, JAVA_LONG));
+    private static final MethodHandle MADVISE = link("madvise", FunctionDescriptor.of(JAVA_INT, ADDRESS, JAVA_LONG, JAVA_INT));
+    private static final AtomicBoolean MADVISE_LOGGED = new AtomicBoolean();
+
+    private static MethodHandle link(String name, FunctionDescriptor descriptor) {
+        if (!System.getProperty("os.name", "").startsWith("Linux") || "false".equals(System.getProperty("graphhopper.thp")))
+            return null;
+        try {
+            Linker linker = Linker.nativeLinker();
+            return linker.downcallHandle(linker.defaultLookup().find(name).orElseThrow(), descriptor);
```
otbutz: We should also probe if the host supports THP, see https://man7.org/linux/man-pages/man2/madvise.2.html

> `madvise(0, 0, advice)` will return zero if `advice` is supported by the kernel and can be relied on to probe for support.

core/src/main/java/com/graphhopper/storage/ForeignMemoryDataAccess.java
```diff
@@ -40,6 +49,33 @@ public final class ForeignMemoryDataAccess extends AbstractDataAccess {
...
+    private static final long HUGE_PAGE_SIZE = 2L * 1024 * 1024;
+    private static final int MADV_HUGEPAGE = 14, PROT_READ = 1, PROT_WRITE = 2, MAP_PRIVATE = 2, MAP_ANONYMOUS = 0x20; // linux/mman.h
+    private static final MethodHandle MMAP = link("mmap",
+            FunctionDescriptor.of(ADDRESS, ADDRESS, JAVA_LONG, JAVA_INT, JAVA_INT, JAVA_INT, JAVA_LONG));
+    private static final MethodHandle MUNMAP = link("munmap", FunctionDescriptor.of(JAVA_INT, ADDRESS, JAVA_LONG));
+    private static final MethodHandle MADVISE = link("madvise", FunctionDescriptor.of(JAVA_INT, ADDRESS, JAVA_LONG, JAVA_INT));
+    private static final AtomicBoolean MADVISE_LOGGED = new AtomicBoolean();
+
+    private static MethodHandle link(String name, FunctionDescriptor descriptor) {
+        if (!System.getProperty("os.name", "").startsWith("Linux") || "false".equals(System.getProperty("graphhopper.thp")))
```
otbutz: Boolean.getBoolean() ?

core/src/main/java/com/graphhopper/storage/ForeignMemoryDataAccess.java
karussell: As I understand it, yes, we need mmap here. The JVM zero-initializes the memory which is the first touch of every page leading to 4K pages and so madvise would come after this, i.e. too late.

Also note that in my perf tests (Measurement.java) I couldn't see a benefit of THP. Neither for on-heap via JVM params (RAMDataAccess) nor for off-heap (ForeignMemoryDataAccess). With ForeignMemoryDataAccess, OSM import and off-heap THP it was even significantly slower, but I'm currently testing this further.

karussell: Maybe I did something wrong but I can't reproduce any speed up when enabling THP. Note that our default settings for the latest debian already was enabling THP but the JVM did not seem to use it (smaps_rollup -> AnonHugePages was 0). But it is also likely that I did not fully understand the setup and configured something wrong. But to include the code for enabling THP I would need a reproducer that justifies it (at least 5% speed up for import or routing).

otbutz: > Maybe I did something wrong but I can't reproduce any speed up when enabling THP. Note that our default settings for the latest debian already was enabling THP but the JVM did not seem to use it (smaps_rollup -> AnonHugePages was 0). 

You can use the JVMs built in logging to troubleshoot that:

```sh
java -XX:+UseTransparentHugePages -Xlog:gc+init,pagesize=info -version
```

On my local workstation:

```
[0.002s][info][pagesize] Explicit hugepage support:
[0.002s][info][pagesize]   hugepage size: 2M
[0.002s][info][pagesize]   hugepage size: 1G
[0.002s][info][pagesize]   default hugepage size: 2M
[0.002s][info][pagesize] Transparent hugepage (THP) support:
[0.002s][info][pagesize]   THP mode: madvise
[0.002s][info][pagesize]   THP pagesize: 2M
[0.002s][info][pagesize] Shared memory transparent hugepage (THP) support:
[0.002s][info][pagesize]   Shared memory THP mode: never
[0.002s][info][pagesize] UseLargePages=1, UseTransparentHugePages=1
[0.002s][info][pagesize] Large page support enabled. Usable page sizes: 4k, 2M. Default large page size: 2M.
[0.005s][info][pagesize] CodeHeap 'profiled nmethods':  min=4M max=116M base=0x0000751f64400000 size=116M page_size=2M
[0.005s][info][pagesize] CodeHeap 'non-nmethods':  min=4M max=6M base=0x0000751f6b800000 size=6M page_size=2M
[0.006s][info][pagesize] CodeHeap 'non-profiled nmethods':  min=4M max=118M base=0x0000751f6be00000 size=118M page_size=2M
[0.006s][info][gc,init ] CardTable entry size: 512
[0.006s][info][pagesize] Heap:  min=8M max=16000M base=0x0000000418000000 size=16000M page_size=2M
[0.006s][info][pagesize] Block Offset Table: req_size=32M req_page_size=2M base=0x0000751f62400000 size=32M page_size=2M
[0.006s][info][pagesize] Card Table: req_size=32M req_page_size=2M base=0x0000751f60400000 size=32M page_size=2M
[0.006s][info][pagesize] Mark Bitmap: req_size=250M req_page_size=2M base=0x0000751f50a00000 size=250M page_size=2M
[0.008s][info][gc,init ] Version: 25.0.2+10-LTS (release)
[0.008s][info][gc,init ] CPUs: 20 total, 20 available
[0.008s][info][gc,init ] Memory: 63985M
[0.008s][info][gc,init ] Large Page Support: Enabled (Transparent)
[0.008s][info][gc,init ] NUMA Support: Disabled
[0.008s][info][gc,init ] Compressed Oops: Enabled (Zero based)
[0.008s][info][gc,init ] Heap Region Size: 8M
[0.008s][info][gc,init ] Heap Min Capacity: 8M
[0.008s][info][gc,init ] Heap Initial Capacity: 1000M
[0.008s][info][gc,init ] Heap Max Capacity: 16000M
[0.008s][info][gc,init ] Pre-touch: Disabled
[0.008s][info][gc,init ] Parallel Workers: 15
[0.008s][info][gc,init ] Concurrent Workers: 4
[0.008s][info][gc,init ] Concurrent Refinement Workers: 15
[0.008s][info][gc,init ] Periodic GC: Disabled
openjdk version "25.0.2" 2026-01-20 LTS
OpenJDK Runtime Environment Temurin-25.0.2+10 (build 25.0.2+10-LTS)
OpenJDK 64-Bit Server VM Temurin-25.0.2+10 (build 25.0.2+10-LTS, mixed mode, sharing)
```

Note the `page_size=2M` in the `Heap` log. That should be enough to diagnose why the heap isn't using THPs.

As for off-heap, let's roll without THP for now. I'll try to wrap my head around it once i've time for it.

karussell: > You can use the JVMs built in logging to troubleshoot that:

I confirmed this already via smaps_rollup and the JVM used THP. But there might be other things I might have missed.

karussell: > Also note that in my perf tests (Measurement.java) I couldn't see a benefit of THP. 

If I enable THP for offheap I think I can see now a tiny 3% improvement, but only for CH requests. But I dont think this is worth the effort.

core/src/main/java/com/graphhopper/storage/GHDirectory.java
```diff
@@ -195,8 +216,8 @@ public DAType getDefaultType() {
      */
     public DAType getDefaultType(String dataAccess, boolean preferInts) {
         DAType type = getDefault(dataAccess, typeFallback);
-        if (preferInts && type.isInMemory())
-            return type.isStoring() ? RAM_INT_STORE : RAM_INT;
+//        if (preferInts && type.isInMemory())
+//            return type.isStoring() ? RAM_INT_STORE : RAM_INT;
         return type;
```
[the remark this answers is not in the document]
karussell: I think this is already fixed and you used an older version

PR #3375, opened 2026-07-03
Racingbike priority refactoring

karussell:
This change avoids the use of the BikeCommonPriorityParser.collect method as it is too different in too many places and not easy to adapt this. /cc @ratrun 

See [this commit](https://github.com/graphhopper/graphhopper/commit/d011c418a5012c6216390c1ff8dee250fd64530f) which shows the changes best (after copying everything from the super collect method).

This PR includes the following changes

 * The pushing section code for racingbike should not `PREFER` if `bicycle=yes` as it is inconsistent with the case `way.hasTag("foot", INTENDED)`. See e.g. [this route](https://graphhopper.com/maps/?point=51.17958%2C14.390842&point=51.179635%2C14.394708&profile=racingbike).
 * The case `preferHighwayTags.contains(highway)` should ignore `segregated` as we do for `way.hasTag("foot", INTENDED)`
 * no avoidSpeedLimit for racingbike
 * `class:bicycle:roadcycling` should not fallback to `class:bicycle`
 * put `highway=service`, `residential` and `unclassified` all into avoidHighwayTags instead of fixing the priority afterwards
 * removed case `if (way.hasTag("bicycle", "optional_sidepath"))`
 * `highway=path` and `track` should be handled in one place only
 * make highway=footway+bicycle=designated consistent with highway=cycleway+foot=yes

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonPriorityParser.java
@@ -94,23 +94,23 @@ public abstract class BikeCommonPriorityParser implements TagParser {
         }
 
         double maxSpeed = Math.max(OSMMaxSpeedParser.parseMaxSpeed(way, false), OSMMaxSpeedParser.parseMaxSpeed(way, true));
-        if ("cycleway".equals(highway) && preferHighwayTags.contains(highway)) {
+        boolean highSpeed = maxSpeed != MaxSpeed.MAXSPEED_MISSING && maxSpeed >= 71;
+        if ("cycleway".equals(highway)) {
             if (way.hasTag("foot", INTENDED) && !way.hasTag("segregated", "yes"))
                 weightToPrioMap.put(100d, PREFER);
             else
                 weightToPrioMap.put(100d, VERY_NICE);
         } else if (preferHighwayTags.contains(highway) || maxSpeed <= 30) {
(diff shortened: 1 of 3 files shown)


ratrun (commented):
I'm fine with these changes!

karussell:
There is still at least one suboptimal route [here](https://graphhopper.com/maps/?point=47.476849%2C14.988092&point=47.467949%2C14.988407&profile=racingbike) (even with the changes in this PR here) where we prefer the detour because maxspeed=30 although highway=unclassified is in the avoidHighwayTags.

Shall we remove the maxspeed=30 handling entirely for racingbike or shall we try to combine avoidHighwayTags+preferHighwayTags and have individual boosts or penalty for each different highway and then boost for maxspeed<=30 and minor penalty if foot=yes or tunnel?

ratrun:
Difficult to decide. But I would leave this as it is. From my perspective `highway=unclassified` are the best roads for racebikes, at least in Europe. According to the [wiki](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dunclassified) it says"Unlike [residential](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dresidential) and [track](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtrack), they are generally outside of residential areas and are not primarily used for agriculture."  Therefore, this example might be a tagging error. It probably better should be `residential`.

PR #3379, opened 2026-07-10
Support `maxweight:hgv`

otbutz:
Similar to `maxweightrating:hgv` there's also a HGV variant of `maxweight`. See [Taginfo](https://taginfo.openstreetmap.org/keys/maxweight%3Ahgv)


karussell:
Thanks!

PR #3380, opened 2026-07-10
use Arrays.binarySearch in GHLongLongBTree

karussell:
This change makes the OSM import 8% faster (the get method itself is around 30% faster).

It looks simple but was rather tricky to find. Due to special circumstances I found via AI that the import got faster when I replaced the following code in binarySearch:

```java
if (guessedKey < key) {
    low = guess;
} else {
    high = guess;
}
```

with
```java
if (guessedKey < key) {
   low = guess;
   continue;
}
high = guess;
```

Because then, the AI stated, the C2 compiler of the JVM no longer uses CMOV and instead is able to load multiple values in parallel (ie several iterations' loads are in flight at once). Making this method significantly faster for our case.

And it turns out the inbuilt Arrays.binarySearch is already optimized like this and so it can be used instead. It has an explicit range check which is not measurable (TODO wait for confirmation from benchmark).

### DIFF
--- a/core/src/main/java/com/graphhopper/coll/GHLongLongBTree.java
+++ b/core/src/main/java/com/graphhopper/coll/GHLongLongBTree.java
@@ -72,6 +72,25 @@ public class GHLongLongBTree implements LongLongMap {
         clear();
     }
 
+    // Like Arrays.binarySearch but without range check
+    private static int binarySearch(long[] a, int fromIndex, int toIndex, long key) {
+        int low = fromIndex;
+        int high = toIndex - 1;
+
+        while (low <= high) {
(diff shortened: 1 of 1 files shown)


easbar:
This is what I measured for the OSM import time (using maps from 2025, no elevation, car profile with turn costs and ignored footway,cycleway,path). 

| map     | master | this PR | 
|-------|-------------|----------------------|
| germany | 261.1    | 240.1  (−8.0%)    | 
| europe  | 2101.5   |  1857.2  (−11.6%)  |
| planet  | 6754.8   |  6108.3  (−9.6%)  |

PR #3384, opened 2026-07-18
Increase bike speed for surface=wood

### LINKED ISSUE #3383: Bicycle average speed on surface=wood is too slow
The average speed assumed for bicycle on surface=wood is much slower than people ride, leading to poor routing suggestions.

For example, in Ottawa-Gatineau, the Alexandra bridge is one of the best options for crossing the Ottawa River by bike. It has segregated bike lanes on a deck of wood planks, and is tagged in OSM as surface=wood. Cyclists usually seem to ride at roughly the same speed as they do on the cement and asphalt pathways at either end, which are also a little bumpy and shared with pedestrians. This includes both commuter bikes and road bikes. The main speed limitation is from interaction with pedestrians.

**To reproduce**

Request a route in Graphhopper Maps from one end of the Alexandra bridge to the other: https://graphhopper.com/maps/?point=45.431582%2C-75.708373&point=45.429042%2C-75.701005&profile=bike&layer=Omniscale

Graphhopper proposes turning around to avoid the bike path, and taking the road instead. If you switch the direction of travel, it will propose staying on the bike lane but claim the 650m trip will take 9 minutes. If you switch to the road bike profile, it will suggest going around to an entirely different bridge instead.

**Expected behaviour**
The proposed route should take the bike path, with a travel time of roughly 2 minutes.

**Screenshots**

Bike profile going backwards to get on the road:
The route on this map crosses the bridge, but does so by turning backwards for a section at each end

Road bike profile taking a large detour to avoid the Alexandra bridge:
The route on this map starts and ends on opposite ends of one bridge, but the route goes on 3 sides of a rough square, 3 times as long, to cross a different bridge instead

**Discussion**

The average speed for the various bike models has changed a few times:

|Bike   |MTB    |Racing |PR     |Date
|-------|-------|-------|-------|-------|
|10     |20     |2      |#150   |2014-02-05
|6      |10     |2      |#200   |2014-05-06
|4      |10     |2      |#2631  |2022-08-22
|4      |unset (4?)  |2      |#3214  |2025-08-24

However, there are no comments on why the change for surface=wood was made.

Also, d9fcbc5be5 (2019-11-05) marks 'wood' as unpaved, and 56a6620012 (2023-04-14) and 050d56a9ab (2023-04-22) order wood after ground. But the wiki categorizes wood as 'paved', and that makes sense to me as an improved, stable surface.

I don't know what kind of wood surface those low speeds are in response to. Why would the road bike speed be even less than PUSHING_SECTION_SPEED? Wooden slats are a little bumpy but not really any more than some rough asphalt or sidewalk concrete or paving stones.

I propose increasing the base average speed for all bike profiles to at least the speed for paving_stones (16), since the paving stone surface has pretty similar qualities to wooden boards: deliberately improved surface for wheels, flat, solid, has regular minor grooves between elements.

- **Bike**: 16 is just slightly slower than asphalt (18) and consistent with other hard but slightly grooved materials
- **MTB**: 16, same as bike, as the comment `// +4km/h on certain surfaces (max 16km/h) due to wide MTB tires` doesn't apply
- **Racing**: I think up to 20 (same as concrete:plates or concrete:lanes) would be justified, as smoothness can be applied for wood surfaces in worse condition

ratrun:
Fixes #3383

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/BikeCommonAverageSpeedParser.java
@@ -63,7 +63,7 @@ public abstract class BikeCommonAverageSpeedParser extends AbstractAverageSpeedP
         setSurfaceSpeed("pebblestone", 14);
         setSurfaceSpeed("salt", PUSHING_SECTION_SPEED);
         setSurfaceSpeed("sand", PUSHING_SECTION_SPEED);
-        setSurfaceSpeed("wood", PUSHING_SECTION_SPEED);
+        setSurfaceSpeed("wood", 16);
 
         setHighwaySpeed("steps", MIN_SPEED);
 
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikeAverageSpeedParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikeAverageSpeedParser.java
@@ -43,8 +43,6 @@ public class RacingBikeAverageSpeedParser extends BikeCommonAverageSpeedParser {
         setSurfaceSpeed("pebblestone", PUSHING_SECTION_SPEED);
         setSurfaceSpeed("salt", MIN_SPEED);
         setSurfaceSpeed("sand", MIN_SPEED);
-        setSurfaceSpeed("wood", 16);
-
         setHighwaySpeed("track", MIN_SPEED); // assume unpaved
 
         setHighwaySpeed("trunk", 24);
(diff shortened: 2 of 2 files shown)


karussell:
Why would we change the speed only for racingbike? I would indeed change the speed for all to 16? And then just wait until there is a real world case where avoiding the (potential wet) wood improves the route.

ratrun:
>  I would indeed change the speed for all to 16?

@karussell This is what I did! Please recheck the changes. In `BikeCommonAverageSpeedParser.java` for MTB and touring bike and in `RacingBikeAverageSpeedParser.java`.
Or am I missing something even more simple?

karussell:
ah, ok. Then we can just remove setSurfaceSpeed("wood", 16) in racingbike?

ratrun:
> ah, ok. Then we can just remove setSurfaceSpeed("wood", 16) in racingbike?

Oh! You are right, I improved it.

karussell:
Thanks!

PR #3385, opened 2026-07-22
Ignore unknown tag values in OSM parsing

karussell:
Currently unknown or incorrect values for restrictions confuse the tag parsers. E.g. for

```
way.setTag("highway", "service");
way.setTag("access", "no");
way.setTag("bicycle", "xyz");
```

The parser would not just ignore the bicycle tag `xyz` but would even ignore the `access=no`. This is rare, but still worth a fix and the fix is rather simple via a loop like we already do in ModeAccessParser.

We also should not have `unknown` in the restrictedValues (added [here](https://github.com/graphhopper/graphhopper/commit/5626d066dbe856653cb8c3aea00fab657e4e2998) because of the problem explained above). As a result a few ways [like this](https://www.openstreetmap.org/way/28841096) and barriers like [this](https://www.openstreetmap.org/node/2922058392) are no longer restricted.


ratrun (commented):
I like this improvement. 👏

PR #3386, opened 2026-07-24
Racingbike: simplify 'collect'

karussell:
After the collect method from racingbike is now decoupled I tried to simplify the code further and I like it.

wdyt @ratrun ?

Finally mainly three cases:

1. priority via simple highwayToPrio HashMap
2. "track", "cycleway" and "steps" exceptions
3. boost wider roads with bikeDesignated
4. (tunnel handling -> not so sure if really needed for racingbike)
5. (minor boost for primary if "maxSpeed <= 30". e.g. secondary is already at "slight prefer")

Minor behaviour change: removed shared_lane (no advantage for racingbike IMO) and also share_busway+opposite_track are not that good (not sure why we have this for normal bike?)

This change also fixes some wrong shortcuts/detours:

 * https://graphhopper.com/maps/?point=47.395673%2C15.160168&point=47.396893%2C15.164878&profile=racingbike
 * https://graphhopper.com/maps/?point=47.40608%2C15.262696&point=47.404806%2C15.25618&profile=racingbike
 * https://graphhopper.com/maps/?point=47.741086%2C15.310372&point=47.741103%2C15.307773&profile=racingbike
 * https://graphhopper.com/maps/?point=47.40979%2C15.274089&point=47.406505%2C15.269758&profile=racingbike

One remaining obstacle we can (later) improve: lower bike_network boost so that [certain detours](https://graphhopper.com/maps/?point=47.421552%2C15.273507&point=47.425199%2C15.271171&profile=racingbike) disappear. E.g. local+regional to 1.1 and (inter)national to 1.3

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
@@ -70,7 +70,9 @@ public class RacingBikePriorityParser extends BikeCommonPriorityParser {
             // tunnels are only dangerous on the high-speed roads that we strongly avoid anyway
             if (prio == BAD) prio = REACH_DESTINATION;
             else if (prio == SLIGHT_PREFER) prio = UNCHANGED;
-        } else if (maxSpeed <= 30 && !NARROW_WAYS.contains(highway) && prio.getValue() < SLIGHT_PREFER.getValue()) {
+        } else if (maxSpeed <= 30 && prio == UNCHANGED && !"cycleway".equals(highway)) {
+            // a slow but otherwise neutral road is pleasant, but this must not lift
+            // e.g. residential in a 30 zone above the parallel main road
             prio = SLIGHT_PREFER;
         }
 
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
@@ -5,7 +5,6 @@ import com.graphhopper.routing.ev.*;
 import com.graphhopper.routing.util.PriorityCode;
 
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -17,7 +16,7 @@ import static com.graphhopper.routing.util.parsers.AbstractAccessParser.INTENDED
 public class RacingBikePriorityParser extends BikeCommonPriorityParser {
 
     // usually too narrow for racing bikes, so e.g. bicycle=designated must not boost them
-    private static final Set<String> NARROW_WAYS = new HashSet<>(List.of("cycleway", "path", "footway", "pedestrian", "platform"));
+    private static final Set<String> NARROW_WAYS = Set.of("cycleway", "path", "footway", "pedestrian", "platform");
 
     private static final List<String> CYCLEWAY_KEYS = List.of("cycleway", "cycleway:left", "cycleway:both", "cycleway:right");
     private static final Set<String> CYCLEWAY_LANES = Set.of("lane", "shoulder");
@@ -54,7 +53,7 @@ public class RacingBikePriorityParser extends BikeCommonPriorityParser {
 
     @Override
     void collect(ReaderWay way, boolean bikeDesignated, TreeMap<Double, PriorityCode> weightToPrioMap) {
-        String highway = way.getTag("highway");
+        String highway = way.getTag("highway", "");
         double maxSpeed = Math.max(OSMMaxSpeedParser.parseMaxSpeed(way, false), OSMMaxSpeedParser.parseMaxSpeed(way, true));
         PriorityCode prio = highwayToPrio.getOrDefault(highway, UNCHANGED);
 
@@ -77,9 +76,8 @@ public class RacingBikePriorityParser extends BikeCommonPriorityParser {
             // tunnels are only dangerous on the high-speed roads that we strongly avoid anyway
             if (prio == BAD) prio = REACH_DESTINATION;
             else if (prio == SLIGHT_PREFER) prio = UNCHANGED;
-        } else if (maxSpeed <= 30 && prio == UNCHANGED && !"cycleway".equals(highway)) {
-            // a slow but otherwise neutral road is pleasant, but this must not lift
-            // e.g. residential in a 30 zone above the parallel main road
+        } else if (maxSpeed <= 30 && highway.startsWith("primary")) {
+            // a slow primary is as pleasant as a secondary
             prio = SLIGHT_PREFER;
         }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
```diff
@@ -4,101 +4,95 @@
...
 import static com.graphhopper.routing.util.PriorityCode.*;
 import static com.graphhopper.routing.util.parsers.AbstractAccessParser.INTENDED;
 
 public class RacingBikePriorityParser extends BikeCommonPriorityParser {
 
+    // usually too narrow for racing bikes, so e.g. bicycle=designated must not boost them
+    private static final Set<String> NARROW_WAYS = Set.of("cycleway", "path", "footway", "pedestrian", "platform");
+
+    private static final List<String> CYCLEWAY_KEYS = List.of("cycleway", "cycleway:left", "cycleway:both", "cycleway:right");
+    private static final Set<String> CYCLEWAY_LANES = Set.of("lane", "shoulder");
```
ratrun, line 22: Why is this a set and the CYCLEWAY_KEYS a list of strings? Please change into List as well.

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
```diff
@@ -4,101 +4,95 @@
...
-        preferHighwayTags.add("tertiary_link");
-
-        avoidHighwayTags.put("motorway", BAD);
-        avoidHighwayTags.put("motorway_link", BAD);
-        avoidHighwayTags.put("trunk", BAD);
-        avoidHighwayTags.put("trunk_link", BAD);
-        avoidHighwayTags.put("service", SLIGHT_AVOID);
-        avoidHighwayTags.put("residential", SLIGHT_AVOID);
-        avoidHighwayTags.put("unclassified", SLIGHT_AVOID);
+        highwayToPrio.put("road", SLIGHT_PREFER);
```
ratrun: Please change this into SLIGHT_AVOID.

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
```diff
@@ -4,101 +4,95 @@
...
+        highwayToPrio.put("secondary", SLIGHT_PREFER);
+        highwayToPrio.put("secondary_link", SLIGHT_PREFER);
+        highwayToPrio.put("tertiary", SLIGHT_PREFER);
+        highwayToPrio.put("tertiary_link", SLIGHT_PREFER);
+        highwayToPrio.put("service", SLIGHT_AVOID);
+        highwayToPrio.put("residential", SLIGHT_AVOID);
+        highwayToPrio.put("unclassified", SLIGHT_AVOID);
```
ratrun: Please change this into SLIGHT_PREFER. I remember that we had an incomplete discussion about a specific case where this is relevant. There this lead to leaving the `highway=PRIMARY` I believe it was. But according to the wiki definition these are usually very good for race cycling.

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
```diff
@@ -4,101 +4,95 @@
...
-            // do not overwrite if e.g. designated
-            weightToPrioMap.compute(100d, (key, existing) ->
-                    existing == null || existing.getValue() < prio.getValue() ? prio : existing
-            );
+            // We assume that humans are better in classifying preferences compared to our algorithm above,
+            // but do not degrade e.g. designated
+            PriorityCode classPrio = convertClassValueToPriority(classBicycleValue);
+            if (classPrio.getValue() > prio.getValue())
+                prio = classPrio;
         }
```
ratrun: I vaguely remember #3375 but I must confess I now do not understand the comment and the reasoning behind this additional 
`if (classPrio.getValue() > prio.getValue())`  
check any longer.

Suggestion: change into

``` 
       if (classBicycleValue != null) {
            // We assume that humans are better in classifying preferences compared to our algorithm above,
            prio = convertClassValueToPriority(classBicycleValue);
        }
```

core/src/test/java/com/graphhopper/routing/util/parsers/RacingBikeTagParserTest.java
ratrun, line 1: Please add following test at the end of `testClassBicycle()`:

`
        way.setTag("class:bicycle:roadcycling", "-3");
        assertPriority(REACH_DESTINATION, way);
`

This addition test is needed in order to verify that also negative values have a de-prioritization effect and tests the suggestion in the code

ratrun (commented):
Now I understand what you meant with elimination of the `weightToPrioMap` in `collect`. I thought that you wanted to get rid of it completely, also for bike and mountainbike. This would be nice, but would be much harder compared to `racebike` only.

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
karussell: hasTags requires a list (because of the order property), but in general I would use a set.
Similar to what we have in BikeCommonPriorityParser already.

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
karussell: Note that this PR did not change the code and also `unclassified` was never SLIGHT_PREFER.

In #3369 `unclassified` changed from UNCHANGED to SLIGHT_AVOID. I'm fine with reverting this (here or separately?), but I would argue against SLIGHT_PREFER :)

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
karussell: I see the problem for negative values. Note, the `if` was introduced in #3225.

Not sure about a tagging conflict or if a bike network matches the way (see comment in BikeCommonPriorityParser "`do not overwrite if e.g. designated`"). 

Maybe we should do:

```
 PriorityCode classPrio = convertClassValueToPriority(classBicycleValue);
 if (!bikeDesignated) prio = classPrio;
```

?

Not sure and probably this is all not too important for 2330 values world wide :)

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
ratrun: Thanks. Makes sense, but in this case also the CYCLEWAY_LANES should probably be better a List

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
ratrun: Thank you for applying this change. Now it seems to be a good compromise to me.

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
ratrun: I'm fine with this change now.

ratrun:
Thanks for the changes made. Now its ready for merge!

core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
karussell: Why? The order does not matter and no duplicates are needed/wanted.

ratrun: You are right. I didn't check thoroughly enough. Sorry.

PR #3387, opened 2026-07-28
Further deprioritize bicycle infrastructure for racebikes.

ratrun:
Improves remaining issues from https://github.com/gpxstudio/gpx.studio/issues/325#issuecomment-5096658894 

These are the mentioned cases:

http://localhost:8989/maps/?point=47.437422%2C15.282319&point=47.410845%2C15.273828&profile=racingbike&layer=OpenStreetMap Fixed
http://localhost:8989/maps/?point=47.466121%2C15.336426&point=47.474325%2C15.370978&profile=racingbike&layer=OpenStreetMap Already OK on master
http://localhost:8989/maps/?point=47.630049%2C15.658982&point=47.674609%2C15.489342&profile=racingbike&layer=OpenStreetMap Fixed
http://localhost:8989/maps/?point=47.40398%2C15.224011&point=47.381216%2C15.10702&profile=racingbike&layer=OpenStreetMap   Fixed
http://localhost:8989/maps/?point=47.289413%2C15.310959&point=47.280978%2C15.31001&profile=racingbike&layer=OpenStreetMap  Already OK on master
http://localhost:8989/maps/?point=47.052146%2C16.082349&point=47.065988%2C16.034605&profile=racingbike&layer=OpenStreetMap Fixed - identical to brouter now

Here is the brouter result for comparison:
https://brouter.de/brouter-web/#map=15/47.052146/16.082349/standard&lonlats=16.082349,47.052146;16.034605,47.065988&profile=fastbike

http://localhost:8989/maps/?point=47.073346%2C15.418422&point=47.019181%2C15.43863&profile=racingbike&layer=OpenStreetMap  Already identical to brouter with latest master, correct avoidance of bicycle=no

Here is the brouter result for comparison:
https://brouter.de/brouter-web/#map=15/47.073346/15.418422/standard&lonlats=15.418422,47.073346;15.43863,47.019181&profile=fastbike

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
@@ -38,15 +38,15 @@ public class RacingBikePriorityParser extends BikeCommonPriorityParser {
         highwayToPrio.put("unclassified", UNCHANGED);
         highwayToPrio.put("primary", UNCHANGED);
         highwayToPrio.put("primary_link", UNCHANGED);
-        highwayToPrio.put("cycleway", UNCHANGED);
+        highwayToPrio.put("cycleway", AVOID);
 
-        highwayToPrio.put("road", SLIGHT_AVOID);
-        highwayToPrio.put("service", SLIGHT_AVOID);
+        highwayToPrio.put("road", AVOID);
+        highwayToPrio.put("service", AVOID);
         highwayToPrio.put("residential", SLIGHT_AVOID);
-        highwayToPrio.put("path", SLIGHT_AVOID);
-        highwayToPrio.put("footway", SLIGHT_AVOID);
-        highwayToPrio.put("pedestrian", SLIGHT_AVOID);
-        highwayToPrio.put("platform", SLIGHT_AVOID);
+        highwayToPrio.put("path", AVOID);
+        highwayToPrio.put("footway", AVOID);
+        highwayToPrio.put("pedestrian", AVOID);
+        highwayToPrio.put("platform", AVOID);
         highwayToPrio.put("living_street", AVOID);
         highwayToPrio.put("bridleway", AVOID);
         highwayToPrio.put("track", AVOID_MORE);
@@ -76,7 +76,7 @@ public class RacingBikePriorityParser extends BikeCommonPriorityParser {
             prio = prio.better();
         } else if ("cycleway".equals(highway) && way.hasTag("foot", INTENDED)) {
             // too narrow when shared with pedestrians; wide roads keep their priority
-            prio = SLIGHT_AVOID;
+            prio = AVOID;
         } else if (way.hasTag("tunnel", INTENDED)) {
             // tunnels are only dangerous on the high-speed roads that we strongly avoid anyway
             if (prio == BAD) prio = REACH_DESTINATION;
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/RacingBikePriorityParser.java
@@ -38,7 +38,7 @@ public class RacingBikePriorityParser extends BikeCommonPriorityParser {
         highwayToPrio.put("unclassified", UNCHANGED);
         highwayToPrio.put("primary", UNCHANGED);
         highwayToPrio.put("primary_link", UNCHANGED);
-        highwayToPrio.put("cycleway", AVOID);
+        highwayToPrio.put("cycleway", SLIGHT_AVOID);
 
         highwayToPrio.put("road", AVOID);
         highwayToPrio.put("service", AVOID);
--- a/core/src/main/resources/com/graphhopper/custom_models/racingbike.json
+++ b/core/src/main/resources/com/graphhopper/custom_models/racingbike.json
@@ -8,8 +8,8 @@
 {
   "priority": [
     { "if": "true",  "multiply_by": "racingbike_priority" },
-    { "if": "bike_network == INTERNATIONAL || bike_network == NATIONAL",  "multiply_by": "1.4" },
-    { "else_if": "bike_network == REGIONAL || bike_network == LOCAL",  "multiply_by": "1.2" },
+    { "if": "bike_network == INTERNATIONAL || bike_network == NATIONAL",  "multiply_by": "1.2" },
+    { "else_if": "bike_network == REGIONAL || bike_network == LOCAL",  "multiply_by": "1.1" },
     { "if": "road_environment == FERRY", "multiply_by": "0.5" },
     { "if": "mtb_rating > 2",  "multiply_by": "0" },
     { "else_if": "mtb_rating == 2",  "multiply_by": "0.5" },
--- a/core/src/test/java/com/graphhopper/routing/util/parsers/BikeCustomModelTest.java
+++ b/core/src/test/java/com/graphhopper/routing/util/parsers/BikeCustomModelTest.java
@@ -200,16 +200,16 @@ public class BikeCustomModelTest {
         way.setTag("highway", "path");
         EdgeIteratorState edge = createEdge(way);
         CustomWeighting.Parameters p = CustomModelParser.createWeightingParameters(cm, em);
-        assertEquals(0.9, p.getEdgeToPriorityMapping().get(edge, false), 0.01);
+        assertEquals(0.8, p.getEdgeToPriorityMapping().get(edge, false), 0.01);
         assertEquals(6.0, p.getEdgeToSpeedMapping().get(edge, false), 0.01);
 
         way.setTag("mtb:scale", "0");
         edge = createEdge(way);
-        assertEquals(0.9, p.getEdgeToPriorityMapping().get(edge, false), 0.01);
+        assertEquals(0.8, p.getEdgeToPriorityMapping().get(edge, false), 0.01);
 
         way.setTag("mtb:scale", "1");
         edge = createEdge(way);
-        assertEquals(0.45, p.getEdgeToPriorityMapping().get(edge, false), 0.01);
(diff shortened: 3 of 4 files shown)


bblacher:
Thanks for continuing to work on this! Is there a public instance to experiment with this?

ratrun:
> Thanks for continuing to work on this! Is there a public instance to experiment with this?

Sorry, no.

ratrun:
@karussell : Probably I should have explicitly pointed out that it is mainly [this example](https://graphhopper.com/maps/?point=47.052146%2C16.082349&point=47.065988%2C16.034605&profile=racingbike&layer=OpenStreetMap) in comparison with [brouter](https://brouter.de/brouter-web/#map=15/47.0606/16.0607/standard&lonlats=16.082349,47.052146;16.034605,47.065988&profile=fastbike) which motivated me for reducing the priority of cycleway like cases by one more step.

karussell:
I think the change in core/src/main/resources/com/graphhopper/custom_models/racingbike.json is the most important. I'm unsure if the other changes are a good idea. Which examples require the further changes?

ratrun:
> Which examples require the further changes?

I don't remember the results of the other examples. I would need to check these once again. But before that please investigate the [example provided above two weeks ago](https://github.com/graphhopper/graphhopper/pull/3387#issuecomment-5133812155). Please check the situation in the east of this example specifically. I believe that it is an excellent case to discuss the additional changes.

karussell:
I'm unsure what I should see. But somehow this change does IMO improve the example you mention:



I'm also unsure if the path brouter suggests via the residential roads is really better with all the junctions/roundabouts and I thought residentials are suboptimal for racingbikes or at least not better than cycleways?

Maybe the use of residential is not 100% correct there near the industrial complex :) ? But I'm no tagging expert.

What I definitely agree are the changes to the racingbike.json regarding the decrease of bike network boost.

ratrun:
Yes, this is the spot where my additional changes are necessary. With these changes the result looks like this: 

grafik

I like the example because it forces you to think hard about which option you really want to take with a racebike:

1. The Grazerstraße `highway=primary` is blocked because of the [`bicycle=no`](https://www.openstreetmap.org/way/323763378) tag on a segment
2. The alongside `highway=cycleway` has all the downsides for a racebike like
- slow cyclists you need to take care of
- probably narrow such that overtaking is probably not that easy
- lots of junctions where you need to slow down and which are dangerous especially when you are faster than 25 km/h
- It breaks the "flow"
3. The alongside `highway=residential` has roundabouts which  
- doesn't break the flow that much
- you usually can keep your pace (depends on the traffic)

Theses reasons made me trigger to apply the additional changes.

@bblacher I would be interested on your opinion about what the users of gpxstudio would choose in this example

> Maybe the use of residential is not 100% correct there near the industrial complex :) ? But I'm no tagging expert.

You are right that the residential tag is probably not 100% correct in this case, but the arguments remain the same even if it would be a residential area.

bblacher:
The `bicycle=no` tag is actually wrong there, but the discussion is still valid.

I do think @ratrun 's example is better, especially since after first leaving the `highway=primary`, you are already on Industrieweg, in this case I would just stay there. I also agree about roundabouts being better for keeping the flow, if there are no cars, you can usually just go almost straight across, not having to reduce speed at all or only by a bit.

karussell:
@ratrun did you try with mapterhorn elevation data that we use in production? Using this elevation data and this branch I cannot reproduce the corrected behavior that you see.

This route detour also looks strange but this is probably the result of the bicycle=no tagging mistake.

ratrun:
> @ratrun did you try with mapterhorn elevation data that we use in production? Using this elevation data and this branch I cannot reproduce the corrected behavior that you see.

The reason that you now see a different result is [this recent change in OSM data](https://www.openstreetmap.org/changeset/186535601#map=19/47.058620/16.067975), where the `bicycle=no` was removed on one of the `highway=primary` segments.  

With todays geofabrik OSM Austria extract and my branch I get this result:

grafik

Unfortunately the example case I wanted to use as basis for discussion of this  PR is not actual any longer :cry: and will completely vanish once on the remaining segments the `bicycle=no`also gets removed from the OSM data.

I see following options now:

1. We keep discussing this case on the basis of OSM data I used for testing the PR. I could extract the OSM data from the past with overpass or cut out this data from my old OSM Austria data extract, which I still have and provide it somehow to @karussell for testing
2. I temporarily revert changeset 186535601 from above in the OSM data such that we can finish our discussion here
3. @karussell It could be that the screenshots provided so far in combination with the review of the changes convinced you to merge the PR without further discussion using this particular example

karussell:
I do not see the point to tune the priority towards a no-longer-existing example. Aren't there more of these or similar good examples?

btw: I'm already convinced that the bike_network boost needs the reduction you implemented here and I'm fine with AVOID for the platform/pedestrian involvement.

But I'm unsure about cycleway=AVOID instead of e.g. UNCHANGED or SLIGHT_AVOID. Any examples for this where it is necessary? (Same for the service=SLIGHT_AVOID -> AVOID change.)

ratrun:
> But I'm unsure about cycleway=AVOID instead of e.g. UNCHANGED or SLIGHT_AVOID. Any examples for this where it is necessary? (Same for the service=SLIGHT_AVOID -> AVOID change.)

The easiest example is a separated cycleway along a `highway=residential` with a starting point on the cycleway. 

I mean such an [example](https://graphhopper.com/maps/?point=47.060706%2C16.063066&point=47.062978%2C16.054208&profile=racingbike&layer=OpenStreetMap). Here currently the cycleway is choosen.

With a racebike you would do [this](https://graphhopper.com/maps/?point=47.060706%2C16.063066&point=47.060552%2C16.063025&point=47.062978%2C16.054208&profile=racingbike&layer=OpenStreetMap).

karussell:
This particular issue is already fixed with the bike_network changes in racing_bike.json. Do you have an example where cycleway=AVOID is required and cycleway=UNCHANGED (or SLIGHT_AVOID) is not sufficient?

ratrun:
Sorry, you are right. I didn't see that this cycleway is part of a bicycle route relation.

Here is a new [better example](https://graphhopper.com/maps/?point=48.080053%2C16.32785&point=48.08006%2C16.323816&profile=racingbike&layer=OpenStreetMap).

karussell:
Changing "cycleway"=UNCHANGED is still not required to fix this:



The change regarding "cycleway" && foot=INTENDED to `prio = AVOID` is sufficient.

I'm just a bit pedantic here because I'm not fully convinced cycleway should be equally bad as pedestrian/foot.

ratrun:
> I'm just a bit pedantic here because I'm not fully convinced cycleway should be equally bad as pedestrian/foot.

Pedantic is fine for a review :smile: . And now I believe I understand your concerns. I think that with the racebike you still want to take the residential if in this example the cycleway was not also intended for foot. It is not only pedestrians which make you slow down, but also slow cyclists like kids.

For me the question now is if you mean that I should prioritize footway, pedestrian , platform, living_street and bridleway down to AVOID_MORE?

karussell:
> For me the question now is if you mean that I should prioritize footway

No, I just want to reduce changes to a minimum as every change could cause something unexpected :)

> It is not only pedestrians which make you slow down, but also slow cyclists like kids.

Sure, but cycleway is IMO still slightly better :)

What about assigning cycleway to "slight_avoid" and if in future there are examples where "avoid" would improve the routing, than we'll change it later on?

ratrun:
> What about assigning cycleway to "slight_avoid" and if in future there are examples where "avoid" would improve the routing, than we'll change it later on?

I'm still not convinced, but this small detail shall not block the rest of the PR. Therefore I changed it in 7b750a3 as requested.

karussell:
Ok, thanks :)

karussell:
The routes you show above work all fine now.

Except [this one](https://graphhopper.com/maps/?point=47.399428%2C15.176449&point=47.395915%2C15.161116&profile=racingbike). But all segments have racingbike_prioity=1 and so usually the shorter route should be picked but isn't because of the elevation data. There are two problems here: the tunnel interpolation has a bug and would avoid this problem. But also, independently, a non-step function for elevation slowdown (instead of bike_elevation.json) would likely also fix this bug (and a couple of other problems). The first fix will come soon and the second fix is already in the works but will take longer...

ratrun:
> Except [this one](https://graphhopper.com/maps/?point=47.399428%2C15.176449&point=47.395915%2C15.161116&profile=racingbike). But all segments have racingbike_prioity=1 and so usually the shorter route should be picked but isn't because of the elevation data. There are two problems here: the tunnel interpolation has a bug and would avoid this problem. But also, independently, a non-step function for elevation slowdown (instead of bike_elevation.json) would likely also fix this bug (and a couple of other problems). The first fix will come soon and the second fix is already in the works but will take longer...

This is strange. The example works for me as well with the latest master branch, todays Austria geofabric extract and following elevation data configuration:
```
  graph.elevation.provider: pmtiles
  graph.elevation.pmtiles.location: /tmp/graphhoppertest/pmtiles/data/europe.pmtiles
  graph.elevation.pmtiles.zoom: 12
  graph.elevation.cache_dir: /tmp/graphhoppertest/pmtiles/cache
  graph.elevation.interpolate: bilinear
  graph.elevation.edge_smoothing: ramer
  graph.elevation.edge_smoothing.ramer.max_elevation: 2
  graph.elevation.long_edge_sampling_distance: 60
  graph.elevation.way_point_max_distance: 5
```

Here is my result:
grafik

karussell:
We use moving_average in production so a small difference might comes from this and boost the slope above a certain step in bike_elevation.json.

However, the recent commit should fix this elevation problem around the problem (the elevation smoothing turns out to be rather tough, but we are getting better :) ).

ratrun:
Now that you mentioned `bike_elevation.json` I detected that I wasn't using this for racebike.  With this I can reproduce the issue! Looking forward to your improvements!

karussell:
Should already work with the latest master - can you confirm :) ?

ratrun:
> Should already work with the latest master - can you confirm :) ?

Yes, confirmed! :+1:

PR #3388, opened 2026-08-08
fix: throw on elevation download IO failures instead of elev=0

### LINKED ISSUE #2867: CGIAR Elevation seems to be broken.
**Describe the bug**
When using the CGIARElevation provider, `0.0` will be returned for all queries.
This probably happens because the default base URL (https://srtm.csi.cgiar.org/) does time out now. (This is different from #2642). At least on my machine, this fails silently.
I don't know if this is only temporary or if this dataset has been removed.
**To Reproduce**

1. Create an instance of CGIARElevation.
2. Query the instance for any lat/long
3. 0.0 will be returned.
4. Look inside the cache folder. The files do exist, but they are only one 
**Expected behavior**
The correct elevation should be returned. Or an exception must be thrown.

**System Information**

MacOS,  OpenJDK 20.0.2, Graphopper 7.0

arimu1:
## Summary

Fixes #2867

When TIFF-based elevation providers (CGIAR, GMTED) failed to download a tile, `AbstractTiffElevationProvider` treated **any** `IOException` as “sea level” and returned `0.0`. That made timeouts / connection errors look like successful ocean tiles and could leave a useless in-memory cache entry.

This matches the intent called out by @otbutz on the issue (the catch around the download path) and aligns more closely with `AbstractSRTMElevationProvider`, which only treats missing tiles specially and fails loudly on other problems.

### Changes
- **`FileNotFoundException`** (HTTP 404 / missing tile, e.g. ocean): still sea level → `0` (unchanged behavior, covered by existing tests)
- **Other `IOException`s** (timeout after retries, connection reset, etc.): throw `IllegalStateException` with a clear message and cause
- Remove the failed tile from the in-memory cache so a later `getEle` can retry
- Re-throw `SocketTimeoutException` after retries as `IOException` instead of wrapping in `RuntimeException`

## Test plan

- [x] `CGIARProviderTest` / `GMTEDProviderTest` (Temurin 25): FileNotFound → 0; SocketTimeout and generic IOException → `IllegalStateException` with correct cause; no sea-level `.gh` left after failed download
- [x] `mvn -am -pl core -Dtest=CGIARProviderTest,GMTEDProviderTest -Dsurefire.failIfNoSpecifiedTests=false test` → BUILD SUCCESS

### DIFF
--- a/core/src/main/java/com/graphhopper/reader/dem/AbstractTiffElevationProvider.java
+++ b/core/src/main/java/com/graphhopper/reader/dem/AbstractTiffElevationProvider.java
@@ -23,6 +23,7 @@ import com.graphhopper.util.Downloader;
 import javax.net.ssl.SSLException;
 import java.awt.image.Raster;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.SocketTimeoutException;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


karussell:
Thanks! LGTM. @otbutz can you also have a quick look?

otbutz:
Sounds reasonable

PR #3396, opened 2026-08-31
Use continue on when the street name stays

AzazelSensei:
Continue and keep used onto even when the road name did not change.

I used on when the name stays the same, and onto when it changes. Turns are still onto.

Added continue_on and turn_on to en_US (other locales fall back to English).

#1287

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -9,6 +9,7 @@ Here is an overview:
  * andreaswolf, flag encoder versioning and more
  * andreylh, polygon for blocked area #1306
  * Anvoker, fixes like #1614 and helped with JUnit 5 migration #1632 
+ * AzazelSensei, continue on vs onto wording #1287
  * b3nn0, Android improvements
  * baumboi, path detail and landmark improvements
  * baybatu, improved error for EncodingManager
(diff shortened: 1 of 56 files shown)


karussell:
Thanks. Looks good.

Can you edit [our spreadsheet](https://github.com/graphhopper/graphhopper/blob/master/docs/core/translations.md), i.e. for now copy the continue_onto and keep_onto keys&values and modify it then for English (and rerun tests)? This way we keep all the other language translated until they decide to tune the translation for e.g. continue_onto.

AzazelSensei:
Copied continue_onto into continue_on and turn_onto into turn_on for every locale so the other languages keep the current wording. English still uses on. There's no keep_onto key — keep left/right goes through turn_onto.

I don't have access to the spreadsheet. TranslationMapTest and InstructionListTest passed.

karussell:
Please see the linked document where you do not edit the json files yourself. You edit the spreadsheet and run the script to update the json files. Otherwise the next update would remove these newly translations.

AzazelSensei:
Still can't edit the spreadsheet from here. If you add continue_on and turn_on (copy continue_onto / turn_onto; English continue on %1$s and %1$s on %2$s) I'll run update-translations.sh and drop the hand-edited locale files.

There's no keep_onto key — keep left/right with a street name goes through turn_onto / turn_on.

karussell:
> Still can't edit the spreadsheet from here.

Honest question, which AI model are you that you can't follow the link?

AzazelSensei:
I opened the sheet — I can view it, but I don't have edit access so I can't add rows. continue_onto and turn_onto are there; keep_onto isn't.

If you can grant edit access I'll add continue_on and turn_on and run update-translations.sh.

karussell:
Please do not waste our time with AI comments. Letting the AI help your code is one thing but communication should be done without (except maybe some translation)

Issue #215, opened 2014-05-29
Indoor routing


hieuhcmus:
Hi,

I am new to Graphhopper, I am building indoor routing.

Currently everything worked fine in a floor.
I was wondering how to use graphhopper to navigate/routing between floors?

I know that Graphhopper support elevation, can I use it?

Can someone give me advice.

Thanks

PR #3398, opened 2026-09-03
Add `level` encoded value.

natedx:
The goal is to improve indoor routing #215, seen through the prism of accessible pedestrian routing for the blind and low vision.

As preliminary step, I have added `level` as an encoded value.

## Discussion

There are many things to do in order to get better indoor routing.

1. We want to know what level we are on.
2. We want our waypoints to stick to the correct floor.
3. We want to know if we are going up or down stairs.
4. Some people might want to route across areas.

This PR focuses exclusively on 1.

### IntEncodedValue: Bad
I started out trying to do 1 and 3 together.
Using an IntEncodedValue seemed logical as levels are usually Integers (`level=1`).
However in the case of stairs or escalators, we get two values (`level=0;1`). If a staircase serves several floors, we can even get a range (`level=2-6`).
I briefly tried to fit both numbers into a single encoded value, by storing one for each direction, but it just doesn't make any sense.

Furthermore, with an Integer, it doesn't seem desirable to store only the first or second number, a default number or a NaN. 
E.g : setting `level=0` on a staircase going from 0 to 4th floor seems like information loss.

### DecimalEncodedValue : Good

Switching to a decimal encoded value takes a few more bits, but allows us to store regular integer values, and have a "good enough" value for things like stairs :

`decimalLevel = (startLevel + endLevel) / 2`

This method :  
- works great for routing details (for post-processing client side).
- works for Custom Models priorities
- would work in a future waypoint snapping LevelEdgeFilter, by setting the level to a given value, even though there might be rare ambiguous values (`level=0;2`and `level=0;4` parallel right next to each other ??) 

### No incline, will need another encoded value.

Unfortunately this encoded value does not encode the **incline** of such steps. My opinion is that trying to guess incline through levels would be an antipattern anyway, since many steps in the wild don't have levels associated to them, they just have the [`incline=up|down |...`](https://wiki.openstreetmap.org/wiki/Key:incline) tag.

## Next steps

This is my first contribution, so I keep it simple.
Next step is allowing waypoints to snap to a given level thanks to these encoded values, which already works on my fork.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -65,6 +65,7 @@ Here is an overview:
  * michaz, one of the core developers
  * mprins, improvements for travis CI and regarding JDK9 #806
  * msbarry, fixes like #1733
+ * n8dx, contributions for accessible pedestrian routing
  * nakaner, documentation
  * naser13, fixes like #1923
  * njanakiev, fixes like #1560
--- a/core/src/main/java/com/graphhopper/routing/ev/DefaultImportRegistry.java
+++ b/core/src/main/java/com/graphhopper/routing/ev/DefaultImportRegistry.java
@@ -172,6 +172,11 @@ public class DefaultImportRegistry implements ImportRegistry {
                     (lookup, props) -> new OSMCyclewayParser(
                             lookup.getEnumEncodedValue(Cycleway.KEY, Cycleway.class))
             );
+        else if (Level.KEY.equals(name))
+            return ImportUnit.create(name, props -> Level.create(),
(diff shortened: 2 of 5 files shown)


natedx:
As mentioned by @karussell (https://github.com/graphhopper/graphhopper/pull/3402#issuecomment-5574401724), the factor should be set to 0.5 in order to save some bits.

The best value is probably 8 bits, with factor 0.5, allowing us to store 512 total values, or 256 floors with half-floors.
E.g. we could store floors -80 to 176, with all half-floors.

PR #3401, opened 2026-09-07
Treat separate sidewalks like foot=use_sidepath

ratrun:
This PR contains an implementation as suggested in #3042 to fix the issue described there.

See this [test](http://localhost:8989/maps/?point=50.099099%2C8.760847&point=50.09977%2C8.765526&profile=foot&layer=OpenStreetMap).

### DIFF
--- a/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
+++ b/core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
@@ -30,6 +30,7 @@ public class OSMRoadAccessParser<T extends Enum> implements TagParser {
     private final List<String> restrictions;
     private final Function<String, T> valueFinder;
     private final RoadAccessDefaultHandler<T> roadAccessDefaultHandler;
+    private static Set<String> SIDEWALKS_NO_VALS = Set.of("no", "none", "separate");
 
     public OSMRoadAccessParser(EnumEncodedValue<T> accessEnc, List<String> restrictions,
                                RoadAccessDefaultHandler<T> roadAccessDefaultHandler,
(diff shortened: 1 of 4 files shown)


ratrun:
Now I added setting of `bike_road_access` value USE_SIDEPATH for separated cycleways in [bcbee36](https://github.com/graphhopper/graphhopper/pull/3401/commits/bcbee3672bcd96208c2a7db580ebcadea86f4467) for symmetry reasons. But here I am not sure, because the `bike_road_access` encoded value would need to have a separate value for the forward and reverse direction.

karussell:
> But here I am not sure, because the bike_road_access encoded value would need to have a separate value for the forward and reverse direction.

Yes, this is the current ugly limitation of XYRoadAccess. The idea in the future is to make it aware of two directions and remove xy_access encoded value.

core/src/main/resources/com/graphhopper/custom_models/foot.json
```diff
@@ -9,6 +9,7 @@
   "priority": [
     { "if": "!foot_access || hike_rating >= 2", "multiply_by": "0" },
     { "else": "", "multiply_by": "foot_priority"},
+    { "if": "foot_road_access == USE_SIDEPATH", "multiply_by": "0.9" },
```
karussell: Is the soft avoidance intended? Before we had a stronger avoidance (VERY_BAD). Maybe you have any real world examples for this?

update: [here is one](https://graphhopper.com/maps/?point=50.098865%2C8.760995&point=50.099584%2C8.765556&profile=foot) where it shouldn't use the main road

core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
```diff
@@ -300,10 +298,33 @@ public static OSMRoadAccessParser<RoadAccess> forCar(EnumEncodedValue<RoadAccess
...
+        return new OSMRoadAccessParser<>(roadAccessEnc, toOSMRestrictions(TransportationMode.BIKE), BIKE_HANDLER, BikeRoadAccess::find) {
+            @Override
+            public void handleWayTags(int edgeId, EdgeIntAccess edgeIntAccess, ReaderWay way, IntsRef relationFlags) {
+                if (way.hasTag("cycleway", "separate")
+                        || way.hasTag("cycleway:both", "separate")
+                        || (way.hasTag("cycleway:left", "separate") && way.hasTag("cycleway:right", SIDEWALKS_NO_VALS))
+                        || (way.hasTag("cycleway:right", "separate") && way.hasTag("cycleway:left", SIDEWALKS_NO_VALS)))
+                    accessEnc.setEnum(false, edgeId, edgeIntAccess, BikeRoadAccess.USE_SIDEPATH);
+                else
+                    super.handleWayTags(edgeId, edgeIntAccess, way, relationFlags);
```
karussell: Can you add a test for this? And do you have any real world examples for this?

karussell (commented):
Thanks! See my comments

core/src/main/resources/com/graphhopper/custom_models/foot.json
ratrun: > Is the soft avoidance intended? Before we had a stronger avoidance (VERY_BAD). Maybe you have any real world examples for this?

I just took the value you suggested [here](https://github.com/graphhopper/graphhopper/pull/3042#issuecomment-2500974858). Now changed into

```
    { "if": "foot_road_access == USE_SIDEPATH", "multiply_by": "0.3" },
```

ratrun:
> > But here I am not sure, because the bike_road_access encoded value would need to have a separate value for the forward and reverse direction.
> 
> Yes, this is the current ugly limitation of XYRoadAccess. The idea in the future is to make it aware of two directions and remove xy_access encoded value.

I tried to start with this one, but it is not trivial and needs further discussion. It probably also would require the possibility to assign two separate priority values in the custom rule for forwards and backward. I'm going to create a separate issue for it.

core/src/main/java/com/graphhopper/routing/util/parsers/OSMRoadAccessParser.java
ratrun: > Can you add a test for this? And do you have any real world examples for this?

No, not yet. As indicated above I was and I am still not sure if we should apply this change because of the issue about the direction. I made this change for bike just because of symmetry to foot. Bike is different to foot in the sense that one also needs to take the direction into regard. Do you think that this change makes sense now?

karussell: Ok, then lets remove the bike part for now an only apply it to foot.

ratrun: Done in [5b07e34](https://github.com/graphhopper/graphhopper/pull/3401/commits/5b07e341aaafb44186856e4a6cc2d94475b52a11)

ratrun:
Instead of creating a new issue started a discussion on the [graphhopper forum](https://discuss.graphhopper.com/t/bike-road-access-ev-should-be-direction-dependent/10022/2).

core/src/main/resources/com/graphhopper/custom_models/foot.json
karussell: Thanks for the link. I just tried the example and indeed the 0.9 is sufficient already. 

At one point there is a missing footpath as passage (probably just not mapped) and the resulting route with 0.9 looks like:



And then with 0.5 it does the longer detour:



If you are fine with it we can also use 0.5. Or is there a reason why it should be such a stronger?

Also can you use else_if instead of if in the custom model line of the foot_road_access == NO. Just for clarity.

ratrun: > If you are fine with it we can also use 0.5. Or is there a reason why it should be such a stronger?

I just implemented your suggestion from above with `VERY_BAD`. 
Now changed to 0.5 in [3f60465](https://github.com/graphhopper/graphhopper/pull/3401/commits/3f60465c69c5ffc30e4bba069889c5c312434da4) together with using `else_if`.

Issue #646, opened 2016-01-20
Indoor Routing


karussell:
Indoor routing is already possible in principle. But we could improve here:
 * read the indoor OSM data
 * provide a changed location lookup API to being able to specify the elevation too
 * improved turn instructions

TommasoDb:
> Indoor routing is already possible in principle

There is documentation or examples for this?

karussell:
With this I meant that we internally operate on a Graph regardless if it outside or inside of a building. So any test like GraphHopperStorageTest acts as an example here. The issues above are necessary for real world indoor routing.

kuchmiyalex:
@karussell Can you clarify how should LocationIndexTree/EdgeFilter be used to find closest match along with same 'level' tag? I have already added 'level' encoding into CarEncoder to put this info into graph.

karussell:
Inside the EdgeFilter you would do something like `return carFlagEncoder.getLevel(edge.getFlags()) == intendedLevel`, where the intendedLevel would be a property of this new EdgeFilter (pass via constructor) and the EdgeFilter is created per request and passed into the LocationIndex.findClosest method.

17631553844:
Excuse me, so there is no way to navigate indoors, including between floors?

17631553844:
Indoor navigation has been implemented using your graphhopper. Thank you very much, but I have a problem. I can't navigate between floors. emmm... I ‘m a Chinese

karussell:
Please use our forum for questions and discussion: https://discuss.graphhopper.com/

Davomat:
Hello, I have copied a functional indoor version of graphhopper [here](https://github.com/Davomat/graphhopper_indoor). It should be possible to add the changes to the original so that outdoor and indoor routing is possible.

However, the indoor version is based on a graphhopper commit from 2018 and I probably won't be able to fuse it with the current version. It would be great if there is someone who could help with or take on the task.

Davomat:
From an overview of the commit history it is mainly about floor level data. I see these overall steps:

- Level Catching: The OSMReader must capture the level data (not sure whether for nodes or edges). The level information must also be handed to various instances like BaseGraph, EdgeIterator, EdgeFilter, GH itself ... and be utilized there.
- Server: aesthetic changes in the web resources (html) as well as adjustments in .js files and GHServer to receive and process the level as a third coordinate
[...]

PR #3402, opened 2026-09-07
Allowing via routing to snap to levels.

natedx:
This is a follow-up PR related to #3398, and a big step towards solving #215.

This PR basically adds a new LevelEdgeFilter which uses the previously added `level` encoded value to snap any routing point (start, end, waypoint) to a given level.

GHRequest, ViaRouting and other files were updated to enable this behavior, mimicking what is done for the `heading` routing parameter, which is very close to the usage I was working toward.

Docs were updated as well, with an example.

I will leave this PR in draft until the previous #3398 PR is merged in order to see fewer "Files changed".

This will also give me the opportunity to add corrections, as this code has just been deployed to one of our GH instances for an indoor routing project, and we might see some obvious bugs.

### DIFF
--- a/client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
+++ b/client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
@@ -357,7 +357,7 @@ public class GraphHopperWeb {
             for (Double heading : ghRequest.getHeadings())
                 url += "&heading=" + heading;
         
-        if (ghRequest.getLevels().stream().anyMatch(h -> !Double.isNan(h)))
+        if (ghRequest.getLevels().stream().anyMatch(h -> !Double.isNaN(h)))
             for (Double level : ghRequest.getLevels())
                 url += "&level=" + level;
 
--- a/client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
+++ b/client-hc/src/main/java/com/graphhopper/api/GraphHopperWeb.java
@@ -261,6 +261,8 @@ public class GraphHopperWeb {
             requestJson.putArray("point_hints").addAll(createStringList(ghRequest.getPointHints()));
         if (!ghRequest.getHeadings().isEmpty())
             requestJson.putArray("headings").addAll(createDoubleList(ghRequest.getHeadings()));
+        if (!ghRequest.getLevels().isEmpty())
+            requestJson.putArray("levels").addAll(createDoubleList(ghRequest.getLevels()));
         if (!ghRequest.getCurbsides().isEmpty())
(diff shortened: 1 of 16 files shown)


karussell:
Thanks! Looks interesting.

Please note that the main issue is #646.

Also I would do it backwards: first add the highways and make this configurable (on/off). And only later change the snapping (via the level).

Minor: the level encoded value should probably get a factor of 0.5 ... this would reduce the number of required bits.

natedx:
Hello @karussell,

> Also I would do it backwards: first add the highways and make this configurable (on/off). And only later change the snapping (via the level).

I might have misunderstood you, but did you see that in my previous #3398 PR, I add the encoded value to the graph, which can be turned ON/OFF by adding the encoded value to the config.yml. 

Compared to #3398, the code in this draft PR only contains logic for snapping, via the level EV, and only if the api routing request contains a mention of level.

This PR still needs a bit of work, such as throwing errors if both "heading" and "levels" are requested (similarly to point hints), throwing an error if level snapping is requested but level encoded values haven't been added.

> Minor: the level encoded value should probably get a factor of 0.5 ... this would reduce the number of required bits.

Great idea, I didn't know you could do that ! Will explore that in #3398.

karussell:
Yes, I saw the other PR but again: we won't merge things that are not usable. So the first step would be to include indoor ways in the routing graph and then improve it by adding the snapping + level encoded value. 

> which can be turned ON/OFF by adding the encoded value to the config.yml.

Yes, of course this is optional. What I mean is the indoor routing. As I understand it you did not add this part. But once you do (which IMO should be the first step) it should be possible to disable it until it is mature enough.

natedx:
Ok, I understand your point now !

My experience with graphhopper has always been that indoor highways are included in the graph, unless you specifically filter out `indoor=yes` from the input OSM data. [Here's an example in your public instance](https://graphhopper.com/maps/?point=48.876194%2C2.324794&point=48.876183%2C2.325651&profile=foot&layer=Omniscale).

> So the first step would be to include indoor ways in the routing graph

The indoor ways are already in the graph by default, although they are currently indistinguishable from other footways. We could and probably should add a boolean "indoor" encoded value to be able to prevent pedestrian routing from taking shortcuts through buildings.

I'd be glad to do that too !

karussell:
> has always been that indoor highways are included in the graph

Thanks for the clarification. Not sure why I assumed they are something special ... good to know that they are just footways and this of course changes the approach and the snapping + level enc value is probably the right way to go forward. 

> We could and probably should add a boolean "indoor" encoded value to be able to prevent pedestrian routing from taking shortcuts through buildings

Yes, but lets keep them included as such an exclusion was not yet requested :)

Issue #2982, opened 2024-04-21
should locked=yes have lower priority?


karussell:
[This node](https://www.openstreetmap.org/node/10720988349) is tagged bicycle=yes and foot=yes and locked=yes. And we [cannot pass it](https://graphhopper.com/maps/?point=52.593502%2C13.249002&point=52.593465%2C13.249132&profile=foot).

Or is this incorrect tagging?

bt90:
An example from the OSM wiki:

![Example](https://wiki.openstreetmap.org/w/images/thumb/9/97/16052010022.JPG/1280px-16052010022.JPG)

It's locked, yet still passable for bikes and pedestrians.

bt90:
Seems like a tagging mistake to me:

https://maps.app.goo.gl/yw1dCcgVTrKYRFBo7

BURMAUG:
If bikes are allowed and pedestrians can walk freely they yes to both but not a vehicle.

caspg:
I asked question about this on forum https://discuss.graphhopper.com/t/barrier-lift-gate-is-blocking-bike/8398
I have similar node near me (Poland) which blocks routing through official cycling route.

I would vote for ignoring `locked=yes` if there is `[access]=yes`.

caspg:
I did this change in my fork to allow bicycles when there is `locked=yes`.  Maybe something like this could go to the main branch?

[commit](https://github.com/caspg/graphhopper/commit/d8eae9499cb5e7803d4eac88eee243328f275956)
```java
        if (restrictedValues.contains(firstValue))
            return true;
        else if (node.hasTag("locked", "yes") && !intendedValues.contains(firstValue))
            return true;
 ```

**Comment by ratrun** [comment]
I think that one also needs to take the transportation mode into regard. See https://github.com/graphhopper/graphhopper/compare/master...ratrun:biketourplanner:issue2982 for a draft

**Comment by caspg** [comment]
@ratrun Don't we already return `restrictionKeys` based on transportation mode?

https://github.com/graphhopper/graphhopper/blob/master/core/src/main/java/com/graphhopper/routing/util/parsers/AbstractAccessParser.java#L37

```java
restrictionKeys.addAll(OSMRoadAccessParser.toOSMRestrictions(transportationMode));
```

Then below lines will return `true` for `bicycle=yes` only for bike profile, right?

```java
String firstValue = node.getFirstValue(restrictionKeys);
intendedValues.contains(firstValue))
```

[...]

PR #3406, opened 2026-09-18
Respect locked barriers when ignoring private restrictions

ratrun:
Triggered by an [OSM forum discussion](https://community.openstreetmap.org/t/routing-through-gate/147753)  it was detected that the public graphhopper instance routes through barrier nodes tagged as `locked=yes` in combination with `access=private`. This only happens when `graph.encoded_values:` contains `foot_access|block_private=false`.

Here is the [example](https://graphhopper.com/maps/?point=49.566741%2C3.626259&point=49.565677%2C3.625968&profile=foot&layer=OpenStreetMap) which currently results in

image

although such a [barrier node](https://www.openstreetmap.org/node/1610339996#map=19/49.566520/3.626229) is contained along the result.

With this change the direct connection is avoided. Also see the previous [Graphhopper forum discussion](https://discuss.graphhopper.com/t/not-blocking-private-locked-yes-barrier-for-foot/10024/5).

Additionally I added `barrier=wall` to the always obeyed barrier list. According to the [OSM Wiki walls ](https://wiki.openstreetmap.org/wiki/DE:Tag:barrier%3Dwall) should be tagged as ways, but there are now more than 2000 occurrences for nodes.


karussell:
Reading [some previous discussion](https://github.com/graphhopper/graphhopper/issues/2982) I am really unsure how to handle locked=yes properly. Is the condition 
`&& !allowedValues.contains(firstValue)`
really correct? Shouldn't it be more like 
`&& !INTENDED.contains(firstValue))`
? Then no firstValue.equals("private") complexity is necessary and e.g. permit would be covered too.

Note with this locked=yes + access=destination becomes a barrier, which is fine (?)

ratrun:
> Note with this locked=yes + access=destination becomes a barrier, which is fine (?)

Your solution is simpler and according to [taginfo](https://taginfo.openstreetmap.org/tags/access%3Ddestination#combinations) the combination of `locked=yes` + `access=destination` appears only about 1000 times:

grafik

I agree with you that it is  impossible to guess the intention of the mapper for this combination.

Consequently I changed the implementation according to your suggestion and updated the test.

karussell:
Thanks!

